Issue #5, opened 2015-01-09
Introduce the build process


fredck:
The build process must be introduced.

fredck:
This has been prototyped here:
https://github.com/ckeditor/ckeditor5-design/tree/structure

fredck:
Fixed with #6.

Issue #4, opened 2015-01-09
Plugins path resolution


fredck:
The RequireJS "plugin" plugin must be created as well as the features to resolve CKEditor plugins paths.

fredck:
This is prototyped here:
https://github.com/ckeditor/ckeditor5-design/tree/structure

fredck:
Fixed with #6.

Issue #1, opened 2015-01-09
Introduce tests and Bender.js


fredck:
The infrastructure for testing must be created, including proper bender configuration.

PR #6, opened 2015-01-15
t/5: Introducing the CKEDITOR namespace, the AMD API and the Builder.

### LINKED ISSUE #2: Introduce the AMD API
The AMD API must be introduced according to [its prototype](https://github.com/ckeditor/ckeditor5-design/tree/structure).

fredck:
The proposed code means to close #2, #4 and #5.

I've pushed t/2 and t/4 as well, but I'm putting t/5 only under review because all those features are extremely related. They're actually defining the basics for our AMD approach. So it is simpler to review t/5 only and close all those issues at once.

There is a cross-requirement for this PR in ckeditor5-core:
https://github.com/ckeditor/ckeditor5-core/issues/1

Therefore to review this one, ckeditor5-core must be at t/1.

The builder is simply merging the main scripts for now. Other features are to be added in the future so a complete and real distribution version will be created. For now, the focus is creating an interchangeable build version of the root `ckeditor.js`.

No tests are included because it was not possible to bring the infrastructure necessary for the test environment (#1) without any API to get tested. Therefore, tests will be added later once this PR is masterised and #1 gets closed. A ticket will be opened as a followup to this PR.

The builder can be executed by simply running `grunt build`.


Reinmar:
Just a very quick check :smiling_imp: 

*  130b2c7 - would be good to fix these issues directly in the commits which introduced them, so the code is correct at every point.
* 255d2b5 - the comments should be either indented (and followed by a space) or outdented - not both at the same time.

dev/tasks/build.js
```diff
@@ -0,0 +1,158 @@
...
+		} else {
+			if ( done ) {
+				done();
+			}
+		}
+	}
+
+	function cleanup( callback ) {
+		var del = require( 'del' );
+		del.sync( target );
```
gregpabian: Why synchronously when you have a callback to call when ready?

dev/tasks/build.js
```diff
@@ -0,0 +1,158 @@
...
+
+		return callback();
+	}
+
+	function copyToTmp( callback ) {
+		var ncp = require( 'ncp' ).ncp;
+		var path = require( 'path' );
+		var fs = require( 'fs' );
+
+		var deps = JSON.parse( fs.readFileSync( 'package.json', 'utf8' ) ).dependencies;
```
gregpabian: You could use a plain `require` here:
```js
var deps = require('../../package.json').dependencies;
```

dev/tasks/build/start.frag
```diff
@@ -0,0 +1,13 @@
+( function ( root, factory ) {
+	// Register the CKEDITOR global.
+	root.CKEDITOR = factory();
```
gregpabian: Why do we pollute the global namespace even though someone wants to use CKEditor5 as an AMD module? That's completely against AMD principles...

dev/tasks/build.js
```diff
@@ -0,0 +1,158 @@
...
+			preserveLicenseComments: false,
+			wrap: {
+				startFile: [ 'dev/tasks/build/start.frag', require.resolve( 'almond' ) ],
+				endFile: 'dev/tasks/build/end.frag'
+			}
+		};
+
+		requirejs.optimize( config, function() {
+			callback();
+		} );
```
gregpabian: There's no need to wrap it in a function.

src/plugin.js
```diff
@@ -0,0 +1,36 @@
...
+		var path = name.split( '/' );
+		path.splice( 1, 0, 'src' );
+
+		if ( path.length === 2 ) {
+			path.push( path[ 0 ] );
+		}
+
+		path = '../../ckeditor5-plugin-' + path.join( '/' );
+
+		require( [ path ], function( value ) {
```
gregpabian: Could be shorter:
```js
require( [ path ], onload );
```

fredck:
@Reinmar:
> 130b2c7 - would be good to fix these issues directly in the commits which introduced them, so the code is correct at every point.

I think that, to properly handle PRs, we should avoid force-pushes so we'll have to live this as it is. In any case, in both points the code work. Those are just small details.

> 255d2b5 - the comments should be either indented (and followed by a space) or outdented - not both at the same time.

Ok, I've removed those comments as we should not really have them at this stage.

dev/tasks/build.js
fredck: Just to make the code simpler, as we have another call right after that. In any case we need to wait for you anyway, so it should not be a big del, right?

gregpabian: > should not be a big del

Right ^_^

dev/tasks/build/start.frag
fredck: I thought about that... but just the fact that `require` is available in the page doesn't guarantee that developers will be using CKEditor in the AMD way. They could still go with the namespace way. For example when copying pasting code from documentation or other sources.

For instance, developers may not even be developing with AMD at all, but maybe their CMSs are making `require` available and they don't know it. They'll try to use the CKEDITOR global then and have troubles.

Therefore, it sounded like the safest approach to enable both ways.

fredck:
@gregpabian, I've fixed all your remarks with the above commits.

dev/tasks/build/start.frag
gregpabian: Safe doesn't always mean right. They can still expose the global namespace themselves.

fredck: I understand, but you must understand that this level of flexibility is required because we deal with all kinds of developers out there. Some would be prepared for AMD but many will just have troubles.

Note that our goal is not making CKEditor an AMD module, but instead **enable** it as a AMD module to satisfy AMD devs.

For the sake of being "right" we may even consider a configuration option for the builder where one could force "pure AMD", so no global is created. But by default we should go the "satisfy all" way for easy adoption.

gregpabian: Yeah, having this configurable might do the trick.

fredck: Ok, I'll open an issue for this once this PR goes accepted.

dev/tasks/build.js
```diff
@@ -0,0 +1,146 @@
...
+
+			include: [ 'ckeditor' ],
+			stubModules: [ 'ckeditor-dev' ],
+
+//			optimize: 'none',
+			optimize: 'uglify2',
+			preserveLicenseComments: false,
+			wrap: {
+				startFile: [ 'dev/tasks/build/start.frag', require.resolve( 'almond' ) ],
+				endFile: 'dev/tasks/build/end.frag'
```
mlewand: I think we should put start and end fragments somewhere near the sources, not in `dev/tasks/build`. Eg. jQuery uses `src/intro.js`, `src/outro.js`, angular uses eg. `src/module.prefix`.

dev/tasks/build.js
```diff
@@ -0,0 +1,146 @@
...
+
+module.exports = function( grunt ) {
+	grunt.registerTask( 'build', 'Build a release out of the current development code.', function() {
+		var done = this.async();
+		module.exports.build( done );
+	} );
+};
+
+// Exports the build method so it can be used from plain node code as well.
+module.exports.build = function( done ) {
```
mlewand: Can't we create a class for the builder? And then wouldn't it be good to have unit tests for builder too?

It's not super complicated right now, but there's no doubt that it will grow, and tdd would be very handy.

dev/tasks/build.js
jodator: +1 from me because those are closer to cource code then to tasks source code/configuration.

Be careful with extension. The `.frag` is good because it can be white listed on jshint/jscss  and in IDE/text editor. Those files can be "invalid" like a function that is cut in half and put in two separate files.

dev/tasks/build.js
fredck: @mlewand 
> Can't we create a class for the builder?

Good suggestion. I've just done that with c3fc22c.

> And then wouldn't it be good to have unit tests for builder too?

I thought about that but I didn't want to spend time on this as there is no infrastructure for such tests in place. There are too many things to be bootstrapped right now, so I would propose to move ahead without tests here and make this a future enhancement.

dev/tasks/build.js
fredck: It makes sense. I've moved them with 76d64c6.

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,217 @@
...
+			var fs = require( 'fs' );
+			var tmp = this.tmp;
+
+			var deps = require( '../../../package.json' ).dependencies;
+
+			var toCopy = Object.keys( deps ).filter( function( name ) {
+				return name.indexOf( 'ckeditor5-' ) === 0;
+			} );
+
+			if ( !fs.existsSync( tmp ) ) {
```
Reinmar: Since the `tmp/` directory is removed at the end of the task we risk removing someone's directory (with all his uncommitted work :D). Even if this is only a theoretical situation, it's a good practice IMO to abort if a directory which should not exist already exists.

LICENSE.md
```diff
@@ -24,6 +24,15 @@ Where not otherwise indicated, all CKEditor content is authored by CKSource engi
 intellectual property. In some specific instances, CKEditor will incorporate work done by developers outside of CKSource
 with their express permission.
 
+(Ignore this line: %REMOVE_START%)
+
+Software available at our [repository](https://github.com/ckeditor/ckeditor5) and developer version only:
+
+ - RequireJS (lib/requirejs) <br>
```
Reinmar: Use `*` insteaf of `-`.

LICENSE.md
```diff
@@ -24,6 +24,15 @@ Where not otherwise indicated, all CKEditor content is authored by CKSource engi
...
+   Licensed under the terms of the MIT or new BSD license. <br>
+   Copyright (c) 2010-2014, The Dojo Foundation All Rights Reserved.
+
+(Ignore this line: %REMOVE_END%)
```
Reinmar: If you remove the first or last blank line inside the above block we will be able to add a blank line after it, so the next header will be visually separated from the contents of this section :P.

ckeditor.js
```diff
@@ -0,0 +1,105 @@
...
+
+	// Define a new "ckeditor" module, which overrides the core one with the above and the dev stuff.
+	define( 'ckeditor', [ 'ckeditor-core', 'ckeditor-dev', 'utils' ], function( core, dev, utils ) {
+		utils.extend( core, root.CKEDITOR, ( dev || {} ) );
+		root.CKEDITOR = core;
+
+		return core;
+	} );
+
+	function getBasePath() {
```
Reinmar: This function should be testable. We already had problem in CKEditor 4 that in order to test the base path we had to load multiple iframes.

ckeditor.js
```diff
@@ -0,0 +1,105 @@
...
+		if ( window.CKEDITOR_BASEPATH ) {
+			return window.CKEDITOR_BASEPATH;
+		}
+
+		var scripts = document.getElementsByTagName( 'script' );
```
Reinmar: Why so much spacing between variable declarations?

ckeditor.js
```diff
@@ -0,0 +1,105 @@
...
+
+( function( root ) {
+	var CKEDITOR = root.CKEDITOR = {
+		/**
+		 * The full URL for the CKEditor installation directory.
+		 *
+		 * It is possible to manually provide the base path by setting a global variable named `CKEDITOR_BASEPATH`. This
+		 * global variable must be set **before** the editor script loading.
+		 *
+		 *		alert( CKEDITOR.basePath ); // e.g. 'http://www.example.com/ckeditor/'
```
Reinmar: `console.log` will be more appropriate these days :D.

package.json
```diff
@@ -7,8 +7,15 @@
     "Editor",
     "WYSIWYG"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "ckeditor5-core": "ckeditor/ckeditor5-core"
```
Reinmar: Is it correct? What does the right side mean?

lib/requirejs/require.js
```diff
@@ -0,0 +1,36 @@
+/*
```
Reinmar: Wouldn't it be better to use a non-minified version of Require.JS for development purposes? Couldn't we use an npm version https://www.npmjs.com/package/requirejs?

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,217 @@
...
+			requirejs.optimize( config, callback );
+		},
+
+		/**
+		 * Deletes `tmp` directory.
+		 *
+		 * @param {Function} callback Function to be called when the task is done.
+		 * @returns {Object} The callback returned valued.
+		 */
+		cleanupTmp: function( callback ) {
```
Reinmar: It should be `cleanUpTmp`. Cleanup is a noun.

ckeditor.js
```diff
@@ -0,0 +1,105 @@
...
+
+			if ( match ) {
+				path = match[ 1 ];
+
+				return true;
+			}
+		} );
+
+		if ( path.indexOf( ':/' ) == -1 && path.slice( 0, 2 ) != '//' ) {
+			if ( path.indexOf( '/' ) === 0 ) {
```
Reinmar: `path[ 0 ] == '/'` would be shorter.

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * @class Builder
+ */
+module.exports = Builder = function( target ) {
+	/**
+	 * The target directory where to create the build.
+	 *
+	 * **Warning**: if existing, this directory will be deleted before processing.
+	 *
+	 * @type {string}
```
Reinmar: We don't use the `@type` tag.

> The use of this tag is deprecated. It's only supported for backwards compatibility with old ext-doc. For new code always use the @property tag instead.

https://github.com/senchalabs/jsduck/wiki/@type

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,217 @@
...
+	 *
+	 * All methods here MUST be called in the builder context by using
+	 * `builder.tasks.someMethod.call( builder, callback )`.
+	 */
+	tasks: {
+		/**
+		 * Deletes the `target` and `tmp` directories.
+		 *
+		 * @param {Function} callback Function to be called when the task is done.
+		 * @returns {Object} The callback returned valued.
```
Reinmar: Typo. Should be "value".

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,217 @@
...
+		 */
+		cleanup: function( callback ) {
```
Reinmar: The same as below - `cleanUp`.

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,217 @@
...
+				if ( !fs.existsSync( dest ) ) {
+					fs.mkdirSync( dest );
+				}
+
+				// Copy the "src" directory only.
+				ncp( path.join( 'node_modules', module, 'src' ), path.join( dest, 'src' ), {
+					dereference: true
+				}, function( err ) {
+					if ( err ) {
+						throw( err );
```
Reinmar: `throw` doesn't need `()`.

src/plugin.js
```diff
@@ -0,0 +1,34 @@
...
+// class.
+
+define( 'plugin', [ 'plugin-core' ], function( CorePlugin ) {
+	// Called when a "plugin!" module is to be loaded.
+	// http://requirejs.org/docs/plugins.html#apiload
+	CorePlugin.load = function( name, require, onload ) {
+		var path = name.split( '/' );
+		path.splice( 1, 0, 'src' );
+
+		if ( path.length === 2 ) {
```
Reinmar: Why `===`?

dev/tasks/build/builder.js
mlewand: Having a talk about doc tags we can also discus if we want to stick with JSDuck, see ckeditor/ckeditor5-design#27.

src/build/start.frag
```diff
@@ -0,0 +1,27 @@
...
+	// Register the CKEDITOR global.
+	root.CKEDITOR = factory();
+
+	// Make the build an AMD module.
+	if ( typeof define == 'function' && define.amd ) {
+		define( [], function() {
+			return root.CKEDITOR;
+		} );
+	}
+} )( this, function () {
```
Reinmar: There should be no space after `function`.

src/build/start.frag
```diff
@@ -0,0 +1,27 @@
...
+// This is the code that will precede the compiled source code in the ckeditor.js build.
+//
+// The following are the parts that compose the build:
+//
+//  * start.frag
+//  * Almond.js source code
+//  * CKEditor source code
+//  * end.frag
+
+( function ( root, factory ) {
```
Reinmar: There should be no space after `function`.

package.json
fredck: Pick the one you prefer:

A1: [RTFM](https://docs.npmjs.com/files/package.json#github-urls)! (just kidding... :D)

A2: That's the way to point npm to GitHub repos: `organization-name/repo-name` ;)

src/build/start.frag
```diff
@@ -0,0 +1,27 @@
...
+	// Register the CKEDITOR global.
+	root.CKEDITOR = factory();
+
+	// Make the build an AMD module.
+	if ( typeof define == 'function' && define.amd ) {
```
Reinmar: What does the `define.amd` check does?

fredck: That's the right way to check for AMD:
https://github.com/amdjs/amdjs-api/wiki/AMD#defineamd-property-

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,217 @@
...
+'use strict';
+
+var Builder;
+
+/**
+ * A CKEditor 5 release builder.
+ *
+ * @class Builder
+ */
+module.exports = Builder = function( target ) {
```
Reinmar: It's slightly better to use function declaration and then assign it to the `module.exports`, because it leaves the function named.

ckeditor.js
fredck: Fixed with  9a86ebb.

ckeditor.js
fredck: Fixed with 10246ac.

dev/tasks/build/builder.js
fredck: Fixed with 10246ac.

dev/tasks/build/builder.js
fredck: Fixed with 10246ac.

dev/tasks/build/builder.js
fredck: Fixed with 10246ac.

dev/tasks/build/builder.js
fredck: Fixed with 10246ac.

src/plugin.js
fredck: Fixed with 10246ac.

src/build/start.frag
fredck: Fixed with 10246ac.

src/build/start.frag
fredck: Fixed with 10246ac.

dev/tasks/build/builder.js
fredck: Fixed with 66297bb.

dev/tasks/build/builder.js
fredck: Fixed with 950871b.

ckeditor.js
fredck: Fixed with 3d4fbbe.

ckeditor.js
fredck: Fixed with 10246ac.

ckeditor.js
fredck: Fixed with 10246ac.

LICENSE.md
fredck: Fixed with 10246ac.

dev/tasks/build/builder.js
fredck: Good one... fixed with f4074f0.

LICENSE.md
fredck: Sorry, I didn't understand this one. Feel free to change it yourself.

package.json
Reinmar: Whoops. I actually checked the manual, but haven't seen this :D Too many options!

src/build/start.frag
Reinmar: Please add a comment in the code about this (the link will do)..

lib/requirejs/require.js
fredck: Non-minified: I don't think it makes much difference for us but we can do it at any moment.

npm: Unlike Almond, RequireJS is not needed for a strictly dev operation in this case (like build). It is required to simply running the editor after cloning the repo, so I would still opt to have it in our code.

src/build/start.frag
fredck: Er... you know what... this is much like RTFM :D

lib/requirejs/require.js
Reinmar: Non-minified: Perhaps it will be easier to debug some things. But you're right that we change it at any moment.

npm: I understand it as a dependency (not as a dev dependency). Just like CKEditor-core or plugins, it's needed to run a dev version of the project. But I would choose the npm also because of the maintenance - we won't need to manually upgrade the lib or write a grunt task for that. These are small things, but remembering about stuff like this (e.g. upgrading jQ) in CKEditor 4 is tiring.

BTW. do we anticipate having more libraries like this one?

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,236 @@
...
+	 *
+	 * @param {Function} [callback] Function to be called when build finishes. It receives `false` on error.
+	 */
+	build: function( callback ) {
+		var that = this;
+		var stepCounter = 0;
+
+		// Before starting, run the initial checkups.
+		if ( !this.checkUp() ) {
+			console.log( 'Build operation aborted.' );
```
Reinmar: You can use http://gruntjs.com/api/grunt.log - e.g. `grunt.error.log`.

BTW. I think I've read somewhere that using `console.log` in Node.JS for something more than debugging isn't good. @gregpabian: Do you confirm?

Reinmar: AFAIR it was because of the process output. If you use `proc.stdout` you can pipe or redirect this, while with `console.log` you can't.

gregpabian: Not really - console.log is just a wrapper around stdout.write so the logs will be redirected if we want to.

fredck: Additionally, the class is coded in a way that it is independent from grunt so it can use by pure node code as well.

Reinmar: So I remembered something wrong. Just one thing then - `console.error()` is IMO more appropriate.

lib/requirejs/require.js
fredck: You’re totally right. I said something stupid.

I’ve remove the RequireJS code with 49b17a2, getting rid of the lib directory (hurray!).

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,236 @@
...
+	this.tmp = 'build_tmp';
+
+	/**
+	 * The list of tasks to be executed by the `build()` method. Each entry is an Array containing the name of the
+	 * method inside `tasks` to execute and the message to show to the end user when executing it.
+	 *
+	 * @property {Array}
+	 */
+	this.taskList = [
+		[ 'cleanUp', 'Cleaning the "' + target + '" directory...' ],
```
Reinmar: Now (49b17a2) I've got:

```
Running "build" task
1. Cleaning the "undefined" directory...
```

BTW. I miss very much an information about where the build was created - where should I look for it now.

fredck: Fixed with 8880b20.

dev/tasks/build/builder.js
fredck: Fixed with 990f135.

src/build/start.frag
Reinmar: Well - you could save us 5 minutes of googling by a one comment. I think it's worth it. Besides that, first you need to know what manual you should be looking at, and for me this code was context-less.

src/build/start.frag
```diff
@@ -0,0 +1,27 @@
...
+//  * CKEditor source code
+//  * end.frag
+
+( function( root, factory ) {
+	// Register the CKEDITOR global.
+	root.CKEDITOR = factory();
+
+	// Make the build an AMD module.
+	if ( typeof define == 'function' && define.amd ) {
+		define( [], function() {
```
Reinmar: http://requirejs.org/docs/api.html#deffunc - the array seems to be unnecessary.

src/build/start.frag
fredck: Ok…ok… added this with 2c89d5d. I hope you like the commit message (signed, “the expert” :D)

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,238 @@
...
+		 * Removes the `CKEDITOR` namespace from AMD calls in the `tmp` copy of the source code.
+		 *
+		 * @param {Function} callback Function to be called when the task is done.
+		 * @returns {Object} The callback returned value.
+		 */
+		removeAmdNamespace: function( callback ) {
+			var replace = require( 'replace' );
+
+			replace( {
+				regex: /^\s*CKEDITOR\.(define|require)/mg,
```
Reinmar: Do we have to do this? Can't the minified code be using `define` and `require` through the `CKEDITOR` namespace?

fredck: This is necessary for r.js to find dependencies.

src/build/start.frag
fredck: Fixed with ef856f4.

dev/tasks/build/start.frag
Reinmar: I found an interesting thing - https://github.com/jrburke/requirejs/wiki/Updating-existing-libraries#anon

If I understand correctly, if `ckeditor.js` was concatenated with other scripts (e.g. by an assets pipeline like in RoR) it may lead to an error if we use an anonymous module definition (which we do now).

This means that there must be (somehow available) a totally non-AMD version of `ckeditor.js` available for situations like this - concatenation by some simple tool. Therefore at least the builder must have an option for that.

The remaining questions are:

1. What options the builder should have?
2. What do we want to ship in a release?

Ad 1. I think that 3 - AMD with polluting (could be the default), AMD without polluting (for pedants like many of us), no AMD at all.

Ad 2. jQuery is doing the "AMD with polluting", so I think we can do it as well.

Reinmar:
Hurray! The end of the review marathon :dancer: 

PS. Before merging please check 1a37506.

Reinmar:
I'll extract https://github.com/ckeditor/ckeditor5/pull/6#discussion-diff-23005728 to a new ticket.

dev/tasks/build/start.frag
fredck: @Reinmar, it makes sense.

I think we can proceed with the existing code and the do what I already said, to avoid such important thoughts to be left inside commits comments.:

> Ok, I'll open an issue for this once this PR goes accepted.

Reinmar: Follow-up: #8.

dev/tasks/build/builder.js
```diff
@@ -0,0 +1,238 @@
...
+
+		// Before starting, run the initial checkups.
+		if ( !this.checkUp() ) {
+			console.error( 'Build operation aborted.' );
+			callback( false );
+
+			return;
+		}
+
+		console.log( 'Building CKEditor into the "' + this.target + '" directory:')
```
gregpabian: JSHint and JSCS throw errors here - missing semicolon and missing space before closing round bracket.

fredck: Fixed with cccf119.

Damn WebStorm that doesn't work with our pre-commit hook. They fixed it some good time ago but still it didn't land into any of their releases.

PR #11, opened 2015-01-20
t/1: Introduce tests and Bender.js

### LINKED ISSUE #1: Introduce tests and Bender.js

fredck:
Fixes #1.

Tests for ckeditor5 code included with 100% code coverage.


dev/bender/plugins/ckeditor5/static/extensions.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+/* globals bender, CKEDITOR */
+
+'use strict';
+
+( function() {
+	// Override bender.start to wait until the "ckeditor" module is ready.
+	var originalStart = bender.start;
+	bender.start = function() {
```
gregpabian: I'm not sure if we discussed this already, but I think we should consider adding some space between the variable definition and the rest of the code. It would improve the readability of the code because right now it looks a bit crammed.

dev/bender/plugins/ckeditor5/static/extensions.js
```diff
@@ -0,0 +1,18 @@
...
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals bender, CKEDITOR */
+
+'use strict';
+
+( function() {
+	// Override bender.start to wait until the "ckeditor" module is ready.
+	var originalStart = bender.start;
```
gregpabian: why not `bender.defer()`?

tests/amd/amd.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals describe, it, expect, CKEDITOR */
```
gregpabian: We're gonna have describe, it and expect in every test, maybe we can "globalize" them in a single place, like `.jshintrc` file in the `tests/` directory?

tests/amd/amd.js
```diff
@@ -0,0 +1,31 @@
...
+
+'use strict';
+
+CKEDITOR.define( 'testModule', [ 'ckeditor' ], function( ckeditor ) {
+	return {
+		test: ( typeof ckeditor == 'object' )
```
gregpabian:
```js
typeof null == 'object'
```
Touché! :trollface:

tests/amd/amd.js
```diff
@@ -0,0 +1,31 @@
...
+	};
+} );
+
+describe( 'amd', function() {
+	it( 'require() should work with defined module', function( done ) {
+		CKEDITOR.require( [ 'testModule' ], function( testModule ) {
+			expect( testModule ).to.have.property( 'test' ).to.be.true();
+
```
gregpabian: Unnecessary empty line (I know I'm bit*hing right now, but I think this should be consistent across our tests).

tests/ckeditor/basepath.js
```diff
@@ -0,0 +1,70 @@
...
+beforeEach( function() {
+	// Ensure that no CKEDITOR_BASEPATH global is available.
+	delete window.CKEDITOR_BASEPATH;
+
+	// Remove all script elements from the document.
+	removeScripts();
+} );
+
+describe( 'ckeditor.basePath', function() {
+	it( 'Full URL', function( done ) {
```
gregpabian: Name of the test case is obscure. It should be written using natural language, e.g. "Full URL should do something".

tests/ckeditor/basepath.js
```diff
@@ -0,0 +1,70 @@
...
+		CKEDITOR.require( [ 'ckeditor' ], function( CKEDITOR ) {
+			addScript( 'http://bar.com/ckeditor/ckeditor.js' );
+			expect( CKEDITOR._getBasePath() ).equals( 'http://bar.com/ckeditor/' );
+			done();
+		} );
+	} );
+
+	it( 'CKEDITOR_BASEPATH', function( done ) {
```
gregpabian: Same as above.

tests/ckeditor/basepath.js
```diff
@@ -0,0 +1,70 @@
...
+		CKEDITOR.require( [ 'ckeditor' ], function( CKEDITOR ) {
+			window.CKEDITOR_BASEPATH = 'http://foo.com/ckeditor/';
+			expect( CKEDITOR._getBasePath() ).equals( 'http://foo.com/ckeditor/' );
+			done();
+		} );
+	} );
+
+	it( 'Ensure that no browser keep script URLs absolute or relative', function( done ) {
```
gregpabian: `it( 'should ensure that...`

tests/ckeditor/basepath.js
```diff
@@ -0,0 +1,70 @@
...
+
+	// Remove all script elements from the document.
+	removeScripts();
+} );
+
+describe( 'ckeditor.basePath', function() {
+	it( 'Full URL', function( done ) {
+		CKEDITOR.require( [ 'ckeditor' ], function( CKEDITOR ) {
+			addScript( 'http://bar.com/ckeditor/ckeditor.js' );
+			expect( CKEDITOR._getBasePath() ).equals( 'http://bar.com/ckeditor/' );
```
gregpabian: `).to.equal(` would read better

tests/ckeditor/basepath.js
gregpabian: Please review all the test case titles - I don't want to comment every single one.

dev/bender/plugins/ckeditor5/static/extensions.js
Reinmar: +1 and in this case it also matches the "margin around blocks" rule that we agreed on.

dev/bender/plugins/ckeditor5/static/extensions.js
fredck: Just because defer() wasn't still available when I wrote it :) I'll change it.

tests/ckeditor/basepath.js
fredck: I'm reviewing them right now and in fact it makes much more sense. We must also get used to use describe() properly.

tests/ckeditor/basepath.js
fredck: Hum, in fact there are many possible variations: `to.equal()`, `equal()`, `equals()`.

We may agree on best practices, but in any case I think we should not enforce this because it'll be a pain in the ass for contributors to have their stuff rejected because of such things, considering that chai gives all these possibilities and they're all right.

I'll change what we have not to `to.equal()`.

fredck: Btw, which is better: `to.equal()` or `to.be.equal()`?

gregpabian: The one that is grammatically correct :)

Reinmar: I remember we've been talking about using the shortest (or reasonably short) versions. Apart from grammatical correctness, I wouldn't like to use `to.be.equal()` because it is terrible to write. Personally, I would keep `expect().equals()`.

gregpabian: `.to.equal()` is just two letters longer but reads way better, just sayin'...

fredck: Actually to make it right, it should be `to.be.equal.to()`... too long!!!

`to.equal` is grammatically right as well, so let's use it.

fredck: @Reinmar, we use `().to` a lot for other cases so this would just get similar to others.

I'll make the change and lets see how it looks like.

Reinmar: > .to.equal() is just two letters longer but reads way better, just sayin'...

I know, but while writing my comment I noticed that it's not going through my fingers too well :P

Anyway - `to.equal()` is not bad. I'll perhaps use to type it so it will be faster. Just saying so we remember to keep them short.

tests/amd/amd.js
fredck: i tried this already but it doesn't work in the way we're using jshint right now (by pointing to a specific configuration file).

If we would like to do it it looks like we would have to bloat (further) the root by adding . jshintrc there (and at this point .jscs) as well.

fredck:
@gregpabian, I've fixed all your comments.

tests/amd/amd.js
Reinmar: Heh... I've been already thinking about this too. I think that the easiest (thus, doable) solutions are:

* Define little in the global configs and use comments extensively. The thing is that the global config of the ckeditor5-core module should for example contain `CKEDITOR` because it should be available in most of the files, but global config of the ckeditor5 module does not have to, because many files in this repository are related to grunt tasks. The rest will need to be defined in comments, which is not that bad, because you only need to do this when you start using some variable.
* Define a lot in the global configs. We can add most of the stuff that can be available somewhere in tests or dev code to the global config and we won't need to add so many comments. This makes the check less precise, but it's still good.

In CKEditor 4 we use now the second approach and it's ok. For instance, you never use `it` or `describe` in the dev code, so you don't do mistakes related to them.

fredck: Gonna have an issue for this once this PR gets merged.

tests/ckeditor/ckeditor.js
```diff
@@ -0,0 +1,43 @@
...
+			var originalGetPluginPath = CKEDITOR.getPluginPath;
+			CKEDITOR.getPluginPath = core.getPluginPath;
+
+			// This test is good for both the development and production codes.
+			var basePath = CKEDITOR.basePath;
+			var path = CKEDITOR.getPluginPath( 'test' );
+
+			expect( path ).to.equal( basePath + 'plugins/test/' );
+
+			CKEDITOR.getPluginPath = originalGetPluginPath;
```
Reinmar: It's not going to be executed if `expect().equal()` throws. I don't know how it works in Mocha+Chai, but in CKE4 you would need to handle teardown outside the test or in `try-catch-finally` :D.

fredck: Nice catch... I've just pushed a fix for it by simply reverting the override before the assertion.

fredck:
Just pushed commits that introduce `bender.amd.require()`, making the tests much cleaner.

tests/ckeditor/ckeditor.js
```diff
@@ -0,0 +1,39 @@
...
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals describe, it, expect */
+
+'use strict';
+
+bender.amd.require( 'ckeditor', 'ckeditor-core' );
+
+describe( 'getPluginPath()', function( CKEDITOR, core ) {
```
gregpabian: I'm not so convinced with this solution - this means we have to keep in mind the order of required modules and write them down in every `describe`'s callback. I think this solution would work better:
```js
var modules = bender.amd.require( 'ckeditor', 'ckeditor-core' );
```
And then we can use `modules.ckeditor` etc.

Besides, I feel it's too closely coupled with Mocha (which could be avoided thanks to `bender.defer`).

fredck: Still you would much probably not call `modules.ckeditor` directly, but instead have `var CKEDITOR = modules.ckeditor` on every single test, messing things again. This describe() helper cleans things up by defining this variables once only as its callback parameters.

Also, to make it work right, `var modules` would have to go in the global scope. Maybe not a big deal, but not better than the current solution as well.

Additionally, the proposed gives this sensation of being much similar to AMD (list of requires and modules as params).

I don't know, I still like it a lot and I don't see the big deal.

> Besides, I feel it's too closely coupled with Mocha (which could be avoided thanks to bender.defer).

I don't understand this.

Reinmar: > this means we have to keep in mind the order of required modules and write them down in every describe's callback

Hm... Makes sense. Usually we're testing one thing in one file, so there will be one `describe()` call. But eventually this mechanism will be adopted by benderjs-amd and then other projects will start using it and may want to use far more `describe()`s.

Besides that, I'm ok with both APIs.

fredck: > But eventually this mechanism will be adopted by benderjs-amd

Just to underline this, this feature has been coded with the CKEditor 5 dev environment in mind, exclusively, and is proposed in the ckeditor5 plugin for bender right because of this. It is optimized for our needs and the way we code tests.

gregpabian: > I don't understand this.

I mean the current solution is for BDD frameworks only (it's based on `describe()` method) - you can't use other testing framework/style with it, whereas the solution with the `modules` variable would work.

fredck: I agree, but please check my previous comment. This is intentional.

A generic solution may look different, in fact, but we should not miss the chance to bring an optimized solution to be used by us only.

dev/bender/plugins/ckeditor5/static/extensions.js
```diff
@@ -0,0 +1,90 @@
...
+				done();
+			} );
+		}
+	};
+
+	var originalDescribe = describe;
+	var describeQueue = [];
+
+	// Override the original Mocha's describe() so we can pass required modules to it.
+	describe = function( title, fn ) {
```
gregpabian: Did you check if this is enough to also cover the `describe.only` API? That is useful to run just a part of the suite.

dev/bender/plugins/ckeditor5/static/extensions.js
```diff
@@ -0,0 +1,90 @@
...
+
+'use strict';
+
+// ### General extensions.
+( function() {
+	// Make Bender wait to start running tests.
+	var done = bender.defer();
+
+	// Wait for the "ckeditor" module to be ready to start testing.
+	CKEDITOR.require( [ 'ckeditor' ], done );
```
gregpabian: Do we really need this one now?

dev/bender/plugins/ckeditor5/static/extensions.js
fredck: I didn't give attention `describe.only()` and `describe.skip()` and I think it doesn't support them (it even undefines them). 

Making this work for `describe.skip()` is a piece of cake. As for `describe.only()` instead, we would have to do more work, but should be doable.

I think it is worth bringing support for it, even if we'll never use them, because there should be a transparent experience to coders trying to use all mocha's power.

dev/bender/plugins/ckeditor5/static/extensions.js
fredck: It may be needed if `bender.amd.require()` is not used (e.g., the amd test). I think it is better to keep it just in case and it should not be a big deal anyway.

dev/bender/plugins/ckeditor5/static/extensions.js
fredck: Hum... no good news. `describe()` is supposed to return a "suite" object. We don't need it in our tests, but mocha needs it internally for `describe.only()`. As we're overriding `describe()` to delay its call, we cannot return anything and things break.

As a matter of fact, we don't need `describe.only()` as bender makes it possible to click in the test to run it only. So we could leave `only()` undefined and we would not suffer because of it.

I was able to make `describe.skip()` work well though. 

So we have two options:

 1. Go ahead with supporting `describe.skip()` with the current approach and keep `describe.done()` undefined.
 2. Change the whole thing and switch to the `var modules = bender.amd.require()` approach.

I would still go with "1" because it is cleaner, but this time I'll leave this to be decided by you guys.

Reinmar: In this case 2. Less pain with maintaining this.

gregpabian: `describe.only` runs a **selected suite** while Bender's test filter runs a **single test case** so those are two different cases.

I'd also go with the 2nd option for this and all the reasons I gave earlier.

fredck: Agree, your wish is my command :) Gonna work on this now.

fredck:
Changed strategy so now `bender.amd.require()` returns an object that will hold all loaded modules.

dev/bender/plugins/ckeditor5/lib/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+'use strict';
+
+var path = require( 'path' );
+var files = [
+	path.join( __dirname, '../static/extensions.js' )
+];
+
+module.exports = {
+	name: 'bender-framework-ckeditor5',
```
gregpabian: `framework` is no longer needed

fredck: Fixed with 0de6f6b.

tests/amd/amd.js
Reinmar: I reported #13.

.gitignore
```diff
@@ -9,3 +9,4 @@
 
 node_modules/**
 build/**
+coverage/**
```
Reinmar: What's in this directory? AFAICS we didn't have to ignore it in CKE4.

gregpabian: This is created while running `bender run` with coverage plugin enabled. I want to avoid a situation where someone accidentally pushes CC results to the repo.

PR #18, opened 2015-02-16
t/17: CKEDITOR._getBasePath() has insufficient tests

### LINKED ISSUE #17: CKEDITOR._getBasePath() has insufficient tests
It may have 100% tests coverage, but this means that every statement is executed, not that every statement is correct. Since there's a pretty complex regexp used in which we already had bugs in CKE4, more tests are required.

Furthermore, we have good tests for this in v4. Why not porting them?

Reinmar:
Fixes #17.

### DIFF
--- a/tests/ckeditor/basepath.js
+++ b/tests/ckeditor/basepath.js
@@ -18,12 +18,13 @@ beforeEach( function() {
 } );
 
 describe( 'basePath', function() {
-	it( 'should work with script tags', function() {
-		var CKEDITOR = modules.ckeditor;
-
-		addScript( 'http://bar.com/ckeditor/ckeditor.js' );
-		expect( CKEDITOR._getBasePath() ).to.equal( 'http://bar.com/ckeditor/' );
-	} );
+	testGetBasePathFromTag( 'http://bar.com/ckeditor/ckeditor.js', 'http://bar.com/ckeditor/' );
+	testGetBasePathFromTag( '/ckeditor/ckeditor.js', /\/ckeditor\/$/ );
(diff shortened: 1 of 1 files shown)


tests/ckeditor/basepath.js
```diff
@@ -31,6 +32,20 @@ describe( 'basePath', function() {
 		window.CKEDITOR_BASEPATH = 'http://foo.com/ckeditor/';
 		expect( CKEDITOR._getBasePath() ).to.equal( 'http://foo.com/ckeditor/' );
 	} );
+
+	function testGetBasePathFromTag( url, expectedBasePath ) {
```
gregpabian: Even though it's cool, DRY, KISS and everything else, it will break Bender.js, because you can't filter out test cases created on the fly. So no support for regressions and single test runs...

gregpabian: I still have to review this, but at the moment, for some reason the filtering mechanism doesn't work as it should.

fredck: Hum... we need to figure out something here, because this is a must have for us.

Reinmar: I don't understand. All the `it()` calls should add test cases and Bender.js should be able to filter them. Regardless of whether they are dynamically created or not. Actually... every single TC is created dynamically, because every single TC is created inside `describe()`'s callback.

gregpabian: I know, but still it does not work with your cases, I hope it's just a matter of escaping the descriptions.

gregpabian: Alright, I'm glad I can say it was just a filtering mechanism issue. Please update benderjs-mocha to 0.1.3 and it should work fine.

fredck: Ok, update with 31e347a.

PR #35, opened 2015-10-15
Add pagebuilder before the includes pagebuilder.

### LINKED ISSUE #34: The CKEditor plugin for Bender should add its files before the includes pagebuilder
In #31 we introduced a small test tools namespace. It can only work reasonably if files included by the tests (e.g. in the `ckeditor5-core` or plugins repositories) are loaded after the main tools file, which is loaded by the CKEditor plugin for Bender.

Unfortunately, the CKEditor plugin is added at the very end of the queue, so the main tools file is loaded at the very end too.

Reinmar:
Fixes #34.

### DIFF
--- a/bender.js
+++ b/bender.js
@@ -32,7 +32,8 @@ var config = {
 			applications: [ 'ckeditor' ],
 			paths: [
 				'tests/**',
-				'node_modules/ckeditor5-*/tests/**'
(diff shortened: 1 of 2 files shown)


Reinmar:
No objections from @oleq and @pjasiun, so merging...

PR #37, opened 2015-11-04
Added grunt dev:init task.

szymonkups:
This push request adds `grunt dev:init` task. It initializes development workspace for current configuration:
1. Scans trough `package.json` file and search for dependencies starting with `ckeditor5-`. 
2. If dependency has a valid GitHub URL ( starting with `ckeditor/`), task will clone its GitHub repository and link it into node_modules using `npm link`. For now new repositories are cloned inside the same directory where CKEditor5 repository is located (one level above the location of gruntfile.js).
3. If [commit-ish suffix](https://docs.npmjs.com/files/package.json#github-urls) is included, task will check out repository using that suffix.


Reinmar:
I made a very quick look at the code and it's generally OK. Before we merge it though, please use some of the ES6 goodies like classes, simpler method definitions in objects, arrow functions, etc.

Reinmar:
> If dependency has a valid GitHub URL ( starting with ckeditor/),

I don't think we should require the URL to start from `ckeditor/` because we want to allow all developers to create repositories with their plugins and use them the same way we use the official ones. So the only filter would be that the package name must start wit `ckeditor5-`.

szymonkups:
I will close this and open new PR with a group of grunt task.

PR #40, opened 2015-11-16
T/36 - Tools for project related task

szymonkups:
Added grunt tasks:
- dev-init - initialize development workspace
- dev-status - printing Git status of repositories in dev workspace
- dev-update - updating repositories in dev workspace
- dev-plugin-create - create new plugin
- dev-plugin-install - install plugin
- dev-boilerplate-update - update boilerplate project


dev/tasks/dev.js
```diff
@@ -0,0 +1,59 @@
...
+const pluginInstallTask = require( './utils/dev-plugin-install' );
+const pluginUpdateTask = require( './utils/dev-update' );
+const pluginStatusTask = require( './utils/dev-status' );
+const boilerplateUpdateTask = require( './utils/dev-boilerplate-update' );
+const ckeditor5Path = process.cwd();
+
+module.exports = ( grunt ) => {
+	const packageJSON = grunt.config.data.pkg;
+
+	grunt.registerTask( 'dev-init', function() {
```
Reinmar: Use arrow fn.

dev/tasks/dev.js
```diff
@@ -0,0 +1,59 @@
...
+		const options = getOptions( this );
+		pluginStatusTask( ckeditor5Path, packageJSON, options, grunt.log.writeln, grunt.log.error );
+	} );
+
+	grunt.registerTask( 'dev-boilerplate-update', function() {
+		const options = getOptions( this );
+		boilerplateUpdateTask( ckeditor5Path, packageJSON, options, grunt.log.writeln, grunt.log.error );
+	} );
+
+	function getOptions( context ) {
```
Reinmar: AFAIR `context == grunt`, so better to use a more meaningful name.

dev/tasks/tests/dev-init.js
```diff
@@ -0,0 +1,119 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+/* global describe, it, beforeEach, afterEach */
```
Reinmar: Move JSHint comment before `'use strict'`.

dev/tasks/tests/dev-init.js
```diff
@@ -0,0 +1,119 @@
...
+			expect( spies.parseRepositoryUrl.calledTwice ).to.equal( true );
+			expect( spies.cloneRepository.called ).to.equal( false );
+			expect( spies.checkout.calledTwice ).to.equal( true );
+			expect( spies.linkDirectories.calledTwice ).to.equal( true );
+			expect( spies.npmInstall.calledTwice ).to.equal( true );
+			expect( spies.installGitHooks.calledTwice ).to.equal( true );
+		} );
+	} );
+
+	function createSpies() {
```
Reinmar: We usually divide test file into two sections, preparations and then TCs. It means that it's semantically OK to put such functions at the end of the first section.

dev/tasks/utils/git.js
```diff
@@ -0,0 +1,154 @@
...
+
+	/**
+	 * Updates boilerplate project in specified repository.
+	 * @param {String} repositoryPath Absolute path to repository.
+	 */
+	updateBoilerplate( repositoryPath ) {
+		// Try to add boilerplate remote if one is not already added.
+		try {
+			tools.shExec( `cd ${ repositoryPath } && git remote add boilerplate ${ this.BOILERPLATE_REPOSITORY }` );
+		} catch ( error ) { }
```
Reinmar: Never ever catch to forget. It's better to let the whole execution blow up than to hide an error.

Reinmar: OK, turns out it makes sense to hide this error because it may be thrown if boilerplate remote existed. However, what if different error than this is thrown?

dev/tasks/utils/dev-init.js
```diff
@@ -0,0 +1,71 @@
...
+			const urlInfo = git.parseRepositoryUrl( repositoryURL );
+			const repositoryAbsolutePath = path.join( workspaceAbsolutePath, dependency );
+
+			// Check if repository's directory already exists.
+			if ( directories.indexOf( dependency ) === -1 ) {
+				try {
+					writeln( `Clonning ${ repositoryURL }...` );
+					git.cloneRepository( urlInfo, workspaceAbsolutePath );
+				} catch ( error ) {
+					writeError( error );
```
Reinmar: If this fails, the below should not be executed, so one try-catch would be better.

dev/tasks/utils/dev-init.js
```diff
@@ -0,0 +1,71 @@
...
+				}
+			}
+
+			// Check out proper branch.
+			try {
+				writeln( `Checking out ${ repositoryURL } to ${ urlInfo.branch }...` );
```
Reinmar: I think that before checking out we should pull changes. Otherwise we'll be checking to an old branch.

Second thing is a missing check if the repository contains some uncommitted changes.

dev/tasks/utils/dev-init.js
```diff
@@ -0,0 +1,71 @@
...
+				git.checkout( repositoryAbsolutePath, urlInfo.branch );
+
+				writeln( `Linking ${ repositoryURL }...` );
+				tools.linkDirectories( repositoryAbsolutePath, path.join( ckeditor5Path, 'node_modules' , dependency ) );
+
+				writeln( `Running npm install in ${ repositoryURL }.` );
+				tools.npmInstall( repositoryAbsolutePath );
+
+				writeln( `Installing GIT hooks in ${ repositoryURL }.` );
```
Reinmar: It's "Git" not "GIT".

dev/tasks/utils/dev-plugin-create.js
```diff
@@ -0,0 +1,90 @@
...
+ * 9. Install Git hooks in plugin repository.
+ *
+ * @param {String} ckeditor5Path Path to main CKEditor5 repository.
+ * @param {Object} options grunt options.
+ * @param {Function} writeln Function for log output.
+ * @param {Function} writeError Function of error output
+ * @returns {Promise} Returns promise fulfilled after task is done.
+ */
+module.exports = ( ckeditor5Path, options, writeln, writeError ) => {
+	return new Promise( ( resolve, reject ) => {
```
Reinmar: Do you need this promise? I think that you can return that chain which starts with `inquiries.getPluginName()`.

dev/tasks/utils/dev-plugin-create.js
```diff
@@ -0,0 +1,90 @@
...
+
+					writeln( `Linking ${ pluginName } to node_modules...` );
+					tools.linkDirectories( repositoryPath, path.join( ckeditor5Path, 'node_modules', pluginName ) );
+
+					writeln( `Running npm install in ${ pluginName }.` );
+					tools.npmInstall( repositoryPath );
+
+					writeln( `Installing GIT hooks in ${ pluginName }.` );
+					tools.installGitHooks( repositoryPath );
+				} catch ( error ) {
```
Reinmar: Can't you handle this error in `.catch()` of the promise chain?

dev/tasks/dev.js
```diff
@@ -0,0 +1,59 @@
...
+
+	grunt.registerTask( 'dev-init', function() {
+		const options = getOptions( this );
+		initTask( ckeditor5Path, packageJSON, options, grunt.log.writeln, grunt.log.error );
+	} );
+
+	grunt.registerTask( 'dev-plugin-create', function() {
+		const done = this.async();
+		const options = getOptions( this );
+		pluginCreateTask( ckeditor5Path, options, grunt.log.writeln, grunt.log.error ).then( done );
```
Reinmar: You need to handle rejected promise here too.

Reinmar: I know that it is (or will be) handled inside `pluginCreateTask()` but if this function returns a promise it's necessary to handle it. Otherwise, if after some refack code gets broken somewhere "outside" that handler which is used inside `pluginCreateTask()` then you won't know what's going on.

dev/tasks/utils/tools.js
```diff
@@ -136,5 +138,120 @@ module.exports = {
...
+	 * @param {String} source
+	 * @param {String} destination
+	 */
+	linkDirectories( source, destination ) {
+		// Remove destination directory if exists.
+		if ( this.isDirectory( destination ) ) {
+			this.shExec( `rm -rf ${ destination }` );
+		}
+
+		this.shExec( `ln -s ${ source } ${ destination }` );
```
Reinmar: Please check with @mlewand that it's going to work on Windows. If not, perhaps we should use some npm module that would gracefully degrade to Windows env.

dev/tasks/utils/inquiries.js
```diff
@@ -0,0 +1,55 @@
...
+const DEFAULT_GITHUB_URL_PREFIX = 'ckeditor/';
+
+module.exports = {
+	getPluginName() {
+		return new Promise( ( resolve ) => {
+			inquirer.prompt( [ {
+				name: 'pluginName',
+				message: 'Enter plugin name without ' + DEFAULT_PLUGIN_NAME_PREFIX + ' prefix:',
+				validate: ( input ) => {
+					var regexp = /^[a-zA-Z0-9-_]+$/;
```
Reinmar: Sounds like `/[\w-]+$/`.

dev/tasks/utils/tools.js
Reinmar: Or... npm's fs module may have a cross-os method for that.

dev/tasks/utils/dev-plugin-create.js
```diff
@@ -0,0 +1,90 @@
...
+					writeError( error );
+				}
+			} )
+			.catch( reject );
+	} );
+};
```
Reinmar: OK, this task should also fill out couple of gaps in the files coming from boilerplate. To simplify that you can also change `boilerplate/ckeditor5` branch so e.g. README or LICENSE or CONTRIBUTE files contain something useful to work with.

dev/tasks/utils/dev-update.js
```diff
@@ -0,0 +1,50 @@
...
+		for ( let dependency in dependencies ) {
+			const repositoryURL = dependencies[ dependency ];
+			const urlInfo = git.parseRepositoryUrl( repositoryURL );
+			const repositoryAbsolutePath = path.join( workspaceAbsolutePath, dependency );
+
+			// Check if repository's directory already exists.
+			if ( directories.indexOf( dependency ) > -1 ) {
+				try {
+					writeln( `Updating ${ repositoryURL }...` );
+					git.pull( repositoryAbsolutePath, urlInfo.branch );
```
Reinmar: Similarly like `dev-init` this task also needs to do safety check for uncommitted changes.

dev/tasks/dev.js
szymonkups: I'm using grunt `async()` and `options()` methods that are available in function's scope.

dev/tasks/utils/dev-init.js
szymonkups: If  some uncommitted changed are present - `git.checkout` method will throw an exception with `git checkout` command's error message. This message will be printed to the user using `grunt.log.error`.

dev/tasks/utils/dev-update.js
szymonkups: Standard error output is thrown when `git pull` command fails. Do we need more sophisticated check?

Reinmar:
OK, there are couple of followups like:

* #43
* #44 
* #42 
* #41 

But this is a good start.

Issue #66, opened 2016-01-18
Dev tasks should not use the word "plugin" – "package" is the right one


Reinmar:
See https://github.com/ckeditor/ckeditor5/issues/47#issuecomment-172535870

Reinmar:
At least in those two places:

* `gulp dev-install --plugin`
* `gulp dev-plugin-create`

Reinmar:
Fixed by #67.

Issue #44, opened 2015-11-26
Get rid of Grunt


Reinmar:
To be totally replaced with Gulp. But it means that we need to figure out what to do with the boilerplate repo.

Reinmar:
Fixed by #67.

PR #67, opened 2016-01-19
T/44 Get rid of Grunt

szymonkups:
PR for #44 issue. It also covers #66 and removes `dev-boilerplate-update` as we are no longer using boilerplate project.


Reinmar:
The `pre-commit` task should be defined in the main `gulpfile.js` (similarly to the `default` task), because it may do more than just linting in the future.

Reinmar:
Second, we need a branch in the ckeditor5-core repo which will update that repo as well.

Reinmar:
```(t/44 0b5e100) p@m /www/ckeditor5/ckeditor5> gulp lint
[17:31:07] Using gulpfile /www/ckeditor5/ckeditor5/gulpfile.js
[17:31:07] Starting 'lint'...
Incompatible values for the 'strict' and 'globalstrict' linting options. (13% scanned) at tests/_utils/utils.js :
     4 |  */
     5 |
     6 | /* jshint node: false, browser: true, globalstrict: true */
---------^
     7 | /* globals before, afterEach, sinon */
     8 |

1 code quality error found.
[17:31:11] Finished 'lint' after 3.81 s```

**Comment by Reinmar** [comment]
```
const src = [ '**/*.js' ].concat( config.IGNORED_FILES.map( i => '!' + i ), getGitIgnore() );
```

Seeing how long it takes to scan all files in this repo I think that we could change this to:

```
const src = [ '@(src|tests|dev)/**/*.js', '*.js' ].concat( config.IGNORED_FILES.map( i => '!' + i ), getGitIgnore() );
```

It speeds up the task twice. On the other hand we risk missing some files in the future, if other directories will be used than those... so we can keep the current pattern. After all, most of the time we'll be using pre-hooks.

Reinmar:
I'm considering removing the `dev-` prefix from the dev tasks. It makes the names unnecessarily long. We should not have conflicts anyway.

Reinmar:
And... that would be all. Very good job :) It's nice to see this simplified too.

Reinmar:
One minor detail – when `dev-create-package` changes `package.json` it does not sort dependencies as npm does natively. Would be nice to have it.

szymonkups:
Pushed commits to PR branch that should resolve issues mentioned above.

PR #81, opened 2016-02-11
Remove unused version of files when building

### LINKED ISSUE #50: Remove unused version of files when building
Moved from https://github.com/ckeditor/ckeditor5-core/issues/130.

szymonkups:
Attempt to resolve #50.

### DIFF
--- a/dev/tasks/build/utils.js
+++ b/dev/tasks/build/utils.js
@@ -200,20 +200,37 @@ require( [ 'tests' ], bender.defer(), function( err ) {
 	},
 
 	/**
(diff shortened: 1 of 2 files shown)


Reinmar:
An idea here – I think that it worked that if inside `rename()`'s callback you clear the name of the file totally it gets removed. I can't find such usage of `rename()` now, but I recall doing something like this. And it would simplify the code significantly.

Reinmar:
Hm... it doesn't seem to work. So maybe it was a different function.

Reinmar:
OK, I refactored the code which allowed simplifying it and improving its performance.

Issue #84, opened 2016-02-17
Run tests in Node.js environment.


szymonkups:
We should be able to run tests in Node.js environment, more information can be found in this [ckeditor5-design issue](https://github.com/ckeditor/ckeditor5-design/issues/87).

Reinmar:
Fixed by #85.

PR #85, opened 2016-02-17
Enable testing under Node.js

szymonkups:
This PR tries to resolve issue #84 and includes:
- changes made to builder when transpiling to CommonJS format:
  - Bender.js launcher is not appended to test files for `cjs` format
  - paths between required modules are resolved do be relative
  - ~~`babel-plugin-transform-es2015-parameters` is used to transpile `rest parameter` not supported yet by Node.js~~
- `tests/_utils/module__cjs.js` - `define` and `require` utility methods used by tests, this is the part that need to be checked very carefully. This code tries to emulate behavior of AMD version of `define` using CommonJS and module mocking. I've tried different approaches and came up with a solution that is included in this PR. Can this be done simpler or in another way?
- new gulp task `gulp test-editor --node` that runs tests under Node.js ( after building with `gulp build --formats cjs` )


Reinmar:
This branch has conflicts. Please update it.

szymonkups:
Rebased with master.

Reinmar:
```(t/84 7e89724) p@m /www/ckeditor5/ckeditor5> gulp test-editor
[13:33:38] Using gulpfile /www/ckeditor5/ckeditor5/gulpfile.js
[13:33:38] Starting 'test-editor'...
[13:33:38] Finished 'test-editor' after 429 μs
```

I don't think that the tests were really executed.

I also noticed that `gulp test-editor` doesn't run `gulp build formats=cjs` while it would be great to have this automated.

Reinmar:
Ah, I see that `gulp test-editor --node` works. I forgot that we decided to define it this way. But still, it would be nice to:

* run the builder automatically
* thrown when `--node` isn't passed

Reinmar:
One more thing – can we do something to better show which test cases is in which file? When running tests in the browser we know the test file which has the red test. In Node.js I need to based on the `describe()` calls which aren't always that clear.

If Mocha doesn't have some option to print test files or something, then we'll need to create a ticket in the design repo to recommend that main describe() calls should be identifiable.

szymonkups:
_Updated info about this PR_.

This PR contains changes in `ckeditor/ckeditor5` repository. Changes in other repositories are kept in branches specified in `package.json` file in this PR. 

**Running tests**
To run tests under Node.js, please use node in version v 6.1.0 or higher (no other transpilers are needed since v6). If older version of node will be used - task will throw proper error informing about this.

To start testing use:
```
gulp test-node
```
If project build is needed, specify a `build` option:
```
gulp test-node --build
```
To calculate code coverage use:
```
gulp test-node --coverage
```

**Ignored tests**
Tests from `ui`, `ui-*` and `theme-*` repositories are ignored. Also single test files can be ignored by adding `browser-only` tag to Bender.js tags list.

Most of the ignored tests use DOM elements and cannot be tested under Node.js. Also view test utilities `parse` and `setData` are using `HtmlDataProcessor` ( which is using browser's `DOMParser`). This is probably the most important part to change as test utilities will be used across `ckeditor5-engine` and others.

szymonkups:
I've created follow ups:
https://github.com/ckeditor/ckeditor5/issues/167
https://github.com/ckeditor/ckeditor5-engine/issues/424

Reinmar:
And one more followup: https://github.com/ckeditor/ckeditor5/issues/168

Reinmar:
I'm close to R+ this, but we need to improve the way how istanbul and test runner are configured. Currently the paths are hardcoded while they are very similar to what we have in `bender.js`. In fact, the istanbul options should copied directly from `bender.js` (now CC covers also lodash, because the option is insufficient). Let's keep all this in one place, because we'll have problems updating the code.

So that's the last thing we need to improve and we're done.

Reinmar:
PS. I pushed lots of code to all branches so make sure to run `gulp update`.

Reinmar:
That was a huge PR, thanks @szymonkups!

It also covered:

* https://github.com/ckeditor/ckeditor5-engine/commit/01105a6d2137a04f23bc16bc7cab02cc9f5a8e11
* https://github.com/ckeditor/ckeditor5-undo/commit/e95b197129453bafb154973b9f2a14356267f999
* https://github.com/ckeditor/ckeditor5-utils/commit/999201ffd02d23c18b394f407af22fcdf10c0235

Issue #90, opened 2016-02-19
Dev tasks: gulp update fails to pull changes


Reinmar:
Check the branch "ui" in this repo. Check it out and run `gulp update`. I get:

```sh
(ui 3b99f9b) p@macintosh /www/ckeditor5> gulp update
[16:07:24] Using gulpfile /www/ckeditor5/gulpfile.js
[16:07:24] Starting 'update'...
Checking out ckeditor5-core to t/209...
Pulling changes to ckeditor5-core...
[16:07:26] 'update' errored after 1.8 s
[16:07:26] Error: Error while executing `cd /www/ckeditor5-core && git pull origin t/209`:


    at Object.shExec (/www/ckeditor5/dev/tasks/dev/utils/tools.js:20:10)
    at Object.pull (/www/ckeditor5/dev/tasks/dev/utils/git.js:95:9)
    at module.exports (/www/ckeditor5/dev/tasks/dev/tasks/update.js:45:10)
    at Gulp.<anonymous> (/www/ckeditor5/dev/tasks/dev/tasks.js:36:3)
    at module.exports (/www/ckeditor5/node_modules/orchestrator/lib/runTask.js:34:7)
    at Gulp.Orchestrator._runTask (/www/ckeditor5/node_modules/orchestrator/index.js:273:3)
    at Gulp.Orchestrator._runStep (/www/ckeditor5/node_modules/orchestrator/index.js:214:10)
    at Gulp.Orchestrator.start (/www/ckeditor5/node_modules/orchestrator/index.js:134:8)
    at /usr/local/lib/node_modules/gulp-cli/lib/versioned/^3.7.0/index.js:46:20
    at nextTickCallbackWith0Args (node.js:433:9)
```

Reinmar:
Turns out I had a problem inside that repository:

```
(t/209 9a4406b) p@macintosh /www/ckeditor5-core> git pull origin t/209
From github.com:ckeditor/ckeditor5-core
 * branch            t/209      -> FETCH_HEAD

It seems that there is already a rebase-apply directory, and
I wonder if you are in the middle of another rebase.  If that is the
case, please try
	git rebase (--continue | --abort | --skip)
If that is not the case, please
	rm -fr "/www/ckeditor5-core/.git/rebase-apply"
and run me again.  I am stopping in case you still have something
valuable there.
```

So the problem is that the gulp task doesn't print the error to the console. I'll report a separate ticket for this.

Reinmar:
[...]

Issue #78, opened 2016-02-08
Dev tasks should output the result of shell commands


Reinmar:
I've just executed `gulp update --npm-update` and I realised that I have no idea what has actually happened (e.g. which npm packages where updated). 

We can either print everything on the stdout or to a log file.

Reinmar:
Another case to cover: https://github.com/ckeditor/ckeditor5/issues/90#issuecomment-186252895

Reinmar:
Fixed by #91.

PR #91, opened 2016-02-22
Introduced simple log mechanism for dev tasks output.

szymonkups:
Created log object that is used across dev-tasks to log output. It is currently configured to log using `console.log` and `console.error` methods. We can provide other log methods in the future (e.g. logging to file). This should help resolving #78 issue.
I've added some changes to `shExec` tool function. Now it is using stderr output when throwing an error (#90).


Reinmar:
I wonder when we need to use gutil.log (https://github.com/gulpjs/gulp-util#logmsg) and when something more than that. The shell tasks are a problem actually, because all other things log from JS. Shell tasks' output is just intercepted and it would be best to avoid affecting it.

szymonkups:
Currently shell tasks do not output to console directly. After executing the task, `stdout` and `stderr` outputs are passed to `log` object and then logged to `console.log` / `console.error`. Problem with this approach is that all the output is logged after execution is finished - this is how [shelljs](https://github.com/shelljs/shelljs) library works. If this will be a major problem for us, we should consider changing the library/create our own solution.

Reinmar:
I'd move some of those tools (including the logger) to `dev/tasks/utils` in the future, because they may need to be reused by many tasks. But that applies also to the builder and other utils, so it needs to be handled in a separate ticket.

PR #155, opened 2016-05-11
Adjusted EditorConfig get method to new behaviour of get method from inherited class.

oskarwrobel:
Related to: https://github.com/ckeditor/ckeditor5-utils/pull/34


src/editorconfig.js
```diff
@@ -22,11 +23,19 @@ export default class EditorConfig extends Config {
...
 	get() {
-		// Try to take it from this editor instance.
-		let value = super.get.apply( this, arguments );
+		let value;
 
-		// If the configuration is not defined in the instance, try to take it from CKEDITOR.config.
-		if ( typeof value == 'undefined' ) {
+		try {
+			// Try to take it from this editor instance.
+			value = super.get.apply( this, arguments );
```
Reinmar: BTW. You can do this now with `get( ...args ) { super.get( ...args ); }`.

Reinmar:
I proposed a different solution – removing the code: https://github.com/ckeditor/ckeditor5/issues/157.

PR #180, opened 2016-05-24
Fixed unit tests and addded integrational tests.

### LINKED ISSUE #179: Integration tests for Editor.setData/getData
There was an issue that arguments passed to `DataController` from `Editor.setData` were in wrong order. No test catch that, so we should write some integration tests that check if data and other arguments are passed properly to `DataController` and then down to the model.

Reinmar:
Fixes #179.

Complementary change in the classic creator https://github.com/ckeditor/ckeditor5-creator-classic/pull/6.


Reinmar:
BTW. I added integrational tests which use HtmlDP to `tests/editor/baseeditor.js`. They are not going to pass in Node.js right now, but I didn't want to complicate the tests by moving them to `editor/editor.js`. We need to solve HtmlDP compat with Node.js anyway.

PR #183, opened 2016-05-24
Implemented keystroke handler

### LINKED ISSUE #115: Implement simple keystroke handler
Moved from https://github.com/ckeditor/ckeditor5-core/issues/231

Reinmar:
Fixes #115.

### DIFF
--- a/src/keystrokehandler.js
+++ b/src/keystrokehandler.js
@@ -6,25 +6,54 @@
 'use strict';
 
 import EmitterMixin from './utils/emittermixin.js';
-import KeyObserver from './engine/treeview/observer/keyobserver.js';
+import KeyObserver from './engine/view/observer/keyobserver.js';
 import { getCode, parseKeystroke } from './utils/keyboard.js';
 
 /**
- * Keystroke handler.
+ * Keystroke handler. Its instance is available in {@link ckeditor5.Editor#keystrokes} so features
+ * can register their keystrokes.
+ *
+ * E.g. an undo feature would do this:
+ *
+ *		editor.keystrokes.set( 'ctrl + Z', 'undo' );
+ *		editor.keystrokes.set( 'ctrl + shift + Z', 'redo' );
+ *		editor.keystrokes.set( 'ctrl + Y', 'redo' );
  *
  * @memberOf ckeditor5
  */
 export default class KeystrokeHandler {
 	/**
+	 * Creates an instance of the keystroke handler.
+	 *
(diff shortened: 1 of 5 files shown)


src/keystrokehandler.js
```diff
@@ -0,0 +1,112 @@
...
+		 */
+		this.editor = editor;
+
+		/**
+		 * Observer used to listen to events for easier keystroke handler destruction.
+		 *
+		 * @private
+		 * @member {utils.Emitter} ckeditor5.KeystrokeHandler#_observer
+		 */
+		this._observer = Object.create( EmitterMixin );
```
pjasiun: I called it `_listenters` in EditingController not to confuse with `Obeserver`s.

Reinmar: `_listeners.listenTo` looks very bad. Maybe `_subscribers` (as in pub-sub)? But I'd really like to use `_observers` as that's the naming we use elsewhere. As I wrote many times – trying to figure out a unique name every single time you use a similar concept leads to a mess.

pjasiun: What's wrong in `_listeners.listenTo`? If it can listen or stop listening it seems to be exactly a `listener`. Observer should observe, not listen.

pjasiun: But, in fact, any name is fine for me, as long we do not have a naming conflict. When I saw `_observer`, I was sure that it is an `KeyObserver` used by `KeystrokeHandler`. So I am fine with `_listeners`, `_subscribers` or renaming `Observer` class to `Listener` (in fact most of them are DOM listeners), but I do not like using same name for both meanings.

Reinmar: > What's wrong in _listeners.listenTo? If it can listen or stop listening it seems to be exactly a listener. Observer should observe, not listen.

The object is called "listeners". It does not contain nor represent multiple listeners. It's a single object which observes other instances. Plural names would be good for collections, but this is not one. So if anything, I'd go with `_subscriber` or `_observer`. `_observer` is good, because it's the name we've already been using. `_listener` would be nice because the method we have is called `listenTo()`, but I like it much less. That makes me think that this is another point for our Emitter refactoring issue.

I'll rename it to `_listener` (SINGULAR).

pjasiun: This is exactly what I realised and did in my PR 14 munutes ago: https://github.com/ckeditor/ckeditor5-engine/pull/442/commits/39d74636581ca631c4668f4bb31a3d0b724f6ad1 😄

PR #192, opened 2016-05-31
Implemented the editor

### LINKED ISSUE #185: Review editor creation and base architecture
Because engine introduced both data (https://github.com/ckeditor/ckeditor5-engine/issues/432) and editing (https://github.com/ckeditor/ckeditor5-engine/issues/288) controllers, all creators need to be updated so the editor will use engine for real.

I am not sure if it makes sense to update creators, they need to be changed anyway. We could take this opportunity to apply change introduced here: https://github.com/ckeditor/ckeditor5-design/issues/140

Reinmar:
Simple changes. We've implemented the editor. Fixes #185.

For more info see https://github.com/ckeditor/ckeditor5/issues/185#issuecomment-222739997


Reinmar:
Only 55 files changed. Nothing to see here.

PR #193, opened 2016-06-01
Bundling

### LINKED ISSUE #187: Bundling
We need to bundle whole editor files into two files: `ckeditor.js` and `ckeditor.css`. For JS bundling we decided to use `Rollup` with tree-shaking option (inlcludes only uses library code instead of whole library). CSS file is already bundled (builded theme has only one CSS file), but we still need to minify it.

It seems to be pretty easy, but:

- At this moment we don't know a list of every dependency needed in the bundle. It is because editor features load automatically during initialization process. To work around this problem @Reinmar idea was to create custom entry file with necessary imports and wrapped `CKEDITOR.create` function. Something like that:
```
import CKEDITOR from 'yyy.js';
import FeatuerA from 'xxx.js';

export default function createEditor( element, config ) {
    return CKEDITOR.create( element, Object.assign( config, { features: [ FeatureA ] } ) );
}
```

- There is no (or I couldn't find) JS minifier fully supporting esnext syntax. Harmony branch of UglifyJS seems to be closest to reach this goal, but still needs to convert some of code to ES5 before minification.

oskarwrobel:
Fixes #187 

screenshot 2016-06-01 13 06 13

screenshot 2016-06-01 13 05 59

@TODO:
- source maps
- move import of babel helpers to gulp https://github.com/ckeditor/ckeditor5/pull/193/files#diff-5392e8a5c2208af620b6fc87880ad21aR19


Reinmar:
Awesome!

And just out of curiosity – the sizes are given before gzipping, right?

oskarwrobel:
> And just out of curiosity – the sizes are given before gzipping, right?

Yes.

Reinmar:
And how much does it take after gzipping (you can use `gzip -9` on your console).

oskarwrobel:
screenshot 2016-06-01 13 40 30

Reinmar:
Yep, that's better now :). There's lots to compress, because unlike CKE4's code CKE5's code is super object oriented and there are a lot of long names which cannot be minified, but gzip super well. That's a result :). For comparison, iirc CKE4's compression was able to cut off 50% of the file size.

oskarwrobel:
I think we can still reduce size of js file a little bit. ~1/5 of file size is a babel stuff. Right now we are transforming everything, but probably we don't have to. We can check what we really need to transform (according to browsers [including mobile] we want to support), and configure babel with only necessary plugins. But we also need to keep in mind that not every ES6 features can be minified, because there is no minifier fully supports ES6.

fredck:
We're exactly half way to the v4  standard distribution size, which is 520KB, gzipped to 152KB. Hopefully we'll do better, with the same amount of features.

It's great to start seeing bundling happening though!

Reinmar:
> We're exactly half way to the v4 standard distribution size, which is 520KB, gzipped to 152KB.

Oops, I remembered that CKE4 isn't gzipping that well. So yes... we're half way :D.

dev/tests/bundle/tasks.js
```diff
@@ -0,0 +1,63 @@
+/**
```
oskarwrobel: Unfortunately unit tests not cover whole code, because testing gulp is really difficult :/

Reinmar: In the dev tasks we don't need 100%. Covering the most important paths will be enough.

Reinmar: And a general remark (I haven't checked the code yet) regarding testing gulp – it's just a set of streams. Expose them in functions and test those functions generating streams. Still it takes more code than usual to do that, but it's not that hard.

dev/tasks/bundle/buildclassiceditor.js
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Babel helper.
+ *
+ * @TODO: Move to bundle task.
+ * Should be injected by bundle task, because our source code don't have to know that will be transformed.
+ * But it is not so easy. It is only possible to pass string with file path to rollup,
+ * so to do it we need to create a temporary file.
+ */
+import '../../../node_modules/regenerator-runtime/runtime.js';
```
jodator: Would be cool if that could be resolved by some other way. This does not work if CKEditor is used as dependency (all ckeditor5 repos are in one node_modules folder). This is not a big issue I think but something to think about since bundling is not crucial for including CKEditor5 as dependency (in contrast to transpiling for tests).

dev/tasks/bundle/buildclassiceditor.js
```diff
@@ -0,0 +1,60 @@
+/**
```
Reinmar: We need a generator for such a file and/or make it possible to pass path to such a file to the builder. The generator would accept feature paths (BTW. see `PluginCollection.getPluginPath()`) and generate this file. Passing a path to the file will be useful for some integration cases when you have e.g. your own plugins in a totally different place.

oskarwrobel: Maybe something as `bundleConfig.json`?
```javascript
{
    "moduleName": "MyCKEditor",
    "editor": "classic", // Or path to custom creator
    "features": [
        'undo',
        'redo',
        'path/to/custom/feature.js'
    ]
}


gulp bundle --config path/to/bundleConfig.json


const editor = new MyCKEditor( ... )
```

or/and

```javascript
gulp bundle --moduleName MyCKEditor --editor classic --features undo,redo,path/to/custom/feature.js
```

oskarwrobel: Maybe something as `bundleConfig.json`?
```javascript
{
    "moduleName": "MyCKEditor",
    "editor": "classic", // Or path to custom creator
    "features": [
        'undo',
        'redo',
        'path/to/custom/feature.js'
    ]
}


gulp bundle --config path/to/bundleConfig.json


const editor = new MyCKEditor( ... )
```

or/and

```javascript
gulp bundle --moduleName MyCKEditor --editor classic --features undo,redo,path/to/custom/feature.js
```

Reinmar: `bundle-config.js` makes sense. Passing all this in the commandline would be hard anyway. Too many things are needed there.

dev/tasks/bundle/utils.js
```diff
@@ -33,11 +33,25 @@ const utils = {
...
+	 *
+	 * @param {String} name
+	 * @returns {String} Path to the module.
+	 */
+	getPluginPath( name ) {
+		if ( name.indexOf( '/' ) > 0 ) {
+			return name;
+		}
+
+		return './build/esnext/ckeditor5/' + ( name + '/' + name ) + '.js';
```
Reinmar: You can use tpl strings.

dev/tasks/bundle/utils.js
```diff
@@ -0,0 +1,188 @@
...
+	},
+
+	/**
+	 * Render content for entry file which needs to be passed as main file to the Rollup.
+	 *
+	 * @param {Object} data
+	 * @param {String} [data.moduleName] Name of the editor class exposed in bundle. e.g. MyCKEditor.
+	 * @param {String} [data.creator] Path to the editor creator.
+	 * @param {Array<String>} [data.features] List of paths or names to features which need to be included in bundle.
```
Reinmar: I think it should be `Array.<String>`.

Reinmar:
Congrats!

Reinmar:
We reported couple of followups, but the task works great and only misses some features.

Issue #186, opened 2016-05-25
Alias `gulp update` as `gulp pull`


Reinmar:
I always do this in `ckeditor5/`:

```
> git pull
> gulp pull
< error >
```

Reinmar:
And `gulp status` as `gulp st`.

PR #196, opened 2016-06-01
Additional aliases for Gulp

pomek:
Fixes: #186 

Added two aliases:
* `gulp st` => `gulp status`
* `gulp pull` => `gulp update`


Reinmar:
Please remember to update https://github.com/ckeditor/ckeditor5/wiki/Development-Workflow.

And please add some basic tests.

pomek:
@Reinmar, tests are ready. I also prepared changes on [Development-Workflow](https://github.com/ckeditor/ckeditor5/wiki/Development-Workflow), but I will save my changes after closing PR.

dev/tasks/dev/tasks.js
```diff
@@ -69,4 +62,19 @@ module.exports = ( config ) => {
...
+
+	function updateTaskHandler() {
+		const options = minimist( process.argv.slice( 2 ), {
+			boolean: [ 'npm-update' ],
+			default: {
+				'npm-update': false
+			}
+		} );
+
+		updateTask( installTask, ckeditor5Path, packageJSON, config.WORKSPACE_DIR, options[ 'npm-update' ] );
```
Reinmar: Task may return some value – this is important for gulp to control async flow.

Reinmar:
Hm... at first the idea with running the same tests for alias looked pretty well, but that give us very little. You may have a simple error in your test and execute them twice for the same task and you won't even know, because CC is the same as you're testing exactly the same code.

So I would say that to check an alias it's enough to check that such a task exists. There's very little chance that alias won't work while the main one does.

Reinmar:
One test is failing on this branch:

``` 
158 passing (3s)
  1 failing

  1) lint lintStaged should throw error when linting fails:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

Reinmar:
Everything's fine now. Congrats for your first closed PR!

Please remember about updating the wiki now.

Issue #89, opened 2016-02-18
Dev tasks: I cannot push a freshly created package


Reinmar:
I created a package and once I try to push it to an existing repo I get:

```
(master 1b34661) p@m /www/ckeditor5/ckeditor5-ui-default> git push origin master
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

It seems that the origin remote has to be defined.

Reinmar:
Yep, I had to do this: `git remote add origin git@github.com:ckeditor/ckeditor5-ui-default.git`.

PR #197, opened 2016-06-02
Adds remote into created package.

pomek:
Fixes: #89 

During `gulp create-package`, remote was not created. Now, after init a repository, remote will be added and `git push` shouldn't throw any errors.

### DIFF
--- a/dev/tasks/dev/tasks/create-package.js
+++ b/dev/tasks/dev/tasks/create-package.js
@@ -14,7 +14,7 @@ const log = require( '../utils/log' );
 /**
  * 1. Ask for new package name.
  * 2. Ask for initial version.
- * 3. Ask for GitHub URL.
+ * 3. Ask for GitHub path.
  * 4. Initialize repository.
  * 5. Add remote.
  * 6. Copy files to new repository.
@@ -59,7 +59,7 @@ module.exports = ( ckeditor5Path, workspaceRoot ) => {
 	let packageFullName;
 	let repositoryPath;
 	let packageVersion;
-	let gitHubUrl;
+	let gitHubPath;
 	let packageDescription;
 
 	return inquiries.getPackageName()
@@ -77,10 +77,10 @@ module.exports = ( ckeditor5Path, workspaceRoot ) => {
 		.then( result => {
 			packageVersion = result;
 
-			return inquiries.getPackageGitHubUrl( packageName );
+			return inquiries.getPackageGitHubPath( packageName );
 		} )
 		.then( result => {
-			gitHubUrl = result;
+			gitHubPath = result;
 
 			return inquiries.getPackageDescription();
 		} )
@@ -91,14 +91,14 @@ module.exports = ( ckeditor5Path, workspaceRoot ) => {
 			git.initializeRepository( repositoryPath );
 
 			log.out( `Adding remote ${ repositoryPath }...` );
-			git.addRemote( repositoryPath, gitHubUrl );
+			git.addRemote( repositoryPath, gitHubPath );
 
 			log.out( `Copying files into ${ repositoryPath }...` );
 
 			for ( let destination in fileStructure ) {
 				tools.copyTemplateFiles( fileStructure[ destination ], path.join( repositoryPath, destination ), {
 					'{{AppName}}': packageFullName,
-					'{{GitHubRepositoryPath}}': gitHubUrl,
+					'{{GitHubRepositoryPath}}': gitHubPath,
 					'{{ProjectDescription}}': packageDescription
 				} );
 			}
@@ -116,7 +116,7 @@ module.exports = ( ckeditor5Path, workspaceRoot ) => {
 				if ( !json.dependencies ) {
 					json.dependencies = {};
 				}
-				json.dependencies[ packageName ] = gitHubUrl;
+				json.dependencies[ packageName ] = gitHubPath;
 				json.dependencies = tools.sortObject( json.dependencies );
 
 				return json;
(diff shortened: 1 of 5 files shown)


Reinmar:
Please squash last commits. There shouldn't be ones like fixing variable names when that variable was introduced by previous commit.

dev/tasks/dev/utils/git.js
```diff
@@ -142,5 +143,20 @@ module.exports = {
...
+	/**
+	 * Adds remote to repository under specified path.
+	 *
+	 * @param {String} pluginName
+	 * @param {String} repositoryPath
+	 */
+	addRemote( pluginName, repositoryPath ) {
+		const addRemoteCommands = [
+			`cd ${ repositoryPath }`,
+			`git remote add ${ defaultOrigin } git@github.com:ckeditor/${ pluginName }.git`
```
Reinmar: GitHub url cannot be hardcoded. See the variables available in `create-package.js` - there's `gitHubUrl`.

Reinmar:
I left a comment in the code about one minor issue.

pomek:
@Reinmar, fixed. You can review one more time.

dev/tasks/dev/utils/git.js
```diff
@@ -142,5 +143,20 @@ module.exports = {
...
+	/**
+	 * Adds remote to repository under specified path.
+	 *
+	 * @param {String} repositoryPath
+	 * @param {String} repositoryUrl
+	 */
+	addRemote( repositoryPath, repositoryUrl ) {
+		const addRemoteCommands = [
+			`cd ${ repositoryPath }`,
+			`git remote add ${ defaultOrigin } ${ repositoryUrl }`
```
Reinmar: This is adding `ckeditor/ckeditor5-foo` as an origin. That's not a correct path.

Reinmar:
Now the code adds an incorrect URL as a remote.

pomek:
I did some fixes. Now, if a name will not contain a valid URL, `git.addRemote` will add string `https://github.com/` as a prefix for the name. Otherwise, I assume the name is a correct URL, so I will do nothing.

Reinmar:
I don't understand. Why may the variable not contain the default value that I can see during creation of a new package? There's the default value coded somewhere and this method should return it: https://github.com/ckeditor/ckeditor5/pull/197/files#diff-f087c1ace7f9141506883a96952cf5fdR80

pomek:
The variable `gitHubUrl` contains exactly the same string what user provided. It can be a correct URL, but it also can be some string (like default value in a configuration prompt - `ckeeditor/ckeeditor5-{package_name}`).

If I want to get a full URL - I have two ways:
1. Validate user's input - only correct URL should pass the validation,
2. Modify user's input - it still does not ensure that we'll build a correct URL.

Reinmar:
GitHub URL contains what the user was asked for – the path on GitHub. It's not meant to contain a working Git URL. These are two totally different things. So you need to transform one into another.

pomek:
I guess I [did it](https://github.com/ckeditor/ckeditor5/pull/197/files#diff-9c86580a845bd95e8a2694f275c4b70dR158).

dev/tasks/dev/utils/git.js
```diff
@@ -142,5 +144,26 @@ module.exports = {
...
 		tools.shExec( commitCommands.join( ' && ' ) );
+	},
+
+	/**
+	 * Adds remote to repository under specified path.
+	 *
+	 * @param {String} repositoryPath
+	 * @param {String} repositoryUrl
+	 */
+	addRemote( repositoryPath, repositoryUrl ) {
```
Reinmar: That's a wrong argument name. You get `gitHubUrl` and you turn it into `repositoryUrl`.

Reinmar:
Not yet. The `addRemote()` method should accept only one type of param and it's `gitHubPath`. Right now it accepts `gitHubUrl` which is parsed to an url and a protocol is checked... this is unnecessary. There's no use case for that.

So:

1. Fix naming of `gitHubUrl` – it's not a URL, but a path. Only inside `addRemote()` it's extended to a full URL.
2. Remove the parsing, because there's no use case for that.
3. Remember about updating the inquiry text from `GitHub URL` to `GitHub path`.

pomek:
> Remove the parsing, because there's no use case for that.

I'm not sure. Look on the pictures below:

image

image

It does not matter what user has provided. `gitHubUrl` can be a valid URL, but it can also be a not valid URL, but valid `gitHubPath`. 

`Create-package` tool won't throw any error. User can provide (as `gitHubUrl`) following data:
* `ckeditor/ckeditor5-package`,
* `https://github.com/ckeditor/ckeditor5-package`,
* `foo`

Since two ways are correct, creating package tool has to detect, what user has provided.

Reinmar:
> Create-package tool won't throw an error. User can provide (as gitHubUrl) following data:
> ...
> And nothing will happen.

It's not a problem for this ticket. You need to apply YAGNI here. There was no report that anyone typed there an incorrect value. Especially that we're going to use the value that's suggested automatically.

So again, KISS :)

dev/tasks/dev/tasks/create-package.js
```diff
@@ -79,7 +80,7 @@ module.exports = ( ckeditor5Path, workspaceRoot ) => {
 			return inquiries.getPackageGitHubUrl( packageName );
```
Reinmar: Here it's called `getPackageGitHubUrl` but below it's `gitHubPath`.

Reinmar:
Still a small R- because of inconsistent naming.

dev/tasks/dev/tasks/create-package.js
pomek: Ye, I was not sure whether I should also rename method or not…

Reinmar: Better do it now, once we realised that it's incorrect. Also the inquiry message should be changed as I already commented in https://github.com/ckeditor/ckeditor5/pull/197#issuecomment-225570029.

Reinmar:
All fine now :).

Issue #122, opened 2016-03-18
Period should be automatically appended to project description (if missing)


Reinmar:
... or we can explain it in the inquiry that "one sentence that ends with a period" :D

To test check:

```
> gulp create-package

[14:12:21] Starting 'create-package'...
? Enter package name without ckeditor5- prefix: x
? Enter application full name: x
? Enter package's initial version: 0.0.1
? Enter package's GitHub URL: ckeditor/ckeditor5-x
? Package description (one sentence):
```

Package description is then used in couple templates. In each of them it's expected that it ends with a period.

PR #202, opened 2016-06-13
Forcing period at the end of package description.

maxbarnas:
Resolves https://github.com/ckeditor/ckeditor5/issues/122

### DIFF
--- a/dev/tasks/dev/utils/inquiries.js
+++ b/dev/tasks/dev/utils/inquiries.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const inquirer = require( 'inquirer' );
+const sanitize = require( './sanitize' );
 const DEFAULT_PLUGIN_NAME_PREFIX = 'ckeditor5-';
 const DEFAULT_PLUGIN_VERSION = '0.0.1';
 const DEFAULT_GITHUB_URL_PREFIX = 'ckeditor/';
@@ -70,19 +71,8 @@ module.exports = {
 				name: 'description',
 				message: 'Package description (one sentence, must end with period):'
 			} ], ( answers ) => {
-				resolve( appendPeriodIfMissing( answers.description ) );
+				resolve( sanitize.appendPeriodIfMissing( answers.description ) );
 			} );
 		} );
 	}
 };
-
-function appendPeriodIfMissing( text ) {
-	text = text ? text.trim() : '';
-	const length = text.length;
-
-	if ( length > 0 && text[ length - 1 ] !== '.' ) {
-		text += '.';
-	}
-
-	return text;
-}
--- a/dev/tasks/dev/utils/inquiries.js
+++ b/dev/tasks/dev/utils/inquiries.js
@@ -77,10 +77,10 @@ module.exports = {
 };
 
 function appendPeriodIfMissing( text ) {
-	text = text ? String.prototype.trim.call( text ) : '';
+	text = text ? text.trim() : '';
 	const length = text.length;
 
-	if ( length > 0 && text[length - 1] !== '.' ) {
+	if ( length > 0 && text[ length - 1 ] !== '.' ) {
 		text += '.';
 	}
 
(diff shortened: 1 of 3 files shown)


dev/tasks/dev/utils/inquiries.js
```diff
@@ -68,10 +68,21 @@ module.exports = {
...
 			} );
 		} );
 	}
 };
+
+function appendPeriodIfMissing( text ) {
+	text = text ? String.prototype.trim.call( text ) : '';
+	const length = text.length;
+
+	if ( length > 0 && text[length - 1] !== '.' ) {
```
Reinmar: Code style issue (not sure why JSCS haven't caught it) – there should be spaces inside `[]`.

dev/tasks/dev/utils/inquiries.js
```diff
@@ -68,10 +68,21 @@ module.exports = {
...
 			} ], ( answers ) => {
-				resolve( answers.description || '' );
+				resolve( appendPeriodIfMissing( answers.description ) );
 			} );
 		} );
 	}
 };
+
+function appendPeriodIfMissing( text ) {
+	text = text ? String.prototype.trim.call( text ) : '';
```
Reinmar: Why not simply `text.trim()`?

dev/tasks/dev/utils/inquiries.js
maxbarnas: Yep, mine was silent too on this.

dev/tasks/dev/utils/inquiries.js
maxbarnas: This might be paranoid approach, but if text is `Number`, it will throw SyntaxError because `Number` has no `trim()` method. With the solution above number will be cast to string.

Reinmar:
Make sure to run dev utils tests (`npm test`). Also, you can think about a simple test for that too.

dev/tasks/dev/utils/inquiries.js
Reinmar: Do you know any real case in which argument of that function may be a number? I don't see any. It's used in one place and its parameter is called "text". So KISS.

If there was any possibility that `answers.description` could be a number, we should've fixed it there, not in this function.

dev/tasks/dev/utils/sanitize.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+'use strict';
+
+module.exports = {
+	appendPeriodIfMissing( text ) {
+		text = text ? text.trim() : '';
+		const length = text.length;
+
+		if ( length > 0 && text[ length - 1 ] !== '.' ) {
```
Reinmar: `text.match( /\.$/ )`? ;)

Reinmar: Or actually: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

dev/tasks/dev/utils/sanitize.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+module.exports = {
+	appendPeriodIfMissing( text ) {
+		text = text ? text.trim() : '';
```
Reinmar: I'd keep defaulting to an empty string outside this function – in the `inquiries.js`. E.g. if we add more functions like this, we won't need to check the value in all of them. Plus, the documentation can be simpler – this function accepts a string and nothing more.

dev/tests/dev/sanitize.js
```diff
@@ -0,0 +1,49 @@
...
+const expect = chai.expect;
+
+describe( 'utils', () => {
+	describe( 'sanitize', () => {
+		describe( 'appendPeriodIfMissing', () => {
+			it( 'should be defined', () => expect( sanitize.appendPeriodIfMissing ).to.be.a( 'function' ) );
+
+			it( 'should trim whitespace/new lines to empty string', () => {
+				// jscs: disable validateQuoteMarks
+				const sanitized = sanitize.appendPeriodIfMissing( "\n\t\r " );
```
Reinmar: Why haven't you used `'\n\t\r '`?

PR #207, opened 2016-06-17
Improved compatibility with code coverage.

### LINKED ISSUE #190: Bring compatibility with code coverage
Apparently, Istanbul (even the latest one) doesn't support some ES6 constructs that we use.  This means that we're unable to run jobs, because they take ridiculous amount of time.

There are two things we need to do:

1. Fork benderjs-coverage to use a standard Istanbul (benderjs-coverage uses a hacked one, because, AFAIR, CKE4 requires that) and in its latest version.
2. Check what ES6 features break Istanbul and stop using them. My guess will be that deconstructuring in default params.

See https://github.com/benderjs/benderjs/issues/253 for more.

oskarwrobel:
Fixes #190 

Related: https://github.com/ckeditor/ckeditor5-engine/pull/497

### DIFF
--- a/dev/tests/build/utils.js
+++ b/dev/tests/build/utils.js
@@ -188,20 +188,6 @@ describe( 'build-utils', () => {
 		} );
 	} );
 
-	describe( 'getBabelPlugins', () => {
-		it( 'should return plugins for amd format', () => {
-			expect( utils.getBabelPlugins( 'amd' ) ).to.be.an( 'array' );
-		} );
-
-		it( 'should throw an exception when incorrect format is provided', () => {
-			const format = 'incorrect-format';
-
-			expect( () => {
-				utils.getBabelPlugins( format );
-			} ).to.throw( Error, `Incorrect format: ${ format }` );
-		} );
-	} );
-
 	describe( 'getConversionStreamGenerator', () => {
 		beforeEach( () => {
 			sandbox.stub( utils, 'getBabelOptionsForSource', () => 'src' );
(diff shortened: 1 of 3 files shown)


dev/tests/build/utils.js
```diff
@@ -188,20 +188,6 @@ describe( 'build-utils', () => {
 		} );
 	} );
 
-	describe( 'getBabelPlugins', () => {
```
Reinmar: Why removing a correct test?

oskarwrobel: Was duplicated.

Reinmar:
I pushed a commit upgrading also other Babel plugins, because there was incompatibility between packages I had.

Reinmar:
I noticed that with this set of Babel plugins it also transpiles arrow functions :( That's sad. Can we avoid that?

Reinmar:
To confirm - check transpiled version of `utils/locale.js` and `this` that appeared in the constructor.

Reinmar:
Hm... on the other hand – that doesn't happen in .e.g. `undo/undoengine.js`  in the `init()` function.

Reinmar:
screen shot 2016-06-20 at 14 39 26

:(

Reinmar:
This is due to the destructuring transpilation. If we disable it then default params with destructuring still don't work.

Reinmar:
https://github.com/gotwarlost/istanbul/issues/613

Reinmar:
https://github.com/estools/escodegen/issues/296 has a PR, but escodegen looks to be quite dead ;/. Things don't look super well.

I'm afraid we'll need to just get rid of default params from our code :( Though, it's worth checking if we can't use default params with destructuring or default params in general or destructuring in general.

oskarwrobel:
I also think that we should avoid transforming code in build task as long as possible. Babel might inject some inline helpers to each file what could increase size of a bundle. We can try to use https://babeljs.io/docs/plugins/external-helpers/ but it makes everything more complicated.

Issue #220, opened 2016-06-23
gulp create-package throws an error


pomek:
During `gulp create-package`, after command `npm install` (called in a new created directory), Gulp throws errors:

```
[10:41:29] Linking ckeditor5-docs-builder to node_modules...
[10:41:29] Running npm install in ckeditor5-docs-builder.
/// …
npm WARN ckeditor5-docs-builder@0.0.1 license should be a valid SPDX license expression

[10:42:29] Finished 'create-package' after 1.97 min
[10:42:29] 'create-package' errored after 1.97 min
```

This issue can be solved the same way like in https://github.com/ckeditor/ckeditor5/commit/a8411f98ac3e9591744b1736b21a87a768e39ca2.

PR #221, opened 2016-06-23
Removed invalid returning from create-package task.

pomek:
See: https://github.com/ckeditor/ckeditor5/commit/a8411f98ac3e9591744b1736b21a87a768e39ca2

Fixes: #220


Reinmar:
I'd normally expect a test verifying the change, especially that these bugs went unnoticed, but I also expect that testing this gulp task will be unpleasant.

PR #225, opened 2016-07-04
Running task on all subrepositories

maxbarnas:
For now there are just few sample tasks:
- `bump-year.js` updates Copyright date
- `git-commit.js` commits changes in all `ckeditor5-*` repositories
- `git-push.js` pushes all the changes from `ckeditor5-*` repositories

I moved some of the utils files to common directory `ckeditor5/dev/lib`

Note: This is still work in progress


dev/lib/git.js
```diff
@@ -158,5 +158,33 @@ module.exports = {
...
+	/*
+	 * Creates commit on repository under specified path
+	 *
+	 * @param {String} message
+	 * @param {String} repositoryPath
+	 */
+	commit( message, repositoryPath ) {
+		const commitCommands = [
+			`cd ${ repositoryPath }`,
+			`git commit --dry-run --all --message "${ message }"`
```
Reinmar: Why dry-run? This means that this method won't do anything.

maxbarnas: Correct, forgot to remove it after human testing.

dev/tasks/exec/functions/bump-year.js
```diff
@@ -0,0 +1,34 @@
...
+ * Replaces license date in source files with new date
+ *
+ * @param {String} workdir
+ * @returns {Object} stream
+ */
+module.exports = ( workdir ) => {
+	// Change this to correct year
+	const year = '2017';
+
+	const licenseRegexp = /(@license Copyright \(c\) 2003-)[0-9]{4}/g;
```
Reinmar: First capture group could be non-capturing (`(?:...)`). Also, I would add the next token after that piece which is `, CKSource` so we are sure we only touched our headings.

dev/tasks/exec/functions/git-commit.js
```diff
@@ -0,0 +1,28 @@
...
+ * For licensing, see LICENSE.md.
+ */
+
+ 'use strict';
+
+const git = require( '../utils/git' );
+const PassThrough = require( 'stream' ).PassThrough;
+
+/**
+ * Adds only modified files to git repository and commits them with provided message
```
Reinmar: Some usage example would be helpful here.

dev/tasks/exec/functions/git-commit.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @param {String} workdir
+ * @param {Object} params
+ * @returns {Object} stream
```
Reinmar: Do we need to return a stream if this is a synchronous function? I guess we don't need to return anything to gulp.

dev/tasks/exec/functions/git-commit.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+module.exports = ( workdir, params ) => {
+	if ( !( params.message || params.m ) ) {
+		throw new Error( 'You must provide commit message with parameter: --message | -m ' );
+	}
+	const message = params.message || params.m;
```
Reinmar: First define `const message` then check whether it was passed.

dev/tasks/exec/functions/git-push.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+ 'use strict';
```
Reinmar: Space before.

dev/tasks/exec/functions/git-commit.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+ 'use strict';
```
Reinmar: Space before.

dev/tasks/exec/functions/bump-year.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+ 'use strict';
```
Reinmar: Space before.

dev/lib/git.js
```diff
@@ -158,5 +158,33 @@ module.exports = {
 		];
 
 		tools.shExec( addRemoteCommands.join( ' && ' ) );
+	},
```
Reinmar: Missing blank line.

dev/tasks/exec/package.json
```diff
@@ -0,0 +1,12 @@
+{
```
Reinmar: No needed for this file yet.

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+
+			try {
+				execTask = require( `./functions/${ options.task }` );
+			}
+			catch ( error ) {
+				log.err( error );
+			}
+
+			if ( execTask ) {
+				return exec( execTask, ckeditor5Path, packageJSON, config.WORKSPACE_DIR, options );
```
Reinmar: Why are all those params passed to a task? Do we have a use case for them already?

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+				alias: { t: 'task' },
+				stopEarly: false
+			} );
+
+			let execTask;
+
+			try {
+				execTask = require( `./functions/${ options.task }` );
+			}
+			catch ( error ) {
```
Reinmar: `catch` should be in the same line as `}` unless it has a comment before it.

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+	log.configure(
+		( msg ) => gutil.log( msg ),
+		( msg ) => gutil.log( gutil.colors.red( msg ) )
+	);
+
+	const tasks = {
+		execOnRepositories() {
+			// Omit `gulp exec` part of arguments
+			const options = minimist( process.argv.slice( 3 ), {
+				alias: { t: 'task' },
```
Reinmar: Do we need aliases? The same with `-m|--message` in the commit function. Let's KISS and avoid those.

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,53 @@
...
+
+const exec = require( './tasks/exec' );
+const log = require( '../../lib/log' );
+
+module.exports = ( config ) => {
+	const ckeditor5Path = process.cwd();
+	const packageJSON = require( '../../../package.json' );
+
+	// Configure logging.
+	log.configure(
```
Reinmar: I think this code isn't needed here... it'd be executed also by some other tasks, what means that it perhaps should be moved to `gulpfile.js` to be executed once or to `lib/log`. But please check if tests still pass.

dev/tasks/exec/tasks/exec.js
```diff
@@ -0,0 +1,56 @@
...
+const path = require( 'path' );
+const log = require( '../../../lib/log' );
+const merge = require( 'merge-stream' );
+
+/**
+ * @param {Function} execTask Task to use on each dependency.
+ * @param {String} ckeditor5Path Path to main CKEditor5 repository.
+ * @param {Object} packageJSON Parsed package.json file from CKEditor5 repository.
+ * @param {String} workspaceRoot Relative path to workspace root.
+ * @param {Boolean} dryRun
```
Reinmar: This looks like out of sync.

dev/tasks/exec/tasks.js
maxbarnas: They are used in dev/tasks/exec/functions/git-commit.js

Reinmar: OK, it looks like all this is actually used in the `exec()` function. BTW. I would move `exec()` to this file. It doesn't make sense to keep it in a separate one if there won't be more tasks in this file anyway.

dev/tasks/exec/tasks/exec.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+module.exports = ( execTask, ckeditor5Path, packageJSON, workspaceRoot, params ) => {
+	const workspaceAbsolutePath = path.join( ckeditor5Path, workspaceRoot );
+
+	// Get all CKEditor dependencies from package.json.
+	const dependencies = tools.getCKEditorDependencies( packageJSON.dependencies );
+	const streams = merge();
+
+	if ( dependencies ) {
```
Reinmar: Why checking dependencies and directories? Checking the latter isn't enough?

dev/tasks/exec/tasks/exec.js
```diff
@@ -0,0 +1,56 @@
...
+				}
+			}
+		} else {
+			log.out( 'No CKEditor5 plugins in development mode.' );
+		}
+	} else {
+		log.out( 'No CKEditor5 dependencies found in package.json file.' );
+	}
+
+	return streams;
```
Reinmar: It's a single stream that merges multiple ones. Naming it with plural "streams" confused me at first (I thought it's an array and was already commenting that gulp doesn't support such return value AFAIK).

dev/tasks/exec/tasks/exec.js
maxbarnas: From what I understand they serve different purpose - `log.out(...)` calls below explain the difference.

dev/tasks/exec/tasks/exec.js
maxbarnas: So, `merged` or `mergedStream` seems to be more correct?

dev/tasks/exec/tasks/exec.js
```diff
@@ -0,0 +1,56 @@
...
+		const directories = tools.getCKE5Directories( workspaceAbsolutePath );
+
+		if ( directories.length ) {
+			for ( let dependency in dependencies ) {
+				const repositoryURL = dependencies[ dependency ];
+				const urlInfo = git.parseRepositoryUrl( repositoryURL );
+				const repositoryAbsolutePath = path.join( ckeditor5Path, 'node_modules', dependency );
+
+				// Check if repository's directory already exists.
+				if ( directories.indexOf( urlInfo.name ) > -1 ) {
```
Reinmar: I don't understand what this code does (and most importantly – why we need to check something).

Reinmar:
Time for the first review :)

1. Always end sentence with a period (in commit messages and in code comments).
2. I thought that `dev/lib` should rather be `dev/utils`. We tend to group such functions under word `utils` in other places in the code.
3. New functions should have some tests. At least those simpler ones in utils and the main task.
4. It seems that you committed an empty file `dev/tasks/exec/functions/remove-use-strict.js`.
5. We need the option to execute a task on a single repo so you can easily check its result before applying it to all repos.
6. And I left some comments in the code.

dev/tasks/exec/tasks/exec.js
maxbarnas: It started as a copy of `ckeditor5/dev/tasks/dev/tasks/status.js` - there you can find most of the reasons behind this structure.

dev/tasks/exec/tasks/exec.js
Reinmar: But that doesn't explain why we need to check all that. Can't we just take CKEditor directories? Perhaps I just don't understand what the existing functions do but that may mean that they should be improved. The logic you put here, with getting deps and then dirs should be enclosed in some other, reusable function.

dev/tasks/exec/tasks/exec.js
Reinmar: `mergedStream` sounds fine.

dev/tasks/exec/tasks/exec.js
maxbarnas: Difference lies in listing regular repositories, and checking repositories in development mode. But I am happy to extract that and make it nicer.

dev/tasks/exec/functions/bump-year.js
maxbarnas: AFAIK with non-capturing grouping I have to repeat searched parts in replacement parameter. With capturing groups I can use what I just captured to recreate that line of text. Or I am missing something.

dev/tasks/exec/functions/git-commit.js
maxbarnas: I can't mix streamable and non-streamable functions in current task runner. Synchronous functions must return dummy stream. Otherwise Gulp won't notice that we finished the task.

dev/tasks/exec/tasks.js
maxbarnas: The `ckeditor5/dev/lib/log` module is used only in `ckeditor5/dev/tasks/dev/tasks.js` and `ckeditor5/dev/tasks/exec/tasks.js`. Seems like it is close copy of `ckeditor5-utils/log`.
Having that in mind we could consider removing all that "configuration" and hardcode it into `ckeditor5/dev/lib/log`. What is your opinion about this?

dev/tasks/exec/functions/git-commit.js
Reinmar: Ah, true. Now I get it.

dev/tasks/exec/tasks.js
Reinmar: Harcoding will be the simplest option. We don't have a use case for configuring that logger other than tests. That's why I asked whether they pass.

dev/tasks/exec/utils/gitignore-filter.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+const fs = require( 'fs' );
+const filter = require( 'gulp-filter' );
+const gitignore = require( 'parse-gitignore' );
```
Reinmar: That variable name is confusing. It's a function that does something so it should contain a verb – `parseGitignore` for instance.

dev/tasks/exec/utils/gitignore-filter.js
```diff
@@ -0,0 +1,27 @@
...
+const PassThrough = require( 'stream' ).PassThrough;
+
+module.exports = () => {
```
Reinmar: If a module exports a function only it (the file) should be called after it. Also, the function should be named for easier debugging. This would be `filtergitignore.js` which exports `function filterGitignore()`.

dev/tasks/exec/functions/bump-year.js
Reinmar: Yep, you're right. I thought that non capturing group won't be replaced... but that obviously doesn't make sense :D.

dev/tasks/exec/functions/git-commit.js
Reinmar: So the missing thing is documenting that the return value type is a Stream not a simple Object.

dev/tasks/exec/tasks.js
```diff
@@ -13,29 +13,39 @@ const log = require( '../../utils/log' );
 const tools = require( '../../utils/tools' );
 const git = require( '../../utils/git' );
 
-module.exports = ( config ) => {
+/**
+ * Run task over ckeditor5 repositories.
```
Reinmar: This API doc isn't formatted correctly. After each paragraph there should be a blank line. And code samples should be indented with two tabs (and no spaces).

dev/tasks/exec/functions/git-commit.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+'use strict';
+
+const git = require( '../utils/git' );
+const PassThrough = require( 'stream' ).PassThrough;
+
+/**
+ * Adds only modified files to git repository and commits them with provided message
+ * Example: gulp exec --task git-commit --message "Commit message."
```
Reinmar: Inline code samples need to be wrapped with the "backtick" characters..

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,121 @@
...
+ * @returns {Stream} Merged stream of processed files.
+ */
+function execute( execTask, ckeditor5Path, packageJSON, workspaceRoot, parameters ) {
+	const workspacePath = path.join( ckeditor5Path, workspaceRoot );
+	const devDirectories = getCKE5DevDirectories( workspacePath, packageJSON, ckeditor5Path );
+	const shouldRunOnce = parameters[ 'one-repo' ];
+	const mergedStream = merge();
+
+	for ( let i = 0, len = devDirectories.length; i < len; i++ ) {
+		const devDir = devDirectories[i];
```
Reinmar: Missing spaces inside `[]`.

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,121 @@
...
+ * @param {String} ckeditor5Path Path to main CKEditor5 repository.
+ * @param {Object} packageJSON Parsed package.json file from CKEditor5 repository.
+ * @param {String} workspaceRoot Relative path to workspace root.
+ * @param {Object} parameters Parameters provided to the task via command-line.
+ * @returns {Stream} Merged stream of processed files.
+ */
+function execute( execTask, ckeditor5Path, packageJSON, workspaceRoot, parameters ) {
+	const workspacePath = path.join( ckeditor5Path, workspaceRoot );
+	const devDirectories = getCKE5DevDirectories( workspacePath, packageJSON, ckeditor5Path );
+	const shouldRunOnce = parameters[ 'one-repo' ];
```
Reinmar: It may be better to let the user specify in which repository this task will be executed. In this case the user won't know what to expect.

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,121 @@
...
+ * Run task over ckeditor5 repositories.
+ * Example:
+ *	 gulp exec --task task-name
+ * Example of running task just for one repository:
+ *   gulp exec --task task-name --one-repo
+ *
+ * @param {Object} config Task runner configuration.
+ * @returns {Stream} Stream with processed files.
+ */
+module.exports = function tasks( config ) {
```
Reinmar: In other tasks we use an arrow function here.

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,121 @@
...
+				if ( parameters.task ) {
+					task = require( `./functions/${ parameters.task }` );
+				} else {
+					throw new Error( 'Missing task parameter: --task task-name' );
+				}
+			} catch ( error ) {
+				log.err( error );
+			}
+
+			if ( task ) {
```
Reinmar: Why checking the task for the second time?

dev/tasks/exec/tasks.js
```diff
@@ -0,0 +1,121 @@
...
+ * @param {String} ckeditor5Path Absolute path to ckeditor5 root directory.
+ * @returns {Array} Array of objects.
+ */
+function getCKE5DevDirectories( workspacePath, packageJSON, ckeditor5Path ) {
+	const directories = tools.getCKE5Directories( workspacePath );
+	const dependencies = tools.getCKEditorDependencies( packageJSON.dependencies );
+
+	let devDirectories = [];
+
+	for ( let dependency in dependencies ) {
```
Reinmar: Looks like a case for some `reduce()` or `map().filter()` ;). But you can leave it this way too.

dev/tasks/exec/utils/filtergitignore.js
```diff
@@ -0,0 +1,27 @@
...
+const PassThrough = require( 'stream' ).PassThrough;
+
+module.exports = function filterGitignore() {
+	const fp = '.gitignore';
+
+	if ( !fs.existsSync( fp ) ) {
+		return new PassThrough( { objectMode: true } );
+	}
+
+	let glob = parseGitignore( fp );
```
Reinmar: `const` in both cases. Don't use `let` unless you expect the variable to change.

Issue #149, opened 2016-05-04
We can remove 'use strict' since we use ES6 modules


Reinmar:
AFAIR (needs to be confirmed) ES6 modules enable strict mode out of the box.

oskarwrobel:
According to documentation http://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code 

> Module code is always strict mode code.

So looks like we can remove 'use strict' from source code, but probably we need to add it automatically to CommonJS and AMD versions of builded packages.

We can also configure eslint for it http://eslint.org/docs/rules/strict

Reinmar:
Thanks for confirming this.

> We can also configure eslint for it http://eslint.org/docs/rules/strict

We use JSHint (and JSCS).

Reinmar:
TODO: Update https://github.com/ckeditor/ckeditor5-design/wiki/Code-Style#strict-mode.

Reinmar:
Wiki updated: https://github.com/ckeditor/ckeditor5-design/wiki/Code-Style/_compare/c5276b3bec1bcd7a8f8408c2cf0be0e3d6735417...466587838d548bb9e716eb67e32c2575719c840d

PR #237, opened 2016-07-12
We can remove 'use strict' since we use ES6 modules

maxbarnas:
Fixes: https://github.com/ckeditor/ckeditor5/issues/149

Allows for `'use strict';` directive removal from our source code. Respects `.gitignore`'s rules. What it also does is update `.jshintrc` option `"strict"` with `"implied"` value - instead of usual `"global"`.

Task will be considered done when all code in every `ckeditor5-*` package will be processed with this script.

Example of usage:
```
gulp exec --task remove-use-strict
```

I've also added support for processing root package `ckeditor5`. This can be achieved with:
```
gulp exec --task remove-use-strict --include-root
```

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,13 +1,12 @@
 {
-	"browser": true,
-	"esnext": true,
-
-	"immed": true,
-	"loopfunc": true,
-	"noarg": true,
-	"nonbsp": true,
-	"undef": true,
-	"unused": true,
-	"strict": "global",
-	"varstmt": true
+  "browser": true,
+  "esnext": true,
+  "immed": true,
+  "loopfunc": true,
+  "noarg": true,
+  "nonbsp": true,
+  "undef": true,
+  "unused": true,
+  "strict": "implied",
+  "varstmt": true
 }
--- a/dev/tests/utils/ckeditor5-dirs.js
+++ b/dev/tests/utils/ckeditor5-dirs.js
@@ -141,6 +141,29 @@ describe( 'utils', () => {
 						repositoryPath: '/workspace/path/ckeditor5/node_modules/ckeditor5-core'
 					} );
 			} );
+
+			it( 'should return only ckeditor5 directories in development mode, including root directory', () => {
+				sandbox.stub( ckeditor5Dirs, 'getDirectories', () => sourceDirectories );
+				sandbox.stub( ckeditor5Dirs, 'getDependencies', () => dependencies );
+				sandbox.stub( git, 'parseRepositoryUrl' ).returns( repositoryInfo );
+				const includeRoot = true;
+
+				const directories = ckeditor5Dirs.getDevDirectories( workspacePath, packageJSONDependencies, ckeditor5Path, includeRoot );
+
+				expect( directories.length ).equal( 3 );
+				expect( directories[ 0 ] ).eql( {
+						repositoryURL: 'ckeditor/ckeditor5-plugin-image',
+						repositoryPath: '/workspace/path/ckeditor5/node_modules/ckeditor5-plugin-image'
+					} );
+				expect( directories[ 1 ] ).eql( {
+						repositoryURL: 'ckeditor/ckeditor5-core',
+						repositoryPath: '/workspace/path/ckeditor5/node_modules/ckeditor5-core'
+					} );
+				expect( directories[ 2 ] ).eql( {
+						repositoryURL: 'ckeditor/ckeditor5',
+						repositoryPath: '/workspace/path/ckeditor5'
+					} );
+			} );
 		} );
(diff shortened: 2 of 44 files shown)


.jshintrc
```diff
@@ -8,6 +8,6 @@
 	"nonbsp": true,
 	"undef": true,
 	"unused": true,
-	"strict": "global",
+	"strict": "implied",
 	"varstmt": true
 }
```
szymonkups: Also `.jshintrc` file from `test/` directory should be changed since we write tests using ES6 and after this task execution `'use strict';` is removed from test files. Additionally, this PR should remove `'use strict';` from test files located in `ckeditor5` repository.

szymonkups: There is also `.jshintrc` file in `dev/` directory that should be changed. We can use `implied` there too I think.

maxbarnas: I just added `tests/` and `dev/` directories to script.

dev/tests/utils/ckeditor5-dirs.js
```diff
@@ -141,6 +139,29 @@ describe( 'utils', () => {
...
+
+			it( 'should return only ckeditor5 directories in development mode, including root directory', () => {
+				sandbox.stub( ckeditor5Dirs, 'getDirectories', () => sourceDirectories );
+				sandbox.stub( ckeditor5Dirs, 'getDependencies', () => dependencies );
+				sandbox.stub( git, 'parseRepositoryUrl' ).returns( repositoryInfo );
+				const includeRoot = true;
+
+				const directories = ckeditor5Dirs.getDevDirectories( workspacePath, packageJSONDependencies, ckeditor5Path, includeRoot );
+
+				expect( directories.length ).equal( 3 );
```
Reinmar: We use `to.equal()` to make it a correct statement.

dev/tests/utils/ckeditor5-dirs.js
```diff
@@ -141,6 +139,29 @@ describe( 'utils', () => {
...
+				expect( directories[ 0 ] ).eql( {
```
Reinmar: Please don't use `eql()`. I find it super confusing that it's something else than `equal()` (it's deep equal right?).

package.json
```diff
@@ -75,6 +75,7 @@
     "parse-gitignore": "^0.2.0",
     "pretty-bytes": "^3.0.1",
     "regenerator-runtime": "^0.9.5",
+    "parse-gitignore": "^0.2.0",
```
Reinmar: This is doubled.

dev/utils/ckeditor5-dirs.js
```diff
@@ -72,18 +70,26 @@ module.exports = {
...
 	 * @param {String} ckeditor5Path Absolute path to ckeditor5 root directory.
+	 * @param {Boolean} includeRoot Include main `ckeditor5` package.
 	 * @returns {Array.<Object>}
 	 */
-	getDevDirectories( workspacePath, packageJSON, ckeditor5Path ) {
+	getDevDirectories( workspacePath, packageJSON, ckeditor5Path, includeRoot ) {
 		const directories = this.getDirectories( workspacePath );
 		const dependencies = this.getDependencies( packageJSON.dependencies );
 
+		if ( includeRoot ) {
```
Reinmar: I don't like implementation of this option. The root repo doesn't have to be `ckeditor5`. What if someone forks this repo? Also, in the loop you shouldn't check `isRootDep` in every iteration – the dependencies should be correctly filled before you get to the loop (so in this condition).

Reinmar: BTW. I would consider moving this condition to `getDevDirectories()`.

Reinmar: Right... I'm commenting part of getDevDirectories :D Sorry for the last comment.

Reinmar: As for the implementation, I would move the whole condition after the loop and simply push (or better – shift) the CWD there.

dev/tests/utils/ckeditor5-dirs.js
maxbarnas: Changed to `to.deep.equal()`

Reinmar:
Unfortunately this task changes also files in `dev/*` which are Node.js modules and AFAIK CJS modules don't have strict mode implied in them.

Reinmar:
I've had to revert some changes made in this branch and by mistake I rebased the whole thing. 

But the Node.js thing is now fixed and I also realised that gulpfiles and `bender.js` need to be fixed as well, plus it'll be good to reformat remaining `.jshintrc` and `.jscsrc` files to match the touched ones.

Issue #211, opened 2016-06-22
Use lower case for const


pjasiun:
Execute: https://github.com/ckeditor/ckeditor5-design/issues/148.

PR #240, opened 2016-07-14
Changed const to lower case.

oskarwrobel:
Fixes: #211 

Related to:
- https://github.com/ckeditor/ckeditor5-utils/pull/48
- https://github.com/ckeditor/ckeditor5-typing/pull/23
- https://github.com/ckeditor/ckeditor5-engine/pull/528/files
- https://github.com/ckeditor/ckeditor5-basic-styles/pull/7/files

### DIFF
--- a/dev/tasks/dev/utils/inquiries.js
+++ b/dev/tasks/dev/utils/inquiries.js
@@ -7,23 +7,23 @@
 
 const inquirer = require( 'inquirer' );
 const sanitize = require( './sanitize' );
-const DEFAULT_PLUGIN_NAME_PREFIX = 'ckeditor5-';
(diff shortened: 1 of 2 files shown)


scofalik:
I am afraid that you have not understood the issue correctly. The issue was about "string constants" which we use throught the project, not `const` variables.

So we are talking about things like:
https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/treewalker.js#L306
https://github.com/ckeditor/ckeditor5-utils/blob/master/src/comparearrays.js#L38
etc.

I'll close this PR because I don't think we will change those things that are changed by it, but we can leave it for some time and discuss it when people come back from holidays.

Issue #216, opened 2016-06-22
Extract linting tasks to a separate package


Reinmar:
Part of #215. Plus, requires changes in all existing packages and the `gulp create-package` task.

PS. Don't break automatic git hooks installation.

Reinmar:
I expect the `gulpfile.js` in all packages (except `ckeditor5` and `ckeditor5-utils`) to look like this:

```js
/* jshint browser: false, node: true, strict: true */

'use strict';

const gulp = require( 'gulp' );
const ckeditor5Lint = require( 'ckeditor5-dev-task-lint' )( {
	ROOT_DIR: '.',

	// Files ignored by jshint and jscs tasks. Files from .gitignore will be added automatically during tasks execution.
	IGNORED_FILES: [
		'src/lib/**'
	]
} );

gulp.task( 'lint', ckeditor5Lint.lint );
gulp.task( 'lint-staged', ckeditor5Lint.lintStaged );
gulp.task( 'pre-commit', [ 'lint-staged' ] );
```

This shows that the package should accept root dir and ignored files as a config option and also that it shouldn't register gulp task. Rather, it should allow to register them under any name (so e.g. tasks exposed by this package will be useful in completely different project where CKEditor is only a dependency.

PR #265, opened 2016-07-25
Use ckeditor5-dev-task-lint package for linting

maxbarnas:
Resolves: #216.


maxbarnas:
I'll review and update `package.json` dependencies. Please do not close this PR before that.

package.json
```diff
@@ -60,8 +59,6 @@
     "gzip-size": "^3.0.0",
     "inquirer": "^0.11.0",
     "jsdoc": "^3.4.0",
-    "jshint": "^2.9.1",
```
Reinmar: This can't be removed. Please see https://github.com/ckeditor/ckeditor5-dev-utils/issues/5#issuecomment-235074578

Reinmar: It works without jshint listed here because ckeditor5-dev-utils requires it, so it gets installed in `ckeditor5` because it's the topmost package. But it may not be reliable unfortunately. If there's some conflict with some other package jshint may be installed in other place and it's going to break.

So, we could live without jshint being listed here and there's a high chance that nothing will break, but it's unsafe.

Reinmar: Turns out I missed one thing – that ckeditor5-dev-lint installs jshint as its normal dependency which will work even when there are version conflicts. See also https://github.com/ckeditor/ckeditor5-dev-utils/issues/5#issuecomment-235244998.

Reinmar:
All fine. Sorry for the confusion with jshint.

PR #290, opened 2016-08-10
Using compiler instead of builder

### LINKED ISSUE #289: Use ckeditor5-dev-compiler instead of ckeditor5-dev-builder
After changing `ckeditor5-dev-builder` name to `ckeditor5-dev-compiler` we should use compiler, and name the step as `compiling`. We also should change name of the gulp task to `gulp compile`.

szymonkups:
Fixes #289.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -72,8 +72,8 @@ gulp.task( 'bundle', ( callback ) => {
 	);
 } );
 
-// Build tasks.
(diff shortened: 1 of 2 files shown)


pomek:
Please remember about update [Wiki](https://github.com/ckeditor/ckeditor5/wiki/Development-Environment). Should use task `compile` instead of `build`.

Issue #260, opened 2016-07-22
Testing samples and releases


Reinmar:
**tl;dr: How to test releases (bundles) and perform integration tests using the set of samples we'll create for CKEditor 5 packages. And much more...**

## Goals

The goal is to enhance our testing suite with:

* integration tests (so far we're mainly writing unit tests),
* mechanisms which would allow testing CKEditor bundles that we'll be officially shipping (there will be many kind of builds and it sometimes happen that something fail in a build, while it work in a dev version).

At the same time we know that we'll have a set of samples (most packages will have some) and we
want to cover these samples by tests too.

## The Idea

All samples of the editor that we'll use on our websites and in the documentation will come from the samples that we'll keep in the repos and which we'll cover by tests.

This means that every important functionality will have its integration tests in form of that sample. Also, these samples will have a form which may allow testing them using a bundled editor.

So the idea is to **use the samples and heir testing mechanisms for testing releases**. This should provide us a sufficient coverage level.

### Sample + Tests

A typical sample JS file will look like this:

```js
ClassicEditor.create( document.querySelector( '#editor' ), {
    features: [ 'paragraph', 'typing', ... ],
    toolbar: [ ... ]
} )
.then( function( editor ) {
    window.editor = editor;
} )
.catch( function( err ) {
    console.error( err.stack );
} );
```

In order to run this as a test we need to import `ClassicEditor`. The `import` statement should not be present in a sample's code in documentation, so it needs to be removed when building docs. To avoid magic, we can mark it with some comment:

```js
import ClassicEditor from '/ckeditor5/editor-classic/classic.js'; // docs:remove

ClassicEditor.create( document.querySelector( '#editor' ), {
    features: [ 'paragraph', 'typing', ... ],
    toolbar: [ ... ]
} )
.then( function( editor ) {
    window.editor = editor;
} )
.catch( function( err ) {
    console.error( err.stack );
} );
```

Then, we need to add code of the tests, but:

* tests need to wait until the editor is ready,
* tests should fail if the editor hasn't been initialised correctly.

Surprisingly, these points are tricky because the code needs to stay simple for the docs purposes and needs to accommodate some additional code for tests.

One of the ideas we had was to do this:

```js
import ClassicEditor from '/ckeditor5/editor-classic/classic.js';

before( function() {
    // docs:start
    return ClassicEditor.create( document.querySelector( '#editor' ), {
        features: [ 'paragraph', 'typing', ... ],
        toolbar: [ ... ]
[...]

PR #299, opened 2016-08-16
Introduced tasks and configuration for testing releases.

pomek:
Fixes: #260 

**Requires:** https://github.com/ckeditor/ckeditor5-dev-compiler/pull/23

This PR will introduce new tasks:

* `test:samples:local` - building tests with local modules,
* `test:samples:bundled` - building tests with bundled editor,
* `test:samples` - an alias for above tasks,
* `compile:samples` - building editors (needed for `test:samples:bundled`),
* `compile:samples:clean` - removes built editors. 

For testing it, you need to create some sample files:

**`ckeditor5-typing/docs/samples/typing.js`**
```js
/**
 * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
 * For licensing, see LICENSE.md.
 */

/* global console:false, bender: false */

import ClassicEditor from '/ckeditor5/editor-classic/classic.js'; // editor:module

const done = bender.defer();

// sample:start
ClassicEditor.create( document.querySelector( '#editor' ), { // editor:name
	features: [ 'enter', 'typing', 'paragraph', 'undo', 'basic-styles/bold', 'basic-styles/italic' ], // sample:remove|editor:features
	toolbar: [ 'bold', 'italic', 'undo', 'redo' ]
} )
.then( editor => {
	window.editor = editor;
	done(); // sample:remove
}, err => {
	console.error( err.stack );
} );
// sample:end

describe( 'Typing sample', () => {
	it( 'contains default value', () => {
		const output = '<p><em>This</em> is an <strong>editor</strong> instance...</p>';

		expect( editor.getData() ).to.equal( output );
	} );
} );
```

**`ckeditor5-typing/docs/samples/typing.html`**
```html
<div id="editor">
	<p><em>This</em> is an <strong>editor</strong> instance...</p>
</div>
```

**`ckeditor5-basic-samples/docs/samples/basic-styles.js`**
```js
/**
 * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
 * For licensing, see LICENSE.md.
 */

/* global console:false, bender: false */

import ClassicEditor from '/ckeditor5/editor-classic/classic.js'; // editor:module

const done = bender.defer();

// sample:start
ClassicEditor.create( document.querySelector( '#editor' ), { // editor:name
	features: [ 'enter', 'typing', 'paragraph', 'undo', 'basic-styles/bold', 'basic-styles/italic' ], // sample:remove|editor:features
	toolbar: [ 'bold', 'italic', 'undo', 'redo' ]
} )
.then( editor => {
	window.editor = editor;
	done(); // sample:remove
}, err => {
	console.error( err.stack );
} );
// sample:end

describe( 'Basic Styles sample', () => {
	it( 'contains default value', () => {
		const output = '<p><em>This</em> is an <strong>editor</strong> instance...</p>';

		expect( editor.getData() ).to.equal( output );
	} );
} );
```

**`ckeditor5-basic-samples/docs/samples/basic-styles.html`**
```html
<div id="editor">
	<p><em>This</em> is an <strong>editor</strong> instance...</p>
</div>
```


pomek:
**`r-` because there is still one bug. It will be fixed tomorrow (probably).**

gulpfile.js
```diff
@@ -24,6 +24,15 @@ const config = {
 	// Path to the default configuration file for bundler.
 	BUNDLE_DEFAULT_CONFIG: 'dev/bundles/build-config-standard.js',
 
+	DOCUMENTATION: {
+		// Path to the built editors for samples.
+		EDITORS_DIST: '.docs/assets/scripts/samples',
```
Reinmar: `.docs` looks like a mistake.

Reinmar: OK, I can see that you meant to use `.docs`. Then this is bad. You should use some directory inside `build/docs/`. Also, aren't those files needed only temporarily?

pomek: During this issue, those editors aren't needed. But, in the future we will introduce `docs-builder` and that tool will need to have the editors (for sample pages).

Reinmar:
There are couple of things that I noticed already:

1. The sample tests (not bundled) should be generated together with all other tests (so when using `gulp compile [--watch]`). The goal is to use sample tests as part of integration tests. This, however, may require more coding, so we don't have to work on this in this PR.
2. In the bundled scenario, the editor package is injected directly to test JS file. It should be saved on a disk as a separate file and loaded using `bender-include` directive (like [here](https://github.com/ckeditor/ckeditor-dev/blob/846f2a1a8f45db9b2c59538b59cbbfb1080b84ed/tests/plugins/clipboard/inline.js#L3)).
3. When marking lines (`// sample:remove|editor:features`), I think that comma (followed by a space of course!) would be more suitable than `|` to separate directives.

pomek:
I agree with all the points and I will fix.

Reinmar:
One thing that I want to think through is the naming of the commands. Right now there's lack of consistency and a lack of plan visible, because the tasks were created by many devs at different stages.

Reinmar:
I extracted the first point from https://github.com/ckeditor/ckeditor5/pull/299#issuecomment-240974790 to https://github.com/ckeditor/ckeditor5-dev-compiler/issues/24.

Reinmar:
I pushed one more commit which cleans up few things in the gulpfile. I also reported a few tickets:

* https://github.com/ckeditor/ckeditor5/issues/310
* https://github.com/ckeditor/ckeditor5/issues/309
* https://github.com/ckeditor/ckeditor5/issues/308

Reinmar:
There are couple of things that still need to be taken care of:

* Some garbage is left in `build/dist/` after running `gulp compile:bundled-sample-tests`.
* There's an inconsistency in config options – there's `DOCS_DIR` and `DOCUMENTATION`. Do I understand correctly that the first option will be removed soon? If so, then it can stay there for a while.
* `compiler.compile.tests` should be renamed to `compiler.compile.sampleTests`.

Also, I start to think that we need a new repository for the docs things or something, because the code which bundles editors for release tests don't match any repository that we have now... We should think about this after closing these PRs.

pomek:
@Reinmar,

1. Done.
2. Right, `docs-builder` will unify it.
3. Done.

> I start to think that we need a new repository for the docs things or something

@oskarwrobel [mentioned](https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/pull/15/files/51f8d0d6fce878a6bd016c44581802043282188e#r74061718) about it. I also thought about it.

Reinmar:
OK, I've done some refactoring here and there, moved some code from dependent PRs to this one (temporarily) and merged this code to branch `docs`. I changed deps to use GH links so running `npm i` after cleaning up `node_modules/@ckeditor` should switch all deps correctly.

Let's keep this code out of master until we have the documentation ready.

Issue #300, opened 2016-08-18
Extend chai instance to allow checking for thrown `CKEditorError` errors.


maxbarnas:
Related: https://github.com/ckeditor/ckeditor5-utils/issues/2.

maxbarnas:
I pushed very early version of such extension to https://github.com/ckeditor/ckeditor5/tree/t/300.

Reinmar:
The code looks nice :) Especially that it's validating itself (though, `describe()` and `it()` is missing, so I'm not sure what an assertion error would do).

maxbarnas:
I added tests just to have quick feedback about the behaviour of extension. When those tests breaks, they break whole test suite. And it doesn't look nice at all. I'll check if I can put in in `describe` block to make it spot easier if in future those tests break.

maxbarnas:
I set this as invalid and reverted the code because of problems with Chai extension in bundles. We'll use Babel plugin, that allows to extend builtin types - this will make `instanceof` usable again.

PR #301, opened 2016-08-19
Added expect( fn ).to.throwCKEditorError() extension to ChaiJS

maxbarnas:
Resolves: https://github.com/ckeditor/ckeditor5/issues/300.


Reinmar:
Please correct linting errors and comments. Please also add the test that I mentioned above.

Issue #319, opened 2016-08-29
Adjust Gulp tasks to the latest changes in Compiler


pomek:
Related to: https://github.com/ckeditor/ckeditor5-dev-compiler/issues/32

PR #320, opened 2016-08-29
Adjust gulp tasks to the latest changes in ckeditor5-dev-compiler.

pomek:
Fixes: #319 

Depends on: https://github.com/ckeditor/ckeditor5-dev-compiler/pull/33


Reinmar:
This can't be merged to master, because master still uses versioned dev-compiler. This can only be merged to docs.

pomek:
You're right.

Ok. I changed the base branch. You can review again :)

Issue #322, opened 2016-08-31
Docs: Remove `buildEditorsForSamples()` function from Gulpfile


pomek:
#### The issue is related with [`docs`](https://github.com/ckeditor/ckeditor5/tree/docs) branch.

Related to: https://github.com/ckeditor/ckeditor5-dev-tests/issues/2

Because this function has moved to another repository, there is no sense to keeping a duplicate of them.

PR #324, opened 2016-08-31
Removed code which was moved to dev-tests.

pomek:
Fixes: #322


Reinmar:
Let's not use this package for now. Let's only move the code there and get back to testing releases some day.

Reinmar:
I pushed two commits fixing the PR.

PR #351, opened 2016-10-20
Fixed invalid disallowMultipleVarDecl setting

wwalc:
See http://jscs.info/rule/disallowMultipleVarDecl.


Reinmar:
We're switching to ESLint and we have this:

```js
		'one-var': [
			'error',
			{
				uninitialized: 'always',
				initialized: 'never'
			}
		],
```

Issue #374, opened 2016-12-06
Aligning gulpfile to the bundler-rollup changes 


ma2ciek:
Update gulpfile to match changes introduced in https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/issues/37.

PR #379, opened 2016-12-14
Changes in gulpfile needed by Rollup bundler

ma2ciek:
Changes in gulpfile needed by https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/pull/38 

Fixes https://github.com/ckeditor/ckeditor5/issues/374.


build-config.js
```diff
@@ -0,0 +1,36 @@
+/* jshint browser: false, node: true, strict: true */
+'use strict';
+
+/**
+ * @see https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/issues/37 for config options
```
Reinmar: If we want to point someone to options documentation then it should be in the README of that package.

build-config.js
```diff
@@ -0,0 +1,36 @@
+/* jshint browser: false, node: true, strict: true */
```
Reinmar: Missing new line after comment.

gulpfile.js
```diff
@@ -126,34 +126,21 @@ gulp.task( 'compile:themes:esnext', ( callback ) => {
...
-
-gulp.task( 'bundle:clean', ckeditor5DevBundler.cleanFromConfig );
-gulp.task( 'bundle:minify:js', ckeditor5DevBundler.minify.jsFromConfig );
-gulp.task( 'bundle:minify:css', ckeditor5DevBundler.minify.cssFromConfig );
-
-// Generates the bundle without minifying it.
-gulp.task( 'bundle:generate',
-	[
-		'bundle:clean',
-		'compile:js:esnext',
```
Reinmar: Please check whether the `compile:*` tasks couldn't be removed now.

ma2ciek: `docs-buider` is still using compiler here, but this part of code can be also moved from the main gulpfile, so here we can expose only `docs`.

PR #388, opened 2017-01-10
Introduced `mgit`.

pomek:
Fixes: #???.

When everything will work as we expected, we can separate this code to another repository.


pomek:
`R-` because the code have a lot of `console.log`.

I need to find some logger instead of.

mgit/README.md
```diff
@@ -0,0 +1,87 @@
+# MGit
+
+Tool for management the multi-repository architecture.
+
+## Instalation
+
+```bash
+npm install @cksource/mgit
```
Reinmar: I'd call it `mgit2` (but the executable would be `mgit`).

mgit/README.md
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+```
+
+* `packages` - the directory where dependencies will be cloned,
+* `dependencies` - list of dependencies.
+
+The following dependencies are proper and can be placed into configuration file:
+
+Package will be installed in `packages/@scope/package` from branch `develop`:
```
Reinmar: Explain what `scope/package` means. And it should rather be `organization/repository`, because here it is GH URL.

mgit/README.md
```diff
@@ -0,0 +1,87 @@
...
+{
+	"@scope/package": "scopes/package#develop"
+}
+```
+
+Package will be installed in `packages/package` from branch `master`:
+
+```json
+{
+	"package": "https://github.com/github/package"
```
Reinmar: This is not a correct form – you shouldn't clone by `https://` if you want to work on this package (you won't be able to push). Only the GH URL (`org/repo`) or full git URL are correct forms.

mgit/index.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+'use strict';
+
+const path = require( 'path' );
+const meow = require( 'meow' );
+const mgit = require( './lib/index' );
+
+const cli = meow( `
+                     _ _
```
Reinmar: Haha :D Not very readable unfortunately :P Better to remove it or find a better generator (unless you designed it yourself :P).

mgit/lib/utils.js
```diff
@@ -0,0 +1,32 @@
...
+const utils = {
+	/**
+	 * Executes a command.
+	 *
+	 * @param {String} command Command to execute.
+	 * @returns {Promise}
+	 */
+	exec( command ) {
+		return new Promise( ( resolve, reject ) => {
+			execSh( command, true, ( err, stdout, stderr ) => {
```
Reinmar: Why async? Do we gain anything by using async execSh? I don't think so, because we've got a linear flow of each process (one command needs to finish before the next one is started). So it's better to use synchronous versions of IO and other methods, because then the code is easier to read.

I can see that `exec-sh` doesn't have a synchronous mode, so this means that perhaps `shelljs` will better fit our use case?

mgit/lib/utils.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+const execSh = require( 'exec-sh' );
+
+const utils = {
```
Reinmar: To follow the convention begun in https://github.com/ckeditor/ckeditor5-dev/tree/master/packages/ckeditor5-dev-tests/lib/utils and used in CKEditor 5 code, let's keep every util file separately.

mgit/lib/resolver.js
```diff
@@ -0,0 +1,20 @@
...
+
+'use strict';
+
+const path = require( 'path' );
+
+/**
+ * Resolves repository for given package.
+ *
+ * @param {String} name Package name.
+ * @returns {String}
```
Reinmar: May return `undefined` if a name isn't found. BTW, it'd be good to call this the `default-repository-resolver.js`. `lib/resolver.js` is too vague.

mgit/lib/createforkpool.js
```diff
@@ -0,0 +1,70 @@
...
+'use strict';
+
+const childProcess = require( 'child_process' );
+const genericPool = require( 'generic-pool' );
+
+/**
+ * @param {String} childPath Path to module that will be forked.
+ * @returns {Object} response
+ * @returns {Boolean} response.isDone
+ * @returns {Promise} response.enqueue
```
Reinmar: It's not a promise – it's a function which returns a promise.

mgit/lib/commands/exec.js
```diff
@@ -0,0 +1,48 @@
...
+	 * @param {Object} data.parameters Additional arguments provided by user.
+	 * @param {String} data.name Name of current package that will be parsed.
+	 * @param {Object} data.mgit MGit configuration.
+	 * @param {Object} data.mgit.dependencies MGit dependencies.
+	 * @returns {Promise}
+	 */
+	execute( data ) {
+		return new Promise( ( resolve, reject ) => {
+			const cwd = process.cwd();
+			const newCwd = path.join( cwd, data.mgit.packages, data.name );
```
Reinmar: In order to make this tool more useful the cwd should not be taken from the `process.cwd()` but should be the directory which holds `mgit.json`. Just like git does with `.git/`. 

It doesn't make sense to write a code now which resolves the cwd to `mgit.json`'s directory, but at least keep this in mind.

mgit/lib/commands/bootstrap.js
```diff
@@ -0,0 +1,71 @@
...
+				const parsedRepository = parseGithubUrl( repositoryName );
+
+				//jscs:disable maximumLineLength
+				const command = `git clone -b ${ parsedRepository.branch } git@github.com:${ parsedRepository.repository }.git ${ destinationPath }`;
+				//jscs:enable maximumLineLength
+
+				utils.exec( command )
+					.then( () => {
+						if ( !data.options.recursive ) {
+							resolve();
```
Reinmar: This code will call `resolve()` twice.

mgit/lib/commands/update.js
```diff
@@ -0,0 +1,100 @@
...
+					name: data.name,
+					mgit: data.mgit
+				};
+
+				return bootstrapCommand.execute( bootstrapOptions )
+					.then( resolve )
+					.catch( reject );
+			}
+
+			try {
```
Reinmar: You don't need try-catch in Promises' callback.

mgit/lib/commands/update.js
```diff
@@ -0,0 +1,100 @@
...
+					.then( () => {
+						resolve();
+					} )
+					.catch( reject );
+			} catch ( err ) {
+				reject( err.stack );
+			}
+		} );
+
+		/**
```
Reinmar: Don't use `/**` comments for private code.

mgit/lib/forkexecutable.js
```diff
@@ -0,0 +1,18 @@
+/**
```
Reinmar: This file should be called something like `child-process.js` or at least `child.js`. It doesn't `fork` anything, especially not `executable`.

mgit/index.js
ma2ciek: Nice! :D

mgit/lib/index.js
```diff
@@ -0,0 +1,110 @@
...
+			name: packageName,
+			mgit: mgitConf
+		};
+
+		forkPool.enqueue( data )
+			.then( ( response ) => {
+				progress.tick();
+
+				if ( Array.isArray( response ) ) {
+					progress.total += response.length;
```
Reinmar: Please structure the `response` a bit more. The bootstrap command should return deps of a handled package as `response.dependencies` or something. Otherwise, this is nightmare to understand :P

mgit/index.js
pomek: I assume you don't run the mgit in CLI :D 

image

mgit/README.md
pomek: It does not matter because the `bootstrap` command creates a command to clone a repository manually:

```js
const command = `git clone -b ${ parsedRepository.branch } git@github.com:${ parsedRepository.repository }.git ${ destinationPath }`;
```

The configuration has to provide correct URLs, `bootstrap` will build correct git command and everything will work.

pomek: In the future, we can consider support for other repositories (like bitbucket, etc). It should not be a problem.

Reinmar: But that's not the point. You can clone repo by this URL. But you're not gonna push to it (at least on GH). You need to use the SSH+Git protocol.

pomek: You didn't check the tool :(

For the config:

```json
{
  "packages": "packages/",
  "dependencies": {
    "@ckeditor/ckeditor5-autoformat": "https://github.com/ckeditor/ckeditor5-autoformat#mgit"
  }
}
```

image

As I said, git command does not base on the configuration parameters. The tool parses the configuration and creates custom command for git.

pomek:
A few screenshots which show how the tool works:

A configuration file contains only one library.

**Bootstrap packages (and their dependencies)**

image

MGit cannot resolve repositories for dependencies.

**Bootstrap packages (and their dependencies) with custom repository resolver**

image

Custom resolver handles packages with scope (@ckeditor/repository).

**Bootstrap packages from invalid branches**

image

**Update from correct branch**

image

**Update to another branch**

image

**Update to another branch which does not exist**

image

The configuration file contains two packages now.

** Execute `git status` on all repositories**

image

**Execute invalid task on all repositories**

image

**Execute valid and invalid tasks on all repositories**

image

---

After a short talk with @Reinmar, we will change the repository name styles. Repository name will be displayed on a yellow background with black text:

image

pomek:
@Reinmar, PR is ready to review once again.

pomek:
Hm, I forgot to add `git fetch` (and option `--no-fetch`) command for `update` process.

pomek:
Ok, done.

Reinmar:
I'm closing this PR as the code was moved to and will follow the work on https://github.com/ckeditor/ckeditor5/issues/389 in t/389.

PR #428, opened 2017-04-26
Initial integration with Umberto

### LINKED ISSUE #427: Initial integration with Umberto. Prepare example directory structure for documentation.

m-turek:
Internal: Initial integration with Umberto. Example directory structure for documentation. Closes #427.


.gitignore
```diff
@@ -8,4 +8,6 @@ node_modules/**
 build/**
 coverage/**
 packages/**
+umberto-output/**
```
Reinmar: Can we keep this in `build/docs/`?

package.json
```diff
@@ -41,7 +41,8 @@
     "gulp": "^3.9.0",
     "guppy-pre-commit": "^0.4.0",
     "lerna": "^2.0.0-rc.1",
-    "mgit2": "^0.5.1"
+    "mgit2": "^0.5.1",
+    "umberto": "0.0.6"
```
Reinmar: Please use `^` ranges.

umberto.json
```diff
@@ -0,0 +1,80 @@
+{
```
Reinmar: I'm thinking... maybe it would be cleaner if this file was kept in `docs/`? There are already many files in the root dir of every package.

Reinmar:
The commit message (the thing which starts with "Internal:...") should be in the PR body, not in the title. See other PRs for examples.

.gitignore
m-turek: Yes, I will change default output path to be `build/docs/`. Also any project can change the output path by creating a custom config file for hexo.

gulpfile.js
```diff
@@ -25,11 +25,15 @@ gulp.task( 'docs', () => {
...
 				'!' + process.cwd() + '/packages/ckeditor5-*/src/lib/**/*.js'
-			],
-			destinationPath: path.join( process.cwd(), 'build', 'docs' )
+			]
 		} );
 } );
 
+gulp.task( 'umberto', () => {
+	return require( 'umberto' )
+		.buildSingleProject( true );
```
ma2ciek: Maybe it's not a part of this PR, but we used to passing objects instead of flags as the objects are more explicit in such situations and easier to understand.

Consider this:
```js
.buildSingleProject( { clean: true } );
```

Reinmar: Yep, I wanted to comment on this too. It's indeed not part of this PR though :D

m-turek: Sure, thx

Reinmar:
OK, it seems that we're running into https://github.com/lerna/lerna/issues/789. I disabled concurrency for now. It seems to slow down `lerna bootstrap` from 140s to 250s, but now Lerna has a lot more packages to install (this PR introduces them) so there may not be a difference really. I need to check that on master to have comparable results.

However, it made me think that perhaps we shouldn't install umberto by default. This slows down the whole process significantly as this entire tool is huuge. Maybe, we could have it as an optional dependency. I'd also move there ckeditor5-dev-env which is also pretty big. I need to think on that.

PR #597, opened 2017-10-10
Internal: Remove gulp dependency for pre-commit linting.

### LINKED ISSUE #473: Consider switching from gulp to npm scripts to reduce bloat
We have `ckeditor5` repository and `ckeditor5-*` sub repositories. Right now all of them use gulp.

As for `ckeditor5` itself, it doesn't matter that much, but we already have some npm scripts there, so it's a bit messy. We could try to standardise this.

A bigger issue is with all the other packages. All of them base on gulp which isn't the most lightweight dependency and we use it to expose two simple tasks – `lint` and `lint-staged`. 

So, every `ckeditor5-*` package depends on `@ckeditor/ckeditor5-dev-lint`, `gulp`, `guppy-pre-commit` which have around 380 dependencies: https://gist.github.com/Reinmar/d9633ea5bf583c6a85111a269d273995.

We install them using Lerna and only `guppy-pre-commit` is installed in every package physically. But it still takes a time to calculate all the deps for every package to see what needs to be hoisted to `ckeditor5`.

Another thing is that we have one additional file – `gulpfile.js` that we could remove if we switched to npm scripts. And, we could drop the `ckeditor5-dev-lint` package altogether if we'd be lucky.

So, the plan would be to try to implement:

* three scripts `lint`, `lint-staged` and `pre-commit`,
* automatic git pre-commit hook installation (just like `guppy-pre-commit` but triggering npm instead of gulp).

It seems that it can be easily done using https://github.com/okonet/lint-staged. We just need to figure out whether this will be any real improvement in terms of dependency bloat.

jodator:

Internal: Remove gulp dependency for pre-commit linting. Closes #473.

---

### Additional information

* Requires https://github.com/ckeditor/ckeditor5-dev/pull/295 as it changes how Travis tests are run
* All repositories have required changes on `t/ckeditor5/473` branches

Installation step:

```
rm -rf .git/hooks/pre-commit
mgit exec 'rm -rf .git/hooks/pre-commit'
lerna clean && lerna bootstrap
```


.eslintignore
```diff
@@ -0,0 +1,4 @@
+node_modules/
```
Reinmar: Do we need to have a new file? Can't we put this into `.eslintrc.json`?

Reinmar: Also, do we have to duplicate here stuff from `.gitignore`?

docs/framework/guides/contributing/development-environment.md
```diff
@@ -96,7 +96,6 @@ total 16
 drwxr-xr-x    5 p  staff   170 31 Jan 10:37 .
 drwxr-xr-x  292 p  staff  9928 20 Jan 00:20 ..
 lrwxr-xr-x    1 p  staff    23 31 Jan 10:37 ckeditor5-core -> ../../../ckeditor5-core
-drwxr-xr-x    7 p  staff   238 20 Jan 00:20 ckeditor5-dev-lint
```
Reinmar: See also the paragraph before this block.

.eslintignore
jodator: Hm... I can only see this: https://eslint.org/docs/user-guide/configuring#using-eslintignore-in-packagejson. So in `package.json`. I'll check how `excludedFiles` works from `eslintrc.js`.

package.json
```diff
@@ -69,6 +70,13 @@
     "test": "node --max_old_space_size=4096 ./node_modules/.bin/ckeditor5-dev-tests --reporter=dots",
     "install-optional-dependencies": "./scripts/install-optional-dependencies.sh",
     "switch-to-dev-dev": "./scripts/switch-to-dev-dev.sh",
-    "build-and-publish-docs": "./scripts/docs/build-and-publish.js"
+    "build-and-publish-docs": "./scripts/docs/build-and-publish.js",
+    "lint": "eslint --quiet '**/*.js'",
```
Reinmar: Please move these to the top. They are more important than the other tasks which are our super-dev tasks, in fact.

.eslintignore
Reinmar: BTW, in https://github.com/ckeditor/ckeditor5-engine/compare/t/ckeditor5/473 you only added stuff additional to `.gitignore`.

jodator: Yep. `node_modules/` are ignored by default.

jodator: OK so we either stick with `.eslintignore` (with additional folders to `node_modules`) or we can add this to `package.json`:
```json
{
  "eslintIgnore": [ "src/lib/**" ]
}
```

Reinmar: I'm for putting this into `package.json`. There are already other paths for linting defined there (in tasks' definitions).

Reinmar:
There are a couple of things which don't work:

1. Hooks don't work in other packages.
2. When linting fails it creates an awful error:

   image

Reinmar:
I fixed the first issue with daa93c3.

Reinmar:
> When linting fails it creates an awful error:

OK, it seems that `npm run -s lint` prevents that. And that there's no other way around. At least not one I found. You can research this after we close this PR.

Reinmar:
We also blew up CI because we use mgit there too (I'm talking about ckeditor5-* repos) and the `packages/` dir is not ignored there. I'll fix it too.

PR #600, opened 2017-10-12
Internal: Use new changelog generator features

Reinmar:

Internal: Use the new changelog generator features for summary changlogs in builds and in the main package. See https://github.com/ckeditor/ckeditor5-dev/issues/289.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-dev/pull/291.


gulpfile.js
```diff
@@ -116,14 +116,29 @@ gulp.task( 'translations:download', () => {
...
-			packages: 'packages'
-		} );
+	const devEnv = require( '@ckeditor/ckeditor5-dev-env' );
+	const commonOptions = {
+		cwd: process.cwd(),
+		packages: 'packages'
+	};
+
+	const optionsForDependencies = Object.assign( {}, commonOptions, {
+		skipPackages: [
```
Reinmar: I don't like hardcoding just these packages. Can't we use some wildcard? `@ckeditor/ckeditor5-build-*`. Besides, it's inconsistent that this option uses strings while the `scope` option below uses a regexp.

gulpfile.js
```diff
@@ -116,14 +116,29 @@ gulp.task( 'translations:download', () => {
...
+			'@ckeditor/ckeditor5-build-inline',
+			'@ckeditor/ckeditor5-build-classic',
+			'@ckeditor/ckeditor5-build-balloon'
+		]
+	} );
+	const optionsForBuilds = Object.assign( {}, commonOptions, {
+		scope: /-build-/
```
Reinmar: Always make the regexp as specific as possible, but not more. This one is going to fail if we ever have a package with `-build-` used anywhere (which is rather improbable but why risking anyway).

PR #602, opened 2017-10-13
T/473: Remove gulp dependency

### LINKED ISSUE #473: Consider switching from gulp to npm scripts to reduce bloat

jodator:

Other: Remove gulp dependency in favor of npm scripts. Closes #473.

BREAKING CHANGE: Gulp task are removed. New npm scripts were introduced for dev related tasks.

---

### Additional information

* Requires https://github.com/ckeditor/ckeditor5-dev/pull/298.
* Most of gulp tasks were refactored to js scripts (located in `./scripts/release` and `./scripts/docs`) and are invoked by appropriate npm `run-script`.


scripts/docs/build-docs.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* eslint-env node */
+
+'use strict';
+
+const assertIsInstalled = require( './../util/assertisinstalled' );
```
Reinmar: Let's call this directory `utils/` to be consistent.

pomek:
I've just checked all the tasks. All works fine.

Reinmar:
@szymonkups I don't want to merge this PR and leave, but it LGTM too. I think you can merge it but please check the docs for some leftovers, cause I haven't checked that bit.

szymonkups:
I've checked both PR's and everything seems fine. But I will not close them: after merging PR in `ckeditor5-dev` we need to release packages, update their versions in `package.json` in this PR. After that we will be able to safely merge this PR.

PR #691, opened 2017-11-24
T/690: Fixed script which validates docs

### LINKED ISSUE #690: Docs valdiation does not break the build
## 🐞 Is this a bug report or feature request? (choose one)

- Bug report

## 💻 Version of CKEditor

Latest master.

## 📋 Steps to reproduce

1. Run `npm run docs:api -- --validate-only`.

## ✅ Expected result

If https://github.com/ckeditor/ckeditor5-utils/issues/209 is not resolved, `npm` should fail.

## ❎ Actual result

`npm` prints the warning but everything is ok.

It's a regression after https://github.com/ckeditor/ckeditor5/issues/473.

pomek:

Fix: Script which validates documentation will work properly. Closes #690.


scripts/docs/buildapi.js
```diff
@@ -24,6 +24,6 @@ module.exports = function buildApiDocs() {
 				'!' + process.cwd() + '/packages/ckeditor5-*/src/lib/**/*.js',
 				'!' + process.cwd() + '/packages/ckeditor5-build-*/src/**/*.js'
 			],
-			validateOnly: process.argv[ 3 ] == '--validate-only'
+			validateOnly: process.argv[ 2 ] == '--validate-only'
```
ma2ciek: `process.argv.includes( '--validate-only' )` would be safer.

pomek: Good one!

pomek:
Because some of the docs are invalid, build for this PR should fail:

image

scripts/docs/build-api-docs.js
```diff
@@ -11,4 +11,7 @@
 
 const buildApiDocs = require( './buildapi' );
 
-buildApiDocs();
+buildApiDocs()
+	.catch( () => {
+		process.exitCode = -1;
```
ma2ciek: Not sure about negative error codes. But I guess that everything despite exitCode that is equal to 0 will be treated as an error.

ma2ciek: screen shot 2017-12-11 at 11 49 06

pomek: You're right. We should use `1` error code (http://tldp.org/LDP/abs/html/exitcodes.html).

`1` means _Catchall for general errors_.

PR #694, opened 2017-11-28
Align docs to the new Translation Service

### LINKED ISSUE #624: Implement translation service (i18n) v2
Followup of https://github.com/ckeditor/ckeditor5/issues/387

In #387 we didn't manage to implement a webpack plugin which would be able to create translation files for CKEditor. E.g. you can see this in builds – there's a single `build/ckeditor.js` file which contains English translations. If you'd like to have the editor in Polish you need to go through the entire [Custom builds](https://docs.ckeditor.com/ckeditor5/latest/builds/guides/development/custom-builds.html) guides (see also [Setting UI language](https://docs.ckeditor.com/ckeditor5/latest/features/ui-language.html)). This is bad. Really bad.

What we need is that each build will contain the following files:

```
build/
    ckeditor.js
    lang/
        en.js
        pl.js
        ...
```

To enable the editor with a Polish interface you'll need to load two files (e.g. using script tags):

```js
<script src="<path-to-build>/build/ckeditor.js"></script>
<script src="<path-to-build>/build/lang/pl.js"></script>
```

And initialize it:

```js
ClassicEditor.
    create( {
        language: 'pl'
    } )
    ...
```

By default, if not specified, the editor will be initialised with English interface (cause that localization is automatically bundled in `ckeditor.js`).

## Implementation

It'd be best if all this was handled by the `CKEditorWebpackPlugin` that we use for builds:

```js
new CKEditorWebpackPlugin( {
	language: 'en',
	buildAlso: 'all' // or e.g.: [ 'pl', 'en' ]
} ),
```

If someone wants to build just a single language and have the most optimised build she/he can do it like this:

```js
new CKEditorWebpackPlugin( {
	language: 'pl'
} ),
```

From CKEditor's code perspective, the language files will look something like this:

```js
CKEDITOR_TRANSLATIONS.add( 'pl', {
    bold: 'Pogrubienie'
} )
```

And this will use the global API exposed by the https://github.com/ckeditor/ckeditor5-utils/blob/master/src/translation-service.js (basically, the `add()` function needs to be exposed).

If it won't be possible to implement this with a webpack's plugin it can also be done independently from webpack – by an external tool. Every build has its `bin/build-ckeditor.sh` file anyway so we can add more scripts there. However, this will need to be somehow synchronised with the settings passed to `CKEditorWebpackPlugin()`.

One more thing to notice is that when that plugin is configured to build just one language it could use the current method which is to replace `t( 'Bold' )` calls with e.g. `t( 'Pogrubienie' )`. When building for multiple languages such calls need to be replaced with `t( 'some-key' )` (see the old ticket for more details about this) and this `some-key` needs to be passed to translation services (so a different translation service is need to be used – cause right now `t()` function returns its first param).

However, for simplicity, we could always replace `t()` with `t( 'some-key' )`, even when building for a single language. Then, it'd be just a matter of feeding the translations.

ma2ciek:

Docs: Aligned docs and snippets to the Translation Service v2. Closes #624.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-dev/pull/314.
TODO: Update snippet-adapter's dependencies after `ckeditor5-dev` releases.
TODO: Update package.json after `ckeditor5-dev` releases.


docs/builds/guides/development/custom-builds.md
```diff
@@ -106,12 +106,12 @@ module.exports = {
...
-	// UI language. Language codes follow the https://en.wikipedia.org/wiki/ISO_639-1 format.
-	language: 'en',
-
 	// Editor configuration.
 	config: {
-		toolbar: [ 'headings', 'bold', 'italic', 'custombutton' ]
+		toolbar: [ 'headings', 'bold', 'italic', 'custombutton' ],
+
+		// UI language. Language codes follow the https://en.wikipedia.org/wiki/ISO_639-1 format.
+		lang: 'en'
```
Reinmar: One problem I see here is that the editor's config option is called `lang` while the Webpack plugin's `language`. This will lead people misconfiguring one or the other.

Since `additionalLangs` and `langs` sound at least bad, I think it'll be better if we used the full `language` word in all cases. So we need to change it in the editor's code.

docs/features/ui-language.md
```diff
@@ -15,7 +15,7 @@ See the demo of the editor in German:
 
 ## Building the editor using a specific language
 
-Currently, it is only possible to change the UI language at the build stage. A single build of the editor supports only the language which was defined in the [CKEditor 5 webpack plugin](https://www.npmjs.com/package/@ckeditor/ckeditor5-dev-webpack-plugin)'s configuration.
+Currently, it is possible to change the UI language at the build stage and after the build. A single build of the editor supports the language which was defined in the [CKEditor 5 webpack plugin](https://www.npmjs.com/package/@ckeditor/ckeditor5-dev-webpack-plugin)'s configuration. See the whole translation process to see how you can change the language later.
```
Reinmar: This guide needs to be pretty much rewritten rather than patched. Since the whole picture changed dramatically  we need to put focus on different things.

1. The current intro can stay.
1. But then we should explain how to initialize the editor from CDN/zip/npm with a different language. This is what most people will want to do. Please also review other guides like `installation.md` and other places where we list files available in a build because that list needs to be updated (and perhaps link to the ui-language.md guide):

   image

2. Then we can provide a few more details on what other options people have:

   * compiling an optimised editor (single language case),
   * compiling a subset of languages (TBH, I'm not sure what's the use case for that – perhaps speeding up the process and reducing the number of files – but let's mention it anyway),
   * handling multiple entry points scenario (as @ma2ciek noticed that has certain limitations)

3. At the end we can mention that compiling editor translations is only available for Webpack and we will be extending support for other bundlers in the future. If we want to mention those tickets about translations services we should explain why we mention them. Something like – if you'd like to tinker with one of the bundlers you can read about CKEditor 5's translation service here and here.

docs/features/ui-language.md
```diff
@@ -37,19 +37,49 @@ If you build CKEditor from scratch or integrate it directly into your applicatio
...
+				// This option can be set to array of language codes or `'all'` to build all found languages.
+				// The bundle is optimized for one language when this option is omitted.
+				additionalLanguages: 'all',
+
+				// Optional directory for emitted translations, `'lang'` by default.
+				// Relative to the webpack's output.
+				outputDirectory: 'lang',
+
+				// Optional flag indicates breaking the build process when the error is found.
+				throwErrorOnMissingTranslation: false
```
Reinmar: Is `false` a default value? In documentation, in code samples, we should present real usage and no one will change `false` to `false` :P Plus, the comment above is unclear now – if you'd turn it on here, we could simply document that it "Will cause the building process to fail if a translation for one of the strings is missing".

ma2ciek: Ok. I just wanted to list available options with their default values.

docs/features/ui-language.md
```diff
@@ -37,19 +37,49 @@ If you build CKEditor from scratch or integrate it directly into your applicatio
...
-3. Run webpack. The CKEditor plugin for webpack will replace the {@link module:utils/locale~Locale#t `t()`} function calls used in the source code with localized language strings.
+3. Run webpack. If the `additionalLanguages` option is not set, the CKEditor plugin for webpack will replace the {@link module:utils/locale~Locale#t `t()`} function call parameters used in the source code with localized language strings. Otherwise the CKEditor plugin for webpack will replace the {@link module:utils/locale~Locale#t `t()`} function call parameters with short ids and emit the translation files that should land in the `'lang'` directory (or different, if the `outputDirectory` option is specified).
+
+4. If you want to change the language after the build ends, you will need to edit the `index.html` file, add the translation file and set the UI language to the target one.
+
+	```html
+	<script src="../build/ckeditor.js"></script>
+	<script src="../build/lang/pl.js"></script>
+	<script>
+		ClassicEditor.create( document.querySelector( '#editor' ), {
```
Reinmar: ~Better~ Consistent formatting:

```js
ClassicEditor
    .create( {
    } )
    .then()
    .catch();
```

npm-debug.log
```diff
@@ -0,0 +1,45 @@
+0 info it worked if it ends with ok
```
jodator: I dunno why I was reading this commit but this caught my attention :): `npm-debug.log`

docs/features/ui-language.md
```diff
@@ -13,6 +13,57 @@ See the demo of the editor in German:
...
+		console.log( editor );
+	} )
+	.catch( error => {
+		console.error( error );
+	} );
+```
+
+### CDN
+
+To use different language than default one, you need to load the editor together with the preferred language:
```
ma2ciek: We can specify, that English will be the main language at CDN.

Issue #578, opened 2017-10-02
[Chrome] Error throws in the image description after random typing


ma2ciek:
## Steps to reproduce
1. Go to https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/image.html or https://ckeditor5.github.io/
2. Type randomly in the image description

Error:
![screen shot 2017-10-02 at 12 44 24](https://user-images.githubusercontent.com/2768047/31074172-a01a8baa-a76f-11e7-9e8e-942e6b5fcc57.png)

Reinmar:
Hm... confirmed. When I type blabla quick enough the error is thrown.

Reinmar:
I can't reproduce this issue on http://localhost:8125/ckeditor5-image/tests/manual/caption.html though. And I can do this on http://localhost/ckeditor5/build/docs/ckeditor5/0.11.0/examples/builds/classic-editor.html. Interesting. It seems that page styling affects it.

Reinmar:
Yeah, that's definitely due to styling because when I loaded the docs before Umberto generated their CSS, I could not reproduce this issue.

ma2ciek:
It's interesting that styling might easily destroy the editor.

Reinmar:
OK, for some reason, a common ancestor of these mutations is a text node. Such situation should not be handled by `_handleContainerChildrenMutations()`. I wonder what's inside the mutations array.

Reinmar:
OK, it turns out that this is Grammarly:

![image](https://user-images.githubusercontent.com/156149/31076957-f56d6fe8-a77c-11e7-9862-a7d009c38dae.png)

It seems that in https://github.com/ckeditor/ckeditor5/issues/267 we missed that Grammarly kicks in in nested editables.

Before we'll ping Grammarly's team, we need to figure out whether we can't just add `.ck-editor__editable` to nested editables.

cc @oleq

oleq:
We could add this class but it will require some refactoring of the styles across the project. There are many usages of the `.ck-editor__editable` in various stylesheets and we must understand their purpose, then rewrite the styles to consider the more generic role of the class.

Reinmar:
@oleq, could you estimate how hard would this change be? And if it makes sense in general? I see that right now, nested editable has only the `.ck-editable` class... which is odd. Unification sound like a thing we should do.

oleq:
[...]

PR #917, opened 2018-03-21
Docs: Updated snippet styles to the latest nested editable CSS class naming convention

oleq:

Docs: Updated snippet styles to the latest nested editable CSS class naming convention (see #578).

---

### Additional information

Requires: https://github.com/ckeditor/ckeditor5-widget/pull/36


Reinmar:
WDYT about this change after making it? Does it make sense to use the same class for root and nested editables? Or does it bring more issues than it needs to do?

docs/assets/snippet-styles.css
```diff
@@ -74,12 +74,12 @@ It breaks CKEditor 5 (see how <p><code>[]</code></p> looks). */
 	padding: 0;
 }
 
-#snippet-inline-editor .ck-editor__editable {
+#snippet-inline-editor .ck-editor__editable:not(.ck-editor__nested-editable) {
```
Reinmar: Can't we use `.ck-content` here? It would pick just the right element. And kinda makes sense semantically too, doesn't it? You want to align the style of the editor contents so it looks well.

Reinmar:
Before you'll answer, please also see https://github.com/ckeditor/ckeditor5/pull/917/files#r176673884.

oleq:
> WDYT about this change after making it? Does it make sense to use the same class for root and nested editables? Or does it bring more issues than it needs to do?

TBH, I haven't seen any issues. It definitely fixes all the issues caused by Grammarly, though.

PR #921, opened 2018-03-23
t/494: Increased the specificity of CSS rules. Introduced the .ck class for editor UI components

### LINKED ISSUE #494: Rise the specificity of CSS reset rules (and all other rules as well)
I added CKEditor samples to CKEditor documentation and guess what? The CKEditor's reset gets overridden by the page styling.

image

The documentation website styles are not very specific. They are actually really common:

image

The `.formatted ...` styles come from Umberto.

So, the problem that we have is that `.ck_reset` has very low specificity. It's enough to have `.mycontent ul` to break that.

I can see that we've been adding rules like `.ck_reset_all *` which would've the same specificity as `.mycontent ul` but I'm thinking that perhaps we could do something else.

What if we had a rule that we're always using at least two classes? Like, what if everything had `.ck` class, so we can always use `.ck.ck_reset` and then `.ck.ck-list__item`. That would always give us minimum specificity of 20 which would always override styles like `.mycontent ul` (which are 11).

oleq:

Docs: Increased the specificity of CSS rules. Introduced the .ck class for editor UI components. Closes #494.

---

### Additional information

All branches in the constellation:

* ~~https://github.com/ckeditor/ckeditor5-basic-styles/compare/t/ckeditor5/494~~
* ~~https://github.com/ckeditor/ckeditor5-block-quote/compare/t/ckeditor5/494~~
* https://github.com/ckeditor/ckeditor5-editor-classic/compare/t/ckeditor5/494
* https://github.com/ckeditor/ckeditor5-engine/compare/t/ckeditor5/494
* https://github.com/ckeditor/ckeditor5-heading/compare/t/ckeditor5/494
* https://github.com/ckeditor/ckeditor5-image/compare/t/ckeditor5/494
* https://github.com/ckeditor/ckeditor5-link/compare/t/ckeditor5/494
* https://github.com/ckeditor/ckeditor5-theme-lark/compare/t/ckeditor5/494
* https://github.com/ckeditor/ckeditor5-ui/compare/t/ckeditor5/494
* https://github.com/ckeditor/ckeditor5-widget/compare/t/ckeditor5/494

----

Message for merging child branches

```
Other: Increased the specificity of CSS rules. Introduced the `.ck` class for editor UI components (see: ckeditor/ckeditor5#494).
```


Reinmar:
Why did you increase the specificity of content styles too? This issue was about UI's styles. Content styles are a different topic – they should be fairly simple to be overridden by the developer.

oleq:
[These](https://github.com/ckeditor/ckeditor5/pull/921/files#diff-e0fe1b0b36c50a30ab6b776353602026) are the documentation styles, not user content styles.

Reinmar:
I didn't mean this. I meant all the changes in basic-stylers, block-quote, image, headings, etc.

oleq:
Maybe you're right. WDYT @dkonopka?

[example](https://github.com/ckeditor/ckeditor5-block-quote/compare/t/ckeditor5/494)

dkonopka:
I can agree, `blockquote` border-color or `code` background should be easy to change, without facing CSS specificity, because as we had agreed, we are skipping CSS Variables in content styles ( https://github.com/ckeditor/ckeditor5-block-quote/issues/21)

So I'm ok with adding extra `.ck` for UI Elements like buttons, icons, tooltips to be sure that it will be displayed properly in most cases and we have a lot of vars to change color/size/background of it.

oleq:
I removed changes from basic-styles, block-quote and image.

Reinmar:
I reviewed all the changes and I'm unsure about these:

* https://github.com/ckeditor/ckeditor5-heading/compare/t/ckeditor5/494 – I assume this styles the headings in the dropdown. I'm not entirely sure whether I think about this more as about the content or about the UI.
* https://github.com/ckeditor/ckeditor5-image/compare/t/ckeditor5/494#diff-1d370f46d756a5c592c29c39a45704fbL11 – I think we miss the `figure` part here. For some reason, the `figure.image` always come together for me. But maybe it's unnecessary? @dkonopka @Comandeer

docs/assets/snippet-styles.css
```diff
@@ -33,27 +33,27 @@ It breaks CKEditor 5 (see how <p><code>[]</code></p> looks). */
...
 }
 
-.ck-reset.ck-list {
+.ck.ck-list {
 	/* See https://github.com/ckeditor/ckeditor5/issues/494 */
 	margin-left: 0;
 }
 
 /* examples/builds/inline-editor.html */
-.live-snippet .image-style-left, .live-snippet .image-style-right {
```
Reinmar: These changes don't make much sense to me. Either `.ck` or `.live-snippet`. The latter was used so far to bump up the specificity. Now, one of these become unnecessary.

Reinmar: Also, `.ck` is part of the UI, and `.image-style-left` of the content. So, if anything, the selector which makes the most sense here is `.ck-content .image-style-left`. But then, the content will not look good before the editor is initialised, so it'd be good to use a class which is always present. Which is... `.live-snippet` ;)

oleq:
> https://github.com/ckeditor/ckeditor5-heading/compare/t/ckeditor5/494 – I assume this styles the headings in the dropdown. I'm not entirely sure whether I think about this more as about the content or about the UI.

I consider it the UI.

> https://github.com/ckeditor/ckeditor5-image/compare/t/ckeditor5/494#diff-1d370f46d756a5c592c29c39a45704fbL11 – I think we miss the figure part here. For some reason, the figure.image always come together for me. But maybe it's unnecessary? @dkonopka @Comandeer

We used to have `figure.image` (1,1) to raise the specificity. I propose `.ck-content image`, which has the same specificity, but clearly defines the purpose of the style.

The [icon](https://github.com/ckeditor/ckeditor5-ui/compare/t/ckeditor5/494#diff-19674b75fd3ef7d3d181c112f1aa5a58R6), OTOH, is a different story. We had `svg.ck-icon` (1,1) to raise specificity but since this is the UI we can use `.ck.ck-icon` (2,0), which is clearer (I did it wrong in the PR — `svg` is unnecessary).

docs/assets/snippet-styles.css
oleq: Let's stick to `.l-s` :P

PR #1039, opened 2018-05-23
Introduced a script that automates the ckeditor5's releases

### LINKED ISSUE #1033: Add script for publishing ckeditor5
There are a couple of steps which I need to perform manually now:

1. Run `npm version major|minor` (choose the bump based on the changes in deps).
1. `npm publish` **IMPORTANT** See notes below.
1. Create a release on `ckeditor5`'s GitHub (copy there the changelog section).
1. Update the mention of the last released version in the README.
1. Update all `stable` branches: `git co stable && git merge master && git co master` and `mgit exec 'git co stable && git merge master && git co master'`.
1. Now, make sure that `mgit.json` has the right content in `master` and `stable`:
   1. `git co stable` and add `#stable` to all packages in `mgit.json`; commit this with `git commit -a -m "Internal: Use stable branches. [skip ci]"`. 
   1. `git co master && git merge stable` and remove `#stable` from `mgit.json`; commit this with `git commit -a -m "Internal: Use master branches. [skip ci]"`
   1. `git push origin stable master && mgit exec 'git push origin stable'`
   1. Add stable branches in all repos which don't have them yet: `mgit exec 'git co -b stable && git push origin stable && git co master'`

## Contents of this package?

This package is meant to be empty on npm. This means:

* No dependencies (neither dev nor prod ones).
* `package.json` should contain a description, keywords, license, links, etc.
* No files except `README.md` and `LICENSE.md`.

And that's it. This means that publishing it straight from this repo (which I mistakenly did in v10.0.1) publishes far too many things.

## What I need?

I need one script which will automate this process.

pomek:

Feature: Introduced a script that automates the ckeditor5's releases. Closes #1033.


scripts/release-ckeditor5.sh
```diff
@@ -0,0 +1,42 @@
...
+    # Update the `stable` branch in the ckeditor5 repository.
+	git checkout stable && git merge master && git checkout master
+
+	# Update all `stable` branches in all packages.
+	mgit exec 'git checkout stable && git merge master && git checkout master'
+
+	# Make sure that `mgit.json` for `stable` and `master` branches is correct.
+	# `stable` branch.
+	git checkout stable && \
+	node -e "require( './scripts/release/update-mgit-branches' )( 'stable' );" && \
```
Reinmar: Why not passing `stable/master` through `args` like to any shell command?

scripts/release-ckeditor5.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+# @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
+# For licensing, see LICENSE.md.
+
+set -e
+
+read -p "Did you update the README.md? " -n 1 -r
```
Reinmar: With what? Only if you'll reply "no" you can now learn what you need to update :D

scripts/release/release-ckeditor5.js
```diff
@@ -0,0 +1,121 @@
...
+		log.info( 'Bumping the version...' );
+
+		// Bump the version.
+		tools.shExec( `npm version ${ changelogVersion } --message "Release: v${ changelogVersion }."`, { verbosity: 'error' } );
+
+		log.info( 'Updating the package.json...' );
+
+		// Update the `package.json`. Published package shouldn't have any dependencies and devDependencies.
+		// Also, we should omit our environment scripts and their configs.
+		tools.updateJSONFile( packageJsonPath, packageJson => {
```
Reinmar: I'd rather preferred if we had all the files that are used during a release in a separate directory. I don't like changing existing files because e.g. if we'll add some new property to the main `package.json` we will need to remember to update this script.

So, it's better to have templates of these files in some other directory (`scripts/release/template`) and modify them there (e.g. copying some props from the main `package.json`.

Reinmar:
The script seems to does all the things which I've been doing manually, which is great. It actually does even more – it should save me from some mistakes :)

One thing that I didn't like though is how it publishes ckeditor5 to npm. It's unsafe to do that in place. It's very hard to validate in fact. Duplication is easier to handle.

scripts/release-ckeditor5.sh
pomek: Mgit does not allow changing branches in `mgit.json` file via CLI. We can save hashes or check out the branches in the repo but there is no option to save particular branch in the file.

Reinmar: I meant calling this:

```
./scripts/release/update-mgit-branches stable
```

Wouldn't it be easier?

scripts/release/release-ckeditor5.js
pomek: I created a template with all required properties. They are set by default as `null`. It means that we will copy values from original (source) `package.json`. If a value is other than `null`, the script won't touch it.

I leave the `"files"` section in the release script because we cannot put a comment in JSON file and I feel that there should be some explanation.

pomek:
> One thing that I didn't like though is how it publishes ckeditor5 to npm. It's unsafe to do that in place. It's very hard to validate in fact. Duplication is easier to handle.

Done. Now it will be executed from the template directory which contains a template for `package.json`. Required files will be copied here during the release and will be removed after.

scripts/release/release-ckeditor5.js
```diff
@@ -0,0 +1,130 @@
...
+
+		// Copy additional files.
+		for ( const file of additionalFiles ) {
+			tools.shExec( `cp ${ path.resolve( cke5Path, file ) } ${ path.resolve( templatePath, file ) }` );
+		}
+
+		log.info( 'Publishing on npm...' );
+
+		// Publish the package on npm.
+		tools.shExec( `cd ${ templatePath } && npm publish && cd ${ cke5Path }` );
```
pomek: If you would like to test it – replace `npm publish` with `npm pack`. Remember about adding `exit 1;` to the shell script (`scripts/release-ckeditor5.sh`) in L15 and committing these changes because NPM throws an error if the repository is dirty.

pomek: After finishing the test phase, remember about removing the tag created by `npm version`. :D

PR #1063, opened 2018-06-18
Improved documentation for building the project

### LINKED ISSUE #1046: [Docs] Fix regexp in webpack configuration
The regexps here: https://docs.ckeditor.com/ckeditor5/latest/builds/guides/integration/advanced-setup.html#webpack-configuration do not handle Windows paths (unless webpack normalizes them – this must be verified first!)

The other thing is that I'd replace those CKEditor 5-only regexps from comments to the normal code because in majority of situations people will be integrating this webpack config with their projects, so let's make sure that only CKEditor 5 files are treated in this way.

Finally, we need to add a note at the end of this section that if you're adding CKEditor 5 to an existing project where there are other loaders for SVG or CSS files already you need to take care of excluding CKEditor 5 files from those loaders' configuration. Just like @szymonkups described in https://github.com/ckeditor/ckeditor5-react#changes-that-need-to-be-made-to-both-config-files-webpackconfigdevjs-and-webpackconfigprodjs.

However, what I'd also like to have checked here is whether having loaders needed for CKEditor 5 at the top of the loaders definitions will not be enough. I mean  – can a single file be handled by multiple loaders? Or is only the first loader applied?

pomek:

Docs: Improved documentation for building the project. Closes #1046.

### DIFF
--- a/docs/builds/guides/integration/advanced-setup.md
+++ b/docs/builds/guides/integration/advanced-setup.md
@@ -147,7 +147,7 @@ module.exports = {
 				use: [ 'raw-loader' ]
 			},
 			{
-				test: /ckeditor5-[^/\\]+[/\\]theme[/\\][\w-/]+\.css$/,
+				test: /ckeditor5-[^/\\]+[/\\]theme[/\\][\w-/\\]+\.css$/,
 				use: [
 					{
 						loader: 'style-loader',
@@ -194,7 +194,7 @@ module.exports = {
             
             // Loader for CKEditor5's theme.
             {
-            	test: /ckeditor5-[^/\\]+[/\\]theme[/\\][\w-/]+\.css$/,
+            	test: /ckeditor5-[^/\\]+[/\\]theme[/\\][\w-/\\]+\.css$/,
             	use: [
             		{
             			loader: 'style-loader',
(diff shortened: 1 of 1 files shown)


docs/builds/guides/integration/advanced-setup.md
```diff
@@ -143,16 +143,11 @@ module.exports = {
 	module: {
 		rules: [
 			{
-				// Or /ckeditor5-[^/]+\/theme\/icons\/[^/]+\.svg$/ if you want to limit this loader
-				// to CKEditor 5 icons only.
-				test: /\.svg$/,
-
+				test: /ckeditor5-[^/\\]+[/\\]theme[/\\]icons[/\\][^/\\]+\.svg$/,
```
Reinmar: It deserves a comment. Something like:

> Process CKEditor 5 icons (ckeditor5-*/theme/icons/*.svg) using raw-loader.

This comment makes sense before `{`, just like you did in the below example in the new section.

I'm also thinking about simplifying this regexp to:

```
/ckeditor5-[^/\\]+[/\\]theme[/\\]icons[/\\].+\.svg$/
```

It will cover a bit more cases (e.g. `theme/icons/some-group-of-tem/foo.svg`).

docs/builds/guides/integration/advanced-setup.md
```diff
@@ -176,6 +171,83 @@ module.exports = {
 };
 ```
 
+#### Adding CKEditor 5 to existing project
+
+If you are not build the Webpack's configuration from scratch, you need to adjust your configuration to handle the CKEditor 5's assets.
```
Reinmar: s/Webpack/webpack

And I need to think on what should be here in general.

Reinmar:
I want to prepare a new PR for this.

PR #1118, opened 2018-06-28
Update webpack to version 4.12

ma2ciek:

Other: Updated webpack to version 4.12.


ma2ciek:
I'll update one test in `BalloonPanelView`, as I got the following error message:

```
TypeError: ES Modules cannot be spied
          at Function.spy (base-dev/node_modules/sinon/pkg/sinon.js?51ef4aa3de92c43bc4721758a1e445182048177d:2342:15)
          at Object.spy (base-dev/node_modules/sinon/pkg/sinon.js?51ef4aa3de92c43bc4721758a1e445182048177d:863:34)
          at Context.it (webpack:///packages/ckeditor5-ui/tests/panel/balloon/balloonpanelview.js:176:0 <- build/.automated-tests/entry-point.js:4877:105)
```

ma2ciek:
Can I change the `BalloonPanelView` class, so it will contain a static method `getOptimalPosition` which will be an alias to the original `getOptimalPosition` method from `@ckeditor/ckeditor5-utils/src/dom/position`? Then I'll be able to mock that method in tests.

https://github.com/ckeditor/ckeditor5-ui/blob/0ed2b79379e1ea78559cf16ddc0863f6546ebb69/src/panel/balloon/balloonpanelview.js#L11

cc @Reinmar

Reinmar:
cc @oleq

oleq:
I guess so if there's no other way ;-) Why did that thing get broken in the first place, anyway?

ma2ciek:
We import position utils, which is a module

```js
import * as positionUtils from '@ckeditor/ckeditor5-utils/src/dom/position';
```

And then we stub the module's export with:

```js
const spy = testUtils.sinon.spy( positionUtils, 'getOptimalPosition' );
```

Which becames invalid now, as `positionUtils` is an ES Module, not a normal object. (I guess it).

PR #1123, opened 2018-07-03
Script for updating version of dependencies

### LINKED ISSUE #1122: Script for saving package versions in package.json
I was doing this manually doing a release and I already made two mistakes. E.g. this: https://github.com/ckeditor/ckeditor5/commit/7c87a7e01a73c4b49f7b589233d7521b140d9a81

pomek:

Internal: Added a script that updates version of CKEditor 5 dependencies during the release. Closes #1122.


scripts/release/update-dependencies-versions.js
```diff
@@ -0,0 +1,42 @@
...
+'use strict';
+
+const path = require( 'path' );
+const { tools, logger } = require( '@ckeditor/ckeditor5-dev-utils' );
+
+const log = logger();
+const rootPath = path.join( __dirname, '..', '..' );
+const packageJsonPath = path.join( rootPath, 'package.json' );
+
+const mgitJson = require( path.join( rootPath, 'mgit.json' ) );
```
Reinmar: I don't like basing this script on `mgit.json`. Mgit provides us infrastructure. It makes sure that subrepos are available. What if one of packages will be defined somewhere else? It's theoretical, but it shows that this is not the right starting point. Let's go through all dependencies of `ckeditor5/package.json` and simply get their versions by `require( packageName + '/package.json' ).version`.

scripts/release-ckeditor5.sh
```diff
@@ -10,6 +10,9 @@ echo ""
 
 if [[ $REPLY =~ ^[Yy]$ ]]
 then
+	# Update version of CKEditor 5 dependencies in package.json
+	node ./scripts/release/update-dependencies-versions.js
```
Reinmar: It's not part of this script. We need to update versions in `package.json`, commit it, push and wait for CI. It's a separate step.

scripts/release/update-dependencies-versions.js
```diff
@@ -0,0 +1,42 @@
...
+		const dependencyPath = path.join( rootPath, mgitJson.packages, packageName.split( '/' )[ 1 ] );
+		const dependencyPackageJson = require( path.join( dependencyPath, 'package.json' ) );
+
+		packageJson.dependencies[ dependencyPackageJson.name ] = '^' + dependencyPackageJson.version;
+	}
+
+	return packageJson;
+} );
+
+log.info( 'Done.' );
```
Reinmar: Please print something more. E.g. "Updated versions of 45 packages."

scripts/release-ckeditor5.sh
Reinmar: It can. As:

```
git commit -a -m "Internal: Updated dependencies."
```

pomek: Done.

scripts/update-dependencies-versions.js
```diff
@@ -0,0 +1,52 @@
...
+
+if ( counter ) {
+	log.info( 'Committing updated "package.json" file...' );
+
+	const response = tools.shExec( 'git add package.json && git commit -m "Internal: Updated dependencies."', { verbosity: 'error' } );
+
+	log.info( response.trim() );
+}
+
+log.info( 'Updating dependencies has been finished.' );
```
Reinmar: Sounds like a robot talking. The previous "Done" was better. We missed some status, but nothing else.

Issue #667, opened 2017-11-15
Allow adding plugins to a CKEditor 5 build


pjasiun:
Right now there is no easy way to add a real plugin to the CKEditor 5 build. Even if it is theoretically possible, you will not be able to create a position or fire an event.

The problem is that build has all classes build in and expose only editors API. If a custom plugin imports `Position` from the `ckeditor5-engine` and use it in the plugin it will not be the same instance as the one used in the engine in the build. Because of it, this position will return `false` when `position instanceof Position` will be called, so the user will get an ugly bug.

There is a simmilar problem with symbols, which are used in the events system.

Note that it is even worse because the same issue will appear when a custom plugin will have a different version of the engine on the list of dependencies. We could say that in such case semantic versioning will save us, but it still means that with every major engine release all 3rd party plugins will stop working immediately, even if changes in the engine breaks nothing important for them. The version of the dependency will not match anymore and 2 separate engines will be included with not matching classes.

This very ticky problem, but there are some improvements we could think about.

Since editor usually needs all engine features, we could expose them as `document` property, have something like `document.createPosition`. Then all plugins will use instances of the same class and the problem disappears. In fact, when all engine utiles will be exposed on the `document` model (or controller), plugins will not need to depends on the engine and only use engine exposed by the editor.

Unfortunately, it looks fine only in the theory. Only for position we have:
```js
Position.constructor( root, path )
Position.createAt( itemOrPosition, offset )
Position.createAfter( item )
Position.createBefore( item )
Position.createFromParentAndOffset( parent, offset )
Position.createFromPosition( position )
```
Then there is range:
```js
Range.constructor( start, end )
Range.createFromPositionAndShift( position, shift )
Range.createFromParentsAndOffsets( startElement, startOffset, endElement, endOffset )
Range.createFromRange( range )
Range.createIn( element )
Range.createOn( item )
Range.createCollapsedAt( itemOrPosition, offset )
Range.createFromRanges( ranges )
```

Then, `treeWalker`... But hey! It's only the model. We have also all conversion utils to be exposed because one will need them for sure. And working with converters one will need to be able to use view API, so multiple constructors in the view need to be exposed too. It means a lot of factories and a lot of ugly APIs. And I'm talking here only about the engine, but note that there is the same problem with utils and UI. The additional problem with utils is that we can not expose all of them because we don't know which are really needed.

[...]

PR #1140, opened 2018-07-06
T/1005: Implemented a module exposing the `CKEDIOR_VERSION` to the global scope

### LINKED ISSUE #1005: Include CKEditor version number somewhere
Currently when I have a js bundle it's hard to determine what version I'm running actually.

oleq:

Feature: Implemented a module exposing the `CKEDIOR_VERSION` to the global scope. Closes #1005.

---

Requires https://github.com/ckeditor/ckeditor5-core/pull/139.

**Note**: To test it properly, the module must export a function executed somewhere in `core/editor`. Besides, I'm not sure tests will run for `ckeditor5` at all.


src/version.js
```diff
@@ -0,0 +1,39 @@
...
+	 * `<script>` elements or make sure your page builder/bundler includes CKEditor only once.
+	 *
+	 * @error ckeditor-version-collision
+	 * @param {String} collidingVersion The version of the build which has already been (incorrectly) loaded.
+	 * @param {String} version The version of the build which is supposed to be loaded.
+	 */
+	throw new CKEditorError( 'ckeditor-version-collision: The global CKEDITOR_VERSION constant has already been set.', {
+		collidingVersion: windowOrGlobal.CKEDITOR_VERSION,
+		version
+	} );
```
pjasiun: I would throw only if the version is different. If the version is the same, but it is loaded multiple times there should be a (big) warning in the console. Otherwise, we make the editor stop working in some cases when people try to prototype something. And, even though, such prototyping will not work in some cases (because of https://github.com/ckeditor/ckeditor5/issues/667) we will make it stop working in all cases. In other words, loading this module twice does not necessarily mean that something will be broken I think that warning will be much better.

oleq: cc @Reinmar

Reinmar: Right now the editor doesn't work anyway when things are duplicated (we're talking here about serious duplication of the entire `Editor` module) so an error makes a lot of sense. We may change it once we'll fix the situation.

Reinmar: Alternatively, we can change it into `console.error()`.

oleq: `log.error()` then?

src/version.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @module ckeditor5/version
+ */
+
+/* global window, global */
+
+import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+import { version } from '../package.json';
+
+const windowOrGlobal = window || global;
```
Reinmar: How's this supposed to work?

image

I think that you need to use `typeof window`.

src/version.js
Reinmar: `log.error()` is not available here, unfortunately. That'd mean that `ckeditor5` depends `ckeditor5-utils`. 

We could depend on `ckeditor5-utils` if all this code was in `ckeditor5-core`. Maybe we should go this way?

PS. Using `log.error()` is better because we'll also log a link to docs.ckeditor.com.

oleq: Maybe let's move the `version.js` to utils?

Reinmar: And what would be left here, in `ckeditor5`? It'd be empty, yes? So this change would be limited to `ckeditor5-core` and `ckeditor5-utils`?

oleq: Yes.

Reinmar: Let's do this, then.

oleq:
Moved to https://github.com/ckeditor/ckeditor5-utils/pull/249.

Issue #1061, opened 2018-06-15
Make ckeditor5 a peer dependency of all other packages


Reinmar:
The next iteration (iteration19) will contain some breaking changes (e.g. https://github.com/ckeditor/ckeditor5/issues/1038, https://github.com/ckeditor/ckeditor5-core/issues/64), so it's a good moment to squeeze some more breaking/infrastructural things too.

There's one idea about making `ckeditor5` a peer dependency of other source packages (and a real dependency of builds). I explained it in https://github.com/ckeditor/ckeditor5/issues/667#issuecomment-382057035.

I'm not sure whether we'll be able to also work on reducing dependency on other pkgs, but those changes will be a progressive enhancement, so we can do them later on. Adding `ckeditor5` everywhere, though, is something which will affect other people, so we should do it now.

Reinmar:
I'm leaving this as a "discussion" because we haven't made a final decision yet. I want to track it, though.

Reinmar:
@pomek, could you test whether out release tools will properly bump version of `ckeditor5` if it's a peer dependency of other packages? I guess not, because I don't think we're touching the `peerDependencies` key at all.

I've been also thinking a bit about releasing all this and it seems that... we'll need to release all packages at the same time – `ckeditor5-*` and `ckeditor5`. Otherwise, the version of `ckeditor5` bumped in `ckeditor5-*` pkgs will make it impossible to install these packages. So, you could also check in our complete release process what blocks us from making the process continuous from the `npm run release:dependencies` to `npm run release`.

pomek:
> @pomek, could you test whether out release tools will properly bump version of ckeditor5 if it's a peer dependency of other packages? I guess not, because I don't think we're touching the peerDependencies key at all.


You're right. This tool does not touches `peerDependencies` key.

pomek:
The release tool now supports updating the `peerDependencies`.

I also introduced `dry run` mode which allows testing everything without pushing/publishing anything. We talked about it last time. It's very helpful.

I was able to generate changelogs for all packages (I used the `12.0.0` version). Then I run the release script and everything worked fine. The version of `ckeditor5` as a peer dependency was updated. 

I also didn't have any troubles with executing the whole release process.

I'm preparing pull requests now.

Reinmar:
## Some findings...

[...]

PR #1143, opened 2018-07-09
Improved the release scripts

pomek:

Internal: Added support for generating the changelog for all packages with the same version. The script for releasing packages support `dry-run` mode now. See ckeditor/ckeditor5#1061.

---

Requires: https://github.com/ckeditor/ckeditor5-dev/pull/420

### DIFF
--- a/scripts/release-ckeditor5.sh
+++ b/scripts/release-ckeditor5.sh
@@ -20,7 +20,7 @@ then
 	mgit exec 'git checkout -b stable 2> /dev/null && git push origin stable && git checkout master'
 
 	# Update all `stable` branches in all packages.
-	mgit exec 'git checkout stable && git merge master && git checkout master'
+	mgit exec 'git checkout stable && git pull origin stable && git merge master && git checkout master'
 
 	# Make sure that `mgit.json` for `stable` and `master` branches is correct.
 	# `stable` branch.
--- a/scripts/release/changelog.js
+++ b/scripts/release/changelog.js
@@ -9,10 +9,14 @@
 
 'use strict';
 
+// In order to use the same version for all packages (including builds and ckeditor5 itself), you can call:
+// npm run changelog [newVersion]
+
 const devEnv = require( '@ckeditor/ckeditor5-dev-env' );
 const commonOptions = {
 	cwd: process.cwd(),
-	packages: 'packages'
+	packages: 'packages',
+	newVersion: process.argv[ 2 ] || null
 };
 const editorBuildsGlob = '@ckeditor/ckeditor5-build-*';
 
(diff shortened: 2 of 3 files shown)


scripts/release-ckeditor5.sh
```diff
@@ -20,7 +20,7 @@ then
 	mgit exec 'git checkout -b stable 2> /dev/null && git push origin stable && git checkout master'
 
 	# Update all `stable` branches in all packages.
-	mgit exec 'git checkout stable && git merge master && git checkout master'
+	mgit exec 'git checkout stable && git pull origin stable && git merge master && git checkout master'
```
pomek: We need to pull the `stable` branch because it could be behind to the remote.

scripts/release/changelog.js
```diff
@@ -9,10 +9,14 @@
...
 
+// In order to use the same version for all packages (including builds and ckeditor5 itself), you can call:
+// npm run changelog [newVersion]
+
 const devEnv = require( '@ckeditor/ckeditor5-dev-env' );
 const commonOptions = {
 	cwd: process.cwd(),
-	packages: 'packages'
+	packages: 'packages',
+	newVersion: process.argv[ 2 ] || null
```
Reinmar: New version of what? Does this option make sense in CKEditor 5?

pomek: For the real release - nope. But for testing (like I often do) - yes.

We can skip that and I'll modify this file when I'll need that.

pomek: newVersion for changelog of course. It's a version that you need to specify based on history of commits.

Reinmar: > newVersion for changelog of course. It's a version that you need to specify based on history of commits.

For which changelog? This task builds 30x changelogs. Does it affect all pkgs?

pomek: Yes.

Reinmar: Anyway, I just want to have a clarity here. You're adding an option so you should be clear why. If it's an option for tests, then please add a comment clarifying this in the code.

pomek: That was my goal. It speeds up testing this tool. I don't lose time for parsing commits from repositories and pasting the same version as an answer for the question "type new version".

Reinmar: It could even be quite useful for me in the future, once I'd have to release for e.g. the next major version of every package because of something. Then, we could make this option to also accept values such as `major` or `next-alpha`. But that's for the future, of course.

pomek: Reported as a new issue: https://github.com/ckeditor/ckeditor5-dev/issues/421

Issue #416, opened 2017-03-21
Responsive toolbar


SammyM:
## Is this a bug report or feature request?

🆕 Feature request

## 💻 Version of CKEditor

CKEditor v5 @ 12.4.0

## 📋 Steps to reproduce

Currently the toolbar gets a little ugly in narrow viewports. Namely it flows into two, disorganized lines.

Example below:

<img width="320" alt="screen shot 2017-03-21 at 15 44 26" src="https://cloud.githubusercontent.com/assets/1324289/24156122/9cfbaaaa-0e4d-11e7-864c-aab844154170.png">

A solution for this would be to collapse overflowing buttons.

---

If you'd like to see this feature implemented, add 👍 to this post.

fredck:
Hey @SammyM, can you elaborate a bit, guiding us on your expectations on this? How do you think CKEditor should behave when it comes to responsive design?

fredck:
I mean, we know there’s work to be done, and we have plans for it, but I would love to hear your expectations, because it may point us to things we didn't take in consideration so far.

SammyM:
I've not spent a lot of time playing with the new version of CKEditor so I'm not sure of all the functionality yet, but I've noticed that the popups/tooltips can't fully be seen on smaller viewports. For example:

<img width="320" alt="screen shot 2017-03-21 at 15 44 26" src="https://cloud.githubusercontent.com/assets/1324289/24156122/9cfbaaaa-0e4d-11e7-864c-aab844154170.png">

I'm glad to hear it's in your plans.

HassanAmmori:
Well, I am sure these are minor things that I wouldn't be worried about. Most of the elements act really well already

SammyM:
Is there an update on this at all? I know the UI of CKEditor has changed a bit since I last raised this issue, but the popups still go off the screen.

![screen shot 2018-09-07 at 16 44 02](https://user-images.githubusercontent.com/1324289/45229289-5d60fb00-b2bd-11e8-8b3e-c96375fcd614.png)

oleq:
I proposed 2 improvements to the UI:

## Responsive in–balloon forms

This [feature](https://github.com/ckeditor/ckeditor5/pull/1299) is on review now.

<img width="331" alt="image" src="https://user-images.githubusercontent.com/1099479/46874026-761e7c80-ce38-11e8-9a7b-e3fc61a556ed.png">

[...]

PR #1299, opened 2018-10-12
t/416: Improved responsiveness of the forms in narrow viewports

oleq:

Feature: Improved responsiveness of the forms in narrow viewports (see ckeditor/ckeditor5#416).

---

### Additional information

* This PR **should not** close the issue. This is only the first action we take to improve the responsiveness. More is yet to come.
* This is a CI/tests PR.
* PRs with actual changes:
   * https://github.com/ckeditor/ckeditor5-link/pull/205
   * https://github.com/ckeditor/ckeditor5-image/pull/239
   * https://github.com/ckeditor/ckeditor5-media-embed/pull/60
   * https://github.com/ckeditor/ckeditor5-theme-lark/pull/204

### Things to decide

* Is the media query breakpoint correct https://github.com/ckeditor/ckeditor5-theme-lark/compare/t/ckeditor5/416?expand=1#diff-5ac1a3c4aac63c20ab773daf9fbe1983R7? I picked it at random from the net (it's the screen size of some iPhone :P) but maybe we can do better than that. 
   * Also... should we use `max-device-width` instead?
* If you browse the branches you'll notice that a lot of code repeats because the UI is almost the same in each case: [ input/url preview button, save button, cancel button ]. Maybe we could give such a structure a class name, implement it somewhere, and make the existing code and this new responsive approach shorter. This could be a follow–up.

image

image

image

image


oleq:
P.S. I consider this situation and an edge case and a sacrifice for the sake of an easy solution

image

dkonopka:
First of all with this PR I would recommend to add viewport settings in manual tests:
`<meta name="viewport" content="width=device-width, initial-scale=1">`.

https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-tests/lib/utils/manual-tests/template.html#L4

dkonopka:
> If you browse the branches you'll notice that a lot of code repeats because the UI is almost the same in each case: [ input/url preview button, save button, cancel button ]

I agree this is not perfect solution, but ATM we can live with it I guess. 

Anyway, I was thinking about something like mixin for buttons in the forms (`_rwd-form-buttons`) to not duplicate code, but as you said it can be a follow-up.

dkonopka:
And my last thought, if we are using `--ck-spacing-standard` for labels, then it would be great to use it for buttons too (instead of `--ck-spacing-tiny`).

### ATM (left), proposal (right)
spacings

oleq:
That's a good idea @dkonopka! It's way easier to tap. I added it in https://github.com/ckeditor/ckeditor5-theme-lark/commit/33c974c6bc1320176703164832ef2c10b4986337.

PR #1322, opened 2018-10-24
t/ckeditor5-vue/1: Created a guide for the Vue.js component

oleq:

Docs: Created a guide for the Vue.js component (see ckeditor/ckeditor5-vue#1).


pomek:
I based on this documentation when I was reviewing https://github.com/ckeditor/ckeditor5-vue/pull/2 and everything is described clearly and understandable for me.

docs/builds/guides/frameworks/vuejs.md
```diff
@@ -0,0 +1,645 @@
...
+				editorConfig: {
+					// Configuration of the editor.
+				}
+			};
+		}
+	}
+</script>
+```
+
+## Plugin configuration
```
Reinmar: What's "plugin" here? Do you mean "`<ckeditor>` component"? If so, it'd be good to avoid using "plugin" because I thought that this is about CKEditor plugins.

docs/builds/guides/frameworks/vuejs.md
```diff
@@ -0,0 +1,645 @@
...
+				editorType: ClassicEditor,
+
+				// ...
+			};
+		}
+	}
+</script>
+```
+
+## Using CKEditor built from source
```
Reinmar: There are two main ways to include the editor – via a build and from source. In the latter scenario, we should not use the term "editor build" because it does not apply. There's no intermediate file create for that.

Therefore, the "Advanced setup" guide, even though it describes two methods, it does not use the term "editor build" in `builds/guides/integration/advanced-setup.html#scenario-2-building-from-source`. 

I think that the Vue's guide should follow the React guide where in the "from source" section we use the most direct, close method possible – importing everything directly. No additional file is created. The goal there is to show the absolute minimum setup, something most "native". Then we can link to the "Advanced setup" guide with "you can learn about other methods and scenarios in".

oleq: I rephrased the documentation. Check out and tell me if it makes sense now.

Reinmar:
I still don't like that we're proposing to use some intermediate `ckeditor.js` file with the build-like setup. It's a complicated method that we should avoid if possible.

PR #1345, opened 2018-11-07
Switched development environment to Yarn

### LINKED ISSUE #1214: Upgrade to lerna 3 or change it to something else
It introduced many backwards incompatible changes. We need to align our setup to it because people will be installing it instead of the support learna@2 due to how our [docs](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/development-environment.html) install it.

Perhaps, the right moment to update all this will be when we'll release a new version of mgit2 because that guide will need to update then anyway.

pomek:

Other: Switched development environment to Yarn. Closes #1214.

---

### Additional information

- CI for all packages will fail because it requires changes from https://github.com/ckeditor/ckeditor5-dev/pull/453. 
- Build that confirms that changes work: https://travis-ci.org/ckeditor/ckeditor5-table/builds/451805447
- All packages (except the builds) contain a branch `ckeditor5/t/1214` that contains changes for `ckeditor5-*` packages (https://github.com/ckeditor/ckeditor5-font/compare/ckeditor5/t/1214)

Side note: Indentations in `.travis.yml` have changed once again because I used mgit and a script (specified below) for modifying the configuration file. 

```js
#!/usr/bin/env node

const path = require( 'path' );
const fs = require( 'fs' );
const cwd = process.cwd();
const configPath = path.join( cwd, '.travis.yml' );

if ( !fs.existsSync( configPath ) ) {
	return;
}

// https://www.npmjs.com/package/js-yaml
const YAML = require( 'js-yaml' );
const config = YAML.load( fs.readFileSync( configPath, 'utf-8' ) );

config.sudo = 'required';

config.dist = 'trusty';

config.addons = {
	firefox: 'latest',
	apt: {
		sources: [ 'google-chrome' ],
		packages: [ 'google-chrome-stable' ]
	}
};

config.language = 'node_js';

config.node_js = [ '8' ];

config.cache = { yarn: true };

config.branches = {
	except: [ 'stable' ]
};

config.before_install = [
	'export DISPLAY=:99.0',
	'sh -e /etc/init.d/xvfb start',
	'npm i -g yarn'
];

config.install = [
	'yarn add @ckeditor/ckeditor5-dev-tests',
	'ckeditor5-dev-tests-install-dependencies'
];

config.script = [ 'ckeditor5-dev-tests-travis' ];
config.after_success = [ 'ckeditor5-dev-tests-save-revision' ];

fs.writeFileSync( configPath, YAML.safeDump( config, { flowLevel: -1, lineWidth: 10000} ), 'utf8' );
```


Reinmar:
Is there a chance that changes in `package.json` (merged from `master`) caused this? I can see that the committed version of deps are old:

image

.travis.yml
```diff
@@ -7,25 +7,25 @@ addons:
...
-  - npm install
-  - npm install mgit2 lerna
+  - yarn add mgit2 --ignore-workspace-root-check
   - mgit bootstrap --resolver-url-template="https://github.com/\${ path }.git"
-  - lerna bootstrap
+  - yarn
 script:
-  - npm t -- --reporter=dots
-  - npm run docs:api -- --validate-only
+  - yarn run test --reporter=dots
```
Reinmar: Can it be like this?

```suggestion
  - yarn test --reporter=dots
```

ma2ciek:
Changing the `package.json` or calling `yarn upgrade` may change the `yarn.lock`. Calling `yarn install` shouldn't change.

.travis.yml
```diff
@@ -7,25 +7,25 @@ addons:
...
+  - yarn run docs:api --validate-only
```
Reinmar: Also no need to `run`.

docs/framework/guides/contributing/development-environment.md
```diff
@@ -102,7 +100,7 @@ lrwxr-xr-x    1 p  staff    25 31 Jan 10:37 ckeditor5-engine -> ../../../ckedito
 If everything worked correctly, you should be able to run some tests:
 
 ```bash
-npm run test -- --files=core
+yarn run test --files=core
```
Reinmar:
```suggestion
yarn test --files=core
```

docs/framework/guides/contributing/development-environment.md
```diff
@@ -117,48 +115,32 @@ git pull
...
 
-Lerna is a tool used by many well-known projects such as [Babel.js](https://github.com/babel/babel). It has an amazing community and, relying on it ourselves, we hope that it will become a standard for managing multi-package projects.
-
 ## Running tests
 
 In order to run tests you need to use the `test` and `manual` tasks.
 
 ```bash
-npm test -- --watch --coverage --source-map --files=engine
+yarn run test --watch --coverage --source-map --files=engine
```
Reinmar:
```suggestion
yarn test --watch --coverage --source-map --files=engine
```

.travis.yml
ma2ciek: `yarn` allows to do it. But IMO it might be risky to do it in `.travis` file.

.travis.yml
Reinmar: Actually, I made a couple of suggestions here in this PR, but it'll be easier to change this automatically in all places. `s/yarn run/yarn` seems to work for me.

ma2ciek:
IMO using `yarn <our command>` is risky as we don't know if `yarn` won't introduce its own command under the same name in the future.

pomek:
image

This is a reason why I propose using `yarn run ...`

Reinmar:
OK, good points.

Reinmar:
I miss the old good `lerna clean` task. Can you add `yarn run clean` which would do the same thing?

Reinmar:
Also, we need to update this in `package.json`:

```
    "reset": "rm -rf ./packages ./node_modules && mgit bootstrap && lerna bootstrap",
    "reinstall": "lerna clean --yes && lerna bootstrap",
```

Reinmar:
I left 2 minor comments in https://github.com/ckeditor/ckeditor5-dev/pull/453/files

Other than that, I'd love if we could return to the previous `.travis.yml` formatting. Otherwise, if we'd use `travis encrypt` in new packages (or existing ones) the formatting will differ.

pomek:
> Other than that, I'd love if we could return to the previous .travis.yml formatting. 

Done.

PR #1368, opened 2018-11-22
t/ckeditor5-core/146: Documented a possible use–case of config.extraPlugins.

oleq:

Docs: Documented a possible use–case of `config.extraPlugins` (see ckeditor/ckeditor5-core#146).

---

### Additional information

A part of https://github.com/ckeditor/ckeditor5-core/pull/149.

### DIFF
--- a/docs/builds/guides/integration/configuration.md
+++ b/docs/builds/guides/integration/configuration.md
@@ -64,7 +64,58 @@ ClassicEditor.builtinPlugins.map( plugin => plugin.pluginName );
 
 ## Adding features
 
+### Adding complex features
+
 As CKEditor builds do not include all possible features, the only way to add more features to them is to {@link builds/guides/development/custom-builds create a custom build}.
+
+### Adding simple (standalone) features
+
+There is an exception to every rule. Although it is impossible to add plugins that have dependencies to {@link api/core `@ckeditor/ckeditor5-core`} or {@link api/engine `@ckeditor/ckeditor5-engine`} (that includes nearly all existing official plugins) without rebuilding the build, it is still possible to add simple, **dependency-free** plugins.
+
+You can do that using the {@link module:core/editor/editorconfig~EditorConfig#extraPlugins `config.extraPlugins`} configuration. The {@link module:core/plugin~PluginInterface plugin interface} allows plugins to be simple functions and you can define them in just a few lines, for instance:
+
+```js
+function MyPlugin( editor ) {
+	// ...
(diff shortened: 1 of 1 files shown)


docs/builds/guides/integration/configuration.md
```diff
@@ -64,8 +64,59 @@ ClassicEditor.builtinPlugins.map( plugin => plugin.pluginName );
 
 ## Adding features
 
+### Adding complex features
+
 As CKEditor builds do not include all possible features, the only way to add more features to them is to {@link builds/guides/development/custom-builds create a custom build}.
 
+### Adding simple (standalone) features
+
+There is an exception to every rule. Although it is impossible to add plugins that have dependencies to {@link api/core `@ckeditor/ckeditor5-core`} or {@link api/engine `@ckeditor/ckeditor5-engine`} (that includes nearly all existing official plugins) without rebuilding the build, it is still possible to add simple, **dependency-free** plugins.
```
scofalik: > rebuilding the build

Maybe "rebuilding the editor" ?

docs/builds/guides/integration/configuration.md
```diff
@@ -64,8 +64,59 @@ ClassicEditor.builtinPlugins.map( plugin => plugin.pluginName );
...
+
+You can do that using the {@link module:core/editor/editorconfig~EditorConfig#extraPlugins `config.extraPlugins`} configuration. The {@link module:core/plugin~PluginInterface plugin interface} allows plugins to be simple functions and you can define them in just a few lines, for instance:
+
+```js
```
scofalik: Do we post examples of "ES5" classes in other parts of docs?

Reinmar: That's not an ES5 class, it's a mention that you can write a simple function plugin. Perhaps if it was named `myPlugin()` it would be clearer, but I'd keep it like this.

Reinmar:
> I am not sure, but I have a feeling that it should work even if the plugin has dependencies, as long as everything is available (and it was because I had the snippet in that feature's repo).

It'll certainly work. But this guide is about adding plugins to builds. This guide should not go further than explaining this because it will create confusion.

> Second thing is that I miss mentioning `Plugin` class from core.

No need to mention it at it doesn't change anything here. You can write such standalone plugins as a simple function or class (which gets the constructor). Maybe a comment that "init()" get called automatically might clarify this a bit, but since we mention it I don't think that it will be confusing.

PR #1444, opened 2019-01-07
[Docs] Introduced CSS Frameworks compatibility. [skip ci]

### LINKED ISSUE #1421: Docs: Extend the bootstrap modal integration docs to consider other frameworks (or generalise it)
A follow-up of https://github.com/ckeditor/ckeditor5/issues/1409.

It seems that there are other frameworks providing modals have the same problem with CKEditor5.  We mention in our [docs](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/overview.html#compatibility-with-bootstrap):

> Configure Bootstrap to not steal focus from rich text editor fields.

And it is actually done in the linked code sample (but for Bootstrap only). 

I think we need do something because more users will have the same problem in the future with other frameworks:

* We could either make a [FAQ entry](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/faq.html) explaining the topic in a more generic manner.
* We could extend the [frameworks compatibility guide](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/overview.html) to consider other frameworks.

cc @AnnaTomanek @dkonopka @Reinmar

dkonopka:

Docs: Introduced CSS Frameworks compatibility section. Closes #1421.

### DIFF
--- a/docs/builds/guides/frameworks/css.md
+++ b/docs/builds/guides/frameworks/css.md
@@ -1,0 +1,98 @@
+---
+menu-title: CSS Frameworks
+category: builds-integration-frameworks
+order: 50
+---
+
+# Compatibility with CSS Frameworks
+
+Some of popular CSS Frameworks uses stronger [CSS Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) of their components that's why sometimes it can overrides CKEditor 5 elements like floating balloons (e.g. URL input).
+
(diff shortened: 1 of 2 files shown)


AnnaTomanek:
This should actually go to `stable`, right? Because it could be published anytime, and we could start linking to it/promoting it before the next release?

Issue #1472, opened 2019-01-18
Toolbar group plugin


dashingswags:
🆕 Feature request

For CKEditor 5
Worked on a plugin that does something like this

## 📋 Steps to reproduce

Add this to the editor config:
 
 ```
     toolbar: {
                items: [ 'bold', 'italic','underline','highlight','toolbarGroup']
     },
     toolbarGroup: {
                options: [
                    { model: 'paragraph', title: 'Paragraph' },
                    { model: 'heading1',  title: 'Heading 1' },
                    { model: 'heading2', title: 'Heading 2' },
                    { model: 'link',  title: 'Insert Link' },
                    { model: 'ckfinder',title:'Upload Image'}
                ]
       },
```

## ✅ Expected result

![screenshot 180](https://user-images.githubusercontent.com/15422269/51387842-ddc72080-1b27-11e9-8ae6-6d6c3cb547ef.png)

## 📃 Other details that might be useful

Can work with any of the builds.
Developed it for apps where grouping independent plugins might be necessary for better UI as i required something like that on the app iam working on. would like to know if i can go ahead to contribute this plugin to the community

---

If you'd like to see this feature implemented, please add 👍to this comment.

Reinmar:
Hi! Thanks for the ticket and ❤️ for the PR (https://github.com/ckeditor/ckeditor5/pull/1473). It looks great!

Ability to group toolbar items in dropdowns via a declarative API was something that we've discussed in the past, but I can't find the ticket. This is a pitty because that old ticket contained a lot of ideas 😞. Maybe one of my colleagues will be able to find it (cc @oleq @oskarwrobel).

Before we'll move to the implementation, we need to figure out what we want to do. Your solution focuses on adding the "more" option. In fact, this is more a space-saving solution which is in line with what we've been researching in https://github.com/ckeditor/ckeditor5/issues/894 and https://github.com/ckeditor/ckeditor5/issues/416. However, it rather seems that such a grouping should be automatic – we can't expect people to configure their editors always for the smallest screen size.

I like the UI of your solution, though. It's cleaner that what we had so far. However, I'm not sure how dropdowns would look in it. Perhaps it would work with normal buttons only (which could be fine too).

Automatic grouping is one thing. The other is declarative grouping. This, however, should be more flexible than just ability to create one group. We need something like:

```js
	toolbar: {
		items: [
			'heading',
			'|',
			'bold',
			'italic',
			'link',
			{
				type: 'dropdown',
[...]

PR #1473, opened 2019-01-18
Developed a plugin to group toolbar items

dashingswags:
This package contains CKEditor 5 features allowing to group multiple toolbar items in a dropdown list using any custom ckeditor5 build

1. Add "toolbargroup" to the toolbar.items array as shown below
2. add a toolbarGroup object to the editor.config and set the required parameters.

The toolbarGroup accepts an option array with a model and title property respectively.

Values for the model is same as toolbar.items array, while the values for the title is optional. if you want a custom title, then you can set that option as desired

### DIFF
--- a/packages/ckeditor5-toolbar-group/README.md
+++ b/packages/ckeditor5-toolbar-group/README.md
@@ -1,0 +1,34 @@
+## Documentation
+```html
+This package contains CKEditor 5 features allowing to group multiple toolbar items in a dropdown list using any custom ckeditor5 build
+
+1. Add "toolbargroup" to the toolbar.items array as shown below
+2. add a toolbarGroup object to the editor.config and set the required parameters.
+
+The toolbarGroup accepts an option array with a model and title property respectively.
+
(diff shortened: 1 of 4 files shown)


Reinmar:
Hi! This PR is really amazing – thank you for sharing it!

I started analyzing in https://github.com/ckeditor/ckeditor5/issues/1472 what we really need, though, and I'm not sure if this is the right way to propose this functionality. And what kind of functionality we need. I'll keep your PR open for the time being, before my colleagues comment on this, but I'm afraid we'll need to implement it in a different way. Still, your plugin is very interesting and I'd love to see it on npm – I'm sure other developers will enjoy using it.

Reinmar:
My late review landed in https://github.com/ckeditor/ckeditor5/issues/1472#issuecomment-485793801. As I explained there, such a feature that you proposed would make a great sense as a 3rd party addon. Thanks again for the great PR, but we'd like to go in a different direction with the core.

PR #1481, opened 2019-01-23
Removed warning about CRA@2

pomek:

Docs: Removed warning about CRA@2. See https://github.com/ckeditor/ckeditor5-react/issues/40#issuecomment-456368847.

### DIFF
--- a/docs/builds/guides/frameworks/react.md
+++ b/docs/builds/guides/frameworks/react.md
@@ -14,16 +14,6 @@ The easiest way to use CKEditor 5 in your React application is by choosing one o
 
 ## Quick start
 
-<info-box>
-	### Compatibility with `create-react-app@2`
-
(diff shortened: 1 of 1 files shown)


ma2ciek:
Without ejecting the SVGs are broken with our beloved error:

```
TypeError: Cannot read property 'getAttribute' of null
at IconView._updateXMLContent (iconview.js:121)
    at IconView.render (iconview.js:94)
	...
```

pomek:
> Is it possible to run react app without ejecting config now?

That's the goal of changing our sources…

> Without ejecting the SVGs are broken with our beloved error:

The later in the same documentation:

> **The configuration needs to be ejected in order to be able to customize webpack configuration.** To build CKEditor 5 from source you must load inline SVG images and handle CKEditor 5's CSS as well as correctly minify the ES6 source.

You can use ready-to-use builds without ejecting.

pomek:
But if you would use CKEditor 5 Framework, you must eject the configuration.

pomek:
Or create a new build and attach it to the React Component. Both ways are described in the documentation.

PR #1497, opened 2019-01-31
Slack integration

### LINKED ISSUE #1493: Integrate Travis into our Slack
We had notifications about failed Travis jobs (only for https://travis-ci.org/ckeditor/ckeditor5) posted in the `#ckeditor5-code` channel. We stopped getting those, probably when we renamed the channel.

However, it was not very useful because there's a lot of mess with new issues/prs/forks. So, I'd rather use a new channel (`#cke5-ci`).

The other improvement could be to post only results for jobs failed on `master` and `master-revisions` branches. This way we'll avoid being pinged about failing jobs for constellation branches.

Do we have other ideas regarding possible improvements?

pomek:

Internal: Replaced an integration with Slack provided by Travis with a custom script that sends a notification for specified branches. Closes #1493.

---

Unfortunately, it does not detect a change of status between two builds. But the notification will be sent only for `push` builds and only for `master` and `master-revisions` branches. I guess it won't create a huge mess.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
   fi'
 after_script:
 - export END_TIME=$( date +%s )
-- yarn add slack-notify @octokit/rest@^16.13.4 --ignore-workspace-root-check
+- yarn add slack-notify --ignore-workspace-root-check
 - node ./scripts/notify-travis-status.js
 env:
   global:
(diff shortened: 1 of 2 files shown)


pomek:
image

pomek:
After the latest changes, the script provides such notifications:

image

Reinmar:
Looking forward to see first info on #cke5-ci :D

Issue #1520, opened 2019-02-12
Get rid of symbols used to recognize view elements in features


oleq:
A follow–up of https://github.com/ckeditor/ckeditor5/issues/1514.

---

When writing the following converters which extend the stock Link feature

```
editor.model.schema.extend( '$text', { allowAttributes: 'linkTarget' } );

editor.conversion.for( 'downcast' ).attributeToElement( {
    model: 'linkTarget',
    view: ( attributeValue, writer ) => {
        return writer.createAttributeElement( 'a', { target: attributeValue }, { priority: 5 } );
    },
    converterPriority: 'low'
} );

editor.conversion.for( 'upcast' ).attributeToAttribute( {
    view: {
        name: 'a',
        key: 'target'
    },
    model: 'linkTarget',
    converterPriority: 'low'
} );
```

I found out that links that were processed by the target (later) upcast converter don't display the UI when clicked.

I researched and I found out we [use symbols (custom props) to mark links](https://github.com/ckeditor/ckeditor5-link/blob/master/src/utils.js#L25-L37) and then we [look for them when recognizing them](https://github.com/ckeditor/ckeditor5-link/blob/master/src/utils.js#L21-L23).

Custom props are pruned when processed by the custom converter so links with `target` were no longer recognized as links. And if a link isn't a link... it does not trigger the balloon UI.

I see 2 solutions (we can apply both):
* Get rid of symbols–as–tags in link and other features (image, image caption, media embed, widget).
* Fix the conversion so the custom props are preserved in this case. We may use them in the future for something else (not related to tag–symbols) and they will be broken anyway.

Quick PoC in the link feature shows that all tests still pass when droppped tag–symbols.

Reinmar:
> Custom props are pruned when processed by the custom converter so links with `target` were no longer recognized as links. And if a link isn't a link... it does not trigger the balloon UI.

I'm sorry, but it seems I didn't get this. Can you elaborate?

oleq:
Closed by PRs in respective packages.

PR #1521, opened 2019-02-12
t/1520: Got rid of the symbol used to recognise view elements

oleq:

CI constellation–only PR for #1520.

PRs:
* https://github.com/ckeditor/ckeditor5-link/pull/213
* https://github.com/ckeditor/ckeditor5-image/pull/275
* https://github.com/ckeditor/ckeditor5-media-embed/pull/76
* https://github.com/ckeditor/ckeditor5-table/pull/171
* https://github.com/ckeditor/ckeditor5-widget/pull/71

### DIFF
--- a/mgit.json
+++ b/mgit.json
@@ -27,20 +27,20 @@
     "@ckeditor/ckeditor5-font": "ckeditor/ckeditor5-font",
     "@ckeditor/ckeditor5-heading": "ckeditor/ckeditor5-heading",
     "@ckeditor/ckeditor5-highlight": "ckeditor/ckeditor5-highlight",
-    "@ckeditor/ckeditor5-image": "ckeditor/ckeditor5-image",
-    "@ckeditor/ckeditor5-link": "ckeditor/ckeditor5-link",
+    "@ckeditor/ckeditor5-image": "ckeditor/ckeditor5-image#t/ckeditor5/1520",
+    "@ckeditor/ckeditor5-link": "ckeditor/ckeditor5-link#t/ckeditor5/1520",
     "@ckeditor/ckeditor5-list": "ckeditor/ckeditor5-list",
     "@ckeditor/ckeditor5-markdown-gfm": "ckeditor/ckeditor5-markdown-gfm",
-    "@ckeditor/ckeditor5-media-embed": "ckeditor/ckeditor5-media-embed",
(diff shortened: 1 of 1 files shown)


Reinmar:
The changes look good. But I wonder if we should be more precise in checking whether something is some element. Like checking also that `<figure class=image>` not only has `image` class but also is a `figure`. WDYT?

Reinmar:
Actually... I wonder if we can make it bulletproof in one case – when someone changes the output format of e.g. the image feature. That it's not a `figure` but something else. In fact, if someone does it – what else might break? If this is the only thing which might break, then it'd be worth considering doing this in a smarter way. We can't use symbols, but perhaps we could use something else.

oleq:
Or

> * Fix the conversion so the custom props are preserved in this case. We may use them in the future for something else (not related to tag–symbols) and they will be broken anyway.

Reinmar:
> Custom props are pruned when processed by the custom converter so links with `target` were no longer recognized as links. And if a link isn't a link... it does not trigger the balloon UI.

I'm sorry, but it seems I didn't get this. Can you elaborate?

Reinmar:
Oh... I think I get it. The problem isn't "pruning" (like – literally removing things)... the problem is more complex.

When you have two converters which create two attribute elements and these attribute get merged (`<a href="x"><a target="y">foo</a></a>`), that merging logic does not care about custom props. It forgets about them. We could indeed fix that.

Still, using symbols was rather unnecessary. Plus, we weren't doing this consistently – we were doing this for a small percentage of identifiers. So the change to remove symbols, in general, would be good. But I'm not sure if we shouldn't just replace them with normal strings.

Reinmar:
Or another option – expose symbols (e.g. on plugin instances) + fix the custom properties handling.

The goal, in all those cases, is to allow tinkering with features. So, to e.g. write a custom converters which define different view representation of a feature but still works with the UI.

Reinmar:
Let's talk about this tomorrow. @pjasiun and @scofalik are for fixing custom props handling and not using symbols (but strings instead). I'm 👍 with that, but we need to talk about the details. Plus... explain somewhere why we made such a decision because, in general, it's not optimal but we don't have time now to make bigger changes here.

oleq:
Replaced by the https://github.com/ckeditor/ckeditor5/pull/1532 constellation with a string approach.

PR #1586, opened 2019-02-28
Docs: Multiple root editor sample and tutorial added

### LINKED ISSUE #1447: Provide a docs example on how to create custom editor with multiple roots
Extracted from https://github.com/ckeditor/ckeditor5-editor-decoupled/issues/21 (and blocked by #1445 ATM).

We have concluded (together with @pjasiun and @Reinmar) that the first step towards multiple root editor should be providing a sample in docs which will show:

* How to create a custom editor creator which enables usage of multiple roots.
* In the same time it should be a solid example showing how to create custom editor creator.

The sample itself should be added under `Examples > Framework` section:

image

And the tutorial with all the code in `Framework > Guides > Architecture > Deep dive > User interface` section:

image

f1ames:

Docs: Multiple root editor sample and tutorial added. Closes #1447.

---

### Additional information

See #1447 and https://github.com/ckeditor/ckeditor5-editor-decoupled/issues/21.

There are two things I'm not 100% sure about. The naming of the editor type which is **Multiple root editor** - generally it makes sense for us, but not sure how easy to find it will be for the users/developers interested in such editor type. I took a look on other editors and usually it is something like **shared toolbar**, but it covers a little different cases.

The other thing is the content of the sample itself, I used the content from Inline build sample, because it has four containers so works nicely with multiple root editor and can be used to see what is the difference between multiple root editor and the approach using multiple editor instances. My doubt is that it may be confusing that we have two samples with the same content.


Reinmar:
Wow🎉 It works and looks great. From my side I see that some wording will need to be polished, but nothing big. Great job :)

Reinmar:
Reviewed from my side. Leaving the source code to @oleq.

Reinmar:
image

docs/_snippets/examples/multi-root-editor.js
```diff
@@ -0,0 +1,378 @@
...
+			this.focusTracker.add( editableElement );
+
+			// Let the editable UI element respond to the changes in the global editor focus
+			// tracker. It has been added to the same tracker a few lines above but, in reality, there are
+			// many focusable areas in the editor, like balloons, toolbars or dropdowns and as long
+			// as they have focus, the editable should act like it is focused too (although technically
+			// it isn't), e.g. by setting the proper CSS class, visually announcing focus to the user.
+			// Doing otherwise will result in editable focus styles disappearing, once e.g. the
+			// toolbar gets focused.
+			editable.bind( 'isFocused' ).to( this.focusTracker );
```
oleq: Why all roots are marked as focused when just one has focus at a time? This is not the right UX. 

image


The relation with the FT should be different in this multi–root editor:
* if any editable is focused -> `FT#isFocused === true`
* if an editable is focused, only that editable gets focused styles as per [`EditableUIView`](https://github.com/ckeditor/ckeditor5-ui/blob/master/src/editableui/editableuiview.js#L121-L138)

docs/framework/guides/custom-editor-creator.md
```diff
@@ -0,0 +1,429 @@
...
+			this.focusTracker.add( editableElement );
+
+			// Let the editable UI element respond to the changes in the global editor focus
+			// tracker. It has been added to the same tracker a few lines above but, in reality, there are
+			// many focusable areas in the editor, like balloons, toolbars or dropdowns and as long
+			// as they have focus, the editable should act like it is focused too (although technically
+			// it isn't), e.g. by setting the proper CSS class, visually announcing focus to the user.
+			// Doing otherwise will result in editable focus styles disappearing, once e.g. the
+			// toolbar gets focused.
+			editable.bind( 'isFocused' ).to( this.focusTracker );
```
oleq: See my comment in `MultirootEditorUI#init()`. All editables cannot share the same `isFocused` state.

docs/framework/guides/custom-editor-creator.md
```diff
@@ -0,0 +1,429 @@
...
+
+			// Bind the editable UI element to the editing view, making it an end– and entry–point
+			// of the editor's engine. This is where the engine meets the UI.
+			editingView.attachDomRoot( editableElement, editable.name );
+		}
+
+		this._initPlaceholder();
```
oleq: How does it work? I assume the same placeholder text is set to all roots, which does not make sense in most cases IMO. It would take `editor.config.placeholder` defined as `{ root: 'Root placeholder text', ... }` to make this right.

Reinmar:
I'm going to merge this because it actually works (with some minor issues). Let's fix the rest on `stable`.

docs/framework/guides/custom-editor-creator.md
f1ames: > I assume the same placeholder text is set to all roots, which does not make sense in most cases IMO.

Yes, this is the way it works in this implementation. It would make much more sense to use it with `{ root: 'Root placeholder text', ... }` as you mentioned, however it will require some changes in `placeholder` code (probably mostly docs?). I assume we could cover this as a follow-up?

docs/_snippets/examples/multi-root-editor.js
f1ames: We had some discussion with @pjasiun some time ago about this case - https://github.com/ckeditor/ckeditor5-editor-decoupled/issues/21#issuecomment-448943315 (the outcome was that it is acceptable). However, if it can be easily changed with the current API, I'm in favor of making this change to have only one root marked as focused at the same time.

pjasiun: If you want to change it, that's fine, but TBH I like it this way. It stresses the fact that this is a single editor.

PR #1605, opened 2019-03-07
Improved the React integration guide

### LINKED ISSUE #1597: [Docs] Does CRA2 still require replacing webpack.optimize.UglifyJsPlugin?
image

Is it still needed? Could we review the guide and move all pieces which are required for CRA1 to a separate section at the end?

pomek:

Docs: Improved the React integration guide. Introduced a section about `create-react-app@2`. Closes #1597.


pomek:
image

Reinmar:
Does this apply to CRA2?

image

pomek:
In CRA@2 eject is required only if a user wants to use CKEditor 5 Framework because the webpack configuration requires loaders for CSS and SVG files.

I don't remember how it was in CRA1 but the argument about SVG/CSS seems to be valid in this case.

Reinmar:
@pomek, could you check my changes?

pomek:
LGTM except one thing:

image

The blue header could be attached to CRA@1 section.

PR #1733, opened 2019-04-29
Docs: updated fields image size in the `Image` feature.

### LINKED ISSUE #1717: [Docs] Image used in the docs is too small
The image we use is only 800px wide hence the white margins on its side.

image

We should:

* use a wider image there (1000px will do)
* ask ourselves why we don't use `display:table` for `figure.image`?

cc @oleq @dkonopka

dkonopka:

Docs: Updated fields image size in the `Image` feature. Closes #1717.

---

### Additional information

Changed image size from 800px to 960px.


mlewand:
Though technically this PR should be targeted at `stable` branch so that it gets to the documentation before the release. 😉

PR #1805, opened 2019-06-10
Command for printing content styles

### LINKED ISSUE #773: Content styles
## Current state in CKEditor 5

* Currently, the content is styled by features and these styles are built together with the editor UI styles.
* In a WYSIWYM editor it makes sense to propose content styles for editing. These styles don’t have to reflect the target page’s styles in 100%. After all, it’s all about the content. However, we're aware that in some cases developers will want to have the same styles in the editor as on the target pages.
* At the same time, certain features (tables, images) require careful styling also on the target page.
* The styles that we have are all prefixed with the `.ck-editor__editable` class. On the target pages such a class will not exist so those styles would not work OOTB anyway.

## CKEditor 4

In CKEditor 4 there's [`contents.css`](https://github.com/ckeditor/ckeditor-dev/blob/master/contents.css) – a separate file containing some of the feature's styles. The problem with this approach is that:

* It works automatically only in iframed editors. When you initialize an inline editor you need to load this style manually.
* It's really generic and not really useful for target pages. Actually, it's also not very useful for inline editors too because of its generic look.
* It's a single file which "knows" about all the features of the editor. Such file might have a tendency to grow. Especially that in CKEditor 5 more features use classes (e.g. font size feature, or highlights).
* In CKEditor 5 we can’t have a separate `contents.css` file loaded automagically like in CKEditor 4 because CKEditor 5 must not load any additional resources itself. It neither can do this (because it requires dubious tricks) nor should (because it severely breaks DX – bundling becomes a nightmare).

## The goal

The goal is to:

* Make it simpler for developers to learn about the content styles. To see a compiled version of all the editor content styles.
* If possible, make this stylesheet easily usable for them.

## How?

* Through a manually maintained guide in the docs. This is the simplest solution and perhaps completely satisfactory. We need some guide about this anyway so there's not a problem to copy&paste there all the features' styles. Of course, it means that we'll need to maintain this but these styles don't change so often. Also, to not forget about this we should have a comment in every content style CSS file saying that "a compilation of content styles can be found in".
  
    In this guide, we could also mention that you need to prefix all these styles with your container's class/id. Or we could even leave the `.ck-editor__editable` prefix or write them like:

   ```css
   .your-container img {
      ...
   }
   ```
* By some API method (e.g. `ClassicEditor.getContentStyles()`)?

   It could accept a prefix-selector to be added to the content styles to ensure proper specificity
Editor content styles often cannot be reused 1:1 to style target website because of their low-specificity nature.

   This would be a better solution because it would let you retrieve the styles of exactly the features that you use and we wouldn't need to maintain anything manually. However, it's far more tricky to implement (I don't even know how we could do that).

pomek:

Other: Added an npm script `docs:content-styles` which build an editor with all plugins and save all occurrence of`.ck-content` class into a separate CSS file. Closes #773.


pomek:
@oleq, could you look on this once again?

pomek:
@oleq, fixed.

pomek:
image

pomek:
(2) and (3) – fixed.

(4) could be reported as a follow-up.

oleq:
I posted some thought on the tool in https://github.com/ckeditor/ckeditor5/issues/773#issuecomment-517212368.

oleq:
@Reinmar Can you please take a look at the docs I created?

Reinmar:
That's a very good start. I've got one proposal, though – we could move "By generating it using the yarn docs:content-styles script in the root package of the editor" to framework/guides/contributing/development-environment.html because it's mainly for us. I'm not sure why would anyone else be interested in that. In the "content styles" guide we may have a short link to that, but I don't think that even this is necessary.

Reinmar:
BTW, the other option after adding `.ck-content` to your container would be to replace `.ck-content` with something custom. It should be fairly straightforward and safe.

PR #1819, opened 2019-06-17
Renamed and replaced usage: mgit => mrgit

pomek:

Other: Adjusted repository, documentation and scripts to changes done in `mgit/mrgit`. See: cksource/mgit2#85. 

---

### Additional information

1. Requires: https://github.com/cksource/mgit2/pull/106 and published package as `mrgit` (should it be version `1.0.0`?).
2. Requires reinstall dependencies in `yarn.lock` (I'd remove `yarn.lock` then `yarn install`)

IMO – this change is a BC for this repository. Even if someone won't update `mrgit`, the `mgit.json` file has been renamed to `mrgit.json` which mgit cannot understand. As a workaround, people can create a symlink `ls -s mrgit.json mgit.json` and it should work.


scripts/release/update-mgit-branches.js
```diff
@@ -18,12 +18,12 @@ if ( !branchName ) {
 const path = require( 'path' );
```
pomek: I forgot to rename this file…

Reinmar: Not a problem. We're removing it very soon.

PR #1835, opened 2019-06-25
Add word count plugin to mgit.json

msamsel:

Feature: Introduce word count plugin.

---

### Additional information

Requires: ckeditor/ckeditor5-word-count#2


Reinmar:
I did that manually already.

Also, when introducing a package it needs to be added in 3 places – mgit.json, package.json and README.md.

PR #1905, opened 2019-07-23
Simplified updating the "#stable" branches process

### LINKED ISSUE #1899: Use base branches instead of modified mgit.json on #stable
Since we introduced the support for more base branches than just `master`, we can now omit modifying `mgit.json` on `stable`.

pomek:

Internal: Simplified updating the `#stable` branches process during releases. Closes #1899.


Reinmar:
I'll also cherry-pick this to `#stable` now and remove `#stable` from `mrgit.json`.

Reinmar:
Or, actually, I'll leave it to you because we need to verify that the documentation builder works with mrgit and all those changes.

PR #1933, opened 2019-07-31
Snippet Adapter works with additional languages

### LINKED ISSUE #1932: Docs: Multiple languages support is needed for code snippets
ATM you can only configure only one language:

```
/* config { "language": "es" } */
```

but when the guide consists of few snippets with editors and each one of them with a different `editor.config.language`, you'll (most likely) need `additionalLanguages` in [`CKEditorWebpackPlugin`](https://www.npmjs.com/package/@ckeditor/ckeditor5-dev-webpack-plugin) to be configurable.

For instance, there can be 3 different editors: one in English, another one in Spanish, and another one Arabic. Something like this would solve the problem:

```
/* config { "additionalLanguages": [ "es", "ar" ] } */
```

and this would be possible:

```js
ClassicEditor
	.create( document.querySelector( '#english-editor' ) )

ClassicEditor
	.create( document.querySelector( '#spanish-editor' ), {
		language: 'es',
		// ...
	} )

ClassicEditor
	.create( document.querySelector( '#arabic-editor' ), {
		language: 'ar',
		// ...
	} )
```

**Note:** You can't configure `/* config { "language": "..." } */` per snippet in this situation because you'd build `ClassicEditor` thrice and end–up with "some modules were duplicated" error in the runtime.

cc @pomek @ma2ciek

pomek:

Docs: Snippet Adapter works with additional languages. Closes #1932.

### DIFF
--- a/scripts/docs/snippetadapter.js
+++ b/scripts/docs/snippetadapter.js
@@ -178,6 +178,13 @@ module.exports = function snippetAdapter( snippets, options, umbertoHelpers ) {
 					if ( wasCSSGenerated ) {
 						cssFiles.unshift( path.join( snippetData.relativeOutputPath, snippetData.snippetName, 'snippet.css' ) );
 					}
+
+					// Additional languages must be imported by the HTML code.
+					if ( snippetData.snippetConfig.additionalLanguages ) {
+						snippetData.snippetConfig.additionalLanguages.forEach( language => {
+							jsFiles.push( path.join( snippetData.relativeOutputPath, 'translations', `${ language }.js` ) );
+						} );
+					}
 				}
 
 				const cssImportsHTML = getHTMLImports( cssFiles, importPath => {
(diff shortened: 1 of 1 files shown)


pomek:
R- because one of the snippets does not work:

image

pomek:
Works!

image

scripts/docs/snippetadapter.js
```diff
@@ -261,6 +273,27 @@ function getWebpackConfig( snippets, config ) {
 		definitions[ definitionKey ] = JSON.stringify( config.definitions[ definitionKey ] );
 	}
 
+	const additionalLanguages = new Set();
+
+	// Find additional languages that must be built.
+	for ( const snippetData of snippets ) {
+		if ( snippetData.snippetConfig.additionalLanguages ) {
+			snippetData.snippetConfig.additionalLanguages.forEach( language => {
```
ma2ciek: I'd use `for ( const language of snippetData.snippetConfig.additionalLanguages || [] )` here to not mess `forEach` with `for of`.

oleq:
Looks like other guides (without additionalLanguages–enabled snippets) suffer some issues now:

image

ma2ciek:
Oh, Now I understand the problem. We create the `ckeditorWebpackPluginOptions` for all snippets at once, while some snippets specify only the `language` while the other specifies only the `additionalLanguages`. I guess that the easiest way is to:

* make the `ckeditorWebpackPluginOptions.additionalLanguages` at least an empty array (this will ensure that all translation files will be produced).
* add translations for the main language file (and translation files for `additionalLanguages` if set in the snippet) to the HTML file.

pomek:
I changed the way how additional languages are built. Instead of building everything together, I separated editors with multi-languages to a separate group. Thanks to that existing snippets weren't touched and multi-languages editors work fine.
I've checked ~10 random snippets and everything worked for me. Could you confirm?

oleq:
LGTM. @ma2ciek?

ma2ciek:
It looks good to me as well.

PR #1956, opened 2019-08-12
doc(advanced-setup): add section for Webpack Encore

Kocal:

Docs: Added docs for building from source in Webpack Encore

---

Hi, today I've had some issues while trying to make [Webpack Encore](https://github.com/symfony/webpack-encore) able to build CKEditor5, but I've finally make it works and I think it will be helpful for other people.

Thanks

### DIFF
--- a/docs/builds/guides/integration/advanced-setup.md
+++ b/docs/builds/guides/integration/advanced-setup.md
@@ -194,12 +194,14 @@ Encore.
   	})
 	
 	// Configure PostCSS loader
-	.enablePostCssLoader(options => {
-    	Object.assign(options, styles.getPostCssConfig({
+	.addLoader({
+    	test: /ckeditor5-[^/\\]+[/\\].+\.css$/,
+    	loader: 'postcss-loader',
+    	options: styles.getPostCssConfig({
       		themeImporter: {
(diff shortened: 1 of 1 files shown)


oleq:
I'm glad you managed to get it to work. And thanks for the PR! 

I'm not sure this is the right place for this kind of docs, though. Feels more like Builds/Guides/Frameworks but we need to figure out where to keep it because this section is devoted mainly to official CKEditor 5 components for 3rd-party frameworks. 

cc @Reinmar

Kocal:
Hum, I don't really know.
A separate page for Webpack Encore would be a copy/paste from Webpack page, but with a different `webpack.config.js` :confused:

Reinmar:
I think this place is ok. It's not ideal but we don't have something more fitting.

I also had doubts whether we should cover such things because there are naturally lots of wrappers and environments similar to this. But I guess we can cover those as well. If we'll have a problem with maintaining it (e.g. we'll change something in our configuration and won't be able to tell how to change this section of the guide) we can always either remove it or mark as outdated and waiting for contributors. The latter choice would actually be completely fine to me.

Reinmar:
Merged! Thanks for your contribution. I'll fix the code style on `master`.

Issue #1960, opened 2019-08-14
mrgit is not used on stable branch


mlewand:
Task

## 📋 Steps to reproduce

#1819 should have been merged also to the stable branch. It will cause clean docs build to fail, let's fix it.

mlewand:
Fixed with #1961.

PR #1961, opened 2019-08-14
Ported mrgit change to the stable branch

mlewand:

Internal: Ported mrgit change to the stable branch #1960.


mlewand:
Tested, works fine. The only thing would be bumping `yarn.lock`, but this will be updated once we merge master to the stable branch.

PR #1962, opened 2019-08-14
t/ckeditor5-paste-from-office/79: Created a new "Paste from office" category in feature guides

oleq:

Docs: Created a new "Paste from office" category in feature guides (see ckeditor/ckeditor5-paste-from-office#79).

---

### Additional information

A piece of https://github.com/ckeditor/ckeditor5-paste-from-office/pull/82.


docs/umberto.json
```diff
@@ -216,6 +216,11 @@
 					"name": "Image upload",
 					"id": "features-image-upload",
 					"slug": "image-upload"
+				},
+				{
+					"name": "Paste from office",
```
AnnaTomanek:
```suggestion
					"name": "Paste from Office",
```
"Paste from office" is confusing, as "office" is more about the physical location than an app. Still, if this category is going to include Microsoft Word and Google Docs, then "Office" is not the best name for the category. "Office" would be fine for Word and Excel, but not really Google Docs. "Paste from external applications"? "Paste from other applications"? Just "Pasting"? The last one would make sense if the category also included some more generic stuff about e.g. pasting content from other websites or why the content you paste is stripped. Do we plan any such content?

Reinmar: Can we extract this discussion to https://github.com/ckeditor/ckeditor5-paste-from-office/issues/83? I bet we'll be asking ourselves in 2 years time why we called it in a certain way and this will be easier to find in a ticket :D

oleq: Yeah, it seems like a confusing issue and we can discuss it somewhere else. But for now, we need to go with something and "Paste from Office" is a no–go for me as a category for the "Paste from Google Docs" guide.

AnnaTomanek: Let's go with "Pasting" simply then, maybe?

Reinmar: ➡️  https://github.com/ckeditor/ckeditor5-paste-from-office/issues/83 ;P

PR #1968, opened 2019-08-16
The building docs process will continue if JSDoc fails due to errors

pomek:

Internal: The building docs process will continue if JSDoc fails due to errors during the validation. See ckeditor/ckeditor5-dev#550.

---

### Additional information

Requires: https://github.com/ckeditor/ckeditor5-dev/pull/551


pomek:
image

scripts/docs/build-docs.js
```diff
@@ -27,7 +27,10 @@ function buildDocs() {
 	if ( skipApi ) {
 		promise = Promise.resolve();
 	} else {
-		promise = buildApiDocs();
+		promise = buildApiDocs()
+			.catch( () => {
```
ma2ciek: We can miss a random error easily with such catching.

pomek: I didn't find a better way.

ma2ciek: even the `catch( err => { console.error( err ); process.exitCode = 1; } )` will be better.

PR #2019, opened 2019-09-02
Unify the asset regexps

### LINKED ISSUE #1046: [Docs] Fix regexp in webpack configuration

Reinmar:

Docs: Unify the asset regexps. Closes #1046.

### DIFF
--- a/docs/builds/guides/frameworks/react.md
+++ b/docs/builds/guides/frameworks/react.md
@@ -203,7 +203,7 @@ Then, add two new elements to the exported object under the `module.rules` array
 	use: [ 'raw-loader' ]
 },
 {
-	test: /ckeditor5-[^/\\]+[/\\]theme[/\\].+\.css/,
+	test: /ckeditor5-[^/\\]+[/\\]theme[/\\].+\.css$/,
 	use: [
 		{
 			loader: 'style-loader',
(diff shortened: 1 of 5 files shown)


Reinmar:
@Mgsy could you check one of the guids – e.g. a tutorial for building a block widget that it still works. First couple of steps will be enough – to verify that svgs and css are loaded correctly.

PR #5609, opened 2019-10-17
[Docs] Improved RWD of demos in Examples section

### LINKED ISSUE #5598: RWD of the examples gives strange results
I know that no one is resizing the browser window like that IRL, but:

* developers are testing CKE5 so they may do that,
* it's actually kinda a mistake that when the ToC space disappears, the editor gets that wide,
* there's a point where the editor overflows the viewport and that's docs' issue IMO

Oct-16-2019 12-32-02

---

Also, this would be nice to improve. I think the max width can be further increased because now the editor will nicely adjust to that. As long as the "paper sheet" displayed in the editor has a constant width, this is fine.

image

dkonopka:

Docs: Improved RWD of demos in the Examples section. Closes #5598.

---

### Additional information

1. Enlarged container width of document editor demo

Screenshot 2019-10-17 at 10 48 03

2. Fixed document editor demo for mobiles because ATM it looks like this:

Before:
Screenshot 2019-10-17 at 10 24 28

### DIFF
--- a/docs/_snippets/examples/document-editor.html
+++ b/docs/_snippets/examples/document-editor.html
@@ -200,4 +200,40 @@
 	.document-editor .ck-content table td {
 		vertical-align: middle;
 	}
+
+	@media only screen and (max-width: 960px) {
+		/* Because on mobile 2cm paddings are to big. */
+		.document-editor__editable-container .document-editor__editable.ck-editor__editable {
+			padding: 1.5em;
+		}
+	}
+
+	@media only screen and (max-width: 1200px) {
+		.main__content-wide .document-editor__editable-container .document-editor__editable.ck-editor__editable {
+			width: 100%;
+		}
+	}
+
+	/* Style document editor a'ka Google Docs.*/
+	@media only screen and (min-width: 1360px) {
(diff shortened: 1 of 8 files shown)


Reinmar:
> Enlarged container width of document editor demo

Could we keep the textual content limited to some max-width there? So only the editor stretches? I think that this is the only demo where we need something like this, so it can be hardcoded there, I guess...

oleq:
> Could we keep the textual content limited to some max-width there?

I'm confused. Isn't it exactly what Damian did below?

Screenshot 2019-10-17 at 10 48 03

dkonopka:
> I'm confused. Isn't it exactly what Damian did below?

@Reinmar means title & paragraph above the editor. But it's not easiest thing to do, because we can't "catch" these elements easily...

```css
.main__content-inner .main__content-title,
.main__content-inner > p {
  max-width: some-number;
}
```

Screenshot 2019-10-17 at 11 26 59

dkonopka:
@Reinmar I'm not so sure about this change:

Screenshot 2019-10-17 at 11 37 38

Issue #1353, opened 2018-11-12
Should we switch our builds to Terser?


Reinmar:
Webpack 4 switched to `uglify-es` but according to https://github.com/mishoo/UglifyJS2/issues/3156#issuecomment-421509930 `uglify-es` is dead:

> > so it seems strange to abandon it. Since you haven't deprecated it on NPM, should I assume you're only taking a break from maintaining it?
> 
> `uglify-es` has been forked and the new project is called `terser`. For projects that need ES6 support, use it instead of `uglify-es`, since all new development will be happening over there (and things like webpack are switching over to it shortly too):
> https://github.com/fabiosantoscode/terser

Looking at the links leading to https://github.com/mishoo/UglifyJS2/issues/3156 everyone is switching to [Terser](https://github.com/terser-js/terser) now. [Webpack 5 will use Terser as well](https://github.com/webpack/webpack/issues/7923). 

Question – when should we do the transition ourselves? Should we wait for webpack 5 or do that before it?

One thing to consider would be – is Terser ready? According to [some comments](https://github.com/facebook/react/issues/13987#issuecomment-436689914) it's not particularly stable yet. Also, looking at how frequently this changes... I'm afraid that before webpack 5 will be released they'll change the decision again to something else (while this is meant to be a joke, I'd like to understand what's the background behind big players switching to Terser so suddenly).

Reinmar:
OK, it seems that webpack 4 may switch to terser as well because they consider this a non-breaking change:

https://github.com/webpack/webpack/issues/8350#issuecomment-437076507

ma2ciek:
So after the webpack changes the minifier it might be a good moment to change it too. Maybe terser isn't the best and most stable, but it's actively maintained. - see e.g. the https://github.com/terser-js/terser/issues/165.

Reinmar:
> it's actively maintained.

As I wrote in https://github.com/webpack/webpack/issues/8350#issuecomment-438265506, I'm not worried about today. I'm worried what will be next year :D

Reinmar:
It's high time to finally move to terser. Lots of docs to update because we have plenty of `uglifyjs-webpack-plugin` occurrences.

Reinmar:
Followup to review the docs in https://github.com/ckeditor/ckeditor5/issues/5836.

PR #5788, opened 2019-11-24
Replaced UglifyJS with Terser

pomek:

Docs: Replaced UglifyJS with Terser. See #1353.

---

More PRs below the mark – https://github.com/ckeditor/ckeditor5/issues/1353#event-2683863323


pomek:
I need to review docs so I marked the PR as WIP.

pomek:
I added a note about Terser to two guides but I guess, we should consider rewriting a part describing webpack when we'll use webpack@5.

Reinmar:
> I added a note about Terser to two guides but I guess, we should consider rewriting a part describing webpack when we'll use webpack@5.

Extracted to https://github.com/ckeditor/ckeditor5/issues/5836

Issue #5880, opened 2019-12-02
Performance research


Reinmar:
Let's start from creating a set of manual tests on which we'll be able to reliably and quickly test typical performance-oriented scenarios.

Some ideas for those scenarios:

*   Loading large HTML data with a lot of attributes.
    *   Coming from an editor (so, only elements and structures that the editor could've produced).
    *   Coming from external sources (e.g. other website), so with structures, elements and attributes that the editor doesn't understand and many inline styles (so we can test the new style normalization).
*   An editor with a long content (same cases as above) but focused on testing features inside the editor (so the content should be already preloaded).
*   Inserting content (like pasting) into an editor.
    *   Also, 3-4 data sets.
    *   The editor content should be diverse as well so you can test various scenarios (pasting into a table, image caption, middle of something, etc.)
*   Loading a couple of editors at once. We expect to see a linear time growth, so we could test just one, but let's have ~3-5 as at those points the load times start to be noticeable.

Other considerations:

*   The action to be tested should be triggered by a button (instead of happening automatically).
*   Use rich setups of editor plugins (so we cover diverse types of those).

Subtasks:

- [x] Test with empty editor and `setData` button (https://github.com/ckeditor/ckeditor5/pull/6103)
- [x] Test with content initialization (after pressing a button) (#6124)
- [ ] Generate content (small, medium, large)
    - [ ] Base text (content that will serve as a reference for other small, medium, large variations) (https://github.com/ckeditor/ckeditor5/pull/6147)
    - [ ] Attributes known to the editor
    - [ ] Attributes unknown to the editor
    - [ ] Inline, combined CSS styles (https://github.com/ckeditor/ckeditor5/pull/6148)
- [x] Editor with many features enabled (content preloaded) (https://github.com/ckeditor/ckeditor5/pull/6149)
- [ ] Loading multiple editors at once (https://github.com/ckeditor/ckeditor5/pull/6144)
- [x] Review performance, check for any potential bottlenecks (https://github.com/ckeditor/ckeditor5/issues/5880#issuecomment-581922363)

zadam:
It's great to see effort on this topic. As an example I'm attaching a HTML file which takes couple of seconds (5-10 on my machine) to load into CKEditor. The file is big-ish, but still something you might expect to work with in the CKEditor.

[BMesh Operators.zip](https://github.com/ckeditor/ckeditor5/files/4014457/BMesh.Operators.zip)

mlewand:
[...]

PR #6103, opened 2020-01-21
Added editor setData manual performance test

mlewand:

Internal: Added editor setData manual performance test. Part of #5880.

---

### Additional information

In this micro PR I added a manual test for calling setData on the editor.

Fixtures (so the data that we consider small/medium/large) are extracted as separate files to keep it DRY. Fun fact, it couldn't be `.html` file as it would not be served by our Karma/Webpack config so I went with `.txt`. Fixtures contain a dummy placeholder as setting it with proper content will be done in another PR.

### DIFF
--- a/tests/manual/performance/setdata.md
+++ b/tests/manual/performance/setdata.md
@@ -1,0 +1,3 @@
+# Performance: `editor.setData()`
+
+1. Use any of "set data" buttons to perform a `editor.setData()` operation.
--- a/tests/manual/performance/setdata.md
+++ b/tests/manual/performance/setdata.md
@@ -2,4 +2,4 @@
 
 1. Begin performance recording in devtools.
 1. Use any button to call `editor.setData()` with chosen data size.
-1. (optionally) Stop performance recording.
+1. Stop performance recording.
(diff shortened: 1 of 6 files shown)


tests/manual/performance/setdata.md
```diff
@@ -0,0 +1,5 @@
+# Performance: `editor.setData()`
+
+1. (optionally) Begin performance recording in devtools.
```
jodator:
```suggestion
1. Begin performance recording in devtools.
```
Is this really optional? I'd remove that.

tests/manual/performance/setdata.md
```diff
@@ -0,0 +1,5 @@
...
+1. Use any button to call `editor.setData()` with chosen data size.
+1. (optionally) Stop performance recording.
```
jodator:
```suggestion
1. Stop performance recording.
```

mlewand:
Ture, the only case when you can gut tell this are these "omg it takes so long that I'm going to take a coffee" kind of cases, which luckily I'm not aware of in our editor :smile:

PR #6124, opened 2020-01-22
Added editor initialization manual performance test

mlewand:

Tests: Added editor setData manual performance test. Part of #5880.

---

### Additional information

File `utils.js` was placed above of manual test directory, as otherwise it was parsed as a manual test and could not be included at all to the test.

### DIFF
--- a/tests/manual/performance/editorinit.html
+++ b/tests/manual/performance/editorinit.html
@@ -1,0 +1,17 @@
+<div id="test-controls">
+	Initialize editor with:
+	<button id="small-content" data-file-name="small" disabled>small</button>
+	<button id="medium-content" data-file-name="medium" disabled>medium</button>
+	<button id="large-content" data-file-name="large" disabled>large</button>
+	HTML payload.
+</div>
+
+<style>
+	.editor {
+		display: none;
+	}
+</style>
+
+<div id="editor_small" class="editor"></div>
+<div id="editor_medium" class="editor"></div>
+<div id="editor_large" class="editor"></div>
--- a/tests/manual/performance/editorinit.html
+++ b/tests/manual/performance/editorinit.html
@@ -1,3 +1,9 @@
+<style>
+	.editor {
+		display: none;
+	}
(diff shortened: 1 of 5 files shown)


tests/manual/performance/editorinit.html
```diff
@@ -0,0 +1,17 @@
+<div id="test-controls">
+	Initialize editor with:
+	<button id="small-content" data-file-name="small" disabled>small</button>
+	<button id="medium-content" data-file-name="medium" disabled>medium</button>
+	<button id="large-content" data-file-name="large" disabled>large</button>
+	HTML payload.
+</div>
+
+<style>
```
jodator: `<style>` could be moved to the top

tests/manual/performance/editorinit.html
```diff
@@ -0,0 +1,17 @@
...
+	.editor {
+		display: none;
+	}
+</style>
+
+<div id="editor_small" class="editor"></div>
```
jodator: The `<hr>` would be nice here as in other manual tests (small issue but would be nice to have same visual pattern for those tests).

tests/manual/performance/editorinit.md
```diff
@@ -0,0 +1,8 @@
+# Performance: editor initialization
+
+1. You might want to enable devtools profiling.
```
jodator: This steps should describe to other guys how to measure performance (in short). So I'd not use `might` or `optional` here - just the steps required to perform the check.

tests/manual/performance/setdata.html
```diff
@@ -0,0 +1,11 @@
+<div id="test-controls">
```
jodator: Those files are mangled from another PR. For atomic PRs we can't create branches from one another (only from `master`). The problem might be shared code but for those quick PRs we should minimize the round-trip time for review.

ps.: I'll see how this will look after merging https://github.com/ckeditor/ckeditor5/pull/6103.

PR #6148, opened 2020-01-27
Performance manual tests: inline css variant of fixtures

mlewand:

Tests: Added inline css variant of fixtures for manual performance tests. See #5880.

---

### Additional information

**Note:** I have added example styling only for one fixture, to make sure this is the right direction before we go with bigger fixtures. This way we'll avoid losing a lot of time in case we're looking for something else.

### DIFF
--- a/tests/_utils/utils.js
+++ b/tests/_utils/utils.js
@@ -3,7 +3,16 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
-/* globals window */
+/* globals window, console */
+
+import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor';
+import ArticlePluginSet from '@ckeditor/ckeditor5-core/tests/_utils/articlepluginset';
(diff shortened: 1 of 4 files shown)


mlewand:
@Reinmar please have a look at the `tests/manual/performance/_utils/small-inline-css.txt` file, let me know if this is what you meant in #5880 

> *   With many inline styles (so we can test the new style normalization).

Reinmar:
No, that's not what I meant. Many means... many :D Like dozens per element. Various forms, some to be picked by the editor. Some not. 

Example – content copied in Chrome from some website. It usually has hundreds of inline styles on all kind of elements. A relatively good case was the markup we used for testing remove formatting, but it was too repetitive (would touch just one/two cases).

Also, this is under "large HTML" so I don't understand why you proposed such a short content.

mlewand:
@Reinmar didn't want to spend too much time on it in case there are any changes requested.

I'll change it in that case.

mlewand:
As mentioned in https://github.com/ckeditor/ckeditor5/pull/6176#issue-370246175 we decided to go with a bigger pull request instead of series of micro prs. This PR is merged there.

PR #6361, opened 2020-02-29
Added all-features manual test

### LINKED ISSUE #6323: All-in manual test

It would be good to have a manual test that includes all the editor's features. During a testing phase, if I want to check a complicated scenario, I always need to create a custom editor first, because we don't provide any demo with all features.

pomek:

Tests: Added a manual test that contains an editor with all features developed by CKEditor 5 team. Closes #6323.


FilipTokarski:
Adding `TextTransformation`, `Autoformat` and `MathType` might be useful too

Reinmar:
As @FilipTokarski mentioned, however, without `MathType` for now due to #6362.

Reinmar:
Actually, `Autoformat` is part of `ArticlePluginSet`. `TextTrnasformation` is not, though.

PR #6384, opened 2020-03-06
i/6324: Add a warning notification in demo pages when user tries to paste a content from MS Word or Google Docs

### LINKED ISSUE #6324: Add a warning to all demos that pasting from Word/GDocs should be tested in a particular demo

The issue: most people expect for the formatting to be maximally preserved when testing pasting from Word.

In most of our demos we enable a minimal number of features (usually none formatting ones). Hence, styling is lost.

We should automatically detect that one is pasting from Word/GDocs and show a nice warning that pasting in this demo may not work as expected and give a link to the Word/GDocs guides (where we have many features enabled and configured).

---

From the technical perspective, the easiest way to plug this behaviour automatically to all editors on a page would be to do something like:

```js
setTimeout( () => {
     const editables = document.queryElementAll( '.ck-content' );
     
    editables.forEach( editable => {
        if ( !editable.ckeditorInstance ) return;

        editable.ckeditorInstance.plugins.get( 'Clipboard' ).on( 'clipboardInput', ... );
     } );
}, 3000 );
```

This code should be included in one of the assets that loaded on **all** our test pages. I presume that Umberto support such things.

panr:

Other: Added a warning notification in demo pages when user tries to paste content from MS Word or Google Docs. Closes #6324.

### DIFF
--- a/docs/assets/snippet.js
+++ b/docs/assets/snippet.js
@@ -3,7 +3,99 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
-/* globals window, document */
+/* globals window, document, setTimeout */
+
+// Show clipboard input notification when user tries to paste a content from MS Word or Google Docs.
+setTimeout( () => {
+	const editables = document.querySelectorAll( '.ck-content' );
+	const googleDocsMatch = /id=("|')docs-internal-guid-[-0-9a-f]+("|')/i;
+	const msWordMatch1 = /<meta\s*name="?generator"?\s*content="?microsoft\s*word\s*\d+"?\/?>/i;
+	const msWordMatch2 = /xmlns:o="urn:schemas-microsoft-com/i;
+
+	editables.forEach( editable => {
+		if ( !editable.ckeditorInstance ) {
+			return;
+		}
+
+		const editor = editable.ckeditorInstance;
+
+		editor.plugins.get( 'Clipboard' ).listenTo( editor.editing.view.document, 'clipboardInput', ( evt, data ) => {
+			const htmlString = data.dataTransfer.getData( 'text/html' );
+			const match = msWordMatch1.test( htmlString ) ||
+				msWordMatch2.test( htmlString ) ||
+				googleDocsMatch.test( htmlString );
+
+			if ( match ) {
+				createClipboardInputNotification();
+			}
+		} );
+	} );
+}, 3000 );
+
+// The notification should contain the links to the demos where user can test rich formatting from.
+function createClipboardInputNotification() {
+	const title = 'Hello!';
+	/* eslint-disable max-len */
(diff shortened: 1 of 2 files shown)


docs/assets/snippet.js
```diff
@@ -3,7 +3,99 @@
...
+		notification.style.top = parseInt( notification.style.top ) + moveOffset + 'px';
+		notification.style.right = parseInt( notification.style.right ) + moveOffset + 'px';
+	}
+
+	// Append notification to the `.main__content` element.
+	const main = document.querySelector( '.main__content' );
+	main.appendChild( notification );
+
+	// A handler for closing a notification.
+	const onClose = () => {
```
Reinmar: Not necessary to be a separate function.

docs/assets/snippet.js
```diff
@@ -3,7 +3,99 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
-/* globals window, document */
+/* globals window, document, setTimeout */
+
+// Show clipboard input notification when user tries to paste a content from MS Word or Google Docs.
+setTimeout( () => {
```
Reinmar: We shouldn't execute this code in snippets from https://github.com/ckeditor/ckeditor5-paste-from-office/tree/master/docs/\_snippets/features. I guess the easiest way to prevent showing the notification on those demos is to register something like `window.preventPasteFromOfficeNotification = true` in those 2 snippets and abort in this `setTimeout()` when this variable is `true`.

docs/assets/snippet.js
panr: 👍

docs/assets/snippet.js
panr: Done here https://github.com/ckeditor/ckeditor5-paste-from-office/pull/95

docs/assets/snippet.js
```diff
@@ -3,7 +3,99 @@
...
+			}
+		} );
+	} );
+}, 3000 );
+
+// The notification should contain the links to the demos where user can test rich formatting from.
+function createClipboardInputNotification() {
+	const title = 'Hello!';
+	/* eslint-disable max-len */
+	const message = '<p>We detected that you tried to paste content from <strong>Microsoft Word</strong> or <strong>Google Docs</strong>.</p><p>Please bear in mind that the editor demo to which you try to paste does not have all the features enabled. Due to that, unsupported formatting is being stripped.</p><p>Check out the <a href="/docs/ckeditor5/latest/features/pasting/paste-from-word.html#demo">Paste from Word</a> or <a href="/docs/ckeditor5/latest/pasting/paste-from-google-docs.html#demo">Paste from Google Docs</a> demos for the best experience.</p>';
```
panr: `href="/docs/ckeditor5/latest/features/pasting/paste-from-word.html#demo"`

Links like this will work...? 🤔

Reinmar: It won't work locally or on nightly docs. But I think it's enough that it works on ckeditor.com. I don't know how we could improve it anyway.

panr: That's why I asked. I was wondering if there is any other solution for that ;-)

Reinmar:
Looks good. Works good. Just two minor things:

*   What do you think about not showing this warning for the second time when pasting twice or more times to a single editor? If you saw it once, there's no need to see it more. And you may want to test pasting into this editor anyway.
*   The close button does not have a hover style. Just sayin' :P

panr:
@Reinmar added, ready for another review.

PR #6728, opened 2020-05-04
I/6115: Table cells selection with keyboard.

### LINKED ISSUE #3203: Handle Shift+arrows inside table cells
While reviewing [selection post-fixer](https://github.com/ckeditor/ckeditor5-engine/pull/1431) I found out that the biggest problem with it is how it fixes selection made by <kbd>Shift</kbd>+<kbd>arrow</kbd> if it started inside a table cell.

The expected result is that once you reach a cell's limit we either:

* Trim the selection to that cell only.
* Select the starting table cell and the one where you moved the caret (which requires https://github.com/ckeditor/ckeditor5-table/issues/63).

I described how we can implement it in https://github.com/ckeditor/ckeditor5-engine/pull/1431#issuecomment-398022824.

### LINKED ISSUE #6115: Table selection with keyboard

It depends on custom navigation in tables: #3267 and #6113.

Adding keyboard navigation would enable us to use <kbd>shift</kbd> with arrows to extend selection over multiple cells.

---

niegowski:

Feature (table): Table cells selection with keyboard. Closes #6115. Closes #3203.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/src/tablenavigation.js
+++ b/packages/ckeditor5-table/src/tablenavigation.js
@@ -7,13 +7,15 @@
  * @module table/tablenavigation
  */
 
-import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
-import { getSelectedTableCells, getTableCellsContainingSelection } from './utils';
-import { findAncestor } from './commands/utils';
+import TableSelection from './tableselection';
 import TableWalker from './tablewalker';
+import { findAncestor } from './commands/utils';
+import { getSelectedTableCells, getTableCellsContainingSelection } from './utils';
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import Rect from '@ckeditor/ckeditor5-utils/src/dom/rect';
-import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
 import priorities from '@ckeditor/ckeditor5-utils/src/priorities';
+import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
 
 /**
  * This plugin enables keyboard navigation for tables.
@@ -27,6 +29,13 @@ export default class TableNavigation extends Plugin {
 	 */
 	static get pluginName() {
 		return 'TableNavigation';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	static get requires() {
+		return [ TableSelection ];
 	}
 
 	/**
@@ -188,9 +197,16 @@ export default class TableNavigation extends Plugin {
 		const selectedCells = getSelectedTableCells( selection );
 
 		if ( selectedCells.length ) {
-			const tableCell = isForward ? selectedCells[ selectedCells.length - 1 ] : selectedCells[ 0 ];
+			if ( expandSelection ) {
+				const tableSelection = this.editor.plugins.get( 'TableSelection' );
+				const focusCell = tableSelection.getFocusCell();
 
-			this._navigateFromCellInDirection( tableCell, direction );
+				this._navigateFromCellInDirection( focusCell, direction, expandSelection );
+			} else {
+				const tableCell = isForward ? selectedCells[ selectedCells.length - 1 ] : selectedCells[ 0 ];
+
+				this._navigateFromCellInDirection( tableCell, direction, expandSelection );
+			}
 
 			return true;
 		}
@@ -206,7 +222,7 @@ export default class TableNavigation extends Plugin {
 
 		// Let's check if the selection is at the beginning/end of the cell.
 		if ( this._isSelectionAtCellEdge( selection, isForward ) ) {
-			this._navigateFromCellInDirection( tableCell, direction );
+			this._navigateFromCellInDirection( tableCell, direction, expandSelection );
 
 			return true;
 		}
@@ -228,7 +244,7 @@ export default class TableNavigation extends Plugin {
 		const textRange = this._findTextRangeFromSelection( cellRange, selection, isForward );
 
 		if ( !textRange ) {
-			this._navigateFromCellInDirection( tableCell, direction );
+			this._navigateFromCellInDirection( tableCell, direction, expandSelection );
 
 			return true;
 		}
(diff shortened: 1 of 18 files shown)


niegowski:
Moved from https://github.com/ckeditor/ckeditor5-table/pull/313.

packages/ckeditor5-table/src/tableselection.js
```diff
@@ -425,39 +459,30 @@ export default class TableSelection extends Plugin {
 		const startColumn = Math.min( startLocation.column, endLocation.column );
 		const endColumn = Math.max( startLocation.column, endLocation.column );
 
-		const cells = [];
+		const selectionMap = new Array( endRow - startRow + 1 ).fill( null ).map( () => [] );
```
jodator, line 462: `fill( [] )` ?

niegowski: That would make the same array instance for all rows.

packages/ckeditor5-table/tests/tableselection.js
```diff
@@ -831,6 +831,120 @@ describe( 'table selection', () => {
 		} );
 	} );
 
+	describe( 'getAnchorCell() and getFocusCell()', () => {
```
jodator, line 834: After refactoring to a getter the proper describe would be:

```javascript
describe( '#anchorCell and #focusCell', ... );
```

jodator: The other things is that those tests can be split so one set of tests for `#anchor` and another for `#focus`. After reading that cases for anchor and focus are tested separately.

But I see them tangled somehow - so if you don't see a reason - don't split ;)

packages/ckeditor5-table/tests/tableselection.js
```diff
@@ -831,6 +831,120 @@ describe( 'table selection', () => {
...
+			tableSelection = editor.plugins.get( TableSelection );
+
+			setModelData( model, modelTable( [
+				[ '[]00', '01', '02' ],
+				[ '10', '11', '12' ],
+				[ '20', '21', '22' ]
+			] ) );
+		} );
+
+		it( 'should not return element if the table cell is not selected', () => {
```
jodator, line 848:
```suggestion
		it( 'should return null if no table cell is selected', () => {
```

packages/ckeditor5-table/src/tablenavigation.js
```diff
@@ -426,18 +442,19 @@ export default class TableNavigation extends Plugin {
 	/**
 	 * Moves the selection from the given table cell in the specified direction.
 	 *
-	 * @private
-	 * @param {module:engine/model/element~Element} tableCell The table cell to start the selection navigation.
+	 * @protected
```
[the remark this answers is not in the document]
niegowski: To make it "officially" testable from automatic tests.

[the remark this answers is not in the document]
jodator: We do not tests private methods :man\_shrugging: so tests should be changed :D 

I'll take a look at those one more time :)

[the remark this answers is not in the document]
jodator: I've missed that in a previous PR - we shouldn't test that. Because we test it already they we must make it @Protected so closing this comment.

packages/ckeditor5-table/src/tableselection.js
```diff
@@ -382,6 +382,40 @@ export default class TableSelection extends Plugin {
 		} );
 	}
 
+	/**
+	 * Returns the focus cell from the current selection.
+	 *
+	 * @returns {module:engine/model/element~Element}
+	 */
+	getFocusCell() {
```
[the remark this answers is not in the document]
niegowski: As discussed f2f this is an util that eases getting anchor and focus cells from current document selection and not a property of the plugin.

packages/ckeditor5-table/src/tablenavigation.js
```diff
@@ -426,18 +442,19 @@ export default class TableNavigation extends Plugin {
...
+	 * @param {module:engine/model/element~Element} focusCell The table cell that is current multi-cell selection focus.
 	 * @param {'left'|'up'|'right'|'down'} direction Direction in which selection should move.
+	 * @param {Boolean} expandSelection If the current selection should be expanded.
 	 */
-	_navigateFromCellInDirection( tableCell, direction ) {
+	_navigateFromCellInDirection( focusCell, direction, expandSelection ) {
```
jodator, line 450: After reading tests: `expandSelection = false` - WDYT?

jodator, line 450: After reading tests: `expandSelection = false` - WDYT?

jodator: FFS GH.. what happens?

packages/ckeditor5-table/src/tablenavigation.js
```diff
@@ -474,20 +491,28 @@ export default class TableNavigation extends Plugin {
...
 		}
 
 		const cellToSelect = tableMap.find( cellInfo => cellInfo.row == row && cellInfo.column == column ).cell;
 		const isForward = [ 'right', 'down' ].includes( direction );
-		const positionToSelect = model.createPositionAt( cellToSelect, isForward ? 0 : 'end' );
 
-		model.change( writer => {
-			writer.setSelection( positionToSelect );
-		} );
+		if ( expandSelection ) {
```
jodator, line 504: test

packages/ckeditor5-table/src/tablenavigation.js
```diff
@@ -474,20 +491,28 @@ export default class TableNavigation extends Plugin {
...
+			const tableSelection = this.editor.plugins.get( 'TableSelection' );
+			const anchorCell = tableSelection.getAnchorCell() || focusCell;
```
jodator, line 506: test 2

packages/ckeditor5-table/src/tableselection.js
```diff
@@ -107,6 +107,58 @@ export default class TableSelection extends Plugin {
 		} );
 	}
 
+	/**
+	 * Sets the model selection based on given anchor and target cells (can be the same cell).
+	 * Takes care of setting the backward flag.
+	 *
+	 * @param {module:engine/model/element~Element} anchorCell
+	 * @param {module:engine/model/element~Element} targetCell
+	 */
```
jodator, line 116: Requires some sample - public API generally comes with it. It doesn't have to be a full-blown code. Just a few lines of how to use it.

jodator:
```suggestion
	 * Sets the model selection based on given anchor and target cells (can be the same cell).
	 * Takes care of setting the backward flag.
	 *
	 *		const modelRoot = editor.model.document.getRoot();
	 *		const firstCell = modelRoot.getNodeByPath( [ 0, 0, 0 ] );
	 *		const lastCell = modelRoot.getNodeByPath( [ 0, 0, 1 ] );
	 *
	 *		const tableSelection = editor.plugins.get( 'TableSelection' );
	 *		tableSelection.setCellSelection( firstCell, lastCell );
	 *
	 * @param {module:engine/model/element~Element} anchorCell
	 * @param {module:engine/model/element~Element} targetCell
	 */
```

packages/ckeditor5-table/src/tablenavigation.js
```diff
@@ -197,17 +197,16 @@ export default class TableNavigation extends Plugin {
 		const selectedCells = getSelectedTableCells( selection );
 
 		if ( selectedCells.length ) {
-			if ( expandSelection ) {
-				const tableSelection = this.editor.plugins.get( 'TableSelection' );
-				const focusCell = tableSelection.getFocusCell();
+			let focusCell = null;
```
jodator, line 200:
```suggestion
			let focusCell;
```

I see that `null` as an overhead. Usually we do not set `null` in similar constructs.

PR #6739, opened 2020-05-05
Docs: Add notes about using AttributeElement's priority and id attributes in mentions.

### LINKED ISSUE #6587: Adding styling to a mention built with MentionCustomization adds the style tags in the wrong place

The bug occurs any time when you use the mention feature with the MentionCustomization function.
You can see the bug in action in the ckeditor docs page on the example [here](https://ckeditor.com/docs/ckeditor5/latest/features/mentions.html#fully-customized-mention-feed):

In that demo,

1. Highlight the @tdog text .
2. Click the Bold button to add the bold style to the text
3. Inspect the element

### ✔️ Expected result

The expected result is that the `<strong>` tags should be added outside  of the `<span>` that creates the mention.
Eg `<strong><span>my mention</span></strong>`
This works correctly when you do the same actions with the demo at the top of the page [here](https://ckeditor.com/docs/ckeditor5/latest/features/mentions.html#demo)

### ❌ Actual result

What actually happens is that the `<strong>` tag is inside the `<span>` tag 
Eg `<span><strong>my mention</strong></span>`

This creates an issue when pulling data back from a database to re-render the mention.  It seems like because the contents of the `<span>` tag don't match up with what is expected, it gets flagged as a partial mention and the mention span is not rendered.

I'm thinking maybe  [here](https://github.com/ckeditor/ckeditor5-mention/blob/4d678d3f95e73b048e711cda942324286d4579b9/src/mentionediting.js#L187) in the source code is where the issue happens.

## 📃 Other details

* Browser: Safari 13.0.5
* OS: MacOS 10.15.3
* CKEditor version: 18.0.0
* Installed CKEditor plugins: 
  EssentialsPlugin,
  UploadAdapterPlugin,
  AutoformatPlugin,
  BoldPlugin,
  ItalicPlugin,
  Underline,
  BlockQuotePlugin,
  EasyImagePlugin,
  HeadingPlugin,
  ImagePlugin,
  ImageCaptionPlugin,
  ImageStylePlugin,
  ImageToolbarPlugin,
  ImageUploadPlugin,
  LinkPlugin,
  ListPlugin,
  ParagraphPlugin,
  Mention,
  MentionCustomization,
  Table,
  TableToolbar,
  TableProperties,
  TableCellProperties,
  Indent,
  IndentBlock,
  RemoveFormat,
  Font,
  Highlight,
  HorizontalLine,
  PasteFromOffice,
  Alignment,
  PageBreak,
  Subscript,
  Superscript,

---

jodator:

Docs (mention): Add notes about using \`AttributeElement\`'s priority and id attribute in mentions. Closes #6587.


Reinmar:
I've got a big problem with using a private `._uid` property there. Shouldn't we also make it public now?

jodator:
I've totally forgot that you already commented here. This was done AFAIR to not mess with users' attributes set on mention object. We could name it somehow distinguishably, like `mentionId`?

Opening this makes sense anyway. WDYT?

Reinmar:
> Opening this makes sense anyway. WDYT?

Yep, definitely. `mentionUid` or `mentionId` sounds good to me.

Reinmar:
I can actually change that when checking the code.

PR #6748, opened 2020-05-06
#6549: Adding a new row in the table copies structure of the selected row.

### LINKED ISSUE #6549: Adding a new row to table with colspans should repeat the structure of the previous row

1. Create a table with horizontally merged cells at the beginning.
2. Start creating new rows.

### ✔️ Expected result

the geometry of the previous row is preserved.

### ❌ Actual result

The geometry of the first row is repeated. If anyone wants to create 10 rows like this, they'll need to do 30 merge operations (happened to me).

---

niegowski:

Other (table): Adding a new row in the table copies the structure of the selected row. Closes #6549.

---

### Additional information


packages/ckeditor5-table/src/tableutils.js
```diff
@@ -132,37 +137,52 @@ export default class TableUtils extends Plugin {
...
+			const columnsSpans = new Array( columns ).fill( 1 );
 
-			for ( const { row, rowspan, colspan, cell } of tableIterator ) {
-				const isBeforeInsertedRow = row < insertAt;
-				const overlapsInsertedRow = row + rowspan > insertAt;
+			for ( const { row, column, rowspan, colspan, cell } of tableIterator ) {
+				const lastCellRow = row + rowspan - 1;
 
-				if ( isBeforeInsertedRow && overlapsInsertedRow ) {
+				if ( row < insertAt && insertAt <= lastCellRow ) {
```
jodator, line 159: AFAICS the logic here wasn't changed. Please do not remove helper `is...` from `if` - they helps to read what that case handles. We may know at this point in time why there is `<` and what we compare here but after a while we might forget.

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -712,7 +732,8 @@ function createEmptyRows( writer, table, insertAt, rows, tableCellToInsert, attr
 // @param {module:engine/model/position~Position} insertPosition
 function createCells( cells, writer, insertPosition, attributes = {} ) {
 	for ( let i = 0; i < cells; i++ ) {
-		createEmptyTableCell( writer, insertPosition, attributes );
+		const cellAttributes = Array.isArray( attributes ) ? attributes[ attributes.length - i - 1 ] : attributes;
```
jodator, line 735: I tried to follow the code for a while and I couldn't found out whit this is in reverse order. I would expect `attributes[ i ]` here.

But I wonder if we couldn't change implementation a bit as it is very complicated now. From the tests we copy table structure only in some edge cases and in most this array is `[ null, null, null ]`. This makes me think that maybe we do too much here and maybe on/off switch would be sufficient.

What I think now is that the default behaviour should be the one described in the ticket. Passing flag (like `useSingleColumns`) could enable the old behavior.

Let's discuss this later on as I'm not sure which way would be better and it may be better to think on it for a while rather re-implement over and over :)

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -132,37 +137,52 @@ export default class TableUtils extends Plugin {
...
 
-				if ( isBeforeInsertedRow && overlapsInsertedRow ) {
+				if ( row < insertAt && insertAt <= lastCellRow ) {
 					// This cell overlaps inserted rows so we need to expand it further.
 					writer.setAttribute( 'rowspan', rowspan + rowsToInsert, cell );
+
+					// Mark this column as spanned from the rows above so no new cell will be needed.
+					columnsSpans[ column ] = -colspan;
+					cellsToInsert -= colspan;
+				} else if ( row <= copyStructureFrom && copyStructureFrom <= lastCellRow ) {
```
jodator, line 166: This gets complicated here.

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -132,37 +137,52 @@ export default class TableUtils extends Plugin {
...
-				if ( row === insertAt ) {
-					cellsToInsert += colspan;
+			const cellsAttributes = [];
+
+			// Prepare array of the new cells' attributes.
+			for ( let i = 0; i < columnsSpans.length; i++ ) {
+				const colspan = columnsSpans[ i ];
+
+				if ( colspan > 0 ) {
+					cellsAttributes.push( colspan > 1 ? { colspan } : null );
```
jodator, line 179: And here :)

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -132,37 +137,52 @@ export default class TableUtils extends Plugin {
...
+				const lastCellRow = row + rowspan - 1;
 
-				if ( isBeforeInsertedRow && overlapsInsertedRow ) {
+				if ( row < insertAt && insertAt <= lastCellRow ) {
 					// This cell overlaps inserted rows so we need to expand it further.
 					writer.setAttribute( 'rowspan', rowspan + rowsToInsert, cell );
+
+					// Mark this column as spanned from the rows above so no new cell will be needed.
+					columnsSpans[ column ] = -colspan;
+					cellsToInsert -= colspan;
```
jodator, line 165: This is also "unexpected".

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -132,37 +137,52 @@ export default class TableUtils extends Plugin {
...
+					columnsSpans[ column ] = colspan;
+					cellsToInsert -= colspan - 1;
 				}
+			}
 
-				// Calculate how many cells to insert based on the width of cells in a row at insert position.
-				// It might be lower then table width as some cells might overlaps inserted row.
-				// In the table above the cell 'a' overlaps inserted row so only two empty cells are need to be created.
-				if ( row === insertAt ) {
-					cellsToInsert += colspan;
```
jodator, line 161: Previously it was additive calculations. Now everything is susbstractive. It looks like it require more calculations.

ATM I'm thinking about simpler switch which would allow to:

```javascript
cellsToInsert += useSingleColumns ? colspan : 1;
```

But I'm just "thinking" aloud here - I'm not sure if this would be the best way because we need to copy some attributes.

packages/ckeditor5-table/src/tableutils.js
niegowski: As discussed f2f: if we implement flag for old behavior it would be a breaking change and still we would need to indicate which row structure we want to copy (since above/below is resolved to `at`command level).

niegowski: The `attributes` were applied in reverse order because the table cells was inserted in reverse order (they all was inserted at the same position). This was changed in [399356e](https://github.com/ckeditor/ckeditor5/pull/6748/commits/399356e33a26c979820d7e950985e311bcf5fbcf).

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -116,12 +116,17 @@ export default class TableUtils extends Plugin {
...
 	 * @param {Number} [options.at=0] The row index at which the rows will be inserted.
 	 * @param {Number} [options.rows=1] The number of rows to insert.
+	 * @param {Number} [options.copyStructureFrom] The row index used as structure reference.
 	 */
 	insertRows( table, options = {} ) {
 		const model = this.editor.model;
 
 		const insertAt = options.at || 0;
 		const rowsToInsert = options.rows || 1;
+		const copyStructureFrom = options.copyStructureFrom !== undefined ? options.copyStructureFrom : -1;
```
jodator, line 126: I would like to make this more digestible. Using `-1` as an indicator for something looks like a smart shortcut. I'm worried here that using magic number is harder to analyse why some comparison are made in such way.

I think that a switch `isCopyStructure = copyStructureFrom !== undefined` would be simpler to follow.

The second thing I think it would be better to have is to revert the additions to `createEmptyRows()` to it previous form. And use calculated spans directly to create empty cells (using `createEmptyCell()` ). This way we shouldn't have to create two arrays for that same information (`columnSpans` and `cellsAttributes`) and can use one of them directly.

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -701,18 +721,20 @@ function createEmptyRows( writer, table, insertAt, rows, tableCellToInsert, attr
 
 		writer.insert( tableRow, table, insertAt );
 
-		createCells( tableCellToInsert, writer, writer.createPositionAt( tableRow, 'end' ), attributes );
+		createCells( tableCellToInsert, writer, tableRow, attributes );
```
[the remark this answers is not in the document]
jodator: This comment can be ignored if you revert the changes in `createEmptyRows` . However, the above fix to logic is welcome as the previous one with inserting always at the `'end'` was a simplification.

PR #6751, opened 2020-05-06
Enabled code linting and dependency checks on the CI and better integration with coveralls

### LINKED ISSUE #6733: Improve collecting coverage reports from packages
I really like how the output on Travis looks now. Great job. However, we still need to improve how we collect what we send to coverage service.

Example build: https://travis-ci.org/github/ckeditor/ckeditor5/builds/682833183

As for now, we're sending a report from each package separately. It means Coveralls will handle only the first report: https://coveralls.io/jobs/6220886. Unfortunately, it does not merge reports if you send more than one.

image

The next packages are ignored. Coveralls responds with the error:

```json
{"message":"service_job_id (682833184) must be unique for Travis Jobs not supplying a Coveralls Repo Token","error":true}
```

Our test tool attaches `karma-coveralls` reporter on CI: https://github.com/ckeditor/ckeditor5-dev/blob/24b718dc8912051dfe3f3bdff498d6642a3be838/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js#L130-L132

We need to remove it.

AFAIR, we merge coverage reports and we want to send a single (merged) report after executing tests for all packages.

For sending the single report, we can use the [`coveralls`](https://www.npmjs.com/package/coveralls) package.

Things to do:
1. Do not attach additional reporters on CI (`ckeditor5-dev`)
2. Find a way how to merge all reports into a single file.
3. Send the merged coverages (as the single file) to Coveralls.

### LINKED ISSUE #6742: Restore additional checks on CI
Before merging to the monorepo, for `ckeditor5-*` packages we had a few checks/tasks:

- JS linter
- CSS linter
- Dependency check
- Save revision

ATM, the repository executes only tests and checks the coverage.

We could consider restoring the scripts listed above:

- Linter – they are obvious,
- Dependency check – when to execute the script? For PR? For push to the master branch? For every commit? I guess we need to check all packages.
- Save revision script is no longer needed.

mlewand:

Internal (ckeditor5): Enabled code linting and dependency checks on the CI. Closes #6742.

Internal (ckeditor5): Combine results from all the separate CC results into one and send combined report to Coveralls service. Closes #6733.

---

### Additional information

As a part of this ticket I ported the bash script to a node script - as bash was getting unmaintenable with adding third check (dependency check).

Also few extra directories had to be excluded from the global linter configuration.


pomek:
Something does not work:

image

Also, I don't see the package name in the label (on the right side):

image

The rest looks good.

mlewand:
@pomek You're right, that's because I attempted to change fold labels to like `package:link` which included a semicolon. That was a mistake, as Travis' fold format uses semicolons as a delimiter, so it was breaking that format.

Fixed in 2651fc4.

scripts/continuous-integration-script.js
```diff
@@ -0,0 +1,152 @@
...
+	childProcess.execSync( 'cp coverage/*/coverage-final.json .nyc_output' );
+
+	const nyc = [ 'nyc', 'check-coverage', '--branches', '100', '--functions', '100', '--lines', '100', '--statements', '100' ];
+	runSubprocess( 'npx', nyc, simplePackageName, 'codeCoverage', 'doesn\'t have required code coverage' );
+
+	travis.foldEnd( foldLabelName );
+}
+
+console.log( 'Uploading combined code coverage report…' );
+childProcess.execSync( 'npx coveralls < .out/combined_lcov.info' );
```
pomek, line 94: One thing I'm missing here is a link to the report on Coveralls:

image

I'm not sure what will do: `coveralls < .out/combined_lcov.info -v` but maybe it's a solution.

mlewand: All right let's give the verbose flag a try with coveralls command, committed in 380e646.

mlewand: Turned out that verbose mode doesn't print anything more.



Reverted in 1560389.

PR #6762, opened 2020-05-07
#3274: Updated table layout post-fixer and upcast converters to properly fix empty rows.

### LINKED ISSUE #3274: Editor crashes with a downcasting error when input contain an empty <tr></tr>
This was fixed if i hard replace the source html by `replace(/<tr[^<]*><\/tr>/gm, '<tr><th></th></tr>')`

niegowski:

Fix (table): Table layout post-fixer and upcast converters should properly handle empty rows . Closes #3274.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/src/converters/table-cell-paragraph-post-fixer.js
+++ b/packages/ckeditor5-table/src/converters/table-cell-paragraph-post-fixer.js
@@ -100,6 +100,8 @@ function fixTableRow( tableRow, writer ) {
 function fixTableCellContent( tableCell, writer ) {
 	// Insert paragraph to an empty table cell.
 	if ( tableCell.childCount == 0 ) {
+		// @if CK_DEBUG_TABLE // console.log( 'Post-fixing table: insert paragraph in empty cell.' );
+
 		writer.insertElement( 'paragraph', tableCell );
 
 		return true;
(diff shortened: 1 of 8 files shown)


jodator:
I think that this fix is too late in the whole progress. We must always have valid model and all algorithms that operates on model expects that and must create valid table.

We have post-fixers for cases that general algorithms on the model could produce invalid table. In most of the cases it is due to collaboration changes. This might happen when applying colliding changes from an external editor. The most obvious case is: user A insert row user B inserts column. Such changes should only be the reason to fix improper model structure.

The table upcast converters should work with invalid HTML - just because not everybody reads the specs and can create something invalid. We will not fix them all and let the table blow-up :man\_shrugging: . But in cases that an invalid HTML breaks the editor should be addressed there.

In other words we should create valid model from an invalid input during upcast conversion if _feasible_.

I think that the provided change is not in line with [a valid table model](https://www.w3.org/TR/html52/tabular-data.html#forming-a-table) and will yield table model error:

> 22\. If there exists a [row](https://www.w3.org/TR/html52/tabular-data.html#row) or [column](https://www.w3.org/TR/html52/tabular-data.html#column) in the table containing only [slots](https://www.w3.org/TR/html52/tabular-data.html#slots) that do not have a [cell](https://www.w3.org/TR/html52/tabular-data.html#cell) anchored to them, then this is a [table model error](https://www.w3.org/TR/html52/tabular-data.html#table-model-error).

I think that we should remove empty row from the model or not allow it to land it in the model in the first place. The fixer for rowspan attributes should cover cases where all cells occupies slots of another row. So that could be another issue.

ps.: I like the idea of "slot" - I didn't probably noticed that before and maybe it would be a better to use this term for "spanned" cells.

jodator:
I've forgot about this tool:  https://validator.w3.org/nu/#textarea.

Example set of errors:

PR #6764, opened 2020-05-07
#6667: Table heading rows should be properly updated after removing rows as a side effect of merging cells

### LINKED ISSUE #6667: Merging cells in header do not update heading rows properly

1.  Open "coffee table" sample.
2.  Merge cells "milk" and two below.

### ✔️ Expected result

The table features works properly (merging cells in body, setting heading rows, etc).

### ❌ Actual result

Table has one heading row (visually) but model have heading rows = 2.

## 📃 Other details

I don't know which command should take care of this (\`TableUtils.removeRow()\`, commands or a post-fixer).

Similar thing happen on `v18.0.0` using merge table cell left/right/up/down buttons:

---

niegowski:

Fix (table): Table heading rows should be properly updated after removing rows as a side effect of merging cells. Closes #6667.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/src/commands/mergecellcommand.js
+++ b/packages/ckeditor5-table/src/commands/mergecellcommand.js
@@ -80,11 +80,15 @@ export default class MergeCellCommand extends Command {
 		const model = this.editor.model;
 		const doc = model.document;
 		const tableCell = getTableCellsContainingSelection( doc.selection )[ 0 ];
-		const cellToMerge = this.value;
-		const direction = this.direction;
 		const table = findAncestor( 'table', tableCell );
 
-		model.change( writer => {
+		const cellToMerge = this.value;
+		const direction = this.direction;
+
+		// Use single batch to modify table in steps but in one undo step.
+		const batch = model.createBatch();
+
+		model.enqueueChange( batch, writer => {
 			const isMergeNext = direction == 'right' || direction == 'down';
 
 			// The merge mechanism is always the same so sort cells to be merged.
@@ -107,7 +111,7 @@ export default class MergeCellCommand extends Command {
 			// Remove empty row after merging.
 			if ( !removedTableCellRow.childCount ) {
 				const tableUtils = this.editor.plugins.get( 'TableUtils' );
-				tableUtils.removeRows( table, { at: table.getChildIndex( removedTableCellRow ) } );
+				tableUtils.removeRows( table, { at: table.getChildIndex( removedTableCellRow ), batch } );
 			}
 		} );
 	}
--- a/packages/ckeditor5-table/src/commands/mergecellcommand.js
+++ b/packages/ckeditor5-table/src/commands/mergecellcommand.js
@@ -109,7 +109,7 @@ export default class MergeCellCommand extends Command {
 			if ( !removedTableCellRow.childCount ) {
 				const tableUtils = this.editor.plugins.get( 'TableUtils' );
 
-				tableUtils.removeRows( table, { at: table.getChildIndex( removedTableCellRow ), batch: writer.batch } );
+				tableUtils.removeRows( table, { at: removedTableCellRow.index, batch: writer.batch } );
 			}
 		} );
 	}
--- a/packages/ckeditor5-table/src/commands/mergecellcommand.js
+++ b/packages/ckeditor5-table/src/commands/mergecellcommand.js
@@ -9,10 +9,7 @@
 
 import Command from '@ckeditor/ckeditor5-core/src/command';
 import TableWalker from '../tablewalker';
-import {
-	updateNumericAttribute,
-	isHeadingColumnCell
-} from './utils';
+import { isHeadingColumnCell, findAncestor } from './utils';
 import { getTableCellsContainingSelection } from '../utils';
 
 /**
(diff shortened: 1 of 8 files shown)


jodator:
I've almost merged it with some changes but I had to revert it by force push (sorry for that).

Unfortunatelly it doesn't play well with undo (current master works):



the broken structure is a bug but the other thing is that number of heading rows isn't restored properly or isn't well handled. I had similar case but i can't reproduce it without that broken structure.

Could you check why this happens in this PR? Maybe a merge with master will be enough. The other hint is to check it how remove rows command handle this.

packages/ckeditor5-table/src/commands/mergecellcommand.js
```diff
@@ -108,7 +107,9 @@ export default class MergeCellCommand extends Command {
 
 			// Remove empty row after merging.
 			if ( !removedTableCellRow.childCount ) {
-				removeEmptyRow( removedTableCellRow, writer );
+				const tableUtils = this.editor.plugins.get( 'TableUtils' );
+
+				tableUtils.removeRows( table, { at: table.getChildIndex( removedTableCellRow ), batch: writer.batch } );
```
jodator, line 112:
```suggestion
				tableUtils.removeRows( table, { at: removedTableCellRow.index, batch: writer.batch } );
```

A bit shorter. Basically the same and

packages/ckeditor5-table/src/commands/mergecellcommand.js
```diff
@@ -83,6 +80,8 @@ export default class MergeCellCommand extends Command {
 		const model = this.editor.model;
 		const doc = model.document;
 		const tableCell = getTableCellsContainingSelection( doc.selection )[ 0 ];
+		const table = findAncestor( 'table', tableCell );
```
jodator, line 83: Can be moved to `if` below - AFAICS it will not be needed in other cases.

packages/ckeditor5-table/src/commands/mergecellscommand.js
```diff
@@ -57,40 +57,25 @@ export default class MergeCellsCommand extends Command {
 			updateNumericAttribute( 'colspan', mergeWidth, firstTableCell, writer );
 			updateNumericAttribute( 'rowspan', mergeHeight, firstTableCell, writer );
 
+			const emptyRows = [];
```
jodator, line 60: Maybe `emptyRowsIndexes` - just to be ultra precise.

packages/ckeditor5-table/src/commands/mergecellscommand.js
```diff
@@ -57,40 +57,25 @@ export default class MergeCellsCommand extends Command {
...
+
 			for ( const tableCell of selectedTableCells ) {
 				const tableRow = tableCell.parent;
+
 				mergeTableCells( tableCell, firstTableCell, writer );
-				removeRowIfEmpty( tableRow, writer );
+
+				if ( !tableRow.childCount ) {
+					emptyRows.push( table.getChildIndex( tableRow ) );
```
jodator, line 68:
```suggestion
					emptyRows.push( tableRow.index );
```

packages/ckeditor5-table/src/commands/mergecellscommand.js
```diff
@@ -57,40 +57,25 @@ export default class MergeCellsCommand extends Command {
...
+				}
 			}
 
+			emptyRows.reverse().forEach( row => tableUtils.removeRows( table, { at: row, batch: writer.batch } ) );
```
jodator, line 72: And since a `table` is only needed here maybe a wrap in `if( emptyRows.length )` ? But IDK for sure - this can be left as is.

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -289,17 +289,21 @@ export default class TableUtils extends Plugin {
 		const last = first + rowsToRemove - 1;
 		const batch = options.batch || 'default';
 
-		// Removing rows from table requires most calculations to be done prior to changing table structure.
+		model.enqueueChange( batch, writer => {
+			// Removing rows from table requires most calculations to be done prior to changing table structure.
+			// Preparations must be done in enqueueChange callback to use the current table structure.
```
jodator, line 294:
```suggestion
			// Removing rows from the table require that most calculations to be done prior to changing table structure.
			// Preparations must be done in the same enqueueChange callback to use the current table structure.
```

Minor wording change.

packages/ckeditor5-table/src/tableutils.js
```diff
@@ -753,17 +757,19 @@ function adjustHeadingColumns( table, removedColumnIndexes, writer ) {
 
 // Calculates a new heading rows value for removing rows from heading section.
 function updateHeadingRows( table, first, last, model, batch ) {
-	const headingRows = table.getAttribute( 'headingRows' ) || 0;
+	// Must be done after the changes in table structure (removing rows).
+	// Otherwise the downcast converter for headingRows attribute will fail. ckeditor/ckeditor5#6391.
+	//
```
jodator, line 762:
```suggestion
	// Otherwise the downcast converter for headingRows attribute will fail.
	// See https://github.com/ckeditor/ckeditor5/issues/6391.
	//
```

Issue #6790, opened 2020-05-11
Unify color definition type among components.


tomalec:
## 📝 Provide a description of the improvement

Currently, the number of components uses a color definition object. Unfortunately, there are two slightly different types, which introduces confusion and cognitive burden for developers, requires more documentation, and makes code complicated and less performance (due to translations needed).

#### TableColorConfig

*   [docs](https://ckeditor.com/docs/ckeditor5/latest/api/module_table_table-TableColorConfig.html)
*   [source](https://github.com/ckeditor/ckeditor5-table/blob/016175e1f49d2340420ad38c289cc8a1c7b3d56a/src/table.js#L75-L101)

   `Array.<String | Object>`

   ```js
   const colors = [
	   {
	       color: 'hsl(0, 0%, 60%)',
	       label: 'Grey'
	   },
	   'hsl(0, 0%, 80%)',
	   {
	       color: 'hsl(0, 0%, 90%)',
	       label: 'Light grey'
	   },
	   {
	       color: 'hsl(0, 0%, 100%)',
	       label: 'White',
	       hasBorder: true
	   },
	   '#FF0000'
   ]
   ```

#### ColorDefinition

*   [docs](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_colorgrid_colorgrid-ColorDefinition.html)
*   [source](https://github.com/ckeditor/ckeditor5-ui/blob/d2a66f096ed98c94b0356a14406cda6319f29121/src/colorgrid/colorgridview.js#L180-L200)

   `Object`
   
   ```js
   {
       color: 'hsl(0, 0%, 75%)',
	   label: 'Light Grey',
	   options: {
	       hasBorder: true
	   }
   }
   ```

#### undocumented, unnamed structure - let's call it ColorValue - for the sake of this post.

*   [source](https://github.com/ckeditor/ckeditor5-ui/blob/d2a66f096ed98c94b0356a14406cda6319f29121/src/colorgrid/colorgridview.js#L110-L112)

   `Object`

   ```js
   {
	   value: 'hsl(0, 0%, 75%)',
	   label: 'Light Grey',
	   hasBorder: true
   }
   ```

As you can see they are similar, but slightly off. I don't see any particular reason why is it so.

### Who uses what?

... that was the tricky part to me, but AFAI have found:

[...]

PR #6801, opened 2020-05-12
#6791: Stop normalizing color values passed to the UI, relax matching criteria for ColorInputView's text input.

### LINKED ISSUE #6791: `normalizeColorOptions` in TablePropertiesUI breaks UX of ColorInputData

This issue was discovered during the review of https://github.com/ckeditor/ckeditor5/pull/6784

1. Open table properties (or table cell properties).
2. Set border color and background color to any defined value, for example, "Dim grey".
3. Close the toolbar.
4. Reset the state by typing `editor.setData(editor.getData())` in the browser console.
5. Open table (cell) properties once again.

Animated demo of steps above

### ✔️ Expected result

"Dim grey" should be shown in both input boxes.

### ❌ Actual result

Border color input shows actual color value: "hsl(0, 0%, 30%)".

## 📃 Other details

After a brief investigation with @Reinmar we discovered that removing `.replace( / /g, '' )`
https://github.com/ckeditor/ckeditor5/blob/3c7c28d3cff978b9efe6e0f748a742939ea22481/packages/ckeditor5-ui/src/colorgrid/utils.js#L73-L85
Solves the problem. 
There is a mismatch between normalized and non-normalized values set by the element, and set by the state restore.
We were not sure why this normalization was there, and what side effect we might introduce by removing it. However, it seems to me that we should use, preserve, and accept values in exactly the same shape as they were given in the initial config.

* Browser: Chrome
* OS: Win10
* CKEditor version: master ([@fbec6b2af](https://github.com/ckeditor/ckeditor5/tree/fbec6b2af7a8a45c189388b537ed48d223b9f18a))
* Installed CKEditor plugins: table

---

tomalec:

Fix (table): When the state is restored or the user enters color value manually, color input is able to provide matching color label. Closes #6791.

---

### Additional information

I decided to use more complicated `RegExp` instead of just stripping spaces, I hope it will not affect performance. The main reason was to support, (valid) color definitions, like `rgb(25 50 0)` and not to match them with `rgb(255 0 0)`. (this case is covered in tests)

By the way, I discovered that normalization util uses yet another object type - reported at https://github.com/ckeditor/ckeditor5/issues/6790#issuecomment-627412557/. It's a different type, but function docs refer `ColorDefinition` even though it does not match. It hurts my OCD, but I left it as it was, to minimize the set of changes, and leave the decision up to #6790.

### DIFF
--- a/packages/ckeditor5-table/src/ui/colorinputview.js
+++ b/packages/ckeditor5-table/src/ui/colorinputview.js
@@ -294,8 +294,9 @@ export default class ColorInputView extends View {
 	 */
 	_setInputValue( inputValue ) {
 		if ( !this._stillTyping ) {
+			const normalizedInputValue = normalizeColor( inputValue );
 			// Check if the value matches one of our defined colors.
-			const mappedColor = this.options.colorDefinitions.find( def => inputValue === def.color );
+			const mappedColor = this.options.colorDefinitions.find( def => normalizedInputValue === normalizeColor( def.color ) );
 
 			if ( mappedColor ) {
 				this._inputView.value = mappedColor.label;
@@ -304,4 +305,17 @@ export default class ColorInputView extends View {
 			}
 		}
 	}
+}
+
+// Normalizes color value, by stripping extensive whitespace.
+// For example., transforms:
+// * `   rgb(  25 50    0 )` to `rgb(25 50 0)`,
+// * "\t  rgb(  25 ,  50,0 )		" to `rgb(25 50 0)`.
+//
+// @param {String} colorString The value to be normalized.
+// @returns {String}
+function normalizeColor( colorString ) {
+	// Remove any whitespace at the beginning, after `(`, or `,`, at the end, before `)`, `,`, or another whitespace.
+	// Then, replace `,` or whitespace with single space.
+	return colorString.replace( /(?<=^|\(|,)\s*|\s*(?=$|\)|\s|,)|/g, '' ).replace( /,|\s/g, ' ' );
 }
--- a/packages/ckeditor5-table/src/ui/colorinputview.js
+++ b/packages/ckeditor5-table/src/ui/colorinputview.js
@@ -315,7 +315,10 @@ export default class ColorInputView extends View {
 // @param {String} colorString The value to be normalized.
 // @returns {String}
 function normalizeColor( colorString ) {
-	// Remove any whitespace at the beginning, after `(`, or `,`, at the end, before `)`, `,`, or another whitespace.
+	// Remove any whitespace right after `(` or `,`.
+	// Remove any whitespace at the beginning or right before the end, `)`, `,`, or another whitespace.
 	// Then, replace `,` or whitespace with single space.
-	return colorString.replace( /(?<=^|\(|,)\s*|\s*(?=$|\)|\s|,)|/g, '' ).replace( /,|\s/g, ' ' );
+	return colorString.replace( /([(,])\s+/g, '$1' )
+		.replace( /^\s+|\s+(?=[),\s]|$)/g, '' )
+		.replace( /,|\s/g, ' ' );
 }
--- a/packages/ckeditor5-table/src/ui/colorinputview.js
+++ b/packages/ckeditor5-table/src/ui/colorinputview.js
@@ -315,10 +315,11 @@ export default class ColorInputView extends View {
 // @param {String} colorString The value to be normalized.
 // @returns {String}
 function normalizeColor( colorString ) {
-	// Remove any whitespace right after `(` or `,`.
-	// Remove any whitespace at the beginning or right before the end, `)`, `,`, or another whitespace.
-	// Then, replace `,` or whitespace with single space.
-	return colorString.replace( /([(,])\s+/g, '$1' )
+	return colorString
+		// Remove any whitespace right after `(` or `,`.
+		.replace( /([(,])\s+/g, '$1' )
+		// Remove any whitespace at the beginning or right before the end, `)`, `,`, or another whitespace.
 		.replace( /^\s+|\s+(?=[),\s]|$)/g, '' )
+		// Then, replace `,` or whitespace with a single space.
 		.replace( /,|\s/g, ' ' );
 }
--- a/packages/ckeditor5-table/tests/ui/colorinputview.js
+++ b/packages/ckeditor5-table/tests/ui/colorinputview.js
@@ -218,6 +218,23 @@ describe( 'ColorInputView', () => {
 				expect( inputView.value ).to.equal( 'bar' );
 			} );
 
+			it( `when the color input value is set to one of defined colors, but with few additional white spaces,
+			should use its label as the text input value`, () => {
+				view.value = 'rgb(0,    255, 0)';
+				expect( inputView.value ).to.equal( 'Green' );
+
+				view.value = '   rgb( 255 0  0)    ';
+				expect( inputView.value ).to.equal( 'Red' );
+
+				view.value = ' 		  rgb(0,  0,  255 )';
+				expect( inputView.value ).to.equal( 'Blue' );
+
+				// Blindly stripping spaces may not work.
+				// rgb(25 50 0) != rgb(255 0 0)
+				view.value = ' 		  rgb(25 50  0)';
+				expect( inputView.value ).to.equal( ' 		  rgb(25 50  0)' );
+			} );
+
 			it( `when the color input value is set to one of defined colors,
 			should use its label as the text input value`, () => {
 				view.value = 'rgb(0,255,0)';
--- a/packages/ckeditor5-table/tests/ui/colorinputview.js
+++ b/packages/ckeditor5-table/tests/ui/colorinputview.js
@@ -218,22 +218,25 @@ describe( 'ColorInputView', () => {
 				expect( inputView.value ).to.equal( 'bar' );
 			} );
 
-			it( `when the color input value is set to one of defined colors, but with few additional white spaces,
-			should use its label as the text input value`, () => {
-				view.value = 'rgb(0,    255, 0)';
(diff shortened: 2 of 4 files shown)


packages/ckeditor5-table/src/ui/colorinputview.js
```diff
@@ -305,3 +306,16 @@ export default class ColorInputView extends View {
...
+// For example., transforms:
+// * `   rgb(  25 50    0 )` to `rgb(25 50 0)`,
+// * "\t  rgb(  25 ,  50,0 )		" to `rgb(25 50 0)`.
+//
+// @param {String} colorString The value to be normalized.
+// @returns {String}
+function normalizeColor( colorString ) {
+	// Remove any whitespace at the beginning, after `(`, or `,`, at the end, before `)`, `,`, or another whitespace.
+	// Then, replace `,` or whitespace with single space.
+	return colorString.replace( /(?<=^|\(|,)\s*|\s*(?=$|\)|\s|,)|/g, '' ).replace( /,|\s/g, ' ' );
```
jodator, line 320: Unfortunatelly, fancy ECMA2018 RegExp additions are not supported on Firefox. It blows the editor up.

jodator: We could either have a simpler RegExp for FF or have totally new different RegExp.

packages/ckeditor5-table/tests/ui/colorinputview.js
```diff
@@ -218,6 +218,23 @@ describe( 'ColorInputView', () => {
 				expect( inputView.value ).to.equal( 'bar' );
 			} );
 
+			it( `when the color input value is set to one of defined colors, but with few additional white spaces,
+			should use its label as the text input value`, () => {
```
jodator, line 222: I'm not sure if I like that notation. Could you try to make it fit?

packages/ckeditor5-ui/src/colorgrid/utils.js
```diff
@@ -64,13 +64,16 @@ export function normalizeColorOptions( options ) {
 }
 
 // Creates a normalized color definition from the user-defined configuration.
+// The "normalization" means it will create full
+// {@link module:ui/colorgrid/colorgrid~ColorDefinition `ColorDefinition-like`}
+// object for string values, and add a `view` property, for each definition.
 //
 // @param {String|module:ui/colorgrid/colorgrid~ColorDefinition}
 // @returns {module:ui/colorgrid/colorgrid~ColorDefinition}
 export function normalizeSingleColorDefinition( color ) {
```
jodator, line 73: No tests for this change. (looks like should land in `normalizeColorOptions()` tests.

Sorry :)

packages/ckeditor5-table/src/ui/colorinputview.js
jodator: ps.: This RegExp looks very complicated - couldn't we use something simpler (two-step change, like congesting multiple whitespaces to one space and then replacing quirks?

:point\_up: just a random thought as I didn't dig into the RegExp istelf as it cause FF to not load the editor in the first place.

jodator: If you go with two RegExps, here's a similar case: https://github.com/ckeditor/ckeditor5/blob/dca2d0f620e0fc2895ebb56f57b5d00950ea55ef/packages/ckeditor5-mention/src/mentionui.js#L639-L655.

But I'd check if we could simplify that.

tomalec: It's the lack of support for lookbehind https://bugzilla.mozilla.org/show_bug.cgi?id=1225665 in FF. I'd do separate `.replace` for this assertion.

tomalec: >  couldn't we use something simpler (two-step change, like congesting multiple whitespaces to one space and then replacing quirks

That's exactly, what I do, see the code comment above.
https://github.com/ckeditor/ckeditor5/pull/6801/files/4ea684d2c61271b4d8a8f6f85239a7b6d390bb1c#diff-05504865ae0c3ced8720bd0b06af5c54R318-R319

tomalec: Let me know if it get clearer after https://github.com/ckeditor/ckeditor5/pull/6801/commits/d3513944db643ece09a531abafe479fb38a3a16e

packages/ckeditor5-table/tests/ui/colorinputview.js
tomalec: I hope that's enough https://github.com/ckeditor/ckeditor5/pull/6801/commits/fc82d5cec73a5d1456fc3b742ee05423d0fadc09
I don't think, I can rephrase the text to make it fit.

This is the styling we agreed on with @Reinmar in the previous review. We both agreed it's not perfect, but acceptable.
If you have any better ideas, please let me know.

packages/ckeditor5-ui/src/colorgrid/utils.js
tomalec: Do you mean, there is no test for what I added to the description? Or the fact it does not replace spaces any longer?

I think https://github.com/ckeditor/ckeditor5/pull/6801/files/4ea684d2c61271b4d8a8f6f85239a7b6d390bb1c#diff-8075f08ce8d2dc6e3b560e6c0adce83fR307 tests it

tomalec:
@jodator All comments addressed or commented. PR is ready for a second look :)

packages/ckeditor5-ui/src/colorgrid/utils.js
jodator: The change is `ckeditor5-ui` and you test it in `ckeditor5-table` . The base test must be in `ckeditor5-ui`. Additional tests in `ckeditor5-table` are OK.

packages/ckeditor5-table/tests/ui/colorinputview.js
```diff
@@ -218,6 +218,26 @@ describe( 'ColorInputView', () => {
 				expect( inputView.value ).to.equal( 'bar' );
 			} );
 
+			it(
+				`when the color input value is set to one of defined colors, but with few additional white spaces,
+				should use its label as the text input value`,
```
jodator, line 223:
```suggestion
			it( 'should use its label as the text input value (normalizing color input value to match defined colors)', () => {
```

jodator: I was thinking about something like that. Shorter description. But if you guys already agreed on that I'm :man\_shrugging: on this :)

packages/ckeditor5-table/tests/ui/colorinputview.js
jodator: Just to be precise :) https://github.com/ckeditor/ckeditor5/pull/6801/files#r425600735

The test description is awfully long for me. It's my proposition to shorten this, but I'm not insisting on that form. The fc82d5c is a bit worse IMO here. The trick with template string was OK.

packages/ckeditor5-table/tests/ui/colorinputview.js
jodator: OK... I saw PK was writing that long descriptions already. For me those are too long but as above :man\_shrugging: . Let's keep that tests consistent.

packages/ckeditor5-table/tests/ui/colorinputview.js
jodator: OK I've read rest of the test. As in other comment - leave it as is.

packages/ckeditor5-table/src/ui/colorinputview.js
```diff
@@ -305,3 +306,19 @@ export default class ColorInputView extends View {
...
+//
+// @param {String} colorString The value to be normalized.
+// @returns {String}
+function normalizeColor( colorString ) {
+	// Remove any whitespace right after `(` or `,`.
+	// Remove any whitespace at the beginning or right before the end, `)`, `,`, or another whitespace.
+	// Then, replace `,` or whitespace with single space.
+	return colorString.replace( /([(,])\s+/g, '$1' )
+		.replace( /^\s+|\s+(?=[),\s]|$)/g, '' )
+		.replace( /,|\s/g, ' ' );
```
jodator, line 323:
```suggestion
function normalizeColor( colorString ) {
	return colorString
		// Remove any whitespace right after `(` or `,`.
		.replace( /([(,])\s+/g, '$1' )
		// Remove any whitespace at the beginning or right before the end, `)`, `,`, or another whitespace.
		.replace( /^\s+|\s+(?=[),\s]|$)/g, '' )
		// Then, replace `,` or whitespace with single space.
		.replace( /,|\s/g, ' ' );
```

@tomalec Yep this looks OK now :) However I would move comments closer to code - just to have it on the spot. WDYT?

packages/ckeditor5-ui/src/colorgrid/utils.js
jodator: OK, my bad again. It is OK.

tomalec: Thx, speaking of tests:

After the change 
https://github.com/ckeditor/ckeditor5/pull/6801/files/4ea684d2c61271b4d8a8f6f85239a7b6d390bb1c#diff-8075f08ce8d2dc6e3b560e6c0adce83fR307 

the test below

https://github.com/ckeditor/ckeditor5/blob/4ea684d2c61271b4d8a8f6f85239a7b6d390bb1c/packages/ckeditor5-font/tests/ui/colorui.js#L321-L327

Checks that `'rgb(255, 255, 255)'` is "transformed" to `'rgb(255, 255, 255)'` (no change).
Previously it checked that `'rgb(255, 255, 255)'` is transformed to `'rgb(255,255,255)'` (removed spaces).

Issue #5328, opened 2017-03-31
Balloon panel sticks out of the limiter element while scrolling


oskarwrobel:
Follow-up of: ckeditor/ckeditor5#5320 

![mar-31-2017 12-37-33](https://cloud.githubusercontent.com/assets/1057056/24547209/ec6978b0-160e-11e7-98d8-dec905b55dbc.gif)

oleq:
The issue also appears when h–scrolling some wide content

![image](https://user-images.githubusercontent.com/1099479/66916528-ae443680-f01b-11e9-8ac7-ef259904d0da.png)

dungkvy:
I still got this bug. How about the solution that we can hide ck-ballon-panel when scroll?

<img width="776" alt="Screen Shot 2020-02-17 at 10 02 13 PM" src="https://user-images.githubusercontent.com/54826652/74665068-82a3a980-51d1-11ea-8b59-ba8b9cbb4b41.png">

jswiderski:
![](https://user-images.githubusercontent.com/5314687/142996605-6488c879-0ce3-4172-a61c-5d13414d5f1d.gif)

albertfdp:
👍 We're seeing this issue at Zendesk too for image resize toolbar

lslowikowska:
Hey @albertfdp, thank you for your comment. Would you get in touch with us at support@cksouce.com? We might have more questions for you regarding this issue. Hope to hear back from you soon 👋

Dumluregn:
Let's start with creating a PoC according to @oleq 's guidelines that will appear below soon.

oleq:
I think there's an alternative solution to what I proposed in https://github.com/ckeditor/ckeditor5/issues/5328#issuecomment-294888313.

* We could replicate the powered by logo logic to **some** extent https://github.com/ckeditor/ckeditor5/blob/01ea41e10e296ba2c64afd896f1168ba4e7e2e8c/packages/ckeditor5-ui/src/editorui/poweredby.ts#L321-L327
* The idea is not to hide the element if it becomes "unpositionable" but position it far away from the viewport.
   * 👍 The element retains focus, so we don't have to worry about focus management. The change is also transparent to screen reader users.
   * 👎 In theory, the user could do something by accident because the positioned UI is still there and have focus, so typing in it or pressing enter could execute some actions.
[...]

PR #6810, opened 2020-05-13
#6190: Balloons attached to widgets should always be visible even if the widget is longer than a visible viewport

### LINKED ISSUE #6190: There's no way to access table properties when table longer than the viewport

A follow-up of https://github.com/ckeditor/ckeditor5/issues/6049 and https://github.com/ckeditor/ckeditor5/issues/6112.

1. Create a table longer than the viewport.
2. Try to open the table (not cell) properties form.

### ✔️ Expected result

It should be visible somewhere and accessible.

### ❌ Actual result

It is pinned either to the top or to the bottom, way beyond the visible viewport and user's reach.

## 📃 Other details

image

---

oleq:

Fix (widget): The widget toolbar should always be visible even if the widget is longer than a visible viewport (see #6190).

Fix (table): The table properties balloon should always be visible if the table is longer than a visible viewport. Closes #6190.

---

### Additional information

*   I'm not sure about the name of `centeredBalloonPositionForLongWidgets()` but there's so much going on in there I'm not sure there's an easy way to express that in simple words.
*   At first I wanted `centeredBalloonPositionForLongWidgets()` to be a part of `BalloonPanelView#defaultPositions` but I figured (so far) this is a widget-specific behavior so maybe let's keep it in `ckeditor5-widget` for a while.


oleq:
@ckeditor/qa-team Can you take a look at this? You can test both scenarios in http://fake.ckeditor.com:8125/ckeditor5-table/tests/manual/tableproperties.html using a very long table (or a very small viewport).

FilipTokarski:
Desktop browsers work fine, but on Android the balloon appears halfway when scrolling up:
14_android_1

packages/ckeditor5-widget/src/utils.js
```diff
@@ -339,6 +342,75 @@ export function viewToModelPositionOutsideModelElement( model, viewElementMatche
...
+		};
+	}
+	// Cases (A) and (B).
+	else {
+		return {
+			top: Math.max( widgetRect.top, 0 ) + BalloonPanelView.arrowVerticalOffset,
+			left,
+			name: 'arrow_n'
+		};
+	}
```
niegowski, line 411: I would simplify this function only to cases A and B. C seems to appear only in very "short" scenario and makes balloon jump on the bottom of the widget.
```suggestion
	return {
		top: Math.max( widgetRect.top, 0 ) + BalloonPanelView.arrowVerticalOffset,
		left,
		name: 'arrow_n'
	};
```

packages/ckeditor5-widget/tests/utils.js
```diff
@@ -489,4 +493,123 @@ describe( 'widget utils', () => {
...
+			width: 10,
+			height: 10
+		} );
+
+		beforeEach( () => {
+			testUtils.sinon.stub( global.window, 'innerWidth' ).value( 100 );
+			testUtils.sinon.stub( global.window, 'innerHeight' ).value( 100 );
+		} );
+
+		it( 'should position the balloon inside a widget – at the top + in the middla', () => {
```
niegowski, line 515:
```suggestion
		it( 'should position the balloon inside a widget – at the top + in the middle', () => {
```

packages/ckeditor5-widget/src/utils.js
oleq: I also had some doubts about it but I figured it feels sort of OK when the balloon is attached to the bottom and there's not enough space to keep it there. This way we avoid a rapid change and provide this one extra step `[ underneath, inside at the bottom, inside at the top]` instead of `[underneath, inside at the top]`. 

Anyway, a funny idea popped into my head when I was thinking about it. What if... the position detects which direction the window is scrolling? Like as long as scrolling down (the upper edge of widget disappears) we would attach it to the top of the viewport and to the bottom of the viewport if scrolling in to opposite direction. 

WDYT?

niegowski: As we talked F2F, for this edge scenario there is no point in complicating it.

oleq:
> Desktop browsers & iOS work fine, but on Android the balloon appears halfway when scrolling up:

I guess I'm ready to live with that.

niegowski:
I'm wondering about the scenario where editor is limited in height with `overflow: scroll`. If editor would be in the middle of window, then this balloons would be displayed on the edge of the window (could be far away from editor), maybe we should try to stick it to closest scrollable instead of window?

oleq:
> I'm wondering about the scenario where editor is limited in height with `overflow: scroll`. If editor would be in the middle of window, then this balloons would be displayed on the edge of the window (could be far away from editor), maybe we should try to stick it to closest scrollable instead of window?

That's a bigger topic because it touches all balloons, not only widget balloons. See #5328.

PR #6816, opened 2020-05-14
#6621: Select all in parent limit elements for successive select-all commands.

### LINKED ISSUE #6621: The Select all feature could expand selection with successive Ctrl+A keystrokes

The idea is as follows:

1. Consider the following structure root->table->cell->image->caption.
2. The selection is collapsed in a caption.
3. Ctrl+A to select the image caption.
4. Ctrl+A to select the image.
5. Ctrl+A to select the cell content (if the image is not already its only content).
6. Ctrl+A to select the entire table.
7. Ctrl+A to select the entire root.

Alternatively, the scenario could work without 4. and 6. because it feels like a widget responsibility.

The idea was first described in the [original PR](https://github.com/ckeditor/ckeditor5-select-all/pull/1#issuecomment-614520203).

---

tomalec:

Other: Successive executing select-all in nested editables, selects all in parent ones. Closes #6621.

---

### Additional information

The same effect occurs when the user manually selects the entire content of an editable.


tomalec:
I'm confused with Travis's result. It complains about code coverage, but Coveralls is happy and I don't see which line/case is not covered by tests.

tomalec:
CC issue addressed, by https://github.com/ckeditor/ckeditor5/pull/6816/commits/f00e28284498e545c82082781ce9ffa7dfb23a4d
with an assumption that root element is a limit element.

packages/ckeditor5-select-all/src/selectallcommand.js
```diff
@@ -30,10 +31,20 @@ export default class SelectAllCommand extends Command {
 	 */
 	execute() {
 		const model = this.editor.model;
-		const limitElement = model.schema.getLimitElement( model.document.selection );
+		const selection = model.document.selection;
+		let limitElement = model.schema.getLimitElement( selection );
+		let place = 'in';
+		// If entire element was already selected, try selecting all in a parent limit element (if any).
```
Reinmar, line 37:
```suggestion
	
			// If entire element was already selected, try selecting all in a parent limit element (if any).
```

packages/ckeditor5-select-all/src/selectallcommand.js
```diff
@@ -30,10 +31,20 @@ export default class SelectAllCommand extends Command {
...
+		if ( selection.containsEntireContent( limitElement ) && limitElement.root !== limitElement ) {
```
Reinmar, line 38: Blank lines before `if()` and after `while()` are missing.

We need ESLint rule for that :D

packages/ckeditor5-select-all/src/selectallcommand.js
```diff
@@ -21,6 +21,7 @@ import Command from '@ckeditor/ckeditor5-core/src/command';
  *
  * If the selection was anchored in a {@glink framework/guides/tutorials/implementing-a-block-widget nested editable}
  * (e.g. a caption of an image), the new selection will contain its entire content.
+ * Successive execution - selecting all if entire content is selected - expands selection to the parent editable.
```
niegowski, line 24: This is not exactly how this works. It expands selection to closest limit element and if it's content is entirely selected then tries to select closest object element: caption content -> image object -> table cell content -> table object -> root content

packages/ckeditor5-select-all/src/selectallcommand.js
```diff
@@ -30,10 +31,20 @@ export default class SelectAllCommand extends Command {
...
+				limitElement = limitElement.parent;
+			} while ( !model.schema.isLimit( limitElement ) );
+			place = 'on';
+		}
 
 		model.change( writer => {
-			writer.setSelection( limitElement, 'in' );
+			writer.setSelection( limitElement, place );
 		} );
 	}
```
niegowski, line 49: I would propose sth like this:

```javascript
	execute() {

		model.change( writer => {
			writer.setSelection( getClosestSelection( model, selection ) );
		} );
	}
}

function getClosestSelection( model, selection ) {
	const limitElement = model.schema.getLimitElement( selection );

	// If the entire content of the closest limit element is already selected, try selecting closest object.
	if ( selection.containsEntireContent( limitElement ) ) {
		const objectElement = getObjectElement( model.schema, limitElement );

		if ( objectElement ) {
			return model.createSelection( objectElement, 'on' );
		}
	}

	return model.createSelection( limitElement, 'in' );
}

function getObjectElement( schema, element ) {
	do {
		if ( element.parent ) {
			element = element.parent;
		} else {
			return null;
		}
	} while ( !schema.isObject( element ) );

	return element;
}
```

niegowski: Note:

*   We can use `selection` object, no need for `place` variable;
*   It looks for closest `isObject` not `isLimit`;
*   `getObjectElement` is extracted to helper function to make code more readable;
*   We can check if element is a `root` if it has no `parent`.

packages/ckeditor5-select-all/tests/selectallcommand.js
```diff
@@ -70,13 +70,21 @@ describe( 'SelectAllCommand', () => {
...
 		it( 'should select all (selection in a nested editable)', () => {
-			setData( model, '<paragraph>foo</paragraph><image src="foo.png"><caption>[bar]</caption></image>' );
+			setData( model, '<paragraph>foo</paragraph><image src="foo.png"><caption>b[ar]</caption></image>' );
 
 			editor.execute( 'selectAll' );
 
 			expect( getData( model ) ).to.equal( '<paragraph>foo</paragraph><image src="foo.png"><caption>[bar]</caption></image>' );
 		} );
 
+		it( 'when entire editable is selected, should select all in parent limit element', () => {
```
niegowski, line 80: This is for selecting whole widget/object.

packages/ckeditor5-select-all/tests/selectallcommand.js
```diff
@@ -103,5 +111,59 @@ describe( 'SelectAllCommand', () => {
 				'</table>'
 			);
 		} );
+
+		it( 'consecutive execute() on nested editable, should select all in the parent limit element', () => {
```
niegowski, line 115: I would split this to the separate tests to make sure that the next `execute` is expanding selection based on the current selection, and not on some internal state.

packages/ckeditor5-select-all/src/selectallcommand.js
oleq: During a F2F talk with @niegowski we noticed some details that should help implement things even better:

*   I agree that this `do {} while()` loop and `place` variable tandem is hard to read, so let's split things a little bit.
*   Instead of `model.createSelection()`, let's use `writer.createRange()`. Unlike selections, ranges are lightweight.
*   `Schema#isLimit` is a tricky concept because some model elements like `tableRow` are also limits and, for sure, the select all feature should not select all cells in a row. I think (but didn't debug that) the proposed implementation works by accident because the loop starts from `limitElement.parent`. This may backfire in the future if we have some deeper structure of limit elements.  
      
    I think what we need is something smarter than just a limit check. It could be we need to find the closest ancestor that allows text/paragraph, so more likely we need to exclude elements that fail this check `schema.checkChild( limitAncestor, '$text' ) || schema.checkChild( limitAncestor, 'paragraph' )`.  
      
    OTOH this still sounds a little weak because in theory there could be limits that allow widgets only, WDYT @Reinmar?

Reinmar: Tip: There's `model.createRange*()` too, so you don't have to access the writer to create a range.

Good catch with the limit thing. I didn't realise that before and it's indeed tricky. I can't think of a better check than this:

*   it's a limit element
*   and it can contain a text or paragraph

It's not perfect but it should work in enough cases to be accepted.

oleq: Created a follow-up #6829.

tomalec: Thanks! Reading the current docs, I understood that "limit" is exactly the "closest editable ancestor", like "limit of editing scope", which led to this implementation. I naively thought that "widget" could be editable and have something meaningful to select-all as well (because why not).

tomalec:
Thanks for the deep review and lots of explanations.

-----

> This implementation selects the parent widget if the selection was in the nested editable. 

This is a result of my misunderstanding of what "limit" is. I thought it is "a scope/limit of editing"= editable element. Then, if `selection` `containsEntireContent` of an editable element, I have to start from parent element, if I'm looking for its editable ancestors.

-----

Could you explain what you mean by "expanding selection horizontally" vs "vertically"?

Reinmar:
We've talked with @tomalec about the algorithm and it can go both ways:

*   with the model check for "editable limits" as proposed by @oleq and then me in one of the comments,
*   with a view check as "editable element" is a view concept.

I'm fine with both.

Reinmar:
> Could you explain what you mean by "expanding selection horizontally" vs "vertically"?

I understand that @oleq meant that expanding vertically means adding parent elements to the selection (e.g. from image  caption to image element). Expanding horizontally meant adding siblings.

oleq:
@Reinmar I meant that <kbd>Ctrl+A</kbd> should exclude "parent" steps from the algorithm like I explained here

> If we consider a `root > table > cell > image > caption with selection` scenario, with this implementation it will require 5 keystrokes to select the entire content of the editor. If we skip the widget selection that would be only 3 keystrokes.

Maybe "horizontal/vertical" are not the best terms to describe it.

tomalec:
I changed the behavior to:

In `root > table > cell > image > caption with selection` successive <kbd>Ctrl+A</kbd>s selects all in: `caption > cell > root`.

There is still `do ... while`, but, to `place` and I believe code is shorter and cleaner now.

The code is ready for another round of review.

packages/ckeditor5-select-all/tests/selectallcommand.js
```diff
@@ -103,5 +141,83 @@ describe( 'SelectAllCommand', () => {
...
+					'<tableRow>' +
+						'<tableCell>' +
+							'<paragraph>foo</paragraph>' +
+							'<image src="foo.png"><caption>bar</caption></image>' +
+						'</tableCell>' +
+					'</tableRow>' +
+				'</table>]'
+			);
+
+			// editor.execute( 'selectAll' );
```
oleq, line 193: Obsolete comment.

tomalec:
I added more docs. Ready for another round :)

oleq:
Just FYI @tomalec: I simplified the guide so people without deep understanding of editor internals (who are the target of our guides) can read it without getting into details such as limits etc..

Also, I reverted the demo editor content because the structure you proposed is unlikely to ever happen. Let's demo the editor in such a way that it resembles real–life scenarios (I know that not every sample is perfect, though). It's hard to showcase the editor using academic (manual test–like) content.

PR #6834, opened 2020-05-15
#6785 Table walker API change

### LINKED ISSUE #6785: Improve TableWalker yielded options.
## TableWalker - intro

The purpose of the table walker is to iterate over a table "slots" not only table cells. The latter is a trivial task.

Iterating over table "slots" is required to handle operations on cells covering many "slots" (ie table cell with `rowspan` and/or `colspan` attributes).

TableWalker is used in two "modes":
1. Direct iterator - extracting some information about from a table without modifying it.
2. Constructing "tableMap" to modify a table - in most cases using slot information (and "previous cell").

The idea of "slots" is defined in W3C/WHATWG docs about tables and it makes sense to use that lingo.

### Current `TableWalkerValue`

```js
const tablewWalkerValue = {
    row: 1, // Often used
    column: 2,  // Often used
    cell: TableCell( { colspan: 2, rowspan: 2 } ),  // Often used
    colspan: 2, // Often used
    rowspan: 2, // Often used
    isSpanned: true, // <- confusing
    cellIndex: 0 // rarely used - most often previousCell is needed (batch insertions)
}
```

The most confusing here is `isSpanned` and `cell`. For slots that are covered by a table cell anchored in another slot we set `isSpanned = true` and the `cell` holds reference to the cell that spans over given slot.

---

## Proposal

The language used in `TableWalkerValue` is hard to grasp. The better idea is to use "slot" definition form [processing table model](https://html.spec.whatwg.org/multipage/tables.html#table-processing-model) specification.

### Introduce `TableSlot` typedef

The `TableSlot` is a name that is in line with docs describing table model - it describes over what we actually iterate.

```js
const tableSlot = {
    // Current slot indexes
    row: 1, // starts from 1
    column: 2,
    // Is the slot also an anchor? Better name needed.
    isAnchorCell: false, // True if cell is anchored in current slot.
    // Anchor cell - a cell that covers this slot.
    anchorCell: TableCell( { colspan: 2, rowspan: 2 } ),
    // Indexes of anchor cell - calculating this outside is expensive.
    anchorCellRow: 0,
    anchorCellColumn: 0,
    // Getter to make it read when needed becasue we do not store span=1 in the model.
    anchorCellWidth: () => { this.anchorCell.getAttribute( 'colspan' ) || 1 },
    // Better than `colspan` & `rowspan`.
    anchorCellHeight: () => { this.anchorCell.getAttribute( 'rowpsan' ) || 1 }
}
```

However, we could drop `anchor` here to have a bit more compact names:
```js
const tableSlot = {
    row: 1,
    column: 2,
    isAnchorCell: false,
    cell: TableCell( { colspan: 2, rowspan: 2 } ),
    cellRow: 0,
    cellColumn: 0,
    cellWidth: () => { this.cell.getAttribute( 'colspan' ) || 1 },
    cellHeight: () => { this.cell.getAttribute( 'rowpsan' ) || 1 }
}
```

### Creating a table walker - restricting yield slots

In most cases we do not need whole table information. Some algorithms operate on rows while others on columns. However, some will operate on a table section constrained by rows and columns.

```js
new TableWalker( table, {
    onlyAnchorCells: true, // (default value - which is kinda messy - )
    // hideNonAnchorSlots: false // (better default value, worse name)
    // Limit rows to return.
    startRow: 0, // (default value)
    endRow: tableHeight - 1, // (default value)
    // Limit columns to return.
    startColumn: 0, // (default value)
    endColumn: tableWidth - 1, // (default value)
    // was: column - iterate over one column only - I'd drop that.
} )
```

---

### Modifying table problem

Moved to: https://github.com/ckeditor/ckeditor5/issues/7326

---

niegowski:

Other (table): TableWalker API rebuild. Closes #6785.

MINOR BREAKING CHANGE: The `TableWalker` API and yielded values change.

---

### Additional information


packages/ckeditor5-table/src/tablewalker.js
```diff
@@ -53,69 +55,91 @@ export default class TableWalker {
...
-	 *		'Cell at 1 x 2 : has data'
+	 *		'Cell at 1 x 2 : is anchored'
 	 *		'Cell at 1 x 3 : is spanned'
-	 *		'Cell at 1 x 4 : has data'
-	 *		'Cell at 1 x 5 : has data'
+	 *		'Cell at 1 x 4 : is anchored'
+	 *		'Cell at 1 x 5 : is anchored'
+	 *
+	 * **Note**: Option `row` excludes `startRow` and `endRow` (use `row` or `startRow` and `endRow` but never together).
+	 * Option `column` excludes `startColumn` and `endColumn` (use `column` or `startColumn` and `endColumn` but never together).
```
jodator, line 74:
```suggestion
	 * **Note**: Option `row` is a shortcut that sets both `startRow` and `endRow` to the same row.
	 * (Use either `row` or `startRow` and `endRow` but never together). Similarly the `column` option sets both `startColumn`
	 * and `endColumn` to the same column (Use either `column` or `startColumn` and `endColumn` but never together).
```

packages/ckeditor5-table/src/tablewalker.js
```diff
@@ -370,30 +388,135 @@ export default class TableWalker {
...
+	 *
+	 * @readonly
+	 * @returns {Boolean}
+	 */
+	get isAnchor() {
+		return this.row === this.cellAnchorRow && this.column === this.cellAnchorColumn;
+	}
+
+	/**
+	 * The `colspan` attribute of a cell. If the model attribute is not present, it is set to `1`.
```
jodator, line 488:
```suggestion
	 * The widht of a cell defined by a `colspan` attribute. If the model attribute is not present, it is set to `1`.
```

packages/ckeditor5-table/src/tablewalker.js
```diff
@@ -370,30 +388,135 @@ export default class TableWalker {
...
+	 *
+	 * @readonly
+	 * @returns {Number}
+	 */
+	get cellWidth() {
+		return parseInt( this.cell.getAttribute( 'colspan' ) || 1 );
+	}
+
+	/**
+	 * The `rowspan` attribute of a cell. If the model attribute is not present, it is set to `1`.
```
jodator, line 498:
```suggestion
	 * The height of a cell defined by a `rowspan` attribute. If the model attribute is not present, it is set to `1`.
```

PR #7230, opened 2020-05-19
Fixed a bug where the rendered attribute name is always lower case

### LINKED ISSUE #7228: DOM converter doesn't preserve attribute letter casing

There's a problem in renderer, where it will force attribute name to be output as a lowercase name.

I pushed a branch with the unit test for that:

https://github.com/ckeditor/ckeditor5/blob/5709039272f541cb04939fc421705b427c7dc16a/packages/ckeditor5-engine/tests/view/domconverter/view-to-dom.js#L99-L104

```plaintext
FAILED TESTS:
  DomConverter
    viewToDom()
      × should preserve attribute letter casing
        Chrome 81.0.4044 (Windows 10.0.0)
      AssertionError: expected [ 'foobar' ] to include 'fooBar'
```

---

mlewand:

Fix (engine): Fixed a bug where the rendered attribute name is always lower case. Closes #7228.

MAJOR BREAKING CHANGE: Output HTML attributes are no longer lowercased. This relates to both, markup in the data and the editor output.

---

### Additional information

As mentioned in https://github.com/ckeditor/ckeditor5/issues/7228#issuecomment-630728949:

> And the reason for the whole problem is:
>
> https://github.com/ckeditor/ckeditor5/blob/5709039272f541cb04939fc421705b427c7dc16a/packages/ckeditor5-engine/src/view/domconverter.js#L245
>
> `HTMLElement.setAttribttribute()` forces implicitly converts attribute name to lower-case, as everyone might expect /s
>
> `HTMLElement.setAttribttributeNS()` being more XML oriented doesn't have this wonderful feature 😄

I consider this change to be a major breaking change, as it might wreak havoc for some editor configurations.

### DIFF
--- a/packages/ckeditor5-engine/src/view/domconverter.js
+++ b/packages/ckeditor5-engine/src/view/domconverter.js
@@ -242,7 +242,7 @@ export default class DomConverter {
 
 				// Copy element's attributes.
 				for ( const key of viewNode.getAttributeKeys() ) {
-					domElement.setAttribute( key, viewNode.getAttribute( key ) );
+					domElement.setAttributeNS( null, key, viewNode.getAttribute( key ) );
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


mlewand:
See https://github.com/ckeditor/ckeditor5/issues/7228#issuecomment-630776927

PR #7240, opened 2020-05-19
#5901: Use text nodes content directly instead of stringifying using data processor in code-block

### LINKED ISSUE #5901: Code Block displays HTML entities
Let me first thank you for the highly anticipated code block feature! Apart from this issue it seems to work great and covers all basics already.

1. Create empty code block:
image

2. Call `textEditor.getData()` and get `<pre><code class="language-text-plain">&nbsp;</code></pre><p>text</p>`

3. Create new instance of CKEditor and call `textEditor.getData('<pre><code class="language-text-plain">&nbsp;</code></pre><p>text</p>')` - i.e. with what `.getData()` previously returned.

### ✔️ Expected result

image

### ❌ Actual result

image

## 📃 Other details

Similar problem happens when code block contains "&", after which is returned from `.getData()` as `&amp;` but then displayed literally from `.setData()`.

* Browser: latest chrome/firefox
* OS: Ubuntu
* CKEditor version: 16.0.0

---

tomalec:
as suggested by @oleq at https://github.com/ckeditor/ckeditor5/issues/5901#issuecomment-624605013.
Ignore HTML elements inside `<pre><code>`.
Closes: #5901.

Fix (code-block): Stop rendering `&nbsp;` when restoring the state of an empty code block. Closes #5901.

---

### Additional information

There is still an issue with dom to view conversion
https://github.com/ckeditor/ckeditor5/compare/i/5901-code-blocks?expand=1#diff-82a5336a00edfc515fb87b52fc38eafbR892-R901
as you can see, filler element is not cleared on `getData`

From what I have found, `domConverter` https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/view/domconverter.js#L1027 uses `getDataWithoutFiller` which strips only **`INLINE_FILLER`**. I tried to change [`_processDataFromDomText`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/view/domconverter.js#L1023) to strip the block filler. However, it does not identify `&nbsp;` inside `<code>` as block filler, because `<code>` is an inline element.

I could do more changes, to let it change it anyway, given it already checked it's inside `<pre>` element which accepts only `Phrasing content`. But I have a doubt regarding that:

From what I understand `domConverter` is not aware of `code-block` feature. Therefore, for it `<code>` is as inline as any other element like `<strong>`, even inside `<pre>`. Shouldn't we use `\u200b` (zero-width space) in the first place? The pre-formatted behavior of `<pre>` guarantees the height for us? 

//cc @Reinmar

### DIFF
--- a/packages/ckeditor5-code-block/src/codeblockediting.js
+++ b/packages/ckeditor5-code-block/src/codeblockediting.js
@@ -128,7 +128,7 @@ export default class CodeBlockEditing extends Plugin {
 		editor.editing.downcastDispatcher.on( 'insert:codeBlock', modelToViewCodeBlockInsertion( model, normalizedLanguagesDefs, true ) );
 		editor.data.downcastDispatcher.on( 'insert:codeBlock', modelToViewCodeBlockInsertion( model, normalizedLanguagesDefs ) );
 		editor.data.downcastDispatcher.on( 'insert:softBreak', modelToDataViewSoftBreakInsertion( model ), { priority: 'high' } );
-		editor.data.upcastDispatcher.on( 'element:pre', dataViewToModelCodeBlockInsertion( editor.data, normalizedLanguagesDefs ) );
+		editor.data.upcastDispatcher.on( 'element:pre', dataViewToModelCodeBlockInsertion( editor.editing.view, normalizedLanguagesDefs ) );
 
 		// Intercept the clipboard input (paste) when the selection is anchored in the code block and force the clipboard
 		// data to be pasted as a single plain text. Otherwise, the code lines will split the code block and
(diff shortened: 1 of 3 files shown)


oleq:
The problem comes down to `&nbsp;` being rendered in the code element in editing when the data with `&nbsp;` is set:

```javascript
editor.setData( '<pre><code>&nbsp;</code></pre>' );
```

and the raw editing view looks as follows (and you can move the caret before/after the nbsp):

```html
<pre data-language="Plain text" spellcheck="false"><code class="language-plaintext">&nbsp;</code></pre>
```

but we'd rather expect the same behavior as with the paragraph:

```javascript
editor.setData( '<p>&nbsp;</p>' );
```

when the raw editing view is:

```html
<p><br data-cke-filler="true"></p>
```

Honestly... after a F2F call with @tomalec I came to the conclusion that we should follow through because from the end-user standpoint the [original issue](https://github.com/ckeditor/ckeditor5/issues/5901) is serious and we can discuss this extra space quirk in a follow-up. IMO it's not a big deal and it could end up in a bigger refactoring around the `DomCoverter` class.

tomalec:
Follow-up issue created https://github.com/ckeditor/ckeditor5/issues/7259

PR #7256, opened 2020-05-20
#6829: Improve docs for isLimit and isObject

### LINKED ISSUE #6829: The editing engine guide section about schema needs information about limits and objects

When working on #6816 I figured some key concept explanations are missing:

*   What is "limit" element? Why limit is not just closest editable (for instance, `tableRow` is also a limit).
*   What is object?
    *   What's the difference between an object and a limit?
*   API docs for `Schema#isLimit` and `Schema#isObject` are also vague, they explain the implementation not the semantics.

Something else @niegowski?

tomalec:

Docs: Improve the docs for "limit elements" and "object elements". Closes #6829.

---

### Additional information

Travis and Umberto complain about invalid links. However, ``{@link #allowIn `allowIn`}`` looks perfectly fine to me. Using full ``{@link module:engine/model/schema~SchemaItemDefinition#allowIn `allowIn`}`` within `@typeDef module:engine/model/schema~SchemaItemDefinition`looks redundant and confusing to me.
https://github.com/ckeditor/ckeditor5/compare/master...i/6829-isLimit-doc#diff-375a78efc71168f3eacd6055760a3389R1021

Should I change them to full ones, or adapt Umberto check to allow such cases?

### DIFF
--- a/packages/ckeditor5-engine/docs/framework/guides/deep-dive/schema.md
+++ b/packages/ckeditor5-engine/docs/framework/guides/deep-dive/schema.md
@@ -44,47 +44,72 @@ While this would be incorrect:
 </$root>
 ```
 
-## Declaring as a limit element
+## Defining additional semantics
 
-Consider a feature like an image caption. The caption text area should construct a boundary to some internal actions.
+In addition to setting allowed structures, the schema can also define additional traits of model elements. By using the `is*` properties a feature author may declare how a certain element should be treated by other features and the engine.
 
- - A selection that starts inside should not end outside.
- - Pressing <kbd>Backspace</kbd> or <kbd>Delete</kbd> should not delete the area. Pressing <kbd>Enter</kbd> should not split the area.
+### Limit elements
 
- It should also act as a boundary for external actions:
+Consider a feature like an image caption. The caption text area should construct a boundary to some internal actions:
 
- - A selection that starts outside, should not end inside.
+* A selection that starts inside should not end outside.
+* Pressing <kbd>Backspace</kbd> or <kbd>Delete</kbd> should not delete the area. Pressing <kbd>Enter</kbd> should not split the area.
+
+It should also act as a boundary for external actions. This is mostly enforced by a selection post-fixer that ensures that a selection that starts outside, should not end inside. That means that most actions will either apply to the "outside" of such an element or to a content inside it.
+
+Taken these characteristics, the image caption should be defined as limit element by using the {@link module:engine/model/schema~SchemaItemDefinition#isLimit `isLimit`} property.
 
 ```js
 schema.register( 'myCaption', {
-    isLimit: true
+	isLimit: true
 } );
 ```
 
-{@link module:engine/model/schema~SchemaItemDefinition#isLimit `isLimit`} makes the engine construct such boundaries, and let {@link module:engine/model/utils/selection-post-fixer `selection-post-fixer`} update the user's selection if needed.
+The engine and various features then check it via {@link module:engine/model/schema~Schema#isLimit `Schema#isLimit()`} and can act accordingly.
 
 <info-box>
-    "Limit element" does not mean "editable element". The concept of "editability" is reserved for the view and expressed by {@link module:engine/view/editableelement~EditableElement `EditableElement` class}.
+	"Limit element" does not mean "editable element". The concept of "editability" is reserved for the view and expressed by the {@link module:engine/view/editableelement~EditableElement `EditableElement` class}.
 </info-box>
 
-## Declaring as a self-sufficient object
+### Object elements
 
 For the image caption as in the example above it does not make much sense to select the caption box, then copy or drag it somewhere else.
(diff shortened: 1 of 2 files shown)


Reinmar:
Make sure to run `yarn docs` before pushing such changes ;)

packages/ckeditor5-engine/docs/framework/guides/deep-dive/schema.md
```diff
@@ -44,6 +44,48 @@ While this would be incorrect:
...
+    isObject: true
+} );
+```
+
+Every "object" is also a "limit" element. 
+
+<info-box>
+    It means for every element with `isObject` set to `true`, {@link module:engine/model/schema~Schema#isLimit `schema.isLimit( element )`} will always return `true`. 
+
+    However, {@link module:engine/model/schema~Schema#getDefinition `schema.getDefinition( 'element' )`} may return `false` in a case when `{ isLimit: true, isObject: true}` was registered.
```
oleq, line 86: Given that `Schema#getDefinition()` is only used internally in our codebase and it's unlikely anyone will use it in their integration (maybe it should be private BTW?) I think this clarification is unnecessary, even confusing.

mlewand:
Also it's best to target your docs changed at `stable` branch rather than `master` so that they get published sooner :+1:

packages/ckeditor5-engine/src/model/schema.js
```diff
@@ -219,15 +219,28 @@ export default class Schema {
...
 	 *		schema.isLimit( 'image' ); // -> true
 	 *
+	 * See the {@glink framework/guides/deep-dive/schema "Schema" deep dive} guide.
+	 *
+	 * It considers `isObject` with higher priority than `isLimit`. Meaning, even if
+	 * `isLimit` was explicitly  set to `false` in `SchemaItemDefinition`, but `isObject` was set to `true`,
+	 * `isLimit()` will return `true`.
+	 *
+	 * **Note**: In the above scenario {@link module:engine/model/schema~Schema#getDefinition `schema.getDefinition( item ).isLimit`}
+	 * will be `false`, to match the exact definition given.
```
oleq, line 242: As I mentioned before, I think this clarification about `Schema#getDefinition()` is unnecessary. 

> It considers \`isObject\` with higher priority than \`isLimit\`

Also, these API docs explain the implementation (which is very simple and anyone can see that) instead of the meaning (which is unclear when looking at the code). What we only need to know is:

*   that every object is a limit,
*   that a limit is a boundary that model selection cannot cross (both from the inside or from the outside),
*   that there's no operation in the editor that would end up in a the limit split into more fragments.

oleq: Another thing that is missing is a mention that every object is a limit but not very limit is an object.

packages/ckeditor5-engine/src/model/schema.js
```diff
@@ -241,15 +254,21 @@ export default class Schema {
 	}
 
 	/**
-	 * Returns `true` if the given item is defined to be
-	 * an object element by {@link module:engine/model/schema~SchemaItemDefinition}'s `isObject` property.
+	 * Returns `true` if the given item is should be treated as an object element.
+	 * If it should act as a self-sufficient element.
```
oleq, line 258: We should mention here that every object is automatically a limit.

Reinmar:
Guys, I'm doing some bigger changes in this documentation.

Unfortunately, the proposed changes added to already strongly duplicated content. That's hard to fix in this PR so I'm doing just a small cleanup of this. However, I'm going to revive the deep dive guide to be more complete.

packages/ckeditor5-engine/docs/framework/guides/deep-dive/schema.md
Reinmar: I don't agree. This method is part of public API. So it's ok to mention it in the deep dive as this is something that may be confusing.

tomalec:
> Unfortunately, the proposed changes added to already strongly duplicated content. That's hard to fix in this PR so I'm doing just a small cleanup of this. However, I'm going to revive the deep dive guide to be more complete.

Does it mean, I should stop any further work for this and #6829, and wait for you to finish?

Reinmar:
> Make sure to run `yarn docs` before pushing such changes ;)

Ooopsie. I see you did that :D

So, you cannot use such short links in some cases. You may need to expand them to `~Class#sth` or even entire `module:path~Class#sth`.

Reinmar:
> Does it mean, I should stop any further work for this and #6829, and wait for you to finish?

Yup.

I pushed my changes and I leave them to your review now.

Reinmar:
You may notice that I removed some information in some  places and moved as much as I could to the schema deep dive. Also, the guide is linked in more places. We should be forwarding people to one place where there's the most info. Having all this repeated in so many places is terrible for maintenance. If you'll find some more things that can be improved in line with this, please do.

tomalec:
> So, you cannot use such short links in some cases. You may need to expand them to `~Class#sth` or even entire `module:path~Class#sth`.

Thanks. I found `module:path~Class#sth` to verbose in that case, and `#sth` was generating working links. I didn't know that `~Class#sth` would pass the validation,  Will use that in future.

> I pushed my changes and I leave them to your review now.

LGTM.

PR #7270, opened 2020-05-22
#6680: Model#deleteContent() should not change the type of the trailing block 

### LINKED ISSUE #3115: Should deleteForward change the direction of model#deleteContent() merge blocks?
Let's see the two following cases:

TC1: `<h1>Foo</h1><p>[]Bar</p>` + delete (backspace)
TC2: `<h1>Foo[]</h1><p>Bar</p>` + forward delete

Now, we could handle these cases differently. Right now, in both of them the result will be `<h1>Foo[]Bar</h1>`. However, perhaps the user's intention is a bit different when pressing the backward and forward delete keys.

When using the editor I'm expecting that forward delete will merge the current block "to the right", so it will keep that block (`<p>Foo[]Bar</p>`). We should perhaps check some other editors to see if they handle forward delete differently.

### LINKED ISSUE #4434: Removing empty paragraph is merging and is bad UX
oct-01-2018 10-57-22

As you can see, there is an empty paragraph, and then there is a heading below it. Then, a user removes the empty paragraph. But under the hood, this is realized as a merge. So the heading is merged into an empty paragraph and the header becomes a paragraph.

AFAIR, I requested this "action" to be a merge because of collaboration :D. So when one user types in an empty paragraph and the other clicks <kbd>Backspace</kbd> in it, the text that was written by the first user is not deleted.

However, outside of a collaboration, it looks really bad. We should do something about it. It looks like a bug.

### LINKED ISSUE #6680: Backspace should not change the type of the trailing block (probably shouldn't merge it at all)

When having the following selection:

```
<paragraph>[Foo</paragraph>
<paragraph>Bar</paragraph>
<heading1>]Bom</heading1>
```

I expect to end with this after pressing Backspace:

```
<paragraph>[]</paragrph>
<heading1>Bom</heading1>
```

## Actual result

The format of the trailing block is changed to the format of the former block. Technically speaking, the heading is being merged into the first paragraph.

## UX

This is especially irritating when you try to delete e.g. paragraphs between 2 headings:

I get the "Proposal" heading turned into a paragraph. What I'd expect is having the two paragraphs above deleted, one left and the heading not touched at all:

## Background

The trailing block problem is a well-known DOM selection representation issue. For instance, when you triple-click a line of text to select the entire line, the selection looks like this:

```
<paragraph>Foo</parargaph>
<paragraph>[Trippleclicked line</parargaph>
<paragraph>]Foo</parargaph>
```

Unfortunately:

*   We didn't plan to select the "Foo" block. It got selected by the browser because that's the only way how it represents a "new line is selected too" situation which is a different situation than selecting just the content of this block.
*   You don't see in Chrome (FF and Safari handle this better) that the end of line is selected too.

We already handled this situation in the `selection.getSelectedBlocks()` method, so e.g. when yoy apply a block format to this selection, it does not affect the "Foo" block.

It'd now be good to also handle this in `deleteContent()` by simply ignoring the trailing block if the selection starts at its beginning.

---

niegowski:

Fix (engine): `Model#deleteContent()` should preserve the properties of more meaningful element. Closes #6680. 

---

### Additional information


niegowski:
This change replaces this behavior:



With this:



It's using `merge` operation as original `mergeBranches` to keep CF compatibility as described here: #4434.

niegowski:
@ckeditor/qa-team could you test it? This is affecting selected elements and also <kbd>Backspace</kbd>/<kbd>Delete</kbd> behavior.

FilipTokarski:
Seems to be ok in all browsers 👍

packages/ckeditor5-engine/src/model/utils/deletecontent.js
```diff
@@ -97,89 +97,254 @@ export default function deleteContent( model, selection, options = {} ) {
...
+	}
+
+	// Check if operations we'll need to do won't need to cross object or limit boundaries.
+	// E.g., we can't merge endElement into startElement in this case:
+	// <limit><startElement>x[</startElement></limit><endElement>]</endElement>
+	return checkCanBeMerged( startPosition, endPosition, schema );
+}
+
+// Returns the elements that are the ancestors of the provided positions that are direct children of the common ancestor.
+function getElementsNextToCommonAncestor( positionA, positionB ) {
```
jodator, line 337: This does something different than its name. This method returns elements under common ancestor not next to common ancestor.

`<ancestor><h1>[</h1><h2>]</h2></ancestor><nextToAncestor></nextToAncestor>`

The method will return `<h1>` and `<h2>` not `<nextToAncestor>`.

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -283,8 +322,20 @@ describe( 'DataController utils', () => {
 
 			test(
 				'leaves just one element when all selected',
+				'<paragraph>[x</paragraph><paragraph>foo</paragraph><paragraph>y]bar</paragraph>',
+				'<paragraph>[]bar</paragraph>'
+			);
+
+			test(
+				'leaves just one element when all selected (different name)',
```
jodator, line 330: I think that this test should reflect what merging happen.

Reading the test description I'm not sure what is the expected outcome and how this test differs from `leaves just one (first) element when all selected (different name)`.

Why first not last?

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -235,19 +237,56 @@ describe( 'DataController utils', () => {
 				'<paragraph>x</paragraph><paragraph>fo[]ar</paragraph><paragraph>y</paragraph>'
 			);
 
+			test(
+				'removes first element when it\'s empty but second element is not empty (same name)',
+				'<paragraph>x</paragraph><paragraph>[foo</paragraph><paragraph>b]ar</paragraph><paragraph>y</paragraph>',
```
jodator, line 242: The test description is a bit misleading and maybe a test case should properly show which node is removed. This would be showcased when using differnt node names.

`<blockA>[foo</blockA><blockB>b]ar</blockB` -> `<blockB>[]ar<blockB>` ?

Also the test name describes some intermediate state. The `<paragraph>foo</paragraph>` is not empty but the test describes it as "empty".

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -235,19 +237,56 @@ describe( 'DataController utils', () => {
...
+				'<paragraph>x</paragraph><paragraph>[]ar</paragraph><paragraph>y</paragraph>'
+			);
+
+			test(
+				'excludes end block from deletion if selection ends at start position of it',
```
jodator, line 247:
```suggestion
				'do not remove end block if selection ends at start position of it',
```

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -235,19 +237,56 @@ describe( 'DataController utils', () => {
...
+				'<paragraph>x</paragraph><paragraph>[foo</paragraph><paragraph>]bar</paragraph><paragraph>y</paragraph>',
+				'<paragraph>x</paragraph><paragraph>[]</paragraph><paragraph>bar</paragraph><paragraph>y</paragraph>'
+			);
+
+			test(
+				'removes empty element',
+				'<paragraph>x</paragraph><paragraph>[</paragraph><paragraph>]bar</paragraph><paragraph>y</paragraph>',
```
jodator, line 254: Test description similar to:

"'removes first element when it\\'s empty but second element is not empty (same name)',"  but the initial state is different.

Needs explanation.

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -235,19 +237,56 @@ describe( 'DataController utils', () => {
...
+				'<paragraph>x</paragraph><paragraph>[]bar</paragraph><paragraph>y</paragraph>'
+			);
+
+			test(
+				'does not remove first element when first element contains object (same name)',
```
jodator, line 259: It is probably describing this _intermediate_ state - AFAICS ii is about having content (\`model.hasContnet()\` check in code).

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -317,6 +368,36 @@ describe( 'DataController utils', () => {
 				expect( mergeSpy.called ).to.be.true;
 			} );
 
+			it( 'uses merge operation if first element is empty (because of content delete) and last is not', () => {
```
jodator, line 371: Maybe using "merges elements if..." - the merge operation is an test/implementation detail.

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -344,138 +425,197 @@ describe( 'DataController utils', () => {
...
+					'does not remove block element with nested element and object',
+					'<paragraph><pchild><widget></widget>[foo</pchild></paragraph><paragraph><pchild>b]ar</pchild></paragraph>',
+					'<paragraph><pchild><widget></widget>[]ar</pchild></paragraph>'
+				);
 
-					model.change( writer => {
-						writer.setSelection( range );
-					} );
+				test(
+					'merges heading element',
```
jodator, line 441: Not exactly IMO :) it does "merges `pchild` to `nchild` (and removes `<paragraph>`). However I'm not sure which case is that and probably more fluent description is needed.

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -344,138 +425,197 @@ describe( 'DataController utils', () => {
...
+					'<heading1><hchild>x[foo</hchild></heading1><paragraph><pchild>b]ar</pchild></paragraph>',
+					'<heading1><hchild>x[]ar</hchild></heading1>'
+				);
 
-					deleteContent( model, doc.selection );
+				test(
+					'removes heading element',
```
jodator, line 447: Why? Description does not say anything why we expect that in this situation.

packages/ckeditor5-engine/tests/model/utils/deletecontent.js
```diff
@@ -344,138 +425,197 @@ describe( 'DataController utils', () => {
...
+					'<heading1><hchild>[foo</hchild></heading1><paragraph><pchild>b]ar</pchild></paragraph>',
+					'<paragraph><pchild>[]ar</pchild></paragraph>'
+				);
 
-					expect( getData( model ) )
-						.to.equal( '<pparent>x<paragraph>x<pchild>fo[]ar</pchild>y</paragraph>y</pparent>' );
-				} );
+				test(
+					'does not remove heading element if contains object',
```
jodator, line 453: But it would remove  heading even though it contains an object?

`'<heading1><hchild>[<widget></widget>foo</hchild></heading1><paragraph><pchild>b]ar</pchild></paragraph>',`

The test description is about a state after deletion and before merging not the input state.

packages/ckeditor5-engine/src/model/utils/deletecontent.js
```diff
@@ -97,89 +97,254 @@ export default function deleteContent( model, selection, options = {} ) {
...
+	}
+
+	return [
+		LivePosition.fromPosition( startPosition, 'toPrevious' ),
+		LivePosition.fromPosition( endPosition, 'toNext' )
+	];
+}
+
+// Finds the lowest element in position's ancestors which is a block.
+// It will search until first ancestor that is a limit element.
```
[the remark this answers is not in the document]
niegowski: This is simplified copy&paste from https://github.com/ckeditor/ckeditor5/blob/893b873002c838fb10efda35515ccfaf44c3e39a/packages/ckeditor5-engine/src/model/selection.js#L838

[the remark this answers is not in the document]
jodator: A more obvious info about `null` in that situation would be enough.

ps.: I saw a git blame for that :P I was too lazy back than probably :smile:

Issue #4665, opened 2019-07-05
Customize keys to accept mention


felipegiotto:
Currently, the plugin always select the current item when the user hit Enter, Tab or Space keys. My suggestion is to allow the user to customize these keys. For instance, I may want to disable the "Space key" shortcut, so if the user has the mention box opened and hit Space, the box closes and the editor keeps only the plain text, without mentions.

zadam:
From my perspective the "space" issue is severely limiting usefulness of this plugin to only simple use cases - even the standard use case of mentioning somebody by name does not fully work since names contain space between first name and last name.

If having this configurable would take too much time then a quickfix could be considered of just allowing space (resp. not treating it as a "commit key") as a more sensible default.

cnickless:
Not allowing a space to be written after the marker makes for a broken typing experience. It restricts users from being able to simply type marker characters (e.g. #, @) with a space following it.  Right now if I wanted to just type "#[space]" the first option will be selected.  If I delete those characters to go back to just the #, the feed is shown again.

To get past this, the user would have to either:

1. Type text after the tag, then use their mouse to select the cursor to the tag, delete the characters and then move the cursor back to the end of the string.
2. Type in sequence:  #[esc][space]

That's quite the unnatural behavior placed on the user.

I'd recommend just updating the default keys to select being ENTER and TAB, as those feel like the more specialized keys for this.

To test the ideal scenario -- type a comment on GitHub here down below ... type "#[Space]" and notice how the mention feed appears and then disappears.

jaigupta10:
Any update on this one? Is it working and if so then how.

zadam:
@jaigupta10 There's PR ckeditor/ckeditor5-mention#92 which fixes the space behavior but unfortunately it wasn't merged yet and has not activity :-(

mlewand:
Note that this issue focuses on being able to customize the keys used for committing the mention.

@zadam @cnickless @neongreen your feature request is valid (ability for mentions to have a space) but that's [a separate issue dedicated to handling space in mention value](https://github.com/ckeditor/ckeditor5/issues/9741). Please add 👍 reaction there to show your support.

Unfortunately mentioned PR ckeditor/ckeditor5-mention#92 (which became #7306 after our repo migration) merely removed space from list of keys that accept a mention. While logic of mention (which were broken by spaces) remained unchanged.

PR #7306, opened 2020-05-27
Removed the confirmation of mention selection by using the space key

### LINKED ISSUE #6394: Mention: do not confirm selection on "space"

Currently, it is cumbersome to simply type text that uses the mention marker but is not in the mention list, in some cases.

For example, suppose this:

*   My app loads just the last 500 issue numbers of CKEditor 5 in the mention list.
*   Let's say we have 1200 issue is total -> therefore issues from 700 to 1200 are loaded.
*   My app uses hash (#) as the marker to mention issues.
*   I want to mention issue 1. I type `#1`.
    *   At this point, the mention list is shown. As the list is filtered by what I typed, I see issues 1000, 1001, 1002... with issue 1000 pre-selected.
*   I want to continuer typing after `#1`, so I hit `space`.
    *   **Issues**: my typed `#1` is replaced with `#1000`. That's because `space` was used to confirm my selection in the list, which is clearly not what I wanted.

This seems to be caused by this line, which tells me that even `tab` could be used to confirm my select :thinking:   
https://github.com/ckeditor/ckeditor5-mention/blob/9a514ccb0dc8b595100fd4459a82fa67cc329d1d/src/mentionui.js#L124

I propose to allow confirming the selected item by using the `enter` key, exclusively.

## 📃 Other details

*   CKEditor version: v17.0.0
*   Installed CKEditor plugins: mention

---

fredck:

Other (mention): The `space` keys is not anymore used to confirm a mention selection from the list. Closes #6394.

---

### Additional information

Original PR: ckeditor/ckeditor5-mention#92


pammann-work:
@fredck 
Why not doing the same with keyCodes.space as with keyCodes.esc, calling `this._hideUIAndRemoveMarker();`? e.g. in github when using the @, if you enter space after the @, the menu goes away....

fredck:
> Why not doing the same with keyCodes.space as with keyCodes.esc, calling `this._hideUIAndRemoveMarker();`? e.g. in github when using the @, if you enter space after the @, the menu goes away....

@pammann-work, I'm not sure I've understood this. With the provided patch, if you enter space after @ the menu goes away, exactly as you described. In fact, the final behavior will match exactly what you see in GitHub. Am I missing something?

pammann-work:
@fredck, good to hear to it will work like in GitHub. I was just looking at the code changes and did not try it out. Thanks for your confirmation.

robjbrain:
I came to this issue by following a path of closed issues 
https://github.com/ckeditor/ckeditor5/issues/4665#issuecomment-633198527
https://github.com/ckeditor/ckeditor5-mention/pull/92#issuecomment-634623779

The original request was that we be able to mention names that have a space in them e.g. "John Doe" rather than closing the mention dialog after typing the space in "John ".

I can see that this change was merged and I can see the updated code in my node_modules directory so i'm definitely up to date but the functionality is still the same, the menu is closed when a space is typed.

Is there a solution that prevents the menu from closing on a space?

EricPaulson:
This PR does not solve the problem. As soon as I press the spacebar, the mention dropdown disappears. This is not how any user would expect this feature to work.

PR #7309, opened 2020-05-27
I/6015

### LINKED ISSUE #6015: Core: provide a `Editor` creation factory
Before proposing anything, let me illustrate the two cases I've faced recently, which motivated this issue.

### Case 1

While working on [https://github.com/ckeditor/ckeditor5-markdown-gfm/pull/30](https://github.com/ckeditor/ckeditor5-markdown-gfm/pull/30), I've noticed that, although the plugin is setting `editor.data.processor` to the GFM data processor, the editor classes (classic, inline, etc.) are in any case pointlessly (at that point) setting that same property to the HTML data processor (which later on is overwritten by the plugin).

I figured that it would be safe to set the data processor only after initing all plugins, if no plugin set it. Then, I've also figured that this would involve touching all four official editor classes, basically deleting and pasting the same lines, in the same places. It would not fix other editors out there though.

### Case 2

While coding a custom editor class, it is depressing the amount of copy/paste that has do be done. One does it without even thinking why, copying the boilerplate from the [relative documentation guide](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/ui/custom-editor-creator.html).

All that sounded overcomplicated and pointless. I see this as a weak part of the API.

### Proposal

The creation of a factory that will remove most of the boilerplate from the editor classes and centralize it in a single place.

---

fredck:

Feature (core): Introduced a factory for classes implementing `Editor`. Closes #6015.

Other (editor-balloon): Refactored the editor class to use EditorFactory. Follow-up to #6015.

Other (editor-classic): Refactored the editor class to use EditorFactory. Follow-up to #6015.

Other (editor-decoupled): Refactored the editor class to use EditorFactory. Follow-up to #6015.

Other (editor-inline): Refactored the editor class to use EditorFactory. Follow-up to #6015.

---

### Additional information

Proposing a full implementation for the factory:

*   Includes a factory for `create()`.
*   Includes a factory for `destroy()`.
*   Makes it possible to customize every step of the creation process. The changes in ClassicEditor are a good example of it.
*   Reduced the code of all four editor implementations to their very minimum, holding now just code specific for each implementation (no code duplication... kinda... we could make that textarea check a util).

Before moving forward with more tests and documentation, **I'm asking here for feedback. Thanks!**

---

Original PRs:

*   ckeditor/ckeditor5-core#206
*   ckeditor/ckeditor5-editor-balloon#41
*   ckeditor/ckeditor5-editor-classic#101
*   ckeditor/ckeditor5-editor-decoupled#46
*   ckeditor/ckeditor5-editor-inline#63


fredck:
I've been trying to debug the Travis fail with no success. [It insists breaking on a test](https://travis-ci.org/github/ckeditor/ckeditor5/builds/692089432#L1443) which I was able to confirm locally but then fixed and pushed back into the PR branch (force-push).

When I run the very same test locally, it works fine. This is the output I have: https://gist.github.com/fredck/bd0c6ba5484cd347fafd608409df001a

Don't know what else to do. Can someone confirm that the "media-embed" is not passing in the i/6015 branch?

mlewand:
It's been a while since this proposal has been made. For now, the idea of swapping data processor using a plugin works well (just like it is [done in md plugin](https://github.com/ckeditor/ckeditor5/blob/82cc5431f2a9b17149bc872d8b4cb5c71554f754/packages/ckeditor5-markdown-gfm/src/markdown.js#L27)). Technically it's possible to potentially have conflicting plugins that will override the data processor, but this case boils down to the editor configuration. We expect that one doesn't add Markdown (processing) plugin and, say, XML (processing) plugin at the same time.

Other than that the current implementation stood the test of time pretty well, so we don't see a need to change things in this area. I'm closing the PR but leave the issue open in case it's to get a bigger support in the future.

PR #7351, opened 2020-06-01
Re-enable Memory tests

### LINKED ISSUE #1731: It looks like memory leaks tests are broken again
The first occurence of `master-revision` CI started to fail today: https://travis-ci.org/ckeditor/ckeditor5-editor-classic/jobs/524842961. The previous one is OK: https://travis-ci.org/ckeditor/ckeditor5/builds/523466495.

It looks like Chrome introduces some changes as previos Jobs were executed on Chrome `73.x` and the failing one is on Chrome `74.x`.

I've tested locally and all previous builds on `master-revisions` branch fails locally.

jodator:

Internal: Enable memory leak tests. Closes #1731.

---

### Additional information

*   This is a long-running draft PR to check if the memory tests are stable on master.


tests/manual/memory/memory.md
```diff
@@ -0,0 +1,31 @@
...
+
+* Browser extensions might attach event listeners to the DOM so the safest way to run this test is by running Chrome from command line:
+
+    ```
+    google-chrome \
+        --enable-precise-memory-info \
+        --disable-extensions \
+        --disable-plugins \
+        --incognito \
+        http://localhost:8125
```
mlewand, line 26:
```suggestion
        http://localhost:8125
```

Extra tailing space. This behavior is defined in our [`.editorconfig`](https://github.com/ckeditor/ckeditor5/blob/5948bcff409d79bb6b1f8b6dce949eb1d5c6d38b/.editorconfig#L11) double check your editor settings.

mlewand: Actually it might be even more convenient to link straight to manual test: `http://localhost:8125/ckeditor5/tests/manual/memory/memory.html`?

tests/manual/memory/memory.html
```diff
@@ -0,0 +1,186 @@
+<p>
+	<button id="destroyEditor">Destroy editor</button>
+	<button id="initEditor">Init editor</button>
+</p>
+
+<head>
+	<meta http-equiv="Content-Security-Policy" content="script-src 'self' https://ckeditor.com 'unsafe-inline' 'unsafe-eval'">
```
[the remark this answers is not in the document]
jodator: Thanks, that was a copy-paste from two files to check editor with all the features enabled. This is fixed now.

jodator:
> I think it's OK to disable it on win only, as there's a good value in having it running on CI.

I don't remember having that run on win previously but that's a pity. I agree that it would be valuable to have it at least on the CI.

mlewand:
The thing is that we need to disable these tests in windows environment before merging. Otherwise `yarn run test` fails for me every time.

I would assume that it's a client side (win browser) so we should do a user agent check to verify whether it's a windows and if it is then skip the test for it.

We already have a check for mac, so it could be used as an inspiration: https://github.com/ckeditor/ckeditor5/blob/ced964d3d4364e7969ca46d1f404e5288bb53717/packages/ckeditor5-utils/src/env.js#L78-L80

Issue #1108, opened 2018-06-25
Typing a letter on the end of the link followed by the same character creates another link


f1ames:
## Is this a bug report or feature request? (choose one)

🐞 Bug report

## 💻 Version of CKEditor

`v10.1.0`

## 📋 Steps to reproduce

1. Go to https://ckeditor5.github.io.
1. Place caret on the end of the link (outside link) - the link has to be the last element inside a block element.
1. Type `a`.
1. Place caret on the end of the same link, but now inside of it.
1. Type `a`.

![chrome link](https://user-images.githubusercontent.com/1061942/41849540-9714f30e-7881-11e8-82e2-a77e2e469849.gif)

## ✅ Expected result

The `a` character is part of the link.

## ❎ Actual result

The `a` character creates new link after non-linked `a`.

## 📃 Other details that might be useful

Tested on Chrome `67.0.3396.87`.

Reinmar:
Will be fixed once we use ZWSes on both ends of inline elements.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity over the last year. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

Issue #946, opened 2018-04-06
The two-step caret movement requires 3 steps on the left side of a link


jodator:
## Is this a bug report or feature request? (choose one)

🐞 Bug report

## 💻 Version of CKEditor

Latest master (f31e6e1)

## 📋 Steps to reproduce

1. I've encountered it in build classic sample.
2. Navigate through text by cursor towards link (from it's left side so using right arrow).
3. Place a caret before last character of a word preceding a link.
4. press right arrow twice (you are "before" link)
5. press right arrow once (you are still "before" link)
6. press right arrow again (you are inside a link)

## ✅ Expected result
you should only need one right arrow press to enter link

## ❎ Actual result

you need press right arrow twice

## 📃 Other details that might be useful

![screencast 2018-04-06 13_30_52](https://user-images.githubusercontent.com/247363/38419049-2558f97a-399f-11e8-9fb5-0d2ba730ad9a.gif)

jodator:
That's weird but the first try (freshly instantiated editor) does not show this error. Only the second and next movements will show this.

scofalik:
We know about problems with 2-step-caret-movement. It will be refactored, maybe rewritten. There are more bugs.

oleq:
I'm almost sure this is a DUP of https://github.com/ckeditor/ckeditor5-engine/issues/1301. But let's wait for the research first.

Reinmar:
We've got so many tickets for two-step caret movement that it'd be good to create an epic for them.

PR #7356, opened 2020-06-01
Enable two-step caret movement for `code` attribute.

### LINKED ISSUE #6722: Allow distinguishing between "cursor inside/outside inline code" (same as for links)

Currently, if you are at the end of a link, you can press <kbd>Right Arrow</kbd> once to "escape" it and type outside of the link. The link is highlighted when I'm inside it, but not when I'm outside. This is very convenient.

It would be great to have the same behavior for inline code. The only other editor I know that does this is the one used by Bitbucket. In this demo, note how I can escape the `<code>` merely by pressing <kbd>Right Arrow</kbd>.

---

tomalec:

Other (basic-styles): Enable typing around for `code` element. Closes #6722.

---

### Additional information

The change is mostly copy-pasted from `LinkEditing`

I'm not sure whether such a tiny detail is worth mentioning in high-level docs at https://ckeditor.com/docs/ckeditor5/latest/features/basic-styles.html#available-text-styles.

Also, I wasn't sure whether highlighting should go into the scope of this issue. I made a draft at [`i/6722-code-caret-highlight`](https://github.com/ckeditor/ckeditor5/compare/i/6722-code-caret-highlight?expand=1). I think If it should be implemented here, I'd make a shared util/mixin so the code could be shared between `Link` and `Code`.


tomalec:
We (@oleq, @niegowski and me) agreed today on F2F meeting, that visual feedback is necessary. 

I'll continue as started on [`i/6722-code-caret-highlight`](https://github.com/ckeditor/ckeditor5/compare/i/6722-code-caret-highlight?expand=1) branch, move highlighting code to [`..engine/src/utils/bindtwostepcarettoattribute.js`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/utils/bindtwostepcarettoattribute.js), make `Link` use it as well, and mention it in the docs.

tomalec:
I've made link's highlight code slightly more re-usable,
used it for `code` as well, mentioned two-step movement in code feature docs.

The highlight effect is not visible in docs, due to a selector with higher specificity at https://github.com/cksource/umberto/blob/b8c9196cd2e653f2cf820be3291279a53bf83d9c/themes/umberto/src/css/_formatted.scss#L100

If the changes in this PR get approved, I'll create a followup PR there.

tomalec:
Thanks, I'll cover those cases and add tests. I was not able to reproduce the first case, but maybe I'll get to it when investigating and solving others. Otherwise, I'll bother you a little, to help me with that.

oleq:
The last one looks like #1108.

tomalec:
I was not able to reproduce the highlight issue - I have [added a test](https://github.com/ckeditor/ckeditor5/pull/7356/commits/cd33448bb9723a0ccfbfd763abba5052e3f16134) to check against that.

I was trying to address "the same character" issue, but with no luck. I've [added the test to make sure it's not the result of 2SCM](https://github.com/ckeditor/ckeditor5/pull/7356/commits/32f18eaa0a7ea68b55d3622a4bf31ca605b71e08). Also, the issue is reproducible after commenting out the 2SCM and highlight code. 

I'll spend a few hours more trying to find a solution. In case I would not find any by Monday, I think this PR could be reviewed again, and follow it up at #1108.

oleq:
> I was not able to reproduce the highlight issue - I have [added a test](https://github.com/ckeditor/ckeditor5/pull/7356/commits/cd33448bb9723a0ccfbfd763abba5052e3f16134) to check against that.

Take a look at this screencast, the left arrow is pressed thrice. 

## Actual

1.  The first key press moves the selection and overrides the gravity. If you start typing, a plain text is entered (✅ OK)
2.  The second one does nothing, it does not change the selection/selection attributes/gravity. If you start typing, a plain text is entered again (❌ NOK)
3.  The third one restores the gravity which gives the selection both `code` and `linkHref` attributes. If you start typing, a linked code is extended (✅ OK)



## Expected

1.  The first key press moves the selection and overrides the gravity.
2.  The second key press allows you to continue typing in a link (but not a code)
3.  The third key press allows you to type a linked code.

tomalec:
Ok, I'll get into this, looks similar to https://github.com/ckeditor/ckeditor5/issues/946

tomalec:
Closing in favor of https://github.com/ckeditor/ckeditor5/pull/7577

PR #7362, opened 2020-06-02
#6609: Removing empty columns (with no cells anchored)

### LINKED ISSUE #6609: Cleanup colspans on single table row and single column.

See: https://github.com/ckeditor/ckeditor5-table/pull/303#pullrequestreview-393616120.

```
+----+----+----+
| 00 | 01      |
+----+----+----+
| 10 | 11 | 12 |  <-- remove this row
+----+----+----+
```

### ✔️ Expected result

```
+----+----+
| 00 | 01 |
+----+----+
```

❌ Actual result:

```
+----+----+----+
| 00 | 01      |
+----+----+----+
```

---

niegowski:

Other (table): Removing empty columns (with no cells anchored). Closes #6609.

---

### Additional information


packages/ckeditor5-table/src/commands/mergecellscommand.js
```diff
@@ -69,10 +70,21 @@ export default class MergeCellsCommand extends Command {
...
+				} );
+			} else {
+				// If there were some rows removed then empty columns were already verified.
+				const emptyColumnsIndexes = getEmptyColumnsIndexes( table );
+
+				if ( emptyColumnsIndexes.length ) {
+					emptyColumnsIndexes.reverse().forEach( column => {
+						tableUtils.removeColumns( table, { at: column, batch: writer.batch } );
+					} );
+				}
```
jodator, line 87: AFAICS:

1.  This is commonly used together.
2.  Finding empty rows is cheap and can be done without `emptyRowIndexes` in above algorithm.
3.  Part of fixing logic is used as util and part is here.

I think it would be beneficial to check if this logic can be merged to one: `removeEmptyRowsColomuns` or two (`removeEmptyRows()` and `removeEmptyColumns()` ). This way:

*   it is a bit better to have whole "cleanup" logic in one place - it is used this way in the code
*   no need to think how to use `getEmptyColumnIndexes()`
*   the logic of removing empty rows can be in near removing empty columns.

Depending on the outcome, name, etc I would also think where to put this function(s) - `structure.js` is OK.

packages/ckeditor5-table/src/utils/structure.js
```diff
@@ -305,3 +305,36 @@ function addHeadingsToCroppedTable( croppedTable, sourceTable, startRow, startCo
...
+ *     +----+----+----+----+         +----+
+ *     | 10 | 11      | 13 |         | 16 |
+ *     +----+----+----+----+----+----+----+
+ *     | 20 | 21      | 23 | 24      | 26 |
+ *     +----+----+----+----+----+----+----+
+ *                  ^--- empty ---^
+ *
+ * Will return: [ 2, 5 ]
+ */
+export function getEmptyColumnsIndexes( table ) {
```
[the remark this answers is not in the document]
jodator: The above comments are invalidated in favor of the above comment in this util usage :point\_up: .

Side note: When adding a jsdoc to a "public" API (" - its internal) a usage example would be nice. (this is to be removed or make internal anyway so no need to change that ATM - just general note :) .

packages/ckeditor5-table/src/commands/mergecellscommand.js
jodator: ps.: when documenting new funcitons link them together and write when to use which (if needed).

packages/ckeditor5-table/src/utils/structure.js
```diff
@@ -305,3 +305,36 @@ function addHeadingsToCroppedTable( croppedTable, sourceTable, startRow, startCo
...
+	const tableMap = Array.from( new TableWalker( table, { includeAllSlots: true } ) );
+	const lastColumn = tableMap[ tableMap.length - 1 ].column;
```
jodator, line 327: I'm concerned about this - according to HTML table model docs the first row defines the length of a table. Here however it is probably needed to check all columns width. So maybe a comment on that is needed.

But if not, a `TableUtils.gerColumns()` + `TableWalker` without `includeAllSlots` should do the trick to get `lastColumn`

PR #7370, opened 2020-06-03
i/4858: Automatically add a link protocol when not provided in the link form.

### LINKED ISSUE #4858: Automatically add a protocol to a link
## Is this a bug report or feature request?

🆕 Feature request

The plugin should give an option for adding a protocol to a link automatically if a user doesn't provide it.

---

panr:

Feature (link): A link protocol will be automatically added to a link when not provided by the user in the link form. Closes #4858.

MINOR BREAKING CHANGES (link): This PR introduces `config.link.defaultProtocol` API.
  - The `Link` plugin uses `http://` protocol by default, when no configuration provided.

### DIFF
--- a/packages/ckeditor5-link/src/linkui.js
+++ b/packages/ckeditor5-link/src/linkui.js
@@ -9,7 +9,7 @@
 
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import ClickObserver from '@ckeditor/ckeditor5-engine/src/view/observer/clickobserver';
-import { isLinkElement } from './utils';
+import { isLinkElement, DEFAULT_PROTOCOL } from './utils';
 import ContextualBalloon from '@ckeditor/ckeditor5-ui/src/panel/balloon/contextualballoon';
 
 import clickOutsideHandler from '@ckeditor/ckeditor5-ui/src/bindings/clickoutsidehandler';
@@ -143,8 +143,9 @@ export default class LinkUI extends Plugin {
 	_createFormView() {
 		const editor = this.editor;
 		const linkCommand = editor.commands.get( 'link' );
+		const defaultProtocol = editor.config.get( 'link.defaultProtocol' ) || DEFAULT_PROTOCOL;
 
-		const formView = new LinkFormView( editor.locale, linkCommand );
+		const formView = new LinkFormView( editor.locale, linkCommand, defaultProtocol );
 
 		formView.urlInputView.fieldView.bind( 'value' ).to( linkCommand, 'value' );
 
@@ -154,7 +155,16 @@ export default class LinkUI extends Plugin {
 
 		// Execute link command after clicking the "Save" button.
 		this.listenTo( formView, 'submit', () => {
-			editor.execute( 'link', formView.urlInputView.fieldView.element.value, formView.getDecoratorSwitchesState() );
+			const { value } = formView.urlInputView.fieldView.element;
+			const hasProtocol = ( /\w+:\/\//gmi ).test( value );
+			const isEmail = ( /[\w-]+@[\w-]+\.+[\w-]+/gmi ).test( value );
+
+			// It's good to check config second time on submit for more dynamic cases
+			// when protocol might change after the element rendered.
+			const protocol = isEmail && 'mailto:' || editor.config.get( 'link.defaultProtocol' ) || defaultProtocol;
+			const parsedValue = !hasProtocol && value ? protocol + value : value;
+
+			editor.execute( 'link', parsedValue, formView.getDecoratorSwitchesState() );
 			this._closeFormView();
 		} );
 
--- a/packages/ckeditor5-link/tests/linkui.js
+++ b/packages/ckeditor5-link/tests/linkui.js
@@ -8,7 +8,7 @@
 import ClassicTestEditor from '@ckeditor/ckeditor5-core/tests/_utils/classictesteditor';
 import testUtils from '@ckeditor/ckeditor5-core/tests/_utils/utils';
 import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
-import { setData as setModelData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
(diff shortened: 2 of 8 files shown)


panr:
As we spoke, I went with the UI level of change... but when I was writing tests I thought that maybe we should add this feature for the model level as well? 
Now `LinkCommand.execute('ckeditor.com')` won't add any protocol (as before), so maybe we should be more consistent here 🤔

panr:
I still have to update feature docs, but the code can be reviewed.

jodator:
The code generally looks OK so I'll wait for the tests. One thing to change [which slipped us](https://github.com/ckeditor/ckeditor5/issues/4858#issuecomment-635872767):

> 2\. Should be a opt-in feature.

As for the command auto-fix. I'm not 100% sure whether it is a good or bad idea. Could you check in what scenarios the link command is used? My thinking is that if someone (dev) use the command to insert link then we shouldn't alter it but if the link command is executed by user action we should alter it. But I can't think of other options than UI.

The second scenario is handling link in pasted content. But that seems to be out-of scope for this issue.

panr:
> As for the command auto-fix. I'm not 100% sure whether it is a good or bad idea.

Me neither...

> Could you check in what scenarios the link command is used? My thinking is that if someone (dev) use the command to insert link then we shouldn't alter it but if the link command is executed by user action we should alter it. But I can't think of other options than UI.

What I've seen that we're using the command only on submitting the value (but here the command already gets auto-fixed value from the form input) and in the unit tests... I haven't spotted anything else...

And maybe you're right saying that from the developer's perspective such autofixer isn't a good idea, because it can be unpredictable... So maybe we should stick with the user's perspective only.

jodator:
> So maybe we should stick with the user's perspective only.

Sounds good - let's keep that on UI part only then.

packages/ckeditor5-link/src/linkui.js
```diff
@@ -154,7 +155,21 @@ export default class LinkUI extends Plugin {
 
 		// Execute link command after clicking the "Save" button.
 		this.listenTo( formView, 'submit', () => {
-			editor.execute( 'link', formView.urlInputView.fieldView.element.value, formView.getDecoratorSwitchesState() );
+			const { value } = formView.urlInputView.fieldView.element;
+
+			// It's good to check the config for the second time on submit.
+			// This will cover more dynamic cases, when protocol might change after the element has been rendered.
+			const defaultProtocol = editor.config.get( 'link.defaultProtocol' );
```
jodator, line 162: We do not have dynamic config in CKEditor5 - #7383.

At this stage I'd made that static - changing manual test would be a PITA but this might be seen as a side-effect in the current code base. Other features read config once - at init stage.

jodator: ps.: If dynamic options were already established somwhere else we can leave it but please add a comment in #7383.

packages/ckeditor5-link/src/linkui.js
```diff
@@ -154,7 +155,21 @@ export default class LinkUI extends Plugin {
...
+
+			// The regex checks for the protocol syntax ('xxxx://' or 'xxxx:')
+			// or non-word charecters at the begining of the link ('/', '#' etc.).
+			const isProtocolNeeded = !!defaultProtocol && !( /^((\w+:(\/{2,})?)|(\W))/gmi ).test( value );
+			const isEmail = ( /[\w-]+@[\w-]+\.+[\w-]+/gmi ).test( value );
```
jodator, line 167: Externalize those regexpes to not create new RegExp object on each callback.

packages/ckeditor5-link/src/linkui.js
```diff
@@ -154,7 +155,21 @@ export default class LinkUI extends Plugin {
...
+
+			const protocol = isEmail && 'mailto:' || defaultProtocol;
```
jodator, line 169:
```suggestion
			const protocol = isEmail ? 'mailto:' : defaultProtocol;
```

Straight forward if-else requires less time to parse ;)

packages/ckeditor5-link/tests/linkui.js
```diff
@@ -905,6 +905,121 @@ describe( 'LinkUI', () => {
...
+				formView.urlInputView.fieldView.value = 'email@example.com';
+				formView.fire( 'submit' );
+
+				expect( formView.urlInputView.fieldView.value ).to.equal( 'mailto:email@example.com' );
+				expect( getModelData( editor.model ) ).to.equal(
+					'[<$text linkHref="mailto:email@example.com">email@example.com</$text>]'
+				);
+			} );
+
+			it( 'should not add an email protocol when given provided wihitn the value' +
```
jodator, line 1012:
```suggestion
			it( 'should not add an email protocol when given provided within the value' +
```

packages/ckeditor5-link/tests/linkui.js
```diff
@@ -905,6 +905,121 @@ describe( 'LinkUI', () => {
...
+			it( 'should not add a protocol to the relative links even when `config.link.defaultProtocol` configured', () => {
+				editor.config.set( 'link.defaultProtocol', 'http://' );
+
+				formView.urlInputView.fieldView.value = '/test.html';
+				formView.fire( 'submit' );
+
+				expect( formView.urlInputView.fieldView.value ).to.equal( '/test.html' );
+			} );
+
+			it( 'should not add a protocol when given provided wihitn the value even when `config.link.defaultProtocol` configured', () => {
```
jodator, line 958:
```suggestion
			it( 'should not add a protocol when given provided within the value even when `config.link.defaultProtocol` configured', () => {
```

packages/ckeditor5-link/src/linkui.js
jodator: It will require changing the tests but only for the part when configured protocol is set would require a new describe block probably.

packages/ckeditor5-link/src/linkui.js
panr: 🙈

panr: I don't know what happened here 😬

PR #7378, opened 2020-06-04
#6443: BalloonToolbar should not be displayed for multi cell table selection.

### LINKED ISSUE #6443: Table properties balloon not displaying on cell multiselect in Balloon and Balloon Block editors

Occurs on latest master

1. Open Ballon or Balloon Block editor
2. Select multiple cells ( with some text content )

**In Balloon editor you don't event need to add any text**

### ✔️ Expected result

Table properties balloon shows up, you are able to edit a table

### ❌ Actual result

Editor balloon toolbar shows up, you can edit only content

Sample from Balloon Block editor:
16_table1

---

niegowski:

Fix (ui): Table properties balloon should be displayed on cell multi-select in Balloon and Balloon Block editors. Closes #6443.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ui/src/toolbar/balloon/balloontoolbar.js
+++ b/packages/ckeditor5-ui/src/toolbar/balloon/balloontoolbar.js
@@ -364,21 +364,24 @@ function getBalloonPositions( isBackward ) {
 	];
 }
 
+// Returns "true" when the selection has multiple ranges and each range contains an object
+// and nothing else.
+//
+// @private
+// @param {module:engine/model/selection~Selection} selection
+// @param {module:engine/model/schema~Schema} schema
+// @returns {Boolean}
 function selectionContainsOnlyMultipleObjects( selection, schema ) {
 	// It doesn't contain multiple objects if there is only one range.
-	if ( selection.rangeCount <= 1 ) {
+	if ( selection.rangeCount === 1 ) {
 		return false;
 	}
 
-	for ( const range of selection.getRanges() ) {
+	return [ ...selection.getRanges() ].every( range => {
 		const element = range.getContainedElement();
 
-		if ( !element || !schema.isObject( element ) ) {
-			return false;
-		}
-	}
-
-	return true;
+		return element && schema.isObject( element );
+	} );
 }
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-ui/src/toolbar/balloon/balloontoolbar.js
```diff
@@ -356,6 +364,23 @@ function getBalloonPositions( isBackward ) {
 	];
 }
 
+function selectionContainsOnlyMultipleObjects( selection, schema ) {
+	// It doesn't contain multiple objects if there is only one range.
+	if ( selection.rangeCount <= 1 ) {
```
oleq, line 369: A `DocumentSelection` in the model will always have at least one range. AFAIR there's no way to force an editor into a range-less selection state. Only virtual `Selection` instances can be range-less.

oleq, line 369: A `DocumentSelection` in the model will always have at least one range. AFAIR there's no way to force an editor into a range-less selection state. Only virtual `Selection` instances can be range-less.

packages/ckeditor5-ui/tests/toolbar/balloon/balloontoolbar.js
```diff
@@ -374,6 +375,15 @@ describe( 'BalloonToolbar', () => {
 			sinon.assert.notCalled( balloonAddSpy );
 		} );
 
+		// https://github.com/ckeditor/ckeditor5/issues/6443
+		it( 'should not add the #toolbarView to the #_balloon when the selection contains more than one fully contained objects', () => {
```
oleq, line 379:
```suggestion
		it( 'should not add the #toolbarView to the #_balloon when the selection contains more than one fully contained object', () => {
```

packages/ckeditor5-ui/src/toolbar/balloon/balloontoolbar.js
niegowski: So this should be strict `selection.rangeCount == 1` ?

PR #7381, opened 2020-06-04
#7379: Trimming pasted content for table to table pasting.

### LINKED ISSUE #7379: Support for pasting a table into a table with empty elements around pasted table.

This is extracted case from #6817. 

For pasted table that have some empty elements before or after it, we should extract only the table and use it for pasting.

---

niegowski:

Other (table): Pasting a table into a table is more tolerant for whitespaces around pasted table. Closes #7379.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/tests/tableclipboard-paste.js
+++ b/packages/ckeditor5-table/tests/tableclipboard-paste.js
@@ -359,6 +359,33 @@ describe( 'table clipboard', () => {
 			] ) );
 		} );
 
+		it( 'should alter model.insertContent if mixed content is pasted (p + p + table + p + br)', () => {
+			tableSelection.setCellSelection(
+				modelRoot.getNodeByPath( [ 0, 0, 0 ] ),
+				modelRoot.getNodeByPath( [ 0, 1, 1 ] )
+			);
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-table/tests/tableclipboard-paste.js
```diff
@@ -248,6 +248,131 @@ describe( 'table clipboard', () => {
...
 
+		it( 'should alter model.insertContent if mixed content is pasted (table + empty paragraph)', () => {
+			tableSelection.setCellSelection(
+				modelRoot.getNodeByPath( [ 0, 0, 0 ] ),
+				modelRoot.getNodeByPath( [ 0, 1, 1 ] )
+			);
+
+			const table = viewTable( [
+				[ 'aa', 'ab' ],
+				[ 'ba', 'bb' ] ] );
```
jodator, line 259: code style - AFAICS the `] );` would be always in it own line. (in all tests also)

jodator:
@niegowski there was a comment about missing tests for "integration" test for something like: "p+p+table+p+br" but it got loast AFAICS - could you add it also?

PR #7386, opened 2020-06-05
#7330 Editing part for the linking images feature

### LINKED ISSUE #7330: Allow linking images - editing

Implement link on images editing:

* [x]  add proper schema rules and converters to load / save link on image
    *   the `<a>` element should be inside image's `<figure>` and set on `` element
    *   current converters and its utils should work with that structure (might assume first child of `<figure>` in the view
* [ ]   check `LinkCommand` behavior - it should be tested that this is possible
* [x]   manual tests are welcome (loading data)
* [x]   adjust all image features to new view (like image resize)
* [x] check how image/link behaves when plugin is not enabled
* [x] separate plugin in `ckeditor5-link`.
* [ ] should be enabled in `Essentials` plugin

---

pomek:

Feature (link): Introduced the editing part of linking images feature. Closes #7330.

Other (image): `image/image/utils~getViewImgFromWidget()` will look for nested `` element in the widget.

☝️ I'm not sure it's worth mentioning.

---

Since the PR introduced only a part of the entire plugin (we're still missing UI and Docs), I'd suggest reviewing this code and make it as a base for other parts. New branches will be merged to this one and after finishing the entire plugin, we will be able to merge full stuff to the master.

CI: https://travis-ci.org/github/ckeditor/ckeditor5/builds/694996103


packages/ckeditor5-link/src/linkcommand.js
```diff
@@ -187,18 +187,49 @@ export default class LinkCommand extends Command {
...
+						}
+					}
+
+					if ( isRangeToUpdate ) {
+						rangesToUpdate.push( range );
+					}
+				}
+
+				// And finally we can set the attribute.
+				for ( const elementOrRange of rangesToUpdate ) {
```
jodator, line 224: All of those are ranges so the name might be `range` or `rangeToUpdate`.

packages/ckeditor5-link/src/linkcommand.js
```diff
@@ -187,18 +187,49 @@ export default class LinkCommand extends Command {
...
+				// Ranges that accept the `linkHref` attribute. Since we will iterate over `allowedRanges`, let's clone it.
+				const rangesToUpdate = allowedRanges.slice();
+
+				// For all ranges we want to check whether given range is inside an element that accepts the `linkHref` attribute.
+				// If so, we don't want to propagate applying the attribute to its children.
 				for ( const range of ranges ) {
-					writer.setAttribute( 'linkHref', href, range );
+					let isRangeToUpdate = true;
+
+					for ( const allowedRange of allowedRanges ) {
```
jodator, line 210: Could you extract this as a method, like `isRangeToUpdate( range, allowedRanges )`. It will simplify that loop a bit.

Another thing is that it could even make 202-221 shorter (I didn't test the below though):

```javascript
const rangesToUpdate = [ 
    ...allowedRanges,
    ...ranges.filter( range => isRangeToUpdate( range, allowedRanges ) )
];
```

packages/ckeditor5-link/src/linkimageediting.js
```diff
@@ -0,0 +1,143 @@
...
+		editor.conversion.for( 'downcast' ).add( this._downcastImageLink() );
+	}
+
+	/**
+	 * Returns a converter that consumes the 'href' attribute if a link contains an image.
+	 *
+	 * @private
+	 * @returns {Function}
+	 */
+	_upcastLink() {
```
jodator, line 52: Usually we have them as named functions not class methods so this and \`\_upcastLink\` must follow that.

packages/ckeditor5-link/src/linkcommand.js
pomek: The shorter syntax you proposed looks nice. Unfortunately, it does not work so we need to go with the `for of` loop. After extracting its internals to a separate method it looks even clear.

pomek:
@jodator, I applied your suggestion. Ready for review once again.

packages/ckeditor5-link/src/linkcommand.js
jodator: Yeah, I ddin't test that :D Anyway now looks nicer :)

jodator: Yeah, I ddin't test that :D Anyway now looks nicer :)

Issue #6741, opened 2020-05-05
Improve various navigation scenarios around widgets (mouse, keyboard)


oleq:
There are 3 scenarios we could improve when working on magic block because they add up to the overall user experience of the feature:

### Navigating across widgets using the keyboard

**Case**: `p + widget + widget + p`

**Expected**:

1. A selection in a text
2. <kbd>Arrow down</kbd>: The first widget gets selected.
3. <kbd>Arrow down</kbd>: Magicblock gets activated between the widgets.
4. <kbd>Arrow down</kbd>: The second widget gets selected.
5. <kbd>Arrow down</kbd>: The selection lands in the last paragraph.

**Nice to have**:

1. <kbd>Shift</kbd>+<kbd>Arrow</kbd> – a custom handler.

### Making selection across widgets using the mouse (drag)

ATM the selection postfixer breaks the selection (mostly when created backwards).

![2020-05-04 17 30 55](https://user-images.githubusercontent.com/1099479/81046777-6272b500-8eb9-11ea-8069-a3f0a59cc2c9.gif)

### Vertical widget ↔ nested editable navigation

**Expected**:

1. <kbd>Enter</kbd> – Enters the widget (the selection lands in the very first nested editable).
2. <kbd>Esc</kbd> – Moves the slection from the nested editable to the whole widget.
3. <kbd>Arrows</kbd> – They should work like in table cells when in a nested editable.
4. <kbd>Tab</kbd> – It should move the selection to the next nested editable (and cycle).

Reinmar:
Related tickets:

*   #4374
*   #4590

justinkunz:
+1 on needing a solution for this. Navigation between lists of widgets currently feels broken in cke5

Witoso:
>  Navigation between lists of widgets currently feels broken in cke5

Could you elaborate on what feels broken and what would you expect?

gorzelinski:
A client reported a similar issue. Nested tables behave similarly when trying to select them. Fix would probably help here as well.

https://github.com/user-attachments/assets/7a284d36-a93b-4fc5-bfb3-e991e317701d

PR #7387, opened 2020-06-05
#6693: Keyboard support for the widget type around feature

### LINKED ISSUE #6693: Arrow keys should activate the magic block
It should then become ready (**active**) for inserting content (typing, enter, pasting, etc.), but not be in the model yet to not disrupt track changes or collaboration.

This can be implemented by adding an attribute to the model selection so the model structure is untouched. It comes down to:

    1. A widget is selected (last child).
    2. Arrow down in pressed.
    3. Selection gets an attribute.
    4. Conversion for the selection kicks in.
    5. A late listener removes `.ck-selected` from a widget and adds a class (?) to the block (block becomes active).

### LINKED ISSUE #6694: Typing/Enter/Backspace should replace the active magic block with a paragraph
This should only happen when the block is already active, for instance, when a user activated it using the keyboard.

The paragraph would land in the model and be broadcasted to collaboration, track changes, etc.

### LINKED ISSUE #6825: Consider showing the "insert new paragraph" buttons regardless of widget location

I find situations like the one above confusing:

So, there's no button in this case... why? Because this image is anchored before the block quote.

That made me think a bit and I realised that I actually use \<kbd>Enter\</kbd> to start typing after widgets on many occasions – not only when there's no space after/before that widget.

Assuming that the button will show up only when my mouse pointer is near it, I think that it makes sense to have 2 buttons per every widget in the document. This will actually remove some code and make the editor faster too 😁.

---

oleq:

Fix (engine): The model selection post-fixer should not set a new selection if the ranges before and after post-fixing are the same (see #6693).

Tests (image): Aligned image tests to the latest `WidgetTypeAround` plugin API (see #6693).

Other (paragraph): The `InsertParagraphCommand` should split ancestors of the `Position` to find a parent that allows `'paragraph'` (see #6693).

Tests (table): Aligned `Table` plugin tests to the latest `WidgetTypeAround` API. Code refactoring (see #6693).

Feature (theme-lark): Added styles for a "fake caret" brought by the `WidgetTypeAround` plugin (see #6693).

Feature (typing): Created a public `isSafeKeystroke()` helper (see #6693).

Feature (utils): Created `isArrowKeyCode()`, `getLocalizedArrowKeyCodeDirection()`, and `isForwardArrowKeyCode()` helpers (see #6693).

Feature (widget): Implemented keyboard support for inserting paragraphs around block widgets using a "fake horizontal caret" (`WidgetTypeAround`). Both "Insert new paragraph" buttons are now always displayed for all block widgets regardless of their location in the document. Closes #6693. Closes #6825. Closes #6694.

Also: 

*   code refactoring in the `Widget` plugin,
*   aligned `Widget` tests to the new `WidgetTypeAround` keyboard behavior,
*   moved the `Enter` and `Shift+Enter` support from the `Widget` plugin to the `WidgetTypeAround` plugin.

MINOR BREAKING CHANGE: Removed the `getWidgetTypeAroundPositions()` helper since the "Insert new paragraph" buttons are now visible regardless of the widget location in the document

---

### Additional information

@ckeditor/qa-team Can you take a look at this new feature? It supports arrow keys and enter/backspace/typing when a "fake caret" is displayed (no paste support yet).


FilipTokarski:
### Steps to reproduce:
1. With magic block active, press `enter`
2. Type something
3. Press `arrow up`

### Expected:
Caret moves up, creating "fake horizontal caret" on the widget

### Actual:
Caret jumps above the widget

8_magic_1

*Note: this happens only _from bottom up_. If you type something above the widget and press `arrow down` it works as expected

oleq:
@niegowski I remember you struggled with similar problems when implementing table navigation, right? Do we have some universal mechanism that would help us out? If not, are we tracking it in some issue?

BTW: This probably comes down to #6741.

FilipTokarski:
As it comes to table there are two behaviours, depending on whether the caret is in the middle or at the end of paragraph ( in _top down_ scenario, pressing `arrow down` ):
8_magic_2

FilipTokarski:
Other than whats reported above, everything seems to be ok 👍

niegowski:
> [**@niegowski**](https://github.com/niegowski) I remember you struggled with similar problems when implementing table navigation, right? Do we have some universal mechanism that would help us out? If not, are we tracking it in some issue?

We don't have universal mechanism but we could extract this method to some common utils? https://github.com/ckeditor/ckeditor5/blob/6036d4a4983aba4f5e43704300dc38aeba7369f3/packages/ckeditor5-table/src/tablekeyboard.js#L396

oleq:
@FilipTokarski Can you also check how this new feature forks with Track Changes? Thanks 🙏

FilipTokarski:
It seems to work really good with TC and real time in general. The only thing I found now, that this PR introduces is - TC suggestion does not remove empty paragraph before a widget ( which works fine on CF with CKE5 master and on CF demos ):
  
8_magicTC_1

packages/ckeditor5-widget/tests/widgettypearound/widgettypearound.js
```diff
@@ -250,82 +280,842 @@ describe( 'WidgetTypeAround', () => {
...
+	describe( 'typing around view widgets using keyboard', () => {
+		let model, modelSelection, eventInfoStub, domEventDataStub;
 
-			const viewWidget = viewRoot.getChild( 0 );
+		beforeEach( () => {
+			model = editor.model;
+			modelSelection = model.document.selection;
+		} );
+
+		describe( '"fake caret" activation', () => {
```
niegowski, line 291: There isn't any tests for up/down arrows.

niegowski: Also no test for <kbd>Shift</kbd> + <kbd>Arrow</kbd>

niegowski: Also second step is missing (from fake caret to fully selected widget), as it is tested in "deactivation".

packages/ckeditor5-widget/tests/widgettypearound/widgettypearound.js
```diff
@@ -250,82 +280,842 @@ describe( 'WidgetTypeAround', () => {
...
+					expect( modelSelection.getAttribute( 'widget-type-around' ) ).to.be.undefined;
+
+					fireKeyboardEvent( 'enter' );
+
+					expect( getModelData( model ) ).to.equal( '<paragraph>foo</paragraph><paragraph>[]</paragraph>' );
+					expect( modelSelection.getAttribute( 'widget-type-around' ) ).to.be.undefined;
+				} );
+			} );
+
+			describe( 'on typing an "unsafe" character when the "fake caret" is activated ', () => {
```
niegowski, line 816:
```suggestion
			describe( 'on keydown of a "typing" character when the "fake caret" is activated ', () => {
```

oleq:
> It seems to work really good with TC and real time in general. The only thing I found now, that this PR introduces is - you cannot remove empty paragraph before a widget ( which works fine on CF with CKE5 master and on CF demos )

@FilipTokarski I checked this on CKEditor 5 master and it's the same. The change is not related to this PR. Can you please confirm this?

packages/ckeditor5-widget/tests/widgettypearound/widgettypearound.js
oleq: > Also second step is missing (from fake caret to fully selected widget), as it is tested in "deactivation".

[These are tested by these 2 tests.](https://github.com/ckeditor/ckeditor5/blob/i/6693/packages/ckeditor5-widget/tests/widgettypearound/widgettypearound.js#L567-L615)

FilipTokarski:
🤔  True. IDK what happened, but you are right. The PR is 👍

PR #7391, opened 2020-06-05
#7360: Downcast conversion for markers over table cells

### LINKED ISSUE #7360: Provide marker conversion handling for table cells

I am not sure if it is a new feature or improvement, but...

In short, it would be nice if table cells behave like images or other widgets when it comes to marker conversion. Right now, they behave like paragraphs, that is, if a marker is created on a cell, the conversion happens for the cell contents not for the cell itself. As a result, the text is highlighted, while it would be better if the cell had received proper classes:

Model now:

```
[<tableCell><paragraph>Foo</paragraph></tableCell>]
```

View now:

```
<td><span class="marker">Foo</span></td>
```

View expected:

```
<td class="marker">Foo</td>
```

This way we can do better styling, for example change table border or background.

This is fairly easy to achieve as proper code already exists.

Of course, this is needed to implement collaboration features for multi-cell selection (comments in this case). However, since downcasting for tables is quite complicated, it cannot be easily done on collaboration features side (it is possible, I did it, but it would require tens of lines of two function calls and it would be limited to comments only).

I will create a branch with a proposal for changes for table package.

Related: #7359

niegowski:

Other (table): Markers on the table cells should be visible on the cell level, not only on the cell content. Closes #7360.

---

### Additional information


oleq:
A follow-up #7401.

oleq:
@scofalik I assume this is to enable cell properties in TC. Is there any way to test it? I tried the TC manual test but there's no support for cell properties on master yet.

niegowski:
> [**@scofalik**](https://github.com/scofalik) I assume this is to enable cell properties in TC. Is there any way to test it? I tried the TC manual test but there's no support for cell properties on master yet.

I was testing it here: http://localhost:8125/ckeditor5-engine/tests/manual/highlight.html

PR #7399, opened 2020-06-08
Disable resizing while in read-only mode

### LINKED ISSUE #6457: Image can be resized in read-only mode

1.  Select an image.
2.  Switch to read only mode.

### ✔️ Expected result

You can't resize the image. Probably the resize overlay should be gone.

### ❌ Actual result

You can still resize the image

pomek:

Fix: The resize container will not be displayed while an editor is in read-only mode. Closes #6457.

### DIFF
--- a/packages/ckeditor5-widget/src/widgetresize/resizer.js
+++ b/packages/ckeditor5-widget/src/widgetresize/resizer.js
@@ -116,12 +116,6 @@ export default class Resizer {
 
 				that._domResizerWrapper = domElement;
 
-				that.on( 'change:isEnabled', ( evt, propName, newValue ) => {
-					domElement.style.display = newValue ? '' : 'none';
-				} );
-
-				domElement.style.display = that.isEnabled ? '' : 'none';
-
 				return domElement;
 			} );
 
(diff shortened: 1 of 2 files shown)


pomek:
@ckeditor/qa-team, could you check whether the solution fixes the problem?

pomek:
> I'm confused by this PR. I ran the following scenario on **`master`** and it looks like the [original issue](https://github.com/ckeditor/ckeditor5/issues/6457) is not relevant. What am I missing?

That's a good question because I cannot reproduce the issue as well. And I'm pretty sure I could before I prepared the PR.

FilipTokarski:
I'm not able to reproduce the original issue on docs/master either. You can however force image to resize ( both on master and this PR ) when you start holding the resize handle before `setTimeout()` kicks in:
  
8_resize_1

oleq:
@scofalik Can you confirm this is no longer an issue? I checked back to v15 and the resizer has always disappeared when read-only mode kicks in https://ckeditor5.github.io/docs/nightly/ckeditor5/15.0.0/features/image.html#enabling-image-resizing.

scofalik:
I confirm that the issues does not happen anymore (checked with quite-up-to-date version of CF). However, note, that the selection behaves differently now compared to v15 (in v15 the selection on image disappears when the editor becomes readonly, now it does not disappear). So maybe it didn't work for a while and then it got fixed.

oleq:
@scofalik The question is why it didn't work for you on [March 18th 2020](https://github.com/ckeditor/ckeditor5/issues/6457#issue-583637487)... which is very recent. Do you have any idea? If not, I'm just going to close the PR/issue.

oleq:
Closing the issue as invalid.

PR #7403, opened 2020-06-08
Consider comments when comparing nodes. 

### LINKED ISSUE #5734: Balloon editor crashes on init when there is a comment in source html
Hello!

1. Create editor div with comment in content
```html
<div id="editor">
    <!---->
    <p>text</p>
</div>
```
2. Init editor
```js
BalloonEditor
    .create( document.querySelector( '#editor' ) )
    .catch( error => {
        console.error( error );
    } );
```

### ✔️ Expected result

Editor works

### ❌ Actual result

Exception
```
index.html:15 CKEditorError: unexpected-error {"originalError":{"message":"Cannot read property 'toLowerCase' of undefined","stack":"TypeError: Cannot read property 'toLowerCase' of undefined\n    at Xo (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:159540)\n    at u (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:151031)\n    at $o (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:151116)\n    at Jo._findReplaceActions (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:156788)\n    at Jo._updateChildrenMappings (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:153310)\n    at Jo.render (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:152166)\n    at Ls._render (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:188896)\n    at Ls.<anonymous> (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:185429)\n    at Ls.fire (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:96400)\n    at Ls.change (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:188191)","name":"TypeError"}}
    at Function.rethrowUnexpectedError (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:3355)
    at Ls.fire (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:96637)
    at Ls.change (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:188191)
    at Ls._disableRendering (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:188802)
    at Jc.Aa.listenTo.priority (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:244481)
    at Jc.fire (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:96400)
    at Jc._runPendingChanges (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:327297)
    at Jc.enqueueChange (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:324716)
    at $a.init (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:265927)
    at $a.<anonymous> (https://cdn.ckeditor.com/ckeditor5/15.0.0/balloon/ckeditor.js:5:110855)
```

## 📃 Other details

* Browser: Chrome 77.0.3865.120
* OS: Windows 10
* CKEditor version: 15.0.0

---

Here is repo with reproduction - https://github.com/pogromistik/CKEditor-comment-bug. And you can view result here - https://pogromistik.github.io/CKEditor-comment-bug/.

We are trying to integrate ckeditor into Blazor app and face this problem, cause Blazor inserts comments into html for elements tracking.

It can be related to this change - https://github.com/ckeditor/ckeditor5/issues/3860

tomalec:

Fix (engine): Editor no longer crashes when initialized on HTML with comment. Closes #5734.

---

### Additional information

I was not sure about the location of the tests.


packages/ckeditor5-engine/tests/view/renderer.js
```diff
@@ -305,6 +305,33 @@ describe( 'Renderer', () => {
...
+			domRoot.appendChild( domComment );
+
+			renderer.markToSync( 'children', viewRoot );
+			renderer.render();
+
+			expect( domRoot.childNodes.length ).to.equal( 0 );
+		} );
+
+		// https://github.com/ckeditor/ckeditor5/issues/5734
+		it( 'should replace a comment with the added element', () => {
```
oleq, line 320: TBH I don't understand this description even though I read the test and I know what the issue is about 😛

packages/ckeditor5-engine/src/view/renderer.js
```diff
@@ -930,6 +930,7 @@ function addInlineFiller( domDocument, domParentOrArray, offset ) {
 function areSimilar( node1, node2 ) {
 	return isNode( node1 ) && isNode( node2 ) &&
 		!isText( node1 ) && !isText( node2 ) &&
+		node1.nodeType !== 8 && node2.nodeType !== 8 &&
```
oleq, line 933: `Node.COMMENT_NODE` via https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType.

oleq: Also: `DomCoverter` [already implements `isComment()`](https://github.com/ckeditor/ckeditor5/blob/ad3562a2d3b6d5a1e8de276e7f032371ba260d63/packages/ckeditor5-engine/src/view/domconverter.js#L802-L804). I think it should be extracted to `ckeditor5-utils/src/dom` (like `isNode()`, `isText()` and many others), tested, and then used both in `DomConverter` and `Renderer`.

packages/ckeditor5-engine/tests/view/renderer.js
tomalec: Do you have any alternative in mind?
Is "should remove the comment and add a child element" any better?

packages/ckeditor5-engine/src/view/renderer.js
tomalec: Do we really need to add a dependency and bunch of tests for quite straightforward, native, and atomic comparison: `node.nodeType !== Node.COMMENT_NODE`? To me, it seems like overhead.

oleq: I figured since there are 2 places that would benefit from the new thing, then why not? We have `is*()` helpers for most of the native DOM object types anyway. There are even [tests for it already](https://github.com/ckeditor/ckeditor5/blob/ad3562a2d3b6d5a1e8de276e7f032371ba260d63/packages/ckeditor5-engine/tests/view/domconverter/domconverter.js#L172-L183). 

We could use `Object.prototype.toString.apply( obj ) == '[object Comment]';` to make sure it is document-agnostic (like in `isRange()`).

oleq: Or... to make it faster, we could call `isNode()` first and then check the `#type`.

oleq: ...or not :D It looks like `isNode()` is buggy. A comment does not have a `#defaultView`, so `isNode()` will always return `false` even though `Comment` inherits from `Node`

oleq: Maybe let's have a follow-up for this whole story then 😛

packages/ckeditor5-engine/tests/view/renderer.js
oleq: Nothing particular in my mind. The alternative sounds better, though.

PR #7405, opened 2020-06-09
Keep an empty "alt" attribute in the data

### LINKED ISSUE #5033: The alt attribute should always be output
It's required in the data output. Perhaps it should also be required in the model, even when empty.

pomek:

Fix: The `src` and `alt` attributes for the `image` element will be always added to the editor's data. Even if they are empty. Closes #5033.


packages/ckeditor5-image/src/image/converters.js
```diff
@@ -127,6 +127,9 @@ export function modelToViewAttributeConverter( attributeKey ) {
 
 		if ( data.attributeNewValue !== null ) {
 			viewWriter.setAttribute( data.attributeKey, data.attributeNewValue, img );
+		} else if ( attributeKey === 'alt' && !data.attributeNewValue ) {
+			// The `alt` attribute should always be added. Even empty. See #5033.
+			viewWriter.setAttribute( data.attributeKey, '', img );
```
jodator, line 132: Since the coverage dropped because of that I wonder if

a) we should just keep empty `src` and `alt` on an image. While the empty `alt` is needed the empty `src` looks like a bug.

jodator: After reading [the specs](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-src):

> The src attribute must be present, and must contain a valid non-empty URL potentially surrounded by spaces referencing a non-interactive, optionally animated, image resource that is neither paged nor scripted.

So it looks like the image's src must be present and be a non-empty string (to provide valid HTML). If someone would remove the image's src then we will output invalid HTML anyway but assuming that `src` should be always present me might as well keep both attributes here.

pomek:
@jodator, I've added a missing test. CC should be 100% now.

pomek:
@jodator, ready for review once again.

PR #7408, opened 2020-06-09
Do not show an empty widget toolbar

### LINKED ISSUE #5857: Remove image toolbar for empty configuration

I think that the empty image toolbar shouldn't be displayed:

image

I was testing this in restricted editing mode in which image styles are needed to properly display the side image but the toolbar is not needed in this mode.

Removing this by removing the image toolbar plugin probably would fix that, but maybe the default behavior should be to not display the empty toolbar.

ps.: maybe log some warning also?

### ✔️ Expected result

The empty toolbar is not displayed

### ❌ Actual result

That's that arrow hanging if you select an image.

---

pomek:

Fix: The widget toolbar won't be shown if an empty collection of items was specified in the editor's configuration. Closes #5857.

### DIFF
--- a/packages/ckeditor5-core/tests/_utils-tests/articlepluginset.js
+++ b/packages/ckeditor5-core/tests/_utils-tests/articlepluginset.js
@@ -33,7 +33,12 @@ describe( 'ArticlePluginSet', () => {
 		editorElement = document.createElement( 'div' );
 		document.body.appendChild( editorElement );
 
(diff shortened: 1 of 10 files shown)


jodator:
@pomek could you check errors from the CI - is it related with this PR (other packages might be broken) or do we have a problem with `master`?

pomek:
> is it related with this PR

Yes, it is. I'll fix it.

pomek:
@jodator, should be green now.

PR #7410, opened 2020-06-09
6574: Use TableEditing in tests.

### LINKED ISSUE #6574: The table tests are wrong

This issue comes from time to time - most of the table tests works on `defaultSchema()` and `defaultConversion()` utilities.

Why it is this way - mostly because other features' tests work in detachment from it's `FeatureEditing` part. Examples: [list command](https://github.com/ckeditor/ckeditor5-list/blob/c15ac57483903d0dfa05904fa5292969c67f47c4/tests/listcommand.js#L22-L38), [paragraph command](https://github.com/ckeditor/ckeditor5-paragraph/blob/92b1a8db73c812d7c7f97505033684aff22ab5d2/tests/paragraphcommand.js#L23-L29), etc. However I see that other features tests are written using the `FeatureEditing` so it might be a case that we switched the approach somewhere or we have two.

Rewriting all the tests might be a huge task but I see a potential benefit of having proper tests for table.

Unfortunately, some tests cannot be simply rewritten by removing `defaultConversion()` helpers because the tests assumes other configuration (example - the conversion tests assumes operating on data pipeline converters (without widget styling).

jodator:

Other (table): Use real schema and conversion definitions in table tests. Closes #6574.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/tests/commands/insertcolumncommand.js
+++ b/packages/ckeditor5-table/tests/commands/insertcolumncommand.js
@@ -9,9 +9,11 @@ import { getData, setData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model
 
 import InsertColumnCommand from '../../src/commands/insertcolumncommand';
 import TableSelection from '../../src/tableselection';
-import { assertSelectedCells, defaultConversion, defaultSchema, modelTable } from '../_utils/utils';
+import { assertSelectedCells, modelTable } from '../_utils/utils';
 import TableUtils from '../../src/tableutils';
 import { assertEqualMarkup } from '@ckeditor/ckeditor5-utils/tests/_utils/utils';
+import TableCellPropertiesEditing from '../../src/tablecellproperties/tablecellpropertiesediting';
+import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
 
 describe( 'InsertColumnCommand', () => {
 	let editor, model, command;
@@ -19,14 +21,11 @@ describe( 'InsertColumnCommand', () => {
 	beforeEach( () => {
 		return ModelTestEditor
 			.create( {
-				plugins: [ TableUtils, TableSelection, HorizontalLineEditing ]
+				plugins: [ Paragraph, TableCellPropertiesEditing, TableUtils, TableSelection, HorizontalLineEditing ]
 			} )
 			.then( newEditor => {
 				editor = newEditor;
 				model = editor.model;
-
-				defaultSchema( model.schema );
-				defaultConversion( editor.conversion );
 			} );
 	} );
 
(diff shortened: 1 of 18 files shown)


packages/ckeditor5-table/tests/commands/insertcolumncommand.js
```diff
@@ -5,28 +5,27 @@
...
+import InsertColumnCommand from '../../src/commands/insertcolumncommand';
 
 describe( 'InsertColumnCommand', () => {
 	let editor, model, command;
 
 	beforeEach( () => {
 		return ModelTestEditor
 			.create( {
-				plugins: [ TableUtils, TableSelection, HorizontalLineEditing ]
+				plugins: [ Paragraph, TableCellPropertiesEditing, TableSelection, HorizontalLineEditing ]
```
niegowski, line 24: Why `TableCellPropertiesEditing` is used here instead of `TableEditing` ?

packages/ckeditor5-table/tests/converters/downcast.js
```diff
@@ -5,571 +5,512 @@
...
+					'</tbody>' +
+					'</table>' +
+					'</figure>'
+				);
+			} );
 		} );
+	} );
+
+	describe( 'downcastInsertRow()', () => {
+		// The insert row downcast conversion is not executed in data pipeline.
```
niegowski, line 325: Maybe we should add test to make sure that row conversion is not used in data pipeline?

packages/ckeditor5-table/tests/converters/downcast.js
```diff
@@ -606,132 +549,104 @@ describe( 'downcast converters', () => {
...
-					model = editor.model;
-					doc = model.document;
-					root = doc.getRoot( 'main' );
-					view = editor.editing.view;
-
-					defaultSchema( model.schema );
-					defaultConversion( editor.conversion );
-				} );
-		} );
+		// The insert table cell downcast conversion is not executed in data pipeline.
```
niegowski, line 552: Maybe we should add test to make sure that cell conversion is not used in data pipeline?

packages/ckeditor5-table/tests/commands/insertcolumncommand.js
jodator: Looks like typo :D

packages/ckeditor5-table/tests/converters/downcast.js
jodator: I'm not able to think of such test :) It would be very synthetic. The data pipeline conversion runs on whole, static content it do not convert "dynamic" actions (adding/removing stuff, changing attributes, etc).

packages/ckeditor5-table/tests/converters/downcast.js
jodator: As above.

packages/ckeditor5-table/tests/converters/downcast.js
```diff
@@ -1163,20 +1052,6 @@ describe( 'downcast converters', () => {
...
-		beforeEach( () => {
-			return VirtualTestEditor.create( { plugins: [ Paragraph, TableEditing ] } )
-				.then( newEditor => {
-					editor = newEditor;
-					model = editor.model;
-					doc = model.document;
-					root = doc.getRoot( 'main' );
-					view = editor.editing.view;
-
-					markerConversion( editor.conversion, 'marker' );
```
jodator, line 1176: No expclicit test for `classes` as an array. /cc @niegowski

jodator:
@niegowski the tests from #7391 were also fixed.

PR #7432, opened 2020-06-15
(Un)Link command state when an image is selected

### LINKED ISSUE #7429: LinkCommand#value does not work with `<image>`
After introducing the editing part for linking images, the `link` command should read the `linkHref` attribute set on the `<image>` element.

image

image

pomek:

Internal: `#value` and `#isEnabled` properties for `LinkCommand` and `UnlinkCommand` work properly with images. Closes #7429.

### DIFF
--- a/packages/ckeditor5-link/src/linkcommand.js
+++ b/packages/ckeditor5-link/src/linkcommand.js
@@ -62,7 +62,7 @@ export default class LinkCommand extends Command {
 
 		// A check for the `LinkImage` plugin. If the selection contains an element, get values from the element.
 		// Currently the selection reads attributes from text nodes only. See #7429.
-		if ( selectedElement && selectedElement.name === 'image' ) {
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-link/src/linkcommand.js
```diff
@@ -57,13 +58,21 @@ export default class LinkCommand extends Command {
 		const model = this.editor.model;
 		const doc = model.document;
 
-		this.value = doc.selection.getAttribute( 'linkHref' );
+		const selectedElement = first( doc.selection.getSelectedBlocks() );
+
+		// A check for the `LinkImage` plugin. If the selection contains an element, get values from the element.
+		// Currently the selection reads attributes from text nodes only. See #7429.
+		if ( selectedElement && selectedElement.name === 'image' ) {
```
jodator, line 65:
```suggestion
		if ( selectedElement && selectedElement.is( 'image' ) ) {
```

PR #7440, opened 2020-06-16
Typing around the link element

### LINKED ISSUE #1016: It's hard to escape from link areas
🆕 Feature request

## 💻 Version of CKEditor
ckeditor5/10.0.0
classic build

## 📋 Steps to reproduce

1. type a text
2. mark the last word and make it a link\
3. continue typing
4. now we are still in the link  and need to type right 2 times to escape of the link

## ✅ Expected result
if i continue to type i would not expect to enlarge the link area i just defined. It would be totally o.k. to redo the link actions if i need to change this.

## ❎ Actual result
I'm prisoner of a link and had much difficulties to find out how to escape.

none... just change it please. And thank you for your great work!

pomek:

Other (link): The selection after inserting a link will land after the inserted element. Thanks to that a user will be able to type directly after the link without extending the link element. Closes #1016.

Other (link): After clicking at the beginning or end of the link element, the selection will land before/after the clicked element. Thanks to that a user will be able to typing before or after the link element as normal text without extending the link. See #1016.

### DIFF
--- a/packages/ckeditor5-link/src/linkcommand.js
+++ b/packages/ckeditor5-link/src/linkcommand.js
@@ -172,8 +172,8 @@ export default class LinkCommand extends Command {
 						writer.removeAttribute( item, linkRange );
 					} );
 
-					// Create new range wrapping changed link.
-					writer.setSelection( linkRange );
+					// Put the selection at the end of the updated link.
+					writer.setSelection( writer.createPositionAfter( linkRange.end.nodeBefore ) );
 				}
 				// If not then insert text node with `linkHref` attribute in place of caret.
 				// However, since selection in collapsed, attribute value will be used as data for text node.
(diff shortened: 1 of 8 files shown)


pomek:
On my local env, two tests fail:

```
FAILED TESTS:
  LinkUI
    link form view
      link protocol
        ✖ should use the "http://" protocol when it's configured
          Chrome 83.0.4103 (Mac OS X 10.14.6)
        AssertionError: expected 'ckeditor.com' to equal 'http://ckeditor.com'
      + expected - actual
      -ckeditor.com
      +http://ckeditor.com
    at eval (webpack:///./packages/ckeditor5-link/tests/linkui.js?:1007:57)
        ✖ should use the "http://" protocol when it's configured and form input value contains "www."
          Chrome 83.0.4103 (Mac OS X 10.14.6)
        AssertionError: expected 'www.ckeditor.com' to equal 'http://www.ckeditor.com'
      + expected - actual
      -www.ckeditor.com
      +http://www.ckeditor.com
    at eval (webpack:///./packages/ckeditor5-link/tests/linkui.js?:1018:57)
```

Let's see what Travis says about it.

pomek:
Same on CI.

jodator:
@pomek Yeah. It looks like the tests should be written differently. I've fixed that.

Issue #7459, opened 2020-06-18
Unexpected type with link on keyboard move


jodator:
## 📝 Provide detailed reproduction steps (if any)

1.  Insert `<softBreak>` after a link.
2.  Move selection somewhere else.
3.  Move selection back after the `<softBreak>`.

### ✔️ Expected result

You should type without link (the `linkHref` attribute is not set on selection).

### ❌ Actual result

You can type with link.

![](https://user-images.githubusercontent.com/247363/85034059-f787ef80-b181-11ea-81ee-07e2815d54e3.gif)

## 📃 Other details

*   CKEditor version: latest master

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Reinmar:
Support for soft breaks was added after the support for two-step caret movement and quite recently in general, so I'd guess we didn't test the editor for this.

Reinmar:
Sorry, I got it now – it only happens when there's no text after the caret. This is a bit special case. Perhaps it's a symmetrical behavior to this:

![](https://user-images.githubusercontent.com/156149/85105487-dfa87e00-b20a-11ea-9e6f-ca5f56bc1259.png)

Note that if there was a text before the caret then you wouldn't have bold enabled:

![](https://user-images.githubusercontent.com/156149/85105530-edf69a00-b20a-11ea-90c9-8145812cb0bc.png)

So, in some cases, the gravity automatically changes to right. Perhaps it somehow leaks to the case with a soft break. As I mentioned, soft breaks were added recently and for a long time we didn't have support for inline elements so I guess we did consider such a case where an inline element needs to stop the search for a neighbour text node.

Reinmar:
BTW, this doesn't seem to be a regression. I checked it on https://ckeditor.com/docs/ckeditor5/15.0.0/examples/builds/classic-editor.html and it occurred there too.

jodator:
Might be a case that searching for attributes does not stop on inline objects.

Issue #1576, opened 2019-02-22
The shift+arrow moves the selection to the start of the document if the caret is before softBreak


jodator:
## Is this a bug report or feature request? (choose one)

🐞 Bug report 

## 💻 Version of CKEditor

latest master

## 📋 Steps to reproduce

1. split some block (ie `<heading1>`)  with <kbd>enter</kbd> (probably not necessary step)
2. Insert soft break (<kbd>shift</kbd>+<kbd>enter</kbd>)
3. Try to extend selection with keyboard by <kbd>shift</kbd>+<kbd>arrow right</kbd>

## ✅ Expected result

The soft break is selected

## ❎ Actual result

The selection jumps to previous block

## 📃 Other details that might be useful

Happens on Chrome:

![peek 2019-02-22 17-08](https://user-images.githubusercontent.com/247363/53255035-1036e100-36c5-11e9-9c88-dc2b19ee837a.gif)

:point_up: in the screen cast I repeat the <kbd>shift</kbd>+<kbd>arrow down</kbd> in other places to showcas how it behaves.

While Firefox it is OK:

![peek 2019-02-22 17-10](https://user-images.githubusercontent.com/247363/53255038-11680e00-36c5-11e9-8349-0249eb0d43d9.gif)

Reinmar:
The critical part in the TC is that the softbreak must be at the beginning of a block element.

jodator:
@Mgsy  found similar case: https://github.com/ckeditor/ckeditor5/pull/7478#issuecomment-648645026 undo of enqueued auto-link aftera soft-break. Works on Firefox.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity over the last year. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #7478, opened 2020-06-22
4715: Autolink feature

### LINKED ISSUE #4715: Implement autolinking feature
Create autolinking mechanism as separate feature. After new content is inserted to the editor, editor should detect matched UTL and automatically transforms it to link.

In addition to URLs it might also support email addresses.

jodator:

Feature (link): Added the `AutoLink` feature which replaces a plain text with a link if typed or pasted content is the link. Closes #4715.

---

### Additional information

*   I'm not sure how verbose the feature guide should be.
*   I've might forget something due to various distractions so sorry if found something serious :(


jodator:
Hey, @ckeditor/qa-team - there's a new feature to check - `AutoLink`. It has its own manual test and section in Link feature guide.

cc @Reinmar since Olek is out could you take a quick look at the feature guide:

pomek:
Test case:

1. Use [all-features](https://github.com/ckeditor/ckeditor5/blob/master/tests/manual/all-features.js) sample. Add the `AutoLink` plugin.
2. Clear the editor.
3. Type: `Type whatever.`
4. Each word is treated as link.

image

The same result I've got in the [autolink MT](http://localhost:8125/ckeditor5-link/tests/manual/autolink.html).

packages/ckeditor5-link/src/autolink.js
```diff
@@ -0,0 +1,207 @@
...
+
+// Simplified email test - should be run over previously found URL.
+const EMAIL_REG_EXP = /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$/i;
+
+/**
+ * The auto link plugin.
+ *
+ * @extends module:core/plugin~Plugin
+ */
+export default class AutoLink extends Plugin {
```
pomek, line 37: The plugin uses a few things that are provided by other plugins and as I can see, the dependencies are not specified:

* `editor.commands.get( 'enter' )` - what if missing? I know, little chance for such case but still...
* `const shiftEnterCommand = editor.commands.get( 'shiftEnter' );` - the same as above,
* `writer.setAttribute( 'linkHref', linkHrefValue, range );` - missing schema definition or `LinkEditing` which defines it.

packages/ckeditor5-link/src/autolink.js
```diff
@@ -0,0 +1,207 @@
...
+const URL_REG_EXP = new RegExp(
+	// Group 1: Line start or after a space.
+	'(^|\\s)' + // Match .
+	// Group 2: Full detected URL.
+	'(' +
+		// Group 3 + 4: Protocol + domain.
+		'(([a-z]{3,9}:(?:\\/\\/)?)(?:[\\w]+)?[a-z0-9.-]+|(?:www\\.|[\\w]+)[a-z0-9.-]+)' +
+		// Group 5: Optional path + query string + location.
+		'((?:\\/[+~%/.\\w\\-_]*)?\\??(?:[-+=&;%@.\\w_]*)#?(?:[.!/\\\\\\w]*))?' +
+	')$', 'i' );
```
pomek, line 25: Something is wrong in the regexp - https://regex101.com/r/C9sXrU/1.

packages/ckeditor5-link/tests/autolink.js
```diff
@@ -0,0 +1,225 @@
...
+
+		for ( const unsupportedURL of unsupportedURLs ) {
+			it( `should not detect "${ unsupportedURL }" as a valid URL`, () => {
+				simulateTyping( unsupportedURL + ' ' );
+
+				expect( getData( model ) ).to.equal(
+					`<paragraph>${ unsupportedURL } []</paragraph>` );
+			} );
+		}
+	} );
```
pomek, line 147: I miss a test where you check what will happen after typing a text. Not a link, not an email. Just a text.

packages/ckeditor5-link/tests/manual/autolink.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/* globals console:false, window, document */
```
pomek, line 6:
```suggestion
/* globals console, window, document */
```

pomek:
From CI:

```
Checking dependencies...
Found some issue with dependencies.
Missing devDependencies:
- @ckeditor/ckeditor5-code-block

💥 link have a dependency problem 💥
```

pomek:
I've updated the commit/merge message.

packages/ckeditor5-link/tests/autolink.js
jodator: Done.

packages/ckeditor5-link/src/autolink.js
jodator: Ugh - I've must fail while tuning the RegExp. I've used different one.

packages/ckeditor5-link/src/autolink.js
jodator: I've added a tests for `Enter` and `ShiftEnter` - I've forgot about that.

The feature is build similarly to `AutoMediaEmbed` - it assumes `MediaEmbedEditing` to be present. It uses `linkHref` as `editor.execute( 'link' )` does not accept range :/

jodator: Added here: #7490.

jodator:
@pomek I think I've addressed all the issues.

Mgsy:
## Steps to reproduce

1.  Paste a link.
2.  Press Shift + Enter.
3.  Undo.
4.  Press Enter.

## Expected result

A text is converted to a link.

## Current result

A text remains as a plain text.

## GIF

jodator:
> A text remains as a plain text.

I think that the selection remains after `<br>` so the second enter is executed on:

```
https://google.pl<br>
[]
```

In that case it works OK.

Do you recall a bug with rendering selection after <kbd>shift</kbd>+<kbd>enter</kbd>? I think that I saw something like that.

jodator:
@Mgsy - I've found something similar: #1576 looks like Chrome bug.

Mgsy:
> I think that the selection remains after `<br>` so the second enter is executed on:

True, now I see that the selection doesn't change - it stays after `<br>`. Only the caret renders in incorrect place.

jodator:
Yep, FF works fine:

jodator:
@pomek two more things to add. I've forgot about disabling this.

jodator:
@pomek OK I've fixed the issues with disabling `AutoLink` feature in code blocks and when using `forceDisabled()`.

Mgsy:
## Steps to reproduce

1.  Paste a link.
2.  Press Shift + Enter.
3.  Type something.

## Current result

A user types with `linkHref` attribute.



## Expected result

A user types without `linkHref` attribute, just like after inserting a link standard way.

jodator:
@Mgsy dup of: #7459 ?

Mgsy:
> [**@Mgsy**](https://github.com/Mgsy) dup of: [#7459](https://github.com/ckeditor/ckeditor5/issues/7459) ?

Yes, it looks like this is the same issue :)

Mgsy:
I've tested it in various browsers and the feature looks good :+1:

jay2503:
is this feature released with `v20.0.0`? I could not find autolink.js under source with the latest installation.

jodator:
@jay2503 - it will be available in the upcoming release. If everything [goes well](https://github.com/ckeditor/ckeditor5/issues/7479) it should be released next week.

PR #7485, opened 2020-06-23
Make the CI perform full docs build each time

### LINKED ISSUE #6788: CI should build entire docs

There was a bug on master in the docs that I fixed in d08ce3e. However, the CI for the previous commit passed, which probably means that we don't run entire documentation build on CI.

---

mlewand:

Internal: The CI now performs full documentation build for each run. Closes #6788.

---

### Additional information

Strict flag causes umberto to return non-zero CLI code if there's a warning during api building. An example can be seen here: https://travis-ci.org/github/ckeditor/ckeditor5/builds/701048663


mlewand:
Actually I have removed travis wait all together - as it should not take a long time to build.

mlewand:
(oops wrong button, reopening :smile:)

PR #7506, opened 2020-06-26
Change -engine/bindTwoStepCaretToAttribute util to -typing/TwoStepCaretMovement plugin.

### LINKED ISSUE #7444: Move 2-step caret movement to -typing and refactor
_This is a result of a meeting after the review of https://github.com/ckeditor/ckeditor5/pull/7356_

Changes from https://github.com/ckeditor/ckeditor5/pull/7356 shown that current implementation of 2SCM does not work well with multiple attributes, and needs some capability to synchronize gravity between attributes. At the F2F with @Reinmar, @oleq and @niegowski we agreed it requires some refactoring.

We should start with/consider:
1. moving it to `-typing`,
2. make a plugin,
3. Consider having the gravity set per attribute.

We also agreed, there should be only 2 movements required for multiple attributes ending in the same position. E.g. "some <a href="#foo">linking <code>code</code></a>[]" + <kbd>←</kbd><kbd>←</kbd>, results in "some <a href="#foo">linking <code>code[]</code></a>". Meaning stepping inside steps into the deepest attribute at given position. It is to minimize the overhead of additional <kbd>←</kbd>s needed for navigation, simplify the behavior and implementation, especially given, looking at the model with a flat structure of attributes, it's hard to identify the depth of nesting.

## 📃 Other details

* Browser: n/a
* OS: n/a
* CKEditor version: master @
* Installed CKEditor plugins: link, code, engine

## Related issues

- https://github.com/ckeditor/ckeditor5/issues/4341 - idea to rething 2SCM,
- https://github.com/ckeditor/ckeditor5/issues/6722 - add 2SCM to the `code`,
- https://github.com/ckeditor/ckeditor5/issues/4361 - 2SCM doesn't work with `<br>`,
- https://github.com/ckeditor/ckeditor5/issues/4528 - overriding gravity breaks RTL nav

---

tomalec:

Other (engine, typing): The engine's util `bindTwoStepCaretToAttribute` is now typing's plugin TwoStepCaretMovement . Closes #7444.

MINOR BREAKING CHANGE (engine, typing): Package and API for `bindTwoStepCaretToAttribute(…)` changed to `editor.plugins.get( TwoStepCaretMovement ).registerAttribute( 'a' )`

---

### Additional information

I'd also move highlighting code a part of this plugin (currently [hard-coded in `Link`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-link/src/linkediting.js#L215)), but I'd do that in a separate PR, for an easier review.

I will add more tests for multi-attribute scenarios, in a PR that adds 2SCM to the `Code` plugin.

I was thinking about adding the `unregisterAttribute` method for cleanup and scenarios when somebody would like to disable 2SCM for `Link`. Alternatively, we could make a more breaking change and disable 2SCM for Link and require setting it up in config, by just adding 2SCM plugin explicitly.


packages/ckeditor5-typing/src/twostepcaretmovement.js
```diff
@@ -4,27 +4,30 @@
...
- * attribute on arrow right (<kbd>→</kbd>) and left (<kbd>←</kbd>) key press.
+ * This plugin enables the two-step caret (phantom) movement behavior for
+ * {@link module:typing/twostepcaretmovement~TwoStepCaretMovement#registerAttribute registered attributes}
+ * on arrow right (<kbd>→</kbd>) and left (<kbd>←</kbd>) key press.
  *
  * Thanks to this (phantom) caret movement the user is able to type before/after as well as at the
  * beginning/end of an attribute.
  *
- * **Note:** This helper support right–to–left (Arabic, Hebrew, etc.) content by mirroring its behavior
+ * **Note:** This class support right–to–left (Arabic, Hebrew, etc.) content by mirroring its behavior
```
jodator, line 23:
```suggestion
 * **Note:** This plugin support right–to–left (Arabic, Hebrew, etc.) content by mirroring its behavior
```

packages/ckeditor5-typing/src/twostepcaretmovement.js
```diff
@@ -80,70 +83,116 @@ import priorities from '@ckeditor/ckeditor5-utils/src/priorities';
...
+		const editor = this.editor;
+		const model = editor.model;
+		const view = editor.editing.view;
+		const locale = editor.locale;
 
-		const position = modelSelection.getFirstPosition();
-		const contentDirection = locale.contentLanguageDirection;
-		let isMovementHandled;
+		// const twoStepCaretHandler = new TwoStepCaretHandler( model, emitter, attribute );
+		const modelSelection = model.document.selection;
```
jodator, line 120:
```suggestion
		const modelSelection = model.document.selection;
```

left over to rempve ;)

packages/ckeditor5-typing/src/twostepcaretmovement.js
```diff
@@ -80,70 +83,116 @@ import priorities from '@ckeditor/ckeditor5-utils/src/priorities';
...
+		// * "+1" because we would like to avoid collisions with other features (like Widgets), which
+		// take over the keydown events with the "high" priority. Two-step caret movement takes precedence
+		// over Widgets in that matter.
+		//
+		// Find out more in https://github.com/ckeditor/ckeditor5-engine/issues/1301.
+		this.listenTo( view.document, 'keydown', ( evt, data ) => {
+			// Do nothing if the plugin is disabled.
+			// if ( !this.isEnabled ) {
+			// 	return;
+			// }
```
jodator, line 137: Another lef-over. This one makes me think if we need to enable/disable this plugin. On the first thought I think not - as this is purely navigational plugin but I don't have a strong opinion here.

packages/ckeditor5-typing/src/twostepcaretmovement.js
```diff
@@ -80,70 +83,116 @@ import priorities from '@ckeditor/ckeditor5-utils/src/priorities';
...
+	 * @param {String} attribute Name of the attribute to handle.
+	 */
+	registerAttribute( attribute ) {
+		this._handlers.set(
+			attribute,
+			new TwoStepCaretHandler( this.editor.model, this, attribute )
+		);
+	}
 }
+export default TwoStepCaretMovement;
```
jodator, line 192: We tend to keep it together with `class` declaration.

tomalec: But then you do not have the named export, which is usually my personal preference, but if the convention is to always use default imports, then, I'll change it.

packages/ckeditor5-typing/src/twostepcaretmovement.js
tomalec: I think so as well, but I'd like to do that in the next PR, to limit functional changes in this one.

tomalec:
0. 
> It's not entirely clear for me if this should be handled in this PR:
>
> > We also agreed, there should be only 2 movements required for multiple attributes

I'd cover it in the next PR, this one is only to refactor, without changing functionality, except making it a plugin.

----

> a) ticket `1301` manual tests is not working for me

Checking...

> b) two strikes required to move

That's 0., isn't it? If so, I'll cover that in the next PR

packages/ckeditor5-typing/src/twostepcaretmovement.js
jodator: The convention is:

```javascript
export default class TwoStepCareMovement { ... }
```

not the named export.

tomalec:
I reproduced a)
It's FF issue, that also happens on `master` - I'd consider it out of the scope of this PR.

packages/ckeditor5-typing/src/twostepcaretmovement.js
tomalec: addressed

packages/ckeditor5-typing/src/twostepcaretmovement.js
tomalec: Comment removed.

pomek:
> MINOR BREAKING CHANGE (engine, typing): Package and API for bindTwoStepCaretToAttribute(…) changed to editor.plugins.get( TwoStepCaretMovement ).registerAttribute( 'a' )

The changelog generator will handle only the first listed package in the scoped BC.

PR #7508, opened 2020-06-26
#6639: Restoring the document selection to the ranges as they were before undoing table cells merge should not crash the editor.

### LINKED ISSUE #6639: Problem with restoring selection upon merging cells with ranges on them.

The bug occurred here: #6634. The steps are (in terms of operations) setting selection on two cells and merging them (moving contents of second to the end of first and removing second).

```js
// Move contents of cell 2 to cell 1
new MoveOperation(
	new Position( root, [ 0, 0, 1, 0 ], 'toNext' ),
	1,
	new Position( root, [ 0, 0, 0, 1 ] )
);
// Remove cell 2
new MoveOperation(
	new Position( root, [ 0, 0, 1 ], 'toNext' ),
	1,
	new Position( gy, [ 0 ] )
);
// Undo:
// Reinsert cell 2:
new MoveOperation(
	new Position( gy, [ 0, 'toNext' ] ),
	1,
	new Position( root, [ 0, 0, 1 ] )
);
// Move half previous content of cell 1 from cell 1 to cell 2
new MoveOperation(
	new Position( root, [ 0, 0, 0, 1 ], 'toNext' ),
	1,
	new Position( root, [ 0, 0, 1, 0 ] )
);
```

The problem with the above was that there were two ranges set in the model:

1. On cell 1: `[0,0,0]-[0,0,1]`
1. On cell 2: `[0,0,1]-[0,0,2]`

Now, the undo restores the selection it transforms `[0,0,0]-[0,0,1]` and `[0,0,1]-[0,0,2]` selection ranges by transforming it by the above operations.

For the first range we get 2 ranges in return:

1. For range `[0,0,0]-[0,0,1]` we get `[0,0,0]-[0,0,1]` and `[0,0,1,0]-[0,0,1,1]`.
2. For range `[0,0,1]-[0,0,2]` we get `[0,0,1]-[0,0,2]` and `[0,0,1,0]-[0,0,1,1]` (:point_left: wat).

so the base command algorithm fixes the first range to `[0,0,0]-[0,0,1,1]` (one range) which is invalid because it intersects with other transformed range `[0,0,1]-[0,0,2]`.

---

For me, it looks like the range transformation is producing "dangling" ranges which aren't desired _in this situation_. 

The problem could be fixed on two levels: undo and operations. 

In undo it might be somehow doable - probably adressing it in [`transformSelectionRange()`](https://github.com/ckeditor/ckeditor5-undo/blob/6ce19855746ff88754906b67142037dfd1542b81/src/basecommand.js#L172-L194).

On operations level - I don't have the knowledge for now. It looks like that [`Range._getTransformedByMove()`](https://github.com/ckeditor/ckeditor5-engine/blob/029e8b26b420365165e16b2ccf5626e245d7065a/src/model/range.js#L750-L824) is producing invalid ranges. But, as I don't really know what should happen, it might require a bigger effort to fix.

---

niegowski:

Other (table): Restoring the document selection to the ranges as they were before undoing table cells merge. Closes #6639.

Other (engine): Added `Range#getJoined()` method for joining `Range`s.

Internal (table): Removed workaround for `"model-selection-range-intersects"` errors.

Internal (undo): Fixed normalization of restored selection ranges.

---

### Additional information

This is only fixing issues with undo. `LiveRanges` can still lead to unexpected results.


packages/ckeditor5-table/src/commands/removerowcommand.js
```diff
@@ -63,10 +63,7 @@ export default class RemoveRowCommand extends Command {
 		// Use single batch to modify table in steps but in one undo step.
 		const batch = model.createBatch();
 
-		model.enqueueChange( batch, writer => {
-			// This prevents the "model-selection-range-intersects" error, caused by removing row selected cells.
-			writer.setSelection( writer.createSelection( table, 'on' ) );
-
+		model.enqueueChange( batch, () => {
```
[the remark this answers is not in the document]
niegowski: Yes, it's not related. `enqueueChange` is needed due to downcast conversion of `headingRows` attribute into separate view elements ( `thead`, `tbody`)

packages/ckeditor5-engine/src/model/range.js
```diff
@@ -276,6 +276,63 @@ export default class Range {
...
+	 *		);
+	 *		transformed = range.getSum( otherRange ); // range from [ 2, 7 ] to [ 5 ]
+	 *
+	 * @param {module:engine/model/range~Range} otherRange Range to sum with.
+	 * @param {Boolean} [loose=false] Whether the sum intersection check is loose or strict. If the check is strict (`false`),
+	 * summed range is tested for intersection only. If the check is loose (`true`), compared range is also checked if it's "touching"
+	 * current range.
+	 * @returns {module:engine/model/range~Range|null} A sum of given ranges or `null` if ranges have no common part.
+	 */
+	getSum( otherRange, loose = false ) {
```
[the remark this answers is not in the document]
niegowski: I was just trying to be consistent with `getDifference`

packages/ckeditor5-undo/src/basecommand.js
```diff
@@ -93,15 +93,25 @@ export default class BaseCommand extends Command {
...
 
 		// Transform all ranges from the restored selection.
-		for ( const range of ranges ) {
-			const transformed = transformSelectionRange( range, operations );
+		const transformedRangeGroups = ranges.map( range => range.getTransformedByOperations( operations ) );
+		const allRanges = transformedRangeGroups.flat();
+
+		for ( const rangeGroup of transformedRangeGroups ) {
+			// While transforming there could appear ranges that are contained by other ranges, we shall ignore them.
+			const transformed = rangeGroup.filter( range => !allRanges.some( otherRange => (
```
scofalik, line 101: I don't like this part of code, because having `.fliter()` then inside `.some()`, which operate on similar but different data, it seems clunky and reads bad. Can we do that in a different way?

scofalik: Or have the inside function more clear?

scofalik:
It seems that now the only usage of `getJoined()` is with loose flag set to `true`. Also, it is intuitive that the ranges would be joined at least if their boundaries are equal. I'd re-think if we should keep this parameter or make such joining a default behavior. @Reinmar

scofalik:
Other than that the PR is fine with me.

Reinmar:
> I'd re-think if we should keep this parameter or make such joining a default behavior. @Reinmar

:+1: for making it the default behavior and removing this param, unless we can see an obvious case where we'd benefit from having this option.

scofalik:
@Reinmar After giving it a second thought, I came to the conclusion that joining when the ranges are touching is not intuitive and I would not expect that. However, I'd expect that they would be joined if boundaries are exact.

If we want to have this method in the API I think that it should work intuitively. I asked @niegowski to re-introduce the `loose` parameter again. However, I asked him to make joining on exact boundaries a default behavior.

Reinmar:
:+1:  Since my previous comment was strictly related to the approach YAGNI, not to what this function should do, I'm fine with any changes.

packages/ckeditor5-engine/src/model/range.js
```diff
@@ -276,6 +276,63 @@ export default class Range {
...
+	 *		otherRange = model.createRange(
+	 *			model.createPositionFromPath( root, [ 3 ] ),
+	 *			model.createPositionFromPath( root, [ 5 ] )
+	 *		);
+	 *		transformed = range.getJoined( otherRange ); // range from [ 2, 7 ] to [ 5 ]
+	 *
+	 * @param {module:engine/model/range~Range} otherRange Range to be joined.
+	 * @param {Boolean} [loose=false] Whether the intersection check is loose or strict. If the check is strict (`false`),
+	 * ranges are tested for intersection or whether start/end positions are equal. If the check is loose (`true`),
+	 * compared range is also checked if it's "touching" current range.
```
scofalik, line 304: I'd link "touching" to Position#isTouching

niegowski: Done

PR #7509, opened 2020-06-26
Add indent block support to ClassicEditor

deadlyicon:
[Here is a codepen](https://codepen.io/deadlyicon/pen/MWKvoON?editors=1111) of block indent not working in the ClassicEditor


Reinmar:
Hi! Thanks for the PR. We don't plan to change the default setups of the builds, though. Block indentation is a presentational feature and the classic build is focused on semantic content.

PR #7544, opened 2020-07-03
Add Japanese translation for font package and update the Japanese lan…

M1nhNV:
The project I am working on has using Japanese. But the current font package does not. So I added. The translation has been translated by a native speaker, so you can be assured of its quality.
Screen Shot 2020-07-03 at 1 03 59 PM


Reinmar:
Hi! Thanks for the PR, but unfortunately, just like the POT file states we can only accept translations via Transifex: 

```
#         To submit your translations, visit https://www.transifex.com/ckeditor/ckeditor5.
# 
#                   To learn more, check out the official contributor's guide:
#     https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/contributing.html
```

PR #7546, opened 2020-07-03
Move attribute element highlight code from link to typing

tomalec:
Make the code less link-specific, so it could be used for other attributes.

Other: Link's attribute element highlight is now a public utility.

---

### Additional information

This is to address #7444 and #6722.

I do not like the redundancy in tests for highlighting. I copied the set of tests from `linkediting` to `inlineHighlight`. I wanted to remove part of them from `linkEditing` and just check some high-level behavior and that `link` uses `inlineHighlight`. Unfortunately, it's also not a perfect approach, see https://cksource.slack.com/archives/C03URJ97S/p1593637492006600

I'd leave it as it is, just to keep code coverage happy, and move on with the functional change.


packages/ckeditor5-typing/src/utils/inlinehighlight.js
```diff
@@ -0,0 +1,78 @@
...
+ * * The class is added in the view post fixer, after other changes in the model tree were converted to the view.
+ *
+ * This way, adding and removing the highlight does not interfere with conversion.
+ *
+ * @param {module:core/editor/editor~Editor} editor The editor instance.
+ * @param {String} attributeName The attribute name to check.
+ * @param {String} tagName The tagName of a view item.
+ * @param {String} className The class name to apply in the view.
+ */
+export default function setupLinkHighlight( editor, attributeName, tagName, className ) {
```
jodator, line 29: Minimal sample need.

packages/ckeditor5-typing/src/utils/findattributerange.js
```diff
@@ -4,38 +4,44 @@
...
 
 /**
- * @module link/findlinkrange
+ * @module typing/utils/findattributerange
  */
 
 /**
- * Returns a range containing the entire link in which the given `position` is placed.
+ * Returns a range containing the entire {@link module:engine/view/attributeelement~AttributeElement#childCount attribute element}
+ * in which the given `position` is placed.
```
jodator, line 12: This statement is unfortunate. This will return `model` range and we do not have `AttributeElement` in model. Also, the link points to `#childCount`.

packages/ckeditor5-typing/src/utils/inlinehighlight.js
jodator: And wrong function name.

packages/ckeditor5-typing/src/utils/findattributerange.js
```diff
@@ -4,38 +4,44 @@
...
  *
  * @param {module:engine/model/position~Position} position The start position.
- * @param {String} value The `linkHref` attribute value.
+ * @param {String} attributeName The attribute name.
+ * @param {String} value The attribute value.
  * @returns {module:engine/model/range~Range} The link range.
  */
-export default function findLinkRange( position, value, model ) {
-	return model.createRange( _findBound( position, value, true, model ), _findBound( position, value, false, model ) );
+export default function findAttributeRange( position, attributeName, value, model ) {
```
jodator, line 22: I'm thinking of an API of this util. Looking at it's usages:

```javascript
const position = selection.getFirstPosition();
// ...
const range = findAttributeRange( position, 'foo', selection.getAttribute( 'foo' ), model );
```

The above IMO have 1 extra parameter that is not necessary if not 2.

```
findLinkRangeFromSelection( 'foo', selection, model ); // or even
findLinkRangeFromSelection( 'foo', model ); // But the 3 param is kinda better.
```

/cc @Reinmar

packages/ckeditor5-typing/src/utils/inlinehighlight.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+import findAttributeRange from './findattributerange';
+
+/**
+ * @module typing/utils/inlinehighlight
+ */
+
+/**
+ * Adds a visual highlight style to an attribute element in which the selection is anchored.
```
jodator, line 13: Not entirely true - there might be multiple attribute elements. But I don't have a better wording for this :/

packages/ckeditor5-typing/src/utils/inlinehighlight.js
jodator: I think that the naming of the module/method must be improved. The default export should be named after the module AFAIR. Also it is used (in other PR) under totally different name and all of this is confusing.

packages/ckeditor5-typing/src/utils/findattributerange.js
Reinmar: Shouldn't it be more like this?

```javascript
findAttributeRange( model, positionToStartSearchFrom, attributeName, attributeValue )
```

Potentially, attributeValue is not necessary as it can be read at the initial position.

packages/ckeditor5-typing/src/utils/inlinehighlight.js
jodator: Another thing is mabye we should also expose this on `TwoStepCaretMovement` plugin?:

https://github.com/ckeditor/ckeditor5/pull/7577/files#diff-0efb47e00de0c1be680392e506f94f39R68-R71

One things is imported and the second is used from a plugin.

packages/ckeditor5-typing/src/utils/findattributerange.js
jodator: I'd ditch the `attributeValue` - it is always taken from the selection anyway as the position is.

Reinmar: > Potentially, attributeValue is not necessary as it can be read at the initial position.

But it may be actually useful as you may be here: `<$text x=1>aa</$text>[]<$text x=2>bb</$text>` and if you could execute this function with `1` or `2` you'd have a better flexibility.

Reinmar: > Potentially, attributeValue is not necessary as it can be read at the initial position.

But it may be actually useful as you may be here: `<$text x=1>aa</$text>[]<$text x=2>bb</$text>` and if you could execute this function with `1` or `2` you'd have a better flexibility.

Reinmar: BTW, couldn't this method land in one of the model's classes? Maybe a static TreeWalker's method to not mix the layers?

jodator: Sure but the usage isn't that flexible - the question is do we need a general or specialised util here.

For specialised, the below is enough:

```javascript
findLinkRangeFromSelection( 'foo', selection, model );
```

For a general one, probably as you described (but all our plugins will use it as:

```javascript
const positionToStartSearchFrom = selection.getFirstPostion();
const attributeName = 'foo';
const attributeValue = selection.getAttribute( attributeName );
findAttributeRange( model, positionToStartSearchFrom, attributeName, attributeValue );
```

jodator: > BTW, couldn't this method land in one of the model's classes? Maybe a static TreeWalker's method to not mix the layers?

It might be a better solution then (if we want a general util - we might re-use it somewhere where we operate on attribue - I'd check: `Highlight`, `RestrictedEditing` and `Mention` from top of my mind. In those we might do similar things but differently.

jodator: Yep. Highlight: https://github.com/ckeditor/ckeditor5/blob/6612ea68d5f175ce490a86d2092882bbf070fd80/packages/ckeditor5-highlight/src/highlightcommand.js#L68-L73

jodator: Mention and Restricted editing have different mechanisms.

packages/ckeditor5-typing/src/utils/inlinehighlight.js
```diff
@@ -0,0 +1,78 @@
...
+				attributeName,
+				selection.getAttribute( attributeName ),
+				editor.model
+			);
+			const viewRange = editor.editing.mapper.toViewRange( modelRange );
+
+			// There might be multiple view elements in the `viewRange`, for example, when the `a` element is
+			// broken by a UIElement.
+			for ( const item of viewRange.getItems() ) {
+				if ( item.is( tagName ) && !item.hasClass( className ) ) {
```
jodator, line 50: I worry this one :D For `<a>` and `<code>` it should work, but what if someone would downcast code to `<span>`?

It might also, doesn't work for 2SCM enable on attributes which convert to `<span>` - one could do that.

I'll check that.

jodator: OK, a bit artificial, but possible scenario (might be handled by a follow-up?)



The manual test was updated with this:

```javascript
// Custom attribute downcast (I love spans)
editor.conversion.attributeToElement( {
	model: 'italic', view: {
		name: 'span',
		styles: {
			'color': 'red'
		}
	}, converterPriority: 'high'
} );

editor.conversion.for( 'downcast' ).attributeToElement( {
	model: 'bold',
	view: ( attributeValue, writer ) => {
		if ( !attributeValue ) {
			return;
		}

		return writer.createAttributeElement(
			'span',
			{
				style: 'font-weight:bold'
			},
			{
				priority: 80,
				id: uid()
			} );
	}, converterPriority: 'high'
} );

editor.conversion.for( 'downcast' ).attributeToElement( {
	model: 'underline',
	view: ( attributeValue, writer ) => {
		if ( !attributeValue ) {
			return;
		}

		return writer.createAttributeElement(
			'span',
			{
				style: 'text-decoration:underline'
			},
			{
				priority: 20,
				id: uid()
			} );
	}, converterPriority: 'high'
} );

// Don't mind the function name, but the attribute highlighting must be also re-set.
setupLinkHighlight( editor, 'italic', 'span', 'highlight-i' );
setupLinkHighlight( editor, 'underline', 'span', 'highlight-u' );
```

So the question is - should we care? //cc @Reinmar

jodator: TL;DR attribute highlighting for 2SCM will not work for spans.

Reinmar: Yup, not a problem for now.

packages/ckeditor5-typing/src/utils/findattributerange.js
tomalec: When I was moving it out from `Link` I had a gut feeling, that such util has to be used somewhere else as well, but was not able to find an example ;)

@Reinmar Do we agree 
- on moving such util somewhere, and using it in `-highlight` and here?
- if we agree on ☝️ , then where? `-engine`
- and which implementation? I like to one from `-highlight` for its readability, but the call stack get's longer, and I'm not sure which one is faster.

packages/ckeditor5-typing/src/utils/inlinehighlight.js
tomalec: > Adds a visual highlight style to attribute elements bound to the anchored selection.

But "bound" is quite vague.

> Adds a visual highlight style to attribute elements adjacent to the anchored selection.

?

jodator: Let's leave it as is. No need to debate on this ;)

packages/ckeditor5-typing/src/utils/inlinehighlight.js
tomalec: > The default export should be named ...

The default export is named "default" by the definition :trollface: 

I change it all to `inlineHighlight`. I don't like it either :/ Do you have an idea for a better name? I got too attached to the former name, to figure out something new :|

tomalec: > Another thing is mabye we should also expose this on TwoStepCaretMovement plugin?:

I was considering it as well. Given how they were separated before, I thought it because of we foresee use-cases for 2SCM without any highlight.

tomalec:
> should not mix view/model terminology.

Correct me if I'm wrong, but I think this feature in general mixes the view and the model.

That said, I totally support making the docs clearer, but just probably lack the ideas how :)

tomalec:
@jodator I think it's ready for another look.

jodator:
> @jodator I think it's ready for another look.

@tomalec could you take a look at merge conflicts?

jodator:
@tomalec  Sorry, I've forgot to mention that some other PR was using `findLinkRange()`  - it got merged earlier.

tomalec:
@jodator Fix, I hope it's fine now.

jodator:
I've found one issue with the docs only. The rest comments were addressed :+1:

Issue #6544, opened 2020-04-03
Removing more than 7 rows in a 7 column table crashes editor


jodator:
## 📝 Provide detailed reproduction steps (if any)

Have a table like in the gif below and select more than 7 last rows (see that table have 7 columns, sic!).

![](https://user-images.githubusercontent.com/247363/78356075-31981b00-75af-11ea-92c8-3b7908410688.gif)

### ✔️ Expected result

Table rows are removed without error.

### ❌ Actual result

There is an error.

## 📃 Other details

*   Browser: Chrome/Firefox

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

jodator:
Check after: #6439. Might be a similar error.

jodator:
It seems that it no longer crashes on master but have bug found in here: ckeditor/ckeditor5-table#305.

jodator:
Looks like fixed by #6569 or #6437.

Issue #6506, opened 2020-03-27
Problems with the <table> abstraction in the model


jodator:
### Issue description (from #6630)

Atomic converters works in on operation at a time basis. In order to work out some of the above constrains they calculate operations based on the **model** state. This fails if more than two operations happens on a table.

Consider changing header rows from `2` to `1`. The downcast converter:

1.  Takes second **model** row `H2` (because it is at index which is after a heading section) finds its view element and move it to `<tbody>` section.
2.  Taking **model** state calculates which view cells requires renaming to `th`.

Now consider removing multiple rows (`H2` and `B1`). The `RemoveRow` command will:

1.  Remove `H2` row.
2.  Change heading rows to `1`.
3.  Set selection in `B2`.

Now this will fail because heading rows converter sees the same attribute change and will assume that table is valid. But what this converter will do is:

1.  Take the second **model** row (which is now `B2` ) and move it's view to `<tbody>`.

After table's attributes change conversion the remove row converter will step in. Removing elements from model is done on **positions** (offset in parent). So it will:

1.  For a removed row at given position it will get model parent (`table`).
2.  Get view row at an expected offset (let it be `2`).
3.  As the view has been modified (wrongly) it will find row `B2` to be removed instead of `H2`.
4.  :boom: The model to view mapping is broken as view for `B2` was removed instead of `H2`.

So the spotted problems are:

1.  Using model in downcast conversion can cause problems. The fix for changing header rows looks quite simple (I've almost got it working in ckeditor/ckeditor5-table#301).
2.  Table model and its conversion is hard to operate and figure out side effects. Downcast converters are too simplistic and relays on model state / single change.
3.  It may encourage developers to use `enqueueChange()` - I don't see other option for the problem described here. The thing with `enqueueChange()` is that run conversion on each step so it acts like step-by-step conversion but it looks like more work is done than necessary.
4.  I don't like the `headingRows` conversion. One idea is to have `<tableRow isHeading="true">` - this will require some other checks (no `isHeading` gaps have to be checked) than now but conversion "should" be simple for this case. And I don't see how we can make similar change with columns.
5.  I didn't check the columns heading conversion and its problems...

---

### Original issue description

It is rather a general list of spotted problems with current table abstraction which might be unsolvable or would require much work.

1.  No table sections in the model (`<tbody>`, `<thead>`)

WIP - the list should be updated with other issues.

---

Ad. 1 - No table sections in the model

[...]

Issue #6502, opened 2020-03-26
Remove row in complex tables sometimes provides invalid results


jodator:
Extracted from: https://github.com/ckeditor/ckeditor5-table/pull/286#issuecomment-604318239

## 📝 Provide detailed reproduction steps (if any)

Sometimes when the table has complex geometry and you remove a row ( with multiselect or not ), cells are changing their position in table:

![26_table1](https://user-images.githubusercontent.com/34380544/77630313-1859f380-6f4b-11ea-9441-7f0d74129171.gif)

![26_table3](https://user-images.githubusercontent.com/34380544/77630656-b5b52780-6f4b-11ea-83ad-a2724b3efcca.gif)

### ✔️ Expected result

The table geometry should be preserved (ie the cell should stay on the same column)

### ❌ Actual result

It is randomly pushed in weird positions.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

jodator:
This also look like a case for #6439 - to check.

jodator:
The algorithm is wrong :shrug: The order of operations isn't OK. Below is an idea of how this should work:

![](https://user-images.githubusercontent.com/247363/78898747-99010f80-7a74-11ea-8b69-26395eef0f1c.jpg)

Steps should be:

1. Cache all cells that spans over removed rows.
2. Move all cells from removed rows (and reduce their rowspan) to a row after removed rows. (work on cached table structure from `TableWalker`).
3. Remove rows.
4. Trim cached cells from step 1.

Bonus: fewer operations in multi-row removal.

Issue #6437, opened 2020-03-16
The table breaks after removing a last header row


Mgsy:
## 📝 Provide detailed reproduction steps (if any)

1. Insert a table
2. Put the caret in the last row
3. Create a header row
4. Delete header

### ✔️ Expected result

The row disappears properly.

### ❌ Actual result

The table breaks and the selection moves out of the table.
![Screenshot 2020-03-16 at 08 34 41](https://user-images.githubusercontent.com/20988892/76733688-9b1bcb00-6761-11ea-893d-9aa1bbbee16b.png)

## 📃 Other details

**Note**: It seems to be an old issue, it also occurs in v12.1.0

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

jodator:
Might be fixed by #6439 - to check.

PR #7572, opened 2020-07-07
#7454: Down-casting whole table on headingRows attribute change

### LINKED ISSUE #7454: Row-spanned cell gets broken down on removing heading section above that cell

1.  Create a table with 1 heading row an a col-spanned cell below heading section.
2.  Remove heading row.

### ✔️ Expected result

The layout below removed 

### ❌ Actual result

The cell is broken down before removing changing heading section.

## 📃 Other details

*   CKEditor version: latest master

---

niegowski:

Fix (table): Table structure should not be changed while removing heading row. Closes #7454.

Fix (table): Merging multiple cells should not crash editor while removing multiple empty rows and columns that appeared after merge. 

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/src/tableediting.js
+++ b/packages/ckeditor5-table/src/tableediting.js
@@ -35,6 +35,7 @@ import TableUtils from '../src/tableutils';
 import injectTableLayoutPostFixer from './converters/table-layout-post-fixer';
 import injectTableCellParagraphPostFixer from './converters/table-cell-paragraph-post-fixer';
 import injectTableCellRefreshPostFixer from './converters/table-cell-refresh-post-fixer';
+import injectTableHeadingRowsRefreshPostFixer from './converters/table-heading-rows-refresh-post-fixer';
 
 import '../theme/tableediting.css';
 
@@ -115,9 +116,6 @@ export default class TableEditing extends Plugin {
 		// Table heading columns conversion (change of heading rows requires reconversion of the whole table).
 		conversion.for( 'editingDowncast' ).add( downcastTableHeadingColumnsChange() );
 
-		// Table heading rows change requires reconversion of the whole table.
-		this.listenTo( model, 'applyOperation', headingRowsAttributeChangeHandler( model ) );
-
 		// Define all the commands.
 		editor.commands.add( 'insertTable', new InsertTableCommand( editor ) );
 		editor.commands.add( 'insertTableRowAbove', new InsertRowCommand( editor, { order: 'above' } ) );
@@ -144,6 +142,7 @@ export default class TableEditing extends Plugin {
 		editor.commands.add( 'selectTableRow', new SelectRowCommand( editor ) );
 		editor.commands.add( 'selectTableColumn', new SelectColumnCommand( editor ) );
 
+		injectTableHeadingRowsRefreshPostFixer( model );
 		injectTableLayoutPostFixer( model );
 		injectTableCellRefreshPostFixer( model );
 		injectTableCellParagraphPostFixer( model );
@@ -155,23 +154,4 @@ export default class TableEditing extends Plugin {
 	static get requires() {
 		return [ TableUtils ];
 	}
-}
-
-// Model#applyOperation handler for headingRows attribute changes.
-function headingRowsAttributeChangeHandler( model ) {
-	return ( event, [ operation ] ) => {
-		if ( !operation.isDocumentOperation ) {
-			return;
-		}
-
-		if ( operation.type != 'addAttribute' && operation.type != 'removeAttribute' && operation.type != 'changeAttribute' ) {
-			return;
-		}
-
-		const element = operation.range.getContainedElement();
-
-		if ( element && element.is( 'table' ) && operation.key == 'headingRows' ) {
-			model.document.differ.refreshItem( element );
-		}
-	};
 }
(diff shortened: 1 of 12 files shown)


packages/ckeditor5-table/src/tableediting.js
```diff
@@ -113,9 +112,11 @@ export default class TableEditing extends Plugin {
 		conversion.attributeToAttribute( { model: 'colspan', view: 'colspan' } );
 		conversion.attributeToAttribute( { model: 'rowspan', view: 'rowspan' } );
 
-		// Table heading rows and columns conversion.
+		// Table heading columns conversion (change of heading rows requires reconversion of the whole table).
 		conversion.for( 'editingDowncast' ).add( downcastTableHeadingColumnsChange() );
-		conversion.for( 'editingDowncast' ).add( downcastTableHeadingRowsChange() );
+
+		// Table heading rows change requires reconversion of the whole table.
+		this.listenTo( model, 'applyOperation', headingRowsAttributeChangeHandler( model ) );
```
jodator, line 119: @scofalik & @Reinmar - I need to double check this as this changes how table is converted.

After numerous bugs with changing headings + removing rows we concluded that enough is enough. The table models [sucks](https://github.com/ckeditor/ckeditor5/issues/6506). Downcasting `headingRows` change requires moving `<tr>` in the view from `<thead>` to `<tbody>` or vice-versa. It can fail in multiple scenarios ([#6502](https://github.com/ckeditor/ckeditor5/issues/6502), [#6437](https://github.com/ckeditor/ckeditor5/issues/6437), [#6544](https://github.com/ckeditor/ckeditor5/issues/6544), [#6502](https://github.com/ckeditor/ckeditor5/issues/6502), [#6391](https://github.com/ckeditor/ckeditor5/issues/6391), [#6406](https://github.com/ckeditor/ckeditor5/issues/6406) + #7454) and would potentially never be OK.

Here (#7454) there was a problem with table layout post-fixer which was called after each `enqueueChange()` - and thus we hit limit of what API is allowing us to do. The solution is to wait with conversion and post-fixing after **all** table model changes are done and that `table` attribute converter _would know_ that also `table` children are changing.

Possible considered solutions were:

1.  A "hack": set heading rows to 0, remove/add rows, set heading rows to proper value - but it lacks an API for not converting stuff at once.
2.  Some API to mark that some changes need to be either converted at once or should not trigger model post-fixers.
3.  Mark `table` to be re-rendered at once.

We choose option `3` using `differ.refreshItem()` - which is called on every `headingRows` attribute operation as It requires to also take external operations into account when dealing with this bug. Also the API for many `TableUtils` get simplified but not passing a `batch` instance (:tada:).

---

In other words the solution is to re-render _whole_ table in the view if `headingRows` attribute was changed by any means.

---

What is needed - some kind of :+1: / :-1: - it might impact collaboration features in some way. Do we need better API for requiring re-converting whole chunk of content?

ps.: @Reinmar this is also some sort of ADR proposal (without sections - but have intro, considered steps and proposed solution).

jodator:
@ckeditor/qa-team This change requires some testing help. The scenarios which this change touch are for all structural changes inside (or "touching") heading section, like:

*   adding/removing row in table heading
*   adding/removing row just after a heading section
*   changing number of heading rows
*   row-spanned cells (in heading or "touching" heading section) caused problems here

jodator:
@ckeditor/qa-team This change requires some testing help. The scenarios which this change touch are for all structural changes inside (or "touching") heading section, like:

*   adding/removing row in table heading
*   adding/removing row just after a heading section
*   changing number of heading rows
*   row-spanned cells (in heading or "touching" heading section) caused problems here

FilipTokarski:
Seems to be ok, I didn't find any new bugs 👍

scofalik:
@FilipTokarski could you test it with real-time editing?

FilipTokarski:
I found a bug with real time editing:
1. On two clients at the same time add a row above the header row

Result:
- on one client the original header row is now a normal row with normal cells
- on the second client the same, but it keeps the grey background of header row

Screencasts:

packages/ckeditor5-table/src/tableediting.js
scofalik: Why this is done on `applyOperation` not in a post-fixer?

As a rule of thumb, we try to avoid doing stuff on `applyOperation` - this is the last resort solution. Getting between multiple operations is a risky business. OTOH, I see that you only mark the table to refresh, so no changes in the model tree are done. Might be okay.

The one difference that quickly comes to mind is that you might have an attribute operation for heading rows and then an operation that reverses it, so in the end, there's no change. Post-fixer would not refresh the table, AFAIU. 

So, once again, if it can be done through post-fixer and there are no counterarguments, I think I'd go with post-fixer. But I don't see real big disadvantages in doing this with `applyOperation` either, other than we simply avoid it because it is a more vulnerable "place", if you know what I mean.

niegowski:
@FilipTokarski could you check that failing case once again? Maybe last change would fix it.

FilipTokarski:
I'm still getting 2 header rows in the model, and 3 in view.  
  
But apart from this issue, as it comes to real-time editing, thing seem to work fine 👍

scofalik:
I guess that the table is incorrect and not refreshed because `headingRows` attribute does not change in this case. Only new rows are added. So the table is not marked to be refreshed.

jodator:
> I guess that the table is incorrect and not refreshed because `headingRows` attribute does not change in this case. Only new rows are added

It looks like weird bug - the table headings must be updated if you insert row in a heading section.

before: `headingRows=2`

insert row in heading section

after `headingRows=3`

Dunno why the model doesn't reflect that but the view does o\_0.

Issue #5671, opened 2019-10-28
Starting a line with an asterisk and space after a SHIFT+ENTER


wielinde:
![ckeditor-shift-enter](https://user-images.githubusercontent.com/327272/67661063-88465c80-f960-11e9-9d39-0b46914e9a2f.gif)

##  Detailed reproduction steps

1. Go to the official demo page of CKEditor 5 
2. Type a line
3. Finish that line by pressing SHIFT+ENTER
4. Start the next line with the an asterisk (*) followed by a space

### ✔️ Expected result

The asterisk and space should not be transformed. No bullet list should get generated. 

### ❌ Actual result

1. A bullet list is created, starting at the preceding line.
2. The first Letter of the preceding line gets lost.

* Browser: All browsers
* OS: All OSes
* CKEditor version: Latest stable

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

jodator:
Hi, thanks for the report. It looks like I forgot about block autoformat while fixing https://github.com/ckeditor/ckeditor5/issues/1955.

PR #7576, opened 2020-07-08
Enabled block autoformat in non-empty blocks

### LINKED ISSUE #6170: Block autoformat could work in all kind of blocks

So, right now when you're in at the beginning of a heading 2 and type `###` it is not going to change it to a heading 3. It's only possible to turn a paragraph into other types of blocks.

This means that you can't easily toggle between block types. Also, if you made a mistake, you need to get back to a paragraph (e.g. by undoing) and only then you can use autoformatting.

One example of an application where you can always switch between block types is Notion and I use this quite a lot. Especially when I'm changing heading levels (cause I'm reorganizing a document).

Another aspect is that currently you can only apply autoformatting to an **empty block**. We'd also like to apply formatting for blocks that have some content in it, as long as your cursor is at the beginning of the line.

Related issue:

*   [It should be possible to convert a heading into a list item and vice versa](https://github.com/ckeditor/ckeditor5/issues/2368)

cc @jodator @panr @oleq

---

mlewand:

Feature (autoformat): Block autoformat can also triggered in blocks other than paragraphs. Closes #6170.

Feature (autoformat): Enabled autoformat also for blocks that are not empty.

---

### Additional information

* It's possible now to use autoformat for blocks that already have some text in it.
* Block autoformat can be triggered in any block, not just paragraphs.
* Minor names refactoring.
* Corrected the existing manual test (parts were outdated).


mlewand:
@ckeditor/qa-team Gents, I'd like to ask you for some manual testing 🙏

FilipTokarski:
Works really smooth, I checked it also with Track Changes and real-time editing, seems to be ok 👍

packages/ckeditor5-autoformat/tests/blockautoformatediting.js
```diff
@@ -167,6 +167,18 @@ describe( 'blockAutoformatEditing', () => {
 			sinon.assert.notCalled( spy );
 		} );
 
+		it( 'should not call callback when typing in the middle of block text', () => {
+			const spy = testUtils.sinon.spy();
```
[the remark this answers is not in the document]
mlewand: Yea... the main reason why I went with `const spy = testUtils.sinon.spy();` is because I duplicated another test from this suite.

Given that the test is fairly simple, uses only a single spy, I think it's ok to leave it unnamed. And I agree that it would stand out compared to other TCs in this suite.

WDYT @tomalec?

packages/ckeditor5-autoformat/tests/manual/autoformat.md
```diff
@@ -22,6 +20,4 @@
 
 1. For every autoformat pattern: Undo until you'll see just the pattern (e.g. `- `). Typing should be then possible without triggering the autoformatting again.
 
-1. Typing a different pattern in an already converted block **must not** trigger the autoformatting. For example, typing `- ` in a heading should not convert a heading to a list.
```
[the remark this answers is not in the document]
mlewand: Good point, reflected that in f2e7043.

packages/ckeditor5-autoformat/src/blockautoformatediting.js
```diff
@@ -82,18 +85,34 @@ export default function blockAutoformatEditing( editor, plugin, pattern, callbac
...
 		}
 
-		const match = pattern.exec( blockToFormat.getChild( 0 ).data );
+		// In case a command is bound, do not re-execute it over an existing block style which would result with a style removal.
+		// Instead just drop processing so that autoformat trigger text is not lost. E.g. writing "# " in a level 1 heading.
+		if ( command && command.value === true ) {
+			return;
+		}
+
+		const firstNode = blockToFormat.getChild( 0 );
```
Reinmar, line 99: Why is this code needed? Why do we need any new code here (from this point down)?

packages/ckeditor5-autoformat/src/blockautoformatediting.js
```diff
@@ -82,18 +85,34 @@ export default function blockAutoformatEditing( editor, plugin, pattern, callbac
...
+		const match = pattern.exec( firstNode.data );
 
 		// ...and this text node's data match the pattern.
 		if ( !match ) {
 			return;
 		}
 
+		const range = first( editor.model.document.selection.getRanges() );
```
Reinmar, line 107: I don't understand why you're changing this. Why did you remove `$` from the regexps?

mlewand: Without tailing `$` removed it would only match blocks with text noes that contain _only_ the autocomplete trigger character. While related issue requests that non empty blocks like `###^ foo` should also be transformed.

And for the line referenced by this comment, the range check is to only trigger autocompletion when the change occurred in `### ^foo` and not in another place, like `### fo^o` - you don't want to trigger autocomplete when typing in the middle of the block (also handles #5671 referenced in the comment below).

Reinmar: What if the string that we match the regexp to is trimmed to the current selection position. Does it make sense for it to contain characters on the right side of the selection?

Reinmar: We're scanning the text from the very first text node anyway. If there's an element after that text node, or another text node (different attributes), then we'd not try to match it against the given regexps anyway.

Therefore, we can't reliably say that you can also match to the right of the caret. We're looking for the text on the left of the caret. Therefore, in this case:

```
<p>## ^xx</p>
```

In my opinion, we should be matching "## " only, so the regexps could look like this: `/^## $/`.

Right now you're matching "## xx" against `/^## /` and then checking if that occurred right before the selection which is counterintuitive to me. What's also counterintuitive is that you won't be matching against the entire element content anyway – only the first text node is used as I mentioned. That's not very straightforward.

However, it's not a big deal either so if you want to keep this implementation no big harm will be done. I'll just leave comments how you can improve it.

packages/ckeditor5-autoformat/src/blockautoformatediting.js
```diff
@@ -82,18 +85,34 @@ export default function blockAutoformatEditing( editor, plugin, pattern, callbac
...
+
+		// We're only handling a collapsed selection that is right after the matched text (#5671).
+		const expectedPosition = ( new Position( range.root, firstNode.getPath() ) ).getShiftedBy( match[ 0 ].length );
```
Reinmar, line 110: `model.createPositionBefore( firstNode ).getShiftedBy( match[ 0 ].length )`

packages/ckeditor5-autoformat/src/blockautoformatediting.js
Reinmar: `selection.getFirstPosition()` instead of what you're doing this + check before we do ANYTHING that the selection is collapsed.

packages/ckeditor5-autoformat/src/blockautoformatediting.js
```diff
@@ -82,18 +85,34 @@ export default function blockAutoformatEditing( editor, plugin, pattern, callbac
...
+
+		if ( !( range.isCollapsed && range.end.isEqual( expectedPosition ) ) ) {
```
Reinmar, line 112: Checking range.isCollapsed that late is waste of resources – do it at the beginning of everything.

packages/ckeditor5-autoformat/src/blockautoformatediting.js
mlewand: Restored original regexp format in a63f559.

mlewand:
Implementation changed a little since the last time, @ckeditor/qa-team can I ask you for some manual testing to be sure?

FilipTokarski:
Looks good 👍

Reinmar:
The code looks good and resolves the issue so after adding a test you can merge this.

mlewand:
Ok, added a missing TC.

PR #7582, opened 2020-07-08
#6774: Made it possible to disable the WidgetTypeAround plugin

### LINKED ISSUE #6774: There should be some way to globally disable the magic block feature
ATM it's a dependency of `Widget` so even `config.removePlugins` won't work and some integrators may want to get rid of it (for whatever the reason).

When we implement the keyboard support, then using CSS to get rid of it won't be an option because the feature will change the behavior of the selection.

oleq:

Feature (widget): Made it possible to disable the `WidgetTypeAround` plugin on the fly. Closes #6774.

---

### Additional information

This PR brings 2 things:

*   the integration with `WidgetTypeAround#isEnabled`
*   docs explaining how to use it + docs explaining how to hide the buttons only

@Reinmar WDYT about the docs?


oleq:
I added tests and documented the new private method. Let's give it another round of review.

PR #7590, opened 2020-07-09
Typing over the link feature

### LINKED ISSUE #4762: Typing over the link (replacing link text)
This was mentioned in ckeditor/ckeditor5#4720 but requires a separate ticket. Because of UX we need to allow replacing link text.

See https://github.com/w3c/editing/issues/156#issuecomment-263858380.

See also – typing after a link: https://github.com/ckeditor/ckeditor5-link/issues/72.

pomek:

Feature (link): Typing over the selected link will not remove the link itself. Instead, the typed text will replace the link text. Closes #4762.

### DIFF
--- a/packages/ckeditor5-link/src/linkediting.js
+++ b/packages/ckeditor5-link/src/linkediting.js
@@ -527,7 +527,7 @@ function shouldCopyAttributes( model ) {
 	// `nodeBefore` = the position is at the end of the link element.
 	const nodeAtLastPosition = lastPosition.textNode || lastPosition.nodeBefore;
 
-	// Both nodes must indicate the same node in the model tree.
+	// If both references the same node selection contains a single text node.
 	if ( nodeAtFirstPosition === nodeAtLastPosition ) {
 		return true;
 	}
--- a/packages/ckeditor5-link/src/linkediting.js
+++ b/packages/ckeditor5-link/src/linkediting.js
@@ -11,6 +11,7 @@ import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import MouseObserver from '@ckeditor/ckeditor5-engine/src/view/observer/mouseobserver';
 import TwoStepCaretMovement from '@ckeditor/ckeditor5-typing/src/twostepcaretmovement';
 import Input from '@ckeditor/ckeditor5-typing/src/input';
+import Clipboard from '@ckeditor/ckeditor5-clipboard/src/clipboard';
 import LinkCommand from './linkcommand';
 import UnlinkCommand from './unlinkcommand';
 import AutomaticDecorators from './utils/automaticdecorators';
@@ -45,7 +46,8 @@ export default class LinkEditing extends Plugin {
 	 * @inheritDoc
 	 */
 	static get requires() {
-		return [ TwoStepCaretMovement, Input ];
+		// Clipboard is required for handling cut and paste events while typing over the link.
+		return [ TwoStepCaretMovement, Input, Clipboard ];
 	}
 
 	/**
@@ -444,16 +446,6 @@ export default class LinkEditing extends Plugin {
 			deletedContent = true;
 		}, { priority: 'high' } );
 
-		// Detect when the content was cut from the editor.
-		// this.listenTo( view.document, 'cut', () => {
-		// 	deletedContent = true;
-		// }, { priority: 'high' } );
-
-		// Detect when pasting content. Attributes should not be copied.
-		// this.listenTo( view.document, 'paste', () => {
-		// 	selectionAttributes = null;
-		// }, { priority: 'high' } );
-
 		// Listening to `model#deleteContent` allows detecting whether selected content was a link.
 		// If so, before removing the element, we will copy its attributes.
 		this.listenTo( editor.model, 'deleteContent', () => {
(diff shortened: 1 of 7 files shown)


pomek:
@ckeditor/qa-team, could you check how the solution works with CF?

Mgsy:
It works fine :ok\_hand:  We've tested it in all browser and on mobiles.

packages/ckeditor5-link/src/linkediting.js
```diff
@@ -414,4 +418,86 @@ export default class LinkEditing extends Plugin {
...
+		}, { priority: 'high' } );
+	}
+}
+
+// Checks whether passed elements are the same and they represent a text node.
+//
+// @param {module:engine/model/node~Node} nodeA
+// @param {module:engine/model/node~Node} nodeB
+// @returns {Boolean}
+function isCorrectLink( nodeA, nodeB ) {
```
jodator, line 493: Looking at the function alone, it should have a different name:

*   Does not check the link at all.
*   Checks only it is a single text node.

In a broader scope - This check should be improved, the first case that fails is:



The other thing is that this also works for other attributes and if IIRC, it shouldn't (I might skip something when noting). It should preserve a link with other attributes, so bolded link, but not bold alone:

Peek 2020-07-10 12-59

So those two are tied together in this check. We already have [`findLinkRange()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_link_findlinkrange.html#static-function-findLinkRange) - so it might be used here by comparing range from selection and the result of `findLinkRange()`. This should cover cases like this:

Selection_420

What to preserve in such case? I think that anything repeatable is good enough. The goal is to preserve link + other attributes in cases when all those attributes are continous (e.g. bolded link). In a case above whatever it will take - I'm guessing the selection attributes will be preserved, so it would end up as a bolded link and it is OK IMO.

packages/ckeditor5-link/src/linkediting.js
```diff
@@ -414,4 +418,86 @@ export default class LinkEditing extends Plugin {
...
+
+		// Selection attributes when started typing over the link.
+		let selectionAttributes;
+
+		// Whether pressed `Backspace` or `Delete`. If so, attributes should not be preserved.
+		let deletedContent;
+
+		// Detect pressing `Backspace` / `Delete`.
+		editor.editing.view.document.on( 'delete', ( evt, data ) => {
+			/* istanbul ignore else */
```
jodator, line 443: We can't do that - in what other cases it would trigger `delete` event? Shouldn't we test that? If that not occur then `if` might be not needed.

packages/ckeditor5-link/src/linkediting.js
```diff
@@ -414,4 +418,86 @@ export default class LinkEditing extends Plugin {
...
+		const editor = this.editor;
+
+		// Selection attributes when started typing over the link.
+		let selectionAttributes;
+
+		// Whether pressed `Backspace` or `Delete`. If so, attributes should not be preserved.
+		let deletedContent;
+
+		// Detect pressing `Backspace` / `Delete`.
+		editor.editing.view.document.on( 'delete', ( evt, data ) => {
```
jodator, line 442: Everything here works well, but looks messy. Could you add information in the docs why we need to listen on different events to detect this? AFAICS it is because "delete"  observer will handle removing part of text on typing.

packages/ckeditor5-link/src/linkediting.js
```diff
@@ -414,4 +418,86 @@ export default class LinkEditing extends Plugin {
...
+			/* istanbul ignore else */
+			if ( data.domEvent.keyCode == keyCodes.delete || data.domEvent.keyCode == keyCodes.backspace ) {
+				deletedContent = true;
+			}
+		}, { priority: 'high' } );
+
+		editor.model.on( 'deleteContent', () => {
```
jodator, line 449: Use `this.listenTo( editor.model, ... )` - we use this pattern elsewhere. The difference is that it allows to stop all `LinkEditing` listeners on other emitters by `this.stopListening()` - ie when  calling `Plugin.destroy()`.

There might be other problems of which I don't recall now.

Reinmar:
cc @scofalik I think you should know about this change as it touches the very fragile typing/insert/delete content.

Reinmar:
Just to be sure:

*   Does this PR also change the behavior at the beginning of the link? It should, as it was part of the issue too.
*   Does it affect pasting and other forms of inserting content other than typing? It should not.

scofalik:
At first sight, it looks like the change should be safe for track changes:

*   `deleteContent` callback does not modify anything. If I correctly imagine what will happen, this should work just fine.
*   `insertContent` callback, AFAICS, just adds some attributes to the model that hasn't been inserted yet:
    *   this callback and track change's callback have the same priority,
    *   if this callback is fired before track change's callback, track changes will never know that something happened,
    *   if it is fired after it, AFAICS from the execution flow, the new callback will be fired anyway (track change's callback eventually fires `insertContent` too and your callback should be executed then).

This needs testing, but as I said, as I look at it now, it should be fine.

pomek:
Guys, I've just pushed fixes. Could you check the feature once again?

cc: @jodator, @scofalik @ckeditor/qa-team

jodator:
@pomek  for pasting and basic link integration - now it works nice with the typing detection :+1: 

The only thing left is my previous comment: https://github.com/ckeditor/ckeditor5/pull/7590#discussion_r452779004. In case of multiple other attributes over a link. It is kinda corner case but IMO we could treat it as typing over a link.

pomek:
@jodator, should I write additional test for such case? In the code, we copy the selection's attributes.

jodator:
Yes, please :) I think that something like below (sample is without values) should work:

```html
foo [<$text linkHref>b</$text><$text linkHref bold>a</$text><$text linkHref>r</$text>] baz
```

after type "box":

```html
foo <$text linkHref>box[]</$text> baz
```

FilipTokarski:
Checked in all browsers and apart from the case with attributes, it looks ok 👍

pomek:
@jodator, done. Ready for review once again. I hope, it will be the last one.

PR #7591, opened 2020-07-09
i/7457: Add missing indication that image is linked

### LINKED ISSUE #7457: Missing indication that image is linked

Right now there is no major visual difference between a linked image and a "normal" image. Both look the same, and the only small difference is that the linked image has `pointer` cursor on `:hover` and after clicking it `LinkImageUI` component button is enabled.

NOTE: Linked image has a caption set.
image

Follow-up on: #7452

--

panr:

Feature (link): Added missing icon indication in the top-right corner of the image that informs the image is linked. Closes #7457.

Internal (theme-lark): Added the linked image indication styles.


panr:
I decided to add the indication through the downcast instead of CSS pseudo-element.

panr:
image

packages/ckeditor5-link/tests/linkimageediting.js
```diff
@@ -11,6 +11,13 @@ import normalizeHtml from '@ckeditor/ckeditor5-utils/tests/_utils/normalizehtml'
 import { getData as getViewData } from '@ckeditor/ckeditor5-engine/src/dev-utils/view';
 import ImageCaptionEditing from '@ckeditor/ckeditor5-image/src/imagecaption/imagecaptionediting';
 
+const linkIconInditatorElement = '<span class="ck ck-link-image_icon"><svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">' +
```
Reinmar, line 14: That can be an import from the SVG file (the \<svg> part) so you won't need to update it when the SVG changes.

panr: 🤦 oh, right...

panr: Fixed it, but with additional dev comment.

PR #7595, opened 2020-07-09
#7453: Pasting table into table with headings should not break table layout.

### LINKED ISSUE #7453: Pasting table over table when selection covers heading and body sections breaks layout

1.  Create a table with heading rows
2.  Paste table with row-spanned cell (ie whole column) that would span over heading and body sections

### ✔️ Expected result

(or)

## 📃 Other details

*   CKEditor version: latest master

---

niegowski:

Fix (table): Pasting a table into a table with headings should not break table layout. Closes #7453.

---

### Additional information


packages/ckeditor5-table/src/tableclipboard.js
```diff
@@ -316,7 +316,29 @@ function replaceSelectedCellsWithPasted( pastedTable, pastedDimensions, selected
...
+	if ( selection.firstColumn < headingColumns && headingColumns <= selection.lastColumn ) {
+		const rowsLimit = { first: selection.firstRow, last: selection.lastRow };
+		const newCells = doVerticalSplit( selectedTable, headingColumns, rowsLimit, writer );
+
+		cellsToSelect.push( ...newCells );
+	}
+
+	const selectionRanges = cellsToSelect
+		.map( cell => writer.createRangeOn( cell ) )
+		.sort( ( a, b ) => a.start.isBefore( b.start ) ? -1 : 1 );
```
jodator, line 339: Could you expose: https://github.com/ckeditor/ckeditor5/blob/064bbd8ec054c74e5807812ebd5ac7987e188dc3/packages/ckeditor5-table/src/utils/selection.js#L201-L202? Another thing - could you also check (briefly) if sorting ranges isn't done elsewhere (to elevate that to some general util). If not `export` will be fine if there is a similar code outside table a ticket for moving this method is also welcome (apart of `export` here).

packages/ckeditor5-table/src/tableclipboard.js
```diff
@@ -316,7 +316,29 @@ function replaceSelectedCellsWithPasted( pastedTable, pastedDimensions, selected
 		insertPosition = writer.createPositionAfter( cellToInsert );
 	}
 
-	writer.setSelection( cellsToSelect.map( cell => writer.createRangeOn( cell ) ) );
+	// If there are any headings, all the cells that overlap from heading must be splitted.
+	const headingRows = parseInt( selectedTable.getAttribute( 'headingRows' ) || 0 );
+	const headingColumns = parseInt( selectedTable.getAttribute( 'headingColumns' ) || 0 );
+
+	if ( selection.firstRow < headingRows && headingRows <= selection.lastRow ) {
```
jodator, line 323: Could you add some `isFooBar()` or `const isFooBar` to describe that (and below) check?

packages/ckeditor5-table/src/tableclipboard.js
niegowski: I was considering adding global util, but I don't know which place would be ok so that we don't need to import engine package. Also this is a extrenally simple one-liner that i'm not sure if there any sense in exposing it

jodator: I think that exposing that or the comparator is OK - nice to have that "logic" in one place to not duplicate it everywhere. I see also a benefit of additional comments there.

Those utils are also kinda local to table package so re-using existing ones is OK for me.

PR #7596, opened 2020-07-09
i/6559: Fixed opening of the link balloon on touch devices

### LINKED ISSUE #6559: Link balloon does not show on the first click
Not sure if it's related but there are issues with balloon in iOS as well but only on 1st focus. If you load the editor and click straight on link, balloon doesn't appear. It does however work on subsequent clicks or if you focus on other text before the link. Very rarely you need to focus out and on again for it to work properly.

[Example](https://streamable.com/f7p0p)

_Originally posted by @dovydaskukalis in https://github.com/ckeditor/ckeditor5/issues/6289#issuecomment-602704668_

\---

panr:

Fix (link): Fixed opening of the link balloon on touch devices on the first tap. Closes #6559.

Tests (mention): Fix the test for mention integration that was related to the `LinkUI`.


panr:
@FilipTokarski @Mgsy could you test the new behavior on your touch devices, respectively on Android and iOS? 🙏

FilipTokarski:
Looks good on Android 👍

Mgsy:
It works fine on iOS too :+1: It also has been fixed on the desktop.

panr:
**Last-minute question**: should we set timeout only on touch devices? Right now the changes may be "too complex" and unnecessary (especially for the testing purpose) for the pointer devices.

So my new, more direct solution for that is: make it async **only** on touch devices.

This will save us from:
- delaying the code exec for pointers
- changing the tests (this bothers me the most)
- making multiple dev notes complaining why this or that piece of code should wait for the execution etc.

Reinmar:
Interesting idea. However, this change is not that complex for now. You had to change just a couple of tests, which is not bad. Also, code branching is always a cost. For instance – we'd need to have separate tests for touch devices to have code coverage.

So, I'd KISS. As long as we know that this change was really needed and that we don't see a better (i.e. easy and reliable) option, it's fine.

Reinmar:
I explained what's happening and how to reproduce this on every browser in https://github.com/ckeditor/ckeditor5/issues/6559#issuecomment-666338330.

oleq:
> I explained what's happening and how to reproduce this on every browser in [#6559 (comment)](https://github.com/ckeditor/ckeditor5/issues/6559#issuecomment-666338330).

That's what I meant, I tried to follow your reproduction steps but I failed. Maybe some browser extension you use is interacting here?

PR #7598, opened 2020-07-09
#3233: Added model Position#findAncestor() and Element#findAncestor() methods

### LINKED ISSUE #3233: Implement TableUtils#getTable()
We can simplify:

```js
const table = editor.model.document.selection.getFirstPosition().getAncestors().find( item => item.name == 'table' );
```

To:

```js
const table = tableUtils.getTable( editor.model.document.selection );
```

I was surprised to not find this function because I thought I saw it somewhere.

niegowski:

Feature (engine): Added model `Position#findAncestor()` and `Element#findAncestor()` methods. Closes #3233.

Internal (table): Removed `findAncestor` helper.

MINOR BREAKING CHANGE (table): The `findAncestor()` utility function was removed.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/model/element.js
+++ b/packages/ckeditor5-engine/src/model/element.js
@@ -210,6 +210,28 @@ export default class Element extends Node {
 	}
 
 	/**
+	 * Returns the parent element of the given name. Returns null if the element is not inside the desired parent.
+	 *
+	 * @param {String} parentName The name of the parent element to find.
+	 * @param {Object} [options] Options object.
+	 * @param {Boolean} [options.includeSelf=false] When set to `true` this node will be also included while searching.
+	 * @returns {module:engine/model/element~Element|null}
+	 */
+	findAncestor( parentName, options = { includeSelf: false } ) {
+		let parent = options.includeSelf ? this : this.parent;
+
+		while ( parent ) {
+			if ( parent.name === parentName ) {
+				return parent;
+			}
+
+			parent = parent.parent;
+		}
+
+		return null;
+	}
+
+	/**
 	 * Converts `Element` instance to plain object and returns it. Takes care of converting all of this element's children.
 	 *
 	 * @returns {Object} `Element` instance converted to plain object.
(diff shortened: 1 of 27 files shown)


packages/ckeditor5-engine/tests/model/element.js
```diff
@@ -247,6 +247,33 @@ describe( 'Element', () => {
 		} );
 	} );
 
+	describe( 'findAncestor', () => {
+		let p, td, tr, table;
```
jodator, line 251: I hate that model tests already use `p` for `paragraph` - but :man\_shrugging:

packages/ckeditor5-engine/src/model/element.js
```diff
@@ -209,6 +209,28 @@ export default class Element extends Node {
...
+	 * @param {String} parentName The name of the parent element to find.
+	 * @param {Object} [options] Options object.
+	 * @param {Boolean} [options.includeSelf=false] When set to `true` this node will be also included while searching.
+	 * @returns {module:engine/model/element~Element|null}
+	 */
+	findAncestor( parentName, options = { includeSelf: false } ) {
+		let parent = options.includeSelf ? this : this.parent;
+
+		while ( parent ) {
+			if ( parent.name === parentName ) {
```
jodator, line 224: ~`parent.is()` would be better.~

WDYT about using `parent.is()` here? Performance wise it shouldn't be a big impact now. Using `is()` is prefered in most cases but here we might need to check only for name. I don't think that `element.findAncestor( 'mode:element' )` is needed.

niegowski: I was thinking about more generic solution like [view element findAncestor()](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_engine_view_element-Element.html#function-findAncestor) that accepts callback also (i.e., Matcher accepted patterns) but currently there is no use case for that scenario so I thought that we could extend this later if it will be needed.

jodator: OK, maybe we'll change that anyway...

PR #7602, opened 2020-07-10
New marker conversion helpers

### LINKED ISSUE #7556: Introduce a "safer" marker conversion

Currently, marker conversion works in a way that a view element (HTML tag) is created at positions where a marker starts and where a marker ends. This solution creates an incorrect HTML (for example, a custom element is inserted in `<tr>` which accepts only `<td>`s and `<th>`s). This leads to two problems:

1.  Data sanitization.
2.  Data parsing from HTML.

Data sanitization becomes problematic because various libraries that perform it may remove the custom tags that are in incorrect places. Similarly, data parsing may lead to errors.

Thus, I propose changing how we would convert markers. If a marker boundary is before or after a view container element, instead of creating "marker tag", we should add an attribute to that element. The attribute will mark that a marker starts/ends after/before that element.

Some other remarks:

*   Proposed attribute names for view container elements: `data-marker-start`, `data-marker-end`.
*   Those attributes will simply contain a marker name. If there will be multiple, we will separate them with a comma.
*   I think that we should introduce those new converters as default ones. So that when you are creating a new feature that uses a marker you will not need to provide converters. This means that we need also a default converter for "marker tags". At the moment, I don't really understand why we decided to go with `<comment id="...">` instead of unifying it, for example: `<ck-marker data-name="...">`. I propose that when used with default marker conversion, markers will simply use their names and will be converted to `<ck-marker data-name="markerName">`.
*   I wonder whether we should mark already existing marker conversion helpers as deprecated. I think that there should be no need to use anything else other than a marker name to save the marker with the data.
*   Of course, we will be backward compatible. Already existing features should remain as they are.

BTW. we also discussed using HTML comments instead of regular tags to show marker boundaries but we decided that data sanitizers might remove them as well.

scofalik:

Feature: Introduced new marker conversion helpers that produce semantic HTML data output. See `DowncastHelpers#markerToData()` and `UpcastHelpers#dataToMarker()`. Closes #7556.  
Feature: Added `model.Schema` instance to downcast conversion API, available under `conversionApi.schema`.  
Other: `UpcastHelpers#elementToMarker()` is now deprecated. Use `UpcastHelpers#dataToMarker()` instead. `DowncastHelpers#markerToElement()` should be used only for editing downcast.

BREAKING CHANGE (engine): Marker names with "," character are now disallowed.


docs/framework/guides/architecture/editing-engine.md
```diff
@@ -159,9 +159,11 @@ Markers are a special type of ranges.
 * They can only be created and changed through the {@link module:engine/model/writer~Writer model writer}.
 * They can be synchronized over the network with other collaborating clients.
 * They are automatically updated when the document's structure is changed.
-* They can be converted to attributes or elements in the [view](#view).
+* They can be {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToHighlight converted to the editing view}, to show them in the editor.
```
Reinmar, line 162: What about `markerToElement()`? Shouldn't we mention both things:

```
	* They can be converted to the editing view, to show them in the editor (as {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToHighlight highlights} or {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToElement elements).
```

scofalik: For a moment that was deprecated so I didn't include it. But since it will not be deprecated after all, we may mention it.

docs/framework/guides/architecture/editing-engine.md
```diff
@@ -159,9 +159,11 @@ Markers are a special type of ranges.
...
+* They can be {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToData converted to the data view}, to store them with the document data.
+* They can be {@link module:engine/conversion/upcasthelpers~UpcastHelpers#dataToMarker loaded with the document data}.
 
-This makes them ideal for storing and maintaining additional data in the model &mdash; such as comments, selections of other users, etc.
+Markers are ideal for storing and maintaining additional data &mdash; such as comments or selections of other users &mdash; that is related to some portion of the content.
```
Reinmar, line 166:
```suggestion
Markers are ideal for storing and maintaining additional data related to portions of the document &mdash; such as comments or selections of other users.
```

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -90,7 +90,8 @@ export default class DataController {
 		 * @member {module:engine/conversion/downcastdispatcher~DowncastDispatcher}
 		 */
 		this.downcastDispatcher = new DowncastDispatcher( {
-			mapper: this.mapper
+			mapper: this.mapper,
```
Reinmar, line 93: The new param needs to be documented.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -239,22 +239,25 @@ export default class DowncastHelpers extends ConversionHelpers {
 	/**
 	 * Model marker to view element conversion helper.
 	 *
+	 * **Note**: This method should be used only for editing downcast. For data downcast, use {@link #markerToData} which produces
+	 * valid HTML data.
```
Reinmar, line 243:
```suggestion
	 * **Note**: This method should be used only for editing downcast. For data downcast, use
	 * {@link #markerToData `#markerToData()`} that produces valid HTML data.
```

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -357,6 +358,113 @@ export default class DowncastHelpers extends ConversionHelpers {
...
 	}
+
+	/**
+	 * Model marker converter for data downcast.
+	 *
+	 * This conversion creates a representation for model marker boundaries in the view:
+	 *
+	 * * if the marker boundary is at a position where text nodes are allowed, then a view element with specified tag name
+	 * and `name` attribute is added at that position,
+	 * * in other case, a specified attribute is set on an view element that is before/after marker boundary.
```
Reinmar, line 369:
```suggestion
	 * * in other cases, a specified attribute is set on a view element that is before/after marker boundary.
```

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -357,6 +358,113 @@ export default class DowncastHelpers extends ConversionHelpers {
...
+	 *
+	 * Tag and attribute names and values are generated from the marker name:
+	 *
+	 * * templates for attributes are `data-group-start-before="name"`, `data-group-start-after="name"`,
```
Reinmar, line 373: Maybe like that?

```suggestion
`data-[group]-start-before="[name]"`
```

It's hard to understand which parts of that template are variables.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -357,6 +358,113 @@ export default class DowncastHelpers extends ConversionHelpers {
...
+	 * `data-group-end-before="name"` and `data-group-end-after="name"`,
+	 * * templates for view elements are `<group-start name="name">` and `<group-end name="name">`.
+	 *
+	 * Attributes mark whether given marker start or end boundary is before or after given element.
+	 * Attributes `data-group-start-before` and `data-group-end-after` are favored. Other two are used when former two cannot be used.
```
Reinmar, line 378:
```suggestion
	 * Attributes `data-group-start-before` and `data-group-end-after` are favored. The other two are used when the former two cannot be used.
```

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -357,6 +358,113 @@ export default class DowncastHelpers extends ConversionHelpers {
...
+	 *
+	 * Typically, the marker names use `group:uniqueId:otherData` convention. For example: `comment:e34zfk9k2n459df53sjl34:zx32c`.
```
Reinmar, line 380: I'd move that higher. It sets the context. Before "Tag and attribute names..."

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -357,6 +358,113 @@ export default class DowncastHelpers extends ConversionHelpers {
...
+	 *
+	 *		// View:
+	 *		<blockquote><p data-group-end-after="name" data-group-start-after="name">Foo</p></blockquote>
+	 *
+	 * When there are multiple markers from the same group stored in the same attribute of the same element, their
+	 * name parts are put together in the attribute value, for example: `data-group-start-before="name1,name2,name3"`.
+	 *
+	 * Examples of usage:
+	 *
+	 *		// Using the default conversion:
```
Reinmar, line 428: I'd extend these comments. It's a bit hard to understand what is `model: 'content'` without reading deeper.

Reinmar: Missing:

*   which markers are handled
*   eventually, some example data string (simple, short, just to give the context)

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -755,6 +863,141 @@ function removeUIElement() {
...
+		const mapper = conversionApi.mapper;
+		const viewWriter = conversionApi.writer;
+
+		// Adding closing data first to keep the proper order in the view.
+		handleMarkerBoundary( markerRange, false );
+		handleMarkerBoundary( markerRange, true );
+
+		evt.stop();
+
+		function handleMarkerBoundary( range, isStart ) {
```
Reinmar, line 897: I don't like that this function is defined here. It'll be create on every call of `insertMarkerData`. Not that I see performance issues already, but just for the sanity I'd move it outside (if possible).

Reinmar: Same with the other helpers.

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -330,8 +335,61 @@ export default class UpcastHelpers extends ConversionHelpers {
 	 * @returns {module:engine/conversion/upcasthelpers~UpcastHelpers}
 	 */
 	elementToMarker( config ) {
+		console.warn(
```
Reinmar, line 338: Missing error documentation. And you should use http://127.0.0.1:8081/ckeditor5/20.0.0/api/module_utils_ckeditorerror.html#static-function-attachLinkToDocumentation

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -330,8 +335,61 @@ export default class UpcastHelpers extends ConversionHelpers {
...
 		return this.add( upcastElementToMarker( config ) );
 	}
+
+	/**
+	 * View to model marker conversion helper.
+	 *
+	 * Converts view data created by {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToData} back to a model marker.
+	 *
+	 * This converter looks for specific view elements and view attributes that marks marker boundaries. See
+	 * {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToData} to learn what view data is expected by this converter.
```
Reinmar, line 352:
```suggestion
	 * {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToData `#markerToData()`} to learn what view data is expected by this converter.
```

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -330,8 +335,61 @@ export default class UpcastHelpers extends ConversionHelpers {
...
+			'The UpcastHelpers#elementToMarker() method has been deprecated and will be removed in the near future. ' +
+			'Please use #dataToMarker() method instead.' );
+
 		return this.add( upcastElementToMarker( config ) );
 	}
+
+	/**
+	 * View to model marker conversion helper.
+	 *
+	 * Converts view data created by {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToData} back to a model marker.
```
Reinmar, line 349:
```suggestion
	 * Converts view data created by {@link module:engine/conversion/downcasthelpers~DowncastHelpers#markerToData `#markerToData()`} back to a model marker.
```

Reinmar: Otherwise, it looks like this:

image

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -330,8 +335,61 @@ export default class UpcastHelpers extends ConversionHelpers {
...
+	 *
+	 * This converter looks for specific view elements and view attributes that marks marker boundaries. See
```
Reinmar, line 351:
```suggestion
	 * This converter looks for specific view elements and view attributes that mark marker boundaries. See
```

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -330,8 +335,61 @@ export default class UpcastHelpers extends ConversionHelpers {
...
+	 *
+	 * The `config.view` property is equal to the marker group name to convert.
+	 *
+	 * By default, this converter creates markers with `group:name` name convention (to match the default `markerToData` conversion).
+	 *
+	 * The conversion configuration can take a function that will generate a marker name.
+	 * If such function is set as the `config.model` parameter, it is passed the `name` part from the view element or attribute and it is
+	 * expected to return a string with the marker name.
+	 *
+	 *		// Using the default conversion.
```
Reinmar, line 362: Same comment as for the `markerToData()`.

packages/ckeditor5-engine/src/model/markercollection.js
```diff
@@ -92,6 +92,11 @@ export default class MarkerCollection {
 	 */
 	_set( markerOrName, range, managedUsingOperations = false, affectsData = false ) {
 		const markerName = markerOrName instanceof Marker ? markerOrName.name : markerOrName;
+
+		if ( markerName.includes( ',' ) ) {
+			throw new CKEditorError( 'markercollection-incorrect-marker-name: Marker name cannot contain "," character.', this );
```
Reinmar, line 97: Missing error documentation.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
scofalik: I am always afraid that somebody would think that `[]` are also the part of the name (`data-[comment]-start-before="[123:abc]"`) :D

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
scofalik: I had it there originally :).

packages/ckeditor5-engine/src/controller/datacontroller.js
scofalik: I am not sure if we document `conversionApi` "content" anywhere.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
scofalik: I had the same feelings as you but for clarity reasons, I left those helpers inside those functions. They are used only by those functions and this way they can receive fewer parameters.

scofalik:
Ready for another review.

packages/ckeditor5-engine/src/model/markercollection.js
```diff
@@ -92,6 +92,16 @@ export default class MarkerCollection {
 	 */
 	_set( markerOrName, range, managedUsingOperations = false, affectsData = false ) {
 		const markerName = markerOrName instanceof Marker ? markerOrName.name : markerOrName;
+
+		if ( markerName.includes( ',' ) ) {
+			/**
+			 * Marker name cannot contain "," character.
```
Reinmar, line 98:
```suggestion
			 * Marker name cannot contain the "," character.
```

Issue #5566, opened 2019-10-10
Custom Html Widget


turigeza:
## 📝 Provide a description of the new feature
A widget `@ckeditor/ckeditor5-widget` which could contain arbitrary text / html.

*What is the expected behavior of the proposed feature?*

Let's say the source code contains the following html.

```
<div class="custom-html">
    This is some text I don't want to edit just want to preserve it. Same goes
    for the html.

    The div always have the class custom-html.

    Ideally I wanted it to be converted to a non editable widget. 
    Which can then behave as a widget. 

    Can be deleted / selected /copied / pasted etc ...
    <p data-id="22221111"> 
    Unfortunately the p tab gets discarded 
    because it does not fit any model in the editor. </p>    
    <p class="someclass">
    But the content here is arbitrary and I don't want to edit it so I don't need a model for it. 
    I just would like the html not to be modified / stripped from this widget.</p>
</div>
```

So currently I would end up with this in the editor.  

```
<div class="custom-html ck-widget ck-widget_selected" contenteditable="false">This is some text I don't want to edit just want to preserve it. &nbsp; &nbsp; The div always have the class customHtml. &nbsp; &nbsp; Ideally want it to be converted to a non editable widget. &nbsp; &nbsp; Which can then behave as a widget. &nbsp; &nbsp; Can be deleted / selected /copied / pasted etc ... &nbsp; &nbsp; &nbsp; &nbsp;Unfortunately the p tab gets discarded &nbsp; &nbsp; because it does not fit any model in the editor. &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;But the content here is arbitrary and I don't want to edit it so I don't need a model for it. &nbsp; &nbsp; I just would like the html stripped from this widget.</div>

```

I am importing old content and converting it so that I can use ckeditor5 with it. 

Now for the most important things I have created (or will do so) models (plugins) which match the old content. Like for example for my layouts. 

But there is some stuff which really is not worth describing as a model and I would be happy not to be able to edit it. 

Can I achieve this already somehow ? This is how far I got following the docs. I was hoping I can do it with a non editable widget. 

https://github.com/turigeza/ckeditor5-build-balloon-block/tree/master/src/CustomHTML

Thank you
---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

Mgsy:
@jodator, can you take a look at it?

jodator:
[...]

PR #7621, opened 2020-07-14
#4469: Implemented the view RawElement

### LINKED ISSUE #1684: Placeholder does not disappear after inserting media

🐞 Bug report

## 💻 Version of CKEditor

Master and the latest release.

## 📋 Steps to reproduce

1. Go to https://ckeditor.com/docs/ckeditor5/latest/features/editor-placeholder.html or use local build documentation.
2. To the first editor (it must be empty), paste any URL to YT or add the media using the UI button.

## ✅ Expected result

`Type the content here!` should disappear.

## ❎ Actual result

But it does not:

image

I can reproduce it on Firefox, Chrome, and Safari.

### LINKED ISSUE #4469: Introduce RawElement (aka ShadowElement)
See this comment: https://github.com/ckeditor/ckeditor5-engine/issues/1643#issuecomment-456814970

~~The name `UIElement` is misleading because this element can also be used to e.g. output raw content to the data.~~

~~We also need to rename `createUIElement()` and perhaps some other helpers.~~

oleq:

Feature (engine): Implemented the view `RawElement`. Implemented the `DowncastWriter#createRawElement()` method. Closes #4469.

Fix (media-embed): The editor placeholder should disappear after inserting media into an empty editor. Closes #1684.

Docs (ckeditor5): Used the `RawElement` instead of `UIElement` in the "Using a React component in a block widget" guide (see #1684).

Internal (media-embed): Removed the `getFillerOffset()` hack from the `createMediaFigureElement()` helper that is no longer needed since the media content is rendered using `RawElements` (see #1684).

BREAKING CHANGE (engine): The `DomConverter#getParentUIElement()` method was renamed to `DomConverter#getHostViewElement()` because now it supports both `UIElement` and `RawElement` (see #4469).

---

### Additional information

TBH the conversion (mapping) to a `RawElement` should be possible but I don't know where to write tests for it because I couldn't find similar tests for other view element types. Any ideas?


packages/ckeditor5-engine/src/view/domconverter.js
```diff
@@ -887,7 +896,8 @@ export default class DomConverter {
 	 *
 	 * The following places are considered as incorrect for selection boundaries:
 	 * * before or in the middle of the inline filler sequence,
-	 * * inside the DOM element which represents {@link module:engine/view/uielement~UIElement a view ui element}.
+	 * * inside the DOM element which represents {@link module:engine/view/uielement~UIElement a view UI element}.
+	 * * inside the DOM element which represents {@link module:engine/view/rawelement~RawElement a view raw element}.
```
Reinmar, line 900: Something went wrong here.

packages/ckeditor5-engine/src/view/downcastwriter.js
```diff
@@ -272,6 +273,36 @@ export default class DowncastWriter {
...
+	/**
+	 * Creates a new {@link module:engine/view/rawelement~RawElement}.
+	 *
+	 *		writer.createRawElement( 'span' );
+	 *		writer.createRawElement( 'span', { id: 'foo-1234' } );
+	 *
+	 * Custom render function can be provided as third parameter:
+	 *
+	 *		writer.createRawElement( 'span', null, function( domDocument ) {
+	 *			const domElement = this.toDomElement( domDocument );
```
Reinmar, line 285: IMO, in the callback you should get the DOM element as in case of RawElement, unlike like in UIElement, you do not control the host element. It's being controlled by the view.

packages/ckeditor5-engine/src/view/downcastwriter.js
```diff
@@ -272,6 +273,36 @@ export default class DowncastWriter {
 		return uiElement;
 	}
 
+	/**
+	 * Creates a new {@link module:engine/view/rawelement~RawElement}.
```
Reinmar, line 277: Does it make sense to create raw elements without their "internals" (passing the callback)?

Also, I'd focus on having here a good cross-ref between UIElement and RawElement, explaining also briefly what's the difference, etc. These API methods have a high chance of being visited and it's hard to figure out what it is based on this method.

packages/ckeditor5-engine/src/view/downcastwriter.js
```diff
@@ -1572,6 +1603,16 @@ export default class DowncastWriter {
...
 		}
 
+		// If position is placed inside RawElement - throw an exception as we cannot break inside.
+		if ( position.parent.is( 'rawElement' ) ) {
+			/**
+			 * Cannot break inside RawElement instance.
+			 *
+			 * @error view-writer-cannot-break-raw-element
+			 */
+			throw new CKEditorError( 'view-writer-cannot-break-raw-element', this.document );
```
Reinmar, line 1613:
```suggestion
			throw new CKEditorError( 'view-writer-cannot-break-raw-element: Cannot break inside RawElement instance.', this.document );
```

Reinmar: You will need to reformat this line.

packages/ckeditor5-engine/src/view/rawelement.js
```diff
@@ -0,0 +1,163 @@
...
+
+	/**
+	 * Creates DOM element based on this view `RawElement`.
+	 *
+	 * **Note**: Each time this method is called new DOM element is created.
+	 *
+	 * @param {Document} domDocument
+	 * @returns {HTMLElement}
+	 */
+	toDomElement( domDocument ) {
```
Reinmar, line 147: Is it really needed? (See also my comments about the callback's params)

packages/ckeditor5-engine/src/view/rawelement.js
```diff
@@ -0,0 +1,163 @@
...
+	 */
+	_insertChild( index, nodes ) {
+		if ( nodes && ( nodes instanceof Node || Array.from( nodes ).length > 0 ) ) {
+			/**
+			 * Cannot add children to a {@link module:engine/view/rawelement~RawElement}.
+			 *
+			 * @error view-rawelement-cannot-add
+			 */
+			throw new CKEditorError(
+				'view-rawelement-cannot-add: Cannot add child nodes to RawElement instance.',
```
Reinmar, line 108: to a RawElement instance

packages/ckeditor5-engine/src/view/rawelement.js
```diff
@@ -0,0 +1,163 @@
...
+	 * Overrides {@link module:engine/view/element~Element#_insertChild} method.
+	 * Throws {@link module:utils/ckeditorerror~CKEditorError CKEditorError} `view-rawelement-cannot-add` to prevent
+	 * adding any child nodes to a `RawElement`.
+	 *
+	 * @protected
+	 */
+	_insertChild( index, nodes ) {
+		if ( nodes && ( nodes instanceof Node || Array.from( nodes ).length > 0 ) ) {
+			/**
+			 * Cannot add children to a {@link module:engine/view/rawelement~RawElement}.
```
Reinmar, line 103: missing "instance" or remove "a" :D

packages/ckeditor5-engine/tests/view/renderer.js
```diff
@@ -3300,6 +3301,44 @@ describe( 'Renderer', () => {
...
+				viewP._removeChildren( 0, viewP.childCount );
+				viewP._insertChild( 0, [ new ViewText( viewDocument, 'Foo' ), raw2, new ViewText( viewDocument, ' Bar' ) ] );
+
+				renderer.markToSync( 'children', viewRoot );
+				renderer.markToSync( 'children', viewP );
+				renderer.render();
+
+				expect( normalizeHtml( domRoot.innerHTML ) ).to.equal( normalizeHtml(
+					'<p>Foo<span><span id="id2"><b>RAW2</b></span></span> Bar</p>' ) );
+			} );
```
Reinmar, line 3340: I'm missing a test of updating a raw element's attributes via the view.

Reinmar:
I left a couple of comments. The two biggest things are:

*   The form of the raw element's render callback.
*   A missing test for using `toWidget( rawElement )` and then whether e.g. the classes of this element are nicely updated when selection changes – as that's one of the things that wasn't possible before (you had to have a wrapper). Similar test can be written for the data pipeline – for checking that you can indeed implement something like "raw HTML" embedding https://github.com/ckeditor/ckeditor5/issues/5566#issuecomment-542156007. These tests could land in the widget package as some integration tests, although the data pipeline test would not, in fact, be in any way related to widgets (as only the editing pipeline uses it).

Reinmar:
I left a couple of comments. The two biggest things are:

*   The form of the raw element's render callback.
*   A missing test for using `toWidget( rawElement )` and then whether e.g. the classes of this element are nicely updated when selection changes – as that's one of the things that wasn't possible before (you had to have a wrapper). Similar test can be written for the data pipeline – for checking that you can indeed implement something like "raw HTML" embedding https://github.com/ckeditor/ckeditor5/issues/5566#issuecomment-542156007. These tests could land in the widget package as some integration tests, although the data pipeline test would not, in fact, be in any way related to widgets (as only the editing pipeline uses it).

packages/ckeditor5-engine/src/view/domconverter.js
Reinmar: It renders as:

> inside the DOM element which represents {@link module:engine/view/uielement~UIElement a view UI element}. inside the DOM element which represents {@link module:engine/view/rawelement~RawElement a view raw element}

Duplicated "inside..."

packages/ckeditor5-engine/src/view/downcastwriter.js
Reinmar: What about this? I can see that the callback is still optional.

I'll remove the first two examples and leave the rest as is to make it less confusing.

Reinmar: I can see it's a copy-paste from UI element... But it makes more sense there as sometimes UIElements can be used without any content as their purpose is different.

Reinmar: Oh, I remember now that you mentioned that it'd look bad if the last param wasn't optional since attrs are. So, that's fine. I'll just remove those examples.

docs/framework/guides/architecture/editing-engine.md
```diff
@@ -226,14 +226,15 @@ editor.data;                    // The data pipeline (DataController).
...
 
 The element types can be defined as follows:
 
 * **Container element** &ndash; The elements that build the structure of the content. Used for block elements such as `<p>`, `<h1>`, `<blockQuote>`, `<li>`, etc.
 * **Attribute element** &ndash; The elements that cannot contain container elements inside them. Most model text attributes are converted to view attribute elements. They are used mostly for inline styling elements such as `<strong>`, `<i>`, `<a>`, `<code>`. Similar attribute elements are flattened by the view writer, so e.g. `<a href="..."><a class="bar">x</a></a>` would automatically be optimized to `<a href="..." class="bar">x</a>`.
 * **Empty element** &ndash; The elements that must not have any child nodes, for example `<img>`.
 * **UI elements** &ndash; The elements that are not a part of the "data" but need to be "inlined" in the content. They are ignored by the selection (it jumps over them) and the view writer in general. The contents of these elements and events coming from them are filtered out, too.
```
godai78, line 236: Only plural on the list.

Issue #7575, opened 2020-07-08
Upgrade JSDoc to the latest version, align internal JSDoc plugins


ma2ciek:
## Provide a description of the task

JSDoc is the CKEditor 5 API docs extractor. Upgrading it and aligning the code in its plugins that we maintain should allow us for:

*   writing code using the latest JS syntax without errors
*   documenting other projects
*   possible perf improvements

Since JSDoc changed its internal parser to `@babel/parser` a lot has changed internally between its two minor versions, hence a lot of tests should be written to ensure the quality of doclet extraction. These tests can be helpful if we consider #7425 at some point.

PR #7642, opened 2020-07-17
Align API docs to new API docs extractor, fix minor issues

ma2ciek:

Docs: Align API docs to new API docs extractor, fix minor issues. Part of #7575.

---

### Additional information


mlewand:
I see that you R+ed it so in case there are no breaking changes I'm good with it :+1:

PR #7644, opened 2020-07-17
Introduced the Collection.addMany method and Collection.change event

### LINKED ISSUE #7627: Add a method for adding multiple items to the collection at once

When working on #6194 we figured that it would be good to have a way to add multiple items into a collection at once.

The tl;dr motivation is that when the colleciton changes, a resource-heavy logic needs to be run. Currently the only option is to listen to `add` event and run the logic for each added item - which is inefficient.

Instead a `addBatch` method and event pair will solve this.

Note that this new method needs to play well with existing `add` event. Also calling `collection.add()` should trigger `batchAdd` event too.

---

mlewand:

Feature: Introduced the `Collection.addMany()` method for adding multiple items in a single call. Closes #7627.

Feature: Introduced the `Collection.change` event. See #7627.

---

### Additional information

With this I was also able to simplify `DocumentColorCollection` by using new `change` event.

`clear()` method uses almost the entire `remove()` logic except that it must not fire `change` for each removal. Thus I extracted a private base of `remove()` that can be reused in both cases.

### DIFF
--- a/packages/ckeditor5-utils/tests/collection.js
+++ b/packages/ckeditor5-utils/tests/collection.js
@@ -391,7 +391,7 @@ describe( 'Collection', () => {
 
 			collection.addMany( items );
 
-			expect( spy.callCount, 'call count' ).to.equal( 1 );
+			sinon.assert.calledOnce( spy );
 			expect( spy.args[ 0 ][ 1 ] ).to.deep.eql( {
 				added: items,
 				removed: [],
@@ -638,6 +638,23 @@ describe( 'Collection', () => {
 			sinon.assert.calledWithExactly( spy, sinon.match.has( 'source', collection ), item3, 0 );
 		} );
 
+		it( 'should fire the "change" event', () => {
+			const item = getItem( 'foo' );
+			const spy = sinon.spy();
+
+			collection.add( item );
+			collection.on( 'change', spy );
+
+			collection.remove( item );
+
+			sinon.assert.calledOnce( spy );
+			expect( spy.args[ 0 ][ 1 ] ).to.deep.eql( {
+				added: [],
+				removed: [ item ],
+				index: 0
+			} );
+		} );
+
 		it( 'should throw an error on invalid index', () => {
 			collection.add( getItem( 'foo' ) );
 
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-utils/tests/collection.js
```diff
@@ -148,8 +148,31 @@ describe( 'Collection', () => {
 	} );
 
 	describe( 'add()', () => {
-		it( 'should be chainable', () => {
-			expect( collection.add( {} ) ).to.equal( collection );
+		it( 'should proxy its calls to addMany', () => {
```
jodator, line 151: I don't think that we need to test it - it's an implementation detail and might change.

packages/ckeditor5-utils/tests/collection.js
```diff
@@ -148,8 +148,31 @@ describe( 'Collection', () => {
...
+			const spy = sinon.spy( collection, 'addMany' );
+			const item = {};
+
+			collection.add( item );
+
+			sinon.assert.calledOnce( spy );
+			sinon.assert.calledWithExactly( spy, sinon.match.array.contains( [ item ] ), undefined );
+		} );
+
+		it( 'should proxy custom index', () => {
```
jodator, line 161: The same comment as for previous test - I'd rather see those tests as the expected outcome and don't bother how the `Collection` do internally.

It would be much easier to refactor it later if needed).

packages/ckeditor5-utils/tests/collection.js
```diff
@@ -161,15 +184,28 @@ describe( 'Collection', () => {
 			collection.add( {} );
 			expect( collection ).to.have.length( 2 );
 		} );
+	} );
+
+	describe( 'addMany()', () => {
```
jodator, line 189: Here I would leave the old tests for `add()` - those are still valid and we do not care how they are handled internally.

A similar tests for `addMany()` are also needed. Kind duplicates code tests but we're adding new behavior so it would be nice to describe it with tests even though we're be testing similar things twice (because of implementation). Implementation might change but tests should stay the same.

packages/ckeditor5-utils/tests/collection.js
mlewand: Reverted all the old tests in 8cd7ff93.

My intention here was to keep it DRY and avoid same tests in two methods.

packages/ckeditor5-utils/tests/collection.js
mlewand: Resolved in 8cd7ff93.

packages/ckeditor5-utils/tests/collection.js
mlewand: Resolved in e13db78.

PR #7650, opened 2020-07-20
#4060: Undo/redo stacks cleared on DataController set().

### LINKED ISSUE #4060: Setting data doesn't clear undo stack
1. Open any sample.
2. Type few letters (or do any other change).
3. Call `editor.setData( '<p>foo</p>' );`
4. See that undo is enabled.
5. Undo. Error is thrown.

We have two things here:

* It should be possible to manage whether setting data resets the history (whether it's undoable or not) or not. Or, like in CKE4, there should be a way to tell whether `setData()` should create undo step and a way to clear the stack. This is more granular but boils down to the same story.
* If `setData()` was meant to create an undo step then undoing shouldn't throw an error.

   I also thought that features like typing may need to reset their stored batch on such actions. I think that `ChangeBuffer` should be doing this reliably now but it may be good to write a test for it (if there's none).

niegowski:

Fix (undo): Undo/redo stacks should be cleared on `DataController#set()`. Closes #4060.

Feature (engine): `DataController#set()` is now decorated so plugins can listen to `setData()` calls.

MAJOR BREAKING CHANGE: `editor.setData()` clears the undo and redo stacks.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/controller/datacontroller.js
+++ b/packages/ckeditor5-engine/src/controller/datacontroller.js
@@ -453,6 +453,15 @@ export default class DataController {
 	 *
 	 * @event init
 	 */
+
+	/**
+	 * Event fired after {@link #set set() method} has been run.
+	 *
+	 * The `set` event is fired by decorated {@link #set} method.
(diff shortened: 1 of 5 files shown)


Reinmar:
I think it should be marked as a major BC.

PR #7652, opened 2020-07-20
7336: ConversionAPI update

### LINKED ISSUE #7336: Writing an event-based elementToElement upcast converter should be simpler

Right now, if you need to do something at least slightly custom (so `elementToElement()` is not an option to you), you need to copy 90% of this code:

https://github.com/ckeditor/ckeditor5/blob/79221ae/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L544-L618

We should clean up and split this algorithm into set of function that one could pick from.

*   ~https://github.com/ckeditor/ckeditor5/blob/79221ae/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L572 – why do we check whether we can consume it after calling `getModelElement()`? Can't the order be changed?~ (edit: yep, already refactored)
*   https://github.com/ckeditor/ckeditor5/blob/79221ae/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L578-L587 + https://github.com/ckeditor/ckeditor5/blob/79221ae/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L595-L616 => `conversionApi.safeInsert()`?
*   https://github.com/ckeditor/ckeditor5/blob/79221ae/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L590 – position could be replaced with `modelElement` to make it shorter
*   https://github.com/ckeditor/ckeditor5/blob/79221ae/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L593 – I'd check whether this cannot be moved closer to the top or other code related to consuming the element.

---

jodator:

Feature (engine): Introduced new upcast `ConversionApi` helper methods - `conversionApi.safeInsert()` and `conversionApi.updateConversionResult()`. New methods are intended to simplify writing event based element-to-element converters. Closes #7336.

MAJOR BREAKING CHANGE (engine): The `config.view` parameter for upcast element-to-element conversion helpers configurations is again mandatory. You can retain previous "catch-all" behavior for upcast converter using `config.view = /[\s\S]+/`.

---

### Additional information

In addition to changes described previously under the ticket (https://github.com/ckeditor/ckeditor5/issues/7336#issuecomment-654704526):
- I've updated list converters a bit (dd9481f) by using some of the API - however, I had to add a conditional statement to not update the `modelRange` if it was defined already (we might skip that change).
- ~tests are still to be written (as the previous point might change implementation).~ done
- I've updated the API docs
- the `getSplitParts()` and `splitToAllowedParent()` are not used anymore - might be dropped.

TODO:
- [x] the `getSplitParts()` and `splitToAllowedParent()` are not used anymore - might be dropped - let's keep it for a while.
- [x] decide if special handling of `modelRange` is to be preserved (I couldn't find another way for that).
- [x] check if the consuming part shouldn't be simplified - a follow-up but not in this PR.


packages/ckeditor5-list/src/converters.js
```diff
@@ -378,30 +378,16 @@ export function viewModelConverter( evt, data, conversionApi ) {
...
 
-		// When `data.modelCursor` parent had to be split to insert list item...
-		if ( splitResult.cursorParent ) {
-			// Continue conversion in the split element.
-			data.modelCursor = writer.createPositionAt( splitResult.cursorParent, 0 );
-		} else {
-			// Otherwise continue conversion after the last list item.
-			data.modelCursor = data.modelRange.end;
-		}
+		conversionApi.updateConversionResult( listItem, data );
```
jodator, line 390: Check if modelCursor can be ommited.

jodator: I couldn't make it work without full copy-pate of the `updateConversionResult()` so :man\_shrugging: dunno we might live with that `if()` inside it.

jodator:
@niegowski Please make a review for this PR. We've looked at this with @Reinmar and the whole idea and change looks OK.

A conversation on a solution starts here: https://github.com/ckeditor/ckeditor5/issues/7336#issuecomment-649538681.

packages/ckeditor5-paragraph/src/paragraph.js
```diff
@@ -67,6 +67,7 @@ export default class Paragraph extends Plugin {
 
 				return modelWriter.createElement( 'paragraph' );
 			},
+			view: /\S/, // TODO: i/7336 - better API.
```
jodator, line 70:
```suggestion
			view: /\S/, // TODO: i/7336 - better API.
```

Ugh, left-over. I'll check the proper matcher option for this (catch-all).

jodator: We've used `/[\s\S]+/` for attributes matching, so I'd used that for now.

jodator:
@Reinmar I've updated the conversion deep-dive guide if you wish to check it - if not please just mark this PR as :heavy\_check\_mark:  to not block @niegowski with eventual merge.

packages/ckeditor5-table/tests/converters/upcasttable.js
```diff
@@ -285,7 +285,7 @@ describe( 'upcastTable()', () => {
 		);
 
 		expectModel(
-			'<fooTable><fooRow><fooCell></fooCell></fooRow></fooTable>'
+			'<fooTable><fooRow><fooCell><paragraph></paragraph></fooCell></fooRow></fooTable>'
```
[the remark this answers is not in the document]
jodator: `fooCell` does not allow `$text` inside.

packages/ckeditor5-engine/tests/conversion/upcasthelpers.js
```diff
@@ -120,7 +120,8 @@ describe( 'UpcastHelpers', () => {
 			expectResult( new ViewContainerElement( viewDocument, 'p', { 'data-level': 2 } ), '' );
 		} );
 
-		it( 'config.view is not set - should fire conversion for every element', () => {
+		// TODO i/7336: - define behavior
```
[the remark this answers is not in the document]
jodator: Nice catch. Thanks - it was a leftover.

packages/ckeditor5-paragraph/src/paragraph.js
jodator: And this is also "wrong" - there is a set of characters that can be used but `\n` is not one of them. Using `.+` is enough.

packages/ckeditor5-engine/src/conversion/upcastdispatcher.js
```diff
@@ -541,13 +604,73 @@ function createContextTree( contextDefinition, writer ) {
...
+ *			return;
+ *		}
+ *
+ *		// Children conversion may split `myElement`.
+ *		conversionApi.convertChildren( data.viewItem, myElement );
+ *
+ *		conversionApi.updateConversionResult( myElement, data );
+ *
+ * @method #updateConversionResult
+ * @param {module:engine/model/element~Element} element
```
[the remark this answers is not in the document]
jodator: Added.

packages/ckeditor5-engine/docs/_snippets/framework/extending-content-custom-element-converter.js
```diff
@@ -0,0 +1,167 @@
...
+
+	conversionApi.writer.insert(
+		conversionApi.mapper.toViewPosition( data.range.start ),
+		infoBox
+	);
+}
+
+ClassicEditor
+	.create( document.querySelector( '#editor-custom-element-converter' ), {
+		cloudServices: CS_CONFIG,
```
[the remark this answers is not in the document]
jodator: Yes. Any snippet that uses `EasyImage` - so practically any that has `Image` must use that.

jodator:
@Reinmar as @niegowski is off today:

1.  I've fixed found issues.
2.  I've added `UpcastConversionData` typedef: 2e7c739.

So IMO it is good to merge.

Reinmar:
I scanned all the code and I now work on the guide. I want to extend it a bit with some more context. But I may be stuck for a while on how to incorporate it into the existing guides. If you'll see that it may cause problems with maintaining this branch if it takes me 1-2 days, I can merge it as it is.

jodator:
> I scanned all the code and I now work on the guide. I want to extend it a bit with some more context

I'd go with a follow-up as it is only a guide update.

Reinmar:
I pushed guide improvements. Could you scan them, @jodator?

Reinmar:
OK, that's all for my review. I pushed some more changes to the API docs.

Please take a look at them and if I didn't make some bigger mistakes, you can merge this PR.

packages/ckeditor5-engine/src/conversion/upcastdispatcher.js
```diff
@@ -17,28 +17,34 @@ import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
 import mix from '@ckeditor/ckeditor5-utils/src/mix';
 
 /**
- * `UpcastDispatcher` is a central point of {@link module:engine/view/view view} conversion, which is a process of
+ * `UpcastDispatcher` is a central point of the model to view conversion, which is a process of
```
niegowski, line 20:
```suggestion
 * `UpcastDispatcher` is a central point of the view to model conversion, which is a process of
```

jodator:
@ni

jodator:
@Reinmar - nice update :) and @niegowski good catch with that typo. As from me the PR is OK now, so you can merge.

PR #7656, opened 2020-07-21
Optimized initial toolbar render with toolbar grouping enabled

### LINKED ISSUE #6194: Toolbar grouping should recalculate the styles less frequently

Currently toolbar recalculates the styles after adding each toolbar item.

Profiler flame chart

It would be good to optimize it, and ideally it should do it only once.

Style recalculations make for 7.4% of total `_initToolbar` function time (in an editor with small semantic content) and this function makes a significant contribution to editor load time today.

Here is a devtools profiler dump that can be viewed later on: [editor-init-small.zip](https://github.com/ckeditor/ckeditor5/files/4154665/editor-init-small.zip).

### ✔️ Expected result

Less style recalculations 🏃‍♀️

### ❌ Actual result

Styles are recalculated after adding each toolbar item  🐢 🐢 🐢 🐢 🐢 🐢 🐢 🐢 🐢

## 📃 Other details

This is a follow-up of #5880.

*   Browser: Chrome
*   OS: Windows (should be cross system)
*   CKEditor version: latest `master`

---

### LINKED ISSUE #7655: Removing the first hidden toolbar button throws an exception

1.  Go to https://ckeditor.com/docs/ckeditor5/20.0.0/examples/builds/document-editor.html
2.  Resize the window in a way that at least one button gets hidden.

1.  Use the following code to remove the first hidden button:

```javascript

const toolbarView = editor.ui.view.toolbar;
const visibleButtonsCount = toolbarView._behavior.ungroupedItems.length;
const firstHiddenItem = toolbarView.items.get( visibleButtonsCount );
toolbarView.items.remove( firstHiddenItem );
```

### ✔️ Expected result

**The first hidden** toolbar button gets removed, and (if possible) one of hidden buttons gets "promoted" to the toolbar button.

### ❌ Actual result

An error is thrown:

```
Uncaught CKEditorError: collection-remove-404: Item not found. Read more: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-collection-remove-404

    at ViewCollection._remove (http://localhost:8125/ckeditor5/tests/manual/all-features.js:143318:10)
    at ViewCollection.remove (http://localhost:8125/ckeditor5/tests/manual/all-features.js:142909:32)
    at ViewCollection.<anonymous> (http://localhost:8125/ckeditor5/tests/manual/all-features.js:138818:25)
    at ViewCollection.fire (http://localhost:8125/ckeditor5/tests/manual/all-features.js:146827:30)
    at ViewCollection._remove (http://localhost:8125/ckeditor5/tests/manual/all-features.js:143328:8)
    at ViewCollection.remove (http://localhost:8125/ckeditor5/tests/manual/all-features.js:142909:32)
    at <anonymous>:1:19
```

## 📃 Other details

That's a typical OBOE originating in this line: https://github.com/ckeditor/ckeditor5/blob/8f3c81bd98b979ce36521a8eda42ca2d18588752/packages/ckeditor5-ui/src/toolbar/toolbarview.js#L567

It's enough to change it to `if ( index >= this.ungroupedItems.length ) {`

There's a similar situation with adding buttons https://github.com/ckeditor/ckeditor5/blob/8f3c81bd98b979ce36521a8eda42ca2d18588752/packages/ckeditor5-ui/src/toolbar/toolbarview.js#L553 but it doesn't throw an error, instead `_updateGrouping()` call that follows it cleans up the structure, and no problem is visible.

---

mlewand:

Other (ui): Improved toolbar rendering time when multiple items are added or removed at once (e.g. during editor initialization). Closes #6194.

Fix (ui): Removing the first hidden toolbar button will no longer throw an exception. Closes #7655.

---

### Additional information

While working on it I spotted #7655 and decided to fix it while porting old listeners.


packages/ckeditor5-ui/src/toolbar/toolbarview.js
```diff
@@ -302,7 +302,7 @@ export default class ToolbarView extends View {
 				console.warn( attachLinkToDocumentation(
 					'toolbarview-item-unavailable: The requested toolbar item is unavailable.' ), { name } );
 			}
-		} );
+		} ).filter( item => item !== undefined ) );
```
oleq, line 305: Is there any test for this one? Why is it needed anyway?

packages/ckeditor5-ui/tests/toolbar/toolbarview.js
```diff
@@ -575,6 +575,21 @@ describe( 'ToolbarView', () => {
 
 				expect( ungroupedItems.map( i => i ) ).to.have.ordered.members( [ itemB, itemD ] );
 			} );
+
+			it( 'doesnt throw when removing the first of grouped items', () => { // (#7655)
```
oleq, line 579:
```suggestion
			it( 'doesn\'t throw when removing the first of grouped items', () => { // (#7655)
```

packages/ckeditor5-ui/tests/toolbar/toolbarview.js
```diff
@@ -575,6 +575,21 @@ describe( 'ToolbarView', () => {
...
+				const items = [ focusable(), focusable(), focusable(), focusable() ];
+				view.element.style.width = '200px';
+
+				view.items.addMany( items );
+
+				console.log( ungroupedItems.map( i => i ).length );
+				console.log( groupedItems.map( i => i ).length );
```
oleq, line 586: These make the CI red.

packages/ckeditor5-ui/src/toolbar/toolbarview.js
mlewand: In case factory has no matching button then an undefined is returned (from the callback above). It happens if you misspell button name in the configuration.

There's at least one existing test that will fail if it's not handled correctly: https://github.com/ckeditor/ckeditor5/blob/e2befc89a888eb86366eda9fd5d4960a4e0c04dd/packages/ckeditor5-ui/tests/toolbar/toolbarview.js#L411-L426

packages/ckeditor5-ui/tests/toolbar/toolbarview.js
mlewand: Fixed in a4e15fd.

PR #7660, opened 2020-07-21
Improved `--snippets` option in documentation

### LINKED ISSUE #7527: What does --snippets option filter?

I was trying to build only the following guide with its snippet: http://127.0.0.1:8080/ckeditor5/20.0.0/examples/builds/classic-editor.html

So, I tried (using the guide name):

```
yarn docs --skip-api --snippets=classic-editor
```

But it throws errors.

I also tried (using the snippet name):

```
yarn docs --skip-api --snippets=build-classic-source
```

But it also didn't work (the snippet was not built).

The documentation for this option does not help: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/development-environment.html#generating-documentation. But I also think that the option itself is not functioning too well since I could not make it work.

pomek:

Docs (ckeditor5): Improved the `--snippets` option in the documentation for the building documentation process. Closes #7527.


scripts/docs/snippetadapter.js
```diff
@@ -80,12 +80,13 @@ module.exports = function snippetAdapter( snippets, options, umbertoHelpers ) {
...
-	if ( options.whitelistedSnippets ) {
-		filterWhitelistedSnippets( snippets, options.whitelistedSnippets );
+	// Remove snippets that do not match to patterns specified in `options.allowedSnippets`.
+	if ( options.allowedSnippets ) {
+		filterAllowedSnippets( snippets, options.allowedSnippets );
 	}
 
-	console.log( `Building ${ snippets.size } snippets...` );
+	console.log( `Found ${ snippets.size } "{@snippet ...}" tags...` );
+	console.log( `Building ${ countUniqueSnippets( snippets ) } snippets...` );
```
pomek, line 89: The first log is about how many `{@snippet ...}` tags Umberto found. The second one is about how many snippets will be built. The same snippet can be used on many pages but it will be built once.

scripts/docs/snippetadapter.js
```diff
@@ -451,6 +452,21 @@ function getHTMLImports( files, mapFunction ) {
...
 }
 
+/**
+ * Returns a number of unique snippet names that will be built.
+ *
+ * @param {Set.<Snippet>} snippets Snippet collection extracted from documentation files.
+ * @returns {Number}
+ */
+function countUniqueSnippets( snippets ) {
+	const uniqueSnippetNames = new Set();
```
Reinmar, line 462: This would be "nicer":

```
new Set( snippets.map( snippet => snippet.snippetName ) )
```

scripts/docs/snippetadapter.js
```diff
@@ -80,12 +80,13 @@ module.exports = function snippetAdapter( snippets, options, umbertoHelpers ) {
...
-	// Remove snippets that do not match to patterns specified in `options.whitelistedSnippets`.
-	if ( options.whitelistedSnippets ) {
-		filterWhitelistedSnippets( snippets, options.whitelistedSnippets );
+	// Remove snippets that do not match to patterns specified in `options.allowedSnippets`.
+	if ( options.allowedSnippets ) {
+		filterAllowedSnippets( snippets, options.allowedSnippets );
 	}
 
-	console.log( `Building ${ snippets.size } snippets...` );
+	console.log( `Found ${ snippets.size } "{@snippet ...}" tags...` );
```
Reinmar, line 88: I'd move this one to the loop above to log it only when specifying `--snippets` as it does not make much sense in a use without this option. Also "matching {@snippet} tags" would be better. Too many ellipsis and without "matching" it actually does not say the trurth as more tags were found.

pomek: Moving the log before the `for..of` loop (as I understood) produces invalid results:

```
Found 132 matching {@snippet} tags.
[ 'classic-editor', 'build-classic-source' ] // --snippets
```

Filtering is done after the loop and that's why the log is here.

```
Found 8 matching {@snippet} tags.
[ 'classic-editor', 'build-classic-source' ] // --snippets
```

pomek:
@Reinmar, I've just pushed minor improvements. Could you take a look once again?

PR #7662, opened 2020-07-21
Fixed incorrect selection fixing in some multi-cell selection scenarios

### LINKED ISSUE #7659: Selection is incorrectly restored in some multi-cell scenarios

Selection fixing, that happens when a selection range goes to the graveyard, works a bit incorrectly at this moment. The algorithm tries to fix every selection range that gets removed. Consider this scenario:

```
[ ][ ][ ]
[x][x][ ]
[x][x][ ]
```

This is a 3x3 table, with cells marked with `x` being selected.

Now, when you remove the second row, you'd expect this result:

```
[ ][ ][ ]
[x][x][ ]
```

The second row is removed and the selection stays on those cells that are still in the content. However, the actual result is this:

```
[ ][ ][x]
[x][x][ ]
```

Why? Well, two selection ranges were removed and got fixed to select the cell in the upper-right corner.

This happens because the selection post-fixer does not check the selection state. I propose to fix this algorithm. Selection range should be fixed only if it is the last range of the selection. In other cases, we should leave the selection as is.

scofalik:

Fix: Fixed incorrect selection fixing in some multi-cell selection scenarios. Closes #7659.

### DIFF
--- a/packages/ckeditor5-engine/src/model/documentselection.js
+++ b/packages/ckeditor5-engine/src/model/documentselection.js
@@ -1120,7 +1120,7 @@ class LiveSelection extends Selection {
 		return attrs;
 	}
 
-	// Fixes the selection after all it ranges get removed.
+	// Fixes the selection after all its ranges got removed.
 	//
 	// @private
 	// @param {module:engine/model/position~Position} deletionPosition Position where the deletion happened.
--- a/packages/ckeditor5-engine/src/model/documentselection.js
+++ b/packages/ckeditor5-engine/src/model/documentselection.js
@@ -567,7 +567,6 @@ mix( DocumentSelection, EmitterMixin );
 //
 // @extends module:engine/model/selection~Selection
 //
-
 class LiveSelection extends Selection {
 	// Creates an empty live selection for given {@link module:engine/model/document~Document}.
 	// @param {module:engine/model/document~Document} doc Document which owns this selection.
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-engine/src/model/documentselection.js
```diff
@@ -1117,36 +1120,20 @@ class LiveSelection extends Selection {
...
-		liveRange.detach();
+	// @param {module:engine/model/position~Position} deletionPosition Position where the deletion happened.
+	_fixGraveyardSelection( deletionPosition ) {
+		// Find a range that is a correct selection range and is closest to the position where the deletion happened.
+		const selectionRange = this._model.schema.getNearestSelectionRange( deletionPosition );
 
 		// If nearest valid selection range has been found - add it in the place of old range.
-		// If range is equal to any other selection ranges then it is probably due to contents
-		// of a multi-range selection being removed. See ckeditor/ckeditor5#6501.
-		if ( selectionRange && !isRangeCollidingWithSelection( selectionRange, this ) ) {
```
scofalik, line 1142: An explanation behind removing `isRangeCollidingWithSelection`: after the changes I propose, the selection can be fixed only if there are no ranges in it, so there is no possibility that there would be any collision. This is also the reason why `splice` is gone -- I've changed the flow a little and now the range is deleted from selection when it is detected that it was removed from the content. So at this moment, there's no need to remove anything, we just need to add the new range.

packages/ckeditor5-engine/src/model/documentselection.js
```diff
@@ -1117,36 +1120,20 @@ class LiveSelection extends Selection {
 		return attrs;
 	}
 
-	// Fixes a selection range after it ends up in graveyard root.
+	// Fixes the selection after all it ranges get removed.
```
scofalik, line 1123:
```suggestion
	// Fixes the selection after all its ranges got removed.
```

packages/ckeditor5-engine/src/model/documentselection.js
jodator: I like how this get simplified.

Issue #7548, opened 2020-07-03
Millisecond glitch on image upload


fredck:
## 📝 Provide detailed reproduction steps (if any)

1.  In Firefox, go to the [base64 upload doc demo](https://ckeditor.com/docs/ckeditor5/latest/features/image-upload/base64-upload-adapter.html).
2.  Upload an image (easer to see with a small one).

### ✔️ Expected result

Image "uploaded" with no glitches.

### ❌ Actual result

A perceivable glitch happens.

## 📃 Other details

*   Browser: Firefox
*   OS: Mac
*   CKEditor version: 20.0.0

I wasn't able to reproduce the problem in Chrome, but it has been reported in GitHub Writer that a similar issue happens in Chrome as well: ckeditor/github-writer#203.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Reinmar:
@fredck, could you test whether this issue occurs when WidgetTypeAround is disabled (you'd need to remove it from the Widget plugin deps)

fredck:
> @fredck, could you test whether this issue occurs when WidgetTypeAround is disabled (you'd need to remove it from the Widget plugin deps)

I've just tried it with a local build of GitHub Writer as it suffers with this issue ([ckeditor/github-writer#203](https://github.com/ckeditor/github-writer/issues/203)). Removing that plugin didn't change anything and the glitch is still there. I confirmed that the plugin is disabled because there was no fancy arrows and horizontal carets anymore around the image.

I wasn't able to test it with the base64 example, as described in this issue, but I believe you guys should be able to easily do so. Thanks!

Reinmar:
Seems to be related to #7633.

Reinmar:
#7633 makes it quite severe.

oleq:
FYI: #7633 is a different issue.

Reinmar:
Thanks for debugging this, @oleq 

What worries me, though, is that two similar issues appeared shortly one after another. Perhaps something else changed too?

Also, besides image upload I can see that normal image paste (image copied from the editor) is affected too. The resize handles are repositioned with a long delay, even 300ms-500ms. This definitely didn't happen before. And as Fred checked, this is not related to WidgetTypeAround. You can even observe the glitch on this gif:

![](https://user-images.githubusercontent.com/156149/88141394-d3da1e00-cbf3-11ea-8e40-ae78d028599a.gif)

[...]

PR #7664, opened 2020-07-21
#7633: Resizer#redraw() should not change the editing view unless a different size should be set

### LINKED ISSUE #7633: Blue border glitch when undoing image resize

1. Resize some image
2. To see it clearly keep pressing `undo` and `redo`

### ❌ Actual result

The blue border stays as its place for a split second and then jumps to fit the size of the image.  
16_image_4

## 📃 Other details

* Browser: any
* CKEditor version: since 17

---

oleq:

Fix: `Resizer#redraw()` should not change the editing view unless a different size should be set. Closes #7633.

### DIFF
--- a/packages/ckeditor5-widget/src/widgetresize/resizer.js
+++ b/packages/ckeditor5-widget/src/widgetresize/resizer.js
@@ -10,6 +10,7 @@
 import View from '@ckeditor/ckeditor5-ui/src/view';
 import Template from '@ckeditor/ckeditor5-ui/src/template';
 import Rect from '@ckeditor/ckeditor5-utils/src/dom/rect';
+import compareArrays from '@ckeditor/ckeditor5-utils/src/comparearrays';
 
 import ObservableMixin from '@ckeditor/ckeditor5-utils/src/observablemixin';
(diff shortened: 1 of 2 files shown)


oleq:
The glitch should be gone now.

oleq:
FYI: #7548 is a different problem.

Reinmar:
The PR looks good to me and I'd be completely fine to merge it but what I described in https://github.com/ckeditor/ckeditor5/issues/7548#issuecomment-662265194 worries me. Why two similar issues appeared so close one another? What's causing the paste/upload glitch? I have a feeling that there's something wrong elsewhere and that a fix for that thing may fix both issues.

Still, I'll be ok with merging this PR, but after we check whether a fix for the other bug will not resolve this too (if we merge this PR it won't be that easy to check).

oleq:
I couldn't reproduce the upload/paste glitch and I [asked the issue author for more details](https://github.com/ckeditor/ckeditor5/issues/7548#issuecomment-662265194). 

In the meantime, this PR addresses not only the resizer glitch issue but also fixes the nasty `View#render()` loop when an image is selected, which could have other unidentified consequences in the editor.

PR #7668, opened 2020-07-22
Improved removing content after a link element

### LINKED ISSUE #7521: After backspacing into a link, the caret should still stay outside

So we changed recently (#1016) the behavior when clicking after the link. You will now be placed "outside" the link.

However, the same isn't true when you backspace to that place.

I think we should be consistent here.

---

pomek:

Fix (link): When removing the content after a link element, the selection will not preserve the `linkHref` attribute (and decorators if specified). However, if the selection is placed in the link element (in the middle or at the end), those attributes will be preserved. Closes #7521.

### DIFF
--- a/packages/ckeditor5-link/src/linkediting.js
+++ b/packages/ckeditor5-link/src/linkediting.js
@@ -444,7 +444,7 @@ export default class LinkEditing extends Plugin {
 	 * the {@link module:typing/twostepcaretmovement~TwoStepCaretMovement} plugin is active and
 	 * the selection has the "linkHref" attribute due to overriden gravity (at the end), the `linkHref` attribute should stay untouched.
 	 *
-	 * The purpose of this action is to allow removing the link label and keep the selection outside the link.
+	 * The purpose of this action is to allow removing the link text and keep the selection outside the link.
 	 *
 	 * See https://github.com/ckeditor/ckeditor5/issues/7521.
 	 *
--- a/packages/ckeditor5-link/src/linkediting.js
+++ b/packages/ckeditor5-link/src/linkediting.js
@@ -436,11 +436,11 @@ export default class LinkEditing extends Plugin {
 
 	/**
 	 * Starts listening to {@link module:engine/model/model~Model#deleteContent} and checks whether
-	 * removing a content after a link element.
+	 * removing a content right after the "linkHref" attribute.
 	 *
 	 * If so, the selection should not preserve the `linkHref` attribute. However, if
-	 * {@link module:typing/twostepcaretmovement~TwoStepCaretMovement} plugin is activated and
-	 * the selection is inside the link, at the end, the `linkHref` attribute should stay untouched.
+	 * the {@link module:typing/twostepcaretmovement~TwoStepCaretMovement} plugin is active and
+	 * the selection has the "linkHref" attribute due to overriden gravity (at the end), the `linkHref` attribute should stay untouched.
 	 *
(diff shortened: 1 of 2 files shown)


oleq:
@ckeditor/qa-team Can you tak a look at this PR?

packages/ckeditor5-link/src/linkediting.js
```diff
@@ -438,6 +433,92 @@ export default class LinkEditing extends Plugin {
...
+
+	/**
+	 * Starts listening to {@link module:engine/model/model~Model#deleteContent} and checks whether
+	 * removing a content after a link element.
+	 *
+	 * If so, the selection should not preserve the `linkHref` attribute. However, if
+	 * {@link module:typing/twostepcaretmovement~TwoStepCaretMovement} plugin is activated and
+	 * the selection is inside the link, at the end, the `linkHref` attribute should stay untouched.
+	 *
+	 * The purpose of this action is to allow removing the link label and keep the selection outside the link.
```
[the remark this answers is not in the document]
pomek: I meant a text that described the URL `<a href="url">link label</a>`.

packages/ckeditor5-link/src/linkediting.js
```diff
@@ -116,6 +117,9 @@ export default class LinkEditing extends Plugin {
 
 		// Handle typing over the link.
 		this._enableTypingOverLink();
+
+		// Handle removing the content after the link element.
+		this._enableRemovingLinkAttributesWhileRemovingEndOfLink();
```
jodator, line 122: The comment doesn't match the method name. The method name is also long.

For me this method is \[enabling the\] handling what happens when user delete content after a link. So maybe

```
this._handleDeleteContentAfterLink();
```

Especially after reading method docs - it does that (with addition to handling selection attributes, etc). Also, the name might stay if we decide to handle also other delete content cases.

FilipTokarski:
Looks ok 👍

pomek:
@jodator, if I remove `del` checks, after pressing `Del` in the following structure:

```html
<paragraph>Foo <$text linkHref="url">Bar</$text>[ ]</paragraph>
```

link attributes will be removed from the selection as well. Are we fine with that?

packages/ckeditor5-link/src/linkediting.js
```diff
@@ -438,6 +435,93 @@ export default class LinkEditing extends Plugin {
...
+
+	/**
+	 * Starts listening to {@link module:engine/model/model~Model#deleteContent} and checks whether
+	 * removing a content right after the "linkHref" attribute.
+	 *
+	 * If so, the selection should not preserve the `linkHref` attribute. However, if
+	 * the {@link module:typing/twostepcaretmovement~TwoStepCaretMovement} plugin is active and
+	 * the selection has the "linkHref" attribute due to overriden gravity (at the end), the `linkHref` attribute should stay untouched.
+	 *
+	 * The purpose of this action is to allow removing the link label and keep the selection outside the link.
```
jodator, line 447:
```suggestion
	 * The purpose of this action is to allow removing the link text and keep the selection outside the link.
```

will that be OK?

jodator:
> link attributes will be removed from the selection as well. Are we fine with that?

AFAICS it would behave the same way as <kbd>backspace</kbd> so it is OK for me.

If you see that we can squeeze that change with <kbd>del</kbd> quickly then I'd do that. If not it's fine to make that a follow up (https://github.com/ckeditor/ckeditor5/issues/7521#issuecomment-662465385).

PR #7678, opened 2020-07-23
Increase specificity of code styles to enforce 2SCM highlight in the …

### LINKED ISSUE #7671: Active <code> element styles do not work in docs
### ✔️ Expected result

The code element should be highlighted.

### ❌ Actual result

Nothing happens.

---

jodator:

Docs (code): Increase the specificity of code styles to enforce 2SCM highlight in the docs. Closes #7671.

---

### Additional information

Just to double check if it is a proper solution:

Screenshot from 2020-07-23 11-57-05

### DIFF
--- a/docs/assets/snippet-styles.css
+++ b/docs/assets/snippet-styles.css
@@ -53,6 +53,15 @@ It breaks CKEditor 5 (see how <p><code>[]</code></p> looks). */
 	display: none !important; /* Firefox is very stubborn. */
 }
 
+/* Umberto has some high specificity rules - for code highlight we need even more.. */
+.ck.ck-content code:not(.hljs) {
+	background-color: hsla(0, 0%, 78%, 0.3);
+}
+
+.ck-content code.ck-code_selected:not(.hljs) {
+	background-color: hsla(0, 0%, 78%, 0.5);
+}
+
 /* Restore default browser styles for <sub> and <sup>. */
 .ck.ck-content sub {
 	vertical-align: sub;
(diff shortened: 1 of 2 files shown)


Reinmar:
> IMO what we need is (this is regardless of this PR, it is simply wrong):

This is not a content style. This is part of the editing-layer styles. So it should not use `.ck-content`.

oleq:
> This is not a content style. This is part of the editing-layer styles. So it should not use `.ck-content`.

A good catch. `.ck-editor__editable` then.

oleq:
As for this selector in Umberto, it goes back to cksource/umberto@6bc3441.

Probably the commit author wanted to filter all codes that are rendered in HLJS snippets out. So he changed `code` to `code:not( .hljs )`. Then I added `:not( pre ) >` to filter out `<code>` inside editor content (when the \`CodeBlock\` plugin first came out). And since all `code.hljs` are inside `pre` anyway, it could be that we don't need the original `code:not( .hljs )` fix.

Reinmar:
I changed the base of this PR to `release`.

Reinmar:
@oleq, so what you're proposing:

*   Let's check why this style was written in that way first (git blame may help).
*   Let's try to write it differently (how?).
*   Finally, let's verify that we didn't break anything.

oleq:
I just checked it again and it looks like 

```diff
diff --git a/packages/ckeditor5-basic-styles/theme/code.css b/packages/ckeditor5-basic-styles/theme/code.css
index 824be69b55..f689d3f574 100644
--- a/packages/ckeditor5-basic-styles/theme/code.css
+++ b/packages/ckeditor5-basic-styles/theme/code.css
@@ -9,6 +9,6 @@
          border-radius: 2px;
 }
 
-.ck .ck-code_selected {
+.ck.ck-editor__editable .ck-code_selected {
          background-color: hsla(0, 0%, 78%, 0.5);
 }
```

is the only change we need.

PR #7680, opened 2020-07-23
#7679: The config.image.resizeOptions should have the default value.

### LINKED ISSUE #7679: Image resizeOptions config needs defaults

ATM when you just add the resize buttons or a dropdown to the image toolbar nothing happens because these components don't register unless `config.image.resizeOptions` is defined. All you get is the component factory error which does not tell the full story.

Besides, all our features come with the default configuration and so should `ImageResize`.

---

oleq:
## Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Internal (image): The `config.image.resizeOptions` should have the default value. Closes #7679.

### DIFF
--- a/packages/ckeditor5-image/src/imageresize/imageresizeediting.js
+++ b/packages/ckeditor5-image/src/imageresize/imageresizeediting.js
@@ -29,6 +29,37 @@ export default class ImageResizeEditing extends Plugin {
 	/**
 	 * @inheritDoc
 	 */
+	constructor( editor ) {
+		super( editor );
+
+		editor.config.define( 'image', {
+			resizeUnit: '%',
+			resizeOptions: [ {
+				name: 'imageResize:original',
+				value: null,
+				icon: 'original'
+			},
+			{
+				name: 'imageResize:25',
+				value: '25',
+				icon: 'small'
+			},
+			{
+				name: 'imageResize:50',
+				value: '50',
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-image/src/imageresize/imageresizeediting.js
```diff
@@ -26,6 +26,37 @@ export default class ImageResizeEditing extends Plugin {
 		return 'ImageResizeEditing';
 	}
 
+	/**
+	 * @inheritDoc
+	 */
+	constructor( editor ) {
+		super( editor );
+
+		editor.config.define( 'image', {
```
jodator, line 35: WDYT about moving this to `ImageResizeButtons` - the configuration options are defined there (the docs) and is only used by this plugin.

oleq: So default unit stays in editing and `resizeOptions` go to `ImageResizeButtons`? Sounds like a plan.

jodator: Hmm... well :man\_shrugging: for me TBH. Different sub-features defines different configuration options. But since it's already messy we can leave that as is.

PR #7681, opened 2020-07-23
Fix error in function documentation

marcellofuschi:
You use the `toWidget` function in "editingDowncast", not in "dataDowncast".


Reinmar:
:facepalm:  Thank you! And sorry for the confusion it must've caused.

PR #7733, opened 2020-07-28
Fix for data-to-marker upcast for elements that are not converted

### LINKED ISSUE #7732: Data-to-marker upcast crashes if view element has not been converted

In this iteration, we are introducing data-to-marker conversion. I found a bug in it. The new conversion method works partially like attribute conversion, as marker data is set as an attribute if the marker starts/ends before/after an element. The converter expects `data.modelRange` to work properly. However, if the attribute is set on an element that does not have its converter, `data.modelRange` is not available and the conversion crashes.

We already handle this scenario in attribute-to-attribute conversion. If that case happens, the converter first converts the children of the upcasted element and then sets the attribute. We will need to do the same here.

scofalik:

Internal: Fix for data-to-marker upcast for elements that are not converted. Closes #7732.

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -653,6 +653,13 @@ function upcastAttributeToMarker( config ) {
 	return ( evt, data, conversionApi ) => {
 		const attrName = `data-${ config.view }`;
 
+		// Since we are converting to attribute we need an range on which we will set the attribute.
+		// If the range is not created yet, we will create it.
+		if ( !data.modelRange ) {
+			// Convert children and set conversion result as a current data.
+			data = Object.assign( data, conversionApi.convertChildren( data.viewItem, data.modelCursor ) );
+		}
+
 		if ( conversionApi.consumable.consume( data.viewItem, { attributes: attrName + '-end-after' } ) ) {
 			addMarkerElements( data.modelRange.end, data.viewItem.getAttribute( attrName + '-end-after' ).split( ',' ) );
 		}
--- a/packages/ckeditor5-engine/tests/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/tests/conversion/upcasthelpers.js
@@ -861,6 +861,18 @@ describe( 'UpcastHelpers', () => {
 				{ name: 'g:foo', start: [ 0, 3 ], end: [ 0, 3 ] }
 			);
 		} );
+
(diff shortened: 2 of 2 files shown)


packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -653,6 +670,14 @@ function upcastAttributeToMarker( config ) {
 	return ( evt, data, conversionApi ) => {
 		const attrName = `data-${ config.view }`;
 
+		// This converter wants to add a model element, marking a marker, before/after an element (or maybe even group of elements).
+		// To do that, we can use `data.modelRange` which is set on an element (or a group of elements) that has been upcasted.
+		// But, if the processed view element has not been upcasted yet (it does not have a converted), we need to
+		// fire conversion for its children first, then we will have `data.modelRange` available.
+		if ( !data.modelRange ) {
+			data = Object.assign( data, conversionApi.convertChildren( data.viewItem, data.modelCursor ) );
+		}
```
scofalik, line 679: This is the actual "fix" part.

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -634,8 +636,23 @@ function upcastDataToMarker( config ) {
...
+		// To have it work, we need to some extra processing for priority for attribute converter.
+		// Priority `'low'` value should be the base value and then we will change it depending on `config.converterPriority` value.
+		//
+		// This hack probably would not be needed if attributes are upcasted separately.
+		//
+		const basePriority = priorities.get( 'low' );
+		const maxPriority = priorities.get( 'highest' );
+		const priorityFactor = priorities.get( config.converterPriority ) / maxPriority; // Number in range [1, -1].
+
+		dispatcher.on( 'element', upcastAttributeToMarker( config ), { priority: basePriority + priorityFactor } );
```
scofalik, line 655: This came up when I was fixing a test and realized that overwriting converters won't work properly.

scofalik: I don't like this hack. But it seems to work.

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
scofalik: After adding this part, though, test `can be overwritten using converterPriority` started to fail because `high` priority was used and this converter was fired before paragraph converter.

scofalik:
Something went wrong with rebasing... Do not ask...

PR #7739, opened 2020-07-29
#7727: Multi-cell table selection should work properly with single cell selected

### LINKED ISSUE #7727: Multi-cell selection with arrows+shift breaks when trying to go in opposite direction on single row/column

1. Select a cell
2. Hold `shift` + click `arrow right` to create a multicell selection
3. Still holding `shift` start clicking `arrow left`, to go in opposite direction

### ✔️ Expected result

You can create multi-cell selection going in both directions

### ❌ Actual result

Once you go in one direction, if you then go in opposite direction and reach the single original cell, multi-cell selection breaks.
Screencasts:  
  
28_table1
  
  
28_table2

## 📃 Other details
However, it works ok if you already have some multiple cells selected. Then you can go in any direction:  
  
28_table3

* Browser: Chrome, Firefox, Edge
* No errors in the console
* CKEditor version: introduced in 20.0.0, works fine in 19.1.0

---

niegowski:

Tests (table): Fixed tests for multi-cell table selection with single cell selected. See #7727.

---

### Additional information

The issue was fixed in #7770.


niegowski:
@jodator what do you think about introducing changes like this? https://github.com/ckeditor/ckeditor5/issues/7783#issuecomment-669143787

jodator:
> @jodator what do you think about introducing changes like this? [#7783 (comment)](https://github.com/ckeditor/ckeditor5/issues/7783#issuecomment-669143787)

Not in this PR.

packages/ckeditor5-table/src/tablekeyboard.js
```diff
@@ -51,10 +51,18 @@ export default class TableKeyboard extends Plugin {
...
 
+		const priorityAboveWidget = priorities.get( 'high' ) + 10;
+		const priorityBelowWidget = priorities.get( 'high' ) - 10;
+
+		// This listener is handling cases inside table cell.
 		// Note: This listener has the "high-10" priority because it should allow the Widget plugin to handle the default
 		// behavior first ("high") but it should not be "prevent–defaulted" by the Widget plugin ("high-20") because of
 		// the fake selection retention on the fully selected widget.
-		this.listenTo( viewDocument, 'keydown', ( ...args ) => this._onKeydown( ...args ), { priority: priorities.get( 'high' ) - 10 } );
+		this.listenTo( viewDocument, 'keydown', ( ...args ) => this._onKeydown( false, ...args ), { priority: priorityBelowWidget } );
```
jodator, line 61: Boolean handling in this class is getting out of control. Basically every method has `true` / `false` in parameters and this is getting hard to read.

I don't know what this action does `onKeydown( false, ...args )` is very cryptic. If there are two distinguish actions then it should be reflected in the method name.

I would see here two methods:

```javascript
this.listenTo( viewDocument, 'keydown', ( ...args ) => this._handleKeydownOutsideTable( ...args ), ... );
this.listenTo( viewDocument, 'keydown', ( ...args ) => this._handleKeydownInsideTable( ...args ), ... );
```

Alternative might be a method that would return handler:

```javascript
this.listenTo( viewDocument, 'keydown', this._getKeydownOutsideTableHandler(), { ... } );
this.listenTo( viewDocument, 'keydown', this._getKeydownInsideTableHandler(), { ... } );
```

Both give more information about the intention of the method.

The other problem is that `highPriority` as a method parameter does not tell what this flag do.

jodator: The other solution could be string - less refactoring:

```javascript
this.listenTo( viewDocument, 'keydown', this._getKeyDownHandler( 'outside' ), { ... } );
this.listenTo( viewDocument, 'keydown', this._getKeyDownHandler( 'inside' ), { ... } );
```

But still, I found separate handlers better in terms of separation of concerns here.

niegowski: Done :)

packages/ckeditor5-table/src/tablekeyboard.js
```diff
@@ -377,3 +367,22 @@ export default class TableKeyboard extends Plugin {
 	}
 }
 
+// Returns the keydown event handler wrapper.
+function wrapKeydownHandler( editor, callback ) {
```
jodator, line 378: Looks like, you're reading my mind here :+1:

niegowski:
This PR seems to be not needed since #7770 fixed problem with (`isObject` vs `isSelectable`) so the widget isn't taking over while a single table cell is selected. Still, I think that we should update tests to properly verify this scenario. WDYT?

jodator:
> This PR seems to be not needed since #7770 fixed problem with (`isObject` vs `isSelectable`) so the widget isn't taking over while a single table cell is selected. Still, I think that we should update tests to properly verify this scenario. WDYT?

The less code we have the better :+1: 

The more tests we have the better :+1: 

So if reverting the code and keeping the tests will worker after the #7770 then I think that is the way to go.

niegowski:
I removed code changes and left only tests to properly check that problematic issue. There were tests for this case but `VirtualTestEditor` wasn't using `Widget` so its handlers weren't attached.

PR #7744, opened 2020-07-29
Detect target inside a shadow root for clickOutsideHandler

### LINKED ISSUE #7743: Detect target inside a shadow root for clickOutsideHandler

`clickOutsideHandler` currently works by checking whether the click target is one of the context elements. However, when the editor is placed in a shadow root, any click events on the editor are retargeted such that they appear to target the entire editor and not a specific element within the editor. 

Because of event retargeting, `clickOutsideHandler` doesn't function as expected when the editor is in a shadow root. For example, `clickOutsideHandler` is used to detect when a click outside a dropdown view occurs so that the dropdown view can be collapsed. However, in a shadow root, an event clicking on the dropdown view itself is retargeted so that `target` is not the dropdown view, and the click triggers dropdown view to be closed. Looking at the event's composed path instead of just `target` can reveal whether the original target included the context element(s). 
## 📃 Other details
Improving this would be a step in supporting running the editor in shadow DOM (#3891), making a Web Component for CKEditor more possible (#1483)! 

---

ywsang:

Other (ui): Look at event's composed path to detect a target in shadow root. Closes #7743. 

---

### Additional information


Reinmar:
Thanks for the PR and the issue report. Do you think you'd be able to add a test to this PR?

ywsang:
I'll look into it!

packages/ckeditor5-ui/src/bindings/clickoutsidehandler.js
```diff
@@ -25,13 +25,14 @@
...
-	emitter.listenTo( document, 'mousedown', ( evt, { target } ) => {
+	emitter.listenTo( document, 'mousedown', ( evt, domEvt ) => {
 		if ( !activator() ) {
 			return;
 		}
 
 		for ( const contextElement of contextElements ) {
-			if ( contextElement.contains( target ) ) {
+			if ( contextElement.contains( domEvt.target ) ||
+						( 'composedPath' in domEvt && domEvt.composedPath().includes( contextElement ) ) ) {
```
justinfagnani, line 35: Maybe add a comment on why you're checking for the existence of `composedPath`. I'm not sure how this project tracks browser support, but they may want to remove then when all supported browsers have native shadow DOM.

Reinmar: A comment would great! We don't unfortunately track such things in a more organized way yet, so a comment is the best we can do for now.

pomek:
Since I am not able to apply minor changes, I must close the PR and create another one.

image

pomek:
=> https://github.com/ckeditor/ckeditor5/pull/7779

PR #7756, opened 2020-07-31
Docs: Fixes to image guide. [skip ci]

godai78:
CKEditor 5 image feature and image alignment docs - fixes and minor changes. Scope: guide, snippets.  
  
Still missing: live example for defining custom styles.

### DIFF
--- a/packages/ckeditor5-image/docs/_snippets/features/image-resize-buttons-dropdown.html
+++ b/packages/ckeditor5-image/docs/_snippets/features/image-resize-buttons-dropdown.html
@@ -1,5 +1,5 @@
 <div id="snippet-image-resize-buttons-dropdown">
-	<h3>Resize me using the dropdown!</h3>
+	<h3>Click me and resize using the pop-up menu dropdown!</h3>
 
 	<figure class="image">
 		<img src="%BASE_PATH%/assets/img/fields.jpg" alt="Autumn fields">
--- a/packages/ckeditor5-image/docs/_snippets/features/image-resize-buttons-dropdown.html
+++ b/packages/ckeditor5-image/docs/_snippets/features/image-resize-buttons-dropdown.html
@@ -1,5 +1,5 @@
 <div id="snippet-image-resize-buttons-dropdown">
-	<h3>Click me and resize using the pop-up menu dropdown!</h3>
+	<p>Click me and resize using the pop-up menu dropdown!</p>
 
 	<figure class="image">
 		<img src="%BASE_PATH%/assets/img/fields.jpg" alt="Autumn fields">
--- a/packages/ckeditor5-image/docs/_snippets/features/image-resize-buttons-dropdown.html
+++ b/packages/ckeditor5-image/docs/_snippets/features/image-resize-buttons-dropdown.html
@@ -1,5 +1,5 @@
 <div id="snippet-image-resize-buttons-dropdown">
-	<p>Click me and resize using the pop-up menu dropdown!</p>
+	<p>Click me and resize using the contextual toolbar!</p>
 
 	<figure class="image">
 		<img src="%BASE_PATH%/assets/img/fields.jpg" alt="Autumn fields">
--- a/packages/ckeditor5-image/docs/_snippets/features/image-style-presentational.html
+++ b/packages/ckeditor5-image/docs/_snippets/features/image-style-presentational.html
@@ -11,5 +11,5 @@
 		<img src="%BASE_PATH%/assets/img/fields.jpg" alt="Autumn fields">
 	</figure>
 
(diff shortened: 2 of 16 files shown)


AnnaTomanek:
It looks like the image resize API docs have never been reviewed before :wink: 

@godai78 can you have a look and correct the API docs comments in these files (ckeditor5-image/src/imageresize)

AnnaTomanek:
Snippets: I'm fine with using the "meta" content inside, at least for the time being until we figure out something more interesting. However, when reviewing the snippets, I noticed that at some point, when adding the resize and link feature, we deviated a bit, and for consistency reasons, a few things could be changed:

*   [ ] There is no reason to use `<h3>` elements inside, `<p>` was fine.
*   [ ] As a rule, we should not have a situation where the only instructions are inside the editor and not in the guide text. For all features except for resize and link this was done correctly, but now we are missing this sentence or two in the description that would serve as an introduction or instruction to the demo.

packages/ckeditor5-image/docs/_snippets/features/image-style-presentational.html
```diff
@@ -11,5 +11,5 @@
 		<img src="%BASE_PATH%/assets/img/fields.jpg" alt="Autumn fields">
 	</figure>
 
-	<p>Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here.</p>
+	<p>Yet another presentation paragraph and lots of text here, again. Quite a lot of text, just like before. And all the same all this amount of text forms a whole long paragraph. This whole lot of text is here to help you see the image's alignment in context. It serves no other purpose, though. But we still find it useful and plain like it. And it is even longer than the previous one. Go on an use the style buttons to make the most out of this pretty text block and the image alignment demo.</p>
```
AnnaTomanek, line 14:
```suggestion
	<p>Yet another sample paragraph and lots of text here, again. Quite a lot of text, just like before. And this huge amount of text forms a whole long paragraph. This whole lot of text is here to help you see the image's alignment in context. It serves no other purpose, though. But we still find it useful and plain like it. And it is even longer than the previous one. Go on an use the image style buttons to make the most out of this pretty text block and the image alignment demo.</p>
```

packages/ckeditor5-image/docs/_snippets/features/image-style.html
```diff
@@ -11,5 +11,5 @@
 		<img src="%BASE_PATH%/assets/img/fields.jpg" alt="Autumn fields">
 	</figure>
 
-	<p>Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here. Lots of text here.</p>
+	<p>Lots of text here. Quite a lot of text, indeed. In fact there is plenty of text beside this image and it forms a whole long paragraph. This whole lot of text is here to help you see the image's alignment in context. It serves no other purpose, though. But we still find it useful and plain like it.</p>
```
AnnaTomanek, line 14:
```suggestion
	<p>Lots of text here. Quite a lot of text, indeed. In fact, there is plenty of text beside this image and it forms a whole long paragraph. This whole lot of text is here to help you see the image's alignment in context. It serves no other purpose, though. But we still find it useful and plain like it.</p>
```

AnnaTomanek:
@godai78 You've accidentally committed the `yarn.lock` file which as a rule you should never do. You can revert it by running `git reset stable yarn.lock` on the `image-options` branch and then committing the change to `image-options`. Thanks @pomek :heart:

packages/ckeditor5-image/docs/_snippets/features/image-resize-buttons-dropdown.html
```diff
@@ -1,5 +1,5 @@
 <div id="snippet-image-resize-buttons-dropdown">
-	<h3>Resize me using the dropdown!</h3>
+	<p>Click me and resize using the pop-up menu dropdown!</p>
```
AnnaTomanek, line 2: Terminology, here and somewhere else: what you call the "pop-up menu" is what is called the "contextual toolbar" in CKEditor 5 world. As a general UI element it's linked to the feature itself (so: in this case, it's the "image contextual toolbar") and it replaces dialog windows of the past.

packages/ckeditor5-image/docs/features/image.md
```diff
@@ -446,6 +481,7 @@ ClassicEditor
 	.then( ... )
 	.catch( ... );
 ```
+Check out the differnce in this live demo below.
```
AnnaTomanek, line 484: Typo.

godai78: Pushed.

PR #7760, opened 2020-07-31
Fix mistake in documentation

marcellofuschi:
In order to be able to use the "super build" the way it is shown in the sample script (i.e. by creating the editor with `CKEDITOR.ClassicEditor`) you need to have the editor classes not exported using `export default { ... }` but just `export { ... }`. The reason is that, by using the former approach, you get the classes stored in the `default` property of `CKEDITOR` (and not in `CKEDITOR` itself).

Schermata 2020-07-31 alle 22 55 34

### DIFF
--- a/docs/builds/guides/integration/advanced-setup.md
+++ b/docs/builds/guides/integration/advanced-setup.md
@@ -576,7 +576,7 @@ const config = {
 ClassicEditor.defaultConfig = config;
 InlineEditor.defaultConfig = config;
 
-export default {
+export {
 	ClassicEditor, InlineEditor
 };
 ```


Reinmar:
Thanks for pointing this out! @ckeditor/qa-team could you double-check it?

Mgsy:
We've checked it with @FilipTokarski and after following our Superbuilds guide, editor classes are properly stored in `CKEDITOR`. 



Additionally, removing `default` from the export breaks the integration and results in errors:

```
(index):31 Uncaught TypeError: Cannot read property 'ClassicEditor' of undefined
```

marcellofuschi:
I think there's some mismatch between the environments we're using. I suspect the reason for my code not working is that mine is an "ES module" (see `__esModule: true` in the screenshot) while yours is not.
Will try to get back to you with a clearer view.

Reinmar:
Thanks!

marcellofuschi:
Switched back to `export default` in my code and it works after adding `libraryExport: 'default'` (https://webpack.js.org/configuration/output/#outputlibraryexport) to my `webpack.config.js` file.

pomek:
Based on the last comment:

> Switched back to `export default` in my code and it works after adding `libraryExport: 'default'` ([webpack.js.org/configuration/output/#outputlibraryexport](https://webpack.js.org/configuration/output/#outputlibraryexport)) to my `webpack.config.js` file.

I guess, the PR is no longer valid. I'll close it.

PR #7770, opened 2020-08-04
#6432: Introduced SchemaItemDefinition#isSelectable and SchemaItemDefinition#isContent.

### LINKED ISSUE #6432: Should tableCell be defined as object in schema?
## 📝 Provide a description

I am not sure if `tableCell` element should be "an object" (`isObject: true` in schema). Looking at description: https://ckeditor.com/docs/ckeditor5/latest/api/module\_engine\_model\_schema-SchemaItemDefinition.html an object is:

> "self-contained" and should be treated as a whole. Examples of object elements: `image`, `table`, `video`, etc. **Note:** An object is also a limit, so [`isLimit()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-Schema.html#function-isLimit) returns `true` for object elements automatically.

So I am not sure if it applies. I'd add to that, that for me, "an object" is something that carries an information on it's own. For example a video, or image. It can be empty, with no other content/data and it still is meaningful.

Current situation messes up some algorithms.

First, `getNearestSelectionRange()` selects a cell, while I think that even in this situation:

```plaintext
<table>
    <tableRow>
        <tableCell><paragraph></paragraph></tableCell>[]
    </tableRow>
</table>
```

The selection range should be

```plaintext
<table>
    <tableRow>
        <tableCell><paragraph>[]</paragraph></tableCell>
    </tableRow>
</table>
```

Not

```plaintext
<table>
    <tableRow>
        [<tableCell><paragraph></paragraph></tableCell>]
    </tableRow>
</table>
```

(even with table selection turned on).

Second, `Model#hasContent()` will return `true` when an empty cell is selected (like in the example above).

I think it should be a `isLimit` + `isBlock`.

### LINKED ISSUE #7631: Review Schema element properties

We're working on introducing `isSelectable` in #6432 that'll be like `isObject` but without being a content itself.

This made me think that `isObject` is actually a combination of other properties:

*   `isLimit` as it cannot be split
*   `isSelectable` as it can be selected
*   `isContent` as it is a content (as in `model.hasContent()`)

The last one does not exist yet, so the proposal is to introduce it. Nearly all the algorithms in the engine (like the selection post-fixer) and other features should then use these 3 base properties as they represent the actual behavior, while `isObject` would be just a shorthand to them 3.

Thanks to splitting these properties, we'll gain better flexibility. The #6432 proved that sometimes you need just `isLimit` and `isSelectable`. I can also imagine using other combinations like `isLimit` and `isContent` for widget-like things that should not be selectable. Or, one day, `isInline` and `isContent` but no `isLimit` and `isSelectable`.

BTW, to clarify, `$text` would be `isContent`.

So, the question is – do we see some other candidates to be reviewed like that? One guess would be `isBlock` because it seems to have a rather wide application right now (paragraphs, but also images or tables). I don't see issues with it yet, but maybe you'll have some ideas.

---

oleq:

Feature (engine): Introduced the `SchemaItemDefinition#isSelectable` property. Closes #6432.

Feature (engine): Introduced the `SchemaItemDefinition#isContent` property. Closes #7631.

Other (table): The `tableCell` model element brought by the `TableEditing` plugin is no longer an object (`SchemaItemDefinition#isObject`) in the `Schema` but a selectable (`SchemaItemDefinition#isSelectable`) (see #6432).

Internal (ui): Aligned the `BalloonToolbar` plugin behavior to the new  `SchemaItemDefinition#isSelectable` property (see #6432).

Docs (engine): Extended the "Schema" deep dive guide with the new properties and methods (see #6432, #7631).

Tests (image): Aligned tests to the fact that the `tableCell` model element is no longer an object but a selectable in the schema (see #6432).

Tests (horizontal-line): Aligned tests to the fact that the `tableCell` model element is no longer an object but a selectable in the schema (see #6432).

MINOR BREAKING CHANGE (table): The `tableCell` model element brought by the `TableEditing` plugin is no longer an object (`SchemaItemDefinition#isObject`) in the `Schema` but a selectable (`SchemaItemDefinition#isSelectable`). Please update your integration code accordingly (see #6432).

---

### Additional information

*   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-engine/src/model/utils/insertcontent.js#L243-L244 -> `isSelectable()`?
*   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-engine/src/model/utils/modifyselection.js#L108 -> `isSelectable()`?
*   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-engine/src/model/utils/selection-post-fixer.js#L181-L184 -> `isSelectable()`?
*    https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-horizontal-line/src/horizontallinecommand.js#L97 -> `isSelectable()`?  
      
    On one hand, it makes sense because it would disable the command when a single table cell is selected (enabled ATM in this PR).  
      
    OTOH, what if in the future we allow selecting some containers from the outside and converting them into a horizontal line when they are selected makes perfectly sense?  
      
    *    https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-heading/src/headingcommand.js#L95 same as above, except for headings.
    *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-image/src/image/utils.js#L150 same as above, except for images.
    *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-list/src/listcommand.js#L310 same as above, except for lists.
    *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-page-break/src/pagebreakcommand.js#L97 same as above, except for page breaks
    *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-paragraph/src/paragraphcommand.js#L72 ...and paragraphs,


oleq:
F2F call summary:

> https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-engine/src/model/utils/insertcontent.js#L243-L244 -> `isSelectable()`?

waiting for confirmation from @Reinmar 

> https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-engine/src/model/utils/modifyselection.js#L108 -> `isSelectable()`?

Yes.

> https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-engine/src/model/utils/selection-post-fixer.js#L181-L184 -> `isSelectable()`?

Yes.

> https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-horizontal-line/src/horizontallinecommand.js#L97 -> `isSelectable()`?  
> On one hand, it makes sense because it would disable the command when a single table cell is selected (enabled ATM in this PR).  
> OTOH, what if in the future we allow selecting some containers from the outside and converting them into a horizontal line when they are selected makes perfectly sense?
> 
> *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-heading/src/headingcommand.js#L95 same as above, except for headings.
> *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-image/src/image/utils.js#L150 same as above, except for images.
> *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-list/src/listcommand.js#L310 same as above, except for lists.
> *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-page-break/src/pagebreakcommand.js#L97 same as above, except for page breaks
> *   https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-paragraph/src/paragraphcommand.js#L72 ...and paragraphs,

No.

Reinmar:
> https://github.com/ckeditor/ckeditor5/blob/i/6432-schema-is-selectable/packages/ckeditor5-engine/src/model/utils/insertcontent.js#L243-L244 -> `isSelectable()`?
> 
> waiting for confirmation from @Reinmar 

This one is tricky. We actually want to check if it's:

*   a content, as if it's not content itself it could be ignored and only its children inserted in that place – as in `_handleDisallowedNode`,
*   a limit, as it also defines that the element cannot be split (or otherwise ignored),
*   selectable, asc we'll most likely try to select it.

Therefore, the current `isObject()` check is good. WITH ONE condition – that `isObject()` works this way:

```
isObject( item ) {
    const def = this.getDefinition( item );
    
    return def.isObject || ( def.isLimit && def.isSelectable && def.isContent );
}
```

You may remember that I already mentioned that `isObject()` should also return true if you set the 3 granular options to `true`. It'd definitely be safer in this case if we did that.

oleq:
Let's have another round of review then.

PR #7773, opened 2020-08-04
Introduced additional conversion options for downcast conversion.

### LINKED ISSUE #7628: Add options to getData() and DataController#get()

There are scenarios in which, besides the default content returned/accepted by get/setData() we want to output an alternative form of the content. E.g. with previews of some media or without user suggestions.

The options passed to these methods should be available throughout the upcast conversion of all type of nodes (elements and text nodes and document fragments).

---

oskarwrobel:

Feature: Introduced additional conversion options for downcast conversion. Closes #7628.

---

### Additional information


packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -151,10 +151,11 @@ export default class DataController {
 	 * @param {String} [options.trim='empty'] Whether returned data should be trimmed. This option is set to `empty` by default,
 	 * which means whenever editor content is considered empty, an empty string will be returned. To turn off trimming completely
 	 * use `'none'`. In such cases exact content will be returned (for example `<p>&nbsp;</p>` for an empty editor).
+	 * @param {Object} [options.conversionOptions] Additional, custom configuration passed to the conversion process.
```
scofalik, line 154: Okay, let's talk about this. We can have one of three:

1.  The way you implemented it (and we think about the name).
2.  Having just one flat object (`{ root: 'main', showSuggestions: true, foo: 'bar' }`).
3.  Having two parameters (and we think about names).

What I don't like about current solution is that you need to specify additional "level" of object if you want to pass "conversion option":

`editor.getData( { conversionOptions: { showSuggestions: true } } )`

I don't like it. Let's talk about other options.

Having two parameters isn't perfect either:

`editor.getData( {}, { showSuggestions: true } )`

Having it flat is a bit messy and may be difficult to document, but would be the easiest to use:

`editor.getData( { showSuggestions: true } )`  
`editor.getData( { root: 'other' } )`

scofalik: And when it comes to name, it should make sense from the user's perspective. `conversionOptions` doesn't make a lot of sense, especially if someone is not a programmer, does not know CKE5 internals and simply integrates editor and wants to use `.getData()`.

`resultOptions`, `resultConfiguration`, `dataOptions`, `dataConfiguration` is the direction I'd go.

Or just having it all in a one, flat options object :shrug:

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -166,9 +165,12 @@ export default class DowncastDispatcher {
 	 * @fires attribute
 	 * @param {module:engine/model/range~Range} range The inserted range.
 	 * @param {module:engine/view/downcastwriter~DowncastWriter} writer The view writer that should be used to modify the view document.
+	 * @param {Object} [options={}] Additional, custom configuration passed to the converters through
+	 * {module:engine/conversion/downcastdispatcher~DowncastConversionApi conversion api}. Used only with `dataDowncast` group.
 	 */
-	convertInsert( range, writer ) {
+	convertInsert( range, writer, options = {} ) {
```
scofalik, line 171: I am not sure I like this parameter here -- because you have to then document it the way you did. Maybe we could handle it somehow differently?

For example, set a property of `conversionApi` at the beginning of `.get()` and clear at the end? Hmmm.

Or maybe it is fine. But I'd change the documentation. In this place of API docs noone is interested where this parameter is used. This is kind of integration detail. We can write in the tutorial that options passed in `.get()` are later available in `conversionApi` but what you described is too much.

scofalik:
I left my notes. @Reinmar could you take a look at the API?

scofalik:
After thinking for a while I think that I like the "flat object" solution the most.

scofalik:
I'd document the options object this way:

```
* @param {Object} [options] Additional configuration for the retrived data. `DataController` provides two optional properties: `root` and `trim`. Other properties of this object are specified by various editor features.
```

packages/ckeditor5-engine/src/controller/datacontroller.js
Reinmar: I think that having one object would be best. For example, an option like `options.trim` also make sense to be available during conversion. From the integrator's perspective, one will not be able to understand why `showSuggestions` is a conversion option and `trim` is not. This is an implementation detail that should be abstracted.

That's why one `options` object should be enough.

Reinmar:
Quoting myself:

> I think that having one object would be best. For example, an option like `options.trim` also makes sense to be available during conversion. From the integrator's perspective, one will not be able to understand why `showSuggestions` is a conversion option and `trim` is not. This is an implementation detail that should be abstracted.
> 
> That's why one `options` object should be enough.

So, I think we agree with each other @scofalik.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
Reinmar: Correct me if I'm wrong, but in the current form, we'd need to have an additional param in every event and method.

In that case, conversionApi would be better as it's already widely available.

oskarwrobel: >  we'd need to have an additional param in every event and method
Not in every, but in `convertInsert()` and `convertMarkerAdd()`. To those methods that are used in `DataController#toView()`.

oskarwrobel: For me, it sounds good to set options directly to conversionApi (at the beginning of `DataController#toView()`) and remove when the conversion is done (at the end of `DataController#toView()`)

packages/ckeditor5-core/src/editor/utils/dataapimixin.js
```diff
@@ -76,5 +76,6 @@ export default DataApiMixin;
  * @param {String} [options.trim='empty'] Whether returned data should be trimmed. This option is set to `'empty'` by default,
  * which means that whenever editor content is considered empty, an empty string is returned. To turn off trimming
  * use `'none'`. In such cases exact content will be returned (for example `'<p>&nbsp;</p>'` for an empty editor).
+ * @param {Object} [options.conversionOptions] Additional, custom configuration passed to the conversion process.
```
scofalik, line 79: This is no longer valid.

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -146,15 +146,16 @@ export default class DataController {
 	 * Returns the model's data converted by downcast dispatchers attached to {@link #downcastDispatcher} and
 	 * formatted by the {@link #processor data processor}.
 	 *
-	 * @param {Object} [options]
+	 * @param {Object} [options] Additional configuration for the retrieved data. `DataController` provides two optional
+	 * properties: `root` and `trim`. Other properties of this object are specified by various editor features.
```
scofalik, line 150: rootName

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -220,6 +224,9 @@ export default class DataController {
 
 		this.mapper.bindElements( modelElementOrFragment, viewDocumentFragment );
 
+		// Make additional options available during conversion process through conversionSpi.
```
scofalik, line 227:
```suggestion
		// Make additional options available during conversion process through `conversionApi`.
```

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -233,6 +240,9 @@ export default class DataController {
 			}
 		}
 
+		// Remove options from conversionApi.
```
scofalik, line 243:
```suggestion
		// Clean `conversionApi`.
```

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -166,6 +165,7 @@ export default class DowncastDispatcher {
 	 * @fires attribute
 	 * @param {module:engine/model/range~Range} range The inserted range.
 	 * @param {module:engine/view/downcastwriter~DowncastWriter} writer The view writer that should be used to modify the view document.
+	 * {module:engine/conversion/downcastdispatcher~DowncastConversionApi conversion api}. Used only with `dataDowncast` group.
```
scofalik, line 168: ?

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -319,6 +319,7 @@ export default class DowncastDispatcher {
 	 * @param {String} markerName Marker name.
 	 * @param {module:engine/model/range~Range} markerRange Marker range.
 	 * @param {module:engine/view/downcastwriter~DowncastWriter} writer View writer that should be used to modify view document.
+	 * {module:engine/conversion/downcastdispatcher~DowncastConversionApi conversion api}. Used only with `dataDowncast` group.
```
scofalik, line 322: ?

packages/ckeditor5-core/src/editor/utils/dataapimixin.js
```diff
@@ -71,7 +71,8 @@ export default DataApiMixin;
  * the right format for you.
  *
  * @method #getData
- * @param {Object} [options]
+ * @param {Object} [options] Additional configuration for the retrieved data. By default two optional
```
scofalik, line 74: How about ...

```
Additional configuration for the retrieved data. Editor features may introduce more configuration options that can be set through this parameter.
```

scofalik: I think I'd leave `DataController` description as it is now. `editor.getData()` is a very top-level method, its description needs to be as simple as it can.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -669,3 +668,9 @@ function shouldMarkerChangeBeConverted( modelPosition, marker, mapper ) {
  *
  * @member {module:engine/view/downcastwriter~DowncastWriter} #writer
  */
+
+/**
+ * An object with an additional, custom configuration used during conversion process. Available only for `dataDowncast` group.
```
scofalik, line 673:
```suggestion
 * An object with an additional configuration used during conversion process. Available only for data downcast conversion.
```

PR #7778, opened 2020-08-05
Pasting a plain text will inherit selection attributes

### LINKED ISSUE #1006: Pasted plain text should inherit formatting of the surrounding text

🐞 Bug report  

## 💻 Version of CKEditor

1.0.0.beta.2

## 📋 Steps to reproduce

1. Add some content in any H tag and make it bold.
2. Copy some plain text from somewhere else like from address bar of Browser.
3. Paste the content in between the H tag added in step 1

## ✅ Expected result
The new text should be Bold as well.

## ❎ Actual result
The new text is not Bold.

copy-paste-issue

pomek:

Feature (clipboard): Pasting a plain text will inherit selection attributes. Closes #1006.


Reinmar:
I wonder if that's a correct heuristic for detecting plain text paste, as it actually does not check what was in the clipboard but instead what is it in the model. It's not something that was discussed, but it has its merits.

@Mgsy, you know most about the issues that our community and customers are reporting. Could you test it from the user perspective and tell us how do you feel about this approach?

jodator:
> t actually does not check what was in the clipboard but instead what is it in the model.

Well we can't check view data for that also. We _could_ however introduce some additional data for the listeners.

The <kbd>ctrl+shift+v</kbd> pastes `text/plain` only data - so we might think about extending `inputTransformation` data with `isPlainText` attribute.

The other thing is how browser behaves when pasting plain text. AFAICS there are two cases:

1.  text from plain text editor (ie from address bar) - it is the same as <kbd>ctrl+shift+v</kbd> case so the flag would cover that.
2.  text from editor/editable (ie have `'foo'` copied)
    1.  Firefox pastes `foo` as both `text/html` and `text/plain` - a simple string comparison would be enough to set `isPlainText`
    2.  Chrome however, would paste `<meta http-equiv="content-type" content="text/html; charset=utf-8">foo` as `text/html` - here I see a potential fix for this.
    3.  I didn't check Safari nor Edge

But OTOH the @pomek's check operates on model level, after conversion so it would probably catch the situation of pasting not-enabled formatting (let's say strikethrouhg) in editor. In the end we will insert single text node (without attributes) so considering pasting "~foo~" into "**bar | bar**" - user could see either (none of which have unsupported formatting)

1.  "**bar foo| bar**" (formatting copied from selection)
2.  "**bar** foo**| bar**" (pasted as plain-text)

If, we'd like option 1 in this case I think that this check will actually be valid.

If we however would like option 2 better - then we should check for a true plain text in the clipboard.

@pomek - before we get agreement on solution I see that we have too little tests:

1.  Plain text in the clipboard when both `text/html` and `text/plain` are available - let's test both FF and Chrome cases.
2.  Unsupported formatting in the clipboard pasted.

pomek:
@jodator, I added mentioned tests.

Reinmar:
> 1.  Firefox pastes `foo` as both `text/html` and `text/plain` - a simple string comparison would be enough to set `isPlainText`

Good idea, we could consider that. That'd be a bit more complex as the HTML could look like this: `'foo &nbsp;bar&gt;bom'` (`foo  bar<bom`) but with some simple regexp replace it's doable.

As for the other things that @jodator wrote, I think that we need to be predictable:

*   If you pressed ctrl+shift+v it's very likely that you want to inherit selection attributes. But then @pomek's solution won't handle multi-line text as in the model it'd contain soft breaks or paragraphs. Do we want to handle multi-line text the same way as a single line? If yes, we need an additional check for that.
*   If you pressed ctrl+v but you had very simple content in the clipboard, how likely is it that you want it to inherit selection styles? I'd say, quite likely, so I found @pomek's idea really interesting. However, I'm slightly worried about the randomness of this solution. I copied some formatted text from paste from office or other application and that formatting is not supported by the editor – that will make this text to inherit the selection styles. But if a supported style was included in that text, the style will not be inherited. It may feel "random". But that's where I count on your intuition and experience, @Mgsy. Can we live with that?

PR #7779, opened 2020-08-05
Detect target inside a shadow root for clickOutsideHandler (copy of #7744)

### LINKED ISSUE #7743: Detect target inside a shadow root for clickOutsideHandler

pomek:

Other (ui): The `clickOutsideHandler()` function will take into consideration that the editor can be placed in a shadow root while detecting a click. Closes #7743. 

---

### Additional information

Clone of https://github.com/ckeditor/ckeditor5/pull/7744 since we're not able to apply any change in the original PR.

cc: @ywsang


pomek:
Thanks @ywsang for the PR.

PR #7787, opened 2020-08-05
#7753: Upcast conversion should handle wrapping in paragraphs if it would allow conversion

### LINKED ISSUE #6698: Custom inline element disappear after setData() if it is the only or first element in tableCell
I target to create a plugin to insert FontAwesome icon to the editor.

_(I understand this is not a bug, but my lack of knowledge on the editor, anyhow when creating the ticket "bug" was the closest option from the list.)_

## Code

```js
// fontsymbolediting.js
// init()
    //.....

    schema.register('fontSymbol', {
      allowWhere: '$text',
      isInline: true,
      isBlock: true,
      isObject: true,
      allowAttributes: ['name'],
    });

    conversion.for('dataDowncast').elementToElement({
      model: 'fontSymbol',
      view: (modelItem, viewWriter) => {
        console.log('dataDowncast fontSymbol')
        const symbolName = modelItem.getAttribute('name');
        const symbolColor = modelItem.getAttribute(FONT_COLOR);

        const symbolElement = viewWriter.createUIElement(
          'i',
          {
            class: `m-font-symbol fas fa-${symbolName}`,
            style: `color: ${symbolColor}`,
          },
        );

        return symbolElement;
      },
    });

    conversion.for('downcast').elementToElement({
      model: 'fontSymbol',
      view: (modelElement, viewWriter) => {
        console.log('editingDowncast fontSymbol')
        const symbolName = modelElement.getAttribute('name');
        const symbolColor = modelElement.getAttribute(FONT_COLOR);

        const symbolElement = viewWriter.createUIElement(
          'i',
          {
            class: `m-font-symbol fas fa-${symbolName}`,
            style: `color: ${symbolColor}`,
          },
        );

        const viewWrapper = viewWriter.createContainerElement('span', {
          class: 'm-font-symbol-container',
          name: symbolName,
        });
        viewWriter.insert(viewWriter.createPositionAt(viewWrapper, 0), symbolElement);

        return toWidget(viewWrapper, viewWriter);
      },
    });

    conversion.for('upcast').elementToElement({
      view: {
        name: 'i',
        classes: ['m-font-symbol'],
      },
      model: (viewElement, modelWriter) => {
        const cssClasses = viewElement.getAttribute('class') || '';
        const name = (cssClasses.match(/fa-([\w-]+)/) || [])[1];

        if (!name) {
          return null;
        }

        const fontSymbolModel = modelWriter.createElement('fontSymbol', { name });

        return fontSymbolModel;
      },
    });
```

I took reference from the horizontal rule plugin, everything worked fine until I saved the data and reloaded it into the editor.

The icon was not showing up.

upcast() was called, model element returned, but the element did not appear.

I looked into table cell, and realized that table cell accept only block element.

I cannot find any example of putting inline element inside table cell.

However, if I pad an empty space before the icon, it can display properly, with the padded space of course.

### ✔️ Expected result

The inline icon appears in the table cell after setData()

### ❌ Actual result

Nothing shows up in the table cell

## 📃 Other details

* Browser: Chrome
* CKEditor version: 18
* Installed CKEditor plugins: font, table

### LINKED ISSUE #7753: The upcast conversion can't properly handle inline objects without $text accepting context
# The upcast conversion and auto-paragraphing

*   Status: RFC
*   Date: 2020-07-30

Technical Story:

### Context

Inline objects that are allowed only where `$text`, can't be properly auto-paragraphed and it results as broken upcast.

This happens because of a mistake in the auto-paragraphing code (view element is used to check `schema.checkChild()`) https://github.com/ckeditor/ckeditor5/blob/50e3509beec62add39a95163142fa18b4f21b7aa/packages/ckeditor5-paragraph/src/paragraph.js#L196

This error can be easily reproduced using the [placeholder widget](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/tutorials/implementing-an-inline-widget.html) like this:

```javascript
editor.setData('<span class="placeholder">{nested}</span>')
```

1.  The `span` element can't be converted because `splitToAllowedParent` can't find a proper insertion position.
2.  It falls back to the auto-paragraphing converter that is using this `span` view element to check if it's allowed by the schema.
3.  This fails (every time there is no coincidence that view element's name matches the model element's name) and falls back to the default converter that consumes an element without creating a model element and triggers children's conversion.

### Problem

Auto-paragraphing doesn't know what would be the model element after converting view element so can't properly check if it would convert properly if it would be auto-paragraphed.

## Considered Options

1.  Auto-paragraphing should try to convert problematic element in the context of `paragraph` element (without checking if the resulting model element is allowed in the context by the schema) and afterward check if it is compliant with schema.
2.  Auto-paragraphing should be able to decorate `conversionApi.splitToAllowedParent()` to verify if the converted model element would be allowed if auto-paragraphed (it could insert paragraph there and return position inside that paragraph).
3.  Auto-paragraphing for elements should be integrated directly into `conversionApi.splitToAllowedParent()`.
4.  Conversion process could be split into individual events matching steps of conversion (with default handlers taking care of it if not overridden), an example sequence of events:
    *   `'element:span:match'` - testing if view item should be handled
    *   `'element:span:create'` - creating a model element from the view element (`getModelElement()`)
    *   `'element:span:insert'` - finding a proper position and inserting a model element (`splitToAllowedParent()` and `writer.insert()`)
    *   `'element:span:children'` - `convertChildren()`
    *   `'element:span:updatecursors'` - `updateConversionResult()`

### Option 1: Auto-paragraphing and checking afterward

*   Good, because auto-paragraphing is handled in one place (this is not entirely true - eg `model.deleteContent()`).
*   Bad, because there could be elements that should just be ignored and there is no easy way to verify if an element was properly converted after wrapping with a paragraph or it was just handled by default handler (that ignores element and continues conversion of its children)

### Option 2: "Decorating" `splitToAllowedParent()` to extend it with auto-paragraphing

Probably to maintain consistency with conversion API it should be a dedicated event that would also provide `conversionApi` object.

*   Good, because allows extending `splitToAllowedParent()` from external plugins (i.e., paragraph is a separate plugin) so auto-paragraphing would stay in one place.
*   Bad, because there could be other cases that might make use of extending part of built-in conversion steps and it would require breaking change if it would be changed in the future.

### Option 3: Auto-paragraphing elements directly inside `splitToAllowedParent()`

Currently, parts of the paragraph functionality are highly incorporated into the engine (example: `model.deleteContent()` is auto-paragraphing if needed.

*   Good, because probably could simplify auto-paragraphing code.
*   Bad, because if backward compatibility would be maintained then `splitToAllowedParent` is not the best name for it.
*   Bad, because it would tighten paragraph with the engine even more than it is currently.

### Option 4: Splitting the conversion process to a series of events

*   Good, because converters could override only selected parts of the conversion process.
*   Good, because the whole conversion process would expose selective changes.
*   Bad, because this would generate a ton more events during conversion.
*   Bad, because it could be difficult to maintain backward compatibility (but this might be possible).

niegowski:

Fix (engine): Upcast conversion should handle wrapping in paragraphs if it would allow conversion. Closes #7753. Closes #6698.

Internal (paragraph): Auto-paragraphing content conversion moved to the engine package.

Tests (table): Added tests for upcasting inline elements inside a table cell.

---

### Additional information


packages/ckeditor5-engine/tests/conversion/upcastdispatcher.js
```diff
@@ -605,6 +605,39 @@ describe( 'UpcastDispatcher', () => {
...
+				model.schema.register( 'span', {
+					allowIn: 'paragraph'
+				} );
+				model.schema.extend( 'paragraph', {
+					allowIn: 'section'
+				} );
+
+				dispatcher.on( 'documentFragment', ( evt, data, conversionApi ) => {
+					const section = conversionApi.writer.createElement( 'section' );
+					const position = ModelPosition._createAt( section, 0 );
```
jodator, line 623: could you use `conversionApi.writer` to create postions? Using `writer` or `model` is preferable then importing `ModelPosition`.
We left that in tests and in engine source code but we might use proper API in the test.

packages/ckeditor5-engine/tests/conversion/upcastdispatcher.js
```diff
@@ -605,6 +605,39 @@ describe( 'UpcastDispatcher', () => {
 				sinon.assert.calledOnce( spy );
 			} );
 
+			it( 'should insert paragraph if it\'s allowed and if it would allow conversion', () => {
```
jodator, line 608: Maybe, "should auto-paragraph element if it is not allowed at insert postion but is inserted where paragraph is allowed"?

packages/ckeditor5-engine/tests/conversion/upcasthelpers.js
```diff
@@ -969,6 +969,31 @@ describe( 'upcast-converters', () => {
 			expect( conversionResult.getChild( 0 ).data ).to.equal( 'foobar' );
 		} );
 
+		it( 'should wrap text with paragraph if it would allow conversion', () => {
```
jodator, line 972: Maybe 'should auto-paragraph text if it is not allowed at insert postion but is inserted where paragraph is allowed'?

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -459,20 +459,37 @@ export function convertToModelFragment() {
...
-	return ( evt, data, conversionApi ) => {
-		if ( conversionApi.schema.checkChild( data.modelCursor, '$text' ) ) {
-			if ( conversionApi.consumable.consume( data.viewItem ) ) {
-				const text = conversionApi.writer.createText( data.viewItem.data );
+	return ( evt, data, { schema, consumable, writer } ) => {
+		let position = data.modelCursor;
 
-				conversionApi.writer.insert( text, data.modelCursor );
+		// When node is already converted then do nothing.
+		if ( data.modelRange || !consumable.test( data.viewItem ) ) {
```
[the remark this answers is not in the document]
niegowski: This was handled this way (only `data.modelRange`) in the original code for this. I was also surprised so I added `consumable.test` there. It's also strange that CC is 100%

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -459,20 +459,37 @@ export function convertToModelFragment() {
...
 
-				data.modelRange = conversionApi.writer.createRange(
-					data.modelCursor,
-					data.modelCursor.getShiftedBy( text.offsetSize )
-				);
-				data.modelCursor = data.modelRange.end;
+		if ( !schema.checkChild( position, '$text' ) ) {
+			// Check if this text node would be allowed if wrapped with a paragraph.
+			const context = schema.createContext( position );
+
```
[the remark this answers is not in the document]
jodator: Nope :D

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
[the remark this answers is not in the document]
niegowski: Seems that CC can't notice that it's redundant. Updated it.

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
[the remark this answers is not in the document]
niegowski: I think I managed to make it reusable :smile:

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -140,6 +141,12 @@ export default class DataController {
 		this.on( 'init', () => {
 			this.fire( 'ready' );
 		}, { priority: 'lowest' } );
+
+		// Fix empty roots after DataController is 'ready' (note that init method could be decorated and stopped).
+		// We need to handle this event because initial data could be empty and post-fixer would not get triggered.
+		this.on( 'ready', () => {
+			this.model.enqueueChange( 'transparent', writer => autoParagraphEmptyRoots( writer ) );
```
jodator, line 148:
```suggestion
			this.model.enqueueChange( 'transparent', autoParagraphEmptyRoots );
```

packages/ckeditor5-engine/src/model/model.js
```diff
@@ -122,6 +123,9 @@ export default class Model {
 
 		injectSelectionPostFixer( this );
 
+		// Post-fixer which takes care of adding empty paragraph elements to the empty roots.
+		this.document.registerPostFixer( writer => autoParagraphEmptyRoots( writer ) );
```
jodator, line 127:
```suggestion
		this.document.registerPostFixer( autoParagraphEmptyRoots );
```

MrMitch:
@niegowski and @jodator thank you for taking the time to do this and fix #6698, greatly appreciated !

Issue #6575, opened 2020-04-08
ToolbarView should handle items grouping correctly when invisible


oleq:
## 📝 Provide detailed reproduction steps (if any)

1. The grouping should never happen if the toolbar is invisible. Because it always leads to all items groped and weird things when other features (like positioning algorithm) when they see the toolbar in such state

![image](https://user-images.githubusercontent.com/1099479/78762076-b4d8b880-7983-11ea-9b11-fe4fa5ad0451.png)

2. The grouping that was triggered by changing `ToolbarView#maxWidth` when toolbar is invisible should be queued so the next time the toolbar shows up (`ResizeObserver` kicks in automatically), the items are re-grouped. This is a problem when `maxWidth` increased because there's no pressure to ungroup the items when the toolbar showed up.

Related to https://github.com/ckeditor/ckeditor5/issues/6449.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

PR #7795, opened 2020-08-06
#6444: Balloon toolbar should reposition and ungroup items correctly when the window resizes

### LINKED ISSUE #6444: [Safari] Balloon editor toolbar should always expand when there's enough space

0. Open the balloon editor and make a selection.
1. Scale down the browser.
2. Hide the toolbar.
3. Expand the window.
4. Make a selection again.

### ✔️ Expected result

The toolbar should look like before 1.

### ❌ Actual result

The toolbar looks like after 1.

## 📃 Other details

It works correctly in other browsers because unlike Safari, they don't use polyfilled `ResizeObserver`. I think something's wrong with the polyfill, maybe it should be improved.

cc @panr 

---

oleq:

Fix: Balloon toolbar should reposition and ungroup items correctly when the window resizes. Closes #6444.

---

### Additional information

This PR is another attempt at https://github.com/ckeditor/ckeditor5/pull/7721/files.  

There are 2 things in this PR (this is hard but here we go):

1.  A reverted `ResizeObserver` optimization (mentioned here https://github.com/ckeditor/ckeditor5/pull/7721#discussion_r461502549).   
      
    Without it (or... actually with it), the `previousWidth !== entry.contentRect.width` in this `ToolbarView` line https://github.com/ckeditor/ckeditor5/blob/i/6444-toolbar-reposition-on-regroup/packages/ckeditor5-ui/src/toolbar/toolbarview.js#L757 does not work.  
      
    The `ToolbarView` optimizes the number of `_updateGrouping()` calls so they only happen when:  
      
    \- it's the first resize in its lifetime (resize observer always hits when the toolbar first appears),  
    \- the element width changes (but not, for instance, its height) (:point\_left::point\_left::point\_left:  this one matters),  
    \- the update has been queued due to toolbar `maxWidth` change when the toolbar was invisible (editable resized when the toolbar was invisible)  
      
    When the mentioned `ResizeObserver` optimization is in place, the "toolbar visible"->"toolbar invisible" state change is lost (the observer callback is not called then). So `previousWidth` is actually a width **before** the toolbar disappeared. And when it re-appears, well... it's still the same, so the grouping is not updated.  
      
    So this one actually fixes #6444.  
     
2.  A new event fired by `ToolbarView` when it groups or ungroups its items (and `BalloonToolbar` making use of it).  
      
    When 1. is applied, the balloon toolbar does group/ungroup when it re-appears (GIF from #6444). That's fine.  
      
    But a side-effect of this grouping/ungrouping is that it changes its width and being positioned absolutely (to its upper left corner) the balloon triangle starts pointing to nothing (or nonsense) 😕 (see the GIF below).

    In other words: fixing 1. revealed 2.
      
    So the `BalloonToolbar` plugin must listen to the `ToolbarView` event and re-position itself once the geometry of the toolbar has changed (due to grouping/ungrouping).  
      
    **Side note:** At first I wanted to avoid the new event and make the `BalloonToolbar` use another resize observer (observing the toolbar element) and re-position based on this. But then I figured this may end in a race condition since   
      
    \- a toolbar internally uses a resize observer on itself to group/ungroup items,  
    \- there's this resize observer observing the editable that sets toolbar's `maxWidth`  
      
    If I added another resize observer, understanding which executed first and which last and whether the balloon repositioning happens before toolbar items grouping or after it would be a total mess. So an event is safer because it always happens after the internal toolbar's resize observer.

### DIFF
--- a/packages/ckeditor5-ui/src/toolbar/balloon/balloontoolbar.js
+++ b/packages/ckeditor5-ui/src/toolbar/balloon/balloontoolbar.js
@@ -165,6 +165,15 @@ export default class BalloonToolbar extends Plugin {
 				} );
 			} );
 		}
+
+		// Listen to the toolbar view and whenever it changes its geometry due to some items being
+		// grouped or ungrouped, update the position of the balloon because a shorter/longer toolbar
+		// means the balloon could be pointing at the wrong place. Once updated, the balloon will point
+		// at the right selection in the content again.
+		// https://github.com/ckeditor/ckeditor5/issues/6444
+		this.listenTo( this.toolbarView, 'groupedItemsUpdate', () => {
+			this._updatePosition();
+		} );
 	}
 
 	/**
@@ -236,7 +245,7 @@ export default class BalloonToolbar extends Plugin {
 
 		// Update the toolbar position when the editor ui should be refreshed.
 		this.listenTo( this.editor.ui, 'update', () => {
-			this._balloon.updatePosition( this._getBalloonPositionData() );
+			this._updatePosition();
 		} );
 
 		// Add the toolbar to the common editor contextual balloon.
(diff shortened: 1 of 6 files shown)


oleq:
P.S. when reviewing please make sure #6575 didn't re-appear.

panr:
> P.S. when reviewing please make sure #6575 didn't re-appear.
Couldn't reproduce it (even with `BlockToolbar` and `BalloonToolbar` mixed together) 👍

panr:
I closed https://github.com/ckeditor/ckeditor5/pull/7721, since it's no longer needed.

PR #7829, opened 2020-08-12
#7828: DomEmitterMixin should pass "passive" option for listener

### LINKED ISSUE #7828: DomEmitterMixin should pass "passive" option for listener

The current implementation of `DomEmitterMixin` only supports `useCapture` option but no `usePassive` option.

This is needed for non-blocking `scroll` listeners - [https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners)

---

niegowski:

Feature (utils): DomEmitterMixin should support the `passive` option to `addEventListener`. Closes #7828.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-utils/src/dom/emittermixin.js
+++ b/packages/ckeditor5-utils/src/dom/emittermixin.js
@@ -187,10 +187,15 @@ extend( ProxyEmitter.prototype, EmitterMixin, {
 			return;
 		}
 
-		const domListener = this._createDomListener( event, !!options.useCapture );
+		const listenerOptions = options.useCapture || options.usePassive ? {
+			capture: !!options.useCapture,
+			passive: !!options.usePassive
+		} : false;
+
+		const domListener = this._createDomListener( event, listenerOptions );
 
 		// Attach the native DOM listener to DOM Node.
-		this._domNode.addEventListener( event, domListener, !!options.useCapture );
+		this._domNode.addEventListener( event, domListener, listenerOptions );
 
 		if ( !this._domListeners ) {
 			this._domListeners = {};
--- a/packages/ckeditor5-utils/src/dom/emittermixin.js
+++ b/packages/ckeditor5-utils/src/dom/emittermixin.js
@@ -179,6 +179,7 @@ extend( ProxyEmitter.prototype, EmitterMixin, {
 	 * @param {Object} [options={}] Additional options.
 	 * @param {Boolean} [options.useCapture=false] Indicates that events of this type will be dispatched to the registered
 	 * listener before being dispatched to any EventTarget beneath it in the DOM tree.
+	 * @param {Boolean} [options.usePassive=false] Indicates that the function specified by listener will never call preventDefault().
 	 */
 	attach( event, callback, options = {} ) {
 		// If the DOM Listener for given event already exist it is pointless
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-utils/src/dom/emittermixin.js
```diff
@@ -179,6 +179,7 @@ extend( ProxyEmitter.prototype, EmitterMixin, {
 	 * @param {Object} [options={}] Additional options.
 	 * @param {Boolean} [options.useCapture=false] Indicates that events of this type will be dispatched to the registered
 	 * listener before being dispatched to any EventTarget beneath it in the DOM tree.
+	 * @param {Boolean} [options.usePassive=false] Indicates that the function specified by listener will never call preventDefault().
```
oleq, line 182: Well, this is only half of the story. The real use-case for this option is unblocking the browser's main thread and this should be mentioned.

packages/ckeditor5-utils/src/dom/emittermixin.js
```diff
@@ -187,10 +188,15 @@ extend( ProxyEmitter.prototype, EmitterMixin, {
 			return;
 		}
 
-		const domListener = this._createDomListener( event, !!options.useCapture );
+		const listenerOptions = options.useCapture || options.usePassive ? {
```
oleq, line 191: Is there any reason why we cannot pass `listenerOptions = { capture: false, passive: false }` to `addEventListener()`? That would certainly simplify the code and I highly doubt it would have an impact on performance etc..

packages/ckeditor5-utils/src/dom/emittermixin.js
```diff
@@ -227,10 +233,12 @@ extend( ProxyEmitter.prototype, EmitterMixin, {
 	 * @private
 	 * @method module:utils/dom/emittermixin~ProxyEmitter#_createDomListener
 	 * @param {String} event The name of the event.
-	 * @param {Boolean} useCapture Indicates whether the listener was created for capturing event.
+	 * @param {Object} [options] Additional options.
+	 * @param {Boolean} [options.capture=false] Indicates whether the listener was created for capturing event.
+	 * @param {Boolean} [options.passive=false] Indicates that the function specified by listener will never call preventDefault().
```
oleq, line 238: Same comment as before.

packages/ckeditor5-utils/src/dom/emittermixin.js
```diff
@@ -49,6 +49,8 @@ const DomEmitterMixin = extend( {}, EmitterMixin, {
 	 * order they were added.
 	 * @param {Boolean} [options.useCapture=false] Indicates that events of this type will be dispatched to the registered
 	 * listener before being dispatched to any EventTarget beneath it in the DOM tree.
+	 * @param {Boolean} [options.usePassive=false] Indicates that the function specified by listener will never call preventDefault()
+	 * and prevents blocking browser's main thread by this event handler.
```
niegowski, line 53: Added missing option docs.

packages/ckeditor5-utils/src/dom/emittermixin.js
niegowski: Updated.

packages/ckeditor5-utils/src/dom/emittermixin.js
niegowski: Changed.

packages/ckeditor5-utils/src/dom/emittermixin.js
niegowski: Done.

Issue #4969, opened 2019-10-09
Rect.getDomRangeRects doesn't support ranges inside the text nodes


ma2ciek:
Running `documentView.scrollToTheSelection` when selection is inside some text nodes throws the following Error:

`Error: Uncaught TypeError: range.startContainer.getBoundingClientRect is not a function.`

Issue #4952, opened 2019-10-09
Rect utility should work for collapsed DOM Ranges


oleq:
ATM, collapsed Range at the beginning of the block has bounding client rect full of zeros. `getClientRects()[ 0 ]` solves the issue but [when the block is empty, even this method returns an empty array](https://github.com/ckeditor/ckeditor5-ui-default/pull/91#issuecomment-251344158). A smarter approach is needed, probably `startContainer.getBoundingClientRect` with a small correction of width.

A follow-up of https://github.com/ckeditor/ckeditor5-link/issues/46#issuecomment-298966477.

PR #7840, opened 2020-08-13
Fixed Rect results for a sequenced range

### LINKED ISSUE #7838: Rect utility returns wrong sizes in case of sequenced range

While working on #7705 I found out that there are cases when balloon panel being mispositioned, due to the fact that the [`Rect`](https://ckeditor.com/docs/ckeditor5/20.0.0/api/module_utils_dom_rect-Rect.html) utility [cares only for the first range's client rectangle](https://github.com/ckeditor/ckeditor5/blob/f31092d65d97e91b21cd77bd12099f0b80d2b09e/packages/ckeditor5-utils/src/dom/rect.js#L81) (while it can contain multiple).

Example repro case:

1.  Open https://ckeditor.com/docs/ckeditor5/19.0.0/features/link.html
2.  Focus the first editor.
3.  Make the following selection `The [capital] city`
4.  Bold it.
5.  Make a wider selection, e.g. `The [capital city of` [`Malta`](https://en.wikipedia.org/wiki/Malta) `is the top destination this summer.]`
6.  Use the link button to open link balloon.

### ✔️ Expected result

The balloon points to the middle of selection.

### ❌ Actual result

The balloon point to the beginning of the selection.

Reason is that it contains multiple boxes, the first of which (on chrome from my testing is often) 0px width, thus it points to the very beginning.

## 📃 Other details

*   Browser: Any
*   OS: Any

---

mlewand:

Fix (utils): `Rect` utility returns wrong sizes in case of a sequenced range. Closes #7838. 

Feature (utils): Introduced the `Rect#getBoundingRect()` method that returns a `Rect` instance containing all other rectangles. Closes #7858.

---

### Additional information

Initially I was even tempted to simplify it and use `Range.getBoundingClientRect()` to do the computation for us, but then I saw a plethora of issues that this method not always give reliable results:

*   #4969
*   #4952

### DIFF
--- a/packages/ckeditor5-utils/src/dom/rect.js
+++ b/packages/ckeditor5-utils/src/dom/rect.js
@@ -414,7 +414,7 @@ export default class Rect {
 		boundingRectData.width = boundingRectData.right - boundingRectData.left;
 		boundingRectData.height = boundingRectData.bottom - boundingRectData.top;
 
-		return boundingRectData;
+		return new Rect( boundingRectData );
 	}
 }
 
--- a/packages/ckeditor5-utils/src/dom/rect.js
+++ b/packages/ckeditor5-utils/src/dom/rect.js
@@ -78,7 +78,8 @@ export default class Rect {
 			// @if CK_DEBUG // }
 
 			if ( isSourceRange ) {
-				copyRangeRectProperties( this, source );
+				const rangeRects = Rect.getDomRangeRects( source );
+				copyRectProperties( this, Rect.getBoundingRect( rangeRects ) );
 			} else {
 				copyRectProperties( this, source.getBoundingClientRect() );
 			}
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-utils/src/dom/rect.js
```diff
@@ -406,3 +406,25 @@ function isBody( elementOrRange ) {
...
+		combinedRect.left = Math.min( combinedRect.left, rangeRects[ i ].left );
+		combinedRect.bottom = Math.max( combinedRect.bottom, rangeRects[ i ].bottom );
+		combinedRect.top = Math.min( combinedRect.top, rangeRects[ i ].top );
+	}
+
+	combinedRect.width = combinedRect.right - combinedRect.left;
+	combinedRect.height = combinedRect.bottom - combinedRect.top;
+
+	copyRectProperties( rect, combinedRect );
+}
```
niegowski, line 430: In the table package we have a helper function that is very similar (the same purpose but different approach). Maybe we could unify those to one helper function? 

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-table/src/utils/ui/contextualballoon.js#L123-L144

mlewand: Good idea, added #7858 that will also be closed by this PR.

I had a small problem with the naming, first wanted to make it a `Range#expand()` member, but the name didn't feel right. Ended up with static `Range.getBoundingRect()` which feels just right :ok\_hand:

Also I simplified mentioned contextualballoon logic a little bit as the way how `createBoundingRect` [internal logic was implemented in the `getBalloonCellPositionData` function](https://github.com/ckeditor/ckeditor5/blob/1738d0f2296cfc9699cb1d78d7c459f60f1465e0/packages/ckeditor5-table/src/utils/ui/contextualballoon.js#L91-L93) was a little confusing.

niegowski: I guess you mean `Rect` not `Range` 😉

packages/ckeditor5-utils/src/dom/rect.js
```diff
@@ -381,6 +382,40 @@ export default class Rect {
...
+		}
+
+		if ( rectangleCount == 0 ) {
+			return null;
+		}
+
+		boundingRectData.width = boundingRectData.right - boundingRectData.left;
+		boundingRectData.height = boundingRectData.bottom - boundingRectData.top;
+
+		return boundingRectData;
```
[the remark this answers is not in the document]
mlewand: Fixed in f863b19

Issue #3891, opened 2019-10-09
Support for running in shadow DOM


Reinmar:
Apparently, shadow DOM has its own selection (like an iframe), so when running in a shadow root, the engine must use that selection.

Welcome to hell...

ywsang:
@Reinmar Hi there! I’ve been working on creating a web component version of CKEditor5, which involves putting the editor inside of shadow DOM. I’ve made a couple changes to the source code that help increase the amount of support for running the editor in the shadow DOM. So far, these changes are related to the editor’s styles (so that the editor looks visually correct when placed inside shadow DOM) and the engine selection issue. I’d love to contribute these changes back to the repo, though they’re not complete fixes for supporting running the editor in shadow DOM (this issue). Would you be open to accepting incremental changes towards this goal that tackle more fine-grained issues? 

cc: @justinfagnani

yoyo837:
Can we consider this in 2022?

ralfpatric:
Any idea if this will come in 2023? 🤔

Witoso:
@ralfpatric most likely not, we have not currently set it as a priority.

mancristiana:
I reproduced the issue we are facing when adding CKEditor in ShadowDOM
https://stackblitz.com/edit/lit-element-ckeditor5-xd7j6g?file=index.html,index.js,package.json

Reinmar:
I don't want to be the party killer here but the problem with shadow DOM isn't (to our knowledge) really fixable at the moment. It's been a while since we investigated this and AFAICS we failed to share the details here, but all issues except one were rather straightforward to fix.

The one issue being: selection handling 😳🤬

There's currently no standardized API to access selection in a shadow root. If there were even different APIs, but somehow compatible, we would painfully resolve this ticket. But nope – it's 2023 and the state is as follows:

*   Chrome: non-standard API (`shadowRoot#getSelection()`)
*   Webkit: [very fresh implementation](https://bugs.webkit.org/show_bug.cgi?id=163921) based on an incomplete/unreviewed specification
*   Firefox: [https://bugzilla.mozilla.org/show_bug.cgi?id=1430308](https://bugzilla.mozilla.org/show_bug.cgi?id=1430308) "The specification isn't quite ready yet, so I don't know what they have shipped in their beta"

[...]

Issue #1483, opened 2019-01-24
WebComponent for CKEditor 5


Reinmar:
Let's ship every build as a web component 💛 

* [ ] Shadow DOM support
 
    Ability to close the editor in a shadow DOM would greatly improve its integrity by scoping its styles. However, this requires some changes:

    * In the engine: https://github.com/ckeditor/ckeditor5-engine/issues/692
    * In all other places where we create element/nodes?
* [ ] Body collection
   
    The so-called "body collection" is where we keep floating panels (so they are not trimmed by overflowed parents and are easier to position). It's automatically created in the `<body>` element. If we'd enclose CKEditor in a `<ck-editor>`, we should not leak from this element, hence we need to give the developer some control over where this collection is rendered.
* [ ] Naming (`<ck-editor>` or `<ckeditor-classic>` or ?)
   
    The name must contain the `-` character 😿 Also, there may be multiple editor types on one page, so `<ckeditor-*>` seems natural. Or not?
* [ ] Should every editor build expose a web component by default?

    For once, it would be cool that you can use `ClassicEditor.create()` or `<ckeditor-classic>` after importing `@ckeditor/ckeditor5-build-classic` into your page.

    However, I'm not sure how much additional code will the componentization of CKEditor require. I hope it won't increase the final size that much because we'd have to consider having two files in each build (`ckeditor.js` and e.g. `ckeditor.webcomponent.js`).
* [ ] How to componentize specific types of editors. It's fairly easy with the classic editor, but gets more tricky with the inline/balloon editors and super tricky with the decoupled editor. Most likely, we should keep it as consistent with the React/Angular/Vue components as possible.
* [ ] How to make the editor configurable (pass the config object).

Reinmar:
cc @Comandeer @oleq @pjasiun

pjasiun:
> * Should every editor build expose a web component by default?

I think yes. Each editor may have different options as you mentioned already. On the other hand, I agree it would be nice to make it similar to the react integration.

Reinmar:
I added one point to the list:

> * How to build CSS for web components. This may be a blocker for having one build for normal and component use.

danielamorse:
Would this feature include support for Web Components *in* the editor, too?

[...]

PR #7846, opened 2020-08-14
Support text selection when editor is in a shadow root

ywsang:

Other (engine): Use the source element root (document vs. shadow root) to determine which "root" to perform `getSelection` on.
Other (core): Pass the sourceElementRoot to the Editing Controller.

---

### Additional information

Related to #3891 (and more generally, #1483). 

In order to know whether the editor is in shadow DOM or not, and maintain a reference to the shadow root in the former case, this PR uses a `sourceElementRoot` option set in editor's config upon creation. If the editor is created in a shadow root, the shadow root should be passed in this option. If not provided, this option defaults to the document.

### DIFF
--- a/packages/ckeditor5-core/src/editor/editor.js
+++ b/packages/ckeditor5-core/src/editor/editor.js
@@ -7,6 +7,8 @@
  * @module core/editor/editor
  */
 
+/* global document */
+
 import Context from '../context';
 import Config from '@ckeditor/ckeditor5-utils/src/config';
 import EditingController from '@ckeditor/ckeditor5-engine/src/controller/editingcontroller';
@@ -175,6 +177,8 @@ export default class Editor {
 		 */
 		this.data = new DataController( this.model, stylesProcessor );
 
+		const sourceElementRoot = config.sourceElementRoot ? config.sourceElementRoot : document;
+
 		/**
 		 * The {@link module:engine/controller/editingcontroller~EditingController editing controller}.
 		 * Controls user input and rendering the content for editing.
@@ -182,7 +186,6 @@ export default class Editor {
 		 * @readonly
 		 * @member {module:engine/controller/editingcontroller~EditingController}
 		 */
-		const sourceElementRoot = config.sourceElementRoot ? config.sourceElementRoot : document; // eslint-disable-line no-undef
 		this.editing = new EditingController( this.model, stylesProcessor, sourceElementRoot );
 		this.editing.view.document.bind( 'isReadOnly' ).to( this );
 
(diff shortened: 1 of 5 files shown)


Comandeer:
@ywsang, is such a solution compatible with browsers other than Blink-based ones? AFAIR the `shadowRoot.getSelection()` method is Chrome-only and [there is no clear consensus](https://github.com/w3c/webcomponents/issues/79) how to handle selection in shadow roots.

Reinmar:
That'd be good to know, indeed. Also, in order to actually test the PR a [manual test](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/testing-environment.html#running-manual-tests) would be helpful.

Reinmar (changes requested):
As commented – there are concerns about browser compatibility. Also, we need to be able to test it on a manual test.

mfreed7:
Hello, I just wanted to check in on this issue, and point out [a new proposal for a Shadow DOM compatible Selection API](https://github.com/mfreed7/shadow-dom-selection#proposal). This proposal has been being discussed on https://github.com/WICG/webcomponents/issues/79. In particular, [this comment](https://github.com/WICG/webcomponents/issues/79#issuecomment-965900820) gives a quick TL;DR summary of the proposal.

In particular, I'd be curious to hear the opinions of the CKEditor maintainers as to whether this proposal will suit your needs:

 1. Is the [proposal](https://github.com/mfreed7/shadow-dom-selection#proposal) understandable to you?

 2. Does the proposed `getComposedRange()` API work for your use case?

 3. In particular, do the two parameters, `selectionRoot` and `shadowRoots` make sense? And would they allow you to easily make use of cross-scope selection information?

 4. Do the changes to the "setter" APIs like `setBaseAndExtent()` seem like they'll work for your use case?

 5. Do you have any need to get selection information "anywhere on the page", i.e. within any shadow root? This would likely require another parameter to make this easy, if so.

 6. Do you have any other concerns or feedback?

I'd love to hear your feedback either here in this issue, or (better) on the [Selection API thread](https://github.com/WICG/webcomponents/issues/79). These changes are still very much in development, so your feedback can have an impact. Thanks!

wimleers:
FYI: Drupal would also like to see this happen.

pomek:
The PR is a little outdated as the last update was made over a year ago, but I wanted to verify whether it resolves my case. After merging the latest `#master`, the selection in Shadow DOM and the editor's selection are not synchronized. 

Perhaps I failed when merging changes…

Inviz:
I was a bit disappointed to learn that CKEditor does not handle running in the web component. thanks to this PR though i shimmed it like this:

```
var orig = document.defaultView.getSelection;
document.defaultView.getSelection = function() {
  if (document.activeElement.shadowRoot) return document.activeElement.shadowRoot.getSelection(); 
  return orig.apply(this, arguments);
}
```

Obviously it's not great, but at least i didnt have to hack the source code.

PR #7847, opened 2020-08-14
Add CSS styles to classic editor when in a Web Component

ywsang:

Other (build-classic): Update webpack config to save styles to global array.
Change any CSS rules with the :root selector to :root, :host to accommodate Web Components.

---

### Additional information

Related to #3891 and  #1483.

By default, `style-loader` will add the CSS styles to the `<head>` of the document. However, document styles do not get applied to shadow roots as part of Shadow DOM's style encapsulation. Overriding the default behavior by providing a function to the `insert` option saves all of the styles in a global array in addition to adding them to the document `<head>` allows for adding the styles from the array into a shadow root after it's been initialized.  

For example, if creating a web component with CKEditor, one could do the following in the constructor:
```
constructor() {
...
this.attachShadow( { mode: 'open' } );
  if ( window.ckeStylesToInject ) {
      window.ckeStylesToInject.forEach( style => {
        this.shadowRoot.prepend( style.cloneNode(true) );
      });
  } 
  ...
}
```
to add the styles to the shadow root. 

This PR also changes any CSS rules with the :root selector to :root, :host, as :root doesn't select anything inside of a web component.

### DIFF
--- a/docs/assets/snippet-styles.css
+++ b/docs/assets/snippet-styles.css
@@ -2,14 +2,14 @@
  * Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
  */
 
-:root {
+:root, :host {
 	/* This custom property is read by the JS and passed to editor configurations
 	as config.toolbar.viewportTopOffset. */
 	--ck-snippet-viewport-top-offset: 100
 }
 
 @media (max-width: 960px) {
-	:root {
+	:root, :host {
 		/* The mobile layout of Umberto is different and the toolbar offset must be
 		smaller (https://github.com/ckeditor/ckeditor5/issues/1348). */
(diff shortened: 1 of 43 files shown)


Reinmar:
Hi @ywsang! Sorry for the late reply here. 

This is an interesting approach. We'd need to add this snippet to all webpack configs (there are quite many of them :D) and cover this in the docs (in the advanced setup guide). 

What we'd need to be able to take over this from you is a sample of how you use CKEditor 5 in a shadow root and information on which browsers you test. If you could share that information, it'd ensure that we're looking at the same thing.

Reinmar (changes requested):
As commented. Strongly related to https://github.com/ckeditor/ckeditor5/pull/7846#pullrequestreview-488609037.

PR #7845, opened 2020-08-14
#7803: Created the list styles UI

oleq:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Internal (list): Created the list styles UI. Closes #7803.

Feature (theme-lark): Creates styles for the `ListStylesUI` plugin (see #7803).

BREAKING CHANGE (ui): It is now possible to override existing components when [adding new ones](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_componentfactory-ComponentFactory.html#function-add) to the [component factory](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_componentfactory-ComponentFactory.html) (previously an error was thrown) (see #7803).

---

Issue #7803, opened 2020-08-07
Implement UI part for List Styles feature


pomek:
Part of https://github.com/ckeditor/ckeditor5/issues/1031.

pomek:
Blindly estimated as 2.

oleq:
# Designs

## Unordered list styles

![](https://user-images.githubusercontent.com/1099479/90112678-9b3fe700-dd50-11ea-8a8e-46a249a74b5a.png)

## Ordered list styles

![](https://user-images.githubusercontent.com/1099479/90112709-a2ff8b80-dd50-11ea-83d6-e2bd55020abd.png)

## Button states

I followed the hover styles first implemented in [special characters](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/special-characters.html) and combined them with heading dropdown active styles.

![](https://user-images.githubusercontent.com/1099479/90112810-c4f90e00-dd50-11ea-9161-bd64e3b33957.png)

![](https://user-images.githubusercontent.com/1099479/90112824-c9252b80-dd50-11ea-8c4e-62fa4faca46f.png)

![](https://user-images.githubusercontent.com/1099479/90112841-cde9df80-dd50-11ea-9cf8-0cd2b6e438dc.png)

![](https://user-images.githubusercontent.com/1099479/90112859-d3472a00-dd50-11ea-9551-8ff11840811c.png)

## The question of default/revert

### **Context and Problem Statement**

My designs contain the "reset/default/revert" button although I haven't found any WYSIWYG editor out there that would use such thing. I wanted to see how would it look. The button is a tricky thing because it's a feature of CSS world but a potential UX anti-pattern.

### **Decision drivers**

*   The feature should be as useful as it is possible to the users (they should be able to do everything with it – apply styles, change and reset them).
*   Users should not be confused about what each button does.

### **Considered options**

#### Use the "reset" button

1.  When clicked, the "reset" button removes the attribute completely from the model.
2.  In this scenario all buttons work one way (previous style -> new style). Clicking the same button twice will not toggle/reset the style.

#### Don't use the "reset" button

1.  In this scenario, assuming that if a selection is inside a list but there's **no style** applied in the model, simply **none** of the buttons will be active.
    1.  Alternatively, we can activate the "most likely style". If a list is numbered, we can assume that most CMSs/UAs/web pages will style it as "numbered" ("1.").
[...]

PR #7861, opened 2020-08-18
List style feature

### LINKED ISSUE #7801: Implement editing part for List Styles feature
Part of https://github.com/ckeditor/ckeditor5/issues/1031.

pomek:

Feature (list): Introduced the list styles feature that allows customizing the list's marker of the list item elements. Closes #7801.

Feature (theme-lark): Creates styles for the `ListStylesUI` plugin (see #7803).

MINOR BREAKING CHANGE (ui): It is now possible to override existing components when [adding new ones](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_componentfactory-ComponentFactory.html#function-add) to the [component factory](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_componentfactory-ComponentFactory.html) (previously an error was thrown) (see #7803).

---

### Additional information

UI part of the feature was reviewed in https://github.com/ckeditor/ckeditor5/pull/7845. 
Docs: https://github.com/ckeditor/ckeditor5/pull/7881 (can be merged after this PR).


pomek:
Keep in mind that I had some issues with the implementation and the feature could contain some bugs. See https://github.com/ckeditor/ckeditor5/issues/7879.

pomek:
@ckeditor/qa-team, could you check how critical are issues described in the #7879 issue? I mean, are we able to merge the feature core and fix bugs later?

packages/ckeditor5-list/src/liststylesui.js
```diff
@@ -0,0 +1,228 @@
...
+				}
+				// If the style was the same, remove it (the button works as an off toggle).
+				else {
+					editor.execute( 'listStyles', { type: 'default' } );
+				}
+			}
+			// If the content the selection is anchored to is not a list, let's create a list of a desired style.
+			else {
+				editor.execute( parentCommandName );
+				editor.execute( 'listStyles', { type } );
```
pomek, line 220: Shouldn't we wrap the code in the `model#change()` block in order to have single undo step?

Reinmar:
One thing to keep in mind here is that the number of edge cases and weird cases and unclear cases will be significant. I'd heavily prioritize stability over correctness.

The other thing is that we're approaching the code freeze, so minor code refactorings should be split to a separate issue.

packages/ckeditor5-list/src/liststyles.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @module list/liststyles
+ */
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import ListStylesEditing from './liststylesediting';
+import ListStylesUI from './liststylesui';
+
+/**
+ * The to-do list feature.
```
pomek, line 15:
```suggestion
 * The list styles feature.
```

Mgsy:
## Steps to reproduce

1.  Select multiple cells in a table.
2.  Apply a list style.

## Current result

The regular list is applied.

## GIF

Mgsy:
## Steps to reproduce

1.  Set the following data - `<ol><li>test 1</li></ol><ul><li>test 2</li></ul><ol><li>test 3</li></ol>`.
2.  Select all list items.
3.  Apply some list style.

## Current result

Only the first item receives styling.

## Expected result

All lists will be merged.

## GIF

packages/ckeditor5-list/src/liststyles.js
```diff
@@ -0,0 +1,36 @@
...
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import ListStylesEditing from './liststylesediting';
+import ListStylesUI from './liststylesui';
+
+/**
+ * The list styles feature.
+ *
+ * This is a "glue" plugin that loads the {@link module:list/liststylesediting~ListStylesEditing list styles editing feature}
+ * and the {@link module:list/liststylesolistui~ListStylesUI list styles UI feature}.
```
pomek, line 18:
```suggestion
 * and the {@link module:list/liststylesui~ListStylesUI list styles UI feature}.
```

Mgsy:
## Steps to reproduce

1.  Change a paragraph directly to a styled list.

## Current result

Two undo steps are created.

## Expected result

One undo step is created.

Mgsy:
## Steps to reproduce

1.  Select some text in a list item.
2.  Add bold.

## Current result

The editor crashes.

## Error

```
liststylesediting.js:97 Uncaught TypeError: Cannot read property 'getStyle' of null
    at UpcastDispatcher.dispatcher.on.priority (liststylesediting.js:97)
    at UpcastDispatcher.fire (emittermixin.js:209)
    at UpcastDispatcher.convert (upcastdispatcher.js:207)
    at Object.callback (datacontroller.js:414)
    at Model._runPendingChanges (model.js:830)
    at Model.change (model.js:175)
    at DataController.toModel (datacontroller.js:413)
    at MutationHandler._handleContainerChildrenMutations (injecttypingmutationshandling.js:109)
    at MutationHandler.handle (injecttypingmutationshandling.js:64)
```

FilipTokarski:
### Steps: 
1. Add two styled list items 
2. Apply link on the second one
3. Select the second one and press `Bulleted list` on the toolbar

**Expected:** list changes to unordered

**Actual:** list changes to unordered only when you choose the specific style

01_list3

Mgsy:
## Steps to reproduce

1.  Create a new list with one nested item ( `<ol><li>Test<ol><li>Test 2</li></ol></li></ol>` ).
2.  Select all items.
3.  Add styling.

## Current result

Only the root item is changed.

FilipTokarski:
If you select a few indented list items, toggling betweend bulleted and numbered styles all of them. But if you want to add some specific styling - only the first item gets styled:  
01_list4

packages/ckeditor5-list/src/liststylescommand.js
```diff
@@ -0,0 +1,206 @@
...
+			break;
+		}
+
+		// If current parsed item has lower indent that element that the element that was a starting point,
+		// it means we left a nested list. Abort searching items.
+		//
+		// ■ List item 1.       [listIndent=0]
+		//     ○ List item 2.[] [listIndent=1], limitIndent = 1,
+		//     ○ List item 3.   [listIndent=1]
+		// ■ List item 4.       [listIndent=0]
```
jodator, line 162: :+1:  for styling :)

packages/ckeditor5-list/src/liststylesediting.js
```diff
@@ -0,0 +1,511 @@
...
+		}
+	}
+}
+
+// Returns a converter that consumes the `style` attribute and search for `list-style-type` definition.
+// If not found, the `"default"` value will be used.
+//
+// @private
+// @returns {Function}
+function upcastListItem() {
```
jodator, line 93: I'd go with a name that reflects upcasting list item's style. The name `upcastListItem()` might suggest that it handles whole `<li>` upcast.

packages/ckeditor5-list/src/liststylesui.js
jodator: It might do the trick AFAIR. So for now might be OK.

The other thing that came to my mind might be that `ListSyyle` command _might_ also do the `parentCommand` stuff but as per PR requirements to not-block this PR a wrap will be quicker.

packages/ckeditor5-ui/src/componentfactory.js
```diff
@@ -77,20 +77,6 @@ export default class ComponentFactory {
...
-			 *
-			 * @error componentfactory-item-exists
-			 * @param {String} name The name of the component.
-			 */
-			throw new CKEditorError(
-				'componentfactory-item-exists: The item already exists in the component factory.',
-				this,
-				{ name }
-			);
-		}
```
jodator, line 92: This is a change that probably would need to be discussed.

jodator: I've scanned the #7803 but I didn't find why this change.

jodator: OK so this is mostly to @oleq - I don't get from the issue why. The benefits are OK though.

@pomek - please ignore my comments here ;)

packages/ckeditor5-list/src/liststylescommand.js
```diff
@@ -0,0 +1,206 @@
...
+	 *
+	 * @param {Object} options
+	 * @param {String|null} options.type The type of the list styles, e.g. 'disc' or 'square'. If `null` specified, the default
+	 * style will be applied.
+	 * @protected
+	 */
+	execute( options = {} ) {
+		const model = this.editor.model;
+		const document = model.document;
+		const position = findPositionInsideList( document.selection );
```
jodator, line 57: This might cause "single item changed on range selection" issues.

From top of my head:

1.  Get elements/blocks from selection (to would get all selected list items).
2.  Use `selection.getFirst/LastPostion()` to get backward/forward siblings.

jodator (changes requested):
I've saw that there are some issues found so maybe checking if using `differ` instead of command's events might be relevant.

Abut naming - I've asked that question on the Slack channel as I'm not sure why the plural form and I might be lacking some conversation details.

If I commented something in the UI part - please ignore that. I've forgot instantly that this PR was partially reviewed. I would have some insights for retro though :D 

I'll do the jsdocs/tests re-read later on.

---

To sum up and to not block:

1.  The feature works, with minor bugs. Which, at least IMHO, we could pass.
2.  The naming might be the most important one before releasing (PITA to do, but should be fairly fast).
3.  crashes to be addressed/checked first.
4.  The event thing (if no time left) should be at least hidden under @Private to have it refactor later as I'm almost 100% sure that we could do the cleanup without it.

jodator:
So I've also found similar issues to the above:



The 2-commands action behaves strangely because the first command (list type) changes all items to the other type and the second command (list style) applies style only to the top-most item.

So I think that my comment regarding how this command should work is valid (take elements from range or just the same ones as list type command does).

ps.: I've might found a bug in the original command as the one item does not change it's type :open\_mouth: .

jodator:
> ps.: I've might found a bug in the original command as the one item does not change it's type open\_mouth .

It's some kind of downcast rendering glitch. I can't reproduce it on fresh content.

packages/ckeditor5-list/src/indentcommand.js
```diff
@@ -90,6 +91,16 @@ export default class IndentCommand extends Command {
...
+
+			/**
+			 * Event fired by the {@link #execute} method.
+			 *
+			 * It allows to execute an action after executing the {@link ~IndentCommand#execute} method, e.g. adjusting
+			 * attributes of changed list items.
+			 *
+			 * @event executeCleanup
+			 */
+			this.fire( 'executeCleanup', itemsToChange );
```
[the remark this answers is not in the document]
pomek: I marked the event as private and added `_` as prefix. (0143084)
But you're right. Post-fixer could be better solution here.

packages/ckeditor5-list/src/liststylescommand.js
pomek: Thx :D I was wondering how to explain what I am doing in particular functions. I guess this "trick" explains everything.

pomek:
So as for now, we have three main issues:

1. the command changes the most outdented list and ignores nested ones (assuming that selection has some)
2. sometimes render fails (refreshing the same scenario helps)
3. `_executeCleanup` pattern is not correct but it could be done as a followup after merging the PR.

> The naming might be the most important one before releasing (PITA to do, but should be fairly fast).

Now the feature is called `ListStyle`.

jodator:
@pomek I've updated an unrelated test name as it was bugging me ;)

jodator:
@Reinmar if @pomek doesn't need to add anything else here I think that this might go to the testing phase as well.

jodator (approved):

pomek:
Let me add a few tests that you mentioned and I'll check whether `ListStyleCommand` has all use cases covered.

pomek:
@jodator, I added missing stuff. It's ready for merge.

Issue #7383, opened 2020-06-05
Dynamic configuration options


jodator:
## 📝 Provide a description of the improvement

After quick search i can see that most of the features reads configuration options at the plugin init stage. Only few read configuration option afterwards but that feels like a side-effect of the code rather then dynamic configuration.

Should we have dynamic configuration for some options?

Pros:

*   ability to change some values on the fly

Cons:

*   making it dynamic require to re-render UI, update some state and the previously created content may not have sense

References issues: #7278.

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

Reinmar:
I was always against dynamic config as it creates many problems:

*   Which properties should be dynamic? There are some which cannot be due to immense technical issues that that would cause. There are some that will never be changed by anyone. So, the ROI here would need to be checked for every case.
*   The above means that the developer will have to change which property is dynamic anyway.
*   Some properties (e.g. lists of some values) might need a smarter way to be changed than just being overridden.
*   The complexity will rise with every such property.

So, I'm rather looking for API-driven ways to change the editor behavior on the fly. Every case is different and handling it via API, when needed,  makes to me the most sense.

Reinmar:
That could be one reason for having some of the configuration dynamic: ckeditor/ckeditor5-vue#125 (this, and the linked issue).

jodator:
Another reason for having configuration dynamic: #7871. And there's a solution that is kinda OK.

The plugin has an observable property which one can set after the plugin reads a configuration. The plugins listens on this property change and updates internal state.

bendemboski:
I agree that the notion of making _all_ configuration values dynamic is problematic -- it's sometimes not going to be possible for a plugin to handle config values changing, and even when it is it's likely to cause friction in writing plugins because of the effort needed to handle dynamic changes, even in cases where it's not very useful.

In my opinion, it should be opt-in by the plugin authors, which I think means one of three things:

1. Provide some patterns or implement some mixin-style utilities to make it easy for plugins to provide an API for updating configuration values outside the `Config` class
2. Modify the `Config` class to allow plugins to define properties as dynamic, and provide an API on `Config` for updating them (that maybe throws an error when trying to update a non-dynamic config value)
3. Make all config values technically dynamic it that they provide notifications when the values are changed, but plugins aren't expected to respond to any/all changes, but rather to document which ones can and can't be dynamic.

[...]

PR #7871, opened 2020-08-18
Allow dynamic updates to the text transformations

bendemboski:
Opening this PR for discussion.

In our application we want to allow the user to configure their own set of text transformations, and be able to apply those changes to already-instantiated editors. So we want a way to dynamically update the available transformations in an editor.

This PR is a proof-of-concept, but setting a property directly on the plugin may not fit with CKEditor's design patterns. Other options I can think of:

1. Implement a method on the plugin to set the transformations, rather than a settable property
2. Allow [TextTransformationConfig#extra](https://ckeditor.com/docs/ckeditor5/latest/api/module_typing_texttransformation-TextTransformationConfig.html#member-extra) to either be an array or be a function callback.

(2) seems like the closest to other CKEditor features, e.g. the [MentionFeed#feed](https://ckeditor.com/docs/ckeditor5/latest/api/module_mention_mention-MentionFeed.html#member-feed). For our use case, there is no need to support async callbacks, but I could see an argument for it.

Any feedback on this? We're willing to do some more work to develop this idea into something merge-able if someone can share thoughts on the right direction to go.

### DIFF
--- a/packages/ckeditor5-typing/src/texttransformation.js
+++ b/packages/ckeditor5-typing/src/texttransformation.js
@@ -106,6 +106,11 @@ export default class TextTransformation extends Plugin {
 			this.isEnabled = !modelSelection.anchor.parent.is( 'element', 'codeBlock' );
 		} );
 
+		this.on( 'change:transformations', ( evt, propertyName, newValue ) => {
+			this.normalizedTransformations = normalizeTransformations( newValue ) || [];
+		} );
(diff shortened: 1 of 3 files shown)


jodator:
Hi @bendemboski! Thanks for the PR and a use-case. Sorry for late response but we've been busy with the release process.

We already have this issue for dynamic configuration (#7383). However, there's no conclusion on how to resolve that. I'll comment there as well.

Issue #7884, opened 2020-08-20
Pasting with shift pressed produces extra line breaks


mlewand:
## 📝 Provide detailed reproduction steps (if any)

1.  Open [https://ckeditor.com/docs/ckeditor5/21.0.0/examples/builds/classic-editor.html](https://ckeditor.com/docs/ckeditor5/21.0.0/examples/builds/classic-editor.html)
2.  Copy two subsequent paragraphs of text
3.  Put the selection inside of the editor
4.  Paste with <kbd>ctrl/cmd</kbd> + <kbd>shift</kbd> + <kbd>v</kbd>

### ✔️ Expected result

The same behavior as with <kbd>ctrl/cmd</kbd> + <kbd>v</kbd> alone

### ❌ Actual result

![](https://user-images.githubusercontent.com/5353898/90727216-32fa8380-e2c3-11ea-9a6b-42607bcaeec7.png)

## 📃 Other details

[During the course it turned out](https://github.com/ckeditor/ckeditor5/issues/7884#issuecomment-689410602) that this should actually be a feature rather a bug, as this behavior comes from the browser putting extra EOL chars in the clipboard.

*   CKEditor version: v21

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

mlewand:
1. We put a double EOL character after each paragraph.
1. [We are losing soft line breaks](https://github.com/ckeditor/ckeditor5/issues/8045).

I was surprised to learn that the first problem might be actually inherited from contenteditable itself.

A comparison on how other editors handle this:

<table>
<thead>
<tr>
<th>CKEditor 5</th>
<th>Microsoft Word</th>
<th>Google Docs</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<pre><code>Paragraph

new paragraph

new paragraph

new paragraph

new paragraph

softBreak follows:softBreak follows:softBreak follows:

new paragraph</code></pre>
</td>
<td>
<pre><code>Paragraph
new paragraph
new paragraph
new paragraph
new paragraph
softBreak follows:
softBreak follows:
softBreak follows:
new paragraph
</code></pre>
</td>
<td>
<pre><code>Paragraph
new paragraph
new paragraph
new paragraph
new paragraph
softBreak follows:
softBreak follows:
softBreak follows:
new paragraph
</code></pre>
</td>
</tr>

<tr>
<th>CKEditor 4</th>
<th>Plain content editable (Chrome)</th>
[...]

PR #7886, opened 2020-08-20
Improved paste as plain text

### LINKED ISSUE #7799: Polish the plain text pasting
I'm creating a ticket to follow-up on @Reinmar from the #7778 PR.

> 1.  Firefox pastes `foo` as both `text/html` and `text/plain` - a simple string comparison would be enough to set `isPlainText`

Good idea, we could consider that. That'd be a bit more complex as the HTML could look like this: `'foo &nbsp;bar&gt;bom'` (`foo  bar<bom`) but with some simple regexp replace it's doable.

As for the other things that @jodator wrote, I think that we need to be predictable:

*   If you pressed ctrl+shift+v it's very likely that you want to inherit selection attributes. But then @pomek's solution won't handle multi-line text as in the model it'd contain soft breaks or paragraphs. Do we want to handle multi-line text the same way as a single line? If yes, we need an additional check for that.
*   If you pressed ctrl+v but you had very simple content in the clipboard, how likely is it that you want it to inherit selection styles? I'd say, quite likely, so I found @pomek's idea really interesting. However, I'm slightly worried about the randomness of this solution. I copied some formatted text from paste from office or other application and that formatting is not supported by the editor – that will make this text to inherit the selection styles. But if a supported style was included in that text, the style will not be inherited. It may feel "random". But that's where I count on your intuition and experience, @Mgsy. Can we live with that?

_Originally posted by @Reinmar in https://github.com/ckeditor/ckeditor5/pull/7778#issuecomment-669833922_

---

Actions needed to close:

- [ ] 1. check multi-line, plain text <kbd>shift+ctrl+v</kbd> scenario behavior - what is in the clipboard? Add tests. // cc @pomek
- [ ] 2. upon that - yes/no for inheriting text attributes on pasting multi-line text using <kbd>shift+ctrl+v</kbd> // waiting for :point_up:
- [ ] 3. decide on "random" scenario of pasting unsupported formatting resulting in plain text-like paste // cc @Mgsy

mlewand:

Feature: Introduced the `PastePlainText` feature that detects pasting with <kbd>ctrl/cmd</kbd> + <kbd>shift</kbd> + <kbd>ctrl/v</kbd> keystroke. Closes #7799.

---

### Additional information

I believe this fix should be blocked by #7884 - otherwise this feature puts this bug into a spotlight.

I'm not particularly happy with this nasty _glibalish_ variable in https://github.com/ckeditor/ckeditor5/blob/264951d7971f74fa7722f1bad13fd8a625de0fa8/packages/ckeditor5-clipboard/src/pasteplaintext.js#L39-L41 but I didn't find any trace of `shiftKey` in entire clipboard event chain.

I would also do few more things here:

*   Move the code from #7778 also to the `PastePlainText` plugin.
*   ~Decide on where the plugin should be added. I proposed adding it to the builds (afcdf9e) but it might fit better in `Essentials`~ (the plugin became a dependency of Clipboard per [comment#678135091](https://github.com/ckeditor/ckeditor5/pull/7886#issuecomment-678135091))

But since it was out of scope of this issue I didn't touch it.

~During the implementation I found #7887 which required either fixing upstream issue or extra protection added in 15d495f.~ (later I changed implementation so that it is not affected)


pomek (approved):
LGTM. 

I am not merging due to https://github.com/ckeditor/ckeditor5/issues/7884. This should be resolved first. Then we can merge the PR.

PR #7929, opened 2020-08-24
#7928: Balloon panel arrows pointing south should have realistic shadows

### LINKED ISSUE #7928: Give balloon arrows pointing south some realistic shadows
## With the fix

---

oleq:

Other: Balloon panel arrows pointing south should have realistic shadows. Closes #7928.

---


panr:
Looks legit ✅ 

But I have a small question, just to be sure: we all agree that the arrow on the top of the popup doesn't need to have a shadow, right?

image

panr (approved):

oleq:
> But I have a small question, just to be sure: we all agree that the arrow on the top of the popup/balloon doesn't need to have a shadow, right?

Yes, the shadow is always cast to the south across the UI. So to keep things consistent and realistic in this case, the arrow shadow is rendered underneath the balloon and invisible.

panr:
OK, so I'll wait for the CI to be green and I'm gonna merge this 👍

PR #7960, opened 2020-08-28
I/7878: Change the insert image via URL form label depending on whether image is selected or not.

### LINKED ISSUE #7878: Should the label for the "Insert image via URL" form change

We can consider if the label for the form should change to `Update image via URL`, when the user selects the image in the content and then opens the dropdown (the form input is filled). 

ATM, in that scenario, only the `submit` button will change, from `Insert` to `Update`. The label for the **Insert image via URL** is static (doesn't change).

image

---

panr:

Other (image): Change the `insertImageViaUrl` form label depending on whether image is selected or not. Closes #7878.

Other (image): Add new method `ImageUploadPanelView#getIntegration()` to pick desired integration from the collection.


packages/ckeditor5-image/src/imageupload/ui/imageuploadpanelview.js
```diff
@@ -213,6 +215,17 @@ export default class ImageUploadPanelView extends View {
...
 	}
 
+	/**
+	 * Returns a view of the integration.
+	 *
+	 * @param {string} name The name of the integration.
+	 *
+	 * @returns {module:ui/view~View}
+	 */
+	getIntegration( name ) {
```
oleq, line 225: Since you implemented this method, maybe let's use it across https://github.com/ckeditor/ckeditor5/blob/i/7878/packages/ckeditor5-image/tests/imageupload/ui/imageuploadpanelview.js whenever possible? (instead of `view._integrations.get()`).

packages/ckeditor5-image/src/imageupload/ui/imageuploadpanelview.js
```diff
@@ -213,6 +215,17 @@ export default class ImageUploadPanelView extends View {
 		}, { priority: 'high' } );
 	}
 
+	/**
+	 * Returns a view of the integration.
+	 *
+	 * @param {string} name The name of the integration.
+	 *
```
oleq, line 222: Unnecessary new line.

oleq (changes requested):

PR #7992, opened 2020-09-02
Typo

godai78:
Typo in list item


AnnaTomanek:
Looks like you need to correct the PR as there are many more files included in the merge than just a single typo?

godai78:
Invalid.

Issue #4736, opened 2016-09-16
Use smarter balloon positioning for links that wrap to the new line


oleq:
![image](https://cloud.githubusercontent.com/assets/1099479/18589640/edcaf8a0-7c2b-11e6-8086-114a8328d453.png)

The idea is to use [`getClientRects`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects) rather than `getBoundingClientRect` and determine to which rect the balloon should be attached. 

In the first implementation it can be any of the rects, just to avoid the balloon displayed in the middle of editable, attached to nothing.

In the second implementation, some smart positioning algorithm could be used, i.e. stick the balloon to the rect, which encloses the selection (caret). So depending if the user edits in the first line (before wrap) or in the second line (after the wrap), the balloon would appear right there. OTOH, it could produce a "flashing balloon" experience if the user edited textual content of the link close to the wrap limit, so the link wraps and unwraps repeatedly.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity over the last year. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #8015, opened 2020-09-03
Fixed the link balloon positioning in case of a collapsed selection

### LINKED ISSUE #7682: Consider enabling AutoLink in all builds

Demo is here: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/link.html#autolink-feature

Do we want this feature in all builds? I'm :+1:

### LINKED ISSUE #7926: Link balloon displayed in wrong place

1.  Create a link like this: `<$text linkHref="aaa">foo</$text><$text linkHref="aaa" bold="true">bar</$text>[]`
2.  Click the link toolbar button

### ✔️ Expected result

The link Baloon should appear close to the caret.

### ❌ Actual result

The link balloon appears in an incorrect place.

## 📃 Other details

*   Browser: …
*   OS: …
*   CKEditor version: …

---

mlewand:

Fix (link): Fixed the link balloon positioning in case of a collapsed selection in certain edge scenarios. Closes #7926.

---

### Additional information

It was caused by the fact that _sometimes_ empty native DOM range not return client rects (and 0ed bounding client rect). That caused the balloon to be mispositioned. Elements don't have this problem so the easiest solution is to bind balloon to the marker element.


mlewand:
There's a one case where the UX changed. With logic in current branch we go back to the behavior that was in v21.0.0 (it was coincidentally changed in v22.0.0). It's about balloon positioning in case of multiple fake selection parts.

Note the balloon pointing place.

### master (v22.0.0)



### branch `i/7926`



### v21.0.0



---

Personally, I find v22.0.0 behavior better but wanted to ask the team before putting some effort into it as I might be alone in that opinion 🙂 Anyway the fix should not be difficult: in https://github.com/ckeditor/ckeditor5/blob/61845fe9bf4fd0e95e0c016ae8b3ac1b0dffd4ab/packages/ckeditor5-link/src/linkui.js#L596 instead of working with a single marker element we should pick position before the first marker and after the last marker and create a range from these two positions. The only downside is that it will complicate unit tests just a bit.

oleq:
I'm OK with the new approach. This whole thing is related to #4736 BTW.

oleq:
Now that I think about it... we could easily resolve #4736 by making attaching the balloon to the first `Rect` when the selection is forward and the last `Rect` when it's backward. WDYT @mlewand?

oleq:
BTW are you sure 235600c is in the scope of this PR?

oleq:
> I'm OK with the new approach. This whole thing is related to #4736 BTW.

Now that I tested it a bit I have second thoughts about this. It looks confusing from the user perspective that the balloon for creating a link is displayed in one place but when the link is created the balloon moves somewhere else:



Can we smooth this out? The previous implementation was far less confusing.

oleq:
> Pushed `i/7926-combined-range` branch where I implemented what I suggested at the end of [previous comment](https://github.com/ckeditor/ckeditor5/pull/8015#issuecomment-686380359).

I tested this branch and I like it ❤️ Can you use it for this PR?

oleq:
FYI, I found two other (but unrelated) bugs when testing this issue:

*   #8029 
*   #8030

mlewand:
@ckeditor/qa-team can you give it some testing please? The logic affects only the positioning of the balloon but there are quite a few cases possible. As seen in the discussion above we decided to change the positioning to better fit most cases.

FilipTokarski:
Looks ok, I didn't find anything new 👍

Issue #7995, opened 2020-09-02
Implement the beforeInput support detection


oleq:
## 📝 Provide a description of the new feature

The event is supported in all major web browsers [except Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=970802) ([still an experimental feature behind a flag](https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features)). This means for the time being both `beforeInput` and mutation typing will need to exist next to each other. And even after Firefox enables the feature, it will take some time for people to update their browsers (which is slow in corporate environments).

This is why `isBeforeInputSupported = !env.isGecko` does not make sense. We need a feature discovery similar to [the one described on the React issue tracker](https://github.com/facebook/react/issues/11211#issuecomment-355866473) instead.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

Reinmar:
facebook/react#11211 and confirmation in https://stackoverflow.com/questions/49927038/how-to-feature-detect-input-events-level-1-or-2-support

oleq:
Closed in #8021.

PR #8022, opened 2020-09-03
#7995: Implemented the Input Events Level 1 support detection.

oleq:

Feature (utils): Implemented the Input Events Level 1 support detection. Closes [#7995](https://github.com/ckeditor/ckeditor5/issues/7995).

### DIFF
--- a/packages/ckeditor5-utils/src/env.js
+++ b/packages/ckeditor5-utils/src/env.js
@@ -81,7 +81,7 @@ const env = {
 		 *
 		 * @type {Boolean}
 		 */
-		areInputEventsLevel1Supported: areInputEventsLevel1Supported( global.window )
+		isInputEventsLevel1Supported: isInputEventsLevel1Supported( global.window )
 	}
 };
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-utils/src/env.js
```diff
@@ -71,7 +73,15 @@ const env = {
...
-		isRegExpUnicodePropertySupported: isRegExpUnicodePropertySupported()
+		isRegExpUnicodePropertySupported: isRegExpUnicodePropertySupported(),
+
+		/**
+		 * Indicates that the environment supports at least [Input Events Level 1](https://www.w3.org/TR/input-events-1/)
+		 * (`input` and `beforeinput` events).
+		 *
+		 * @type {Boolean}
+		 */
+		areInputEventsLevel1Supported: areInputEventsLevel1Supported( global.window )
```
niegowski, line 84: IMHO we should stick to code style for the boolean properties so `isInputEventsLevel1Supported` and that seems reasonable because we want to check here if the Level 1 of the Input Events is supported.

packages/ckeditor5-utils/src/env.js
```diff
@@ -151,3 +161,16 @@ export function isRegExpUnicodePropertySupported() {
...
+ * Checks if the current environment supports at least [Input Events Level 1](https://www.w3.org/TR/input-events-1/)
+ * (`input` and `beforeinput` events).
+ *
+ * @param {Window} domWindow The DOM Window interface.
+ * @returns {Boolean}
+ */
+export function areInputEventsLevel1Supported( domWindow ) {
+	const inputEvent = new domWindow.InputEvent( 'input' );
+
+	return ( 'inputType' in inputEvent ) && ( 'getTargetRanges' in inputEvent );
```
niegowski, line 175:
```suggestion
    return 'inputType' in inputEvent && 'getTargetRanges' in inputEvent;
```

I just don't like redundant brackets but this might be just my personal preference 😉

PR #8037, opened 2020-09-07
Inheriting the `listStyle` attribute for elements inserted next to the existing list

### LINKED ISSUE #7932: Creating a numbered list above an unordered list results with a unordered

1.  Open [lists feature demo](https://ckeditor.com/docs/ckeditor5/latest/features/link.html) for the upcoming v21 release (not not published yet, so production link won't work).
2.  Focus the editor under the "List styles" section.
3.  Put the selection in the "Ingredients required for making a coffee:" paragraph.
4.  Press "Numbered list" button.

### ✔️ Expected result

Should create a numbered list.

### ❌ Actual result

It makes unordered list instead.

## 📃 Other details

It doesn't happen if there's no list styles demo.

*   Browser: Chrome (didn't check others, but my assumption is that it's a cross-browser thing)
*   OS: Windows
*   CKEditor version: v21.0.0

---

pomek:

Fix (list): The `listStyle` attribute should be inherited when inserting or replacing a `listItem` with the same kind of the list (the `listType` attribute for the inserted/modified item is equal to next/previous sibling list). Closes #7932.

### DIFF
--- a/packages/ckeditor5-list/src/liststyleediting.js
+++ b/packages/ckeditor5-list/src/liststyleediting.js
@@ -189,7 +189,6 @@ export default class ListStyleEditing extends Plugin {
 // Returns a converter that consumes the `style` attribute and search for `list-style-type` definition.
 // If not found, the `"default"` value will be used.
 //
-// @private
 // @returns {Function}
 function upcastListItemStyle() {
 	return dispatcher => {
@@ -206,7 +205,6 @@ function upcastListItemStyle() {
 // Returns a converter that adds the `list-style-type` definition as a value for the `style` attribute.
 // The `"default"` value is removed and not present in the view/data.
 //
-// @private
 // @returns {Function}
 function downcastListStyleAttribute() {
 	return dispatcher => {
@@ -271,7 +269,6 @@ function downcastListStyleAttribute() {
 //     ○ List item 2.[]
 // ■ List item 3.
 //
-// @private
 // @param {module:core/editor/editor~Editor} editor
 // @returns {Function}
 function fixListAfterIndentListCommand( editor ) {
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-list/src/liststyleediting.js
```diff
@@ -573,3 +556,34 @@ function restoreDefaultListStyle( editor ) {
...
+		if ( change.type === 'attribute' ) {
+			return change.range.start.nodeAfter;
+		}
+
+		if ( change.type === 'insert' ) {
+			return change.position.nodeAfter;
+		}
+
+		return null;
+	}
```
jodator: I'd move this out of the parent function. It is re-created on every `getChangedListItems()` call (taking whole scope, etc) and as it is pure it can be placed outside.

No special docs are needed for it as it is small an pretty self-explanatory.

packages/ckeditor5-list/src/liststyleediting.js
```diff
@@ -509,6 +506,10 @@ function fixListStyleAttributeOnListItemElements( editor ) {
 			return false;
 		}
 
+		if ( baseItem.getAttribute( 'listType' ) !== itemToChange.getAttribute( 'listType' ) ) {
+			return false;
+		}
+
```
jodator, line 507: I'm thinking aloud here:

*   I like that it has the same pattern as above check.
*   I don't like the fact that the last check _could_ be a return.

```javascript
return baseItem.getAttribute( 'listType' ) === itemToChange.getAttribute( 'listType' );
```

However, in this scenario both are OK so it's just a matter of a preference here I think.

For now, this function makes all the checks as I could think of and maybe keeping a same pattern has more benefits here rather than going with preferred `if-else` return form.

jodator (approved):
LGTM :+1: I had some minor comments - one dissertation about ifs and second one which I've applied already. 

ps.: Please check them as I'm going to close this PR as generally it was OK and the change was minor.

pomek:
@jodator, LGTM.

PR #8043, opened 2020-09-08
#7996: Improved InputObserver with useful input event data properties

### LINKED ISSUE #7996: Improve InputObserver so the event passes and normalizes additional information

ATM the observer does nothing special; It passes the event through. The `beforeInput` event comes with some interesting properties, though:

*   `type`,
*   `getTargetRanges()`,
*   `data`,
*   `dataTransfer,`
*   `isComposing`

### TODO:

*   maybe we could have some `is()` interface for the type, for instance `is( 'insertText' )`?
*   `data` and `dataTransfer` should be normalized because depending on OS/browser/`type` they complement each other (it's a mess...)
*   ranges should be extracted using `getTargetRanges()`

oleq:

Feature (engine): Improved `InputObserver` with useful input event data properties. Closes #7996.

Other (clipboard): Moved the `DataTransfer` helper class to ckeditor5-engine (see #7996).

Internal (image): Updated import paths and docs since the `DataTransfer` helper class has been moved to [ckeditor5-engine](https://www.npmjs.com/package/@ckeditor/ckeditor5-engine) (see #7996).

MINOR BREAKING CHANGE (clipboard): The `DataTransfer` helper class has been moved to [ckeditor5-engine](https://www.npmjs.com/package/@ckeditor/ckeditor5-engine). Please update the import path in your integrations to `'@ckeditor/ckeditor5-engine/src/view/datatransfer'` (see #7996).

---

### Additional information

\-

### DIFF
--- a/packages/ckeditor5-engine/src/view/observer/inputobserver.js
+++ b/packages/ckeditor5-engine/src/view/observer/inputobserver.js
@@ -8,11 +8,13 @@
  */
 
 import DomEventObserver from './domeventobserver';
+import DataTransfer from '../datatransfer';
 
 /**
  * Observer for events connected with data input.
  *
- * Note that this observer is attached by the {@link module:engine/view/view~View} and is available by default.
+ * **Note**: This observer is attached by {@link module:engine/view/view~View} and available by default in all
+ * editor instances.
  *
  * @extends module:engine/view/observer/domeventobserver~DomEventObserver
  */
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-engine/src/view/observer/inputobserver.js
```diff
@@ -24,21 +26,98 @@ export default class InputObserver extends DomEventObserver {
 	}
 
 	onDomEvent( domEvent ) {
-		this.fire( domEvent.type, domEvent );
+		const domTargetRanges = Array.from( domEvent.getTargetRanges() );
```
niegowski, line 29: Is an iterable always returned from `getTargetRanges`? 

BTW Spec says it returns array so maybe we don't need `Array.from()`?

packages/ckeditor5-engine/tests/view/observer/inputobserver.js
```diff
@@ -5,48 +5,192 @@
 
 import InputObserver from '../../../src/view/observer/inputobserver';
 import View from '../../../src/view/view';
-import env from '@ckeditor/ckeditor5-utils/src/env';
+import createViewRoot from '../_utils/createroot';
+import global from '@ckeditor/ckeditor5-utils/src/dom/global';
 import { StylesProcessor } from '../../../src/view/stylesmap';
+import DataTransfer from '../../../src/view/datatransfer';
```
niegowski, line 11: Those imports could be sorted according to our convention 😉

packages/ckeditor5-engine/tests/view/observer/inputobserver.js
```diff
@@ -5,48 +5,192 @@
...
+				domRange2.setEnd( domEditable.firstChild.firstChild, 2 );
+
+				fireMockNativeBeforeInput( {
+					getTargetRanges: () => [ domRange1, domRange2 ]
+				} );
+
+				expect( evtData.targetRanges ).to.have.length( 2 );
+
+				const viewRange1 = evtData.targetRanges[ 0 ];
+				const viewRange2 = evtData.targetRanges[ 1 ];
```
niegowski, line 139: Maybe it would be good to check if those ranges are view Range instances?

packages/ckeditor5-engine/src/view/observer/inputobserver.js
oleq: [You're right](https://www.w3.org/TR/input-events-2/#dom-inputevent-gettargetranges) :+1: [](https://www.w3.org/TR/input-events-2/#dom-inputevent-gettargetranges)

niegowski (approved):

PR #8044, opened 2020-09-08
i/7983: Prevent showing an error on pressing "enter" when autolink is selected.

### LINKED ISSUE #7983: Autolink throws an exception on pressing enter when the entire first block is selected

1.  Open https://ckeditor.com/docs/ckeditor5/22.0.0/features/link.html.
2.  Navigate to the Autolink feature section.
3.  Focus the editor.
4.  Put the selection in the first paragraph.
5.  Select whole paragraph with the keyboard (e.g. <kbd>home</kbd> followed by <kbd>shift + end</kbd>)
6.  Press <kbd>enter</kbd> key.

### ✔️ Expected result

Paragraph gets empty. No error is thrown.

### ❌ Actual result

An exception is thrown:

```
snippet.js:4 Uncaught TypeError: Cannot read property 'is' of null
    at Function._createAt (snippet.js:4)
    at cn.createPositionAt (snippet.js:4)
    at r.<anonymous> (snippet.js:4)
    at r.fire (snippet.js:4)
    at r.<computed> [as execute] (snippet.js:4)
    at Ft.execute (snippet.js:4)
    at T.execute (snippet.js:4)
    at a.listenTo.priority (snippet.js:4)
    at a.fire (snippet.js:4)
    at a.<anonymous> (snippet.js:4)
```

## 📃 Other details

**It also throws the same way if you select all the content, e.g. using** <kbd><strong>ctr/cmd</strong></kbd> **+** <kbd><strong>a</strong></kbd>**.**

*   Browser: Any
*   OS: Any
*   CKEditor version: v22.0.0

---

panr:

Fix (link): Prevented throwing an error on pressing the <kbd>enter</kbd> key when non-collapsed selection is on autolink. Closes #7983.

### DIFF
--- a/packages/ckeditor5-link/tests/autolink.js
+++ b/packages/ckeditor5-link/tests/autolink.js
@@ -110,6 +110,26 @@ describe( 'AutoLink', () => {
 			);
 		} );
 
+		it( 'does not add linkHref attribute on enter when the link is selected', () => {
+			setData( model, '<paragraph>[https://www.cksource.com]</paragraph>' );
+
+			editor.execute( 'enter' );
+
+			expect( getData( model ) ).to.equal(
+				'<paragraph>[]</paragraph>'
+			);
+		} );
+
+		it( 'does not add linkHref attribute on enter when the whole paragraph containing the link is selected', () => {
+			setData( model, '<paragraph>[This feature also works with e-mail addresses: https://www.cksource.com]</paragraph>' );
+
+			editor.execute( 'enter' );
+
+			expect( getData( model ) ).to.equal(
+				'<paragraph>[]</paragraph>'
+			);
+		} );
+
 		it( 'adds linkHref attribute to a text link after space (inside paragraph)', () => {
 			setData( model, '<paragraph>Foo Bar [] Baz</paragraph>' );
 
--- a/packages/ckeditor5-link/tests/autolink.js
+++ b/packages/ckeditor5-link/tests/autolink.js
@@ -130,23 +130,33 @@ describe( 'AutoLink', () => {
 			);
 		} );
 
+		it( 'adds linkHref attribute on enter when the link (that contains www) is partially selected (end)', () => {
+			setData( model, '<paragraph>https://www.ckso[urce.com]</paragraph>' );
+
+			editor.execute( 'enter' );
+
+			expect( getData( model ) ).to.equal(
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-link/tests/autolink.js
```diff
@@ -110,6 +110,56 @@ describe( 'AutoLink', () => {
...
+
+		it( 'adds linkHref attribute on enter when the link (that contains www) is partially selected (end)', () => {
+			setData( model, '<paragraph>https://www.ckso[urce.com]</paragraph>' );
+
+			editor.execute( 'enter' );
+
+			expect( getData( model ) ).to.equal(
+				'<paragraph><$text linkHref="https://www.ckso">https://www.ckso</$text></paragraph><paragraph>[]</paragraph>'
+			);
+		} );
```
oleq, line 141: Are you sure this is OK? I mean:

*   Why should the `www` prefix change the auto-link behavior? It does now. For instance, starting with `<paragraph>https://www.ckso[urce.com]</paragraph>` vs. `<paragraph>https://ckso[urce.com]</paragraph>` ends with a link (former) and no link (latter).
*   `<paragraph>https://www.ckso[urce.com]</paragraph>`  also ends up as a broken link: `linkHref="https://www.ckso"`. This is not a valid link (it's just a fragment of the original link) and from the user point of view, this is a bug.
*   It looks like something is wrong with `URL_REG_EXP`. I don't think we should ever auto-link if a link is partially selected in the first place.

If you agree with me, can you please create a follow-up issue @panr? The test can stay, it's harmless. Maybe a comment that this is something weird would do the trick for now.

oleq (approved):
The solution is OK. 

I'm having second thoughts about a particular test, though (and the scenario it revealed). Let's have it sorted out first and then I'll merge this PR.

packages/ckeditor5-link/tests/autolink.js
oleq: It's probably because the `URL_REG_EXP` considers `ckso` a domain:

oleq: cc @jodator

panr: @oleq I agree that this incorrect behavior. I was surprised as well at first that this didn't pass the first version of the test, but I thought that this is how should it be, knowing how we handle validation ;-D

I'll create a followup to this issue 👍

PR #8053, opened 2020-09-09
Added sections about Context and Watchdog to the React guide.

ma2ciek:

Docs: Added sections about Context and Watchdog features to the React guide. Part of ckeditor/ckeditor5-react#118.

---

### Additional information


ma2ciek:
Hm. I added two private links there :sweat\_smile:

pomek:
Workaround for private links: https://github.com/ckeditor/ckeditor5/commit/1a0c3fdda3177ba803f2611d4b881b1c1165d38b

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+					/>
+
+					<CKEditor
+						editor={ ClassicEditor }
+						data="<p>Hello from the second editor working with the context!</p>"
+						onReady={ editor => {
+							// You can store the "editor" and use when it is needed.
+							console.log( 'Editor2 is ready to use!', editor );
+						} }
+					/>
```
pomek, line 117: Missing `</CKEditorContext>`.

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+
+The `ckeditor5-react` integration comes with a ready-to-use component for the {@link features/context-and-collaboration-features Context feature} that is useful when used together with some {@link features/collaboration CKEditor 5 collaboration features}.
+
+```jsx
+import React, { Component } from 'react';
+import { CKEditor, CKEditorContext } from '@ckeditor/ckeditor5-react';
+
+// Assuming that application builds the editor from source.
+import Context from '@ckeditor/ckeditor5-core/src/context';
+import ClassicEditor from './ckeditor';
```
pomek, line 93:
```suggestion
import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor';
```

Right now it throws an error.

ma2ciek:
I'll change these links as well.

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+
+class App extends Component {
```
pomek, line 95: After fixing issues, the app builds with two not-looking-good editors:

image

The `[data]` property for both editors is ignored (I guess because of lack of plugins).

ma2ciek:
Fixed.

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+	render() {
+		return (
+			<div className="App">
+				<CKEditorContext context={ Context }>
```
pomek, line 99: I tried to pass an object as the config:

```js
const editorConfiguration = {
	plugins: [ Essentials, Bold, Italic, Paragraph ],
	toolbar: [ 'bold', 'italic' ]
};
```

And I ended with an error:

```
index.js:1 CKEditorError: context-initplugins-invalid-plugin: Only a plugin marked as a context plugin is allowed to be used with a context. Read more: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-context-initplugins-invalid-plugin
 {}
    at Context.initPlugins (http://localhost:3000/static/js/0.chunk.js:3959:15)
    at http://localhost:3000/static/js/0.chunk.js:4109:23
    at new Promise (<anonymous>)
    at Function.create (http://localhost:3000/static/js/0.chunk.js:4107:12)
    at Or._creator (http://localhost:3000/static/js/main.chunk.js:1218:181)
    at http://localhost:3000/static/js/main.chunk.js:1287:79 {phase: "initialization", willContextRestart: false}
```

We should explain which plugins are accepted by the context feature and why we can't provide a list of plugins that should be used for all editors. Also, a way how to do that would be nice.

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+		);
+	}
+}
+```
+
+The `CKEditorContext` component supports the following properties:
+
+* `config` &ndash; The CKEditor 5 context configuration.
+* `context` &ndash; {@link module:core/context~Context The CKEditor 5 context class}.
+* `isLayoutReady` &ndash; A property that delays the editor creation when set to true. Useful when the CKEditor 5 annotations or a presence list is used.
```
pomek, line 128: In other words – the property waits with the editor's initialization for other elements that must be available in the [DOM](https://developer.mozilla.org/en-US/docs/Glossary/DOM).

A question – how should I inform the component that the element (that we wait for) is ready?

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+* `id` &ndash; The context ID. When that property changes the component restarts the context with its editor and reinitialize it based on the current configuration.
```
pomek, line 129: Same with the `CKEditor#id` – is it obligatory?

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+* `onReady` &ndash; A function called when the context is ready and all editors inside were initialized with the `context` instance. This callback can be called after the initialization and after restarts of editors or the context.
```
pomek, line 130: Is the callback called when an editor restarted? Would be good to show a situation when the context restarts itself.

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+* `onError` &ndash; A function called when the context has crashed during the initialization or during the runtime. It receives the error instance and the error details (`phase` and `willContextRestart`) as arguments.
```
pomek, line 131: I would follow the same syntax as I proposed for `CKEditor#onError`.

docs/builds/guides/frameworks/react.md
```diff
@@ -114,19 +173,26 @@ If you use the {@link framework/guides/document-editor Document editor}, you nee
...
 						// Insert the toolbar before the editable area.
 						editor.ui.getEditableElement().parentElement.insertBefore(
 							editor.ui.view.toolbar.element,
 							editor.ui.getEditableElement()
 						);
+
+						this.editor = editor;
+					} }
+					onError={ () => {
+						this.editor.ui.view.toolbar.element.remove();
```
pomek, line 195: We shouldn't remove the toolbar if the editor will not restart. Also, would be good to write a dedicated section about the decoupled editor and the integration with the latest version (3+).

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
 * `editor` (required) &ndash; The {@link module:core/editor/editor~Editor `Editor`} constructor to use.
 * `data` &ndash; The initial data for the created editor. See the {@link builds/guides/integration/basic-api#interacting-with-the-editor Basic API} guide.
 * `config` &ndash; The editor configuration. See the {@link builds/guides/integration/configuration Configuration} guide.
-* `onInit` &ndash; A function called when the editor was initialized. It receives the initialized {@link module:core/editor/editor~Editor `editor`} as a parameter.
```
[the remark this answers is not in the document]
pomek: The CKEditor component will display an error in the console so we could cover that case in the docs.

pomek (changes requested):

docs/builds/guides/frameworks/react.md
```diff
@@ -67,18 +67,73 @@ The `<CKEditor>` component supports the following properties:
...
+* `id` &ndash; The editor ID. When that property changes the component restarts the editor with new data instead of setting it on an initialized editor.
```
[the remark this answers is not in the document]
ma2ciek: Actually, I followed the existing format :smile: Only editor is required and is marked as required. I can add the "optional" keyword, but then it'd be good to add it to other places as well :)

docs/builds/guides/frameworks/react.md
ma2ciek: Yes, because of the lack of plugins. That's why I decided to use `./ckeditor` instead. I can add more info there though.

docs/builds/guides/frameworks/react.md
ma2ciek: `isLayoutReady={true}`. Or maybe I misunderstood you?

docs/builds/guides/frameworks/react.md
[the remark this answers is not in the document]
ma2ciek: I don't agree. Since it isn't a part of the API anymore there is no obligation to mention it in docs at all.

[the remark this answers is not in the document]
ma2ciek: It should be mentioned just in the migration notes, we should not document all versions of our packages in the Readme.

docs/builds/guides/frameworks/react.md
ma2ciek: > We shouldn't remove the toolbar if the editor will not restart.

If the editor won't restart, then the missing toolbar actually won't change anything ;)

docs/builds/guides/frameworks/react.md
ma2ciek: This is up to the Context guide. I wouldn't duplicate the information here. This section is rather designed for people who know the CKEditor 5 Context feature.

docs/builds/guides/frameworks/react.md
ma2ciek: What do you mean by "show" - do you mean creating here a snippet that presents some use cases? Or a sample?

docs/builds/guides/frameworks/react.md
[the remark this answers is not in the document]
pomek: Ok, fine. I see `(required)` for the `editor` property. Let's ignore my comment.

docs/builds/guides/frameworks/react.md
pomek: Instead of creating a new file, we can add the definition of the editor in the component file. We will have all in one place.

docs/builds/guides/frameworks/react.md
pomek: So we could add a link to the context feature that described the options.

docs/builds/guides/frameworks/react.md
pomek: What does the option do?

docs/builds/guides/frameworks/react.md
pomek: I meant an example of a situation when context restarts.

docs/builds/guides/frameworks/react.md
pomek: Right now (version 2.x), if the editor crashes, the toolbar won't disappear.

docs/builds/guides/frameworks/react.md
ma2ciek: Oh, now I have read my comment again :facepalm:

docs/builds/guides/frameworks/react.md
ma2ciek: But the editor restarts and the correct toolbar will be recreated, so where's the problem? In the previous version, the editor would stay crashed.

ma2ciek: I can add an if statement though.

ma2ciek:
I need to check if both `onError()` and `onReady()` callbacks are called when an error occurs in the editor when the `<CKEditor>` component is wrapped in the \<`CKEditorContext>`.

pomek:
While I was searching for an answer to my questions caused another issue: https://github.com/ckeditor/ckeditor5-react/issues/186.

ma2ciek:
Can we move the issue to the follow-up? I mean, the most important thing here, is that the components will restart the editor/context features if an error occurs.

docs/builds/guides/frameworks/react.md
ma2ciek: I checked these docs and they lack some important parts...

docs/builds/guides/frameworks/react.md
ma2ciek: Do you mean creating a sample editor from source in a few lines?

docs/builds/guides/frameworks/react.md
ma2ciek: Also, the problem here is that currently, only collaboration plugins are the context plugins AFAIK so it's hard to create a snippet with private plugins.

docs/builds/guides/frameworks/react.md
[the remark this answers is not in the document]
ma2ciek: Marking as resolved. If you don't agree you can reopen.

docs/builds/guides/frameworks/react.md
pomek: Yes, exactly the same as the editor that is exported by the file. 

```js
class CustomEditor /* ... extends */ { }
CustomEditor.builtinPlugins = [ /* ... */ ];
```

ma2ciek:
I'll make sure that everything is ready and I'll ping you then @AnnaTomanek.

ma2ciek:
@AnnaTomanek, @godai78 could you check these changes?

godai78 (approved):

AnnaTomanek (approved):
I fixed some minor stuff, it looks fine now. Let's 🚀  it finally!

PR #8074, opened 2020-09-14
I/7907: Prevent throwing an error during creating a link from multi-block selection

### LINKED ISSUE #7907: Error thrown when creating a link on a non-collapsed selection

1.  Build docs for production.
2.  Visit http://fake.ckeditor.com/ck/5/ckeditor5/build/docs/ckeditor5/latest/examples/builds/balloon-block-editor.html.
3.  Create a selection like in the screencast.
4.  Attempt to create a link.

```
Uncaught TypeError: t.getCustomProperty(...) is not a function
    at ut.e.on.priority (snippet.js:4)
    at ut.fire (snippet.js:4)
    at ut.convertMarkerRemove (snippet.js:4)
    at ut.convertChanges (snippet.js:4)
    at snippet.js:4
    at nt.change (snippet.js:4)
    at Le.Bt.listenTo.priority (snippet.js:4)
    at Le.fire (snippet.js:4)
    at Le._handleChangeBlock (snippet.js:4)
    at cn._runPendingChanges (snippet.js:4)
i @ snippet.js:4
rethrowUnexpectedError @ snippet.js:4
fire @ snippet.js:4
convertMarkerRemove @ snippet.js:4
convertChanges @ snippet.js:4
(anonymous) @ snippet.js:4
change @ snippet.js:4
Bt.listenTo.priority @ snippet.js:4
fire @ snippet.js:4
_handleChangeBlock @ snippet.js:4
_runPendingChanges @ snippet.js:4
change @ snippet.js:4
_hideFakeVisualSelection @ snippet.js:4
_removeFormView @ snippet.js:4
_closeFormView @ snippet.js:4
(anonymous) @ snippet.js:4
fire @ snippet.js:4
t.listenTo.useCapture @ snippet.js:4
fire @ snippet.js:4
n @ snippet.js:4
```

## 📃 Other details

It's a regression introduced in v20.0.0.

*   Browser: Any
*   Any
*   CKEditor version: release (3928dab)

---

panr:

Fix: Prevented throwing an error during creating a link from multi-block selection. Closes #7907.

### DIFF
--- a/packages/ckeditor5-link/src/linkui.js
+++ b/packages/ckeditor5-link/src/linkui.js
@@ -662,16 +662,27 @@ export default class LinkUI extends Plugin {
 		const model = this.editor.model;
 
 		model.change( writer => {
+			const range = model.document.selection.getFirstRange();
+
 			if ( model.markers.has( VISUAL_SELECTION_MARKER_NAME ) ) {
-				writer.updateMarker( VISUAL_SELECTION_MARKER_NAME, {
-					range: model.document.selection.getFirstRange()
-				} );
+				writer.updateMarker( VISUAL_SELECTION_MARKER_NAME, { range } );
 			} else {
-				writer.addMarker( VISUAL_SELECTION_MARKER_NAME, {
-					usingOperation: false,
-					affectsData: false,
-					range: model.document.selection.getFirstRange()
-				} );
+				if ( range.start.isAtEnd ) {
+					const focus = model.document.selection.focus;
+					const nextValidRange = getNextValidRange( range, focus, writer );
+
+					writer.addMarker( VISUAL_SELECTION_MARKER_NAME, {
+						usingOperation: false,
+						affectsData: false,
+						range: nextValidRange
+					} );
+				} else {
+					writer.addMarker( VISUAL_SELECTION_MARKER_NAME, {
+						usingOperation: false,
+						affectsData: false,
+						range
+					} );
+				}
 			}
 		} );
 	}
@@ -699,4 +710,29 @@ export default class LinkUI extends Plugin {
 // @returns {module:engine/view/attributeelement~AttributeElement|null} Link element at the position or null.
 function findLinkElementAncestor( position ) {
 	return position.getAncestors().find( ancestor => isLinkElement( ancestor ) );
+}
+
+// Returns next valid range for the fake visual selection marker.
+//
+// @private
+// @param {module:engine/model/range~Range} range Current range
+// @param {module:engine/model/documentselection~DocumentSelection#focus} focus Selection focus
+// @param {module:engine/model/writer~Writer} writer Writer
+// @returns {module:engine/model/range~Range|false} New valid range for the fake visual selection marker
+// or "false" if a valid range can't be found.
+function getNextValidRange( range, focus, writer ) {
+	const nextStartPath = [ range.start.path[ 0 ] + 1, 0 ];
+	const nextStartPosition = writer.createPositionFromPath( range.start.root, nextStartPath, 'toNext' );
+	const nextRange = writer.createRange( nextStartPosition, range.end );
+
+	// Block creating a potential next valid range over the current range end.
+	if ( nextRange.start.path[ 0 ] > range.end.path[ 0 ] ) {
+		return writer.createRange( focus );
+	}
+
+	if ( nextStartPosition.isAtStart && nextStartPosition.isAtEnd ) {
+		return getNextValidRange( nextRange, focus, writer );
+	}
+
+	return nextRange;
 }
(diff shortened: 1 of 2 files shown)


pomek (changes requested):
It works when you select non-empty elements.

However, I am still able to reproduce the issue by following the steps described in the issue/comment – https://github.com/ckeditor/ckeditor5/issues/7907#issuecomment-679076029.

For the following model tree:

```
<paragraph>[</paragraph>
<paragraph>Foo]</paragraph>
```

Pressing CMD/Ctrl+K, typing the URL, and applying changes, ends with the error.

But everything works fine when I executed the link command via console.

packages/ckeditor5-link/src/linkui.js
```diff
@@ -700,3 +711,28 @@ export default class LinkUI extends Plugin {
...
 	return position.getAncestors().find( ancestor => isLinkElement( ancestor ) );
 }
+
+// Returns next valid range for the fake visual selection marker.
+//
+// @private
+// @param {module:engine/model/range~Range} range Current range
+// @param {module:engine/model/documentselection~DocumentSelection#focus} focus Selection focus
+// @param {module:engine/model/writer~Writer} writer Writer
+// @returns {module:engine/model/range~Range|false} New valid range for the fake visual selection marker
```
pomek, line 721: Where the function can return `false`?

packages/ckeditor5-link/src/linkui.js
```diff
@@ -700,3 +711,28 @@ export default class LinkUI extends Plugin {
 function findLinkElementAncestor( position ) {
 	return position.getAncestors().find( ancestor => isLinkElement( ancestor ) );
 }
+
+// Returns next valid range for the fake visual selection marker.
+//
+// @private
+// @param {module:engine/model/range~Range} range Current range
+// @param {module:engine/model/documentselection~DocumentSelection#focus} focus Selection focus
```
pomek, line 719:
```suggestion
// @param {module:engine/model/position~Position} focus Selection focus
```

mlewand:
I briefly scanned the code and it looks to me that the proposed solution is a workaround specific for link plugin, rather than a fix for a source of the problem.

The way we initially used markers is intuitive and others that will use markers might face the same problem.

panr:
> However, I am still able to reproduce the issue by following the steps described in the issue/comment – #7907 (comment).

@pomek Hmm... works on my machine 🤔 

image

and after confirmation:

image

panr:
@Mgsy could you check if the bug still appears for that scenario? https://github.com/ckeditor/ckeditor5/issues/7907#issuecomment-679076029

@pomek says it does, but I can't reproduce it after my fixes and there's the unit test for this and gives me ✅  🤔

Mgsy:
> [**@Mgsy**](https://github.com/Mgsy) could you check if the bug still appears for that scenario? [#7907 (comment)](https://github.com/ckeditor/ckeditor5/issues/7907#issuecomment-679076029)

I can't reproduce mentioned issue with changes introduced by this PR.

pomek:
Yeah. I had to use some old version (maybe link/src/* was checked out on the `#master`). I checked once again and the bug is gone.

panr:
Ok, good to hear that 😅 Just added another test, specifically for what @pomek has been testing before.

packages/ckeditor5-link/src/linkui.js
panr: Good catch, I forgot to update the description... 👍

pomek:
We agreed that the bug can be hooked deeper, in the engine. However, the PR fixes the bug and the editor does not crash anymore.

As for now, we're fine with that. But in the future, the bug can occur once again and if we will be fixing that, we should revert changes introduced in the PR.

packages/ckeditor5-link/src/linkui.js
```diff
@@ -700,3 +711,27 @@ export default class LinkUI extends Plugin {
...
 	return position.getAncestors().find( ancestor => isLinkElement( ancestor ) );
 }
+
+// Returns next valid range for the fake visual selection marker.
+//
+// @private
+// @param {module:engine/model/range~Range} range Current range
+// @param {module:engine/model/position~Position} focus Selection focus
+// @param {module:engine/model/writer~Writer} writer Writer
+// @returns {module:engine/model/range~Range} New valid range for the fake visual selection marker
```
pomek, line 721:
```suggestion
// @param {module:engine/model/range~Range} range Current range.
// @param {module:engine/model/position~Position} focus Selection focus.
// @param {module:engine/model/writer~Writer} writer Writer.
// @returns {module:engine/model/range~Range} New valid range for the fake visual selection marker.
```

pomek (approved):
LGTM.

PR #8082, opened 2020-09-15
Change token refresh mechanism to depend on the token expiration time

### LINKED ISSUE #316: Integrate sample tests (with src editor) with the default compile task
Related to: https://github.com/ckeditor/ckeditor5-dev-compiler/issues/24

jczapiewski-cksource:

Other (cloud-services-core): Change tokens refreshing mechanism to depend on the token expiration time. 

---

### Tickets to close

Closes cksource/ckeditor5-internal#316.

_Do not paste it in the commit merge message!_

---

### Additional information

After these changes token should be refreshed after half of a difference between the token expiration time and the actual time.

### DIFF
--- a/packages/ckeditor-cloud-services-core/package.json
+++ b/packages/ckeditor-cloud-services-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ckeditor/ckeditor-cloud-services-core",
-  "version": "22.0.0",
+  "version": "22.0.1",
   "description": "CKEditor Cloud Services Core API.",
   "keywords": [
     "ckeditor5",
--- a/packages/ckeditor-cloud-services-core/package.json
+++ b/packages/ckeditor-cloud-services-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ckeditor/ckeditor-cloud-services-core",
-  "version": "22.0.1",
+  "version": "22.0.0",
   "description": "CKEditor Cloud Services Core API.",
   "keywords": [
     "ckeditor5",
--- a/packages/ckeditor-cloud-services-core/src/token/token.js
+++ b/packages/ckeditor-cloud-services-core/src/token/token.js
@@ -128,6 +128,12 @@ class Token {
 		clearTimeout( this._tokenRefreshTimeout );
 	}
 
+	/**
+	 * Checks whether the provided token follows the JSON Web Tokens (JWT) format.
+	 *
+	 * @protected
+	 * @param {String} tokenValue The token to validate.
+	 */
 	_validateTokenValue( tokenValue ) {
 		// The token must be a string.
 		const isString = typeof tokenValue === 'string';
--- a/packages/ckeditor-cloud-services-core/src/token/token.js
+++ b/packages/ckeditor-cloud-services-core/src/token/token.js
@@ -129,16 +129,19 @@ class Token {
 	}
 
 	_validateTokenValue( tokenValue ) {
-		// the token must be a string
+		// The token must be a string.
 		const isString = typeof tokenValue === 'string';
-		// the token must be a plain string without ""
+
+		// The token must be a plain string without quotes ("").
 		const isPlainString = !/^".*"$/.test( tokenValue );
-		// the token must be in the correct format - has 3 parts joined by dots
+
+		// JWT token contains 3 parts: header, payload, and signature.
+		// Each part is separated by a dot.
 		const isJWTFormat = isString && tokenValue.split( '.' ).length === 3;
 
-		if ( !( isString && isPlainString && isJWTFormat ) ) {
+		if ( !( isPlainString && isJWTFormat ) ) {
 			/**
-			 * A token value must be in JWT format.
+			 * The provided token must follow the [JSON Web Tokens](https://jwt.io/introduction/) format.
 			 *
 			 * @error token-not-in-jwt-format
 			 */
--- a/packages/ckeditor-cloud-services-core/src/token/token.js
+++ b/packages/ckeditor-cloud-services-core/src/token/token.js
@@ -112,8 +112,6 @@ class Token {
 				if ( this._options.autoRefresh ) {
 					this._registerRefreshTokenTimeout();
 				}
-
-				return this;
 			} )
 			.then( () => this );
 	}
--- a/packages/ckeditor-cloud-services-core/src/token/token.js
+++ b/packages/ckeditor-cloud-services-core/src/token/token.js
@@ -175,8 +175,8 @@ class Token {
 	 *
 	 * If the token parse fails, the default DEFAULT_TOKEN_REFRESH_TIMEOUT_TIME is returned.
 	 *
-	 * @returns {Number}
 	 * @protected
+	 * @returns {Number}
 	 */
 	_getTokenRefreshTimeoutTime() {
 		try {
--- a/packages/ckeditor-cloud-services-core/src/token/token.js
+++ b/packages/ckeditor-cloud-services-core/src/token/token.js
@@ -46,6 +46,10 @@ class Token {
 			);
 		}
 
+		if ( options.initValue ) {
+			this._validateTokenValue( options.initValue );
+		}
+
 		/**
 		 * Value of the token.
 		 * The value of the token is null if `initValue` is not provided or `init` method was not called.
@@ -57,10 +61,6 @@ class Token {
 		 * @readonly
 		 */
 		this.set( 'value', options.initValue );
-
-		if ( options.initValue ) {
-			this._validateTokenValue();
-		}
 
 		/**
 		 * Base refreshing function.
@@ -96,8 +96,6 @@ class Token {
 				return;
 			}
 
-			this._validateTokenValue();
-
 			if ( this._options.autoRefresh ) {
 				this._registerRefreshTokenTimeout();
 			}
@@ -113,8 +111,8 @@ class Token {
 	refreshToken() {
 		return this._refresh()
 			.then( value => {
+				this._validateTokenValue( value );
 				this.set( 'value', value );
-				this._validateTokenValue();
 
 				if ( this._options.autoRefresh ) {
 					this._registerRefreshTokenTimeout();
@@ -130,12 +128,15 @@ class Token {
 		clearTimeout( this._tokenRefreshTimeout );
 	}
 
-	_validateTokenValue() {
-		if (
-			typeof this.value !== 'string' ||
-			/^".*"$/.test( this.value ) ||
-			this.value.split( '.' ).length !== 3
-		) {
+	_validateTokenValue( tokenValue ) {
+		// the token must be a string
+		const isString = typeof tokenValue === 'string';
+		// the token must be a plain string without ""
+		const isPlainString = !/^".*"$/.test( tokenValue );
+		// the token must be in the correct format - has 3 parts joined by dots
+		const isJWTFormat = isString && tokenValue.split( '.' ).length === 3;
+
+		if ( !( isString && isPlainString && isJWTFormat ) ) {
 			/**
 			 * A token value must be in JWT format.
 			 *
--- a/packages/ckeditor-cloud-services-core/src/token/token.js
+++ b/packages/ckeditor-cloud-services-core/src/token/token.js
@@ -58,6 +58,10 @@ class Token {
 		 */
 		this.set( 'value', options.initValue );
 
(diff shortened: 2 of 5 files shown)


packages/ckeditor-cloud-services-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ckeditor/ckeditor-cloud-services-core",
-  "version": "22.0.0",
+  "version": "22.0.1",
```
pomek, line 3: This change should be reverted. Our release scripts bump versions automatically.

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +106,55 @@ class Token {
...
 		return this._refresh()
-			.then( value => this.set( 'value', value ) )
+			.then( value => {
+				this.set( 'value', value );
+
+				if ( this._options.autoRefresh ) {
+					this._registerRefreshTokenTimeout();
+				}
+
+				return this;
```
pomek, line 116:
```suggestion
```

Unnecessary since L118 returns `this`.

packages/ckeditor-cloud-services-core/package.json
jczapiewski-cksource: Reverted

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +106,53 @@ class Token {
...
+
+		this._tokenRefreshTimeout = setTimeout( () => {
+			this.refreshToken();
+		}, tokenRefreshTimeoutTime );
 	}
 
 	/**
-	 * Stops value refreshing.
+	 * Returns token refresh timeout time calculated from expire time in token payload.
+	 * If token parse fails, the default DEFAULT_TOKEN_REFRESH_TIMEOUT_TIME is returned.
```
czerwonkabartosz, line 141: \* Returns token refresh timeout time calculated from expire time in the token payload.

 \* If the token parse fails, the default DEFAULT\_TOKEN\_REFRESH\_TIMEOUT\_TIME is returned.

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +106,53 @@ class Token {
...
 	 * Destroys token instance. Stops refreshing.
 	 */
 	destroy() {
-		this._stopRefreshing();
+		clearTimeout( this._tokenRefreshTimeout );
 	}
 
 	/**
-	 * Starts value refreshing every `refreshInterval` time.
+	 * Registers refresh token timeout for time taken from token.
```
czerwonkabartosz, line 127: Registers a refresh token timeout for the time taken from token.

packages/ckeditor-cloud-services-core/tests/token/token.js
```diff
@@ -84,48 +84,61 @@ describe( 'Token', () => {
 				} );
 		} );
 
-		it( 'should start token refresh every 1 hour', done => {
-			const clock = sinon.useFakeTimers( { toFake: [ 'setInterval' ] } );
+		it( 'should refresh token with time specified in token `exp` payload property', async () => {
```
czerwonkabartosz, line 87: should refresh token with the time specified in token \`exp\` payload property

packages/ckeditor-cloud-services-core/tests/token/token.js
```diff
@@ -202,49 +215,39 @@ describe( 'Token', () => {
 		} );
 	} );
 
-	describe( '_startRefreshing()', () => {
-		it( 'should start refreshing', () => {
-			const clock = sinon.useFakeTimers( { toFake: [ 'setInterval' ] } );
+	describe( '_registerRefreshTokenTimeout()', () => {
```
czerwonkabartosz, line 218: Should we test private methods? :thinking:

packages/ckeditor-cloud-services-core/tests/token/token.js
```diff
@@ -202,49 +215,39 @@ describe( 'Token', () => {
...
-					token._stopRefreshing();
+			const timeoutTime = token._getTokenRefreshTimeoutTime();
 
-					clock.tick( 3600000 );
-					clock.tick( 3600000 );
+			expect( timeoutTime ).to.eq( 1800000 );
+		} );
 
-					expect( requests.length ).to.equal( 3 );
+		it( 'should return default refresh timeout time if token parse fails', () => {
```
czerwonkabartosz, line 245: should return default refresh timeout time if the token parse fails

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +106,53 @@ class Token {
...
 	 *
 	 * @protected
 	 */
-	_startRefreshing() {
-		this._refreshInterval = setInterval( () => this.refreshToken(), this._options.refreshInterval );
+	_registerRefreshTokenTimeout() {
+		const tokenRefreshTimeoutTime = this._getTokenRefreshTimeoutTime();
+
```
czerwonkabartosz, line 133: You could remove timeout before set the new one - clearTimeout.

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +106,53 @@ class Token {
 	 */
 	refreshToken() {
 		return this._refresh()
-			.then( value => this.set( 'value', value ) )
+			.then( value => {
+				this.set( 'value', value );
```
czerwonkabartosz, line 110: Maybe we could throw an error if a value is empty or is the token format is incorrect. 

> A well-formed JWT consists of three concatenated Base64url-encoded strings, separated by dots (`.`):

Also, very often the endpoint of customers returns token in "" (`"token"`), so we could check it too. Of course, the token will be validated by the server but we could throw an error earlier. 

This is the only idea.

czerwonkabartosz (changes requested):

pomek:
One of the tests failed on CI:

```
................
Chrome 85.0.4183.102 (Linux x86_64) Token _getTokenRefreshTimeoutTime should return timeout time based on expiration time in token for valid token FAILED
	AssertionError: expected 1799999 to equal 1800000
	    at Context.eval (webpack:///./packages/ckeditor-cloud-services-core/tests/token/token.js?:260:29)
......................
Chrome 85.0.4183.102 (Linux x86_64): Executed 39 of 39 (1 FAILED) (0.217 secs / 0.12 secs)
```

packages/ckeditor-cloud-services-core/tests/token/token.js
jczapiewski-cksource: Tests for private methods were removed.

packages/ckeditor-cloud-services-core/src/token/token.js
jczapiewski-cksource: I've added token validation.

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +112,72 @@ class Token {
...
+				this._validateTokenValue();
```
czerwonkabartosz, line 117: I would prefer to change this method to `this._validateTokenValue( value );` and call it before `this.set( 'value', value );`

We observe this property and send a request to the server when this value changed, so there is no sense to do it if the token is wrong - https://github.com/cksource/cs/blob/master/common/client/src/websocketgateway/websocketgateway.js#L431.

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +112,72 @@ class Token {
...
 	destroy() {
-		this._stopRefreshing();
+		clearTimeout( this._tokenRefreshTimeout );
+	}
+
+	_validateTokenValue() {
+		if (
+			typeof this.value !== 'string' ||
+			/^".*"$/.test( this.value ) ||
+			this.value.split( '.' ).length !== 3
```
czerwonkabartosz, line 137: Could you add some variables to describe these conditions?

```
// the token must be a string
const isString = this.value === 'string';
// the token must be a plain string without ""
const isPlainString = /^".*"$/.test( this.value )'
// the token must be in correct format - has 3 dots.
const isJWTFormat = this.value.split( '.' ).length === 3;
 
if (!(isString && isPlainString && isJWTFormat) ){ throw }
```

czerwonkabartosz (changes requested):

packages/ckeditor-cloud-services-core/src/token/token.js
jczapiewski-cksource: Totally make sense. In addition, I've found that validating token in the init method makes no sense because the token, in that case, was already validated in the constructor.

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +110,75 @@ class Token {
...
+		const isJWTFormat = isString && tokenValue.split( '.' ).length === 3;
+
+		if ( !( isString && isPlainString && isJWTFormat ) ) {
+			/**
+			 * A token value must be in JWT format.
+			 *
+			 * @error token-not-in-jwt-format
+			 */
+			throw new CKEditorError( 'token-not-in-jwt-format', this );
+		}
```
pomek, line 146:
```suggestion
		// The token must be a string.
		const isString = typeof tokenValue === 'string';

		// The token must be a plain string without quotes ("").
		const isPlainString = !/^".*"$/.test( tokenValue );

		// JWT token contains 3 parts: header, payload, and signature.
		// Each of them is separated by a dot.
		const isJWTFormat = isString && tokenValue.split( '.' ).length === 3;

		if ( !( isPlainString && isJWTFormat ) ) {
			/**
			 * The provided token must follow the [JSON Web Tokens](https://jwt.io/introduction/) standard.
			 *
			 * @error token-not-in-jwt-format
			 */
			throw new CKEditorError( 'token-not-in-jwt-format', this );
		}
```

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +110,78 @@ class Token {
...
 
 	/**
 	 * Destroys token instance. Stops refreshing.
 	 */
 	destroy() {
-		this._stopRefreshing();
+		clearTimeout( this._tokenRefreshTimeout );
+	}
+
+	_validateTokenValue( tokenValue ) {
```
pomek, line 131:
```suggestion
	/**
	 * Checks whether the provided token follows the JSON Web Tokens (JWT) format.
	 *
	 * @param {String} tokenValue The token to validate.
	 */
	_validateTokenValue( tokenValue ) {
```

packages/ckeditor-cloud-services-core/tests/token/token.js
```diff
@@ -202,75 +315,41 @@ describe( 'Token', () => {
...
 		} );
 	} );
 } );
+
+/**
+ * Returns valid token for tests with given expiration time offset.
+ *
+ * @param timeOffset {Number}
+ * @returns {String}
+ */
```
pomek, line 352:
```suggestion
// Returns valid token for tests with given expiration time offset.
//
// @param {Number} [timeOffset=3600000]
// @returns {String}
```

In v5, for private code, we're using inline comments for the documentation.

packages/ckeditor-cloud-services-core/src/token/token.js
```diff
@@ -106,33 +110,84 @@ class Token {
...
+		}, tokenRefreshTimeoutTime );
+	}
+
+	/**
+	 * Returns token refresh timeout time calculated from expire time in the token payload.
+	 *
+	 * If the token parse fails, the default DEFAULT_TOKEN_REFRESH_TIMEOUT_TIME is returned.
+	 *
+	 * @returns {Number}
+	 * @protected
```
pomek, line 179:
```suggestion
	 * @protected
	 * @returns {Number}
```

In v5, we're defining the annotations in order how you would like to describe the function:

- It's a **protected** function,
- that requires **parameters**,
- and **returns** something.

czerwonkabartosz (changes requested):
LGTM :+1:

czerwonkabartosz (approved):

pomek:
I'll merge it when CI finishes.

pomek:
CLA was green before we accepted a suggestion from the review: 

image

Issue #1936, opened 2019-08-01
Writing post-fixers is annoying


jodator:
I don't like current post-fixers API in one regard: they require that the post fixer return true if it changed something. I think that we could detect this on the writer and simplify post-fixers a bit.

Idea is that before running post-fixers the document would clear a writer state ([current implementation](https://github.com/ckeditor/ckeditor5-engine/blob/27d25a1c23037af292ce7bb577e1e0f6a77e5c94/src/view/document.js#L167-L179)). Then using any view modifying method of the writer would cause to set that flag:

```js
_callPostFixers( writer ) {
	do {
		for ( const callback of this._postFixers ) {
			writer.clearViewModified();

			!!callback( writer );

			if ( writer.wasViewModified ) {
				break;
			}
		}
	} while ( writer.wasViewModified );
}
```

Why? A bit cleaner API for others. Just look at this part of tableLayouPostFixer:
```js
let wasFixed = false;
for ( const entry of changes ) {
	if ( table && !analyzedTables.has( table ) ) {
		// Step 1: correct rowspans of table cells if necessary.
		// The wasFixed flag should be true if any of tables in batch was fixed - might be more then one.
		wasFixed = fixTableCellsRowspan( table, writer ) || wasFixed;
		// Step 2: fix table rows sizes.
		wasFixed = fixTableRowsSizes( table, writer ) || wasFixed;

		analyzedTables.add( table );
	}
}

return wasFixed;
```
managing `wasFixed` flag is a mess and PITA. Also, it is easy to [forget about it](https://github.com/ckeditor/ckeditor5-link/issues/241).

The modifing view methods are for instance: `writer.remove()`, `writer.setAttribute()`. The other creator methods are not - e.g. `writer.createText()`, `writer.createPosition()`.

ps.: I've checked this with 15 min POC in which I've added such flag and post-fixer handling and only two tests failed (one wasn't passing a writer a second one is probably similar case).

cc @pjasiun / @scofalik / @Reinmar WDYT

ps.: probably nice-to-have milestone :)

pjasiun:
Sounds good 👍

My only doubt is that I am not sure if we will be able to catch all situation when the postfixer changes something. For instance, one may change something inside the UI element. However, we could introduce a special method (`markAsChanged` or something) which can be used in such cases, instead of `return true`.

Also, I think that many developers do not remember that they need to return true if the postfixer change anything, so some automatic mechanism might be better then nothing in such cases.

[...]

PR #8099, opened 2020-09-17
7956: Differ.refreshItem() + triggerBy

### LINKED ISSUE #7956: Implement Option 1: A `triggerBy` option that issues element re-render

Implement the base API for the [slot conversion API](https://github.com/ckeditor/ckeditor5/issues/6510#issuecomment-673417728). This should provide a minimal, top-level API for changes defined there and researched in: #7729.

### Goals:

*   top-level API as described, anything below should support it.
*   Use memoization technique to reduce re-rendering big chunks of the same views.
*   The renderer is out of scope of this task.

jodator:

Feature (engine): Introduce automatic model-to-view reconversion by defining `triggerBy` option for `elementToElement()` conversion helper. Closes #7956.

Other (table): Table cell contents refreshing for the editing view will now make less view updates.

---

### Additional information

This PR:

*   introduces the `triggerBy` configuration option for downcast `elementToElement()` conversion helper config.
*   Handles element re-conversion and view momoization on `DowncastDispatcher` level.
    *   The model element children that have exiting view are not reconverted (memoization).
    *   Newly inserted children are converted by corresponding converters.
    *   Converter for reconverted element can converts also its children (especially in case of layered view)
*   Has fixes for table package `table-cell-refresh-post-fixer` to only re-insert paragraphs not entire table cells.

---

~This PR:~

1.  ~Introduces `'refresh'` type to differ changes (internally represented as `'x'` in differ processing algorithms).~  
    ~This is needed to properly handle refreshing items by the dispatcher. Assumptions:~
    1.  ~The refreshing newly inserted element should be treated as `'insert'`~
    2.  ~Attribute changes on refreshed item must be preserved - allows to have atomic converters for attributes working alongside refreshing.~
2.  ~Introduces automatic refreshing of an item in `elementToElement` downcast by a new configuration option `triggerBy`.~  
    ~It defines which inner events should trigger refreshing. This has limited support for now (follow-up). You can trigger reconversion of the element on attribute change or by insert/remove of a child. This process does not cover three-way-binding of a layerd view elements (follow-up, case of `image -> figure>img` conversion).~
3.  ~Introduces `DowncastDispatcher.convertRefresh()` that handles refreshing elements with~ _~memoization~_~.~  
    ~The memoization process is done by re-using existing child view elements. For instance refreshing `<paragraph>` will trigger `'insert'` + `'attribute'` conversion events for the main element and will use existing view elements for child text nodes and elements. It will also trigger proper events for newly added child elements.~
4.  ~Renames `Differ.refreshItem()` to `Differ.reinsertItem()` and will deprecate it (follow-up).~  
    ~The name is needed for the new `'refresh'` change handling. The `reinsertItem()` is still needed as `triggerBy` doesn't handle table case (missing three-way-binding).~
5.  ~Re-introduces `Differ.refreshItem()` that now uses `'refresh'` change instead of `'remove'` + `'insert'` combo (reserved for `Differ.reinsertItem()`.~
6.  ~The table's `table-cell-refresh-post-fixer` logic was changed to refresh only `<paragraph>` elements inside a table cell.~  
    ~It uses new `Differ.refreshItem()` in order to make smaller changes in the view. The logic behind was also changed as I found a different approach to it.~
7.  ~For this matter, I've changed how `<paragraph>` inside `<tableCell>` is converted.~  
    ~There's new converter for a `<paragraph>` with a `'high'` priority that overrides default conversion for table case (so it can be rendered as `<span>` in the `editing` view. This converter is run on `refreshItem()` as well.~

~Internals of the differ has been~ _~slightly~_ ~refactored so parts of the logic of inserting an item could be re-used in refreshing mechanism. However, I did not review all the changes there and I feel that some might be removed. For instance, I've added `refresh x insert|remove|attribute` changes handling in the `Differ._handleChange()` later than some of the earlier changes.~

---

TODO before merge:

*   [x] review the code with @niegowski
    *   [x] extract paragraph in table cell converter to `downcast.js`
    *   [x] check todos
    *   [x] fix `artilce.js` manual test
    *   [x] create re-conversion manual test (current `article.js`)
    *   [x] check if `mapper.bindSlots()`is needed
    *   [x] refreshing table cell paragraph might cause infinite loops (https://github.com/ckeditor/ckeditor5/issues/1936#issuecomment-698297417)
*   [x] review differ / conversion changes with @scofalik after the initial review.
*   [x] check if the memoization is done properly (it looks like I have this on two levels)


jodator:
@niegowski  - you can check the code but see TODO section if anything is fixed.

I'd assume that a review will not be started ASAP so some of those will be fixed before you would jump to the review.

jodator:
@niegowski  - I think the code is ready to be reviewed. What I need to check:

*   whether the changes/comments are clear enough
*   I suspect that I've made some logic error with how memoization is done - I think that there is two-level memoization. For instance, I didn't manage to remove `mapper.bindSlots()`.
*   changes in tables - see what I did there and validate new approach :) The main thing is that we refresh `<paragraph>` only, not the entire `<tableCell>`  - only the refreshed `<paragraph>` should be re-rendered.
*   anything else :D

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -25,79 +25,89 @@ export default function injectTableCellRefreshPostFixer( model ) {
 function tableCellRefreshPostFixer( model ) {
 	const differ = model.document.differ;
 
-	// Stores cells to be refreshed so the table cell will be refreshed once for multiple changes.
-	const cellsToRefresh = new Set();
-
-	// Counting the paragraph inserts to verify if it increased to more than one paragraph in the current differ.
-	let insertCount = 0;
+	const changesForCells = new Map();
+	const changes = [ ...differ.getChanges() ];
```
niegowski, line 29: The `getChanges()` returns a fresh copy of array so there is no need for copying it again.

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -25,79 +25,89 @@ export default function injectTableCellRefreshPostFixer( model ) {
...
+
+	// 2. For each table cell:
+	for ( const [ tableCell, changes ] of changesForCells.entries() ) {
+		// 2a. Count inserts/removes as diff and marks any attribute change.
+		const { childDiff, attribute } = getChangesSummary( changes );
+
+		// 2b. If we detect that number of children has changed...
+		if ( childDiff !== 0 ) {
+			const prevChildren = tableCell.childCount - childDiff;
+			const currentChildren = tableCell.childCount;
```
niegowski, line 58:
```suggestion
			const currentChildren = tableCell.childCount;
			const prevChildren = currentChildren - childDiff;
```

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -25,79 +25,89 @@ export default function injectTableCellRefreshPostFixer( model ) {
...
+
+			// Might need refresh if previous children was different from 1. Eg.: it was 2 before, now is 1.
+			if ( currentChildren === 1 && prevChildren !== 1 ) {
+				cellsToRefresh.add( tableCell );
+			}
+
+			// Might need refresh if previous children was 1. Eg.: it was 1 before, now is 5.
+			if ( currentChildren !== 1 && prevChildren === 1 ) {
+				cellsToRefresh.add( tableCell );
+			}
```
niegowski, line 68:
```suggestion
			// Might need refresh if previous children was different from 1. Eg.: it was 2 before, now is 1 (or the opposite).
			if ( currentChildren === 1 || prevChildren === 1 ) {
				cellsToRefresh.add( tableCell );
			}
```

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -25,79 +25,89 @@ export default function injectTableCellRefreshPostFixer( model ) {
...
-		// @if CK_DEBUG_TABLE // console.log( `Post-fixing table: refreshing cells (${ cellsToRefresh.size }).` );
+		// @if CK_DEBUG_TABLE // console.log( `Post-fixing table: Checking table cell to refresh (${ cellsToRefresh.size }).` );
+		// @if CK_DEBUG_TABLE // let paragraphsRefreshed = 0;
 
 		for ( const tableCell of cellsToRefresh.values() ) {
-			differ.refreshItem( tableCell );
+			for ( const paragraph of [ ...tableCell.getChildren() ].filter( child => child.is( 'element', 'paragraph' ) ) ) {
+				// @if CK_DEBUG_TABLE // console.log( `Post-fixing table: refreshing paragraph in table cell (${++paragraphsRefreshed}).` );
+				differ.refreshItem( paragraph );
+			}
```
niegowski, line 86: Wouldn't it be enough to refresh only first child? If it's nested in some block quote then it's already <p> element, if there is more than one paragraph then the following ones are already down-casted as <p> elements.

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -25,79 +25,89 @@ export default function injectTableCellRefreshPostFixer( model ) {
...
 		}
-
-		return true;
 	}
 
+	// Always return false to prevent the refresh post-fixer from re-running on the same set of changes and going into an infinite loop.
+	// See https://github.com/ckeditor/ckeditor5/issues/1936.
```
niegowski, line 91: Maybe we should check if the first paragraph is not marked for refresh instead of returning false?

niegowski (commented):
Partial review - test not covered yet.

packages/ckeditor5-engine/tests/manual/slotconversion.js
```diff
@@ -0,0 +1,292 @@
...
+	editor.conversion.for( 'downcast' ).add( dispatcher => dispatcher.on( 'insert', ( event, data, conversionApi ) => {
+		const view = conversionApi.mapper.toViewElement( data.item );
+
+		if ( view ) {
+			const insertCount = nextInsert();
+
+			conversionApi.writer.setAttribute( 'data-insert-count', `${ insertCount }`, view );
+			conversionApi.writer.setAttribute( 'title', `Insertion counter: ${ insertCount }`, view );
+		}
+	}, { priority: 'lowest' } ) );
```
niegowski, line 247: This is not triggered for slots on any edit of the box, those attributes are present only on the first render, and they disappear after any update.

packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
```diff
@@ -113,6 +121,825 @@ describe( 'DowncastHelpers', () => {
...
+								'<div class="slot"><p>bar</p></div>' +
+								'<div class="slot"><p>baz</p></div>' +
+							'</div>' +
+						'</div>'
+					);
+				} );
+			} );
+
+			function getViewAttributes( modelElement ) {
+				// TODO decide whether below is readable:
```
niegowski, line 932: :+1:  for me

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -1136,13 +1237,18 @@ function _generateActionsFromChanges( oldChildrenLength, changes ) {
...
-		} else {
+		} else if ( change.type == 'attribute' ) {
 			actions.push( ...'a'.repeat( change.howMany ).split( '' ) );
 
 			// The last handled offset is at the position after the changed range.
 			offset = change.offset + change.howMany;
 			// We changed `howMany` old nodes, update `oldChildrenHandled`.
 			oldChildrenHandled += change.howMany;
+		} else {
+			actions.push( 'x' );
```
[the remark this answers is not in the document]
jodator: The "refresh" for now is done only for one item at a time.

Maybe I'll add some tests for this to check if this behaves correctly.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +533,115 @@ export default class DowncastDispatcher {
...
+	_getElementsForAutomaticRefresh( differ ) {
+		const found = differ.getChanges()
+			.filter( ( { type } ) => type === 'attribute' || type === 'insert' || type === 'remove' )
+			.map( entry => {
+				const element = getElementFromChange( entry );
+
+				let eventName;
+
+				if ( entry.type === 'attribute' ) {
+					eventName = `attribute:${ entry.attributeKey }:${ element && element.name }`;
```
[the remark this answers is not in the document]
jodator: Yep, those are text attributes changes.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -1349,6 +1364,10 @@ function downcastElementToElement( config ) {
 
 	return dispatcher => {
 		dispatcher.on( 'insert:' + config.model, insertElement( config.view ), { priority: config.converterPriority || 'normal' } );
+
+		if ( Array.isArray( config.triggerBy ) ) {
```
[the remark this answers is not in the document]
jodator: I've added the annotation here. But to decrease doc maintenance I've also removed duplicated descriptions.

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
jodator: I think that this check is too simplistic - a change that adds and removes children would have `currentChildren === 1` and `prevChildren === 1`.

I think that the tests must be updated to check for this case and then let's see.

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
jodator: I think that the tests must be updated to check for this case and then let's see. (as above)

The test should check for the memoized `<p>` and see if it was changed to a new instance (effect of refreshing).

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
jodator: Maybe this will be OK, thanks! I've got lost here TBH.

packages/ckeditor5-engine/tests/manual/slotconversion.js
jodator: I think that this might be a limitation of this approach - I'll double-check this. AFAIR setting this attribute by the writer methods behaved a bit consistently.

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
jodator: Yep, debugging lef-tovleftoverer.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +533,115 @@ export default class DowncastDispatcher {
...
+	}
+
+	_getChangesAfterAutomaticRefreshing( differ ) {
+		const elementsToRefresh = this._getElementsForAutomaticRefresh( differ );
+
+		for ( const element of elementsToRefresh.values() ) {
+			differ.refreshItem( element );
+		}
+
+		return differ.getChanges().filter( entry => !elementsToRefresh.has( getElementFromChange( entry ) ) );
```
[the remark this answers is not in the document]
jodator: The `differ.refreshItem()` invalidates differ cache. This is because `refresh` can remove other changes.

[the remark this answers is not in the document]
jodator: I've updated the code to have that clear.

packages/ckeditor5-engine/src/conversion/mapper.js
```diff
@@ -336,6 +345,31 @@ export default class Mapper {
...
+	 * @param {module:engine/model/element~Element} modelElement Model element.
+	 * @param {module:engine/view/element~Element} viewElement View element.
+	 */
+	bindSlotElements( modelElement, viewElement ) {
+		const existingView = this.toViewElement( modelElement );
+
+		this._slotToViewMapping.set( modelElement, existingView );
+
+		this.bindElements( modelElement, viewElement );
+	}
```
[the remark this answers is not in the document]
niegowski: :+1:

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
niegowski: but this scenario is excluded above by 

```
if ( childDiff !== 0 ) {
```

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
niegowski: :+1:

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
jodator: I'm talking about:  
Before: `<p>`  
Changes:  `[ add p, remove p ]`  
After: `<p>`  
  
But let's test it to be sure.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +540,156 @@ export default class DowncastDispatcher {
...
+		}
+
+		for ( const element of elementsToRefresh.values() ) {
+			differ.refreshItem( element );
+		}
+
+		// The `differ.refreshItem()` invalidates differ cache - we can't re-use previous changes.
+		const changesAfterRefresh = differ.getChanges();
+
+		return changesAfterRefresh.filter( entry => !elementsToRefresh.has( getElementFromChange( entry ) ) );
```
niegowski, line 589: Maybe a comment about why we filter-out some changes would be good.

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
jodator: OMG, I'm not smart with that... there will be no change in the differ itself.

packages/ckeditor5-engine/src/model/differ.js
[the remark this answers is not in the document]
jodator: There is a test for this already - I've added explanation in the code.

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
jodator: > Wouldn't it be enough to refresh only first child?

Tricky - someone might prepend table cell ;) Added a test for this case.

packages/ckeditor5-table/src/converters/table-cell-refresh-post-fixer.js
jodator: Still, there's something odd :/ Look at the current implementation - it hangs other tests.

I'm leaving this as-is for now as I don't see investing more time into this at this stage makes sense.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
jodator: There's an explanation in the method jsdoc. If you think it is not clear I can add it here.

jodator: There's an explanation in the method jsdoc. If you think it is not clear I can add it here.

jodator:
@niegowski Since I've changed some things we could have another look on this.

jodator:
@niegowski Since I've changed some things we could have another look on this.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +540,156 @@ export default class DowncastDispatcher {
...
+	 * @private
+	 * @param {Object} data Event data.
+	 */
+	_reconvertElement( data ) {
+		// Cache current view element of a converted element, might be undefined if first insert.
+		const currentView = this.conversionApi.mapper.toViewElement( data.item );
+
+		// Remove the old view (but hold the reference as it will be used to bring back view items not needed to re-render.
+		// Thanks to the mapper that holds references nothing should blow up.
+		this.conversionApi.writer.remove( currentView );
```
niegowski, line 647: `currentView` is used only here, later in code we use an element from mapper, so this code comment seems invalid.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -636,6 +837,40 @@ function shouldMarkerChangeBeConverted( modelPosition, marker, mapper ) {
...
+	return `${ type }:${ name }`;
+}
+
+function getElementFromChange( entry ) {
+	const { range, position, type } = entry;
+
+	return type === 'attribute' ? range.start.nodeAfter : position.parent;
+}
+
+function rangeIteratorValueToEventData( value ) {
```
niegowski, line 852: This function name is a bit confusing, maybe `walkerValueToEventData`, or `treeWalkerValueToEventData` would be better?

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -636,6 +837,40 @@ function shouldMarkerChangeBeConverted( modelPosition, marker, mapper ) {
...
+	const item = value.item;
+	const itemRange = Range._createFromPositionAndShift( value.previousPosition, value.length );
+
+	return {
+		item,
+		range: itemRange
+	};
+}
+
+function elementWasMemoized( data, mapper ) {
```
niegowski, line 862: Maybe we could name it `elementHasViewMapping` ?

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -982,6 +1045,16 @@ export default class Differ {
 		return diffs;
 	}
 
+	_getRefreshDiff( parent, offset, name ) {
```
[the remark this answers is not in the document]
niegowski: Still missing docs 😉

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +540,156 @@ export default class DowncastDispatcher {
...
+						// Refreshing is done only on elements so skip text attribute changes.
+						return;
+					}
+
+					eventName = `attribute:${ entry.attributeKey }:${ element.name }`;
+				} else {
+					eventName = `${ entry.type }:${ entry.name }`;
+				}
+
+				if ( this._refreshTriggerEventToElementNameMapping.has( eventName ) ) {
```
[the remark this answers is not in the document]
jodator: Another good catch - I had to refactor this as the element name _is_ important (or can be). This must be checked.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
jodator: Differ does not know which elements are related by a `triggerBy` - I think that this is the responsibility of the downcast dispatcher. It can read those mappings and act upon them to alter the conversion process. Mappings are defined by `elementToElement()` conversion helper. So

1.  The `elementToElement()` defines which events should trigger reconversion.
2.  The `DowncastDispatcher` handles the conversion process (which events to fire, which change).
3.  The `Differ` shows what changed, so if an element was inserted, removed, have changed attributes, or if it was marked to be refresh.

The `triggerBy` is a declarative mechanism of reconverting the entire element

But you can use `Differ.refreshItem()` implicitly to mark an element to be re-converted (refreshed), ie in a post-fixer.

niegowski: :+1:

packages/ckeditor5-engine/tests/manual/slotconversion.js
jodator: Fixed.

jodator:
@scofalik I think that you can look at it now. There are some follow-ups or simplifications that didn't make into this ticket. See the main comment of the PR of implemented things and assumptions.

@ckeditor/qa-team I'd like to ask you to test this change. Take a focus on:

1.  A paragraph refreshing in a table cell (the change from `<p>` \<-> `<span>` in the editing view). So single paragraph without attributes (AFAIR only alignment can be set on a `<paragraph>`) should be `<span>` and `<p>` otherwise. So it might impact every aspect of table integrations.
2.  The new manual test slot conversion in the engine. There's one bug (see below). TL;DR: if you change the attribute of a box or add/remove slot the rendered items id of slot children should be unchanged. Editing should work - I've enabled everything inside the slot, but you can focus on content that could be set inside table cell.

Follow-ups:

*   Decide whether refresh post-fixer is OK in the model #8200 
*   Support triggerBy of deeply nested elements #8201 

Bugs:

*   Text node with an attribute is duplicated in slot conversion demo #8203

FilipTokarski:
## Steps:
1. Add table
2. In table cell:  
    - type something
    - press `shift + enter`
    - type something
    - press `enter`  
  
**Expected:** no errors, there are two `<p>` with text in the table cell  
  
**Actual:** editor crashes:  
```js
mapper.js:520 Uncaught TypeError: Cannot read property 'name' of undefined
    at Mapper.getModelLength (mapper.js:520)
    at Mapper.findPositionIn (mapper.js:584)
    at Mapper.on.priority (mapper.js:115)
    at Mapper.fire (emittermixin.js:209)
    at Mapper.toViewPosition (mapper.js:343)
    at elementHasViewMapping (downcastdispatcher.js:876)
    at DowncastDispatcher.convertRefresh (downcastdispatcher.js:317)
    at DowncastDispatcher.convertChanges (downcastdispatcher.js:156)
    at editingcontroller.js:94
    at View.change (view.js:477)
```
  
0_slot1

FilipTokarski:
Scenario with very similar ☝️  stack trace:
1. Add table
2. Apply `bold`
3. Write something
4. Put cursor in the middle of the text
5. Press `enter`
  
```js
Uncaught TypeError: Cannot read property 'name' of undefined
    at Mapper.getModelLength (mapper.js:520)
    at Mapper.findPositionIn (mapper.js:584)
    at Mapper.on.priority (mapper.js:115)
    at Mapper.fire (emittermixin.js:209)
    at Mapper.toViewPosition (mapper.js:343)
    at DowncastDispatcher.<anonymous> (downcasthelpers.js:553)
    at DowncastDispatcher.fire (emittermixin.js:209)
    at DowncastDispatcher.convertRemove (downcastdispatcher.js:233)
    at DowncastDispatcher.convertChanges (downcastdispatcher.js:154)
    at editingcontroller.js:94
```
  
0slot2

FilipTokarski:
### Steps:
1. In table cell, type something ( without any styles )
2. Apply bold
3. Type some more
4. Press `enter`
  
**Result:** bolded text disappears  
  
0slot4

FilipTokarski:
Ok, so far I can't find anythng new apart from what's reported above ☝️

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +553,155 @@ export default class DowncastDispatcher {
...
+	 * will not be converted again. Their existing view elements will be used instead.
+	 *
+	 * @private
+	 * @param {Object} data Event data.
+	 */
+	_reconvertElement( data ) {
+		const currentView = this.conversionApi.mapper.toViewElement( data.item );
+
+		// Remove the old view but do not remove mapper mappings - those will be used to revive existing elements.
+		this.conversionApi.writer.remove( currentView );
```
scofalik, line 659: This is an optimistic assumption ;). This will work for most cases but may fail for more complex ones if model element is mapped to inner view element (something like in list items and \<ul>/\<ol> + \<li> I wonder if it works there correctly if I refresh a singular list item and also change its attribute :thinking: ).

I understand that the problem with "old refreshing" was that remove conversion removed mappings?

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +553,155 @@ export default class DowncastDispatcher {
...
+		// Iterate over new view elements to find "interesting" points - those elements that are mapped to the model.
+		for ( const { item } of this.conversionApi.writer.createRangeIn( viewElement ) ) {
+			const modelItem = this.conversionApi.mapper.toModelElement( item );
+
+			// At this stage we get the update view element, so any mapped model item might be a potential "slot".
+			if ( modelItem ) {
+				const currentViewItem = this.conversionApi.mapper.getExistingViewForSlot( modelItem );
+
+				// This of course needs better API, but for now it works.
+				// Mapper.bindSlot() creates mappings as mapper.bindElements() but also binds view element
```
scofalik, line 675:
```suggestion
				// Mapper.bindSlotElements() creates mappings as mapper.bindElements() but also binds view element
```

scofalik (changes requested):
I added some general comments and wrote more about Differ.

It was difficult for me to understand some parts of this feature, for example slots. I think I finally got it but I wonder if there were other PRs connected with this subject that I haven't seen and maybe because of that I didn't understand some of the code from here (for example, in reconversion, I wondered why we don't simply reconvert the refreshed items and then move all the children of the old view element into the new refreshed view element, but I think I got it and I got why slots are important here - as I understand now, slots are separating view that is generated by model element conversion from the view that is generated by model element children conversion).

From architectural POV I dislike one thing - that differ is affected from conversion layer. In the ideal world, this should be separated and conversion should only read from differ. I think that Differ#refreshItem() is fine and is in a correct place. Then I started wondering about `triggerBy`. And I am not sure if this is in the best place.

In short, `triggerBy` means "Whenever X in model changes, call refreshItem() on Y". I guess it is convenient to specify it when defining a converter but is it really connected with conversion? It is more like a post-fixer, or model mechanism, or a property of given model element (model elements structure). I wonder if it all couldn't be implemented straight in Differ. Maybe it could simplify some code and we wouldn't have this flow of: differ generating changes, dispatcher reading those changes, dispatcher affecting differ, differ generating new set of changes, dispatcher reading those changes again and then also filtering something (`_getChangesAfterAutomaticRefreshing()`).

The code can stay as it is, it will work, AFAICS. However, if I was to suggest some "bigger picture" improvement, I'd suggest that.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -71,6 +86,8 @@ export default class DowncastHelpers extends ConversionHelpers {
 	 * @param {module:engine/view/elementdefinition~ElementDefinition|Function} config.view A view element definition or a function
 	 * that takes the model element and {@link module:engine/conversion/downcastdispatcher~DowncastConversionApi downcast conversion API}
 	 * as parameters and returns a view container element.
+	 * @param {Array.<String>} [config.triggerBy] Events which will trigger element reconversion. Reconversion can be triggered by
```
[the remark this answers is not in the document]
jodator: Makes sense :)

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -1010,6 +1100,34 @@ export default class Differ {
...
 	}
 
+	/**
+	 * Checks whether given element or any of its parents is an element that is buffered as a refreshed element.
+	 *
+	 * @private
+	 * @param {module:engine/model/element~Element} element Element to check.
+	 * @returns {Boolean}
+	 */
+	_isInRefreshedElement( element ) {
```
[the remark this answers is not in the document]
jodator: I'm leaning towards moving this change to `Dispatcher` rather than modifying differ.

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -880,6 +931,26 @@ export default class Differ {
 					}
 				}
 			}
+
+			if ( inc.type === 'refresh' ) {
```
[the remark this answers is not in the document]
jodator: > However, I am missing transformations for X vs "refresh". For example, if there is insert before refresh, refresh offset needs to be moved, etc. See other transformations, for example, insert vs insert.

That's probably because the `_isInInsertedElement()` that was added before most of the `refresh vs X` updates.

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -585,6 +622,20 @@ export default class Differ {
...
 
+	/**
+	 * Saves and handles a refresh change.
+	 *
+	 * @private
+	 * @param {module:engine/model/element~Element} parent
+	 * @param {Number} offset
+	 * @param {Number} howMany
+	 */
+	_markRefresh( parent, offset, howMany ) {
```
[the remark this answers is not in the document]
jodator: Yep, I don't see why we would need to refresh text. It is also unsupported by the rest of the code. It assumes only elements are refreshed.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
jodator: Yes. Those are mappings that were removed. But I think that if would handle memoization on the upper level (ie by taking a snapshot before the whole conversion) or just by moving the wholel logic to dispatcher alone it would be better and we would need "refresh" change at all.the

jodator:
> It was difficult for me to understand some parts of this feature, for example slots. I think I finally got it but I wonder if there were other PRs connected with this subject that I haven't seen and maybe because of that I didn't understand some of the code from here (for example, in reconversion, I wondered why we don't simply reconvert the refreshed items and then move all the children of the old view element into the new refreshed view element, but I think I got it and I got why slots are important here - as I understand now, slots are separating view that is generated by model element conversion from the view that is generated by model element children conversion).

In short yes. There [are three cases](https://github.com/ckeditor/ckeditor5/issues/7956#issuecomment-694241558) here (excluding element without children):

1.  A model element without "slot" children with  "1:1" M-V mapping.
2.  A model element without "slot" children with "1:n" M-V mapping (ie `image - figure>img`)
3.  A model element with n "slot" children with "1:1" or "1:n" mapping in the view. Slots are rendered by the main converter.

The second thing you're also right. The bug reported by Filip is one example of this. I need to re-implement this part to cover all cases above. Right now it only works with "slots" and "1:1" mapping for some reason.

> From architectural POV I dislike one thing - that differ is affected from conversion layer. In the ideal world, this should be separated and conversion should only read from differ. I think that Differ#refreshItem() is fine and is in a correct place. Then I started wondering about `triggerBy`. And I am not sure if this is in the best place.

I think I could remove `differ.refreshItem()` and handle all of this inside converter. But to do that we would probably need to gather all the events beforehand (right now those events are generated on-the-fly during conversion).

> In short, `triggerBy` means "Whenever X in model changes, call refreshItem() on Y". I guess it is convenient to specify it when defining a converter but is it really connected with conversion? It is more like a post-fixer, or model mechanism, or a property of given model element (model elements structure). I wonder if it all couldn't be implemented straight in Differ. Maybe it could simplify some code and we wouldn't have this flow of: differ generating changes, dispatcher reading those changes, dispatcher affecting differ, differ generating new set of changes, dispatcher reading those changes again and then also filtering something (`_getChangesAfterAutomaticRefreshing()`).

I think that I rather see this in conversion (downcast dispatcher) and make the `Differ` simpler (as it was before). Because:

*   `triggerBy` defines under what circumstances the view must be re-converted.
*   Some post-fixers should mark an element to be re-converted (so the model isn't changed).

In order to do that I think that we _might_ need to have all the dispatcher events beforehand. This _could_ make the logic of the re-converting be in one place and make things easier.

> The code can stay as it is, it will work, AFAICS. However, if I was to suggest some "bigger picture" improvement, I'd suggest that.

I could see this as either:

1.  Marking new things "protected" and refactor later.
2.  Refactor now to make the dispatcher handle refreshing.

scofalik:
I was as well thinking of maybe moving everything to dispatcher but it was easier for me to imagine having everything in the differ.

packages/ckeditor5-engine/src/model/differ.js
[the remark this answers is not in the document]
scofalik: Mmm... I don't understand your answer :D. Do you agree with me or do you say that what I mentioned is not needed ("refresh" being `old` change).

jodator:
> I was as well thinking of maybe moving everything to dispatcher but it was easier for me to imagine having everything in the differ.

I'd go with the dispatcher (even if harder) to have it in one place. It just feels like a better place for this.

packages/ckeditor5-engine/src/model/differ.js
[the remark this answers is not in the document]
jodator: I think that `X vs refresh` (I wrote the opposite) is handled by the `_isInInsertedElement` checks. It looks like handling `X vs Y` is done in two steps:

1.  when handing given change in `bufferOperation()`
2.  then afterward in `_handleChange()`

This migth be a reason why `X vs refresh` are not needed in `_handleChange()`.

OTOH, yes adding a test for `insert` before `refresh` (and similar) might be needed.

TL;DR - I agree with you (update the tests if needed)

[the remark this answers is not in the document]
scofalik: Writing from the top of my head but I think I remember how `_isInInsertedElement` works ;).

So, I think you are not right. Following scenario is probably not handled:

1.  Assume model: `<paragraph>Foo</paragraph><paragraph>Bar</paragraph>`.
2.  In one change block:
    1.  Refresh paragraph "Bar".
    2.  Insert paragraph at the beginning.

I think that in this scenario, the refresh change will not get updated, so the differ will return:

*   insert paragraph on index `0`,
*   refresh paragraph on index `1`.

So, in the end, a wrong paragraph will be refreshed.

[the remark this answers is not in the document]
jodator: Thanks for the clarification :) ATM I'm still diving into the `Differ` internals. A test for such scenario(s) should be added.

Additionally, I'm thinking now about moving this to `DowncastDispatcher` anyway. But I don't want to rush into this without re-thinking this.

scofalik:
> In order to do that I think that we _might_ need to have all the dispatcher events beforehand. This _could_ make the logic of the re-converting be in one place and make things easier.

Not sure :thinking:. It seems to me that we don't need them beforehand. We kind of have those events anyway - as differ changes. You do a similar thing now when you get current changes and then get changes after the automatic refresh.

But...

1.  In the dispatcher, we handle changes in the order of them being in the content (so first paragraphs are handled first, and also children are handled after parents).
2.  Reconversion is triggered by attributes and children changes.
3.  We can remember if a given element was inserted.

Having this in mind:

1.  When converting attribute change we can check if a given element has "triggerBy" parameter registered.
2.  When converting children change, before it, we can check if there is an ancestor that has "triggerBy" parameter registered. _Maybe_ we would have to check all children beforehand, I could imagine a situation where maybe a child conversion depends on how its parent looks like in the view but the child itself does not trigger reconversion but other child does. Kind of complicated.

> I think that I rather see this in conversion (downcast dispatcher) and make the `Differ` simpler (as it was before). Because:
> 
> *   `triggerBy` defines under what circumstances the view must be re-converted.

Food for thought: let's imagine that we have two converters for the given model element. One of those converters have `triggerBy` but lower priority. AFAIR, we register `triggerBy` for the element, not for the converter. So, the overwritten converter will still trigger reconversion. I know it is an edge case but that's something that came to my mind.

scofalik:
To sum up:

1.  I commented on some incorrectness in Differ. If you will stick with the current solution, you will need to fix them and then I can re-review again.
2.  I think that the solution can be moved in whole to Differ or to the dispatcher. But I am also fine with having it as it is right now. This was a general comment, not a change request. However, if you also feel the same, let's change it a little.
3.  It seems to me that refreshing inside dispatcher does not have to be complicated but I don't know the whole problem as much as you do.

For now, I think that I am done with this review. Ping me if you decide to go with one or the other route and you would like me to re-review it.

jodator:
> For now, I think that I am done with this review. Ping me if you decide to go with one or the other route and you would like me to re-review it.

Thanks for the input! I'll let you know with the direction nad fixed PR.

jodator:
@scofalik - sorry for bugging you ATM. I've just checked how we could re-implement `triggerBy` to not use `differ.refreshItem()` and it turned out to be pretty easy thanks to your suggestions concerning differ -> dispatcher -> differ -> dispatcher loop there (0f59a7a).

Although not ideal it _looks_ that reverting changes in the differ might be actually quite easy and reduce the size of this PR. :crossed\_fingers: I'll find a place for `refreshItem()` or `reconvertItem()` API since I'd love to get rid of this logic from the differ.

packages/ckeditor5-engine/src/model/differ.js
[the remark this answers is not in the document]
jodator: Done on `DowncastDispatcher` level only.

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -453,6 +484,12 @@ export default class Differ {
 					// there is a single diff for each of them) and insert them into the diff set.
 					diffSet.push( ...this._getAttributesDiff( range, snapshotAttributes, elementAttributes ) );
 
+					i++;
+					j++;
+				} else if ( action === 'x' ) {
+					// Swap action - similar to 'equal'.
```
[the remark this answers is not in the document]
jodator: Reverted.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +553,155 @@ export default class DowncastDispatcher {
...
+	_getElementsForAutomaticRefresh( changes ) {
+		const found = changes
+			.filter( ( { type } ) => type === 'attribute' || type === 'insert' || type === 'remove' )
+			.map( entry => {
+				const element = getElementFromChange( entry );
+
+				let eventName;
+
+				if ( entry.type === 'attribute' ) {
+					if ( !element ) {
```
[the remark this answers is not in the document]
jodator: Reverted.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +553,155 @@ export default class DowncastDispatcher {
...
+
+		// Bring back removed child views on refreshing the parent view.
+		const viewElement = this.conversionApi.mapper.toViewElement( data.item );
+
+		// Iterate over new view elements to find "interesting" points - those elements that are mapped to the model.
+		for ( const { item } of this.conversionApi.writer.createRangeIn( viewElement ) ) {
+			const modelItem = this.conversionApi.mapper.toModelElement( item );
+
+			// At this stage we get the update view element, so any mapped model item might be a potential "slot".
+			if ( modelItem ) {
```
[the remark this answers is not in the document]
jodator: Reverted.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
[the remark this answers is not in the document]
jodator: I've implemented this as this makes the feature much easier to define.

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -136,6 +138,33 @@ export default class Differ {
 		this._cachedChanges = null;
 	}
 
+	/**
+	 * Marks given `item` in differ to be "refreshed".
+	 *
+	 * It means that the item will be marked as a "refreshed" inserted in the differ changes. It will be then re-converted
+	 * when differ changes will be handled by a dispatcher.
```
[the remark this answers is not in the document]
jodator: reverted

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -256,6 +271,57 @@ export default class DowncastDispatcher {
 		this._clearConversionApi();
 	}
 
+	/**
+	 * Starts a refresh conversion - depending on a configuration it would:
+	 *
+	 * * Fire a {@link #event:insert `insert` event} for the element to refresh.
+	 * * Handle conversion of a range insert for nodes under the refreshed item which are not bound as slots.
```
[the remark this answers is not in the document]
jodator: Fixed a bit after refactor.

jodator:
@scofalik I've refactored this PR. I think that it is much straightforward as it does not use differ at all to reconvert an element.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -636,6 +804,38 @@ function shouldMarkerChangeBeConverted( modelPosition, marker, mapper ) {
...
+		item,
+		range: itemRange
+	};
+}
+
+function elementOrTextProxyToView( item, mapper ) {
+	if ( item.is( 'textProxy' ) ) {
+		const mappedPosition = mapper.toViewPosition( Position._createAt( item.parent, item.startOffset ) );
+
+		return mappedPosition.parent.is( '$text' ) ? mappedPosition.parent : null;
```
scofalik, line 833: Are you sure this will always work? I didn't analyze it but it looks suspicious. Does mapper always prefer a position inside a text node? I wonder when it returns null (cause AFAICS you expect it) and if you are ready to handle it.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -1349,6 +1366,21 @@ function downcastElementToElement( config ) {
 
 	return dispatcher => {
 		dispatcher.on( 'insert:' + config.model, insertElement( config.view ), { priority: config.converterPriority || 'normal' } );
+
+		if ( config.triggerBy ) {
+			if ( Array.isArray( config.triggerBy.attributes ) ) {
```
scofalik, line 1371: Why Array.isArray and not simply `!config.triggerBy.attributes` ?

scofalik (approved):
I didn't go through it as thoroughly as the last time. AFAICS, the things that we discussed are changed and obviously, the code looks simpler. I didn't try to analyze if all methods are correctly written. From my POV, I approve this PR, though I left some suggestions in the review.

jodator:
> I didn't go through it as thoroughly as the last time. AFAICS, the things that we discussed are changed and obviously, the code looks simpler. I didn't try to analyze if all methods are correctly written. From my POV, I approve this PR, though I left some suggestions in the review.

Big thanks! Will take it from here with @niegowski :)

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -489,6 +567,96 @@ export default class DowncastDispatcher {
...
+	 * Those are defined by a `triggerBy` configuration for
+	 * {@link module:engine/conversion/downcasthelpers~DowncastHelpers#elementToElement `elementToElement()`} conversion helper.
+	 *
+	 * @private
+	 * @param {String} eventName Event name to check.
+	 * @param {String} elementName Element name to check.
+	 * @returns {Boolean}
+	 */
+	_isReconvertTriggerEvent( eventName, elementName ) {
+		return this._reconversionTriggerEventToElementNameMapping.has( eventName ) &&
```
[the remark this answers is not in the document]
jodator: Yep, you're right.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
```diff
@@ -121,6 +123,14 @@ export default class DowncastDispatcher {
...
 		 */
 		this.conversionApi = Object.assign( { dispatcher: this }, conversionApi );
+
+		/**
+		 * Maps conversion event names that will trigger element reconversion for given element name.
+		 *
+		 * @type {Map<String, String>}
+		 * @private
+		 */
+		this._reconversionTriggerEventToElementNameMapping = new Map();
```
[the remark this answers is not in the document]
jodator: OK after a second thought I've renamed that a bit since it is used in a context anyway.

packages/ckeditor5-engine/src/conversion/downcastdispatcher.js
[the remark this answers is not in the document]
jodator: Actually not - here you can have either element or text. I'll clean things up a bit here to make that clear.

packages/ckeditor5-engine/src/conversion/downcasthelpers.js
jodator: Because I wanted to be sure it is an array. But we do not test for this after all so an early Exception on weird config might be better.

packages/ckeditor5-engine/tests/manual/slotconversion.js
```diff
@@ -0,0 +1,293 @@
...
+	editor.conversion.for( 'upcast' ).add( getBoxUpcastConverter( editor ) );
+
+	editor.conversion.for( 'downcast' ).elementToElement( {
+		model: 'box',
+		view: downcastBox,
+		triggerBy: [
+			'attribute:meta:box',
+			'insert:boxField',
+			'remove:boxField'
+		]
```
jodator, line 209:
```suggestion
		triggerBy: {
			attributes: [ 'meta' ],
			children: [ 'boxField' ]
		}
```

niegowski (commented):
Only some very minor things. Other than that now this code is much more readable than during my previous review 🙂

niegowski (approved):

PR #8102, opened 2020-09-17
Fix the token refresh timeout time calculations

jczapiewski-cksource:

Other (cloud-services-core): Fix the token refresh timeout time calculations. 

---

### Tickets to close

Closes cksource/ckeditor5-internal#329.

_Do not paste it in the commit merge message!_

---

### Additional information

After these changes token should be properly refreshed after the default time (1 hour) if the token payload doesn't contain the `exp` property and refresh timeout time calculations should respect the fact that the `exp` property in the token payload is always in seconds.


czerwonkabartosz (approved):

pomek:
_Merged as internal because it fixes a bug introduced in the same iteration._

PR #8108, opened 2020-09-18
Fixed a quick image flicker of image resize frame when inserting an image

### LINKED ISSUE #7548: Millisecond glitch on image upload

### LINKED ISSUE #8088: Image resize outline flickers when pasting an image
> Also, besides image upload I can see that normal image paste (image copied from the editor) is affected too. The resize handles are repositioned with a long delay, even 300ms-500ms. This definitely didn't happen before. And as Fred checked, this is not related to WidgetTypeAround. You can even observe the glitch on this gif:
> 
> 
> 
> It'd be good to know WTH is happening there. Bisecting it might actually help a lot.

_Originally posted by @Reinmar in https://github.com/ckeditor/ckeditor5/issues/7548#issuecomment-662265194_

mlewand:

Fix (link): Fixed a quick image flicker of image resize frame when inserting an image. Closes #8088.

Other (widget): `WidgetResize#visibleResizer` is now a public property.

Other (widget): `WidgetResize#getResizerByViewElement()` is now a public method.

Other (widget): `WidgetResize` will now automatically set `WidgetResize#visibleResizer` when calling `WidgetResize#attachTo()` if the corresponding resizer's element is focused during the call.

### Additional information

Note: the branch is `i/7548` instead of `i/8088` - sorry about that.

At first I wanted to implement this fix by (synchronously) redrawing the image resizer frame once the image view element gets inserted or has a `src` attribute modified. However this proved to be not enough, as still it **could** randomly happen that at a time element is inserted it still doesn't have the dimensions.

Generally it takes random amount of time between inserting an `img` with `src` and its getting a dimension box, because browser will give it dimension box as soon as it reads proper bytes from the network request - but this happens (sometimes way) before [`window#load` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event).

This means you can't draw wrapper before image is loaded, because browser will change the image size before the actual `load` event occurs - which will cause the flicker.

So my second solution was to simply hide the resizer frame for a given resizer until its image is loaded and this works reliably.

Here's a frame-by-frame dump how it looks like **after** this PR (greatly enlarged resize handles for better visibility):

<table>
	<tr>
		<td width="25%">frame 1</td>
		<td width="25%">frame 2</td>
		<td width="25%">frame 3</td>
		<td width="25%">frame 4</td>
	</tr>
</table>

To see how it looks like **before** PR see [the comment in the original issue](https://github.com/ckeditor/ckeditor5/issues/8088#issuecomment-693963129).


mlewand:
@ckeditor/qa-team can you give it a _rigorous_ (a.k.a. merciless) _testing_ this time?

The gist of change is that the resizer is not visible until the image is **loaded**.

The things that come to my mind:

*   undo integration
*   pasting images that link different resources (image on some slow server, image on a fast local server, maybe embedded images?)
*   pasting multiple images at once

Mgsy:
## Steps to reproduce

1.  Insert two images.
2.  Select them.
3.  Cut.
4.  Undo.

## Current result

Each image displays a resize handler. After using the handler, the image is incorrectly resized and the handler changes its size.

## GIF

Mgsy:
Apart from the above, it looks good :+1:

mlewand:
Excellent catch @Mgsy - that's exactly what I've been counting for. Fixed the problem in 646cd18.

The resizer is no longer visible in the case that you provided. This change also slightly improves editor performance when the focus was moved away before the image loaded.

Mgsy:
It looks good :+1:

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
```diff
@@ -74,13 +78,53 @@ export default class ImageResizeHandles extends Plugin {
...
+		} );
+
+		editingView.document.on( 'imageLoaded', imageLoadCallback );
+
+		function imageLoadCallback( evt, domEvent ) {
+			const handleHost = resizer._getHandleHost();
+
+			if ( domEvent.target.isSameNode( handleHost ) ) {
+				editingView.change( writer => {
+					if ( editor.plugins.get( WidgetResize )._visibleResizer == resizer ) {
```
Reinmar, line 117: Let's add \_visibleResizer to the public API. It's already used cross-package for testing purposes so it should be public already.

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
```diff
@@ -9,6 +9,7 @@
 
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import WidgetResize from '@ckeditor/ckeditor5-widget/src/widgetresize';
+import ImageLoadObserver from './../image/imageloadobserver';
```
Reinmar, line 12:
```suggestion
import ImageLoadObserver from '../image/imageloadobserver';
```

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
```diff
@@ -74,13 +78,53 @@ export default class ImageResizeHandles extends Plugin {
...
+	 * @private
+	 * @param {module:widget/widgetresize/resizer~Resizer} resizer
+	 * @param {module:engine/view/containerelement~ContainerElement} widget
+	 */
+	_hideResizerUntilImageIsLoaded( resizer, widget ) {
+		// Given that #7548 is fixed, this logic might no longer be needed.
+		const editor = this.editor;
+		const editingView = editor.editing.view;
+
+		editingView.change( writer => {
```
Reinmar, line 106: So we first attach a resizer to the image and then hide it. Why we don't attach a hidden resizer? The order seems to be wrong here.

Also, maybe the image could change the class when it's loading and loaded by itself? That may be helpful in other features. So, what I mean is that this code could be part of the core image feature or a new plugin that we could easily reuse.

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
```diff
@@ -74,13 +78,53 @@ export default class ImageResizeHandles extends Plugin {
...
+					if ( editor.plugins.get( WidgetResize )._visibleResizer == resizer ) {
+						// Small optimization to redraw only a resizer that is visible/focused. Hidden resizers should remain
+						// unaffected. It also fixes https://github.com/ckeditor/ckeditor5/pull/8108#issuecomment-695949745.
+						resizer.redraw();
+					}
+					writer.removeClass( 'image_resizer_loading', widget );
+				} );
+
+				// Remove image load listener for optimization, as it is no longer needed.
+				editingView.document.off( 'imageLoaded', imageLoadCallback );
```
Reinmar, line 126: I don't see this tested. If this is going to be needed after the proposed refactoring, it'd be good to cover this.

Reinmar (changes requested):
I think the implementation needs to be improved as it adds a bit of unnecessary complexity now (we do one thing to then kinda redo it).

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
Reinmar: BTW, you could check where this observer is already used and check whether rewriting this as a plugin could not help.

mlewand:
@ckeditor/qa-team again, I'll ask you for a solid test as it was a major rewrite.

mlewand:
@ckeditor/qa-team again, I'll ask you for a solid test as it was a major rewrite.

FilipTokarski:
1. First, in manual test http://localhost:8125/ckeditor5-image/tests/manual/imageresize.html add `imageUpload` to editor toolbar
2. Upload some image

**Result:**
- there are multiple blue image outlines with multiple resize handlers, which have different size than the image  
  
0_image3

- if you start typing somewhere, they disappear but appear again if you try to resize the image  
  
0_image1

mlewand:
Nice catch @FilipTokarski fixed this issue. Can you give it another spin?

FilipTokarski:
Hmm the additional lines are gone now, but it seems like the flicker itself is back. Again, it happens only when you upload. When you paste, it's ok. I'm able to reproduce it more clearly with CPU 4x slowdown:  
  
0_image2

FilipTokarski:
Pasting in Safari:  
  
0_image3

FilipTokarski:
I didn't find anything new apart from those two things ☝️  Resizing works smooth and I didn't see any additional handlers' bugs when undoing/redoing the pasting.

mlewand:
@FilipTokarski unfortunately this #8088 won't fix rendering issues in upload plugin. It's tracked in a separate issue where I made a [decent writeup on issues during the image creation](https://github.com/ckeditor/ckeditor5/issues/7548#issuecomment-692593256) (upload is covered in "Image upload (chrome)" section).

As for the second case (pasting) I can see that the resize wrapper has the correct size during the first paint:

<table><tbody><tr><td><figure class="image"></figure><p>Flash of image-less figure container, that's the scope of #7548.</p></td><td><figure class="image"></figure><p>The first resizer paint - wrapper has proper size.</p></td></tr></tbody></table>

tl;dr - the idea in this PR is to keep the resizer aligned with widget wrapper

I added one unit test that I found to be missing.

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
mlewand: Actually that was a good point, I changed the approach further and instead of playing with HTML classes I'm delaying resizer creation until the image is loaded.

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
mlewand: There's no longer need for that - changed implementation registers only a single listener.

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
```diff
@@ -37,21 +38,45 @@ export default class ImageResizeHandles extends Plugin {
...
+
+		this._addResizerCreationListener();
+	}
+
+	/**
+	 * Attaches the listeners responsible for creating a resizer for each editor image.
+	 *
+	 * @private
+	 */
+	_addResizerCreationListener() {
```
Reinmar, line 52:
```suggestion
	_setupResizersCreator() {
```

Reinmar:
LGTM :heart: 

Waiting for CI.

PR #8119, opened 2020-09-21
Manual link decorators for images will not crash decorators for linked text

### LINKED ISSUE #7975: Link decorators are lost after initialization if LinkImage plugin is enabled
I have link decorators in config
```
link: {
    decorators: {
        openInNewTab: {
            mode: 'manual',
            label: 'Открыть в новом окне',
            defaultValue: true,
            attributes: {
                target: '_blank'
            }
        },
        likeAButton: {
            mode: 'manual',
            label: 'Кнопка',
            defaultValue: false,
            attributes: {
                class: 'likeAButton',
            }
        }
    }
}
```
I add links with decorators on and save data. Then I open this data again, and the decorators seem off. And if I output data it hasn't decorators.

### ✔️ Expected result

Decorators should be turn on.

### ❌ Actual result

Decorators turned off.

UPD: I add `converterPriority: 'high'` to upcast conversion for manual decorators in linkediting.js and now it seems to work fine.

pomek:

Fix (link): Manual decorators will be handled properly for linked images and linked text. Closes #7975.

### DIFF
--- a/packages/ckeditor5-link/tests/linkimageediting.js
+++ b/packages/ckeditor5-link/tests/linkimageediting.js
@@ -684,6 +684,7 @@ describe( 'LinkImageEditing', () => {
 			} );
 
 			it( 'should upcast the decorators when linked image (figure > a > img)', () => {
+				// (#7975)
 				editor.setData(
 					'<figure class="image">' +
 						'<a class="gallery" href="https://cksource.com" target="_blank" rel="noopener noreferrer" download="download">' +
@@ -715,6 +716,7 @@ describe( 'LinkImageEditing', () => {
 			} );
 
 			it( 'should upcast the decorators when linked image (a > img)', () => {
+				// (#7975)
 				editor.setData(
 					'<a class="gallery" href="https://cksource.com" target="_blank" rel="noopener noreferrer" download="download">' +
 						'<img src="sample.jpg" alt="bar">' +
--- a/packages/ckeditor5-link/tests/linkimageediting.js
+++ b/packages/ckeditor5-link/tests/linkimageediting.js
@@ -587,35 +587,57 @@ describe( 'LinkImageEditing', () => {
 		} );
 
 		describe( 'upcast converter', () => {
-			let editor;
+			let editor, model;
 
-			it( 'should upcast attributes from initial data', async () => {
-				editor = await VirtualTestEditor.create( {
-					initialData: '<figure class="image"><a href="url" target="_blank" rel="noopener noreferrer" download="file">' +
-						'<img src="/assets/sample.png"></a></figure>',
-					plugins: [ Paragraph, LinkImageEditing ],
-					link: {
-						decorators: {
-							isExternal: {
-								mode: 'manual',
-								label: 'Open in a new window',
-								attributes: {
-									target: '_blank',
-									rel: 'noopener noreferrer'
-								}
-							},
-							isDownloadable: {
-								mode: 'manual',
-								label: 'Downloadable',
-								attributes: {
-									download: 'file'
+			beforeEach( () => {
+				return VirtualTestEditor
+					.create( {
+						plugins: [ Paragraph, LinkImageEditing ],
+						link: {
+							decorators: {
+								isExternal: {
+									mode: 'manual',
+									label: 'Open in a new tab',
+									attributes: {
+										target: '_blank',
+										rel: 'noopener noreferrer'
+									}
+								},
+								isDownloadable: {
+									mode: 'manual',
+									label: 'Downloadable',
+									attributes: {
+										download: 'download'
+									}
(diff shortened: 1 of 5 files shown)


pomek:
@ckeditor/qa-team, could you take a look at the fix?

tests/manual/all-features.md
```diff
@@ -20,16 +22,30 @@ There should be "two pages" with two sections per each page. Pages should be sep
 
 **Code blocks in the table** - in the table (4x3), in the second and fourth columns should be visible code snippets.
 
+**Horizontal line** - There is the `<hr>` in the source HTML. It should be displayed in the editor.
+
+**Link images + Link decorators** - in the table (3x2), there be a link and two linked images that uses the manual decorators feature:
+
+  - Left column: a plain text with two decorators enabled: `Open in a new tab`, `Downloadable`
+  - Middle column: an image with the caption that is a `Gallery link`
+  - Right column: an image without the caption with enabled all decorators (listed above)
```
pomek, line 31:
```suggestion
**Link images + Link decorators** - in the table (3x2), there are a linked text and two linked images that uses the manual decorators feature:
```

  - Left column: the text with two decorators enabled: `Open in a new tab`, `Downloadable`
  - Middle column: an image with the caption that is a `Gallery link`
  - Right column: an image without the caption with enabled all decorators (listed above)
```

**Comment by FilipTokarski** [comment]
Looks like everything is ok 👍

File: packages/ckeditor5-link/tests/linkimageediting.js [hunk 2]
```diff
@@ -711,4 +711,155 @@ describe( 'LinkImageEditing', () => {
...
+				setModelData( model,
+					'<image alt="bar" src="sample.jpg"></image>' +
+					'<paragraph>' +
+						'<$text>https://cksource.com</$text>' +
+					'</paragraph>'
+				);
+
+				model.change( writer => {
+					// <image>
+					writer.setSelection( model.document.getRoot().getChild( 0 ), 'on' );
```
mlewand, line 830: A much simpler approach is to select image by enclosing it in brackets in line

https://github.com/ckeditor/ckeditor5/blob/dec11ed084ca8670569075d0a697f35a2997536d/packages/ckeditor5-link/tests/linkimageediting.js#L822

Like so `'[<image alt="bar" src="sample.jpg"></image>]' +`.

mlewand: On a side note: I checked that the image does not necessarily need a focus, as the test passes if I remove this particular `model.change()` call (I assume that editor will simply put the focus to the first element by default).

packages/ckeditor5-link/tests/linkimageediting.js
```diff
@@ -711,4 +711,155 @@ describe( 'LinkImageEditing', () => {
...
+								}
+							},
+							isDownloadable: {
+								mode: 'manual',
+								label: 'Downloadable',
+								attributes: {
+									download: 'download'
+								}
+							},
+							isGallery: {
```
mlewand, line 739: I think that for better readability we can simplify it just to one or max two decorators to get the full coverage and keep the markup simpler.

packages/ckeditor5-link/tests/linkimageediting.js
```diff
@@ -711,4 +711,155 @@ describe( 'LinkImageEditing', () => {
...
+					model = editor.model;
+				} );
+		} );
+
+		afterEach( () => {
+			return editor.destroy();
+		} );
+
+		// See: #7975.
+		describe( 'manual decorators: linked text and linked image', () => {
```
mlewand, line 760: We're testing the upcast implementation here and [there are already two tests for handling link decorators during the upcast](https://github.com/ckeditor/ckeditor5/blob/dec11ed084ca8670569075d0a697f35a2997536d/packages/ckeditor5-link/tests/linkimageediting.js#L627-L711).

Wouldn't it make sense to put your unit tests next to it?

I think a good idea would be to nest decorator-related suite in the `upcast converter` suite, there you'd have your new tests as well as the tests that are already there. Also old tests could reuse the editor that you're creating in `beforeEach()` rather than have a [copy&paste editor initialization at the beginning](https://github.com/ckeditor/ckeditor5/blob/dec11ed084ca8670569075d0a697f35a2997536d/packages/ckeditor5-link/tests/linkimageediting.js#L593-L616).

That should result with nice and tight suite.

mlewand (commented):
The code works and looks fine.

Good call with adding it to all features sample 👍

I proposed one minor change to extract common logic for picking the image element in dec11ed. This will protect us from case when one occurrence will be modified without updating the other.

There are, however, couple of minor things that could be polished in the unit tests.

mlewand (approved):
All good 👍

PR #8124, opened 2020-09-22
Use same format for error name in attachLinkToDocumentation as in CKEditorError

### LINKED ISSUE #8097: Followups for the changes in error codes

We changed the format recently (#7822), but the change is not complete:

*   The documentation was not updated (see e.g. https://github.com/ckeditor/ckeditor5/blob/3e6ea99/packages/ckeditor5-utils/src/ckeditorerror.js#L36-L39 or https://github.com/ckeditor/ckeditor5/blob/3e6ea99/packages/ckeditor5-utils/src/ckeditorerror.js#L139-L141).
*   This regepx is unsafe: https://github.com/ckeditor/ckeditor5/blob/3e6ea99/packages/ckeditor5-utils/src/ckeditorerror.js#L147-L148 unless it's kept in perfect sync with a regexp in the ESLint rule. If you're allowed to write an error like `CKEditorError( 'image-invalid-%-size' )`, then this error will not be enhanced with a link to the documentation. This will be super hard to spot – it will be a silent issue for us. We need to at least check that such a situation is not possible and commented in the code that those regexps are in sync ensuring that all errors are well defined.
*   I don't understand why `console.warn` does not have to follow the same convention. What's the reason for that? If it was following it, we could solve the previous point with a more verbose solution (throw a new error :D).
*   Finally, this: cksource/ckeditor5-internal#327

---

jodator:

Internal: Use same format for error name in attachLinkToDocumentation as in CKEditorError. Closes #8097.

---

### Additional information

*   I've added this to ESLint rule as well: ckeditor/eslint-plugin-ckeditor5-rules#1.

### DIFF
--- a/packages/ckeditor5-core/src/plugincollection.js
+++ b/packages/ckeditor5-core/src/plugincollection.js
@@ -117,6 +117,12 @@ export default class PluginCollection {
 		const plugin = this._plugins.get( key );
 
 		if ( !plugin ) {
+			let pluginName = key;
+
+			if ( typeof key == 'function' ) {
+				pluginName = key.pluginName || key.name;
+			}
+
 			/**
 			 * The plugin is not loaded and could not be obtained.
 			 *
@@ -131,15 +137,7 @@ export default class PluginCollection {
 			 * @error plugincollection-plugin-not-loaded
 			 * @param {String} plugin The name of the plugin which is not loaded.
 			 */
-			const errorMsg = 'plugincollection-plugin-not-loaded: The requested plugin is not loaded.';
-
-			let pluginName = key;
-
-			if ( typeof key == 'function' ) {
-				pluginName = key.pluginName || key.name;
-			}
-
-			throw new CKEditorError( errorMsg, this._context, { plugin: pluginName } );
+			throw new CKEditorError( 'plugincollection-plugin-not-loaded', this._context, { plugin: pluginName } );
 		}
 
 		return plugin;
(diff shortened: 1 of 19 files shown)


jodator:
Looks like only the documentation is failing on the CI so not related to this change.

packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -206,12 +206,12 @@ export default class PluginCollection {
 			 * @error plugincollection-plugin-not-found
 			 * @param {Array.<String>} plugins The name of the plugins which could not be loaded.
 			 */
-			const errorMsg = 'plugincollection-plugin-not-found: Some plugins are not available and could not be loaded.';
+			const errorId = 'plugincollection-plugin-not-found';
```
Reinmar, line 209: One more error to fix: https://github.com/ckeditor/ckeditor5/blob/b3a0bba/packages/ckeditor5-core/src/plugincollection.js#L134

mlewand: @jodator Maybe the tests should also expect one space after the `errorName`? So that line https://github.com/ckeditor/ckeditor5/blob/14007d342bdef02009425c059af4308c4012eb17/packages/ckeditor5-core/tests/plugincollection.js#L524 becomes:

```
					/^plugincollection-plugin-not-loaded /, editor, { plugin: 'foo' }
```

jodator: I'd go with those changes in a follow-up: #8140. Let's fix it once and in all places.

Fixing one case is not worth the effor IMO.

mlewand (approved):
All right, I was unbale to find any other error instances.

There are instances in collaboration/external projects though - do we have a ticket for that already?

jodator:
> There are instances in collaboration/external projects though - do we have a ticket for that already?

AFAICS there was no use of `attachLinkToDocumentation` elsewhere.

PR #8199, opened 2020-10-01
Cleaned up usage of the `t()` function for one interpolated argument.

### LINKED ISSUE #6645: Clean usage of `t()` function
In ckeditor/ckeditor5-utils#334 we decided that the second parameter (\`values\`) of the translation function (\`t()\`) will be able to take only a single argument used to interpolate the string - a string or a number. Hence, we should fix places where only one value is used in the `t( 'foo %0', [ value ] )` form.

ma2ciek:

Internal: Cleaned up usage of the \`t()\` function for one interpolated argument. Closes #6645.

---

### Additional information


jodator (approved):
LGTM :+1:

PR #8216, opened 2020-10-06
Media embed command is disabled when selected media is in a table cell

### LINKED ISSUE #7604: Media embed command is disabled when selected media is in a table cell

1.  Put the media in a table cell.
2.  Select the media.

### ✔️ Expected result

The UI is enabled and a user can change the media URL.

### ❌ Actual result

UI is disabled. There's no way to edit the media.

This is because the [`parent` passed to the `Schema#checkChild` is a `tableRow.`](https://github.com/ckeditor/ckeditor5/blob/2f1568d5598437aa37639269568b9012e7febc21/packages/ckeditor5-media-embed/src/mediaembedcommand.js#L43)

## 📃 Other details

*   Browser: \*
*   CKEditor version: master
*   Not a regression (checked against v16.0.0).

---

pkwasnik:

Fix (media-embed): Enable media embed command when selected media is in a table cell. Closes #7604.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-media-embed/src/mediaembedcommand.js
+++ b/packages/ckeditor5-media-embed/src/mediaembedcommand.js
@@ -35,7 +35,8 @@ export default class MediaEmbedCommand extends Command {
 
 		let parent = insertPosition.parent;
 
-		if ( parent.isEmpty && model.schema.isBlock( parent ) ) {
+		// The model.insertContent() will remove empty parent (unless it is a $root or a limit).
+		if ( parent.isEmpty && !model.schema.isLimit( parent ) ) {
 			parent = parent.parent;
 		}
 
@@ -68,4 +69,3 @@ export default class MediaEmbedCommand extends Command {
 		}
 	}
 }
-
--- a/packages/ckeditor5-media-embed/src/mediaembedcommand.js
+++ b/packages/ckeditor5-media-embed/src/mediaembedcommand.js
@@ -30,12 +30,12 @@ export default class MediaEmbedCommand extends Command {
 		const model = this.editor.model;
 		const selection = model.document.selection;
 		const schema = model.schema;
-		const position = selection.getFirstPosition();
+		const insertPosition = findOptimalInsertionPosition( selection, model );
 		const selectedMedia = getSelectedMediaModelWidget( selection );
 
-		let parent = position.parent;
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-media-embed/src/mediaembedcommand.js
```diff
@@ -30,12 +30,12 @@ export default class MediaEmbedCommand extends Command {
...
 		const schema = model.schema;
-		const position = selection.getFirstPosition();
+		const insertPosition = findOptimalInsertionPosition( selection, model );
 		const selectedMedia = getSelectedMediaModelWidget( selection );
 
-		let parent = position.parent;
+		let parent = insertPosition.parent;
 
-		if ( parent != parent.root ) {
+		if ( parent.isEmpty && model.schema.isBlock( parent ) ) {
```
jodator, line 38:
```suggestion
		// The model.insertContent() will remove empty parent (unless it is a $root or a limit).
		if ( parent.isEmpty && !model.schema.isLimit( parent ) ) {
```

I thought the `isBlock()` check will be OK as the behavior suggests. But I wasn't certain why. It turns out that the [insertion logic](https://github.com/ckeditor/ckeditor5/blob/0782dc229381ea185a4d67ec449550ba57a9f1c5/packages/ckeditor5-engine/src/model/utils/insertcontent.js#L560-L583) is a bit different - hence the proposal to change.

I've dug up how this behaves - ATM I'm not sure which is better but let's keep it consistent with `model.insertContent()` checks. I thought it checked for `block`s but it will remove any non-limit empty element on insert.

A "limit" is an element that can contain a selection. In most cases, we do not go past the limit elements. An example is a `<tableCell>` where you can act as like inside editor root. For instance, if you use <kbd>ctrl</kbd> + <kbd>a</kbd> the first use will select all table cell contents.  
The `$root` element is a special case for the limit as you can't go with selection past root.

jodator (changes requested):
Minor fixes are needed but in general good job :)

1. One test uses the wrong schema definition to check the media in a table cell case.
2. I've given you a hint with `isBlock()` check which turned out to be partially OK. I've suggested a small change to the command check as it would match the behaviour of `model.insertContent()`. Sorry for that.

jodator (approved):

PR #8231, opened 2020-10-08
Use npm in favor of yarn for consistency

lauriii:

Docs: Use npm in favor of yarn in the advanced setup for consistency 

---

### Additional information

Assumed that you want to use npm instead of yarn given that in most places you are using npm.


Reinmar:
Hi! We started discussing internally whether we want to promote yarn or not, but the decision is to use npm everywhere except the dev environment docs (the dev env relies on yarn anyway).

PR #8255, opened 2020-10-14
Focusing the URL input when the image upload panel shows up

### LINKED ISSUE #7896: Consider focusing the URL input when the image upload panel shows up

ATM when the user opens the panel, the focus remains in editable and there's no way to type the URL straight away. 

We could adopt the media embed (also: link insertion UI) UX in this matter and focus the field immediately to make it more user-friendly.

---

pkwasnik:

Fix (image): Focusing the URL input when the image upload panel shows up. Closes #7896.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
+++ b/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
@@ -113,6 +113,9 @@ export default class ImageInsertUI extends Plugin {
 					insertImageViaUrlForm.label = t( 'Insert image via URL' );
 				}
 			}
+		// Note: Use the low priority to make sure the following listener starts working after the
+		// default action of the drop-down is executed (i.e. the panel showed up). Otherwise, the
+		// invisible form/input cannot be focused/selected.
 		}, { priority: 'low' } );
 
 		imageInsertView.delegate( 'submit', 'cancel' ).to( dropdownView );
--- a/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
+++ b/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
@@ -113,7 +113,7 @@ export default class ImageInsertUI extends Plugin {
 					insertImageViaUrlForm.label = t( 'Insert image via URL' );
 				}
 			}
-		} );
+		}, { priority: 'low' } );
 
 		imageInsertView.delegate( 'submit', 'cancel' ).to( dropdownView );
 		this.delegate( 'cancel' ).to( dropdownView );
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-image/src/imageinsert/imageinsertui.js
```diff
@@ -113,7 +113,7 @@ export default class ImageInsertUI extends Plugin {
 					insertImageViaUrlForm.label = t( 'Insert image via URL' );
 				}
 			}
-		} );
+		}, { priority: 'low' } );
```
jodator, line 116: Could you add a comment like here:  
https://github.com/ckeditor/ckeditor5/blob/fa810d440464e112bbb5428895c161b09ae9fd26/packages/ckeditor5-media-embed/src/mediaembedui.js#L71-L73

packages/ckeditor5-image/tests/imageinsert/imageinsertui.js
```diff
@@ -278,6 +278,18 @@ describe( 'ImageInsertUI', () => {
...
 				sinon.assert.calledOnce( cancelSpy );
 			} );
+
+			it( 'should focus on "insert image via URL" input after opening', () => {
+				let spy;
+
+				dropdown.on( 'change:isOpen', () => {
+					const first = dropdown.panelView.children.first;
+					spy = sinon.spy( first, 'focus' );
+				}, { priority: 'highest' } );
```
jodator, line 288:
```suggestion
                // The ImageInsertPanelView is added on first open.
                // See https://github.com/ckeditor/ckeditor5/pull/8019#discussion_r484069652
                dropdown.on( 'change:isOpen', () => {
                    const first = dropdown.panelView.children.first;
                    spy = sinon.spy( first, 'focus' );
                }, { priority: 'highest' } );
```

jodator: Additionally, the `const first` might be either:

*   inlined
*   renamed to `imageInsertPanelView` to showcase which view it exactly checks.

That's because you would have to check the code to remember what's that first child.

jodator (changes requested):
The fix works fine as tested :smiley: 

Minor additions are needed to not lose the knowledge we gained during this ticket.

packages/ckeditor5-image/tests/imageinsert/imageinsertui.js
pkwasnik: right, corrected

jodator (approved):
Cool, thanks!

PR #8267, opened 2020-10-15
#1098: Made input labels accessible across the editor UI

### LINKED ISSUE #1098: Inaccessible input for link's URL

🐞 Bug report

## 💻 Version of CKEditor

Version from [docs](https://docs.ckeditor.com/ckeditor5/latest/examples/builds/classic-editor.html).

## 📋 Steps to reproduce

1. Select some text.
2. Click link button.

## ✅ Expected result

Shown `input` has proper label that is announced by screen reader.

## ❎ Actual result

Only `[placeholder]` value is announced by VoiceOver, which is confusing as placeholder is set to a URL.

Chrome devtools Accessibility Inspector shows that current label is empty:

The issue is caused by the fact that `label` for the input is hidden with `display: none`. It should be hidden with more accessible technique, e.g. [one proposed by H5BP](https://github.com/h5bp/html5-boilerplate/blob/7e74f27208ef6e62cda92057dfb853b51d3649dd/dist/css/main.css#L128-L147).

In the ideal situation, link dialog would display both `label` and hint outside the `input` itself:

Super ugly visualisation of input with label above and hint below

If the hint is present, then it should be bound to the `input` with `[aria-describedby]` attribute.

Some context:

* https://www.smashingmagazine.com/2018/06/placeholder-attribute/
* https://www.filamentgroup.com/lab/a11y-form-labels.html
* https://www.scottohara.me/blog/2018/05/24/keep-labels-clean.html

### LINKED ISSUE #8242: Refine the accessible (material) inputs PoC

A follow-up of #1098 and #7952.

oleq:

Feature: Made input labels accessible across the editor UI. Closes #1098. Closes #8242.

BREAKING CHANGE (theme-lark): The look and behavior of the `LabeledFieldView` UI component (used for displaying fields across the project) has changed. This may require changes in your integration if it customizes the `.ck-labeled-field-view` selector (or its internals).

---

### Additional information

⚠️⚠️⚠️ This PR should be merged at the beginning of an iteration because the scope is so vast and it will take some time to detect possible issues by the team. ⚠️⚠️⚠️

Some notes for QA:

*   Below there are screenshots of places affected by this PR.
*   Make sure this PR is tested against other languages (e.g. German, which has longer labels).
*   Make sure this PR does not break CF.


oleq:
@ckeditor/qa-team Can you take a look at this PR?

oleq:
@Reinmar There's [one architectural aspect](https://github.com/ckeditor/ckeditor5/pull/8267/files#diff-df0e9126b9828ee9a6e00208ea0c7c5db3b26aaa0284f58aa1c834bc4f7cacb7R54) of the PR I'd love to know your opinion about.

FilipTokarski:
In _table properties_ and _cell properties_ balloons - `border color` and `background color` inputs have empty labels  
a11y_1

FilipTokarski:
I checked it with screen readers on mac an win, on different browsers and different languages. Apart from lack of `color` labels reported above, the rest seems to be fine 👍   
In some languages the label's text is longer than the input, but it gets truncated with `...` and it looks ok, so I guess it's not an issue.

pkwasnik (approved):

oleq:
> In _table properties_ and _cell properties_ balloons - `border color` and `background color` inputs have empty labels

A nice catch! I'm not sure this is in the scope of this PR, though. In this PR we're talking about visual accessibility while this is screen reader accessibility and this problem also occurs on master. 



Anyway, I checked this in VoiceOver and the screen reader fails to read the label for the input. I'm not sure how to fix it, though, because the field is more complex than this. @Comandeer can you give us some thoughts on this?

oleq:
> A nice catch! I'm not sure this is in the scope of this PR, though. In this PR we're talking about visual accessibility while this is screen reader accessibility and this problem also occurs on master.

I meant: let's have a follow-up for this 😛

Reinmar:
The contrast difference between a placeholder and filled values is IMO too low. I wonder if this makes it harder to understand whether a field is filled:



This is the limit value for WCAG AA:



IDK if reducing the contrast to 4.5 makes sense as the placeholder text in the input is very important (it tells you what that field is about). However, such a form is easier to understand to me.

Alternatively, we could make the default text a bit darker and keep the current placeholder color. However, we should then make the icons stop following this variable (they'd otherwise get darker too):

Reinmar:
This is a blocker:

Reinmar:
Not sure if it's supposed to work on http://localhost:8125/ckeditor5-theme-lark/tests/manual/theme.html where it says that it's supposed to be labeled but there's no label:

Reinmar:
One more, less obvious thing about RTL - the label should be on the right:



The placeholder too:

Reinmar:
As for `disableCSSTransition`:

It's unclear for me from the documentation when this is applicable. My guess is that when we display some form with one of the fields pre-focused, then we don't want to show the transition. Is that correct? If so, then, it'd be lovely to update this documentation with more info.

Other than that – `disableCssTransitions()`.

And finally – the solution is far from perfect in this sense that it leaks presentational logic to JS. Unfortunately, I suppose there's no other way around this and thus:

*   We can live with this because it is what it is and there's no point in lying about this.
*   We can try to generalize it in some fancy way but, hey, leaky abstractions...
*   We can try to combine it with LabeledInput's `focus()`? But what if we have some non-focus-related transition? OTOH, what if some non-focus-related transition should actually happen while the focus-related one should not? The current naming groups all them.
*   We can try to find some slightly more "JSy" name but I can't think of any.

I can see it is used in 3 places, so it's not bad. If you want to keep it this way, I'd be fine with that.

If you'll decide to keep this solution, I think that classes that mix it (like `TextAlternativeFormView`) should either implement some additional interface?

Reinmar (changes requested):
As commented. The RTL issues are blockers. The test issues probably not but since we're not in a hurry and these might be smaller things, it'd be good to fix these too. Finally, as for the whole disabling transitions problem – we may KISS. The naming would need to be fixed and the interface thing.

Reinmar:
> And finally – the solution is far from perfect in this sense that it leaks presentational logic to JS. Unfortunately, I suppose there's no other way around this and thus:

There's another option, BTW, to move these methods to an object inside a view. Like:

```
this.transitionDisabler = injectTransitionDisabler( this );
```

And then use:

```
someView.transitionDisabler.turnOn();
```

The idea here would be to not have two additional methods and a property on a view, but have them grouped.

oleq:
> [This is a blocker:](https://github.com/ckeditor/ckeditor5/pull/8267#issuecomment-720481391)

Addressed in 34d3a80 and ec85982.

---

> [One more, less obvious thing about RTL - the label should be on the right:](https://github.com/ckeditor/ckeditor5/pull/8267#issuecomment-720501554)

Addressed in 34d3a80.

---

[Regarding the dark theme](https://github.com/ckeditor/ckeditor5/pull/8267#issuecomment-720486741), I created a follow-up a few weeks ago #8275.

---

> It's unclear for me from the documentation when this is applicable. My guess is that when we display some form with one of the fields pre-focused, then we don't want to show the transition. Is that correct? If so, then, it'd be lovely to update this documentation with more info.

>  \* A decorator that brings possibility to temporarily disable CSS transitions using  
> \* {@link module:ui/view~View} methods. It is helpful when, for instance, the transitions should not happen  
> \* when the view is first displayed but they should work normally in other cases.

It's pretty clear to me what's the purpose of this helper. It has nothing to do with inputs (although I use it for this purpose). It could be any transition on any DOM element.

> And finally – the solution is far from perfect in this sense that it leaks presentational logic to JS. Unfortunately, I suppose there's no other way around this and thus:

IMO the only sane way to maybe improve this is to narrow the use-case to the inputs and their focus (instead of entire forms in my implementation) because currently, this is the only use case. The implementation of `injectCSSTransitionDisabling` would not change, though. So ¯\\\_(ツ)\_/¯ .

> There's another option, BTW, to move these methods to an object inside a view. Like:

I'd rather not make it available across all views in the editor when there are only a few use-cases. I like the decorator pattern.

oleq:
> Not sure if it's supposed to work on http://localhost:8125/ckeditor5-theme-lark/tests/manual/theme.html where it says that it's supposed to be labeled but there's no label:

It looks like it was broken before. Fixed in 243b0f5.

oleq:
> This is the limit value for WCAG AA:

Increased contrast in 9fad801.

> Alternatively, we could make the default text a bit darker and keep the current placeholder color. However, we should then make the icons stop following this variable (they'd otherwise get darker too):

Can we discuss it in a follow-up? I think the change that I mentioned :point\_up: makes it good enough and we can think about it later. Decoupling icon color and text color would be a serious breaking change because the whole UI/theme depends on this coupling.

PR #8274, opened 2020-10-16
Docs (utils): Add information about using callback with a single property.

jodator:

Docs (utils): Add information about using callback with a single property.

---

### Additional information

I added this example as it was not stated directly that you can use this for one property. The docs say only about merging multiple properties. You can also process the property value for bound property.the

### DIFF
--- a/packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
+++ b/packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
@@ -200,9 +200,9 @@ The property has been "renamed" in the binding and from now on, whenever `comman
 
 #### Processing a property value
 
-Another use case is processing a bounded value. For example: you may need to disable a button only if certain conditions are met. Passing a callback as the third parameter allows implementing custom logic.
+Another use case is processing the bound property value, for instance, when a button should be disabled only if certain conditions are met. Passing a callback as the third parameter allows implementing a custom logic.
 
-In the example below, a button's `isEnabled` property will be set to `true` only when `command.value` will be set to `'heading1`.
+In the example below, the `isEnabled` property will be set `true` only when `command.value` equals `'heading1`.
 
 ```js
 const command = editor.commands.get( 'heading' );
--- a/packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
+++ b/packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
@@ -198,9 +198,9 @@ button.bind( 'isOn' ).to( command, 'value' );
 
 The property has been "renamed" in the binding and from now on, whenever `command.value` changes, the value of `button.isOn` will reflect it.
 
-#### Processing property value
+#### Processing a property value
 
-The other use case is processing bounded value. For instance, you may need to disable a button only when certain conditions are met. Passing a callback as the third parameter allows implementing custom logic.
+Another use case is processing a bounded value. For example: you may need to disable a button only if certain conditions are met. Passing a callback as the third parameter allows implementing custom logic.
 
 In the example below, a button's `isEnabled` property will be set to `true` only when `command.value` will be set to `'heading1`.
 
--- a/packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
+++ b/packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
@@ -202,7 +202,7 @@ The property has been "renamed" in the binding and from now on, whenever `comman
 
 Another use case is processing the bound property value, for instance, when a button should be disabled only if certain conditions are met. Passing a callback as the third parameter allows implementing a custom logic.
 
-In the example below, the `isEnabled` property will be set `true` only when `command.value` equals `'heading1`.
+In the example below, the `isEnabled` property will be set to `true` only when `command.value` equals `'heading1`.
 
 ```js
 const command = editor.commands.get( 'heading' );
--- a/packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
+++ b/packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
@@ -206,7 +206,7 @@ In the example below, a button's `isEnabled` property will be set to `true` only
 
 ```js
 const command = editor.commands.get( 'heading' );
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
```diff
@@ -198,6 +198,17 @@ button.bind( 'isOn' ).to( command, 'value' );
...
 
+#### Processing a property value
+
+Another use case is processing a bounded value. For example: you may need to disable a button only if certain conditions are met. Passing a callback as the third parameter allows implementing custom logic.
+
+In the example below, a button's `isEnabled` property will be set to `true` only when `command.value` will be set to `'heading1`.
+
+```js
+const command = editor.commands.get( 'heading' );
+button.bind( 'isOn', 'isEnabled' ).to( command, 'value', value => value === 'heading1' );
```
oleq, line 209: My gut tells me this won't work. I couldn't find this kind of binding in [tests](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/tests/observablemixin.js#L260-L804).

jodator: It's the "should work with callback #6" test  https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/tests/observablemixin.js#L645-L659 :wink: 

The example was inspired by this:

https://github.com/ckeditor/ckeditor5/blob/b17ae267083c41aefc80863a2e6cf16ffa5989cb/packages/ckeditor5-heading/src/headingui.js#L61

ps.: We should fix those test names as we find them. That's another technical dept.

godai78: I wish I knew better what are you guys talking about...

jodator: @oleq wasn't sure if my additions actually work the way I described them. Which is a valid point since it wasn't documented before. I know it is possible as I already checked it in the tests (the test name could be improved) and the source code that uses this.

The problem was here that it wasn't documented explicitly, rather implicitly, and wasn't discoverable enough. Hence, the proposed additions to the docs.

packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
```diff
@@ -198,6 +198,17 @@ button.bind( 'isOn' ).to( command, 'value' );
 
 The property has been "renamed" in the binding and from now on, whenever `command.value` changes, the value of `button.isOn` will reflect it.
 
+#### Processing a property value
+
+Another use case is processing a bounded value. For example: you may need to disable a button only if certain conditions are met. Passing a callback as the third parameter allows implementing custom logic.
```
oleq, line 203:
```suggestion
Another use case is processing the bound property value, for instance, when a button should be disabled only if certain conditions are met. Passing a callback as the third parameter allows implementing a custom logic.
```

packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
```diff
@@ -198,6 +198,17 @@ button.bind( 'isOn' ).to( command, 'value' );
...
+
+In the example below, a button's `isEnabled` property will be set to `true` only when `command.value` will be set to `'heading1`.
```
oleq, line 205:
```suggestion
In the example below, the `isEnabled` property will be set `true` only when `command.value` equals `'heading1`.
```

packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
oleq: The `.to()` part is OK. The `.bind( 'isOn', 'isEnabled' )` is wrong IMO as there are 2 property names here ( `isOn` will not work here). I think this will throw an error (but this needs to be confirmed).

oleq (changes requested):

packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
jodator: OH, yes it is a typo.

jodator, line 209:
```suggestion
button.bind( 'isOn' ).to( command, 'value', value => value === 'heading1' );
```

jodator: I totally blanked on the first part.

jodator: Fixed. Thanks.

oleq (approved):

packages/ckeditor5-utils/docs/framework/guides/deep-dive/observables.md
```diff
@@ -198,6 +198,17 @@ button.bind( 'isOn' ).to( command, 'value' );
 
 The property has been "renamed" in the binding and from now on, whenever `command.value` changes, the value of `button.isOn` will reflect it.
 
+#### Processing a property value
+
+Another use case is processing the bound property value, for instance, when a button should be disabled only if certain conditions are met. Passing a callback as the third parameter allows implementing a custom logic.
+
+In the example below, the `isEnabled` property will be set `true` only when `command.value` equals `'heading1`.
```
AnnaTomanek, line 205:
```suggestion
In the example below, the `isEnabled` property will be set to `true` only when `command.value` equals `'heading1`.
```

packages/ckeditor5-utils/src/observablemixin.js
```diff
@@ -786,6 +786,10 @@ function attachBindToListeners( observable, toBindings ) {
  *		button.bind( 'isEnabled' ).to( command, 'isEnabled', ui, 'isVisible',
  *			( isCommandEnabled, isUIVisible ) => isCommandEnabled && isUIVisible );
  *
+ * Using custom callback allows to process the value before passing it to the target property:
```
AnnaTomanek, line 789:
```suggestion
 * Using a custom callback allows processing the value before passing it to the target property:
```

Issue #982, opened 2018-04-19
[Docs] Document heading buttons


wwalc:
I couldn't find any documentation about it so I'm not sure if anyone will notice it. And it's a pity as at least for me it's much more intuitive and easy to use. And the UI looks more consistent with it without a dropdown, it fits much better the ballon toolbar and so on.

By documenting it I mean for example having a note about it and a working demo here: https://docs.ckeditor.com/ckeditor5/latest/features/headings.html

Reinmar:
We added this feature very recently and it's not even properly reviewed yet. We'll document it once it's stabilised.

godai78:
Sooo, what do we need to add here? The guide is just being reviewed, but I can plow it tomorrow if it needs additions.

godai78:
Working branch: https://github.com/ckeditor/ckeditor5/tree/docs-heading-buttons

godai78:
OK, so there is now information about the feature option, there is a working demo with buttons and basic install-info. Hope that does the trick.

AnnaTomanek:
PR open here: #8281

PR #8281, opened 2020-10-19
Headings toolbar buttons

godai78:
Expanding the heading feature documentation with toolbar buttons feature info.

Close: #982

@Reinmar: please have a look at the code.  
@AnnaTomanek: review, please.

Thanks to @Mgsy for the helping hand with the snippets.


AnnaTomanek:
I updated the guide and snippet a bit.

I have slightly mixed feelings with two demos with heading buttons shown in the article as they are basically the same. Not a blocker, though, it's better than not having this feature shown at all.

The Common API section is missing information about `HeadingButtonsUI`, `ParagraphButtonUI` and the individual toolbar buttons. @Reinmar it would be cool if you could have a look.

Reinmar:
I'm not sure what you meant, but I see a problem in Heading 1 being equal Heading 2. 



That can be improved by changing it here:



So it's `h1`, `h2` and so on as in the previous example and then changing "Heading1" to `<h1>`.

godai78:
I was inspired by this:  
  
By default this feature is configured to support `<h2>`, `<h3>` and `<h4>` elements which are named: “Heading 1”, “Heading 2” and “Heading 3”, respectively. The rationale behind starting from `<h2>` is that `<h1>` should be reserved for the [page’s main title](https://ckeditor.com/docs/ckeditor5/latest/features/title.html) and the page content will usually start from `<h2>`.  
  
From here: https://ckeditor.com/docs/ckeditor5/latest/features/headings.html  
  
It can be changed easily.

AnnaTomanek:
I fixed this to make it consistent with the existing example. However, it makes me wonder whether all the examples do not really conflict (logically) with the "Heading levels" section that we have at the beginning. Because we first explain why by default we start from `<h2>` and then all the examples actually change that and start from `<h1>`.

Anyway, this was pre-existing so we can fix it later if we want.

Still, the "Common API" section is missing information about `HeadingButtonsUI`, `ParagraphButtonUI` and the individual toolbar buttons. [**@Reinmar**](https://github.com/Reinmar) it would be cool if you could help with this part :wink:

AnnaTomanek (approved):
Looks fine now, the h1 vs h2 stuff can be reconsidered later.

PR #8286, opened 2020-10-19
Exposed API for integration with TableClipboard.

niegowski:

Internal (table): Exposed API for integration with TableClipboard.

---

### Additional information


jodator (approved):
I've asked @scofalik to double-check if we couldn't use commands here.

Other than that, it's just refactoring AFAICS so ready to merge.

scofalik:
If @niegowski confirms we can rewrite this to commands and you both agree it will be prettier in the table plugin, then I am okay.

PR #8290, opened 2020-10-20
Editing and poor UI layer of the HTML embed feature

### LINKED ISSUE #8204: Raw HTML embedding

An idea of the feature is the ability to paste any HTML markup to the editor that will be displayed as a widget in the editing area and the raw HTML in the output data.

View:

```html
<p>Paragraph</p>
<div class="ck-widget">
<!-- HTML Markup -->
</div>
```

`editor.getData()`

```html
<p>Paragraph</p>
<!-- HTML Markup -->
```

Most probably, in the view, we will generate a preview that the user will not be able to interact with.

---

pomek:

Feature (html-embed): Initial implementation. Closes #8204.

Internal (theme-lark): Styles for the HTML embed feature. See #8204.

---

### Additional information

- ~Requires changes from https://github.com/ckeditor/ckeditor5/pull/8243.~
- Misses icons. I used placeholders. It will be updated when proper UI will be implemented. (#8301)


pomek:
Embedding a manual test HTML into the editor:

image

packages/ckeditor5-html-embed/lang/contexts.json
```diff
@@ -0,0 +1,5 @@
+{
```
Reinmar, line 1: Let's remember to review these in the second PR that's coming.

Reinmar:
I've got ~10 followups to report (probably I'll batch them somehow), but I'll do that tomorrow. Nothing critical there – mainly renaming things so I didn't want to do that now as it might create conflicts in your next PR.

From the important things:

*   We'll need to clean up the logic responsible for handling the previews (the toggle button). This is quite messy now and polluted even the button's `execute` listener with some `clicks()`. 
*   Editing an embed does not work - the preview is not updated. I suspect that it may be not implemented yet in this PR so I skipped that too.
*   The model cannot be updated on every change in the textarea as this will flood the collaboration and break the undo.

Other than that, most of the code is :+1:

PR #8292, opened 2020-10-20
Use `link.defaultProtocol` when auto-linking an URL,

### LINKED ISSUE #8079: Use `link.defaultProtocol` when auto-linking an URL

The `defaultProtocol` configuration option is used in the UI only. I think that this might be used for auto-link probably as well as the use-case is pretty similar.

OTOH we should think that through as I'm unsure whether this should be tied or should we have an option to disable it for auto-link or UI only. I see this as either:

1.  If `defaultProtocol` is set use the same logic for auto-link as in link UI.
2.  Another (?) configuration option or object for `defaultProtocol` that could control UI and auto-link behavior separately. If set as `String` would behave as option 1 but providing an object `{ autoLink: 'http://', ui: 'https://' }` would allow for a fine-grined control.

As for now, option 2, look like overly complicated and 1 looks like a way to go.

Original issue: https://github.com/ckeditor/ckeditor5/issues/7909#issuecomment-692173244.1

---

pkwasnik:
add mailto to email link when there is no defaultProtocol

Fix (link): Use \`link.defaultProtocol\` when auto-linking an URL. Closes #8079.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-link/src/autolink.js
+++ b/packages/ckeditor5-link/src/autolink.js
@@ -10,6 +10,7 @@
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import TextWatcher from '@ckeditor/ckeditor5-typing/src/textwatcher';
 import getLastTextLine from '@ckeditor/ckeditor5-typing/src/utils/getlasttextline';
+import { isEmail } from './utils';
 
 const MIN_LINK_LENGTH_WITH_SPACE_AT_END = 4; // Ie: "t.co " (length 5).
 
@@ -48,9 +49,6 @@ const URL_REG_EXP = new RegExp(
 	')$', 'i' );
 
 const URL_GROUP_IN_MATCH = 2;
-
-// Simplified email test - should be run over previously found URL.
-const EMAIL_REG_EXP = /^[\S]+@((?![-_])(?:[-\w\u00a1-\uffff]{0,63}[^-_]\.))+(?:[a-z\u00a1-\uffff]{2,})$/i;
 
 // The regex checks for the protocol syntax ('xxxx://' or 'xxxx:')
 // or non-word characters at the beginning of the link ('/', '#' etc.).
@@ -245,10 +243,6 @@ function getUrlAtTextEnd( text ) {
 	const match = URL_REG_EXP.exec( text );
 
 	return match ? match[ URL_GROUP_IN_MATCH ] : null;
-}
-
-function isEmail( linkHref ) {
-	return EMAIL_REG_EXP.exec( linkHref );
 }
 
 function isLinkAllowedOnRange( range, model ) {
--- a/packages/ckeditor5-link/src/autolink.js
+++ b/packages/ckeditor5-link/src/autolink.js
@@ -52,6 +52,10 @@ const URL_GROUP_IN_MATCH = 2;
 // Simplified email test - should be run over previously found URL.
 const EMAIL_REG_EXP = /^[\S]+@((?![-_])(?:[-\w\u00a1-\uffff]{0,63}[^-_]\.))+(?:[a-z\u00a1-\uffff]{2,})$/i;
 
+// The regex checks for the protocol syntax ('xxxx://' or 'xxxx:')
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-link/src/linkui.js
```diff
@@ -21,8 +21,9 @@ import LinkActionsView from './ui/linkactionsview';
 
 import linkIcon from '../theme/icons/link.svg';
 
+// The regex checks for the protocol syntax ('xxxx://' or 'xxxx:')
+// or non-word characters at the beginning of the link ('/', '#' etc.).
 const protocolRegExp = /^((\w+:(\/{2,})?)|(\W))/i;
-const emailRegExp = /[\w-]+@[\w-]+\.+[\w-]+/i;
 const VISUAL_SELECTION_MARKER_NAME = 'link-ui';
```
jodator, line 27: I can see that we don't have this in our code style, unfortunately.

The rule is to sue `CAPS_LOCK` for such constants. So it should be `PROTOCOL_REG_EXP`.

However, this is duplicated also in the other place :point\_up: .

packages/ckeditor5-link/src/autolink.js
```diff
@@ -222,9 +224,12 @@ export default class AutoLink extends Plugin {
 
 		// Enqueue change to make undo step.
 		model.enqueueChange( writer => {
-			const linkHrefValue = isEmail( url ) ? `mailto:${ url }` : url;
+			const defaultProtocol = this.editor.config.get( 'link.defaultProtocol' );
+			const protocol = isEmail( url ) ? 'mailto:' : defaultProtocol;
+			const isProtocolNeeded = !!protocol && !PROTOCOL_REG_EXP.test( url );
+			const parsedValue = url && isProtocolNeeded ? protocol + url : url;
```
jodator, line 230: I didn't notice that before - this whole logic is duplicated in \`linkui.js\`.

I think that we can extract this logic instead of \`isEmail()\`.

packages/ckeditor5-link/src/utils.js
```diff
@@ -135,3 +138,7 @@ export function isImageAllowed( element, schema ) {
 
 	return element.is( 'element', 'image' ) && schema.checkAttribute( 'image', 'linkHref' );
 }
+
+export function isEmail( linkHref ) {
+	return EMAIL_REG_EXP.test( linkHref );
+}
```
jodator, line 144: A bit PITA, but you can get used to it :)

Every "public" function or class (so here any exported one) needs at least a brief description. Check this file or other similar for inspiration. In most cases, descriptions by PK or Olek would be a good example of such description.

Anyway, you don't have to strees too much on this either as we always double check this on review :smiley:

packages/ckeditor5-link/tests/linkui.js
```diff
@@ -1321,6 +1321,19 @@ describe( 'LinkUI', () => {
 				} );
 			} );
 
+			it( 'should detect an email on submitting the form and add "mailto:" protocol automatically to the provided value ' +
+				'even when defaultProtocol is undefined', () => {
```
jodator, line 1325: :+1: Good catch.

jodator (changes requested):
Good catch with a missing tets :+1: 

I think that extracting the whole logic for default protocol handling will be better and will provide a common point if we would like to change this behavior in the future.

packages/ckeditor5-link/src/utils.js
```diff
@@ -154,16 +154,16 @@ export function isEmail( value ) {
 }
 
 /**
- * Returns an URL with protocol prefix (if applicable).
+ * Adds protocol prefix to the specified link if doesn't contain it already
+ * and there is 'defaultProtocol' config provided or link is an email.
```
jodator, line 158:
```suggestion
 * Adds protocol prefix to the specified `link` when:
 *
 * * it doesn't contain it already, and there is a {@link module:link/link~LinkConfig#defaultProtocol `defaultProtocol` }
 * config value provided
 * * or the link is an email address
 *
```

jodator (approved):
Thanks! I've re-worded this a bit so it is easier to read for me.

PR #8293, opened 2020-10-20
Non-tagged elements will not blow up the list style feature

### LINKED ISSUE #8262: With ListStyles enabled, modeling throws TypeError if a list is preceded by non-tagged content

We've got a variety of legacy HTML data created with older rich text editors, some of which contains elements that aren't surrounded by element tags. For example:
```
If there are missing paragraph or other tags prior to a list
<ol>
    <li>Item one</li>
</ol>
Then the editor won't model
```
1. Given the above source HTML data
2. Load an editor with the ListStyles plugin enabled

### ✔️ Expected result

The content should be modeled and editable by the editor.

### ❌ Actual result

A `TypeError` occurs:
```
TypeError: Cannot read property 'getAttribute' of null
    at Tm (ckeditor.js:5)
    at Pm.setAttribute (ckeditor.js:5)
    at Yp.t.on.priority (ckeditor.js:5)
    at Yp.fire (ckeditor.js:5)
    at Yp._convertItem (ckeditor.js:5)
    at Yp._convertChildren (ckeditor.js:5)
    at Yp.<anonymous> (ckeditor.js:5)
    at Yp.fire (ckeditor.js:5)
    at Yp._convertItem (ckeditor.js:5)
    at Yp._convertChildren (ckeditor.js:5)
```

If the ListStyles plugin is disabled, then the editor functions as expected.

## 📃 Other details

* Browser: Chrome 86
* OS: macOS 10.15
* CKEditor version: 22.0.x
* Installed CKEditor plugins: Bold, Italics, List, ListStyles
* Build ID: eom5f2mlp9sw-494y9pll8mj2

pomek:

Fix (list): The list conversion will not throw an error if the list element is being surrounded by the non-tagged elements (text nodes). Closes #8262.

### DIFF
--- a/packages/ckeditor5-list/src/liststyleediting.js
+++ b/packages/ckeditor5-list/src/liststyleediting.js
@@ -213,7 +213,7 @@ function upcastListItemStyle() {
 		dispatcher.on( 'element:li', ( evt, data, conversionApi ) => {
 			const listParent = data.viewItem.parent;
 			const listStyle = listParent.getStyle( 'list-style-type' ) || DEFAULT_LIST_TYPE;
-			const listItem = data.modelRange.start.nodeAfter;
+			const listItem = data.modelRange.start.nodeAfter || data.modelRange.end.nodeBefore;
 
 			conversionApi.writer.setAttribute( 'listStyle', listStyle, listItem );
 		}, { priority: 'low' } );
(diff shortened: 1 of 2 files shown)


mlewand (approved):
LGTM

Issue #7704, opened 2020-07-24
Buttons separators in RWD mobile should not be full height


oleq:
## 📝 Provide a description of the improvement

[We changed the way separators look in the toolbars](https://github.com/ckeditor/ckeditor5/issues/7493) to better integrate with balloons but we forgot about RWD forms (e.g. link):

![](https://user-images.githubusercontent.com/1099479/88395376-42200b80-cdc1-11ea-99ac-cd163a8f6b32.png)

These styles should also be updated if possible.

cc @panr 

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

PR #8298, opened 2020-10-20
Fixing showing button separators in inappropriate places

### LINKED ISSUE #8296: Button separators should follow every button in the responsive form
Follow-up of #7704.

1.  http://fake.ckeditor.com:8125/ckeditor5/tests/manual/all-features.html
2.  Create a link.

### ✔️ Expected result

The separator should only be displayed after the save button.

### ❌ Actual result

This especially becomes a problem when the form gets some padding:

## 📃 Other details

*   OS: master

---

pkwasnik:

Internal (theme-lark): Fixing showing button separators in inappropriate places. Closes #8296.

---

### Additional information

Follow-up of [#7704](https://github.com/ckeditor/ckeditor5/issues/7704).

### DIFF
--- a/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/responsive-form/responsiveform.css
+++ b/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/responsive-form/responsiveform.css
@@ -69,7 +69,9 @@
 					border-right: 1px solid var(--ck-color-base-border);
 				}
 			}
+		}
 
+		& > .ck-button:nth-last-child(2) {
 			&::after {
 				border-right: 1px solid var(--ck-color-base-border);
 			}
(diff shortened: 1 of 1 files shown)


jodator (changes requested):
Minor fix needed - I've tested it a bit differently than Olek, so I caught one more case.

I've tested the proposed solution by setting border width

```postcss
& .ck.ck-responsive-form > .ck-button:nth-last-child(1),
& .ck.ck-responsive-form > .ck-button:nth-last-child(2) {
    /* other styles */

    &::after {
        border-right: 9px solid var(--ck-color-base-border);
    }
}
```

This is what I've got:



I think that the style for `::after` should be extracted from the media-phone button styling so it only applies to the first button (similarly as it is done for `.vertical-form` class):

```css
& > .ck-button:nth-last-child(2)::after {
    border-right: 1px solid var(--ck-color-base-border);
}
```

PR #8322, opened 2020-10-22
#8321: Added a label and and outline to the html embed widgets to improve contrast.

### LINKED ISSUE #8321: The html embed should have an outline and an label for better distinguishability

ATM an empty embed looks like this

but it could be like this

---

oleq:

Internal: Added a label and an outline to the HTML embed widget to improve contrast. Closes #8321.

---

### Additional information

There's a problem with the widget type around button. I moved if by some offset to avoid collisions but it could be that in some languages "HTML snippet" will translate to some long string and it will overlap with the button. I don't see any easy way to anticipate this, though.


Reinmar:
I also added a fix for #8331 as it was just a copy-paste from media embed.

Issue #8263, opened 2020-10-15
CI engine tests are failing randomly


mlewand:
## Provide a description of the task

This is a reiteration of #6655.

We're keep getting the following warning:

```
Detected `console.warn()`:', 'Selection change observer detected an infinite rendering loop.
```

It also results with a incomplete code coverage and test suite fail. This renders our CI unstable lately.

Last week it caused 5 failed CI builds (I'm considering `master` branch alone):

*   https://travis-ci.org/github/ckeditor/ckeditor5/jobs/735863183
*   https://travis-ci.org/github/ckeditor/ckeditor5/jobs/735683309
*   https://travis-ci.org/github/ckeditor/ckeditor5/jobs/734507719
*   https://travis-ci.com/github/cksource/ckeditor5-integration-ci/jobs/396856902 (was restarted)
*   https://travis-ci.org/github/ckeditor/ckeditor5/jobs/733805236

mlewand:
Something that I'd like to know is which exactly test cases are causing the fail - this should help us better understand what is going on.

I'd suggest wrapping `console.warn` with a proxy that will check whether the console message is equal to `'Selection change observer detected an infinite rendering loop.'` if true then it should throw an exception.

This will tell us which test cases are failing. Once we have this we'll ofc revert this.

I see no problem with the fact that it will _fasely_ fail engine tests - missing CC marks the build as a fail anyway - so no change there.

Reinmar:
> I'd suggest wrapping `console.warn` with a proxy that will check whether the console message is equal to `'Selection change observer detected an infinite rendering loop.'` if true then it should throw an exception.

It could be a good idea _if this error is thrown inside a test_. If it's thrown outside (after a test ends) it will tell us nothing. But it's worth trying.

Testing this locally was a nightmare because it was hard to trigger. If it's still the same issue, we may need to remove these tests or somehow intercept that inf-render-loop to make them pass despite that warning. We spent already quite a lot of time trying to debug this and it does not seem reasonable to spend any more time trying to fix this for real. At this point, hacking them so they pass will be ok too, IMO.

Reinmar:
Another one: https://travis-ci.org/github/ckeditor/ckeditor5/jobs/736187350

Reinmar:
![](https://user-images.githubusercontent.com/156149/102094221-5c1ac300-3e22-11eb-8b20-be6f68e757bc.png)

Didn't we add some piece of code that should help us identify which test fails? cc @mlewand

[...]

PR #8325, opened 2020-10-23
Added a code that will help identify the source of infinite selection loop

mlewand:

Tests (engine): Added the detection for infinite selection rendering loop. See #8263.

---

### Additional information

This is the first step that might show us what is wrong with the test. The trick is to proxy `console.warn` so that it fails if infinite selection loop message is detected (see https://github.com/ckeditor/ckeditor5/issues/8263#issuecomment-708964548).

It's acceptable that it will cause the entire test run to fail - we get a fail from CI anyway (due to lacking CC if infinite rendering is happening).


pomek (approved):
LGTM. Let's wait for CI :D

PR #8330, opened 2020-10-23
Fixing html embed covering image inserted above with align.

### LINKED ISSUE #8324: Inserting image above html embed and changing its align will result in increasing html embed height

1.  insert html embed
2.  insert image above html embed
3.  change image align to left or right

### ✔️ Expected result

Image change align, but html embed will stay unchanged.

### ❌ Actual result

HTML embed increase its height covering image.

## 📃 Other details

*   Browser: Chrome, Firefox
*   OS: Windows
*   CKEditor version: 23.0.0

---

pkwasnik:

Fix (html-embed): Fixing html embed covering image inserted above with align. Closes #8324.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-html-embed/theme/htmlembed.css
+++ b/packages/ckeditor5-html-embed/theme/htmlembed.css
@@ -8,6 +8,7 @@
 	/* Give the embed some air. */
 	margin: 1em auto;
 	position: relative;
+	clear: both;
 
 	/* The switch mode button wrapper. */
 	& .raw-html-embed__buttons-wrapper {


pomek (approved):
LGTM.

pomek:
Merged as internal because the plugin is introduced in the current iteration so, in fact, we don't fix bugs between versions but the initial implementation.

Reinmar:
I'm not sure this is the right solution. It makes the float to just not work. While using `overflow:hidden` or the new smart `display:sth` would avoid the overlap. cc @oleq

Reinmar:
PS. I think that this is a good hotfix, but in the next iteration we could work on a better solution.

oleq:
There are 2 parts to this story:

1.  The image style feature is wrongly configured in this sample. Presentational styles require image width to be set to something sane as per https://ckeditor.com/docs/ckeditor5/latest/features/image.html#presentational-styles. So when the image gets some realistic `width` (e.g. 30%), it looks like this:



1.  Then the only missing piece is `display: flow-root` on the raw HTML widget:

oleq:
I created a follow-up issue for this #8332.

PR #8336, opened 2020-10-23
Docs: new Markdown output demo.

godai78:
A new, functional demo for the Markdown output guide.

Also adding "related features" block.


Mgsy (approved):

packages/ckeditor5-markdown-gfm/docs/_snippets/features/markdown.html
```diff
@@ -1,4 +1,24 @@
...
+
+<h2>Editing with markdown output</h2>
+<p>The <a href="https://ckeditor.com">CKEditor 5 WYSIWYG</a> editor lets the user utilize this flexible yet simple markup language, in the GitHub flavor. The output covers the most important features, like <a href="https://ckeditor.com">links</a>, <strong>different</strong> kinds of <em>emphasis</em>, the <code>inline code formatting</code> or the code blocks:</p>
+
+<pre>
+<code class="language-css">
+	p {
+		text-align: center;
+		color: red;
+	  }
```
Mgsy, line 14: Actually, I think it will display better if we remove the indent, so it won't have an unnecessary indentation in the code block.

```suggestion
p {
  text-align: center;
  color: red;
}
```

packages/ckeditor5-markdown-gfm/docs/_snippets/features/markdown.js
```diff
@@ -9,13 +9,15 @@ import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor'
 
 import ArticlePluginSet from '@ckeditor/ckeditor5-core/tests/_utils/articlepluginset';
 import EasyImage from '@ckeditor/ckeditor5-easy-image/src/easyimage';
```
AnnaTomanek, line 11: It is a bit strange to have Easy Image here but no image feature in the toolbar. The editor itself contains the image toolbar configuration so why don't we enable it?

godai78: Do we support markdown for images? If so, then we may. If not, then it's not really necessary, but I'd rather not tackle with the settings as of yet myself.

AnnaTomanek: It's quite easy to check it out - just drag an image to this sample and see the result. And then you can have a look at other samples with support for this feature, and finally, ping @Mgsy for a review/help with the updated settings. Sounds like a plan, no? :wink:

godai78: I'll try to do it, though I see little purpose in that, to be honest.

Mgsy: I've added the `imageUpload` button to the toolbar.

AnnaTomanek: Thanks, @Mgsy! And why so little faith, @godai78? Inserting images is something that most users expect the editor to support. If we support this with proper Markdown output (and we do), I see no reason to hide it. Does it make sense to you?

AnnaTomanek (approved):

packages/ckeditor5-markdown-gfm/docs/features/markdown.md
```diff
@@ -12,26 +12,40 @@ editor.setData( 'This is **bold**.' );
...
+
+The CKEditor 5 instance below is configured to output GitHub Flavored Markdown. Use the editor to create your content and see the Markdown output displayed as you type below the editor.
+
+{@snippet features/markdown}
+
+## Related features
+
+Some other ways to output the edited content include:
+
+* {@link features/export-word Export to Word} &ndash; Generate editable `.docx` files out of your editor-created content.
```
AnnaTomanek, line 36: These need to be replaced with absolute URLs as agreed with the CKE5 team since they cause the CI to fail.

AnnaTomanek (approved):
The demo is now much nicer, thanks! 👏

PR #8338, opened 2020-10-25
Content styles will display new plugins

### LINKED ISSUE #8217: List new plugins during the release

AFAIR we have already a logic that detects new plugins for the `npm run docs:content-styles` command.

It would be great to list the new plugins during the release so that we can pass on new plugin list to be added to the added to the online builder.

---

pomek:

Internal: The script that generates the content styles stylesheet will display a table with new plugins that were included in the build. Closes #8217.

---

### Additional information

Generating content styles without committing changes:

image

The same, with a commit:

image

No changes at all:

image


mlewand (approved):
One minor detail is that he plugin names are capitalized (rather than preserving the original case) - but that's not a show stopper. Other than that works like a charm 👍

PR #8346, opened 2020-10-26
Added shorthand `focus()` method in generic `Editor` class

### LINKED ISSUE #714: Implement `editor.focus()`
## 🐞 Is this a bug report or feature request? (choose one)

- Feature request

## ✅ Expected result

For implementors, I think a common requirement is moving the focus to the editor either on startup or when doing actions outside of it. I was facing this need.

My first approach for it was trying to use the main API entrypoint for it, the editor instance, and call `editor.focus()`. This is not available.

## ❎ Actual result

Currently, I'm using the following: `editor.element.focus()`, which touches the native web API. I assume this worked just because I'm using the ballon editor and my original div is there. I would be even more confused if I would have to do it with other creators that replace my element.

psmyrek:

Feature (core): Introduced the `focus()` method in the base `Editor` class. Closes #714.

---

### Additional information

The `editor.focus()` method is just a shorthand for `editor.editing.view.focus()`.

### DIFF
--- a/packages/ckeditor5-ui/docs/framework/guides/deep-dive/focus-tracking.md
+++ b/packages/ckeditor5-ui/docs/framework/guides/deep-dive/focus-tracking.md
@@ -73,13 +73,13 @@ To spice things up even more, you should also know `isFocused` will change when 
 Say, for instance, your application wants to focus an editable area of CKEditor 5 when a certain action is executed (e.g. a button is clicked). To do that, use the {@link module:engine/view/view~View#focus `focus()`} method of the editing view:
 
 ```js
-editor.editing.view.focus();
+editor.focus();
 ```
 
 This snippet focuses an editable that has the selection. If the editor has not been focused yet, this will focus the very first editable. If an editor has multiple editing roots and the user was editing content, focus will be brought back where the user left off.
 
 <info-box>
-	Focusing an editor does not change its selection. If you want to focus an editor and move the caret to a specific position, you should call `editor.editing.view.focus()` first and then use the {@link module:engine/model/writer~Writer#setSelection `setSelection()`} method of the {@link framework/guides/architecture/editing-engine#model model writer} to change the selection.
+	Focusing an editor does not change its selection. If you want to focus an editor and move the caret to a specific position, you should call `editor.focus()` first and then use the {@link module:engine/model/writer~Writer#setSelection `setSelection()`} method of the {@link framework/guides/architecture/editing-engine#model model writer} to change the selection.
 </info-box>
 
 ## Focus in the editor UI
--- a/packages/ckeditor5-ui/docs/framework/guides/deep-dive/focus-tracking.md
+++ b/packages/ckeditor5-ui/docs/framework/guides/deep-dive/focus-tracking.md
@@ -70,13 +70,13 @@ To spice things up even more, you should also know `isFocused` will change when 
 
 ### How to focus the editor?
 
-Say, for instance, your application wants to focus an editable area of CKEditor 5 when a certain action is executed (e.g. a button is clicked). To do that, use the {@link module:engine/view/view~View#focus `focus()`} method of the editing view:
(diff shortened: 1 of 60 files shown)


mlewand (changes requested):
the CI fails with:

```
WARN: 'Detected `console.warn()`:', 'There is no selection in any editable to focus.'
Chrome 86.0.4240.111 (Linux x86_64) Editor focus() should call view's focus() method FAILED
	Error: Detected `console.warn()`:
	    at console.<computed> [as warn] (webpack:///./build/.automated-tests/entry-point.js?:40:11)
	    at View.focus (webpack:///./packages/ckeditor5-engine/src/view/view.js?:438:33)
	    at Function.invoke (node_modules/sinon/pkg/sinon.js:1922:47)
	    at View.focus (node_modules/sinon/pkg/sinon.js:2178:26)
	    at TestEditor.focus (webpack:///./packages/ckeditor5-core/src/editor/editor.js?:19:4566)
	    at Context.eval (webpack:///./packages/ckeditor5-core/tests/editor/editor.js?:577:11)
```

https://travis-ci.org/github/ckeditor/ckeditor5/builds/738967351

psmyrek:
CI build has been fixed by explicitly setting `editor.editing.view.document.isFocused = true` in the unit test, because the only thing this particular test is checking is that whether it calls the `focus()` method on the `view` - regardless of its real implementation details.

mlewand:
LGTM, just one last thing that we should do in this PR is to replace existing `editor.editing.view.focus()` calls with the new, nicer form in our codebase.

packages/ckeditor5-ui/docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -73,13 +73,13 @@ To spice things up even more, you should also know `isFocused` will change when
 Say, for instance, your application wants to focus an editable area of CKEditor 5 when a certain action is executed (e.g. a button is clicked). To do that, use the {@link module:engine/view/view~View#focus `focus()`} method of the editing view:
```
mlewand, line 73: Please, be very careful with the context when making a bulk changes like that.

For instance this line no longer makes sense. It says:

To do that, use the {@link **module:engine/view/view~View#focus \`focus()\`**} method of the **editing view**:

This is no longer relevant after your change.

packages/ckeditor5-ui/docs/framework/guides/deep-dive/focus-tracking.md
```diff
@@ -73,13 +73,13 @@ To spice things up even more, you should also know `isFocused` will change when
...
 
 ```js
-editor.editing.view.focus();
+editor.focus();
 ```
 
 This snippet focuses an editable that has the selection. If the editor has not been focused yet, this will focus the very first editable. If an editor has multiple editing roots and the user was editing content, focus will be brought back where the user left off.
```
mlewand, line 79: Not only method name/path should be changed, but sometimes change like this will require you to adjust the text of the documentation.

For instance, since the changed listing shows just a method call, rather than "more complex" method accessing, how about changing the beginning of this line to:

> This method focuses

?

mlewand (changes requested):
Please check the context of remaining changes.

packages/ckeditor5-ui/docs/framework/guides/deep-dive/focus-tracking.md
psmyrek: Done, thanks for vigilance 👍🏻

packages/ckeditor5-ui/docs/framework/guides/deep-dive/focus-tracking.md
psmyrek: Done.

mlewand (approved):
LGTM, minor adjustment in one guide.

PR #8316, opened 2020-10-22
Docs paste plaintext

godai78:
Closes cksource/ckeditor5-internal#414.

Adding plain text guide with demo. Presents the plain text paste and the special keyboard shortcut plain text paste with style inheritance.

Adding related features section to Paste plain text, Paste from Word, Paste from Google Docs.

Related features placeholder for raw HTML paste.

PR #8349, opened 2020-10-27
Docs pastefromoffice demo

### LINKED ISSUE #8181: The Paste from Word demo could have more features
I was trying to check how well our list styles handle Paste from Word and was redirected to a different sample:

Screen Shot 2020-09-29 at 11 01 47 AM

In the recommended "Paste from Word" sample, list styles are not available:

Screen Shot 2020-09-29 at 11 01 54 AM

godai78:
Closes #8181 

New demos for Paste from Word and Paste from Google Docs guides:

*   list styles added to the editor,
*   new placeholder texts to make the demos look more human-appealing.


godai78:
This should ONLY be merged after #8316 is merged, otherwise it will generate errors.

AnnaTomanek:
> This should ONLY be merged after [#8316](https://github.com/ckeditor/ckeditor5/pull/8316) is merged, otherwise it will generate errors.

This is only because the approach that you chose here is wrong.

PRs should be atomic and as independent of one another as possible. This means that if you are adding a new guide, and you want to link to it from other guides, you should add these new links in the same PR that introduces the guide. Otherwise you have no way of really testing it, the CI fails and you create an unnecessary dependency that can be easily avoided.

Also, HTML embed is not a feature that is related to the "pasting" category guides, and especially not with paste from Office features. Paste from Office stuff is intended to be used with business documents and in a corporate environment. HTML embed is an experimental feature for advanced developers/webmasters and should not really be advertised in the business/document setting.

godai78:
OK.

godai78:
As promised.

AnnaTomanek:
It looks like there is some workflow misunderstanding here.

The links to the paste plain text guide should NOT be a part of this PR. At all. They should be transparent to this PR because they should be added as part of the PR creating the paste plain text feature, and then merged to `stable`. As a result, when this PR (#8349) gets merged into `stable`, they will simply be there, period. No additional thing is needed to be done.

If you need these changes sooner (for example, when you see a conflict with the `stable` branch because you changed the same section and GitHub is unable to merge it automatically, you can merge `stable` into this PR first.

godai78:
I got lost. I don't even remember what was added where. Lest just rever the last commit, merge it and once everything is finally in stable I'll just create a new branch and add it.

AnnaTomanek:
We have decided to change the approach. Another PR will replace this one.

PR #7959, opened 2020-08-27
Fix #7958 by checking for valid `firstChild`.

tomi-bigpi:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Fix (paste-from-office): Fix list paste from Google Docs with empty item. Closes #7958.

### Additional information
Note that I didn't include an HTML file named `empty-list-item.html` under `packages/ckeditor5-paste-from-office/tests/_data/paste-from-google-docs/lists/empty-list-item` as I couldn't find any references to the use in the other fixtures that had that corresponding file.

PR #8354, opened 2020-10-27
Pasting a list into another list (fix list-style attr)

### LINKED ISSUE #8160: Pasting list inside other list results in two list types on the same level

1.  Create a two-level list.
2.  Set level two style to bulleted list style (ie circles).
3.  Create a second list with numbered list styles (ie roman).
4.  Select & copy the second list.
5.  Put caret inside the first list item on level two of the first list.
6.  Paste.

### ✔️ Expected result

The styles are set to the same list style & type on every list item from level two.

### ❌ Actual result

The items pasted after the selected list item have their styles retained.

You cannot have this effect using UI and changin list styles as we change all items on the same level when changing list style of one item (second part of the above GIF).

## 📃 Other details

*   CKEditor version: master @ iteration 36
*   Installed CKEditor plugins: ListStyles

---

pomek:
#7959 # Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Fix (list): List styles will be inherited correctly when pasting a list into another list. Closes #8160.

---

### Additional information

_I know, I used the best solution for detecting that content comes from pasting… How could we improve that?_

### DIFF
--- a/packages/ckeditor5-list/package.json
+++ b/packages/ckeditor5-list/package.json
@@ -10,7 +10,6 @@
     "ckeditor5-plugin"
   ],
   "dependencies": {
-    "@ckeditor/ckeditor5-clipboard": "^23.0.0",
     "@ckeditor/ckeditor5-core": "^23.0.0",
     "@ckeditor/ckeditor5-engine": "^23.0.0",
     "@ckeditor/ckeditor5-paragraph": "^23.0.0",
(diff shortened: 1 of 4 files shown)


jodator:
@pomek I thought it would require extracting the `if()` to separate method but it turned out that it has many drawbacks: 00610fd.

*   There was missing `wasFixed = true` and the condition was causing infinite fix-loops as the check caught also cases of the same list types.
*   The logic was _slightly_ different from `shouldInheritListType()` and the `if( prevSibling &&` was duplicated from that function - so I've inlined those checks in new function.
*   Then it was just moving the method around.

jodator:
Additionally, I've moved the check from the `fixListStyleAttributeOnListItemElements()` so the functions will not be re-created on every `fixListStyleAttributeOnListItemElements()` call.

jodator (approved):
LGTM ;)

PR #8359, opened 2020-10-27
The old view element should be properly unbound after reconversion is done

### LINKED ISSUE #8358: Old view element should be unbound after reconversion is done

When reconversion happens, the old view element is removed but it is not unbound, because we want to use it to get its children (without reconverting those children).

However, this element is never unbound which is incorrect. This causes an error, because markers that were created on that view element will still be bound to the old view element.

scofalik:

Fix (engine): The old view element should be properly unbound after reconversion is done. Closes #8358.

### DIFF
--- a/packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
+++ b/packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
@@ -38,6 +38,8 @@ import { expectToThrowCKEditorError } from '@ckeditor/ckeditor5-utils/tests/_uti
 import { StylesProcessor } from '../../src/view/stylesmap';
 import DowncastWriter from '../../src/view/downcastwriter';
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
```diff
@@ -192,6 +194,53 @@ describe( 'DowncastHelpers', () => {
 					expectResult( '<div class="is-classy"></div>' );
 				} );
 
+				it( 'should properly re-bind mapper', () => {
```
jodator, line 197:
```suggestion
				it( 'should properly re-bind mapper mappings and retain markers', () => {
```

jodator (approved):
LGTM :+1: I'll just update the test case name.

jodator:
ps.: I've also changed to `Internal` as this is a fix for the feature from the current release.

PR #8369, opened 2020-10-28
Docs: HTML embed review.

godai78:
HTML Embed review.

Related features block.

### DIFF
--- a/packages/ckeditor5-html-embed/docs/features/html-embed.md
+++ b/packages/ckeditor5-html-embed/docs/features/html-embed.md
@@ -5,11 +5,11 @@ menu-title: HTML embed
 
 # HTML embed
 
-The {@link module:html-embed/htmlembed~HtmlEmbed} plugin allows embedding an arbitrary HTML snippet in the editor. The feature is targeted at more advanced users who want to directly interact with HTML fragments.
+The {@link module:html-embed/htmlembed~HtmlEmbed} plugin allows embedding an arbitrary HTML snippet in the editor. The feature is targeted at more advanced users who want to directly interact with HTML fragments; for less demanding needs the {@link features/media-embed Media embed} feature should suffice.
 
-This feature can be used to embed any HTML code and bypass CKEditor 5's filtering mechanisms. Thanks to that it is possible to enrich content produced by CKEditor 5 with fragments of HTML that are not supported by any other CKEditor 5 feature.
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-html-embed/docs/features/html-embed.md
```diff
@@ -32,6 +32,13 @@ Use the editor below to see the plugin in action.
 
 {@snippet features/html-embed}
 
+## Related features
+
+CKEditor 5 supports a wider range of embed and paste features, including:
```
Reinmar, line 37:
```suggestion
CKEditor 5 supports a wider range of embed and paste features, including:
```

Reinmar, line 37: Rule of thumb – put blank lines the way the content is usually rendered on a page. In this case, after a paragraph, there would normally be a margin separating it from the list.

packages/ckeditor5-html-embed/docs/features/html-embed.md
```diff
@@ -90,22 +97,22 @@ ClassicEditor
 	.catch( ... );
 ```
 
-Currently, the [feature does not execute `<script>` tags](https://github.com/ckeditor/ckeditor5/issues/8326) so content that requires executing JavaScript in order to generate a preview will not show in the editor. However, other JavaScript code – e.g. used in `on*` observers and `scr="javascript:..."` attributes will be executed and therefore a sanitizer still needs to be enabled.
+Currently, the [feature does not execute the `<script>` tags](https://github.com/ckeditor/ckeditor5/issues/8326) so the content that requires executing JavaScript in order to generate a preview will not show in the editor. However, other JavaScript code – e.g. used in `on*` observers and `scr="javascript:..."` attributes will be executed and therefore a sanitizer still needs to be enabled.
```
Reinmar, line 100:
```suggestion
Currently, the [feature does not execute `<script>` tags](https://github.com/ckeditor/ckeditor5/issues/8326) so the content that requires executing JavaScript in order to generate a preview will not show in the editor. However, other JavaScript code – e.g. used in `on*` observers and `scr="javascript:..."` attributes will be executed and therefore a sanitizer still needs to be enabled.
```

These aren't specific script tags. I don't think that "the" is missing here.

packages/ckeditor5-html-embed/docs/features/html-embed.md
godai78: Corrected.

AnnaTomanek:
Added these (and some more) changes directly to `stable` so I'm closing this PR. Thanks for the corrections!

Issue #47, opened 2015-12-15
Documentation for grunt/gulp tasks


Reinmar:
Followup for #43 and #36.

I'm not sure yet where we want to keep it (in the README or in ckeditor5-design and later copy somewhere... dunno). But we definitely need to have a list of them with short descriptions what they are needed for.

Second part is a documentation of a typical workflow (based on #43).

Reinmar:
One thing that would definitely need to be clarified in ckeditor5-design is what we encountered while working on #43. Like a more high-level description of the situation. Without it the behaviour of our tools may seem to be weird.

szymonkups:
For now, I will put tasks description here. Later on we can move them to more suitable place.

## dev-install \< path|GitHub URL|name \>

###### Task description
Installs selected dependency in development mode. Dependency will be cloned into `workspace` directory (if it is not already), linked ( using symlink ) to `node_modules` and added to `package.json` inside CKEditor5 main repository. Dependency can be recognized by:
* path - local path to repository on disk
```
grunt dev-install /home/user/workspace/ckeditor5-core
```
* GitHub URL - it can be either short or full URL, also with branch information
```
grunt dev-install ckeditor/ckeditor5-core
grunt dev-install https://github.com/ckeditor/ckeditor5-core.git
grunt dev-install git@github.com:ckeditor/ckeditor5-core.git
grunt dev-install ckeditor/ckeditor5-core#branch-name
```

* name - module name used in NPM. In this situation task will try to reach NPM to obtain GitHub URL from package information and clone it to `workspace` directory
```
grunt dev-install ckeditor5-core
```
###### Task steps
1. Use either local path, GitHub URL or NPM name to recognize dependency.
1. If dependency directory does not exists - clone dependency repository into `workspace`.
1. If branch information is provided - checkout repository to desired branch.
1. Run `npm install` inside newly created repository to install all dependencies there.
1. Link repository to `node_modules` inside CKEditor5 repository.
1. Add dependency to `package.json` file inside CKEditor5 repository.
1. Install Git hooks inside dependency directory.

Reinmar:
Important note: you use the term "Plugin development mode" and later use the word "plugin" many times.  The thing you're referring to is a "package" (aka "node module").

The important of a correct naming is well visible e.g. when you work on `ckeditor5-core` which is not a plugin – it's simply a package with some modules inside.

Relevant discussion is here: https://github.com/ckeditor/ckeditor5-design/issues/92

[...]

PR #8372, opened 2020-10-28
Inserting images by pasting URL directly into editor

### LINKED ISSUE #8236: Inserting images by pasting URL directly into editor

It could be usefull to have possibility to insert image just by pasting URL to the editor like it works in media-embed.

---

pkwasnik:

Feature (image): Inserting images by pasting URL directly into editor. Closes #8236.

MINOR BREAKING CHANGE (image): There were changes in the module:image/utils~insertImage parameters:

*   removed \`writer\` parameter - now method get it by itself
*   added insertPosition parameter to allow directly specify insert position instead of using findOptimalInsertionPosition logic

---

### Additional information


packages/ckeditor5-image/docs/features/image.md
```diff
@@ -105,6 +106,16 @@ This will add a new **Insert image** dropdown in the toolbar. To open the panel
 
 {@snippet features/image-insert-via-url}
 
+## Inserting images via pasting URL into editor
+
+The {@link module:image/autoimage~AutoImage} plugin recognizes image links in the pasted content and embeds them shortly after they are injected into the document to speed up the editing. Accepted image extensions: jpg, png, gif, ico.
```
jodator, line 111:
```suggestion
The {@link module:image/autoimage~AutoImage} plugin recognizes image links in the pasted content and embeds them shortly after they are injected into the document to speed up the editing. Accepted image extensions: jpg, jpeg, png, gif, ico.
```

packages/ckeditor5-image/src/autoimage.js
```diff
@@ -0,0 +1,172 @@
...
+import Clipboard from '@ckeditor/ckeditor5-clipboard/src/clipboard';
+import LiveRange from '@ckeditor/ckeditor5-engine/src/model/liverange';
+import LivePosition from '@ckeditor/ckeditor5-engine/src/model/liveposition';
+import Undo from '@ckeditor/ckeditor5-undo/src/undo';
+import global from '@ckeditor/ckeditor5-utils/src/dom/global';
+import { insertImage } from './image/utils';
+
+// implements pattern: http(s)://(www.)example.com/path/to/resource.ext?query=params&maybe=too
+const IMAGE_URL_REGEXP =
+	/^(http(s)?:\/\/)?[\w-]+(\.[\w-]+)+[\w._~:/?#[\]@!$&'()*+,;=%-]+\.(jpg|jpeg|png|gif|ico)\??[\w._~:/#[\]@!$&'()*+,;=%-]*$/;
```
jodator, line 20: Some additional tests are needed as I couldn't find them:

*   image with query params
*   all supported extensions
*   special characters
*   case sensitivity: `.JPG` vs `.jpg`
*   some not-supported extensions ( like `svg`, `webp`, `mp3`, `exe` just in case :wink: )

For repetitive test cases we tend to make it in a loop ([example](https://github.com/ckeditor/ckeditor5/blob/9b6f41027da388be5baf5e317caac2b7828b94bf/packages/ckeditor5-link/tests/autolink.js#L228-L285)) - the important thing to remember is that the test name must be clear - in order to see which cases failed in the console. 

A bad example would be `It should work with URL (case 2)`. So in the linked example, I've used the full URL in the test name but you can use [other structures](https://github.com/ckeditor/ckeditor5/blob/e046735d04ee6c650501b610b6e82c6fb740a74a/packages/ckeditor5-engine/tests/dataprocessor/htmldataprocessor.js#L54-L67) as you wish.

jodator (changes requested):
The feature works nicely :tada:. Most of my comments are related to our technical debt but I think that we could iron something out in this PR.

The other thing is that we need more tests, especially for supported vs unsported URLs.

During the review, I've realized that we don't have integration tests for auto media embed and auto image. But I think that we're safe as those _should_ detect different URLs anyway.

Could you also report a ticket for de-duplicating AutoImageand AutoMediaEmbed and link to this issue in both plugins? It will be easier to track in the future.

The last thing is a bug of pasting a link to an image:
Peek 2020-10-29 15-22

jodator:
Hello @ckeditor/qa-team :wave:, could you take a quick look at this feature?

The manual test for this is the same as the insert an image via URL. ~If you think that testing it with other features makes sense let us know.~ I'll add it to the all features tests in a moment :D

jodator:
And after consulting with the docs team we need a separate demo for this feature. It might be a simplified editor with minimal text inside + an input with an example URL to copy.

packages/ckeditor5-image/src/autoimage.js
```diff
@@ -0,0 +1,172 @@
...
+			return;
+		}
+
+		// Position won't be available in the `setTimeout` function so let's clone it.
+		this._positionToInsert = LivePosition.fromPosition( leftPosition );
+
+		// This action mustn't be executed if undo was called between pasting and auto-embedding.
+		this._timeoutId = global.window.setTimeout( () => {
+			// Don't do anything if image element cannot be inserted at the current position (#47).
+			// Condition must be checked after timeout - pasting may take place on an element, replacing it. The final position matters.
```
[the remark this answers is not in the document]
pkwasnik: It does not happen with media embed, because on media the mediaEmbedCommand is enabled. But on image the imageCommand is not enabled, so we need to wait until image become a text before checking this. Condition will pass on image, but still will not pass e.g. in image caption.

FilipTokarski:
Checked in various browsers, seems to be ok 👍

jodator:
@FilipTokarski Thanks!

packages/ckeditor5-image/tests/autoimage.js
```diff
@@ -0,0 +1,620 @@
...
+			clock = sinon.useFakeTimers();
+		} );
+
+		afterEach( () => {
+			clock.restore();
+		} );
+
+		it( 'replaces pasted text with image element after 100ms', () => {
+			setData( editor.model, '<paragraph>[]</paragraph>' );
+			pasteHtml( editor, 'https://ckeditor.com/assets/images/header/ckeditor-5-d7348daebf.png' );
```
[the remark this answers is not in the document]
pkwasnik: Hm, I think changing  it to example.com is ok, works with tests - did it

packages/ckeditor5-image/src/autoimage.js
```diff
@@ -0,0 +1,173 @@
...
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import Clipboard from '@ckeditor/ckeditor5-clipboard/src/clipboard';
+import LiveRange from '@ckeditor/ckeditor5-engine/src/model/liverange';
+import LivePosition from '@ckeditor/ckeditor5-engine/src/model/liveposition';
+import Undo from '@ckeditor/ckeditor5-undo/src/undo';
+import global from '@ckeditor/ckeditor5-utils/src/dom/global';
+import { insertImage } from './image/utils';
+
+// implements pattern: http(s)://(www.)example.com/path/to/resource.ext?query=params&maybe=too
```
jodator, line 18:
```suggestion
// Implements the pattern: http(s)://(www.)example.com/path/to/resource.ext?query=params&maybe=too.
```

packages/ckeditor5-image/src/autoimage.js
```diff
@@ -0,0 +1,173 @@
...
+
+			return;
+		}
+
+		// Position won't be available in the `setTimeout` function so let's clone it.
+		this._positionToInsert = LivePosition.fromPosition( leftPosition );
+
+		// This action mustn't be executed if undo was called between pasting and auto-embedding.
+		this._timeoutId = global.window.setTimeout( () => {
+			// Don't do anything if image element cannot be inserted at the current position (#47).
```
jodator, line 142:
```suggestion
			// Don't do anything if image element cannot be inserted at the current position.
			// See https://github.com/ckeditor/ckeditor5/issues/2763.
```
To explain the suggestion: We have old references in the code after mergin multi-repo into a monorepo.
Additionally `#47` was an old way of linking issues in code. We use full URLs now as it is easier to navigate to the issue from the source code.
Full URLs also works better as we moved issues from all the packages issue trackers into one and GH can redirect them nicely.
Usually you will not have to worry about that :)

jodator (commented):
I've added some minor suggestions.

jodator (approved):

PR #8375, opened 2020-10-29
Typo fix in the engine element comment (to test PR)

mlewand:

Internal (engine): A simple typo fix.

---

### Additional information

The main reason for this PR is to check GH board integration.


mlewand:
PR handling works as expected, while the proposed change is not really needed as both forms are correct 🙂

Issue #7805, opened 2020-08-07
paste-from-office: Losing indent level for nested lists of different type (numbered, bullet)


mmichaelis:
## 📝 Reproduction Steps

1. Created a nested list in MS Word (Office 365) as follows:

    > * Lorem
    > * Ipsum
    >     1. Dolor
    >     2. Sit
    > * Amet

    See [Copy-Paste-Mixed-List-Issue-CKEditor.docx](https://github.com/ckeditor/ckeditor5/files/5040544/Copy-Paste-Mixed-List-Issue-CKEditor.docx) (first list shows error; second list works).

2. Copy the entire list
3. Paste it for example to [CKEditor 5 demo](https://ckeditor.com/ckeditor-5/demo/)

### ✔️ Expected result

Just as all other nested lists (even for mixed outer numbered, inner bullet) the nested indentation levels should be kept as is.

### ❌ Actual result

Indentation level of nested list is set to 0:

> * Lorem
> * Ipsum
> 1. Dolor
> 2. Sit
> * Amet

## 📃 Other details

* Browser: Chrome
* OS: Windows 10
* Office: MS Word; Office 365
* CKEditor version: Most recent (CKEditor 5 demo) as well as custom using `"@ckeditor/ckeditor5-paste-from-office": "^20.0.0"`
* Installed CKEditor plugins: `"@ckeditor/ckeditor5-paste-from-office": "^20.0.0"`

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

FilipTokarski:
Hi, thanks for the report. I confirm this bug. However it seems to occur only if you have numbered items inside bulleted list, otherwise the indentation works fine:  
  
![word3](https://user-images.githubusercontent.com/34380544/89761529-dba02a80-daee-11ea-8839-d0ed076b7441.png)
  
-  occurs also on older versions
- browser: any

pomek:
Lists in Word use the `id` property for marking the whole list as a structure. It means that nested lists have the same value of the property as their parents.

All list items have the same `id`:

![image](https://user-images.githubusercontent.com/2270764/97548592-9b569700-19cf-11eb-96f0-79a9b16d3c78.png)

However, for some, unknown reason, the first list in the attached document has mixed `id` values (a parent and a nested list are not the same).

I would say: I can reproduce the issue and I can't reproduce the issue at the same time. It depends on which list I use.

The list used for testing:

```
• Lorem
    1. Dolor
    2. Sit
[...]

PR #8390, opened 2020-11-02
Support for list style while pasting from Word

### LINKED ISSUE #8080: Make paste-from-office preserve Word list styles

Currently if you paste a list with custom bullets/numbering it's always converted to default editor styles:
0_pfo_list

   
But now that we have list styles feature, it would be nice if styling from Word would be pasted and preserved inside the editor.

pomek:

Feature (paste-from-office): Support for preventing the list styles while pasting from Word. Closes #8080.


jodator:
@ckeditor/qa-team could you guys take a look on this?

Mgsy:
@FilipTokarski, could you please test it?

FilipTokarski:
I cannot reproduce this most of the time, but sometimes list with custom styles breaks when being pasted into editor:  
word_1
  
Here's a gist with data from clipboard: https://gist.github.com/FilipTokarski/b60c888e22cef726ed132b5388b5fb02
  
I'll try to find some steps to reproduce this, but as for now it looks very strange.

FilipTokarski:
Ok, I might have found the cause of those bugs - if you just create a list with varioust styles and copy/paste it into the editor, the styles are preserved:  
word_2_good
  
If you then, inside Word, change some of those styles and copy/paste again, the list breaks:  
word_2_bad

FilipTokarski:
To reproduce this 100% of the times, when changing list styles in Word, you need to fully select all list items. If you just put the cursor on the list item and change the style, it will work ok.

pomek:
The wrong indent bug is already reported here – https://github.com/ckeditor/ckeditor5/issues/7805.

It looks like after resolving #7805, some of the issues reported above should be resolved.

packages/ckeditor5-paste-from-office/src/filters/list.js
```diff
@@ -158,30 +156,114 @@ function findAllItemLikeElements( documentFragment, writer ) {
...
+	}
+
+	const listMarker = listMarkerElement._data;
+
+	if ( listMarker === 'o' ) {
+		return 'circle';
+	} else if ( listMarker === '·' ) {
+		return 'disc';
+	// Word returns '§' instead of '■' for the square list style.
+	} else if ( listMarker === '§' ) {
```
jodator, line 216:
```suggestion
	}
	// Word returns '§' instead of '■' for the square list style.
	else if ( listMarker === '§' ) {
```

AFAICS, we comment `else` as above. Not the ideal code style for that, but should be kept in sync.

packages/ckeditor5-paste-from-office/src/filters/list.js
```diff
@@ -158,30 +156,114 @@ function findAllItemLikeElements( documentFragment, writer ) {
...
+function mapListStyleDefinition( value ) {
+	switch ( value ) {
+		case 'arabic-leading-zero':
+			return 'decimal-leading-zero';
+		case 'alpha-upper':
+			return 'upper-alpha';
+		case 'alpha-lower':
+			return 'lower-alpha';
+		case 'roman-upper':
+			return 'upper-roman';
```
jodator, line 255: Thanks, MS Word :+1:

jodator (approved):
Seeing what words produces amaze me, great job here :+1:.

My comments are mostly to the old code and the MS Word itself :D

jodator:
@pomek if I understand this correctly, those two issues reported by @FilipTokarski are to be resolved with #7805?

pomek:
@jodator, I hope :D

PR #8408, opened 2020-11-03
Code should not be copied to a new line on Enter

### LINKED ISSUE #8144: Code should not be copied to a new line on Enter

Unlike bold or italic, which people seem to expect to be copied to a new line when you press Enter, `<code>` makes no sense at all. I find it very annoying when using GH Writer and writing content like this:

I was absolutely not expecting "The" to be wrapped in `<code>`.

---

pkwasnik:

Fix (basic-styles): Code should not be copied to a new line on Enter. Closes #8144.

---

### Additional information


jodator (approved):
I was wondering for a brief time if we shouldn't add behavioral tests for this case but AFAICS we don't have such tests in other basic styles features. I think we can live with this. Also, the `copyOnEnter` should have tests for this behaviour already.

PR #8411, opened 2020-11-03
Fixed restoring selection on undoing for cases when transformed ranges are interleaved by ranges in the graveyard

niegowski:

Internal (undo): Fixed restoring selection on undoing for cases when transformed ranges are interleaved by ranges in the graveyard.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-undo/src/basecommand.js
+++ b/packages/ckeditor5-undo/src/basecommand.js
@@ -100,7 +100,14 @@ export default class BaseCommand extends Command {
 
 		for ( const rangeGroup of transformedRangeGroups ) {
 			// While transforming there could appear ranges that are contained by other ranges, we shall ignore them.
-			const transformed = rangeGroup.filter( range => !isRangeContainedByAnyOtherRange( range, allRanges ) );
+			const transformed = rangeGroup
+				.filter( range => range.root != document.graveyard )
+				.filter( range => !isRangeContainedByAnyOtherRange( range, allRanges ) );
+
+			// All the transformed ranges ended up in graveyard.
+			if ( !transformed.length ) {
+				continue;
+			}
 
 			// After the range got transformed, we have an array of ranges. Some of those
 			// ranges may be "touching" -- they can be next to each other and could be merged.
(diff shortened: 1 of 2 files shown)


scofalik:
Those changes are counter-intuitive to me. Could you write down step by step what was happening in the incorrect scenario? 

To me, it seems like it would make more sense to include all ranges, transform them, and then filter out graveyard ranges. This way it seems like you have more chances to have a correct range and also you won't end up in the range that is in graveyard after restoring (even if it was in the document at the beginning).

niegowski:
> To me, it seems like it would make more sense to include all ranges, transform them, and then filter out graveyard ranges. This way it seems like you have more chances to have a correct range and also you won't end up in the range that is in graveyard after restoring (even if it was in the document at the beginning).

But look closer at the code, all ranges are transformed then those that ended up in the graveyard are filtered out and the remaining ones are normalized (joined together if they are touching or intersecting).

> Could you write down step by step what was happening in the incorrect scenario?

Using test added in this PR we get ranges after transformation:

```
[ 0,0,1 ] - [ 0,0,1,0 ]
[ 2 ] - [ 2,0 ] $graveyard
[ 1 ] - [ 2 ] $graveyard
[ 0 ] - [ 1 ] $graveyard
[ 0,0,1,0 ] - [ 0,0,2 ]
[ 2,0 ] - [ 3 ] $graveyard
```

and after sorting with `ranges.sort( ( a, b ) => a.start.isBefore( b.start ) ? -1 : 1 )` we have:

```
[ 0,0,1 ] - [ 0,0,1,0 ] 
[ 0 ] - [ 1 ] $graveyard
[ 1 ] - [ 2 ] $graveyard
[ 2 ] - [ 2,0 ] $graveyard
[ 0,0,1,0 ] - [ 0,0,2 ]
[ 2,0 ] - [ 3 ] $graveyard
```

so while normalizing in this code:

```
for ( let i = 1; i < ranges.length; i++ ) {
	const previousRange = ranges[ i - 1 ];
	const joinedRange = previousRange.getJoined( ranges[ i ], true );

	if ( joinedRange ) {
		// Replace the ranges on the list with the new joined range.
		i--;
		ranges.splice( i, 2, joinedRange );
	}
}
```

it ends up with not properly normalized ranges. We could tuneup this normalization but it seems to be easier to filter out ranges in graveyard before trying to normalize them.

PR #8415, opened 2020-11-04
The editing root should remain focused when saving changes in the embed

### LINKED ISSUE #8318: The editing root should remain focused when saving changes in the embed

1.  Run an editor with a dynamic toolbar (inline, balloon) in http://fake.ckeditor.com:8125/ckeditor5-html-embed/tests/manual/htmlembed.html.
2.  Edit an embed.
3.  Save it.

### ✔️ Expected result

Widget is focused. The editor is focused. The toolbar remains visible.

### ❌ Actual result

The editor loses focus. The global toolbar disappears.

## 📃 Other details

*   follow-up of #8301.

---

pkwasnik:

Fix (html-embed): The editing root should remain focused when saving changes in the embed. Closes #8318.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-html-embed/src/htmlembedediting.js
+++ b/packages/ckeditor5-html-embed/src/htmlembedediting.js
@@ -166,7 +166,7 @@ export default class HtmlEmbedEditing extends Plugin {
 						// it's enough to update the model – the entire widget will be reconverted.
 						if ( newValue !== state.getRawHtmlValue() ) {
 							editor.execute( 'updateHtmlEmbed', newValue );
-							editor.focus();
+							editor.editing.view.focus();
 						} else {
 							this.cancel();
 						}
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-html-embed/src/htmlembedediting.js
```diff
@@ -166,6 +166,7 @@ export default class HtmlEmbedEditing extends Plugin {
 						// it's enough to update the model – the entire widget will be reconverted.
 						if ( newValue !== state.getRawHtmlValue() ) {
 							editor.execute( 'updateHtmlEmbed', newValue );
+							editor.focus();
```
jodator, line 169: I think that the a8410bf was on point here.

As per [this comment](https://github.com/ckeditor/ckeditor5/issues/714#issuecomment-721086621) in the PR that introduced `editor.focus()` the semantics of both is a bit different, even though it's a shortcut for the `editor.editing.view.focus()`.

IIRC - the reason behind this is that `editor.focus()` might change in the future. A similar thing is being reverted - buttons should use `editor.editing.view.focus()`.

jodator (changes requested):
As in my only comment for the PR as in general, it is good: the preferred way, AFAICS, is to use `editor.editing.view.focus()` in the UI.

jodator (approved):
Cool, thanks! :+1:

PR #8417, opened 2020-11-05
Improved a check for pasting the nesting lists in PFO

### LINKED ISSUE #7805: paste-from-office: Losing indent level for nested lists of different type (numbered, bullet)

pomek:

Fix (paste-from-office): Improved a mechanism that detects nested lists while pasting from Word. In some scenarios, pasting nested lists could produce invalid results in the editor. Closes #7805.


packages/ckeditor5-paste-from-office/src/filters/list.js
```diff
@@ -370,6 +375,14 @@ function isNewListNeeded( previousItem, currentItem ) {
 	}
 
 	if ( previousItem.id !== currentItem.id ) {
+		// See: #7805.
```
jodator, line 378:
```suggestion
		// See: https://github.com/ckeditor/ckeditor5/issues/7805.
```

packages/ckeditor5-paste-from-office/src/filters/list.js
```diff
@@ -361,6 +361,11 @@ function removeBulletElement( element, writer ) {
 // Whether previous and current item belongs to the same list. It is determined based on `item.id`
 // (extracted from `mso-list` style, see #getListItemData) and previous sibling of the current item.
 //
+// However, it's quite easy to change the `id` attribute for nested lists and it will break the list feature while pasting.
+// Let's check also the `indent` attribute. If between those two elements, the difference is equal to 1, we can assume that
+// the `currentItem` is a beginning of the nested list because lists in CKEditor 5 always starts with the `indent=0` attribute.
```
jodator, line 366:
```suggestion
// However, it's quite easy to change the `id` attribute for nested lists in Word. It will break the list feature while pasting.
// Let's check also the `indent` attribute. If between those two elements, the difference is equal to 1, we can assume that
// the `currentItem` is a beginning of the nested list because lists in CKEditor 5 always starts with the `indent=0` attribute.
```

jodator (approved):
I don't have words for the Word issues anymore... This case looks solved :+1: .

I've added full links to the issue as you need to quickly browse the GH in case you really want to understand the story behind.

I've updated one thing in the comment so hopefully, I get that one right.

Issue #8328, opened 2020-10-23
Can't save an HTML embed after pasting a text from the editor


Mgsy:
## 📝 Provide detailed reproduction steps (if any)

1.  Insert a new HTML embed/edit the existing one.
2.  Select some text in the editor.
3.  Copy it.
4.  Paste it to the HTML embed.
5.  Save changes.

### ✔️ Expected result

Changes are saved.

### ❌ Actual result

Nothing happens.

## 📃 Other details

![](https://user-images.githubusercontent.com/20988892/96963900-c6417680-1509-11eb-80f2-a35a988629df.gif)

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

pomek:
The same kind of error (missing selection on the widget element).

1. **Insert** an empty widget.
2. Enter.
3. **Insert** a second widget.
4. Click the first `textarea`.
5. Type whatever.
6. After saving changes, the second container will be modified.

![image](https://user-images.githubusercontent.com/2270764/96964870-824f7100-150b-11eb-87e7-455b06372e08.png)

pkwasnik:
Another similar bug:

1.  Insert or edit widget (do not save)
2.  Click anywhere outside the widget
3.  Try to save
4.  Nothing happen

psmyrek:
I've created a follow-up #8509 to provide a generic solution for selecting a widget, which has the `data-cke-ignore-events` attribute.

PR #8421, opened 2020-11-05
Reflecting commands state by save button.

### LINKED ISSUE #8357: Raw HTML widget buttons don't reflect commands state

Raw HTML widget buttons don't reflect commands state. When the editing command is disabled, the edit and save buttons should be disabled as well. This happens for example when track changes is turned off (the feature is not yet integrated with TC).

pkwasnik:

Fix (html-embed): Reflecting commands state by save button. Closes #8357.

---

### Additional information


packages/ckeditor5-html-embed/src/htmlembedediting.js
```diff
@@ -352,6 +353,7 @@ function createDomButton( locale, type ) {
 			label: t( 'Save changes' ),
 			class: 'raw-html-embed__save-button'
 		} );
+		buttonView.bind( 'isEnabled' ).to( command, 'isEnabled' );
```
jodator, line 356: Unfortunately, the button is destroyed (see below) here and w only use `ButtonView` to render the DOM element (by using `cloneNode()` below.

The problem with this that you can still click and execute the button. The disabled state is taken into consideration only when the button is created.

The best way to test it in the manual test is to change the editor state to ["read-only"](https://ckeditor.com/docs/ckeditor5/latest/features/read-only.html) by executing in the console:

```
editor.isReadOnly = true;
```

jodator (changes requested):
I think that we didn't check this case when implementing HTML embed so a new solution or a fix must be provided for the buttons used there.

At this moment I wonder if:
- We need to destroy the button at all - thus we would have the `ButtonView` binding working - OTOH I'm not sure if that binding it to a command will not leak memory.
- Simple listener to change button state might be enough (using VanillaJS framework ;) )
- Again, we must ensure that this will not leak the momory.

jodator:
After quick talk during sync - we didn't use `ButtonView` instances there because of memory leaks. There's a work-around to use a view [post-fixer](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_document-Document.html#function-registerPostFixer). You should be able to check which views are not in the view root anymore and run `buttonView.destroy()` to detach listeners (created by `view.bindTo()` call.

The "edit" button must be disabled as well when the command is disabled (note that there are 2 commands - one for inserting and one for editing which I wasn't aware before).

pkwasnik:
Clicking "save" will return to preview only when there are no changes (this a "feature" made on purpose). Otherwise button will not work:



About `edit` button there is a bug that is blocks it. Currently, edit command is disabled when the focus is outside of the editor. If we will link the `edit` state to that command than button will be disabled until the widget will be clicked. There is a bug open for serveral similar cases, including this one: #8328

jodator (approved):
After reading this over and over I agree that in the current state the least we can do is to disable the save button. I don't like this feature UX but let live with it for a while.

Sorry for longer review @pkwasnik, everything is OK now. According to the ticket description and later talks :+1:

PR #8438, opened 2020-11-09
Docs: adding autoformatting to code blocks.

godai78:
Adding information about the autoformatting feature to the code block guide.

### DIFF
--- a/packages/ckeditor5-code-block/docs/features/code-blocks.md
+++ b/packages/ckeditor5-code-block/docs/features/code-blocks.md
@@ -14,7 +14,7 @@ The {@link module:code-block/codeblock~CodeBlock} feature allows inserting and e
 
 ## Demo
 
-Use the code block toolbar button and the type dropdown, to insert desired code block. Alternatively, use the `` ``` `` code provided by the {@link features/autoformat autoformatting feature}.
+Use the code block toolbar button and the type dropdown to insert a desired code block. Alternatively, start the line with `` ``` `` to format it as a code block thanks to the {@link features/autoformat autoformatting feature}.
 
 {@snippet features/code-block}
 
--- a/packages/ckeditor5-code-block/docs/features/code-blocks.md
+++ b/packages/ckeditor5-code-block/docs/features/code-blocks.md
@@ -6,13 +6,15 @@ category: features
 
 {@snippet features/build-code-block-source}
 
-The {@link module:code-block/codeblock~CodeBlock} feature allows inserting and editing blocks of pre–formatted code into the WYSIWYG editor. Each code block has a [specific programming language assigned](#configuring-code-block-languages) (e.g. "Java" or "CSS") and supports basic editing tools, for instance, [changing the line indentation](#changing-line-indentation) using the keyboard.
+The {@link module:code-block/codeblock~CodeBlock} feature allows inserting and editing blocks of pre–formatted code into the WYSIWYG editor. Each code block has a [specific programming language assigned](#configuring-code-block-languages) (e.g. "Java" or "CSS") and supports basic editing tools, for instance, [changing the line indentation](#changing-line-indentation) using the keyboard. A full list of supported languages depends on the highlighter used. Refer to the [common API](#common-api) section for a default supported list.
 
 <info-box>
 	If you would like to use inline code formatting in your WYSIWYG editor, check out the {@link features/basic-styles basic text styles feature} with its support for inline `<code>` element.
 </info-box>
 
 ## Demo
+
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-code-block/docs/features/code-blocks.md
```diff
@@ -6,14 +6,16 @@ category: features
 
 {@snippet features/build-code-block-source}
 
-The {@link module:code-block/codeblock~CodeBlock} feature allows inserting and editing blocks of pre–formatted code into the WYSIWYG editor. Each code block has a [specific programming language assigned](#configuring-code-block-languages) (e.g. "Java" or "CSS") and supports basic editing tools, for instance, [changing the line indentation](#changing-line-indentation) using the keyboard.
+The {@link module:code-block/codeblock~CodeBlock} feature allows inserting and editing blocks of pre–formatted code into the WYSIWYG editor. Each code block has a [specific programming language assigned](#configuring-code-block-languages) (e.g. "Java" or "CSS") and supports basic editing tools, for instance, [changing the line indentation](#changing-line-indentation) using the keyboard. A full list of supported languages depends on the highlighter used. Refer to the [common API](#common-api) section for a default supported list.
```
AnnaTomanek, line 9: I'm not sure about adding this part. What CKEditor 5 does is just display the label, then turn the label into a CSS class in the output. If you have a highlighter integrated, the highlighter then can understand this class and style the snippet accordingly.

But there is nothing really limiting this list, the integrator can have anything they like there. Theoretically, they can configure the editor to "support" languages like `MyNewLanguage` or `MetallicaIsGreat` which would display this label next to the code block and that's pretty much it.

To sum it up, not sure who the audience here is. From the end-users' point of view, the list is configured by the integrator. From the integrators' POV, the list can be anything they want.

packages/ckeditor5-code-block/docs/features/code-blocks.md
```diff
@@ -6,14 +6,16 @@ category: features
...
 
 <info-box>
 	If you would like to use inline code formatting in your WYSIWYG editor, check out the {@link features/basic-styles basic text styles feature} with its support for inline `<code>` element.
 </info-box>
 
 ## Demo
 
+Use the code block toolbar button and the type dropdown, to insert desired code block. Alternatively, use the `` ``` `` code provided by the {@link features/autoformat autoformatting feature}.
```
AnnaTomanek, line 17:
```suggestion
Use the code block toolbar button and the type dropdown to insert a desired code block. Alternatively, start the line with `` ``` `` to format it as a code block thanks to the {@link features/autoformat autoformatting feature}.
```

AnnaTomanek (changes requested):

packages/ckeditor5-code-block/docs/features/code-blocks.md
godai78: The very first thing I personally thought of was "what are the codes that I can use here". Sure, this is defined in the dropdown and rests the case, but as an end-user I wanted to know what can be offered. This info offers some kind of answer to the integrator and via API link answers the user question. We may rephrase it, to point to the default list (I insist on that) and mention this can be expanded further. Which leaves us with a question of how, that should also be addressed.

AnnaTomanek: So maybe I'm missing something but the previous sentence links you exactly to more information about this? Also, you definitely should not send anyone to the API at the beginning of the guide - basics first, common API last. Not to mention the fact that you can find the default list both in the Common API section and in the API itself, so any click to [codeBlock.languages](https://ckeditor.com/docs/ckeditor5/latest/api/module_code-block_codeblock-CodeBlockConfig.html#member-languages) will get you there... :thinking:

AnnaTomanek: Maybe we can mention this is configurable, like: `(e.g. "Java" or "CSS" &mdash; this is configurable)`? Would that solve your issue?

godai78: Yes, mentioning this is configurable, see default list below could do the trick.

godai78, line 9:
```suggestion
The {@link module:code-block/codeblock~CodeBlock} feature allows inserting and editing blocks of pre–formatted code into the WYSIWYG editor. Each code block has a [specific programming language assigned](#configuring-code-block-languages) (e.g. "Java" or "CSS"; this is configurable, see the defauls list in the API section) and supports basic editing tools, for instance, [changing the line indentation](#changing-line-indentation) using the keyboard. A full list of supported languages depends on the highlighter used.
```

AnnaTomanek (approved):

PR #8444, opened 2020-11-10
Don't attach the image resizer to images inside the HTML embed preview

### LINKED ISSUE #8433: The `ImageResize` plugin crashes if `HtmlEmbed` plugin has enabled `showPreviews` option

The following code is sufficient to crash the editor:

```javascript
import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor';
import HtmlEmbed from '@ckeditor/ckeditor5-html-embed/src/htmlembed';
import Image from '@ckeditor/ckeditor5-image/src/image';
import ImageResize from '@ckeditor/ckeditor5-image/src/imageresize';

ClassicEditor
        plugins: [
            HtmlEmbed, Image, ImageResize
        ],
        htmlEmbed: {
            showPreviews: true
        }
    } )
    .then( editor => {
        window.editor = editor;
    } )
    .catch( err => {
        console.error( err.stack );
    } );
```

```html
<div id="editor">
    <div class="raw-html-embed">
        <img src="sample.jpg">
    </div>
</div>
```

### ✔️ Expected result

Editor should not crash.

### ❌ Actual result

There is an `Uncaught TypeError`:

```
Uncaught TypeError: Cannot read property 'is' of null
    at Function._createAt (position.js:349)
    at DowncastWriter.createPositionAt (downcastwriter.js:1015)
    at resizer.js:138
    at View.change (view.js:477)
    at Resizer.attach (resizer.js:117)
    at WidgetResize.attachTo (widgetresize.js:129)
    at Document.<anonymous> (imageresizehandles.js:76)
    at Document.fire (emittermixin.js:209)
    at ImageLoadObserver._fireEvents (imageloadobserver.js:52)
    at ProxyEmitter.listenTo.useCapture (imageloadobserver.js:35)
```

## 📃 Other details

* Browser: Firefox 82, Chrome 86
* OS: Windows 10 x64
* CKEditor version: 23.1.0
* Installed CKEditor plugins:
  - `HtmlEmbed`
  - `Image`
  - `ImageResize`

---

psmyrek:

Fix (image): Don't attach the image resizer to images inside the HTML embed preview. Closes #8433.


packages/ckeditor5-image/src/imageresize/imageresizehandles.js
```diff
@@ -56,6 +56,11 @@ export default class ImageResizeHandles extends Plugin {
 		editingView.addObserver( ImageLoadObserver );
 
 		this.listenTo( editingView.document, 'imageLoaded', ( evt, domEvent ) => {
+			// The resizer must not be attached to images inside HTML embed preview.
+			if ( domEvent.target.matches( 'div.raw-html-embed__preview img' ) ) {
```
jodator, line 60: The fix will work for html embed preview, but only for that.

On the horizon there's an inline image feature so the check should be specific for the CKEditor's image plugin.

In other words it should check if the target image is the one that we need to attach to not if it is invalid one.

TL;DR: the check should be something like `figure.image img` (to check).

The inline image would be just an `img` without a figure, hence the check.

jodator (changes requested):
I like the trick with `matches` as we have to somehow detect this case on the view level.

However, I'm worried that this fix will cover only one case of html embed and could fail in other cases:

- custom image plugin
- upcoming inline image
- dunno if media embed could have img inside?

ps.: I feel that I've might miss something so please double check what I've wrote if that makes sense.

packages/ckeditor5-image/src/imageresize/imageresizehandles.js
psmyrek: It's a good idea to check more than just one unsupported case (the HTML embed image previews) by reversing the logic and _allow_ only known (supported) images to be resized 👍🏻

psmyrek:
The media embed can theoretically have an image inside, because it creates an `iframe`, but it is wrapped by the `figure.media` rather than `figure.image`. Anyway, I've proposed stricter rule to match only supported images: `figure.image.ck-widget > img`.

jodator (approved):
Thanks! LGTM :+1:

PR #8448, opened 2020-11-12
Docs: expanding select all guide.

godai78:
Expanding the select all guide intro.

Adding keyboard shortcuts to Related features block.


AnnaTomanek (approved):
Good catch with the missing stuff! 👏

PR #8449, opened 2020-11-12
Added `toArray()` helper to cast any value to an array

### LINKED ISSUE #7401: Create a helper that normalizes a value to an array (if not already an array)

We use the following normalization `Array.isArray( foo ) ? foo : [ foo ];` many times across the code-base. I think we need an utils for this repeating pattern.

```
packages/ckeditor5-engine/src/conversion/conversion.js:
   77: 		this._downcast = Array.isArray( downcastDispatchers ) ? downcastDispatchers : [ downcastDispatchers ];
   80: 		this._upcast = Array.isArray( upcastDispatchers ) ? upcastDispatchers : [ upcastDispatchers ];
  640: 		upcastAlso = Array.isArray( upcastAlso ) ? upcastAlso : [ upcastAlso ];

packages/ckeditor5-engine/src/conversion/downcasthelpers.js:
  850: 				const classes = Array.isArray( oldAttribute.value ) ? oldAttribute.value : [ oldAttribute.value ];
  869: 				const classes = Array.isArray( newAttribute.value ) ? newAttribute.value : [ newAttribute.value ];

packages/ckeditor5-engine/src/view/element.js:
  730: 		className = Array.isArray( className ) ? className : [ className ];
  748: 		className = Array.isArray( className ) ? className : [ className ];
  795: 		property = Array.isArray( property ) ? property : [ property ];

packages/ckeditor5-image/src/image/imageinsertcommand.js:
  54: 			const sources = Array.isArray( options.source ) ? options.source : [ options.source ];

packages/ckeditor5-image/src/imageupload/imageuploadcommand.js:
  64: 			const filesToUpload = Array.isArray( options.file ) ? options.file : [ options.file ];

packages/ckeditor5-watchdog/src/contextwatchdog.js:
  237: 		const itemConfigurations = Array.isArray( itemConfigurationOrItemConfigurations ) ?
  313: 		const itemIds = Array.isArray( itemIdOrItemIds ) ?

packages/ckeditor5-widget/src/utils.js:
  120: 		return Array.isArray( classes ) ? classes : [ classes ];
```

psmyrek:

Internal: Added `toArray()` helper to cast any value to an array. Closes #7401.

---

### Additional information

All occurrences of `Array.isArray( value ) ? value : [ value ]` and `[].concat( value )` have been replaced by this new `toArray( value )` helper.

### DIFF
--- a/packages/ckeditor5-utils/src/toarray.js
+++ b/packages/ckeditor5-utils/src/toarray.js
@@ -10,9 +10,9 @@
 /**
  * Transforms any value to an array, if it is not already one. If provided value is already an array, it is returned unchanged.
  *
- * @param {*} [data] Value to transform to an array.
+ * @param {*} data Value to transform to an array.
  * @returns {Array} Array created from data.
  */
-export default function toArray( data = [] ) {
+export default function toArray( data ) {
 	return Array.isArray( data ) ? data : [ data ];
 }
--- a/packages/ckeditor5-utils/src/toarray.js
+++ b/packages/ckeditor5-utils/src/toarray.js
@@ -8,7 +8,7 @@
  */
 
 /**
- * Transforms any value to an array, if it is not already one. If provided value is already an array, it is returned unchanged.
+ * Transforms any value to an array. If the provided value is already an array, it is returned unchanged.
  *
  * @param {*} data Value to transform to an array.
  * @returns {Array} Array created from data.
--- a/packages/ckeditor5-utils/src/toarray.js
+++ b/packages/ckeditor5-utils/src/toarray.js
@@ -1,0 +1,18 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module utils/toarray
+ */
+
+/**
+ * Transforms any value to an array, if it is not already one. If provided value is already an array, it is returned unchanged.
+ *
+ * @param {*} [data] Value to transform to an array.
+ * @returns {Array} Array created from data.
+ */
+export default function toArray( data = [] ) {
+	return Array.isArray( data ) ? data : [ data ];
+}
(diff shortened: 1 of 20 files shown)


packages/ckeditor5-utils/src/toarray.js
```diff
@@ -0,0 +1,18 @@
...
+ * @module utils/toarray
+ */
+
+/**
+ * Transforms any value to an array, if it is not already one. If provided value is already an array, it is returned unchanged.
+ *
+ * @param {*} [data] Value to transform to an array.
+ * @returns {Array} Array created from data.
+ */
+export default function toArray( data = [] ) {
```
jodator, line 16: Quick question, as everything else looks very well: Does the default value is needed?

From the changes, I didn't see that we cared for this particular case earlier and we might not need to.

I'm asking here because it does _slightly_ change the logic of this utility:

```javascript
// A. no default:
( toArray() ).length === 1;

// B. with default:
( toArray() ).length === 0;
```

The current code will work with as in the (A) case since we always pass a variable. Passing an `undefined` value would be an error most probably and probably should fail later on. Having a default value would hide this error because other logic would get an empty array.

packages/ckeditor5-utils/src/toarray.js
```diff
@@ -0,0 +1,18 @@
...
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module utils/toarray
+ */
+
+/**
+ * Transforms any value to an array, if it is not already one. If provided value is already an array, it is returned unchanged.
```
jodator, line 11:
```suggestion
 * Transforms any value to an array. If the provided value is already an array, it is returned unchanged.
```

jodator (changes requested):
Nice catch with all the variants of this check, especially `[].concat()` one :)

I do have one question about the need for the default value though.

packages/ckeditor5-utils/src/toarray.js
psmyrek: You are right: the default `[]` is not needed. This new `toArray()` helper should behave exactly the same as the code before the changes 👍🏻

packages/ckeditor5-utils/src/toarray.js
psmyrek: Done 👍🏻

jodator (approved):
LGTM :+1:

PR #8459, opened 2020-11-13
Fix (indent): blockIndent now works with custom headings. Closes #8177.

### LINKED ISSUE #8177: Not possible to use Block Indent Feature with custom headings
1. Create custom header as defined in [Headings](https://ckeditor.com/docs/ckeditor5/latest/features/headings.html) page.
2. install block indentation plugin, as stated in [its page](https://ckeditor.com/docs/ckeditor5/latest/features/indent.html)
3. Select the custom the created custom header 
4. Try to use the indent feature

### ✔️ Expected result

I exptect to be able to use the block indent feature with Custom Headers

### ❌ Actual result

Both buttons are never enabled, so I can not use this feature

## 📃 Other details

Looking into the [source](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-indent/src/indentblock.js#L82-L83), I saw the indent block feature only enable itself to knownElements:
```js
// Enable block indentation by default in paragraph and default headings.
const knownElements = [ 'paragraph', 'heading1', 'heading2', 'heading3', 'heading4', 'heading5', 'heading6' ];
```

This way, we can not use this feature with any custom Headers.

A simple change can allow that. Changing the above line to:
```js
// Enable block indentation by default in paragraph and default headings.
const options = editor.config.get( 'heading.options' );
const headings = options.map(option => option.model);
const knownElements = headings || [ 'paragraph', 'heading1', 'heading2', 'heading3', 'heading4', 'heading5', 'heading6' ];
```
and voilá, block indent works for all defined headers. Another option would be to have a config option, but I think it would only to double the config between headers and blockindent. 

Maybe we do not need the last line, once the [headings plugin](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-heading/src/headingediting.js#L39-L47) always set the heading options. But if the user remove the heading plugin, I do not know the results of that.

Please, let me know if this is the way to fix this. I can make a PR, if you folks approve it. If not, please tell me what should be the solution

   * Browser: Firefox 80.0
   * OS: Linux mint 19.3
   * CKEditor version: 22.0
   * Installed CKEditor plugins: Essentials, Alignment, Paragraph, Heading, List, Bold, Italic, Indent, Outdent

As a final note, the link to see the code at the end of the [contribute section of block indent feature](https://ckeditor.com/docs/ckeditor5/latest/features/indent.html#contribute) is pointing to the font plugin, not the indent plugin.

---

antoniolucasnobar:
Fix (indent): blockIndent now works with custom headings. Closes #8177.

---

### Additional information

Please let me know if you want me to change anything.

### DIFF
--- a/packages/ckeditor5-indent/src/indentblock.js
+++ b/packages/ckeditor5-indent/src/indentblock.js
@@ -79,8 +79,10 @@ export default class IndentBlock extends Plugin {
 		const indentCommand = editor.commands.get( 'indent' );
 		const outdentCommand = editor.commands.get( 'outdent' );
 
-		// Enable block indentation by default in paragraph and default headings.
-		const knownElements = [ 'paragraph', 'heading1', 'heading2', 'heading3', 'heading4', 'heading5', 'heading6' ];
+		// Enable block indentation to headging configuration options. If it is not defined enable in paragraph and default headings.
+		const options = editor.config.get( 'heading.options' );
+		const headings = options && options.map( option => option.model );
+		const knownElements = headings || [ 'paragraph', 'heading1', 'heading2', 'heading3', 'heading4', 'heading5', 'heading6' ];
 
 		knownElements.forEach( elementName => {
 			if ( schema.isRegistered( elementName ) ) {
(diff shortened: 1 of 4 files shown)


jodator (approved):
Hi @antoniolucasnobar, thanks for a great PR!

We really appreciate PRs with tests :)

I've added some additions to the PR just to have those covered in the manual tests - mostly used by the CKEditor 5 team.

I was thinking about removing the `||` since the configuration option should be always present but we do allow implementing the own heading plugin, so in theory, it could be that `heading.options` will not be defined.

Big thanks!

ps.: Bonus points for using `fancy` heading as it was my favorite addition to the docs :D

jodator:
BTW, I've fixed the headings demo: 12b1732.

PR #8471, opened 2020-11-17
Docs: Fix UpcastWriter example in Clipboard Deep Dive

### LINKED ISSUE #8470: Clipboard deep dive instantiates UpcastWriter incorrectly

_What is the purpose and what should be changed?_

In the code block here https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/clipboard.html#processing-input-content, the `UpcastWriter` constructor is called without passing a `viewDocument` argument. If you try to write this plugin according to the docs, you end up with an exception like 

```
Uncaught TypeError: Cannot read property 'stylesProcessor' of undefined at new Element
```

when you try to call `writer.createElement`.

It took me a while to figure this out, but the [UpcastWriter](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_upcastwriter-UpcastWriter.html) docs show the correct way to instantiate the object is by passing in a view document. In the case of the clipboard plugin, it would specifically be the editing view document.

rshipp:

Docs: Fix UpcastWriter example in Clipboard Deep Dive. Closes #8470.

---

### Additional information

Small docs fix.

### DIFF
--- a/packages/ckeditor5-clipboard/docs/framework/guides/deep-dive/clipboard.md
+++ b/packages/ckeditor5-clipboard/docs/framework/guides/deep-dive/clipboard.md
@@ -68,7 +68,7 @@ The default action is to {@link module:engine/model/model~Model#insertContent in
 At this stage the pasted content can be processed by the features. For example, a feature that wants to transform the pasted text into a link can be implemented in this way:
 
 ```js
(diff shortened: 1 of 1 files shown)


jodator:
Thanks! Good catch :)

PR #8474, opened 2020-11-18
Improve the Placeholder behavior to keep shown when focused.

huacnlee:
屏幕录制2020-11-18 11 53 44

### DIFF
--- a/packages/ckeditor5-engine/src/view/placeholder.js
+++ b/packages/ckeditor5-engine/src/view/placeholder.js
@@ -152,18 +152,8 @@ export function needsPlaceholder( element ) {
 	const isEmptyish = !Array.from( element.getChildren() )
 		.some( element => !element.is( 'uiElement' ) );
 
-	const doc = element.document;
-
-	// If the element is empty and the document is blurred.
-	if ( !doc.isFocused && isEmptyish ) {
-		return true;
-	}
-
-	const viewSelection = doc.selection;
-	const selectionAnchor = viewSelection.anchor;
-
-	// If document is focused and the element is empty but the selection is not anchored inside it.
-	if ( isEmptyish && selectionAnchor && selectionAnchor.parent !== element ) {
+	// If the element, keep placeholder
+	if ( isEmptyish ) {
 		return true;
 	}
 
--- a/packages/ckeditor5-engine/tests/view/placeholder.js
+++ b/packages/ckeditor5-engine/tests/view/placeholder.js
@@ -95,7 +95,7 @@ describe( 'placeholder', () => {
 			} );
 
 			expect( element.getAttribute( 'data-placeholder' ) ).to.equal( 'foo bar baz' );
-			expect( element.hasClass( 'ck-placeholder' ) ).to.be.false;
+			expect( element.hasClass( 'ck-placeholder' ) ).to.be.true;
 		} );
 
 		it( 'if element has selection inside but document is blurred should contain placeholder CSS class', () => {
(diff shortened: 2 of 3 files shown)


oleq:
Hi @huacnlee!

Thanks for this contribution. It makes perfect sense that the editor placeholder should act the same as the `<input>` placeholder (it should disappear as you type, not when you focus). I checked your PR (editor root placeholder, the title plugin, RTL language support) and it all works fine. 

There's one problem, though: the image caption placeholder looks off:



I think we can correct it by using `transform: translateX(-50%)` for the `.image figcaption.ck-placeholder::before`:



which corresponds to what an `<input>` with a centered placeholder would look like ([you can play with it live](https://www.w3docs.com/snippets/css/how-to-align-placeholder-text-of-input-field-in-html.html)):



The question remains: do we want this @Reinmar @magda-chrzescian? This still looks off to me.

## Other options

We could have something like below instead (`.image figcaption.ck-placeholder::before { position: static }`) but then:

*   👎 it's incorrect that the caret appears right after the placeholder text instead of before it,
*   👎 the caret would need to go back to the center of the `<figcaption>` when the user starts typing and that looks ugly (the caret jumping around like this)



Finally, the last way would be a perfectly centered placeholder text (like above) but a caret at the beginning of it:

*   :+1: this would at least look like the same as the editor placeholder,
*   ❓ I'm not sure this possible using CSS (I tried but failed),
*   👎  the caret would need to go back to the center of the `<figcaption>` when the user starts typing and that looks ugly (the caret jumping around like this).

huacnlee:
😂  That looks like too complex for me, I'm afraid that to have enough capacity to fix that. 
I make this PR just wants fix the Placeholder behavior in editor init.

magda-chrzescian:
I would vote for center-center position. It's not perfect but I like that it seems consistent with the native input behavior and clear about the text alignment.  
  
However, maybe the caption placeholder should be invisible on focus by default? I think that the main use case of focus-visible placeholder is when you want to prompt the user to fill the input (so it should be focused and should inform of the need to be filled) - it can be common regarding the content of editor, but rather rare in the case of image caption.  
  
I really don't like the idea of caret jumping around right after the user starts to write. It would make an impression that something crashed and the user may stop writing for a second to check what happened. Besides, the initial position of caret can be confusing about how the text will be aligned.

oleq:
> However, maybe the caption placeholder should be invisible on focus by default?

I forked this PR in [https://github.com/ckeditor/ckeditor5/compare/i/8689-smarter-editor-placeholder?expand=1](https://github.com/ckeditor/ckeditor5/compare/i/8689-smarter-editor-placeholder?expand=1) and I came up with a configuration for `enablePlaceholder()` that allows that.



I also checked the `Title` plugin and it also looks fine. 

@ckeditor/qa-team, can you take a look at https://github.com/ckeditor/ckeditor5/compare/i/8689-smarter-editor-placeholder?expand=1 and see if nothing else got broken (CF?) because of this? 🙏

FilipTokarski:
### CKE 5 
It seems to work fine, I didn't find any new bugs.
### CF 
I had errors at the beginning, but after locally merging latest `master` into the branch, everything works ok.  
I see that apart from _editor placeholder_ and _Title plugin_, this affects the behavior of annotations inputs too:  

https://user-images.githubusercontent.com/34380544/105040459-29d43500-5a62-11eb-9f53-5ad217a27870.mp4
  
I'm not sure whether this was intended, but so far I didn't find any bugs related to this too.

oleq:
Thanks @huacnlee for this contribution!

PR #8478, opened 2020-11-18
Fixed saving in HTML embed after it lost selection

### LINKED ISSUE #8328: Can't save an HTML embed after pasting a text from the editor

psmyrek:

Fix (html-embed): Fixed saving in HTML embed after it lost selection. Closes #8328.

### DIFF
--- a/packages/ckeditor5-html-embed/src/htmlembedediting.js
+++ b/packages/ckeditor5-html-embed/src/htmlembedediting.js
@@ -92,7 +92,6 @@ export default class HtmlEmbedEditing extends Plugin {
 		const editor = this.editor;
 		const t = editor.t;
 		const view = editor.editing.view;
-		const model = editor.model;
 
 		const htmlEmbedConfig = editor.config.get( 'htmlEmbed' );
 		const upcastWriter = new UpcastWriter( view.document );
@@ -144,6 +143,21 @@ export default class HtmlEmbedEditing extends Plugin {
 					domContentWrapper = domElement;
 
 					renderContent( { domElement, editor, state, props } );
+
+					// Since there is a `data-cke-ignore-events` attribute set on the wrapper element in the editable mode,
+					// the explicit `click` handler on the `capture` phase is needed to move the selection onto the whole
+					// HTML embed widget.
+					domContentWrapper.addEventListener( 'click', () => {
+						if ( state.isEditable ) {
+							const model = editor.model;
+							const selectedElement = model.document.selection.getSelectedElement();
+
+							// Move the selection onto the whole HTML embed widget if it's currently not selected.
+							if ( selectedElement !== modelElement ) {
+								model.change( writer => writer.setSelection( modelElement, 'on' ) );
+							}
+						}
+					}, true );
 				} );
 
 				// API exposed on each raw HTML embed widget so other features can control a particular widget.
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-html-embed/src/htmlembedediting.js
```diff
@@ -165,6 +166,14 @@ export default class HtmlEmbedEditing extends Plugin {
 						// If the value didn't change, we just cancel. If it changed,
 						// it's enough to update the model – the entire widget will be reconverted.
 						if ( newValue !== state.getRawHtmlValue() ) {
+							const selectedElement = model.document.selection.getSelectedElement();
+							const modelContainer = editor.editing.mapper.toModelElement( viewContainer );
+
+							// The HTML embed widget must be selected to be able to save new value.
+							if ( selectedElement !== modelContainer ) {
+								model.change( writer => writer.setSelection( modelContainer, 'on' ) );
+							}
```
niegowski, line 175: I'm not sure if this is the right place for this. This widget focus behaves quite inconsistent (because of the `data-cke-ignore-events` feature, but it's needed in editing mode).

Have a look at this screencast:

While not in editing mode - click on the widget is handled by the [`mousedown` handler in the widget plugin](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-widget/src/widget.js#L187), but in the editing mode those events are blocked by `data-cke-ignore-events`. Maybe this should be handled in a similar way in this feature - the selection might be moved onto the widget before a button click was handled or a textarea focused?

niegowski (changes requested):

packages/ckeditor5-html-embed/src/htmlembedediting.js
jodator: That would solve other issues as well. I'm thinking about handling this in other PR (#8421).

Similar case: #8328.

packages/ckeditor5-html-embed/src/htmlembedediting.js
```diff
@@ -165,6 +166,14 @@ export default class HtmlEmbedEditing extends Plugin {
 						// If the value didn't change, we just cancel. If it changed,
 						// it's enough to update the model – the entire widget will be reconverted.
 						if ( newValue !== state.getRawHtmlValue() ) {
+							const selectedElement = model.document.selection.getSelectedElement();
+							const modelContainer = editor.editing.mapper.toModelElement( viewContainer );
```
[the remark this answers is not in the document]
psmyrek: You are right, I missed that.

packages/ckeditor5-html-embed/src/htmlembedediting.js
psmyrek: I've changed my solution basing on your comment (I hope I understood you correctly 😅), and I've added a `click` event listener on the `capture` phase to the wrapper to be able to move the selection onto the whole widget in the editing mode.

packages/ckeditor5-html-embed/src/htmlembedediting.js
```diff
@@ -143,6 +143,21 @@ export default class HtmlEmbedEditing extends Plugin {
 					domContentWrapper = domElement;
 
 					renderContent( { domElement, editor, state, props } );
+
+					// Since there is a `data-cke-ignore-events` attribute set on the wrapper element in the editable mode,
+					// the explicit `click` handler on the `capture` phase is needed to move the selection onto the whole
+					// HTML embed widget.
+					domContentWrapper.addEventListener( 'click', () => {
```
pomek, line 150: Wondering why couldn't we attach the listener on`textarea#focus`? Shouldn't it be easier?

niegowski: In widgets, it's handled on `mousedown` so maybe we should do the same here? The click event is triggered on a `mouseup` so there is a visible difference in behavior.

psmyrek: Done - the `click` event listener has been replaced with `mousedown` event listener.

PR #8484, opened 2020-11-19
DocumentSelection#markers collection should be updated only for observed markers groups

niegowski:

Other (engine): The `DocumentSelection#markers` collection is updated only for observed markers groups.

BREAKING CHANGE (engine): The `DocumentSelection#markers` collection is no longer enabled by default. `DocumentSelection#observeMarkersGroup()` must be used to register markers group to be observed.

---

### Additional information

This is performance optimization.


packages/ckeditor5-engine/src/model/documentselection.js
```diff
@@ -619,6 +632,11 @@ class LiveSelection extends Selection {
 		// @type {Set}
 		this._overriddenGravityRegister = new Set();
 
+		// Prefixes of marker names that should affect `LiveSelection#markers` collection.
+		// @private
+		// @type {Array.<String>}
+		this._observedMarkerGroups = [];
```
scofalik, line 638: How about using `Set` for potential dedups?

packages/ckeditor5-engine/src/model/documentselection.js
```diff
@@ -662,7 +680,9 @@ class LiveSelection extends Selection {
 		} );
 
 		// Update markers data stored by the selection after each marker change.
-		this.listenTo( this._model.markers, 'update', () => this._updateMarkers() );
+		this.listenTo( this._model.markers, 'update', ( evt, marker, oldRange, newRange ) => {
```
scofalik, line 683: Could you add a comment here:

```
// This handles only marker changes done through marker operations (not model tree changes).
```

I know this is not connected with your changes but it could be a good tip for someone who is newer to how CKE5 works.

packages/ckeditor5-engine/src/model/documentselection.js
niegowski: done

packages/ckeditor5-engine/src/model/documentselection.js
niegowski: done

packages/ckeditor5-engine/src/model/documentselection.js
```diff
@@ -364,6 +366,17 @@ export default class DocumentSelection {
...
 	}
 
+	/**
+	 * Registers marker group prefix to be collected in {@link ~DocumentSelection#markers selection markers collection}.
+	 *
+	 * See also {@link module:engine/model/markercollection~MarkerCollection#getMarkersGroup `MarkerCollection#getMarkersGroup()`}.
+	 *
+	 * @param {String} prefix Marker group prefix.
+	 */
+	observeMarkersGroup( prefix ) {
```
scofalik, line 376: How about `groupName` instead. of prefix, if we use "markers group" everywhere?. Also could you add an example in the API docs? I know this is straightforward but "marker group" is a loose term that is introduced only in the API docs, this is not a property of a marker or something, so someone may be lost what it means.

packages/ckeditor5-engine/src/model/documentselection.js
```diff
@@ -882,6 +917,54 @@ class LiveSelection extends Selection {
 		}
 	}
 
+	_updateMarker( marker, markerRange ) {
+		if ( !this._observedMarkerGroups.length ) {
```
scofalik: Not needed - this is covered by the next if pretty well.

packages/ckeditor5-engine/src/model/documentselection.js
niegowski: That's why I linked to `MarkersCollection` - there is `prefix` so I kept the same convention (and this seems more reasonable than `groupName` because it is not confusing with any `Marker` property.

packages/ckeditor5-engine/src/model/documentselection.js
niegowski: This was to make it return as early as possible. There is no point in splitting marker name and looking for it in the Set if it is empty

scofalik: The gain here is very minimal. In such situations, I prefer fewer lines of code. I guess it is a bit of a personal preference. But really, you don't gain anything here by leaving early.

niegowski: ok

niegowski: done

scofalik:
I left some review comments, minor stuff.

I have one doubt, though. This is connected with the fact that we never forced any structure for markers and something like "marker group" was not really specified. It just appeared that in 99% of cases there are multiple markers of the same type and it makes sense to handle all of them in the same way. Hence, we have `comment:` and `suggestion:` and "marker groups" were born.

The problem is that it is still possible to create a marker with a name that does not have `:`. We should handle them as well. For example, someone could implement find and replace and use marker named `searchResult` to highlight the current result. So, in places where we check whether we should observe given marker, we should probably check whether prefix matches and if not, whether full name matches. And include this case in API docs for `observeMarkersGroup`. Maybe we should also rename that method to `observeMarkers` or `observeMarkersNames` or something?

packages/ckeditor5-engine/src/model/documentselection.js
scofalik: You are right - we use prefix there.

Issue #1602, opened 2019-03-06
[Docs] Add 3rd party framework integrations section


Reinmar:
I think we could have a section here https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/overview.html or a separate page where we'd list 3rd party integrations.

Reinmar:
I'd keep this page short and only list them. No documentation, nothing. Perhaps a short description only.

wwalc:
Include Omi framework (https://github.com/Tencent/omi) integration: https://github.com/LeeHyungGeun/ckeditor5-omi

Reinmar:
https://github.com/postedin/ember-ckeditor

godai78:
Frameworks mentioned were included in this PR: #8503

godai78:
Also found the following things we might consider interesting:

* https://github.com/hvlads/django-ckeditor-5 (last touched like half a year ago)
* https://github.com/pancodia/django-ckeditor5 (even older)
* https://github.com/plannowtech/tapestry5-ckeditor (now, this one looks abandoned and outdated)

AnnaTomanek:
I'm honestly not convinced that the official documentation is a good place to gather such links. For me, official = covers the stuff that we provide and support.

First of all, even if we add the "third-party integration" heading or note, many users will assume that we somehow promote/approve these solutions while it does not seem probable that we will be checking them for quality, completeness, security, compatibility with the latest CKEditor 5 and framework version, etc. Additionally, the users may then feel like they should report issues about them in our repository or customer channel.

Such a list will also never be complete. What do we do if there are more integrations for the same library (as it is with Django above)? Who should decide which integration will be placed on the list and when? When it is announced on Twitter by its author? When it reaches some maturity? What do we do with the projects that are abandoned? Many of these will be some one-man projects done once and then forgotten.

I cannot seem to understand the value from having this (incomplete and imperfect) list in the official docs when it is not difficult to find them using Google, npm, GitHub search, Stack Overflow, any module/plugin repository for a given framework, or even searching on socials. What am I missing here?

godai78:
It IS of added value for the potential user. Personally, I think we should at least mention there exist 3rd party integrations, even if we do not list these. This is an encouragement and a plus for somebody looking for a solution.

As for the list, I decided to include the ones that seem to be still developed, dropped the ones that were not worked on for years - but I see the reasoning behind dropping this list as troublesome - we'd need to periodically update it.

[...]

PR #8503, opened 2020-11-23
Docs: adding third-party integrations.

godai78:
Addresses this: #1602

Adding third-party integrations to the framework integrations guide.


godai78:
Got outdated pretty badly. Keeping the original issue in case we wanted to revisit this idea.

PR #8529, opened 2020-11-25
Introducing DataProcessor#registerRawContentElementMatcher()

### LINKED ISSUE #5870: DomConverter should not trim whitespaces in nodes that neighbour with black–box elements (e.g. MathML)

**Note**: I used my PoC code here that brings necessary conversion to load `<math>` to showcase the problem. But the issue isn't about the PoC code because it occurs before any conversion kicks in, which is at the raw HTML parsing stage.

1. Try to load this kind of HTML into the editor
```html
<p>Foo <math xmlns="http://www.w3.org/1998/Math/MathML">x</math> bar</p>
```
2. It loads just fine 
image
3. But now try this one (note the trailing space in the `<math>`):
```html
<p>Foo <math xmlns="http://www.w3.org/1998/Math/MathML">x </math> bar</p>
```

### ✔️ Expected result

Spaces after "foo" and before "bar" should be preserved.

### ❌ Actual result

The later is gone:

image

## 📃 Other details

I debugged the issue, and the story is as follows:

1. In `DomConverter` there's a private method that converts native DOM nodes into text.
2. [It does trimming](https://github.com/ckeditor/ckeditor5-engine/blob/e2d134ba62b05900408334c451f30701fe9403e9/src/view/domconverter.js#L1033) from both sides if 
   * there's no adjacent node,
   * adjacent node is an element,
   * adjacent node is text and its last character is a white-space (**this is is the case we're entering here**)
3. The trimming is based on what is [considered a "touching inline node"](https://github.com/ckeditor/ckeditor5-engine/blob/e2d134ba62b05900408334c451f30701fe9403e9/src/view/domconverter.js#L1030).
4. The "touching inline node" is found [using the native tree walker](https://github.com/ckeditor/ckeditor5-engine/blob/e2d134ba62b05900408334c451f30701fe9403e9/src/view/domconverter.js#L1174-L1186).
5. The walker enters a `<math>` and if it ends with some white spaces like, for instance, an indentation used to make the code human-readable (or like a trailing space from the TC), it figures this is it and gives a green light for trimming, which is obviously wrong.

The white spaces in the `<math>` do not render like regular white spaces in a bold or something, so they should never be considered.

We need a fix and here are some ideas:
* Some `DomConverter#objectLike` collection that would collect black–box elements. It would be used in the walker to prevent it from considering black–box content and thus, it would prevent the trimming.
* I think we could safely skip all elements in the tree walker that have `Element#namespaceURI`. I didn't test it but I suppose if we replaced `<math>` with inline `<svg>` (that is also formatted inside), we'd end up with the same issue.

The issue was spotted when creating a [PoC for MathML ](https://github.com/ckeditor/ckeditor5/compare/poc/mathml?expand=1).

---

### LINKED ISSUE #8323: Editor crashes after inserting a specific HTML embed

1.  Go to http://localhost:8080/build/docs/ckeditor5/latest/features/html-embed.html.
2.  Call `editor.getData()` and copy the string.
3.  Paste it to a new HTML snippet.
4.  Switch **`htmlEmbed.showPreviews`** to `false`.

### ❌ Actual result

The editor crashes permanently.

Error from Chrome:

```
snippet.js:4 Uncaught (in promise) Error: Failed to execute 'setAttribute' on 'Element': 'picture-in-picture\"' is not a valid attribute name.
    at b.viewToDom (snippet.js:4)
    at b.viewChildrenToDom (snippet.js:4)
    at viewChildrenToDom.next (<anonymous>)
    at b.viewToDom (snippet.js:4)
    at b.viewChildrenToDom (snippet.js:4)
    at viewChildrenToDom.next (<anonymous>)
    at b.viewToDom (snippet.js:4)
    at o.toData (snippet.js:4)
    at model (snippet.js:4)
    at snippet.js:4
```

Error from Safari:

```
Unhandled Promise Rejection: CKEditorError: The string contains invalid characters.
Read more: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-The string contains invalid characters.
```

## 📃 Other details

I was able to reproduce it only in docs, the manual test works fine for me.

---

niegowski:

Feature (engine): Introducing `DataProcessor#registerRawContentMatcher()`. 

Fix (html-embed): Editor should not crash after inserting a broken HTML. Closes #8323.

Fix (engine): `DomConverter` should not trim whitespaces in nodes that are siblings to inline raw content elements (e.g. MathML). Closes #5870.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/view/domconverter.js
+++ b/packages/ckeditor5-engine/src/view/domconverter.js
@@ -133,7 +133,7 @@ export default class DomConverter {
 		 * Set of encountered raw content DOM nodes. It is used for preventing left trimming of the following text node.
 		 *
 		 * @private
-		 * @type {WeakSet<Node>}
+		 * @type {WeakSet.<Node>}
 		 */
 		this._encounteredRawContentDomNodes = new WeakSet();
 	}
--- a/packages/ckeditor5-engine/src/view/domconverter.js
+++ b/packages/ckeditor5-engine/src/view/domconverter.js
@@ -16,6 +16,7 @@ import ViewRange from './range';
 import ViewSelection from './selection';
 import ViewDocumentFragment from './documentfragment';
 import ViewTreeWalker from './treewalker';
+import Matcher from './matcher';
 import { BR_FILLER, getDataWithoutFiller, INLINE_FILLER_LENGTH, isInlineFiller, NBSP_FILLER, startsWithFiller } from './filler';
 
 import global from '@ckeditor/ckeditor5-utils/src/dom/global';
@@ -118,6 +119,14 @@ export default class DomConverter {
 		 * @member {WeakMap} module:engine/view/domconverter~DomConverter#_fakeSelectionMapping
 		 */
 		this._fakeSelectionMapping = new WeakMap();
+
+		/**
+		 * TODO
+		 *
+		 * @private
+		 * @type {module:engine/view/matcher~Matcher}
+		 */
+		this._cdataElementMatcher = new Matcher();
 	}
 
 	/**
@@ -253,7 +262,7 @@ export default class DomConverter {
 				}
 			}
 
-			if ( options.withChildren || options.withChildren === undefined ) {
+			if ( options.withChildren !== false ) {
 				for ( const child of this.viewChildrenToDom( viewNode, domDocument, options ) ) {
 					domElement.appendChild( child );
 				}
@@ -400,7 +409,7 @@ export default class DomConverter {
 		}
 
 		// When node is inside a UIElement or a RawElement return that parent as it's view representation.
-		const hostElement = this.getHostViewElement( domNode, this._domToViewMapping );
+		const hostElement = this.getHostViewElement( domNode );
 
 		if ( hostElement ) {
 			return hostElement;
@@ -445,9 +454,16 @@ export default class DomConverter {
 				for ( let i = attrs.length - 1; i >= 0; i-- ) {
 					viewElement._setAttribute( attrs[ i ].name, attrs[ i ].value );
 				}
+
+				// Treat this element's content as CDATA if it was registered as such.
+				if ( options.withChildren !== false && this._cdataElementMatcher.match( viewElement ) ) {
+					viewElement._appendChild( new ViewText( this.document, domNode.innerHTML ) );
+
+					return viewElement;
+				}
 			}
 
-			if ( options.withChildren || options.withChildren === undefined ) {
+			if ( options.withChildren !== false ) {
 				for ( const child of this.domChildrenToView( domNode, options ) ) {
 					viewElement._appendChild( child );
 				}
--- a/packages/ckeditor5-engine/src/view/domconverter.js
+++ b/packages/ckeditor5-engine/src/view/domconverter.js
@@ -121,12 +121,13 @@ export default class DomConverter {
 		this._fakeSelectionMapping = new WeakMap();
 
 		/**
-		 * TODO
+		 * Matcher for view elements whose content should be treated as a plain text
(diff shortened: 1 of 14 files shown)


packages/ckeditor5-engine/src/view/domconverter.js
```diff
@@ -911,6 +939,23 @@ export default class DomConverter {
...
+	 * {@link module:engine/view/domconverter~DomConverter#domChildrenToView} processes DOM nodes.
+	 *
+	 * The raw data can be later accessed by {@link module:engine/view/element~Element#getCustomProperty view element custom property}
+	 * `"$rawContent"`.
+	 *
+	 * @param {module:engine/view/matcher~MatcherPattern} pattern Pattern matching all view elements whose content should
+	 * be treated as a raw data.
+	 */
+	registerRawContentElementMatcher( pattern ) {
+		this._rawContentElementMatcher.add( pattern );
```
jodator, line 956: WDYT about changing the name to a bit shorter: `registerRawContentMatcher()` ?

packages/ckeditor5-engine/src/view/domconverter.js
```diff
@@ -911,6 +939,23 @@ export default class DomConverter {
...
+	 * Registers a {@link module:engine/view/matcher~MatcherPattern} for view elements whose content should be treated as a raw data
+	 * and not processed during conversion from DOM nodes to view elements.
+	 *
+	 * This is affecting how {@link module:engine/view/domconverter~DomConverter#domToView} and
+	 * {@link module:engine/view/domconverter~DomConverter#domChildrenToView} processes DOM nodes.
+	 *
+	 * The raw data can be later accessed by {@link module:engine/view/element~Element#getCustomProperty view element custom property}
+	 * `"$rawContent"`.
+	 *
+	 * @param {module:engine/view/matcher~MatcherPattern} pattern Pattern matching all view elements whose content should
```
jodator, line 952:
```suggestion
	 * @param {module:engine/view/matcher~MatcherPattern} pattern Pattern matching view element which content should
```

packages/ckeditor5-engine/tests/view/domconverter/dom-to-view.js
```diff
@@ -158,6 +158,42 @@ describe( 'DomConverter', () => {
 			expect( viewFragment2 ).to.equal( viewFragment );
 		} );
 
+		it( 'should assign $rawContent custom property for view elements registered as raw content elements', () => {
```
jodator, line 161: Too few tests IMO. Right now it tests only one pattern. I think that we could check a bit more:

*   multiple patterns registered
*   what if `div.raw-content-container` has other attributes, classed or styles, ie `div[ data-foo="bar"].raw-content-container`

jodator (changes requested):
I would add more tests for the new method. We should add more test cases to define what happens in a scenario, where two or more features registers raw content.

The method name is a bit long, so I've proposed a shorter one as I think that `Element` part is redundant. But that's up to you to decide whether you think that's a valid comment.

packages/ckeditor5-engine/src/view/domconverter.js
```diff
@@ -118,6 +119,23 @@ export default class DomConverter {
...
+		 * @private
+		 * @type {module:engine/view/matcher~Matcher}
+		 */
+		this._rawContentElementMatcher = new Matcher();
+
+		/**
+		 * Set of encountered raw content DOM nodes. It is used for preventing left trimming of the following text node.
+		 *
+		 * @private
+		 * @type {WeakSet<Node>}
```
Reinmar, line 136:
```suggestion
		 * @type {WeakSet.<Node>}
```

jodator (approved):
Thanks! :+1:

PR #8535, opened 2020-11-26
Added support for RTL in html-embed.

### LINKED ISSUE #8335: HTML embed editing UI is broken when editor uses an RTL language

1.  Go to the HTML embed manual test.
2.  Put `language: 'ar'` in editor config.
3.  Edit an embed.

### ❌ Actual result

---

niegowski:

Fix (html-embed): HTML embed editing UI should not be broken when the editor uses an RTL language. Closes #8335.

Fix (theme-lark): HTML embed editing UI should not be broken when the editor uses an RTL language. 

Feature (ui): Added support for a new south-east position of the `TooltipView`.

Feature (theme-lark): Added styles for a new south-east position of the `TooltipView`. 

---

### Additional information


packages/ckeditor5-theme-lark/theme/ckeditor5-html-embed/htmlembed.css
```diff
@@ -118,3 +122,17 @@
 	}
 }
 
+.ck-content[dir="rtl"] .ck-widget.raw-html-embed {
```
oleq, line 125: These are not content styles and they should not start with `.ck-content`. Both the label and the buttons wrapper are not in the editor output data and the styles should not be [listed in the guide](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/content-styles.html#the-full-list-of-content-styles).

I think this should be done using `@mixin ck-dir rtl { ... }`.

packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/tooltip/tooltip.css
```diff
@@ -117,3 +117,21 @@
 		}
 	}
 }
+
+.ck[dir="rtl"] .ck.ck-tooltip {
```
oleq, line 121: This should not be handled in CSS like this. Instead, this should be done [at the JS level](https://github.com/ckeditor/ckeditor5/blob/i/8335/packages/ckeditor5-html-embed/src/htmlembedediting.js#L350) based on `locale.uiLanguageDirection` (`locale.uiLanguageDirection === 'rtl' ? 'se' : 'sw'`). To do that we need a new `.ck-tooltip_se` style [corresponding to the existing `.ck-tooltip_sw`](https://github.com/ckeditor/ckeditor5/blob/i/8335/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/tooltip/tooltip.css#L84-L98) and documented in the [`TooltipView#position` property docs](https://github.com/ckeditor/ckeditor5/blob/i/8335/packages/ckeditor5-ui/src/tooltip/tooltipview.js#L57).

oleq (changes requested):
There's also one thing not covered by this PR: [These styles](https://github.com/ckeditor/ckeditor5/blob/i/8335/packages/ckeditor5-theme-lark/theme/ckeditor5-html-embed/htmlembed.css#L22-L24) should not work in RTL because the embed label is on the other side and we don't need to compensate for it.

Actual:

image

Expected:

image

PR #8537, opened 2020-11-26
CloudServices Plugin should allow for registering multiple token endpoints

niegowski:

Internal (cloud-services): CloudServices Plugin should allow for registering multiple token endpoints. 

---

### Additional information

### DIFF
--- a/packages/ckeditor5-cloud-services/src/cloudservices.js
+++ b/packages/ckeditor5-cloud-services/src/cloudservices.js
@@ -76,7 +76,11 @@ export default class CloudServices extends ContextPlugin {
 			return;
 		}
 
-		return this.registerTokenUrl( this.tokenUrl, true );
+		this.token = new CloudServices.Token( this.tokenUrl );
+
+		this._tokens.set( this.tokenUrl, this.token );
+
+		return this.token.init();
 	}
 
 	/**
@@ -84,10 +88,10 @@ export default class CloudServices extends ContextPlugin {
 	 * {@link module:cloud-services/cloudservices~CloudServicesConfig#tokenUrl} for more details.
 	 *
 	 * @param {String|Function} tokenUrl The authentication token URL for CKEditor Cloud Services or a callback to the token value promise.
-	 * @param {Boolean} [isDefault=false] Whether this should be a default authentication token provider.
 	 * @returns {Promise.<module:cloud-services-core/token~Token>}
 	 */
-	registerTokenUrl( tokenUrl, isDefault = false ) {
+	registerTokenUrl( tokenUrl ) {
+		// Reuse Token instance in case of multiple features using the same tokenUrl.
 		if ( this._tokens.has( tokenUrl ) ) {
 			return Promise.resolve( this.getTokenFor( tokenUrl ) );
 		}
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-cloud-services/src/cloudservices.js
```diff
@@ -38,6 +39,14 @@ export default class CloudServices extends ContextPlugin {
 			this[ optionName ] = options[ optionName ];
 		}
 
+		/**
+		 * Map of `Token` object instances keyed by `tokenUrl`s.
+		 *
+		 * @private
+		 * @type {Map.<String, module:cloud-services-core/token~Token>}
```
ma2ciek, line 46: This accepts a string or a function as the key (which is actually a bit odd :smile:)

packages/ckeditor5-cloud-services/src/cloudservices.js
```diff
@@ -67,9 +76,61 @@ export default class CloudServices extends ContextPlugin {
...
-		this.token = new CloudServices.Token( this.tokenUrl );
+		return this.registerTokenUrl( this.tokenUrl, true );
+	}
+
+	/**
+	 * Registers an additional authentication token URL for CKEditor Cloud Services or a callback to the token value promise. See the
+	 * {@link module:cloud-services/cloudservices~CloudServicesConfig#tokenUrl} for more details.
+	 *
+	 * @param {String|Function} tokenUrl The authentication token URL for CKEditor Cloud Services or a callback to the token value promise.
+	 * @param {Boolean} [isDefault=false] Whether this should be a default authentication token provider.
```
ma2ciek, line 87: It actually doesn't make sense to expose this flag as only the default value from the plugin initialization can use it.

So I'd try to remove the additional handling of the default token from this function.

packages/ckeditor5-cloud-services/src/cloudservices.js
niegowski: Unfortunately, then it displays in docs like this: 

```
Map.&lt(String|Function), module:cloud-services-core/token~Token>
```

niegowski: but with current notation it's:

niegowski:
```
Map.<String, Token>
```

and `Token` is linking

packages/ckeditor5-cloud-services/src/cloudservices.js
```diff
@@ -67,9 +76,61 @@ export default class CloudServices extends ContextPlugin {
...
+	 * @returns {Promise.<module:cloud-services-core/token~Token>}
+	 */
+	registerTokenUrl( tokenUrl, isDefault = false ) {
+		if ( this._tokens.has( tokenUrl ) ) {
+			return Promise.resolve( this.getTokenFor( tokenUrl ) );
```
ma2ciek, line 92: Should not it throw an error here?

ma2ciek (changes requested):
A few minor comments.

packages/ckeditor5-cloud-services/src/cloudservices.js
niegowski: Nope, it's reusing `Token` instance registered by the other feature - same `tokenUrl` = same token.

packages/ckeditor5-cloud-services/src/cloudservices.js
ma2ciek: Ok. Let's go with this option then.

packages/ckeditor5-cloud-services/src/cloudservices.js
niegowski: I'll add code comment there

ma2ciek: Ok. This makes sense if two different features share the same configuration.

ma2ciek (approved):

PR #8550, opened 2020-11-30
Tooltips in HTML Embed should not affect editable area size

### LINKED ISSUE #8340: Editable gets a weird extra scroll when it ends with a small HTML embed

1.  Create content that and with an HTML embed that has only e.g. one character.
2.  Put the selection in a content preceding that embed.

### ✔️ Expected result

The editable should have no scroll as it's not styled with any overflow and height limitations.

### ❌ Actual result

Bonus point. In our docs the scroll appears only when the widget is not selected. In manual tests I can see that the scroll changes its height but it's shown in both cases.

## 📃 Other details

This happens on every browser. 

I tried to figure out what's leaking out of the embed but I couldn't. My wild guess is that it may be something with using flex, but I can be very wrong.

---

niegowski:

Fix (html-embed): Tooltips in HTML Embed should not affect the editable area size. Closes #8340.

Feature (theme-lark): Added east and west positions for tooltips.

Fix (engine): The fake content selection container should not affect CSS `:last-child` styling in content (use `.ck-content > :not(.ck-fake-selection-container):last-of-type` instead).

---

### Additional information


oleq (changes requested):
Now it gets a weird horizontal scroll instead :P

Screenshot 2020-12-01 at 11 22 10

niegowski:
> Now it gets a weird horizontal scroll instead :P

🤦‍♂️

niegowski:
> Now it gets a weird horizontal scroll instead :P

Should be OK now 😅

PR #8563, opened 2020-12-02
Docs: Update element-reconversion.md - correct code in _defineUI() function

thepriefy:
### Additional information

I follow the example and find out that the code in the example is incorrect.


godai78 (approved):
Looks OK.

PR #8569, opened 2020-12-02
Blockquote should be split on backspace key press. Empty block element at the beginning of the limit element should be converted to a paragraph on backspace key press.

### LINKED ISSUE #7636: How to add text before a blockquote?
If I have a blockquote at the top of the document, I seem to be stuck. No amount of Shift+Enter lets me escape the blockquote.

Perhaps blockquotes should have magic block buttons?

### LINKED ISSUE #8137: Pressing backspace at the beginning of the first block should reset it to a paragraph

Pressing backspace here should turn the first line into a paragraph.

### ✔️ Expected result

The first line is turned into a paragraph.

### ❌ Actual result

The first line remains a heading.

## 📃 Other details

We already support pressing Backspace in this case (when the editor is empty):

```html
<root>
	<heading1>[]</heading1>
</root>
```

But to me, it does not really matter what's in the next line. When I press backspace here I'd expect to see a paragraph. 

---

### LINKED ISSUE #8706: Can't delete a list after widget element using backspace

1. Go to https://ckeditor.com/ckeditor-5/demo/
2. Paste an image
3. Below the image, create a checkbox with any text
4. Go to the beginning of the checkbox text and press Backspace

### ✔️ Expected result

The checkbox becomes a normal paragraph.

### ❌ Actual result

Nothing is deleted, and the image is selected.

Note that everything works as expected without an image.

## 📃 Other details

* Browser: Chrome
* OS: macOS Catalina

---

niegowski:

Feature (typing): Empty block element at the beginning of the limit element should be converted to a paragraph on backspace key press. Closes #8137.

Feature (block-quote): Block-quote should be split on backspace key press at the beginning of the block-quote. Closes #7636.

Fix (list): The `delete` event handler listening on a higher priority to avoid intercepting by the block-quote and widget handler. Closes #8706.

---

### Additional information


niegowski:
@ckeditor/qa-team could you please test it?

FilipTokarski:
Checked in various browsers, I didn't find any bugs. Seems to work fine 👍   
  
The only thing that caught my attention, is that changing _heading_ -> _paragraph_ is not recognized by Track Changes ( _block quote_ works ok ), but I guess this is a case for a follow-up.

niegowski:
> The only thing that caught my attention, is that changing _heading_ -> _paragraph_ is not recognized by Track Changes ( _block quote_ works ok ), but I guess this is a case for a follow-up.

I changed a bit the way it's handled, now it should properly integrate with TC.

FilipTokarski:
Yes, now it shows a proper suggestion 👍

packages/ckeditor5-list/src/listediting.js
```diff
@@ -168,7 +169,7 @@ export default class ListEditing extends Plugin {
 
 			data.preventDefault();
 			evt.stop();
-		}, { priority: 'high' } );
```
Reinmar, line 171: Why? It's useful if such quirks are documented in the code.

packages/ckeditor5-block-quote/src/blockquoteediting.js
```diff
@@ -102,31 +103,49 @@ export default class BlockQuoteEditing extends Plugin {
...
 				evt.stop();
 			}
-		} );
+		}, { priority: priorities.normal - 10 } );
+
+		// Overwrite default Backspace key behavior.
+		// If Backspace key is pressed with selection collapsed in first empty block inside a quote, break the quote.
+		//
+		// Priority high + 5 to override widget's feature listener but not list's feature listener.
+		this.listenTo( viewDocument, 'delete', ( evt, data ) => {
```
Reinmar, line 135: Do you think this applies to blockQuotes only? Maybe it'd be simpler if this was generalized a bit and put directly in the delete handler?

packages/ckeditor5-block-quote/src/blockquoteediting.js
```diff
@@ -102,31 +103,49 @@ export default class BlockQuoteEditing extends Plugin {
...
 		// If Enter key is pressed with selection collapsed in empty block inside a quote, break the quote.
-		// This listener is added in afterInit in order to register it after list's feature listener.
-		// We can't use a priority for this, because 'low' is already used by the enter feature, unless
-		// we'd use numeric priority in this case.
-		this.listenTo( this.editor.editing.view.document, 'enter', ( evt, data ) => {
-			const doc = this.editor.model.document;
-			const positionParent = doc.selection.getLastPosition().parent;
+		//
+		// Priority normal - 10 to override default handler but not list's feature listener.
+		this.listenTo( viewDocument, 'enter', ( evt, data ) => {
```
Reinmar, line 115: Why did you need to change anything in the Enter key support in this PR? This PR from its history and description was all about the Backspace key, so this is confusing.

Reinmar:
I left a couple of questions and suggestions.

I'm mostly worried about:

*   why some things were changed/needed,
*   whether we couldn't make them simpler (these priorities thing is best to avoid).

Reinmar (changes requested):
As commented.

packages/ckeditor5-block-quote/src/blockquoteediting.js
niegowski: It seems more reasonable to use `-10` instead of using `afterInit` just to order listeners.

packages/ckeditor5-block-quote/src/blockquoteediting.js
niegowski: Unfortunately, it's a different case for list items (that are not wrapped) and blockQuotes (that are wrapped).

packages/ckeditor5-list/src/listediting.js
niegowski: I added a code comment.

niegowski:
As we discussed F2F I created a followup: #8640

packages/ckeditor5-list/tests/listediting.js
```diff
@@ -291,6 +291,21 @@ describe( 'ListEditing', () => {
...
+			sinon.spy( editor, 'execute' );
+
+			setModelData(
+				model,
+				'<paragraph>x</paragraph><blockQuote><listItem listType="bulleted" listIndent="0">[]</listItem></blockQuote>'
+			);
+
+			editor.editing.view.document.fire( 'delete', domEvtDataStub );
+
+			sinon.assert.calledWithExactly( editor.execute, 'outdentList' );
```
Reinmar, line 307: I know you were consistent with all the other tests here, but just to clarify – I think all of them should be checking the actual content of the model too, not only whether the command was executed. We don't know if some other handler didn't react or the command didn't do something unexpected (quite probable when changing its behavior).

Issue #8517, opened 2020-11-24
Implement DLLs for CKEditor 5 features


jodator:
## 📝 Provide a description of the new feature

As the DLL spike research proved the approach, we will implement the DLLs for the CKEditor 5 as multiple DLLs.

During F2F talks with @Reinmar & @pomek, we've identified that implementing DLLs might be created on multiple levels to provide reuse of core CKEditor packages and to minimize, if possible, the number of `<script>` tags needed to insert into the document. To do that, we would choose among below 3 options:

1st option:

*   base DLL (core + utils + engine + ui + widget? + essentials? + paragraph?): `ckeditor5`
*   creator DLL (classic, balloon, ...): `@ckeditor/ckeditor5-editor-*`
*   other package DLLs (image, basic-styles, ...): `@ckeditor/ckeditor5-*`

2nd option:

*   base DLL (core + utils + engine + UI): `ckeditor5`
*   creator DLL (classic, balloon, ... + + widget + essentials + paragraph + basic-styles + image, etc.): `@ckeditor/ckeditor5-editor-*`
*   other package DLLs (image, basic-styles, ...): `@ckeditor/ckeditor5-*`

3rd option:

*   base DLL
*   build DLL
*   other package DLLs, not included in build DLL

Potential issues:

*   Angular, React, Vue: integrations must not import anything.
*   Collaboration features.
*   How to import from base DLL as sane imports (ideally the same as we use today). We should not alter current imports to not break current builds (tests, manual tests, etc.).

Additionally, I see option 0: every package is a DLL.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

jodator:
OH, I've just stumbled on first obstacle, that might be hard to overcome in a granular approach. It looks like intermediate DLL is not possible. So it's either a DLL or a DLL consumer.

If I cannot overcome this issue, I'll probably make a shortlist of CKEditor plugins that need to be in every build.

jodator:
At this stage, using top-level webpack's plugins for DLLs I don't see that we could implement options 0-3.

### Problem 1: No DLL in other DLL

[...]

PR #8592, opened 2020-12-07
Move DLL definition to main CKEditor5 package.

### LINKED ISSUE #8578: Check if DLL can be exposed in ckeditor5 main repository

At this point, I think that we need to validate the last problem, which is - where to expose those paths (questions 2 & 3 in https://github.com/ckeditor/ckeditor5/issues/8517#issuecomment-737719336). The idea is to add a bigger reason for the `ckeditor5` package. It would also make imports a bit more sane, in that term that it would provide imports as:

```javascript
import { Plugin } from 'ckeditor5/core';
```

At the time of writing this issue, the #8517 issue branch uses `ckeditor5-dll` to provide DLL build. The imports paths look like below:

```javascript
import { Plugin } from '@ckeditor/ckeditor5-core/core`
```

This is because we would like to keep the ability to have imports to work the same way as in today's build using direct imports.

This of course requires an entry file in each package that exports public-public API:

```javascript
// file `core.js` in the `ckeditor5-core` package direclyt (not in /src).
export { default as Plugin } from './src/plugin';
export { default as Command } from './src/command';
export { default as Editor } from './src/editor/editor';
```

This is also duplicated by the `core.js` file inside `ckeditor5-dll` package:

```javascript
export * from '@ckeditor/ckeditor5-core/core';
```

---

Things to validate when moving those files from `ckeditor5-dll` to `ckeditor5`:

*   Does the DLL is built the same way?
*   I think that we need to change from where we're importing stuff. So every import form `@ckeditor/ckeditor5-feature` should be changed to `ckeditor5/feature`.
*   Do the yarn/npm workspace handle cyclic import?
    *    We put every package to the `ckeditor5/packages` but every package might also require one `ckeditor5` instead of the current mix of packages (mostly `uitls`, `ui` & `core`).
*   Do the test, manual tests, and docs works?

I do not consider this as a blocker to the main task, rather as an improvement.

jodator:

Internal: Move DLL definition to main CKEditor5 package. Closes #8578.

---

### Additional information

This PR adds a meaning to the ckeditor5 package :wink: 

AFICS this setup works:

1.  The `ckeditor5` package has `src` dir with public-public API of base/core CKEditor5 packages as identified in the #8517 task. So `ckeditor5-core` re-exports are in `src/core.js` file.
2.  The dependant plugin, verified with `BasicStyles` package, defines dependency as: `ckeditor5: "^23.1.0"` .
3.  The `webpack.dll.config` is simpler now and has only one entry: 

```javascript
new webpack.DllReferencePlugin( {
	manifest: require( '../../build/ckeditor5-dll.manifest.json' ),
	scope: 'ckeditor5/src'
} )
```

AFAICS the `yarn install worked`.


jodator:
@ckeditor/qa-team  I'll be working on this a bit more to verify other packages. But for now, could you also verify that it works for you?

1.  Please use clean install (no previous `node_modules`).
2.  Do `yarn install` as normal.
3.  Run `yarn run build:dll` task in the main folder.
4.  Run `webpack` in `ckeditor5/sample/plugin` folder.
5.  Open the `ckeditor5/sample/index.html` in the browser.

If the plugins `HtmlEmbed` and `Image` fail, please remove them from the sample `index.html`.

jodator:
@pomek please take a look at this as it might invalidate your task about automation. At least in some sort.

pomek (approved):
LGTM.

Issue #8598, opened 2020-12-07
Set the default language of WProofreader to EN_AI on our demos


wwalc:
Page: https://ckeditor.com/docs/ckeditor5/latest/features/spelling-and-grammar-checking.html
Language code: `EN_AI` https://docs.webspellchecker.net/display/WebSpellCheckerServer55x/Default+languages+list
Option name: `lang` https://webspellchecker.com/docs/api/wscbundle/Options.html

![Screen Shot 2020-12-07 at 2 55 23 PM](https://user-images.githubusercontent.com/545175/101359545-b7860780-389c-11eb-8f49-e496b7f55f7a.png)

godai78:
I made the changes, but can't seem to get it to work and test it on my local server (didn't work before changes either). @ckeditor/qa-team can you have a look? Working branch is /docs-spelling-ai.

godai78:
Thanks @Mgsy for workaround. Pushed.

PR #8603, opened 2020-12-08
Docs: changing spellcheck language.

godai78:
Closes: #8598

Updating web spell checker language.

Additional: multiline wrap option.


AnnaTomanek:
The WProofreader part is fine :+1: 

I'm not sure about the multiline toolbar one though - the toolbar is pretty short still and the guides do various strange things depending on the page width (collapsing the sidebar and TOC) so the result for me was mostly that if I managed to hit a width where the toolbar finally wrapped, the second line just contained 1 or 2 buttons. So nothing spectacular.

I'd say multiline toolbar makes more sense in the declarative option, i.e. when you have a toolbar that contains a lot of things, and you just declare how it should split in advance.

godai78:
Let's drop the multiline then for this one.

AnnaTomanek (approved):
Thanks!

PR #8620, opened 2020-12-09
Improved performance of Differ#getChanges().

### LINKED ISSUE #8188: Formatting large content freezes the editor

1. Add some large content to editor ( I'm using the one from [pagination demo](https://ckeditor.com/docs/ckeditor5/latest/features/pagination.html#demo) )
2. Select all
3. Change font size/ apply bold / apply some other formatting

### ✔️ Expected result

Formatting happens immediately.

### ❌ Actual result

Editor freezes, sometimes for more than ~30 seconds:  
  
0_freeze1

## 📃 Other details

* Browser: Chrome, Firefox
* On Safari the freeze is almost unnoticeable
* CKEditor version: old issue, happens even on v15

---

niegowski:

Other (engine): Improved performance of `Differ#getChanges()`. Closes #8188.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/model/differ.js
+++ b/packages/ckeditor5-engine/src/model/differ.js
@@ -393,7 +393,7 @@ export default class Differ {
 		}
 
 		// Will contain returned results.
-		const diffSet = [];
+		let diffSet = [];
 
 		// Check all changed elements.
 		for ( const element of this._changesInElement.keys() ) {
@@ -483,8 +483,8 @@ export default class Differ {
 		} );
 
 		// Glue together multiple changes (mostly on text nodes).
-		for ( let i = 1; i < diffSet.length; i++ ) {
-			const prevDiff = diffSet[ i - 1 ];
+		for ( let i = 1, prevIndex = 0; i < diffSet.length; i++ ) {
+			const prevDiff = diffSet[ prevIndex ];
 			const thisDiff = diffSet[ i ];
 
 			// Glue remove changes if they happen on text on same position.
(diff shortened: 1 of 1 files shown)


niegowski:
Before this change `getChanges()` for a test content lasted for ~600ms but after this change, it lasts ~150ms.

@ckeditor/qa-team could you test it if you see performance improvement? Also, it never took more than 900ms for me so the case of 30s freeze is a mystery, could you try to find a case when it takes so long if it still happens?

Reinmar:
@niegowski, could you provide more information about the change? What did you identify as slow and what did you do with it. It's hard to review without the context.

niegowski:
> @niegowski, could you provide more information about the change? What did you identify as slow and what did you do with it. It's hard to review without the context.

I found out that in the case of the pagination demo content there is no need for post-fixing because the content is proper from the beginning (all post-fixers return `false` - they made no changes). While checking why it's taking so long (the one round of post-fixing) I noticed that the very first post-fixer is spending most of its time in `Differ#getChanges()` which took most of the time by itself so the long task was right inside `getChanges`. After putting performance tags between the biggest parts of that method I found that "gluing" changes were taking the biggest part of the method call time. It was iterating over a 26k diff entries and for every one of them that was glued with the previous one the merged entry was removed from the array (this was causing shifting elements in an array).

FilipTokarski:
In manual test I got max ~500ms, on local docs max ~530ms.  In comparison, on `master` in manual test I got ~1.18s, the same on local docs. So there's definitely an noticeable improvement here 👍 
The only thing that bothers me, is that this very long freeze I reported, happens for me only on live docs.

scofalik:
In my opinion the change is okay.  
  
@FilipTokarski

> The only thing that bothers me, is that this very long freeze I reported, happens for me only on live docs.

Considering this, do you see this issue as resolved or not? If yes, the PR can be merged.

FilipTokarski:
The originally reported case is tricky, because it can only be reproduced on live docs. So I cannot state now with 100% certainty, whether it will fix the issue once it goes live.
OTOH, changes proposed by @niegowski improve the performance of this particular case quite substantially, when tested locally, so I think we can go with this.

niegowski:
> The originally reported case is tricky, because it can only be reproduced on live docs. So I cannot state now with 100% certainty, whether it will fix the issue once it goes live.  
> OTOH, changes proposed by @niegowski improve the performance of this particular case quite substantially, when tested locally, so I think we can go with this.

I just tested it in live docs by replacing snippet (with Chrome Dev tools overrides) with the one built locally and it seems that some other performance improvement also helped there - it takes same time as in local docs.

PR #8627, opened 2020-12-10
Add DLL modules to the global scope

### LINKED ISSUE #8521: Implement the CKEditor5 global object based off the CKEditor5 DLLs

It would be nice to have all (DLLs and plugins) exposed as UMD global (or just global).

*   Every package has its `index.js` (probably also defined in `package.json` if needed - other plugins _might_ use it to import it).
*   Webpack plugin to extends `index.js` so the exports are exposed as `window.CKEditor5[ packageName ]`.
    *   Additionally, we should move other globals (version, translations) there. (🎉 breaking change).
    *   We can also do runtime checks for:
        *   versions
        *   required DLLs (core, widget, ui, etc) based on dependencies listed in package.jsons of these packages
    *   In the end, we have:
        *   `window.CKEditor5.htmlEmbed.HtmlEmbed`
        *   `window.CKEditor5.htmlEmbed.HtmlEmbedCommand`
        *   `window.CKEditor5.htmlEmbed.HtmlEmbedUI`
        *   `window.CKEditor5.htmlEmbed.HtmlEmbedEditing`

To check: `CKEditor5` or `CKE5`? Definitely not `CKEDITOR` to avoid conflicts with CKEditor 4's global.

If changes in other packages (version, translations) might be done separately then it would be even better. I think that we might forget about something. Additionally, other improvements might also be done in a follow-up.

---

pomek:

Internal: DLL modules will be added to the global scope. Closes #8521.

### DIFF
--- a/scripts/dll/dll-loader.js
+++ b/scripts/dll/dll-loader.js
@@ -38,7 +38,7 @@ module.exports = function dllLoader( source, map ) {
 				// Define the global scope.
 				'window.CKEditor5 = window.CKEditor5 || {};',
 				// Load modules into the global scope using webpack loader.
-				`window.CKEditor5.${ windowScope } = CKEditor5_DLL( './src/${ scope }.js' );`
+				`window.CKEditor5.${ windowScope } = window.CKEditor5_DLL( './src/${ scope }.js' );`
 			].join( '' );
 
 			source += attachModuleToGlobalScope;
--- a/scripts/dll/dll-loader.js
+++ b/scripts/dll/dll-loader.js
@@ -1,0 +1,49 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* eslint-env node */
+
+/**
+ * This loader attaches exported modules to the global (`window`) scope.
+ *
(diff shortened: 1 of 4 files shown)


scripts/dll/dll-loader.js
```diff
@@ -0,0 +1,49 @@
...
+		// ...that re-export the public API.
+		if ( this.resourcePath.match( /index.js$/ ) ) {
+			const scope = this.resourcePath.match( /ckeditor5?-([^/\\]+)/ )[ 1 ];
+			const windowScope = scope.replace( /-([a-z])/g, ( match, p1 ) => p1.toUpperCase() );
+
+			const attachModuleToGlobalScope = [
+				// Define the global scope.
+				'window.CKEditor5 = window.CKEditor5 || {};',
+				// Load modules into the global scope using webpack loader.
+				`window.CKEditor5.${ windowScope } = CKEditor5_DLL( './src/${ scope }.js' );`
```
pomek, line 41:
```suggestion
				`window.CKEditor5.${ windowScope } = window.CKEditor5_DLL( './src/${ scope }.js' );`
```
Hm?

jodator: Sure, it's in the `window` scope anyway.

jodator (approved):
Noice :+1: :tada:

PR #8631, opened 2020-12-10
Export and read DLL from the scoped variable

### LINKED ISSUE #8575: Decide on DLL global location

Might be done as part of #8517 or after it.

The #8517 issue branch introduces `window.CKEditor5_DLL` as a DLL approach POC. However, during the work on #8521 we discovered that we can easily sarge globals using array syntax for `entry` webpack's configuration option.

It should be fairly easy to expose it as `window.CKEditor5.DLL()` or anything else. But first, we should:

1.  Validate that this is possible with `DLLPlugin` & `DLLReferencePlugin`.
2.  Decide the name for `DLL()` method on the `CKEditor5` global.

/cc @Reinmar for his 2¢ on the naming, but maybe after the (1) is confirmed.

pomek:

Internal: Export and read DLL from the scoped variable. Closes #8575.

---

### Additional information

Should be merged together with https://github.com/ckeditor/ckeditor5-dev/pull/681.

### DIFF
--- a/docs/builds/guides/development/dll-builds.md
+++ b/docs/builds/guides/development/dll-builds.md
@@ -18,7 +18,7 @@ A minimal example of an editor using `ClassicEditor` build is presented below:
 <body>
 <div id="editor"></div>
 
-<!-- This adds the CKEditor5_DLL in the global scope, so it is possible to reference its contents -->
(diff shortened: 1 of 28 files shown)


jodator (approved):
Cool :sunglasses:

PR #8632, opened 2020-12-10
Plugin collection will allow to require plugin by name when it is provided in `config.plugins` or if it was already loaded.

### LINKED ISSUE #2907: Affect plugins loading order without creating strict requirements
We already touched this topic [here](https://github.com/ckeditor/ckeditor5-core/issues/148), but it was not implemented then.

Currently, when one plugin requires another plugin you can be sure about two things:
 - the required plugin will be loaded,
 - the required plugin will be loaded before your plugin.

However, sometimes we need only the second part. Another plugin is not necessary, but if it is loaded we want it to be initialized before our plugin. Currently `afterInit` needs to be used what is far from a perfect solution. My proposal is to introduce soft requirements:

```js
class A extends Plugin {
	static get softRequires() {
		return [ 'B' ];
	}

	static get pluginName() {
		return 'A';
	}

	init() {
		if( this.editor.plugins.has( 'B' ) {
			console.log( 'B is also loaded!' );
		}
	}
}

class B extends Plugin {
	static get pluginName() {
		return 'B';
	}
}
```

It means that if the plugin `B` is not loaded then plugin `A` shows no errors nor warnings. However, if plugin `B` is loaded, then it is initialized before plugin `A`, the same way, as it would be a requirement.

Note, that I am not sure about "Soft requirements" name nor about `softRequires` method name. Feel free to propose a better name.

jodator:

Internal: Plugin collection will allow to require plugin by name when it is provided in `config.plugins` or if it was already loaded. Closes #2907.

---

### Additional information

*   The code is not super beautiful.
*   I could see that the plugin loading could be improved by listing all required plugins before trying to load them (follow-up reported here: #8634).


packages/ckeditor5-easy-image/src/easyimage.js
```diff
@@ -18,9 +18,12 @@ import CloudServicesUploadAdapter from './cloudservicesuploadadapter';
  *
  * This is a "glue" plugin which enables:
  *
+ * * {@link module:easy-image/cloudservicesuploadadapter~CloudServicesUploadAdapter}.
+ *
+ * This plugin requires plugin to be present in the editor configuration:
```
pomek, line 23: WDYT about `@dependency` annotation that could be rendered somehow by our documentation tool?

```js
/**
 * ...
 * @dependency {module:image/image~Image}
 * @dependency {module:image/imageupload~ImageUpload}
 * ...
 */
```

I am pretty sure that the dependency list mentioned in the API docs will be missed. As for now, it's ok but the mechanism could be improved in the future.

pomek (approved):
LGTM.

packages/ckeditor5-easy-image/src/easyimage.js
pomek: Extracted to https://github.com/ckeditor/ckeditor5/issues/8645.

Issue #8633, opened 2020-12-10
Autoformat creates code block when typing in bulleted/numbered lists


maxbarnas:
## 📝 Provide detailed reproduction steps (if any)

1.  Create a bulleted/numbered list (from the toolbar or using `1.` / `*` ).
2.  Type ` ``` `.

### ✔️ Expected result

The list has an item with three grave accents.

### ❌ Actual result

The item gets removed and a code block is inserted below.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

PR #8646, opened 2020-12-11
Autoformat will insert horizontal line when line starts with `---`.

### LINKED ISSUE #5720: Autoformat --- for horizontal rule

Typing `---` should insert horizontal rule just like in Markdown (btw this is used below in this issue template). This was mentioned in the #1365 but marked as "not a part of MVP".

---

maxbarnas:

Feature (autoformat): Insert a horizontal line when the line starts with `---`. Closes #5720.

Docs (autoformat, horizontal-line): Updated feature documentation.

---

### Additional information

Right now there is a clash with automatic text transformation when typing three dashes, although it is limited as horizontal line shortcut will be triggered only when `---` are at the beginning of the element.

Also, there is a possible cross-fix with a bug: #8633. Right now both the ` ``` ` and `---` can be created inside the numbered/bulleted list. I can either fix it here, along with incorrect tests, or fix it in separate PR and then update tests for horizontal line autoformatting.


oleq (changes requested):
1. Can you resolve merge conflicts? Looks like we had some changes since this PR was created.
2. There's an unnecessary paragraph left when adding a new HR via `---`:
2.1. `<paragraph>[]</paragraph>` (start with an empty editor)
2.2. `<paragraph>--[]</paragraph>` (type two dashes)
2.3. type the third `-`
2.4. Actual `<paragraph></paragraph><hl /><paragraph>[]</paragraph>`
2.5. Expected `<hl /><paragraph>[]</paragraph>` (like in Notion)

The same happens when instead of `<paragraph>` the initial selection is in `<heading>`. This does not happen if the HL is inserted via the toolbar button into an empty editor, though, so we should figure out what causes it.

maxbarnas:
Seems like the extra paragraph is the artifact of the autoformat. You can replicate it by typing anything in the empty paragraph or heading and clicking the horizontal line button on the toolbar. Will check the autoformat for mitigating that. 

In the meantime, I fixed the conflicts :tada:

oleq:
> Seems like the extra paragraph is the artifact of the autoformat.

I suspected this is the culprit. OTOH we've never had autoformatting content that works like the horizontal line, I mean, it is not editable unlike headings, paragraphs, blockquotes tables, and so on. So maybe there should be some changes in the autoformatting to detect using schema whether it is so (whether it accepts `$text`)? I didn't check the logic there, just guessing. 

BTW: The same would happen if we allowed page breaks in autoformatting. Or any "uneditable" widget made by the 3rd-party integrators.

maxbarnas:
> So maybe there should be some changes in the autoformatting to detect using schema whether it is so (whether it accepts $text)?
Autoformat seems very dumb in this particular case because it only executes a command. All the logic regarding adding extra paragraphs is done inside `HorizontalLineCommand`. I could always add options to stop the command from doing so, but I am not sure it's a good idea.

oleq:
> Autoformat seems very dumb in this particular case because it only executes a command. All the logic regarding adding extra paragraphs is done inside `HorizontalLineCommand`. I could always add options to stop the command from doing so, but I am not sure it's a good idea.

Me neither. Maybe autofromat should get rid of `---` before executing the command then? It comes down to [this part.](https://github.com/ckeditor/ckeditor5/blob/i/5720/packages/ckeditor5-autoformat/src/blockautoformatediting.js#L134-L139) First, the command gets executed, then there's a clean-up. It makes sense for commands like headings because they rename elements (in place) and the clean-up can be done after that. 

In the case of the horizontal line, it does not work because the HL is created after the element that contained `---` and then there's the new paragraph after it. I mean, it does the clean-up but the empty paragraph remains. Maybe after `writer.remove( range );` we should check if the block the `range` was in is now empty and the document selection went somewhere else and just get rid of this block?

oleq:
Maybe something like this

```javascript
		// Use enqueueChange to create new batch to separate typing batch from the auto-format changes.
		editor.model.enqueueChange( writer => {
			// Matched range.
			const start = writer.createPositionAt( blockToFormat, 0 );
			const end = writer.createPositionAt( blockToFormat, match[ 0 ].length );
			const range = new LiveRange( start, end );

			const wasChanged = callback( { match } );

			// Remove matched text.
			if ( wasChanged !== false ) {
				writer.remove( range );

				const selectionRange = editor.model.document.selection.getFirstRange();

				// If the block is empty and the document selection is somewhere else.
				if ( blockToFormat.isEmpty && !writer.createRangeIn( blockToFormat ).containsRange( selectionRange, true ) ) {
					writer.remove( blockToFormat );
				}
			}

			range.detach();
		} );
```

oleq:
This should be working fine

```javascript
const selectionRange = editor.model.document.selection.getFirstRange();
const blockRange = writer.createRangeIn( blockToFormat );

// If the block is empty and the document selection is somewhere else.
if ( blockToFormat.isEmpty && !blockRange.isEqual( selectionRange ) && !blockRange.containsRange( selectionRange, true ) ) {
	writer.remove( blockToFormat );
}
```

oleq:
There's something weird going on in this case, though (when you backspace after the autoformat, the HL becomes "dead"):

oleq:
Aaaaah... now I see. In your manual test, there's no `Widget` plugin now and it looks like neither HL nor Page break require it on their own. Basically, a twin to #8720 that we fixed in this iteration.

Can you create an issue for this, fix it and add it to this iteration @maxbarnas?

maxbarnas:
> This should be working fine
> 
> ```js
> const selectionRange = editor.model.document.selection.getFirstRange();
> const blockRange = writer.createRangeIn( blockToFormat );
> 
> // If the block is empty and the document selection is somewhere else.
> if ( blockToFormat.isEmpty && !blockRange.isEqual( selectionRange ) && !blockRange.containsRange( selectionRange, true ) ) {
> 	writer.remove( blockToFormat );
> }
> ```

With this code, I can't add HL when inside a list item, but I can with the toolbar icon.

> Can you create an issue for this, fix it and add it to this iteration @maxbarnas?

Sure, doing it right now.

oleq:
> With this code, I can't add HL when inside a list item, but I can with the toolbar icon.

Are you sure? I think I didn't work before either. Anyway, this code may need some improvements and tests will reveal that.

maxbarnas:
### Extra paragraph after inserting HL

```
lorem ipsum
--[]
lorem ipsum
```

When you type either `-` (CKE5, Notion) or `-<Space>` (Slack PE), both Slack and Notion will make it look like:

```
lorem ipsum
-------------------------------...
[]
lorem ipsum
```

But CKE5 will behave like this:

```
lorem ipsum
-------------------------------...
[]lorem ipsum
```
Should I try to replicate that?

### Creating HL from within a list

1. Slack will remove `---` at first, then on another try just go to the next line (no HL will get created).
2. Notion will replace the list item with HL.
3. CKE5:
	a. Autoformat will do nothing,
	b. HL triggered from the toolbar will replace a list item with HL.

Considering the inconsistency across editors, I would try to replicate the behavior of HL created from the toolbar. WDYT?

oleq:
> But CKE5 will behave like this:

It is different but, I see no reason why this would be wrong. Let's leave it.

> ### **Creating HL from within a list**

Can you create a follow-up for this? We don't need to focus on this aspect in this iteration (progressive enhancement).

PR #8648, opened 2020-12-11
Improvements in chat with mentions

### LINKED ISSUE #8147: Small improvements to chat with mentions

I've played with ["chat with mentions"](https://ckeditor.com/docs/ckeditor5/latest/examples/framework/chat-with-mentions.html) a bit and I think that adding two additions could make this example a bit closer to a real-world application:

1.  Clear editor contents on "send" (`editor.setData( '' )` should be fine)
2.  Do not send an empty message (Probably checking with [editor.getData()](http://localhost:63341/ckeditor5/build/docs/ckeditor5/latest/api/module_core_editor_utils_dataapimixin-DataApi.html#function-getData) - I don't recall anything in particular here.

magda-chrzescian:
Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Docs (mention): Clearing up and focusing the editor after sending a message, preventing the empty message from being sent. Closes #8147.

### DIFF
--- a/packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.html
+++ b/packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.html
@@ -119,6 +119,10 @@
 		border-top-right-radius: 0;
 	}
 
+	.chat .chat__editor.highlighted + .ck.ck-editor .ck-content {
+		animation: highlight 600ms ease-out;
+	}
+
 	/* ---- In–editor mention list --------------------------------------------------------------- */
 
 	.ck-mentions .mention__item {
--- a/packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.html
+++ b/packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.html
@@ -119,7 +119,7 @@
 		border-top-right-radius: 0;
 	}
 
-	.chat .chat__editor.highlighted + .ck.ck-editor .ck-content {
+	.chat .chat__editor + .ck.ck-editor .ck-content.highlighted {
 		animation: highlight 600ms ease-out;
 	}
 
--- a/packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.js
+++ b/packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.js
@@ -81,10 +81,14 @@ ClassicEditor
 			if ( !message ) {
 				editingView.change( writer => {
 					writer.addClass( 'highlighted', rootElement );
-					setTimeout( () => writer.removeClass( 'highlighted', rootElement ), 650 );
-
 					editingView.focus();
 				} );
+
+				setTimeout( () => {
+					editingView.change( writer => {
+						writer.removeClass( 'highlighted', rootElement );
+					} );
+				}, 650 );
 
 				return;
 			}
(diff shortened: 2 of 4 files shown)


packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.html
```diff
@@ -119,6 +119,10 @@
 		border-top-right-radius: 0;
 	}
 
+	.chat .chat__editor.highlighted + .ck.ck-editor .ck-content {
```
oleq, line 122: I see that you probably did this `+` selector hack because the editor engine was overriding the CSS class that you set on the editing root straight away. You can get rid of the hack, check my later comment.

packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.js
```diff
@@ -73,6 +73,19 @@ ClassicEditor
...
 		// and append to the chat list.
 		document.querySelector( '.chat-send' ).addEventListener( 'click', () => {
+			const message = editor.getData();
+
+			if ( !message ) {
+				const editorContainer = editor.sourceElement;
+
+				editor.editing.view.focus();
+
+				editorContainer.classList.add( 'highlighted' );
```
oleq, line 83: You should do something like this instead:

```javascript
const editingView = editor.editing.view;

editingView.change( writer => {
	writer.addClass( 'highlighted', editingView.document.getRoot() );
} );
```

Once the editor starts, you should not use low-level DOM methods to modify its editing root (e.g. add or remove attributes) because they are under the control of the editor engine. Check out the `DowncastWriter` API for more methods (e.g. to remove the class) and docs.

oleq, line 83: You should do something like this instead:

```javascript
const editingView = editor.editing.view;

editingView.change( writer => {
	writer.addClass( 'highlighted', editingView.document.getRoot() );
} );
```

Once the editor starts, you should not use low-level DOM methods to modify its editing root (e.g. add or remove attributes) because they are under the control of the editor engine. Check out the `DowncastWriter` API for more methods (e.g. to remove the class) and docs.

oleq (changes requested):
UX is OK. Some concerns regarding the implementation, though.

P.S. Don't forget to update the guide .md file with the content of the snippet.

packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.html
magda-chrzescian: Actually, I just followed the convention. 

```
.chat .chat__editor + .ck.ck-editor {
    margin-top: 0;
}

.chat .chat__editor + .ck.ck-editor .ck.ck-toolbar {
    border-top-left-radius: 0;
    border-top-right-radius: 0;
}
```

I wasn't sure why `+` was used in existing styles and assumed there must be some reason for this, i.ex. to avoid the collisions if more editors would be added on the same page.

But indeed, I also didn't want to mess up with the content of the editor.

packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.js
magda-chrzescian: Fixed

PR #8649, opened 2020-12-11
Items baked into editor build can be removed from the toolbar now.

### LINKED ISSUE #7945: removeFromToolbar in config
I have custom build with all tools and buttons that I need. But In some cases I don't want `link` button to be shown. I need option `removeFromToolbar` for that. Otherwise I have to rewrite my whole `toolbar` config, which can change in time.

BTW. I think that option `removePlugins` is not working. I'm adding 
```
removePlugins: [
   'Bold'
],
```
and `bold` is working as always. But still (as in docs) `removePlugins` does not remove corresponding buttons.

maxbarnas:

Feature (ui): Items baked into editor bundles can be removed from the toolbar now. Closes #7945.

Docs: Updated docs/builds/guides/integration/configuration.md with new configuration setting.

MINOR BREAKING CHANGE (ui): Configuration passed to `ToolbarView.fillFromConfig()` will be stripped off of leading, trailing, and duplicated separators (`'|'` and `'-'`).

---

### Additional information

The `ToolbarView.fillFromConfig()` should accept two types of parameters, an array, and an object. This is done to prevent breaking changes in the public API.

### DIFF
--- a/docs/builds/guides/integration/configuration.md
+++ b/docs/builds/guides/integration/configuration.md
@@ -51,7 +51,7 @@ ClassicEditor
 	} );
 ```
 <info-box>
-	Be careful when removing plugins from CKEditor builds using `config.removePlugins`. If removed plugins were providing toolbar buttons, the default toolbar configuration included in a build will become invalid. In such case you need to provide the {@link features/toolbar updated toolbar configuration} as in the example above.
+	Be careful when removing plugins from CKEditor builds using `config.removePlugins`. If removed plugins were providing toolbar buttons, the default toolbar configuration included in a build will become invalid. In such case you need to provide the {@link features/toolbar updated toolbar configuration} as in the example above or by providing only items that need to be removed using `config.toolbar.removeItems`.
 </info-box>
 
 ### List of plugins
--- a/docs/features/toolbar.md
+++ b/docs/features/toolbar.md
@@ -71,6 +71,8 @@ toolbar: {
 
  * **`items`** &ndash; An array of toolbar item names. Most of the components (buttons, dropdowns, etc.) which can be used as toolbar items are described under the {@link features/index Features} tab. A full list is defined in {@link module:ui/componentfactory~ComponentFactory editor.ui.componentFactory} and can be listed using the following snippet: `Array.from( editor.ui.componentFactory.names() )`. Besides button names, you can also use the dedicated separators for toolbar groups (`'|'`) and toolbar lines (`'-'`).
 
+ * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar items of a bundle without the need of recreating it from scratch. If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
+
  * **`viewportTopOffset`** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar. Useful when a page with which the editor is being integrated has some other sticky or fixed elements (e.g. the top menu). Thanks to setting the toolbar offset, the toolbar will not be positioned underneath or above the page's UI.
 
  * **`shouldNotGroupWhenFull`** &ndash; When set to `true`, the toolbar will stop grouping items and let them wrap to the next line when there is not enough space to display them in a single row. This setting is `false` by default, which enables items grouping.
--- a/docs/features/toolbar.md
+++ b/docs/features/toolbar.md
@@ -71,7 +71,7 @@ toolbar: {
 
  * **`items`** &ndash; An array of toolbar item names. Most of the components (buttons, dropdowns, etc.) which can be used as toolbar items are described under the {@link features/index Features} tab. A full list is defined in {@link module:ui/componentfactory~ComponentFactory editor.ui.componentFactory} and can be listed using the following snippet: `Array.from( editor.ui.componentFactory.names() )`. Besides button names, you can also use the dedicated separators for toolbar groups (`'|'`) and toolbar lines (`'-'`).
 
- * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar items of a bundle without the need of recreating it from scratch. If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
+ * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar configuration without the need of defining the entire list (you can specify a couple of buttons that you want to remove instead of specifying all the buttons you want to keep). If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
 
  * **`viewportTopOffset`** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar. Useful when a page with which the editor is being integrated has some other sticky or fixed elements (e.g. the top menu). Thanks to setting the toolbar offset, the toolbar will not be positioned underneath or above the page's UI.
 
--- a/packages/ckeditor5-ui/src/toolbar/toolbarview.js
+++ b/packages/ckeditor5-ui/src/toolbar/toolbarview.js
@@ -287,14 +287,15 @@ export default class ToolbarView extends View {
 		// Items listed in `config.removeItems` should not be added to the toolbar.
 		const items = config.items.filter( name => config.removeItems.indexOf( name ) === -1 );
 
-		this.items.addMany( items.map( ( name, idx ) => {
+		const itemsToAdd = items.reduce( ( itemsToAdd, name, idx, allItems ) => {
 			if ( name === '|' ) {
 				// Omit duplicated separators. This can happen after removing items listed in `config.removeItems`.
-				if ( idx > 0 && items[ idx - 1 ] === '|' ) {
-					return;
+				if ( idx > 0 && allItems[ idx - 1 ] === '|' ) {
+					// No changes to item list.
+					return itemsToAdd;
 				}
 
-				return new ToolbarSeparatorView();
+				itemsToAdd.push( new ToolbarSeparatorView() );
 			} else if ( name === '-' ) {
 				if ( this.options.shouldGroupWhenFull ) {
 					/**
@@ -315,8 +316,10 @@ export default class ToolbarView extends View {
 					 */
 					logWarning( 'toolbarview-line-break-ignored-when-grouping-items', items );
 				}
+
+				itemsToAdd.push( new ToolbarLineBreakView() );
 			} else if ( factory.has( name ) ) {
-				return factory.create( name );
+				itemsToAdd.push( factory.create( name ) );
 			} else {
 				/**
 				 * There was a problem processing the configuration of the toolbar. The item with the given
@@ -337,7 +340,11 @@ export default class ToolbarView extends View {
 				 */
 				logWarning( 'toolbarview-item-unavailable', { name } );
 			}
-		} ).filter( item => item !== undefined ) );
+
+			return itemsToAdd;
+		}, [] );
+
+		this.items.addMany( itemsToAdd );
 	}
 
 	/**
--- a/packages/ckeditor5-ui/src/toolbar/toolbarview.js
+++ b/packages/ckeditor5-ui/src/toolbar/toolbarview.js
@@ -286,10 +286,10 @@ export default class ToolbarView extends View {
 
 		const itemsToAdd = config.items
 			.filter( name => {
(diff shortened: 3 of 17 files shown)


maxbarnas:
Although in my solution we go through `itemsOrConfig.items` twice, we do that once per editor instance. For that reason, I decided to go with a less efficient, but more readable approach.

docs/builds/guides/integration/configuration.md
```diff
@@ -140,13 +140,16 @@ You can use the extended toolbar configuration format to access additional optio
 ```js
     toolbar: {
         items: [ 'bold', 'italic', '|', 'undo', 'redo', '-', 'numberedList', 'bulletedList' ],
+        removeItems: [ 'link' ],
```
maxbarnas, line 143: There should be spaces in the examples? I had to override my editor settings, because, by default, we use tabs.

packages/ckeditor5-ui/src/toolbar/toolbarview.js
```diff
@@ -277,14 +278,24 @@ export default class ToolbarView extends View {
...
+		this.items.addMany( items.map( ( name, idx ) => {
+			if ( name === '|' ) {
+				// Omit duplicated separators. This can happen after removing items listed in `config.removeItems`.
+				if ( idx > 0 && items[ idx - 1 ] === '|' ) {
+					return;
+				}
+
 				return new ToolbarSeparatorView();
-			} else if ( name == '-' ) {
+			} else if ( name === '-' ) {
```
maxbarnas, line 298: Why there were just double equal signs instead of three?

docs/builds/guides/integration/configuration.md
```diff
@@ -51,7 +51,7 @@ ClassicEditor
 	} );
 ```
 <info-box>
-	Be careful when removing plugins from CKEditor builds using `config.removePlugins`. If removed plugins were providing toolbar buttons, the default toolbar configuration included in a build will become invalid. In such case you need to provide the updated toolbar configuration as in the example above.
+	Be careful when removing plugins from CKEditor builds using `config.removePlugins`. If removed plugins were providing toolbar buttons, the default toolbar configuration included in a build will become invalid. In such case you need to provide the updated toolbar configuration - either as in the example above or by providing only items that need to be removed using `removeItems`.
```
Reinmar, line 54:
```suggestion
	Be careful when removing plugins from CKEditor builds using `config.removePlugins`. If removed plugins were providing toolbar buttons, the default toolbar configuration included in a build will become invalid. In such case you need to provide the updated toolbar configuration - either as in the example above or by providing only items that need to be removed using `config.toolbar.removeItems`.
```

docs/builds/guides/integration/configuration.md
Reinmar: Yup, we use tabs. So perhaps that was misformatted in the past (no eslint for docs unfortunately).

docs/builds/guides/integration/configuration.md
```diff
@@ -140,13 +140,16 @@ You can use the extended toolbar configuration format to access additional optio
...
         items: [ 'bold', 'italic', '|', 'undo', 'redo', '-', 'numberedList', 'bulletedList' ],
+        removeItems: [ 'link' ],
         viewportTopOffset: 30,
         shouldNotGroupWhenFull: true
     }
 ```
 
  * **`items`** &ndash; An array of toolbar item names. Most of the components (buttons, dropdowns, etc.) which can be used as toolbar items are described under the {@link features/index Features} tab. A full list is defined in {@link module:ui/componentfactory~ComponentFactory editor.ui.componentFactory} and can be listed using the following snippet: `Array.from( editor.ui.componentFactory.names() )`. More details could be found in the {@link framework/guides/creating-simple-plugin Creating a simple plugin} guide.
 
+ * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar items of a bundle without the need of recreating it from scratch. If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
```
Reinmar, line 151:
```suggestion
 * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can remove items without recreating the entire items array from scratch. It is especially useful when you want to remove a certain plugin from an editor build that has a predefined toolbar. Moreover, if after removing an item, the toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
```

Reinmar: *   bundle -> build as this is the term we use more often.
*   Separated the last sentence with something that makes it stand out a bit more because.
*   Made it clear when this option is useful to give a bit more context.

Reinmar:
Oh, I wanted to write that you forgot to update the toolbar config typedef but I can see... it's not defined. That it's all inline now. Pity, but let's leave it.

packages/ckeditor5-ui/src/toolbar/toolbarview.js
```diff
@@ -277,14 +278,24 @@ export default class ToolbarView extends View {
...
-	fillFromConfig( config, factory ) {
-		this.items.addMany( config.map( name => {
-			if ( name == '|' ) {
+	fillFromConfig( itemsOrConfig, factory ) {
+		const config = normalizeToolbarConfig( itemsOrConfig );
+
+		// Items listed in `config.removeItems` should not be added to the toolbar.
+		const items = config.items.filter( name => config.removeItems.indexOf( name ) === -1 );
+
+		this.items.addMany( items.map( ( name, idx ) => {
```
Reinmar, line 290: I'd move this `map()` call out of `addMany()`. I thought that `addMany()` accepts a callback at first :D I know it was written this way but it makes sense to change it now.

Also, you can consider using a `reduce()` and not `map()`. Might look better because that empty `return` there is a bit odd now. I actually don't know how it works.

packages/ckeditor5-ui/tests/toolbar/toolbarview.js
```diff
@@ -447,6 +447,47 @@ describe( 'ToolbarView', () => {
...
+				},
+				factory
+			);
+
+			const items = view.items;
+			expect( items ).to.have.length( 1 );
+			expect( items.get( 0 ).name ).to.equal( 'bar' );
+		} );
+
+		it( 'deduplicates consecutive separators after removing items listed in `removeItems`', () => {
```
Reinmar, line 474: Funny, because if for any weird reason someone had two separators already, we're changing the behavior for them. I'd add it to the merge commit message as "MINOR BREAKING CHANGE".

cc @oleq We're changing this because the `config.toolbar.removeItems` option might lead to having two separators one after another if you removed all buttons between them. Do you think it's fine if `fillFromConfig()` makes such an optimization? The other option could be to extract that logic to `normalizeToolbarConfig()`. It'd actually might make more sense there, taken the "normalization" aspect. WDYT, guys?

packages/ckeditor5-ui/src/toolbar/toolbarview.js
maxbarnas: Happy to change that.

packages/ckeditor5-ui/tests/toolbar/toolbarview.js
maxbarnas: On one hand, I am happy to change whatever you all think is the right thing to do. Still learning habits of this tribe. On the other hand, the name suggests it's about normalization, not sanitization/preprocessing.

packages/ckeditor5-ui/src/toolbar/toolbarview.js
maxbarnas: Changed now, but I would discuss this change further.

packages/ckeditor5-ui/tests/toolbar/toolbarview.js
maxbarnas: Or, perhaps, take out the entire logic of handling the config to a separate class. It's not ToolbarViews responsibility to deal with all the config logic. Especially now, when it grew in complexity.

maxbarnas: Entire logic has stayed in the `fillFromConfig()`. We do a few steps of filtering to cover all the cases.

oleq: I'm OK with this optimization. Although I need to say I cannot predict to what extent this is a breaking change for integrators. I guess this could fail if someone relies on the order of buttons when they access them via `editor.ui.view.toolbar.items.get( index )` for some reason.

Also FYI, there are already some micro-optimizations in CSS when it comes to grouping separators in the "kebab" drop-down, for instance, if the first grouped item is a separator it is not rendered in the drop-down (because the toolbar should not start with a separator). Maybe this is the right way to go instead of changing the JS logic? This would be way cheaper and safer I guess. `.ck-toolbar__separator + .ck-toolbar__separator { display: none }`? 

P.S. Since we optimize separators, let's make sure line breaks are also optimized to avoid scenarios like `[ item, '-', '-', item ]`.

maxbarnas: Yup! Can confirm that \`.ck-toolbar\_\_separator + .ck-toolbar\_\_separator { display: none }\` works well.

Reinmar: To me, it makes more sense to filter this in JS whenever we can. No one actually wants two separators one by one. When combinging `toolbar.items` with `toolbar.removeItems`, removing also duplicated and trailed separators seems to be a reasonable optimization.

As for whether this is a breaking change – certainly. But not a major one. I think it's acceptable.

> P.S. Since we optimize separators, let's make sure line breaks are also optimized to avoid scenarios like `[ item, '-', '-', item ]`.

Yeah, makes sense too. Good catch.

maxbarnas: Fixed leading and trailing cases for both of the separators along with duplicates.

docs/features/toolbar.md
```diff
@@ -71,6 +71,8 @@ toolbar: {
 
  * **`items`** &ndash; An array of toolbar item names. Most of the components (buttons, dropdowns, etc.) which can be used as toolbar items are described under the {@link features/index Features} tab. A full list is defined in {@link module:ui/componentfactory~ComponentFactory editor.ui.componentFactory} and can be listed using the following snippet: `Array.from( editor.ui.componentFactory.names() )`. Besides button names, you can also use the dedicated separators for toolbar groups (`'|'`) and toolbar lines (`'-'`).
 
+ * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar items of a bundle without the need of recreating it from scratch. If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
```
Reinmar, line 74:
```suggestion
 * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar configuration without the need of defining the entire list (you can specify a couple of buttons that you want to remove instead of specifying all the buttons you want to keep). If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
```

Reinmar:
At some point this feature stopped working at all:

_(note, to test a build you need to rebuild it by calling `yarn build` in its package)_





I think you should consider adding some integration tests. A situation when unit tests don't catch such a basic usage means that integration tests are a must. You can add some here: https://github.com/ckeditor/ckeditor5/blob/88a6f7b/packages/ckeditor5-build-classic/tests/ckeditor.js#L170

Reinmar (changes requested):
It does not work :(

maxbarnas:
Thanks for catching that and showing me the place for tests. The problem was with an old way of passing toolbar configuration, as seen here: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-editor-classic/src/classiceditorui.js#L151. There should be `this._toolbarConfig` passed to `fillFromConfig`, not just the items.

I'll update that in each build and also write integration tests.

maxbarnas:
I see that the builds have a slightly different set of tests. Should I add those integration tests to each build or it's enough to update only classic one?

Reinmar:
> I see that the builds have a slightly different set of tests. Should I add those integration tests to each build or it's enough to update only the classic one?

One build will be enough. Just for security. It will test how all the things cooperate (e.g. `defaultToolbar` with the config object passed to `create()`). A final, ultimate check that it all works.

More tests should be added to each editor package. These editors differ and I suppose you need to change implementations of all of them. Hence, all need to be covered with tests. In this case, think like in TDD – why changing something if there's no test failing first.

PR #8682, opened 2020-12-15
Loosen up condition to allow Text elements.

### LINKED ISSUE #8179: model.Text cannot be used in consumable

I've just found a funny bug (or oversight) in `ModelConsumable` class. As you know, we support consuming `TextProxy` instances but we do not support consuming actual `Text` nodes :). So, if someone tries to consume text node in their custom converter by hand (something along `consumable.consume( modelItem.getChild( 0 ), 'insert' );` ) it will not work as `getChild()` returns `Text`, not `TextProxy`.

Since `Text` and `TextProxy` have similar interfaces, it should be enough to fix one `if` here:

```
add( item, type ) {
	type = _normalizeConsumableType( type );

	if ( item instanceof TextProxy ) {
		item = this._getSymbolForTextProxy( item );
	}

	if ( !this._consumable.has( item ) ) {
		this._consumable.set( item, new Map() );
	}

	this._consumable.get( item ).set( type, true );
}
```

It might check whether the item is `TextProxy` or `Text` (or use `.is()` -- this was introduced before `.is()` method was available).

maxbarnas:

Fix (engine): `ModelConsumable` now accepts `Text` elements. Closes #8179.

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/modelconsumable.js
+++ b/packages/ckeditor5-engine/src/conversion/modelconsumable.js
@@ -7,8 +7,6 @@
  * @module engine/conversion/modelconsumable
  */
 
-import TextProxy from '../model/textproxy';
-
 /**
  * Manages a list of consumable values for {@link module:engine/model/item~Item model items}.
  *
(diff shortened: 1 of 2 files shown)


Reinmar:
AFAIR we weren't sure whether this direction makes sense and I don't think we'll get back to this, so I'm for closing this PR. WDYT, @maxbarnas ?

maxbarnas:
> AFAIR we weren't sure whether this direction makes sense and I don't think we'll get back to this, so I'm for closing this PR. WDYT, @maxbarnas ?

Yup, it seems not that important after all. I'm fine with closing it.

Issue #8623, opened 2020-12-10
Decide on icons and CSS imports


jodator:
## Provide a description of the task

In the working POC, I didn't change imports of icons and some CSS utils.

We could check if we can re-export them and use them as normal code.

jodator:
For icons, we can create a re-export in core's index.js and use it as other exports:

```javascript
import { Plugin, icons } from 'ckeditor5/src/core';

//...

buttonView.set( {
	label: t( 'Block quote' ),
	icon: icons.quote,				// <- Use single icon here
	tooltip: true,
	isToggleable: true
} );
```

This results in:

The DLL loads icon from DLL (excerpt from a minified view):

```javascript
var n=o(0)
//...
return i.set({label:t("Block quote"),icon:n.icons.quote,tooltip:!0,isToggleable:!0})
```

While in standard build the SVG is inlined in the `core` section (it is also tree-shaked AFAICS):

```
const ul='<svg ...>';

o.set({label:t("Block quote"),icon:ul,tooltip:!0,isToggleable:!0})
```

As for tree-shake. Unused icons were removed in my editor classic build. I see only 10 from ~23 total in the file. (ps.: so only those that I've checked, probably, after updates in the rest of the plugins  all the icons will be used in the build.

psmyrek:
The proposed solution for icons has been merged into `dll-integration` branch.
There are still CSS imports left to figure out.

PR #8690, opened 2020-12-16
Use icons from ckeditor5/src/core not as a direct import

jodator:

Internal: Use icons from the main ckeditor5 package.

---

### Additional information

It turned out to be a quite easy and efficient solution. See https://github.com/ckeditor/ckeditor5/issues/8623#issuecomment-746067537.


Reinmar:
TODO: 

* Replace all remaining core's icon imports to the new format in `src/**` (including CF's repo).
* Replace the same thing in the `iconset.js` manual test so we use the new format also there and hence help remembering that icons need to be added to the core's index file.

psmyrek (approved):
LGTM

psmyrek:
> 
> 
> TODO:
> 
>     * Replace all remaining core's icon imports to the new format in `src/**` (including CF's repo).
> 
>     * Replace the same thing in the `iconset.js` manual test so we use the new format also there and hence help remembering that icons need to be added to the core's index file. (note: also add ckeditor5 as devdep of theme-lark)

Done.

PR #8724, opened 2020-12-28
Provide more context to block formatting function. Update tests.

### LINKED ISSUE #8633: Autoformat creates code block when typing in bulleted/numbered lists

maxbarnas:

Fix (autoformat): Code block characters inside a list will not trigger autoformatting. Closes #8633.

### DIFF
--- a/packages/ckeditor5-autoformat/src/autoformat.js
+++ b/packages/ckeditor5-autoformat/src/autoformat.js
@@ -163,7 +163,14 @@ export default class Autoformat extends Plugin {
 	 */
 	_addCodeBlockAutoformats() {
 		if ( this.editor.commands.get( 'codeBlock' ) ) {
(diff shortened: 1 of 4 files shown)


niegowski (commented):
I’m not sure if this issue is valid, it seems that the behavior described in the issue is an expected one - I checked how this works in Notion and in Slack and it’s the same as current implementation - while typing ``` inside the list item - it’s converted to the code block. Maybe we should only update tests to reflect the real behavior?

Reinmar:
I agree with Kuba.

maxbarnas:
Issue is not valid anymore.

maxbarnas:
Currently, the autoformatting feature works as expected after all.

PR #8726, opened 2020-12-28
I/8664 inline images editing part

### LINKED ISSUE #8664: Inline images - Editing part
Plugin architecture, schema, conversion, and commands, that allow changing of an image type from inline to block and back by the user.

Assumptions:
    1. Let's not duplicate plugins
    2. KISS for now and we may refactor it later in the process

- [x] Transforming a block image to an inline image.
- [x] Transforming an inline image to a block image.

pkwasnik:

Feature (image): Inline images - Editing part. Closes #8664.

BREAKING CHANGE (image): The imageEditing plugin is no longer standalone. It was divided into the imageBlockEditing plugin with some of the old logic and the imageInlineEditing plugin containing the new inline image feature. Now the imageEditing plugin contains common logic for those plugins and is imported by each of them.

BREAKING CHANGE (image): The image plugin now imports both the imageBlock plugin (with previous image behaviour) and the imageInline plugin that introduces inline images.

BREAKING CHANGE (image): The image.converters.srcsetAttributeConverter() and the image.converters.modelToViewAttributeConverter() methods now requires the imageType parameter.

BREAKING CHANGE (image): The createImageViewElement() method was moved from the image.imageediting to the image.utils.

BREAKING CHANGE (image): If the image or the imageInline plugin will be imported, then all images without \<figure> wrapper will be upcasted to inline image type instead of block type.

---

### Additional information


packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -134,3 +164,24 @@ export function createImageViewElement( writer ) {
...
+ */
+function matchImageInsideParent( parentType ) {
+	return element => {
+		const parent = element.parent;
+
+		if ( element.name === 'img' && element.hasAttribute( 'src' ) && // Convert only images with src attribute.
+			parent && parent.name === parentType // Convert only images inside paragraph.
+		) {
+			return { name: true };
+		}
```
niegowski, line 183: This way it's much more readable:

```javascript
		// Convert only images with src attribute.
		if ( !element.is( 'element', 'img' ) || !element.hasAttribute( 'src' ) ) {
			return null;
		}

		// Convert only images inside paragraph.
		if ( !parent || !parent.is( 'element', parentType ) {
			return null;
		}

		return { name: true, attributes: [ 'src' ] };
```

Also, this matcher should indicate that the `src` attribute should be consumed: 

`return { name: true, attributes: [ 'src' ] };` 

And what about inline images inside a list item, or in any other element that allows inline images? According to the schema it's allowed everywhere where the text is allowed.

niegowski: And `parentType` is more like `parentName`.

packages/ckeditor5-image/src/image/imageinlinetoblockcommand.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @extends module:core/command~Command
+ */
+export default class ImageInlineToBlockCommand extends Command {
+	/**
+	 * @inheritDoc
+	 */
+	refresh() {
+		const element = this.editor.model.document.selection.getSelectedElement();
+		this.isEnabled = isElementOfType( element, 'imageInline' );
```
niegowski: Missing blank line.

packages/ckeditor5-image/src/image/imageinlinetoblockcommand.js
```diff
@@ -0,0 +1,39 @@
...
+	 * Executes the command.
+	 *
+	 * @fires execute
+	 */
+	execute( ) {
+		const model = this.editor.model;
+		const selection = model.document.selection;
+		const src = selection.getSelectedElement().getAttribute( 'src' );
+
+		insertImage( model, { src }, selection );
```
niegowski, line 37: This way we are not using `findOptimalInsertionPosition`.

Also, we should apply other attributes (`srcset`, `alt`).

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -65,7 +65,18 @@ export function getSelectedImageWidget( selection ) {
...
+}
+
+/**
+ * Checks if the provided model element is of a given type.
+ *
+ * @param {module:engine/model/element~Element} modelElement
+ * @param {String} type
+ * @returns {Boolean}
+ */
+export function isElementOfType( modelElement, type ) {
```
niegowski: This function should be named `isElementOfName` but I'm not sure if this function is needed at all.

niegowski (changes requested):

packages/ckeditor5-image/src/image/imageinlinetoblockcommand.js
pkwasnik: Yes, but we want to replace `selection.` `findOptimalInsertionPosition`will find another place, not the desired one.

packages/ckeditor5-image/src/image/converters.js
```diff
@@ -77,8 +78,8 @@ export function srcsetAttributeConverter() {
 		}
 
 		const writer = conversionApi.writer;
-		const figure = conversionApi.mapper.toViewElement( data.item );
-		const img = getViewImgFromWidget( figure );
+		const element = conversionApi.mapper.toViewElement( data.item );
+		const img = element.name === 'img' ? element : getViewImgFromWidget( element );
```
niegowski, line 82: Can't this check be inside that helper function?

Also, we should use `element.is( 'element', 'img' )`.

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -58,23 +61,49 @@ export default class ImageEditing extends Plugin {
...
+			.elementToElement( {
+				model: 'image',
+				view: ( modelElement, { writer } ) => toImageWidget( createImageViewElement( writer ), writer, t( 'image widget' ) )
+			} )
+			.elementToElement( {
+				model: 'imageInline',
+				view: ( modelElement, { writer } ) => {
+					const emptyElement = writer.createEmptyElement( 'img' );
+					const element = writer.createContainerElement( 'span' );
+					writer.insert( writer.createPositionAt( element, 0 ), emptyElement );
```
niegowski, line 91:
```suggestion
					const imgElement = writer.createEmptyElement( 'img' );
					const wrapperElement = writer.createContainerElement( 'span' );
					writer.insert( writer.createPositionAt( wrapperElement, 0 ), imgElement );
```

Also it would be good to add some class name to that `span`.

packages/ckeditor5-image/src/image/imageediting.js
niegowski: > And what about inline images inside a list item, or in any other element that allows inline images? According to the schema it's allowed everywhere where the text is allowed.

packages/ckeditor5-image/src/image/imageinlinetoblockcommand.js
niegowski: :+1:

packages/ckeditor5-image/src/image/imageblocktoinlinecommand.js
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		model.change( writer => {
+			const paragraph = writer.createElement( 'paragraph' );
+			const imageInlineElement = writer.createElement( 'imageInline', attrs );
+
+			writer.append( imageInlineElement, paragraph );
+			model.insertContent( paragraph, position );
+
+			if ( imageInlineElement.parent ) {
```
niegowski, line 64: Probably some code comment would be useful here.

packages/ckeditor5-image/src/image/imageinlinetoblockcommand.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module image/image/imagetoinlinecommand
```
niegowski, line 7: Shouldn't this be "imageinlinetoblockcommand"?

packages/ckeditor5-image/src/image/imageblocktoinlinecommand.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module image/image/imagetoinlinecommand
```
niegowski, line 7: Shouldn't this be "imageblocktoinlinecommand"?

packages/ckeditor5-image/src/image/imageediting.js
pkwasnik: > 
> 
> > And what about inline images inside a list item, or in any other element that allows inline images? According to the schema it's allowed everywhere where the text is allowed.
This is the topic for the next tasks. In this ticket the only way to create an inline image is to convert it from block image - so it will always be inserted inside a paragraph.

niegowski: But that paragraph can be easily changed by the user to any other kind of block.

packages/ckeditor5-image/src/image/converters.js
```diff
@@ -107,9 +108,16 @@ export function srcsetAttributeConverter() {
 	}
 }
 
+/**
+ * Converter used to convert a given image attribute from the model to the view.
+ *
+ * @param {String} attributeKey
```
niegowski, line 114: Missing param description.

packages/ckeditor5-image/src/image/imageblocktoinlinecommand.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module image/image/imageinlinetoblockcommand
```
niegowski, line 7:
```suggestion
 * @module image/image/imageblocktoinlinecommand
```

packages/ckeditor5-image/src/image/imageblocktoinlinecommand.js
```diff
@@ -0,0 +1,68 @@
...
+ * @extends module:core/command~Command
+ */
+export default class ImageBlockToInlineCommand extends Command {
+	/**
+	 * @inheritDoc
+	 */
+	refresh() {
+		const element = this.editor.model.document.selection.getSelectedElement();
+
+		this.isEnabled = isImage( element );
```
niegowski, line 26: Maybe this command should be disabled also if the schema does not allow paragraphs with inline images in the selection position?

packages/ckeditor5-image/src/image/imageblocktoinlinecommand.js
```diff
@@ -0,0 +1,68 @@
...
+	 * @fires execute
+	 */
+	execute() {
+		const model = this.editor.model;
+		const selection = model.document.selection;
+		const imageElement = selection.getSelectedElement();
+		const src = imageElement.getAttribute( 'src' );
+		const alt = imageElement.getAttribute( 'alt' );
+		const srcset = imageElement.getAttribute( 'srcset' );
+		const position = findOptimalInsertionPosition( selection, model );
```
niegowski, line 41: Maybe we don't need to search for an optimal insertion position - the selection is already on the block image widget so we could just replace it with a paragraph with an inline image inside it. Also `model.insertContent` would handle removing old content for us.

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -58,23 +61,47 @@ export default class ImageEditing extends Plugin {
...
+		conversion.for( 'editingDowncast' )
+			.elementToElement( {
+				model: 'image',
+				view: ( modelElement, { writer } ) => toImageWidget(
+					createImageViewElement( writer, 'image' ), writer, t( 'image widget' )
+				)
+			} )
+			.elementToElement( {
+				model: 'imageInline',
+				view: ( modelElement, { writer } ) => toWidget(
```
niegowski, line 90: Shouldn't `toImageWidget` be used here as for block images?

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -125,12 +154,37 @@ export default class ImageEditing extends Plugin {
 //
 // @private
 // @param {module:engine/view/downcastwriter~DowncastWriter} writer
+// @param {String} imageType
 // @returns {module:engine/view/containerelement~ContainerElement}
-export function createImageViewElement( writer ) {
+export function createImageViewElement( writer, imageType ) {
+	const parentName = imageType === 'image' ? 'figure' : 'span';
```
niegowski, line 160: Maybe `imageType` should be "block" or "inline" to make it clear?

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -125,12 +154,37 @@ export default class ImageEditing extends Plugin {
...
 	const emptyElement = writer.createEmptyElement( 'img' );
-	const figure = writer.createContainerElement( 'figure', { class: 'image' } );
+	const figure = writer.createContainerElement( parentName, { class: imageType } );
 
 	writer.insert( writer.createPositionAt( figure, 0 ), emptyElement );
 
 	return figure;
 }
+
+// {@link module:engine/view/matcher~Matcher} pattern. Returns function which checks if a given element is `<image>` element that is placed
```
niegowski, line 169:
```suggestion
// {@link module:engine/view/matcher~Matcher} pattern. Returns function which checks if a given element is `<img>` element that is placed
```

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -125,12 +154,37 @@ export default class ImageEditing extends Plugin {
...
+// inside the element of a provided type.
+//
+// @param {String} parentType
```
niegowski, line 172:
```suggestion
// @param {String} parentName
```

packages/ckeditor5-image/src/image/imageinlinetoblockcommand.js
```diff
@@ -0,0 +1,57 @@
...
+ * @extends module:core/command~Command
+ */
+export default class ImageInlineToBlockCommand extends Command {
+	/**
+	 * @inheritDoc
+	 */
+	refresh() {
+		const element = this.editor.model.document.selection.getSelectedElement();
+
+		this.isEnabled = isImageInline( element );
```
niegowski, line 25: Maybe this command should be disabled if a block image is not allowed by the schema?

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -120,6 +130,10 @@ export function isImageAllowed( model ) {
  * @returns {module:engine/view/element~Element}
  */
 export function getViewImgFromWidget( figureView ) {
+	if ( figureView.is( 'element', 'img' ) ) {
```
niegowski, line 133: JSDoc should be updated for this function.

packages/ckeditor5-image/src/image/imageblocktoinlinecommand.js
pkwasnik: But if in this position there is a block image, then always also paragraph with inline image will be allowed - or there is any exception?

packages/ckeditor5-image/src/image/imageinlinetoblockcommand.js
pkwasnik: answered here: https://github.com/ckeditor/ckeditor5/pull/8726#discussion_r550138725

packages/ckeditor5-image/src/image/imageblocktoinlinecommand.js
niegowski: In the current schema configuration that's true, but some integration/feature might disable some cases.

packages/ckeditor5-image/tests/image/imageediting.js
```diff
@@ -295,6 +430,17 @@ describe( 'ImageEditing', () => {
 					.to.equal(
 						'<image alt="alt text" src="/assets/sample.png" srcset="{"data":"small.png 148w, big.png 1024w"}"></image>'
 					);
+
+				editor.setData(
+					'<p><img src="/assets/sample.png" alt="alt text" srcset="small.png 148w, big.png 1024w" /></p>'
+				);
+
+				/* eslint-disable max-len */
```
niegowski, line 438: Maybe it would be better to split the lines below instead? I can see there are multiple places where you disable this liner rule.

packages/ckeditor5-image/tests/image/imageinlinetoblockcommand.js
```diff
@@ -0,0 +1,125 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import VirtualTestEditor from '@ckeditor/ckeditor5-core/tests/_utils/virtualtesteditor';
+import { setData as setModelData, getData as getModelData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
+import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
+
```
niegowski, line 9:
```suggestion
import VirtualTestEditor from '@ckeditor/ckeditor5-core/tests/_utils/virtualtesteditor';
import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
import { setData as setModelData, getData as getModelData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
```

packages/ckeditor5-image/tests/image/imageinlinetoblockcommand.js
```diff
@@ -0,0 +1,125 @@
...
+				plugins: [ Image, Paragraph ]
+			} )
+			.then( newEditor => {
+				editor = newEditor;
+				model = editor.model;
+
+				command = new ImageInlineToBlockCommand( editor );
+
+				const schema = model.schema;
+				schema.extend( 'image', { allowAttributes: 'uploadId' } );
```
niegowski, line 28: Same as for the other introduced command.

packages/ckeditor5-image/tests/image/imageinlinetoblockcommand.js
```diff
@@ -0,0 +1,125 @@
...
+		} );
+
+		it( 'should convert inline image with alt and srcset attributes to block image', () => {
+			const imgSrc = 'foo/bar.jpg';
+
+			setModelData( model, `<paragraph>[<imageInline alt="alt text" src="${ imgSrc }" srcset="{}"></imageInline>]</paragraph>` );
+
+			command.execute();
+
+			expect( getModelData( model ) ).to.equal( `[<image alt="alt text" src="${ imgSrc }" srcset="{}"></image>]` );
```
niegowski, line 114: Maybe some more verbose `srcset` check? Just to be sure that the same data is passed.

packages/ckeditor5-image/src/image/converters.js
```diff
@@ -111,7 +111,7 @@ export function srcsetAttributeConverter() {
 /**
  * Converter used to convert a given image attribute from the model to the view.
  *
- * @param {String} attributeKey
+ * @param {String} [attributeKey] name of the attribute to convert
```
niegowski, line 114:
```suggestion
 * @param {String} attributeKey The name of the attribute to convert.
```

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -154,22 +153,23 @@ export default class ImageEditing extends Plugin {
 //
 // @private
 // @param {module:engine/view/downcastwriter~DowncastWriter} writer
-// @param {String} imageType
+// @param {String} [imageType] type of created image: 'block' or 'inline'
```
niegowski, line 156:
```suggestion
// @param {'block'|'inline'} imageType The type of created image.
```

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -154,22 +153,23 @@ export default class ImageEditing extends Plugin {
...
 // @returns {module:engine/view/containerelement~ContainerElement}
 export function createImageViewElement( writer, imageType ) {
-	const parentName = imageType === 'image' ? 'figure' : 'span';
+	const parentName = imageType === 'block' ? 'figure' : 'span';
+	const className = imageType === 'block' ? 'image' : 'image-inline';
 	const emptyElement = writer.createEmptyElement( 'img' );
-	const figure = writer.createContainerElement( parentName, { class: imageType } );
+	const figure = writer.createContainerElement( parentName, { class: className } );
```
niegowski: Maybe this way it would be clearer?
```javascript
const figure = imageType == 'block' ? 
    writer.createContainerElement( 'figure', { class: 'image' } ) : 
    writer.createContainerElement( 'span', { class: 'image-inline' } );
```

packages/ckeditor5-image/tests/image/imageblocktoinlinecommand.js
```diff
@@ -0,0 +1,129 @@
...
+
+			expect( getModelData( model ) ).to.equal(
+				`<paragraph>[<imageInline src="${ imgSrc }"></imageInline>]</paragraph>`
+			);
+		} );
+
+		it( 'should convert block image with alt and srcset attributes to inline image', () => {
+			const imgSrc = 'foo/bar.jpg';
+
+			setModelData( model, `[<image src="${ imgSrc }" alt="alt text" srcset="{}"></image>]` );
```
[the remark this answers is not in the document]
pkwasnik: Here we are operating only on model, so just want to copy srcset. There are already another tests in imageediting.js to deep check srcset conversion.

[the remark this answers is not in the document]
niegowski: This command is replacing block image with an inline image so we must be sure if it was replaced correctly.

packages/ckeditor5-image/tests/image/imageinlinetoblockcommand.js
```diff
@@ -21,7 +20,7 @@ describe( 'ImageInlineToBlockCommand', () => {
 				editor = newEditor;
 				model = editor.model;
 
-				command = new ImageInlineToBlockCommand( editor );
+				command = editor.commands.get( 'imageInlineToBlock' );
 
 				const schema = model.schema;
 				schema.extend( 'image', { allowAttributes: 'uploadId' } );
```
niegowski: Is this needed?

packages/ckeditor5-image/src/image/imageblocktoinlinecommand.js
pkwasnik: I was wrong in the previous message. In the selected position the converted image will never be allowed, but the conversion command modifies the DOM to allow it. In the block to inline conversion, it adds `paragraph` to allow inserting `inlineImage`. So proposed checking makes no sense.

packages/ckeditor5-image/src/image/imageinlinetoblockcommand.js
pkwasnik: updated comment linked above (similar situation)

packages/ckeditor5-image/tests/manual/imageinline.html
```diff
@@ -0,0 +1,19 @@
...
+</head>
+
+<script src="https://ckeditor.com/apps/ckfinder/3.5.0/ckfinder.js"></script>
+
+<div id="editor">
+	<h2>Inline image</h2>
+	<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</p>
+	<figure class="image ck-widget" contenteditable="false">
+		<img src="https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg" alt="Example image">
+		<figcaption class="ck-editor__editable ck-editor__nested-editable ck-hidden ck-placeholder" contenteditable="true" data-placeholder="Enter image caption">
```
niegowski, line 12: This should be a clean html so `ck-widget`, `contenteditable="false"`, `ck-editor__editable ck-editor__nested-editable ck-hidden ck-placeholder`, `data-placeholder` should not be there.

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -98,17 +98,12 @@ export default class ImageEditing extends Plugin {
 
 		conversion.for( 'upcast' )
 			.elementToElement( {
-				view: matchImageInsideParent( 'p' ),
-				model: ( viewImage, { writer } ) => writer.createElement( 'imageInline', { src: viewImage.getAttribute( 'src' ) } )
+				view: matchImageByType( false ),
```
niegowski, line 101: I would replace `true`/`false` with `'inline'`/`'block'` to make this code clear without checking this function signature.

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -181,8 +176,15 @@ function matchImageInsideParent( parentName ) {
...
 
-		// Convert only images inside paragraph.
-		if ( !parent || !parent.is( 'element', parentName ) ) {
+		const isBlockImage = ( parent &&
+			( parent.is( 'element', 'figure' ) ||
+				( parent.is( 'element', 'a' ) && parent.parent.is( 'element', 'figure' ) )
+			)
+		);
+
+		const incorrectParent = matchInlineImage ? isBlockImage : !isBlockImage;
```
niegowski, line 185: This is hard to read, please try to break it into a more readable form.

niegowski: Also, I think that it should not check for the `'a'` element, it should find the closest ancestor that is a figure with the required class name ([`findAncestor`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_element-Element.html#function-findAncestor)).

niegowski (changes requested):

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -58,16 +60,36 @@ export default class ImageEditing extends Plugin {
 			allowWhere: '$block',
 			allowAttributes: [ 'alt', 'src', 'srcset' ]
 		} );
-
-		conversion.for( 'dataDowncast' ).elementToElement( {
-			model: 'image',
-			view: ( modelElement, { writer } ) => createImageViewElement( writer )
+		schema.register( 'imageInline', {
```
oleq, line 63: Missing new line before `schema.register()`.

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -58,16 +60,36 @@ export default class ImageEditing extends Plugin {
...
+			.elementToElement( {
+				model: 'image',
+				view: ( modelElement, { writer } ) => toImageWidget(
+					createImageViewElement( writer, 'block' ), writer, t( 'image widget' )
+				)
+			} )
+			.elementToElement( {
+				model: 'imageInline',
+				view: ( modelElement, { writer } ) => toImageWidget(
+					createImageViewElement( writer, 'inline' ), writer, t( 'image widget' )
```
oleq, line 90:
```suggestion
					createImageViewElement( writer, 'inline' ), writer, t( 'inline image widget' )
```

packages/ckeditor5-image/src/image/imageediting.js
oleq: I agree. Alternatively:

*   you could use `'image'` and `'imageInline'` strings which are everywhere anyway,
*   there could be `const IMAGE_TYPE_INLINE = 1` and `const IMAGE_TYPE_BLOCK = 2` declared at the top of the file and used here (or anywhere else),

packages/ckeditor5-image/src/image/imageediting.js
oleq: I second that this is hard to read. This should be more like this:

```
if ( matchInlineImage ) {
	if ( ... ) {
		// ...
	}
	
	// ...
	
	if ( ... ) {
		return null;
	}
} else {
	if ( ... ) {
		// ...
	}
	
	// ...
	
	if ( ... ) {
		return null;
	}
}
```

Where did this `'a'` parent check come from BTW?

packages/ckeditor5-image/src/image/imageediting.js
oleq: This notation should also be shared with `createImageViewElement()`.

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -114,6 +135,8 @@ export default class ImageEditing extends Plugin {
 			.add( viewFigureToModel() );
 
 		editor.commands.add( 'imageInsert', new ImageInsertCommand( editor ) );
+		editor.commands.add( 'imageBlockToInline', new ImageBlockToInlineCommand( this.editor ) );
+		editor.commands.add( 'imageInlineToBlock', new ImageInlineToBlockCommand( this.editor ) );
```
oleq, line 139: Why do we need 2 separate commands for this? 

This is a simple state machine and `ImageTypeToggleCommand()` should do the trick. I checked implementations and they are fairly simple. The unified command:

*   would be enabled if the selected model element is either `image` or `imageInline`,
*   would have `value` indicating which one is it (e.g. for the UI to bind to), either `'image'` or `'imageInline'`
*   would have `execute( options )` that toggles the kind of image
    *   (in the future) could support `{String} options.type` that allows integrators to enforce `image` or `imageInline` regardless of the current model state

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -125,12 +148,46 @@ export default class ImageEditing extends Plugin {
...
-	return figure;
+	return container;
+}
+
+// {@link module:engine/view/matcher~Matcher} pattern. Returns function which checks if a given element is `<img>` element that is placed
+// inside the element of a provided type.
+//
+// @param {String} parentName
+// @returns {Function}
+function matchImageByType( matchInlineImage ) {
```
oleq, line 170: This function does not match but it returns a matcher callback. It should be called `getImageTypeMatcher()` (or something like this).

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -125,12 +148,46 @@ export default class ImageEditing extends Plugin {
...
+	writer.insert( writer.createPositionAt( container, 0 ), emptyElement );
 
-	return figure;
+	return container;
+}
+
+// {@link module:engine/view/matcher~Matcher} pattern. Returns function which checks if a given element is `<img>` element that is placed
+// inside the element of a provided type.
+//
+// @param {String} parentName
```
oleq, line 168: The param name does not correspond to the argument in `function( ... )`

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -125,12 +148,46 @@ export default class ImageEditing extends Plugin {
...
+		writer.createContainerElement( 'span', { class: 'image-inline' } );
+
+	writer.insert( writer.createPositionAt( container, 0 ), emptyElement );
 
-	return figure;
+	return container;
+}
+
+// {@link module:engine/view/matcher~Matcher} pattern. Returns function which checks if a given element is `<img>` element that is placed
+// inside the element of a provided type.
```
oleq, line 166: I think this description could be simpler:

```suggestion
// A function returning a {@link module:engine/view/matcher~Matcher} callback for a particular type of View images. 
```

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -116,10 +126,16 @@ export function isImageAllowed( model ) {
  *
  * The `<img>` can be wrapped to other elements, e.g. `<a>`. Nested check required.
  *
+ * If provided `figureView` is an <img> element, than there is no need for extracting it, so this element will be returned.
```
oleq, line 129: I think this explanation is unnecessary, it duplicates a simple code (DRY).

oleq (changes requested):

packages/ckeditor5-image/src/image/imageediting.js
pkwasnik: `a` came from link tests :) When image has a link so there is `<figure><a><image></image></a></figure>` then `figure` is no longer a direct ancestor of an `image`

oleq (changes requested):
Update: We realized that current integrations must have some way to preserve the UX where inline images are disallowed. We can do this either via configuration or via plugin granularity and we figured the latter reflects the spirit of the project better. 

* `Image` will require `ImageBlock` and `ImageInline`. These two plugins will work as current `ImageEditing` except that for different kinds of images. 
* `Image` works as a top-level glue.
* Each of the two will require `[ Widget, ImageTextAlternative ]` so they can work standalone (without `Image`).
* Developers who upgrade the editor will get the new feature (inline images) alongside the old one (block images) because they import `Image`. Most won't mind but for some, this is a no-go.
* Those who want the old feature set (only block, no inline images) will need to update their configs and replace `Image` with `ImageBlock` (needs to be documented in the guide).
* Those who want only inline images will import `ImageInline` instead of `Image` (needs to be documented in the guide).

In theory, we could split every single feature and have `ImageInlineTextAlternative` or `ImageInlineResize`. But this will bloat the codebase and these features can discover what is enabled or not via `editor.plugins.has( 'ImageBlock' )` (or schema) and control their internal flow.

This is a breaking change because there will be no longer `ImageEditing` but I think this has never been a standalone plugin that a lot of integrators used directly.

packages/ckeditor5-ckfinder/src/ckfinderediting.js
```diff
@@ -31,7 +32,7 @@ export default class CKFinderEditing extends Plugin {
 	 * @inheritDoc
 	 */
 	static get requires() {
-		return [ Notification, ImageEditing, LinkEditing ];
+		return [ Notification, ImageEditing, ImageBlock, LinkEditing ];
```
niegowski, line 35: Why `ImageBlock` is needed here? To keep backward compatibility? But what if some integrator will want to use only `ImageInline`? Maybe this shouldn't be here.

packages/ckeditor5-image/src/image.js
```diff
@@ -21,7 +23,8 @@ import '../theme/image.css';
  *
  * This is a "glue" plugin which loads the following plugins:
  *
- * * {@link module:image/image/imageediting~ImageEditing},
+ * * {@link module:image/image/imageblock~ImageBlock},
+ * * {@link module:image/image/imageinline~ImageInline},
```
niegowski, line 27: I would mention `ImageEditing` also.

packages/ckeditor5-image/src/image/converters.js
```diff
@@ -69,6 +69,7 @@ export function viewFigureToModel() {
 export function srcsetAttributeConverter() {
 	return dispatcher => {
 		dispatcher.on( 'attribute:srcset:image', converter );
+		dispatcher.on( 'attribute:srcset:imageInline', converter );
```
niegowski, line 72: Those should be registered separately by `ImageBlock` and `ImageInline`.

packages/ckeditor5-image/src/image/converters.js
```diff
@@ -107,9 +108,16 @@ export function srcsetAttributeConverter() {
...
+/**
+ * Converter used to convert a given image attribute from the model to the view.
+ *
+ * @param {String} attributeKey The name of the attribute to convert.
+ * @returns {Function}
+ */
 export function modelToViewAttributeConverter( attributeKey ) {
 	return dispatcher => {
 		dispatcher.on( `attribute:${ attributeKey }:image`, converter );
+		dispatcher.on( `attribute:${ attributeKey }:imageInline`, converter );
```
niegowski, line 120: Those should be registered separately by `ImageBlock` and `ImageInline`.

packages/ckeditor5-image/src/image/imageblock.js
```diff
@@ -0,0 +1,73 @@
...
+		const schema = editor.model.schema;
+		const t = editor.t;
+		const conversion = editor.conversion;
+
+		// Configure schema.
+		schema.register( 'image', {
+			isObject: true,
+			isBlock: true,
+			allowWhere: '$block',
+			allowAttributes: [ 'alt', 'src', 'srcset' ]
```
niegowski, line 49: This plugin should also register converters for those attributes.

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -44,46 +42,17 @@ export default class ImageEditing extends Plugin {
...
 		conversion.for( 'upcast' )
-			.elementToElement( {
-				view: {
-					name: 'img',
-					attributes: {
-						src: true
-					}
-				},
-				model: ( viewImage, { writer } ) => writer.createElement( 'image', { src: viewImage.getAttribute( 'src' ) } )
-			} )
```
niegowski, line 86: Shouldn't this be in the `ImageBlock` to be able to upcast `img` that is not wrapped with `figure`? Also AFAIR this is used by `viewFigureToModel()` to upcast `img` inside figure.

packages/ckeditor5-image/src/image/imageediting.js
```diff
@@ -110,27 +79,9 @@ export default class ImageEditing extends Plugin {
 						return value;
 					}
 				}
-			} )
-			.add( viewFigureToModel() );
+			} );
 
 		editor.commands.add( 'imageInsert', new ImageInsertCommand( editor ) );
```
niegowski, line 84: This command is inserting a block image or an inline image? Is it supposed to handle both types?

packages/ckeditor5-image/src/image/imageinline.js
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * @module image/image/imageinline
+ */
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+
+import { toImageWidget, createImageViewElement, getImageTypeMatcher } from './utils';
+
+/**
+ * The image engine plugin.
```
niegowski, line 15: Maybe this should mention that it's for inline images?

packages/ckeditor5-image/src/image/imageinline.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * It registers:
+ *
+ * * `<image>` as a block element in the document schema, and allows `alt`, `src` and `srcset` attributes.
+ * * converters for editing and data pipelines.
+ * * `'imageInsert'` command.
```
niegowski, line 21: This is not true.

packages/ckeditor5-image/src/image/imageinline.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @extends module:core/plugin~Plugin
+ */
+export default class ImageInline extends Plugin {
```
niegowski, line 25: This plugin should require `Widget` and `ImageTextAlternative`. And also maybe `ImageEditing` in case some integrator would use only `ImageInline` without `Image` itself?

packages/ckeditor5-image/src/image/imageblock.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * It registers:
+ *
+ * * `<image>` as a block element in the document schema, and allows `alt`, `src` and `srcset` attributes.
+ * * converters for editing and data pipelines.
+ * * `'imageInsert'` command.
+ *
+ * @extends module:core/plugin~Plugin
+ */
+export default class ImageBlock extends Plugin {
```
niegowski, line 27: This plugin should require `Widget` and `ImageTextAlternative`. And also maybe `ImageEditing` in case some integrator would use only `ImageBlock` without `Image` itself?

packages/ckeditor5-image/src/image/imageinline.js
```diff
@@ -0,0 +1,70 @@
...
+		const editor = this.editor;
+		const schema = editor.model.schema;
+		const t = editor.t;
+		const conversion = editor.conversion;
+
+		schema.register( 'imageInline', {
+			isObject: true,
+			isInline: true,
+			allowWhere: '$text',
+			allowAttributes: [ 'alt', 'src', 'srcset' ]
```
niegowski, line 46: Converters for those attributes should be registered in this file.

packages/ckeditor5-image/src/image/imageinsertcommand.js
```diff
@@ -14,7 +14,7 @@ import toArray from '@ckeditor/ckeditor5-utils/src/toarray';
 /**
  * Insert image command.
  *
- * The command is registered by the {@link module:image/image/imageediting~ImageEditing} plugin as `'imageInsert'`.
+ * The command is registered by the {@link module:image/image~Image} plugin as `'imageInsert'`.
```
niegowski, line 17: This is not true.

packages/ckeditor5-image/src/image/imagetypetogglecommand.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @module image/image/imagetypetogglecommand
+ */
+
+import Command from '@ckeditor/ckeditor5-core/src/command';
+import { insertImage, isImage, isImageInline } from './utils';
+
+/**
+ * The image inline command. It is used to convert block image io inline image.
```
niegowski, line 14: Please update the description.

packages/ckeditor5-image/src/image/imagetypetogglecommand.js
```diff
@@ -0,0 +1,68 @@
...
+		const element = this.editor.model.document.selection.getSelectedElement();
+
+		this.isEnabled = isImage( element ) || isImageInline( element );
+	}
+
+	/**
+	 * Executes the command.
+	 *
+	 * @fires execute
+	 */
```
niegowski, line 32: This is less descriptive than in the base class so maybe it should be `@inheritDoc` ?

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -133,6 +147,58 @@ export function getViewImgFromWidget( figureView ) {
...
+//
+//		<figure class="image"><img></img></figure>
+//
+// Note that `alt` and `src` attributes are converted separately, so they are not included.
+//
+// @private
+// @param {module:engine/view/downcastwriter~DowncastWriter} writer
+// @param {'image'|'imageInline'} imageType The type of created image.
+// @returns {module:engine/view/containerelement~ContainerElement}
+export function createImageViewElement( writer, imageType ) {
```
niegowski, line 160: This is an exported function so it should have a proper JsDoc. And should not be private. And it should describe the imageInline HTML.

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -133,6 +147,58 @@ export function getViewImgFromWidget( figureView ) {
...
+	writer.insert( writer.createPositionAt( container, 0 ), emptyElement );
+
+	return container;
+}
+
+// A function returning a {@link module:engine/view/matcher~Matcher} callback for a particular type of View images.
+//
+// @param {'image'|'imageInline'} matchImageType The type of created image.
+// @returns {Function}
+export function getImageTypeMatcher( matchImageType, editor ) {
```
niegowski, line 176: This is an exported function so it should have a proper JsDoc.

packages/ckeditor5-image/src/image/imageediting.js
niegowski: Ok I can see that you moved it to the `getImageTypeMatcher` but this is needed for block images.

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -133,6 +147,58 @@ export function getViewImgFromWidget( figureView ) {
...
+	if ( editor.plugins.has( 'ImageInline' ) ^ editor.plugins.has( 'ImageBlock' ) ) {
+		return {
+			name: 'img',
+			attributes: {
+				src: true
+			}
+		};
+	}
```
niegowski, line 184: I don't like using the binary xor operator for logical conditions. 

Also, this upcast conversion is always needed for block image (as mentioned in another comment).

packages/ckeditor5-image/src/imageresize/imageresizeediting.js
```diff
@@ -74,7 +74,8 @@ export default class ImageResizeEditing extends Plugin {
 	 * @private
 	 */
 	_registerSchema() {
-		this.editor.model.schema.extend( 'image', { allowAttributes: 'width' } );
+		const imageType = this.editor.plugins.has( 'ImageBlock' ) ? 'image' : 'imageInline';
+		this.editor.model.schema.extend( imageType, { allowAttributes: 'width' } );
```
niegowski, line 78: What if both plugins are enabled?

packages/ckeditor5-image/src/imagestyle/imagestyleediting.js
```diff
@@ -40,10 +40,11 @@ export default class ImageStyleEditing extends Plugin {
 
 		// Get configuration.
 		const styles = normalizeImageStyles( editor.config.get( 'image.styles' ) );
+		const imageType = this.editor.plugins.has( 'ImageBlock' ) ? 'image' : 'imageInline';
 
 		// Allow imageStyle attribute in image.
 		// We could call it 'style' but https://github.com/ckeditor/ckeditor5-engine/issues/559.
-		schema.extend( 'image', { allowAttributes: 'imageStyle' } );
+		schema.extend( imageType, { allowAttributes: 'imageStyle' } );
```
niegowski, line 47: What if both plugins are enabled?

packages/ckeditor5-image/src/imageupload/imageuploadediting.js
```diff
@@ -60,9 +60,10 @@ export default class ImageUploadEditing extends Plugin {
 		const fileRepository = editor.plugins.get( FileRepository );
 
 		const imageTypes = createImageTypeRegExp( editor.config.get( 'image.upload.types' ) );
+		const imageType = this.editor.plugins.has( 'ImageBlock' ) ? 'image' : 'imageInline';
 
 		// Setup schema to allow uploadId and uploadStatus for images.
-		schema.extend( 'image', {
+		schema.extend( imageType, {
```
niegowski, line 66: What if both plugins are enabled?

packages/ckeditor5-link/src/linkimageediting.js
```diff
@@ -28,7 +30,7 @@ export default class LinkImageEditing extends Plugin {
 	 * @inheritDoc
 	 */
 	static get requires() {
-		return [ ImageEditing, LinkEditing ];
+		return [ ImageEditing, ImageBlock, ImageInline, LinkEditing ];
```
niegowski, line 33: This would disable an option to select only `ImageBlock` or `ImageInline`.

niegowski (changes requested):
As in the comments. Also, CC is not 100% so I wasn't reviewing the tests.

packages/ckeditor5-ckfinder/src/ckfinderediting.js
pkwasnik: Hm, `ImageEditing` needs `ImageBlock` or `ImageInline` to handle images, so I've assumed that it is necessary everywhere where `ImageEditing` is used for compatibility. Do we need to document it for this and similar features or it is obvious that a user will always import Image, ImageBlock, or ImageInline along with it? (btw. tests passes without ImageBlock)

packages/ckeditor5-image/src/image/utils.js
pkwasnik: This upcast could work only when there is one plugin enabled. For both enabled there is another upcast returned below, which differentiates if `img` is block or inline depending on the context. About `xor` operator - without it, the logic will be more complex and this is the perfect situation for using it. Ofc. I could rewrite it, but first, please @oleq confirm that we don't want `xor` operators in the code :)

packages/ckeditor5-ckfinder/src/ckfinderediting.js
niegowski: > *   Those who want the old feature set (only block, no inline images) will need to update their configs and replace `Image` with `ImageBlock` (needs to be documented in the guide).
> *   Those who want only inline images will import `ImageInline` instead of `Image` (needs to be documented in the guide).

pkwasnik: Done.

packages/ckeditor5-image/src/image.js
pkwasnik: Done.

packages/ckeditor5-image/src/image/converters.js
pkwasnik: Done.

packages/ckeditor5-image/src/image/converters.js
pkwasnik: Done.

packages/ckeditor5-image/src/image/imageblock.js
pkwasnik: Done.

packages/ckeditor5-image/src/image/converters.js
```diff
@@ -64,12 +64,12 @@ export function viewFigureToModel() {
 /**
  * Converter used to convert the `srcset` model image attribute to the `srcset`, `sizes` and `width` attributes in the view.
  *
+ * @param {'image'|'imageInline'} imageType The type of created image.
```
niegowski, line 67:
```suggestion
 * @param {'image'|'imageInline'} imageType The type of image.
```

niegowski, line 67:
```suggestion
 * @param {'image'|'imageInline'} imageType The type of the image.
```

niegowski: I'm not sure if it should be "the" or "an".

packages/ckeditor5-image/src/image/converters.js
```diff
@@ -112,12 +112,12 @@ export function srcsetAttributeConverter() {
  * Converter used to convert a given image attribute from the model to the view.
  *
  * @param {String} attributeKey The name of the attribute to convert.
+ * @param {'image'|'imageInline'} imageType The type of created image.
```
niegowski, line 115: Same as above.

packages/ckeditor5-image/src/imageresize/imageresizeediting.js
pkwasnik: done

packages/ckeditor5-image/src/image/imageinline.js
niegowski: OTOH they are identical (converting only `alt` and `srcset` attributes on an `img` element so maybe there should stay in `ImageEditing` and in this file, it should be mentioned where they can be found.

packages/ckeditor5-image/src/image/converters.js
pkwasnik: done

packages/ckeditor5-image/src/image/converters.js
pkwasnik: done

packages/ckeditor5-link/src/linkimageediting.js
pkwasnik: Done.

packages/ckeditor5-image/src/image/imageediting.js
pkwasnik: I am handling this in another ticket about insert/upload images.

packages/ckeditor5-image/src/image/imageediting.js
pkwasnik: This is needed only when block image plugin is alone (and then getImageTypeMatcher will return it). With both plugins enabled, `img` not wrapped with `figure` will be upcasted to inline image.

packages/ckeditor5-image/src/image/utils.js
pkwasnik: Done.

packages/ckeditor5-image/src/image/utils.js
pkwasnik: Done.

niegowski:
I'd like to propose such plugins dependency tree (top level plugins would be just a glue code):

*   Image requires:
    *   ImageEditing
    *   ImageBlock
    *   ImageInline
    *   Widget
    *   ImageTextAlternative
*   ImageBlock requires:
    *   ImageEditing
    *   ImageBlockEditing
    *   Widget
    *   ImageTextAlternative
*   ImageInline requires:
    *   ImageEditing
    *   ImageInlineEditing
    *   Widget
    *   ImageTextAlternative

The glue plugins would be in the `src/` directory, the editing plugins would go to dedicated directories for image, imageblock, imageinline.

packages/ckeditor5-link/tests/linkimageediting.js
```diff
@@ -177,22 +200,22 @@ describe( 'LinkImageEditing', () => {
 			describe( 'a > img', () => {
 				it( 'should convert a link in an image figure', () => {
 					editor.setData(
-						'<a href="http://ckeditor.com"><img src="/assets/sample.png" alt="alt text" /></a>'
+						'<figure class="image"><a href="http://ckeditor.com"><img src="/assets/sample.png" alt="alt text" /></a></figure>'
```
niegowski, line 203: Those tests should use only `ImageBlockEditing` until we add support for linking inline images. Please avoid changing input data.

niegowski (changes requested):

PR #8737, opened 2020-12-30
Add WEBP support to autoimage

kavanagh21:
Other autoimage.js: Adds WEBP support to the inline pasting of images from source URLs.


maxbarnas:
@kavanagh21, thanks for your contribution!

PR #8757, opened 2021-01-07
fix potential null reference

### LINKED ISSUE #8756: null reference
https://github.com/ckeditor/ckeditor5/blob/a26c6532e2090b47816961c8344b258d237f28d0/packages/ckeditor5-table/src/converters/downcast.js#L382

When attempting ctrl+x (cut) on an image inside of a table, `firstChild` === null. When `firstChild` is passed to `hasAnyAttribute()` that null causes an exception. It seems like the safest resolution is to check for `element === null` in `hasAnyAttribute()`

jrit:
Closes #8756

### DIFF
--- a/packages/ckeditor5-table/src/converters/downcast.js
+++ b/packages/ckeditor5-table/src/converters/downcast.js
@@ -520,5 +520,8 @@ function getViewTable( viewFigure ) {
 // @param {module:engine/model/element~Element element
 // @returns {Boolean}
 function hasAnyAttribute( element ) {
+  if(element === null) {
+    return false;
+  }
 	return !![ ...element.getAttributeKeys() ].length;
 }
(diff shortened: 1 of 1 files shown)


jrit:
CI is failing because some lines are not covered by tests due to my changes.

Reinmar:
Hi! Thanks for the PR. In order to merge it we need test coverage. This means not only covering all lines but also showing (by a test) that this change is needed.

You can read about CKEditor 5's dev env and test env here:

* https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/development-environment.html
*  https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/testing-environment.html

mlewand:
Unit coverage is a must have for us. I also mentioned in [https://github.com/ckeditor/ckeditor5/issues/8756#issuecomment-827416855](https://github.com/ckeditor/ckeditor5/issues/8756#issuecomment-827416855) that we're unable to reproduce the issue anymore.

PR #8764, opened 2021-01-07
Do not upcast contents of inner table

### LINKED ISSUE #8393: Loading nested tables with border-bottom style throws JavaScript error

1.  Start editor with the following content -
```
<table>
<tbody>
<tr>
<td> 

<table>
<tbody>
<tr>
<td style="border-bottom: 0 solid #fff;">

</td>
</tr>
</tbody>
</table>

</td>
</tr>
</tbody>
</table>
```

### ✔️ Expected result

Data gets loaded into editor.

### ❌ Actual result

Java Script error is being thrown - 
```
TypeError: Cannot read property 'getAttribute' of undefined
    at setAttributeOnItem (writer.js:1487)
    at Writer.setAttribute (writer.js:380)
    at UpcastDispatcher.<anonymous> (tableproperties.js:73)
    at UpcastDispatcher.fire (emittermixin.js:209)
    at UpcastDispatcher._convertChildren (upcastdispatcher.js:282)
    at UpcastDispatcher.<anonymous> (upcasthelpers.js:446)
    at UpcastDispatcher.fire (emittermixin.js:209)
    at UpcastDispatcher._convertChildren (upcastdispatcher.js:282)
```

## 📃 Other details

Even after enabling nested tables https://github.com/ckeditor/ckeditor5/issues/3232#issuecomment-528251188 the error is still thrown.

*   Browser: Any
*   OS: Any
*   CKEditor version: 23.1.0
*   Installed CKEditor plugins: N/A

---

maxbarnas:

Fix (table): The contents of nested tables are no longer going through upcasting. Closes #8393.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
+++ b/packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
@@ -214,14 +214,16 @@ describe( 'table properties', () => {
 				describe( 'nested tables', () => {
 					// https://github.com/ckeditor/ckeditor5/issues/6177.
 					it( 'should upcast tables with nested tables in their cells', () => {
-						editor.setData( '<table style="border:1px solid red">' +
-						'<tr>' +
-							'<td>parent:00</td>' +
-							'<td>' +
-								'<table style="border:1px solid green"><tr><td>child:00</td></tr></table>' +
-							'</td>' +
-						'</tr>' +
-					'</table>' );
+						editor.setData(
+							'<table style="border:1px solid red">' +
+								'<tr>' +
+									'<td>parent:00</td>' +
+									'<td>' +
+										'<table style="border:1px solid green"><tr><td>child:00</td></tr></table>' +
+									'</td>' +
+								'</tr>' +
+							'</table>'
+						);
 
 						const table = model.document.getRoot().getNodeByPath( [ 0 ] );
 
@@ -233,22 +235,22 @@ describe( 'table properties', () => {
 						// Previously the test was too loose in that regard.
 						expect( getModelData( editor.model ) ).to.equal(
 							'[<table ' +
-							'borderColor="{"top":"red","bottom":"red","right":"red","left":"red"}" ' +
-							'borderStyle="{"top":"solid","bottom":"solid","right":"solid","left":"solid"}" ' +
-							'borderWidth="{"top":"1px","bottom":"1px","right":"1px","left":"1px"}">' +
-							'<tableRow>' +
-								'<tableCell>' +
-									'<paragraph>' +
-										'parent:00' +
-									'</paragraph>' +
-								'</tableCell>' +
-								'<tableCell>' +
-									'<paragraph>' +
-										'child:00' +
-									'</paragraph>' +
-								'</tableCell>' +
-							'</tableRow>' +
-						'</table>]'
+								'borderColor="{"top":"red","bottom":"red","right":"red","left":"red"}" ' +
+								'borderStyle="{"top":"solid","bottom":"solid","right":"solid","left":"solid"}" ' +
+								'borderWidth="{"top":"1px","bottom":"1px","right":"1px","left":"1px"}">' +
+								'<tableRow>' +
+									'<tableCell>' +
+										'<paragraph>' +
+											'parent:00' +
+										'</paragraph>' +
+									'</tableCell>' +
+									'<tableCell>' +
+										'<paragraph>' +
+											'child:00' +
+										'</paragraph>' +
+									'</tableCell>' +
+								'</tableRow>' +
+							'</table>]'
 						);
 					} );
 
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
```diff
@@ -228,6 +228,38 @@ describe( 'table properties', () => {
...
+					editor.setData( '<table style="border:1px solid red">' +
+						'<tr>' +
+							'<td>parent:00</td>' +
+							'<td>' +
+								'<table style="border:1px solid green"><tr><td>child:00</td></tr></table>' +
+							'</td>' +
+						'</tr>' +
+					'</table>' );
+
+					expect( getModelData( editor.model ) ).to.equal(
```
maxbarnas, line 243: This way of testing stands out here, but because I wanted to focus on the contents, not the style, I went with `getModelData` here.

Reinmar: This is a perfectly valid code comment :) You can add it inline.

packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
```diff
@@ -228,6 +228,38 @@ describe( 'table properties', () => {
 					assertTRBLAttribute( table, 'borderStyle', 'solid' );
 					assertTRBLAttribute( table, 'borderWidth', '1px' );
 				} );
+
+				// https://github.com/ckeditor/ckeditor5/issues/8393.
+				it( 'should not upcast contents of nested table', () => {
```
Reinmar, line 233: This description is too vague. What you're checking here is not whether the nested table is being upcasted. But whether the converter applies exactly to the element on which the border is set.

However, to avoid having to explain this, I'd simply merge this test with the above one. They do the same thing, but check it differently. The former one is just too leaky – it checks something but not enough to catch the issue that we had. So, why not just adding additional check there (the one you added below)?

Also, you should also add a TC for the case from the actual ticket that you're working on. It's a bit different because it features an empty cell.

Finally, perhaps you can think of more cases that should be covered?

Reinmar (changes requested):
I think we can improve tests here.

Reinmar:
As I commented in the code – this is not about "not upcasting contents of an inner table". It's about applying the style only if a table was actually converted.

Reinmar:
Could you check whether this works now?

```
editor.setData(`<figure class="image"><img src="X"><figcaption><table>
<tr>
<td>parent:00</td>
<td>
<table style="border:1px solid green"><tr><td>child:00</td></tr></table>
</td>
</tr>
</table></figcaption></figure>`)
```

It's a different error being thrown and I wonder if this is a different bug or different manifestation of the same thing.

packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
maxbarnas: The other case I can come up with is when we provide the markup with empty nested table, e.g.

```
<table>
<tbody>
<tr>
<td>

<table style="border-bottom: 0 solid #fff;"></table>

</td>
</tr>
</tbody>
</table>
```

This seems like a more general case - an element inside the cell that's not allowed there but has styles. This will trigger the same error as in the issue.

maxbarnas:
> Could you check whether this works now?
> 
> ```
> editor.setData(`<figure class="image"><figcaption><table>
> <tr>
> <td>parent:00</td>
> <td>
> <table style="border:1px solid green"><tr><td>child:00</td></tr></table>
> </td>
> </tr>
> </table></figcaption></figure>`)
> ```
> 
> It's a different error being thrown and I wonder if this is a different bug or different manifestation of the same thing.

Yup, it works now :)

packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
```diff
@@ -227,6 +227,87 @@ describe( 'table properties', () => {
...
+										'child:00' +
+									'</paragraph>' +
+								'</tableCell>' +
+							'</tableRow>' +
+						'</table>]'
+					);
+				} );
+
+				// https://github.com/ckeditor/ckeditor5/issues/8393.
+				it( 'should not throw error when loading nested table with border styles', () => {
```
Reinmar, line 255: It took some effort to come up with these input data. Don't you think that it would be good to use this occassion and also test whether what ends up in the model is correct too?

Reinmar: Also, we prefer not to have one test covering 3 cases, so I'd prefer if you split this into 3 separate scenarios.

maxbarnas: I merged those cases based on the assumption we are just checking the exceptions. If I am going to check the models, I would also go with separating TC into separate tests.

PR #8767, opened 2021-01-07
Formatting lists preceded by tables.

### LINKED ISSUE #8412: Table can overlap with List

1. Insert a table
2. Left align the table
3. Insert a list right after the table (the list will appear on the right of the table)

### ✔️ Expected result

No overlapping between the list and the table.

### ❌ Actual result

There is overlapping. See below:
Screen Shot 2020-11-03 at 1 15 29 PM

## 📃 Other details
Reproducible here: https://ckeditor.com/docs/ckeditor5/latest/features/table.html

---

magda-chrzescian:

Fix (list): Lists should be indented properly if preceded by the block widget. Closes #8412.


oleq (commented):
The solution is very hacky and I'm worried it may backfire in some edge-cases (not sure which ones, though). I mean, the risk factor may outweigh possible benefits.

BTW: I also came across this bug where list bullets disappear when the screen is narrow on Safari/Mac:

Issue #8581, opened 2020-12-03
Add ELSint rule to check inter-packages imports


jodator:
## 📝 Provide a description of the new feature

Follow-up for the #8395. We're going to change how we import things from other packages.

This ESLint rule should check if the imports from other packages are pointing to the package entry file instead of direct import.

In other words, depending on how the #8578 is resolved the imports should point to entry file:

```javascript
import { Plugin } from '@ckeditor/ckeditor5-core/core';
// or (see #8578)
import { Plugin } from 'ckeditor/core';
```

The wrong import would be:

```javascript
import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
```

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

Reinmar:
> Verify that the only thing that a package has in dependencies is ckeditor5 (from the set of pkgs maintained by us). Second rule – only import from the current package or ckeditor5 (also, only applies to the set of our packages). Let's have ESLint rules for that (#8581). Exception: the ckeditor5 package.

Reinmar:
TODO: Update remaining incorrect imports (also in CF and internal).

pomek:
In the `Title` plugin we need to refactor the code a bit:

```diff
-import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
-import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
+import { Plugin } from 'ckeditor5/src/core';
 
-import ViewDocumentFragment from '@ckeditor/ckeditor5-engine/src/view/documentfragment';
-import ViewDowncastWriter from '@ckeditor/ckeditor5-engine/src/view/downcastwriter';
-import first from '@ckeditor/ckeditor5-utils/src/first';
+import { first } from 'ckeditor5/src/utils';
 import {
+	UpcastWriter,
 	needsPlaceholder,
 	showPlaceholder,
 	hidePlaceholder,
 	enablePlaceholder
-} from '@ckeditor/ckeditor5-engine/src/view/placeholder';
+} from 'ckeditor5/src/engine';
```

```diff
@@ -44,7 +42,7 @@ export default class Title extends Plugin {
 	 * @inheritDoc
 	 */
 	static get requires() {
-		return [ Paragraph ];
+		return [ 'Paragraph' ];
 	}
```

```diff
 	/**
@@ -149,10 +147,11 @@ export default class Title extends Plugin {
 		const data = editor.data;
[...]

PR #8783, opened 2021-01-08
DLL - updated imports

pomek:

Internal: Updated import paths. See #8581.

---

### Additional information

After merging the PR, remember about adding the note below to the list of breaking changes here – https://github.com/ckeditor/ckeditor5/pull/8614#issue-535043376.

```
BREAKING CHANGE: The `Title` no longer load `Paragarph`. Make sure to add `Paragraph` to the editor plugins when using `Title` feature.
```


pomek:
This PR requires additional eslint rules:

- https://github.com/ckeditor/eslint-plugin-ckeditor5-rules/pull/2
- https://github.com/ckeditor/eslint-config-ckeditor5/pull/1

In order to test everything, you need to clone additional repositories. Assuming, that your current working directory is set to `ckeditor` workspace:

- `git clone git@github.com:ckeditor/eslint-plugin-ckeditor5-rules.git`
- `cd eslint-plugin-ckeditor5-rules`
- `git checkout i/8581`
- `yarn link` (it should call `yarn install`)
- `cd ../`
- The same for the second repository.
- `git clone git@github.com:ckeditor/eslint-config-ckeditor5.git`
- `cd eslint-config-ckeditor5`
- `git checkout i/8581`
- `yarn link` (it should call `yarn install`)
- `cd ../`
- The main repository:
- `cd ckeditor5`
- `yarn link eslint-plugin-ckeditor5-rules`
- `yarn link eslint-config-ckeditor5`

Then, on the `dll-integration` branch, `yarn run lint` should print some errors. The `i/8581` branch should fix those errors.

packages/ckeditor5-engine/src/index.js
```diff
@@ -16,10 +16,18 @@ export { default as Conversion } from './conversion/conversion';
 
 export { default as HtmlDataProcessor } from './dataprocessor/htmldataprocessor';
 
+export { default as InsertOperation } from './model/operation/insertoperation';
```
Reinmar, line 19: Why do we need to add these?

Reinmar:
What about these?

image

packages/ckeditor5-engine/src/index.js
pomek: CF.

Reinmar:
image

So, my stance on this is that consistency is important. It will be confusing which packages use direct imports and which go via `ckeditor5`. We'd need additional ESLint rules to enforce that packages within the DLL use direct imports.

Therefore, I'd rather state it a bit differently: Every package that we have may only have three types of imports:

* directly from within the same package (local ones)
* from `ckeditor5`
* from 3rd party things

Therefore, it'd be great if you could research how much work would require to enforce this rule. I think it may require adding a few more things to packages' index files but other than that it seems doable.

Reinmar:
Oh, and one more thing – you can also consider my yesterday's finding that plugins that are part of the Essential plugin should be in the DLL.

packages/ckeditor5-engine/src/index.js
Reinmar: If we're adding InsertOperation we should also add all other operations here.

PR #8797, opened 2021-01-11
Fake selection marker for collapsed and non-collapsed selections for link UI should be converted separately

### LINKED ISSUE #8092: Markers in empty elements can cause unexpected errors
The problem occurred for the first time in #7907.

---

We agreed that the bug can be hooked deeper, in the engine. However, the PR fixes the bug and the editor does not crash anymore.

As for now, we're fine with that. But in the future, the bug can occur once again and if we will be fixing that, we should revert changes introduced in the PR.

https://github.com/ckeditor/ckeditor5/blob/eb92cfb7377fa066a4cb08163ade33a73639aab1/packages/ckeditor5-link/src/linkui.js#L664-L687

_Originally posted by @pomek in https://github.com/ckeditor/ckeditor5/pull/8074#issuecomment-693331315_

niegowski:

Other (link): Fake selection marker for collapsed and non-collapsed selections for link UI should be converted separately. Closes #8092.

---

### Additional information

The problem with downcasting `link-ui` marker was that two separate converters were defined for the same marker:

*   `markerToHighlight` - for the non collapsed selection.
*   `markerToElement` - for the collapsed selection (but this converter also works correctly for non collapsed selection - it inserts elements at both ends of the marker range).

### DIFF
--- a/packages/ckeditor5-link/src/linkui.js
+++ b/packages/ckeditor5-link/src/linkui.js
@@ -665,7 +665,7 @@ export default class LinkUI extends Plugin {
 			} else {
 				if ( range.start.isAtEnd ) {
 					const startPosition = range.start.getLastMatchingPosition(
-						( { item } ) => !model.schema.checkAttribute( item, 'linkHref' ),
+						( { item } ) => !item.is( '$text' ) && !item.is( '$textProxy' ) && !model.schema.isLimit( item ),
 						{ boundaries: range }
 					);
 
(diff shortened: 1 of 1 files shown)


scofalik:
I reviewed this solution and I don't really like having two markers for the same thing. This is why I asked @niegowski to check alternative solution, which still has only one marker but the marker range is modified. It is similar to Radek's solution but it should be easier (less lines of code) as I proposed using `position.getLastMatchingPosition()` to move starting position to a proper place.

niegowski:
> I reviewed this solution and I don't really like having two markers for the same thing. This is why I asked @niegowski to check alternative solution, which still has only one marker but the marker range is modified. It is similar to Radek's solution but it should be easier (less lines of code) as I proposed using `position.getLastMatchingPosition()` to move starting position to a proper place.

This solution is much simpler 🙂 I just pushed the fix.

PR #8799, opened 2021-01-11
Added a placeholder for not previewable HTML snippet.

### LINKED ISSUE #8435: Show something when embedding something non-previewable

It currently looks like this:

The problem is not that it's not clear that this is an HTML snippet, but rather – what snippet this is. If you have a couple of them in the content, you may not understand which is which or if they are filled already or not.

Therefore, some way to identify what's inside would be good.

### Option 1. Show non-editable source

We could simply show a disabled textarea with the source of the embed, just like in the case of when previews are completely disabled.

**However, how can we detect that a preview is not available?**

One option is, of course, that when the sanitizer stripped everything, we know that the thing is non-previewable. But what if just one part of the input was removed and it makes the preview non-existent?

The other option is to make this check after rendering the preview. If it is `0px` high, we can tell that the preview didn't render. However, this will be slow and may cause layout trashing so if we go this way we must be smart.

Related ticket (in this option): #8326.

### Option 2. Add a "title" field

The other option would be to add a field for a snippet's title and display it e.g. in the "HTML snippet: \<title>" place.

The title could be stored in `data-title` and we wouldn't have to do anything more – if the preview doesn't render, the user still sees something. We don't have to even detect these situations.

My biggest concern is that the title will leak to data, but I'd be fine with that as it's kind of a "label/caption" like thing.

---

magda-chrzescian:

Fix (html-embed): An information should be displayed if the HTML snippet is not previewable. Closes #8435.


magda-chrzescian:
@oleq Sinc it's totally different solution than the one we discussed, I'd like to wait for some initial review before writing the tests.

magda-chrzescian:
The current state is as follows:  
\- "No preview available" is displayed for all invisible by design elements (scripts, head, comments, elements with `display: none`).  
\- Empty container is displayed if renderable, but invisible element is inside (div with 0x0 size, larger div with no background-color). This one may be a poor UX, but this is a rare case and it seems reasonable that if the user adds content that is invisible, he sees nothing.  
\- Proper information must be added if the snippet is empty,  
\- The content is displayed properly if an element in the HTML snippet is smaller than the placeholder behind it.

packages/ckeditor5-theme-lark/theme/ckeditor5-html-embed/htmlembed.css
```diff
@@ -8,6 +8,8 @@
 	--ck-html-embed-source-height: 10em;
 	--ck-html-embed-unfocused-outline-width: 1px;
 	--ck-html-embed-content-min-height: calc(var(--ck-icon-size) + var(--ck-spacing-standard));
+
+	--ck-html-embed-source-disabled-color: hsl(0deg 0% 45%)
```
magda-chrzescian, line 12: This variable is added to be consistent with not merged yet [https://github.com/ckeditor/ckeditor5/pull/8818/files,](https://github.com/ckeditor/ckeditor5/pull/8818/files,) which also fixed the contrast ratio.

PR #8804, opened 2021-01-11
Make external dependencies for marking UI in the Docs as a part of the build

### LINKED ISSUE #8793: 404 errors caused by tippy and popperjs in nightly docs
Visit: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/mentions.html

```
GET https://ckeditor5.github.io/node_modules/tippy.js/dist/tippy.css net::ERR_ABORTED 404
mentions.html:32 GET https://ckeditor5.github.io/node_modules/tippy.js/themes/light-border.css net::ERR_ABORTED 404
mentions.html:808 GET https://ckeditor5.github.io/node_modules/@popperjs/core/dist/umd/popper.min.js net::ERR_ABORTED 404
mentions.html:809 GET https://ckeditor5.github.io/node_modules/tippy.js/dist/tippy-bundle.umd.min.js net::ERR_ABORTED 404
mentions.html:808 GET https://ckeditor5.github.io/node_modules/@popperjs/core/dist/umd/popper.min.js net::ERR_ABORTED 404
mentions.html:809 GET https://ckeditor5.github.io/node_modules/tippy.js/dist/tippy-bundle.umd.min.js net::ERR_ABORTED 404
```

image

I don't like two things here:

* That we read from node_modules directly. As you can see – it's not published on nightly docs.
* That we need to load 4 more files just for the toolbar hints.

We should try fixing at least the first issue, but if we can also reduce the second issue (e.g. by somehow merging these files into existing assests), then it'd be great. But the latter seems like a bigger deal, so if we won't see the light in the tunnel quickly, it may not be worth digging deeper.

maxbarnas:

Fix (docs): Make external dependencies for marking UI in the Docs as a part of the build. Closes #8793.

---

### Additional information


scripts/docs/snippetadapter.js
```diff
@@ -549,6 +556,86 @@ function countUniqueSnippets( snippets ) {
...
+		resolveLoader: {
+			modules: getModuleResolvePaths()
+		},
+
+		module: {
+			rules: [
+				{
+					test: /\.svg$/,
+					use: [ 'raw-loader' ]
+				},
```
oleq, line 626: Are we going to use this loader? It doesn't seem necessary so far (or am I missing something?).

oleq (changes requested):
Something isn't right.

Master:
image
PR:
image

oleq:
Looks like we forgot to load the [theme](https://github.com/ckeditor/ckeditor5/pull/8804/files#diff-64df00835cabfa8b0cb333f7da1cda6a0abfe60968806d6c6f51c6722db102d9L201).

scripts/docs/snippetadapter.js
maxbarnas: @pomek can you provide your expertise? :)

maxbarnas:
> Looks like we forgot to load the [theme](https://github.com/ckeditor/ckeditor5/pull/8804/files#diff-64df00835cabfa8b0cb333f7da1cda6a0abfe60968806d6c6f51c6722db102d9L201).

Please, check again.

scripts/docs/snippetadapter.js
pomek: ATM it loader can be deleted.

maxbarnas: Fixing it now, thanks @pomek!

Reinmar:
I get this warning: 

```
DevTools failed to load SourceMap: Could not load content for webpack:///node_modules/tippy.js/dist/tippy.esm.js.map: HTTP error: status code 404, net::ERR_UNKNOWN_URL_SCHEME
```

image

It's just a warning, but it's great if the console is clean (you just don't start to worry about things). There's something in this webpack configuration for assets that is incorrect. You can try to google that or compare the config for assets with the other config that we have there (for snippets code) that doesn't cause such issues.

Reinmar (changes requested):
As commented.

maxbarnas:
> I get this warning:
> 
> ```
> DevTools failed to load SourceMap: Could not load content for webpack:///node_modules/tippy.js/dist/tippy.esm.js.map: HTTP error: status code 404, net::ERR_UNKNOWN_URL_SCHEME
> ```
> 
> image
> 
> It's just a warning, but it's great if the console is clean (you just don't start to worry about things). There's something in this webpack configuration for assets that is incorrect. You can try to google that or compare the config for assets with the other config that we have there (for snippets code) that doesn't cause such issues.

I tried to get rid of the source map request, but no amount of Webpack config shenanigans made it disappear. The source map link is hardcoded in the `/node_modules/tippy.js/dist/tippy.esm.js` file. @pomek suggested we either need to strip it using some loader/plugin or build that package ourselves.

maxbarnas:
I just checked, and building the docs with a `--production` flag seems to do the job in removing unnecessary source map links. Can someone confirm on their side?

oleq:
> I just checked, and building the docs with a `--production` flag seems to do the job in removing unnecessary source map links. Can someone confirm on their side?

Confirmed.

maxbarnas:
> > I just checked, and building the docs with a `--production` flag seems to do the job in removing unnecessary source map links. Can someone confirm on their side?
> 
> Confirmed.

Thanks! In that case I would leave it as is. ROI of solving it just for development env seems very low.

PR #8807, opened 2021-01-11
Supporting batchType option in the setData method.

### LINKED ISSUE #7947: setData in ckeditor5-engine dev-utils/model ignores batchType option
It seems that an old PR (https://github.com/ckeditor/ckeditor5-engine/pull/1431) has removed the ability to use the `batchType` option in `setData` in https://github.com/ckeditor/ckeditor5/blob/v21.0.0/packages/ckeditor5-engine/src/dev-utils/model.js#L108. It seems that `setData` used to use `model.enqueueChange` which supports specifying the batch, but this was changed in the aforementioned PR.

It would be very useful for writing tests if we could use this option. But if this change was intentional, please update the documentation (https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_dev-utils_model.html#function-setData) which still lists `batchType` as an option.

1. 
```
// in a plugin test
import { expect } from "chai";
import { setData as setModelData } from "@ckeditor/ckeditor5-engine/src/dev-utils/model";

let editor;
...

editor.model.document.registerPostFixer(writer => {
  expect(writer.batch.type).to.equal("transparent");
});
setModelData(editor.model, "<paragraph>Some data</paragraph>", { batchType: "transparent" });
```

### ✔️ Expected result

The above assertion is true

### ❌ Actual result

The above assertion is false

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

---

magda-chrzescian:

Fix (engine): The setData method in dev-utils/model should support the batchType option. Closes #7947.

### DIFF
--- a/packages/ckeditor5-engine/src/dev-utils/model.js
+++ b/packages/ckeditor5-engine/src/dev-utils/model.js
@@ -128,7 +128,13 @@ export function setData( model, data, options = {} ) {
 		modelDocumentFragment = parsedResult;
 	}
 
-	model.change( writer => {
+	if ( typeof options.batchType === 'string' ) {
+		model.enqueueChange( options.batchType, writeToModel );
+	} else {
+		model.change( writeToModel );
+	}
+
+	function writeToModel( writer ) {
 		// Replace existing model in document by new one.
 		writer.remove( writer.createRangeIn( modelRoot ) );
 		writer.insert( modelDocumentFragment, modelRoot );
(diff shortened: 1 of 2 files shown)


oleq (changes requested):
Looks like the documentation of the method is incorrect. I think `@param {String} [options.batchType='transparent']` should be `@param {String} [options.batchType='default']` because when `model#change()` is executed, [it creates a `new Batch()` with default arguments ](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/model/model.js#L173) and [the default batch type is `'default'`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/model/batch.js#L30).

PR #8809, opened 2021-01-12
Fixed image resizer for images with links

### LINKED ISSUE #8749: Image resizer doesn't work with LinkImage

1. Insert image
2. Add link to image using LinkImage plugin
3. Close windows with editor and open again

### ✔️ Expected result

Resize buttons appeared on the image

### ❌ Actual result

Resize buttons did not appear on the image

This is due to a condition in imageresizehandler.js:

`if ( !domEvent.target.matches( 'figure.image.ck-widget > img' ) ) {
				return;
			}`
But with link on image the selector should be as follows: figure.image.ck-widget > a > img.

Tell me how can I fix this on my own? If I change imageresizehandler.js, then how can I load it correctly into my custom build

psmyrek:

Fix (image): Fixed image resizer for images with links. Closes #8749.

### DIFF
--- a/packages/ckeditor5-image/src/imageresize/imageresizehandles.js
+++ b/packages/ckeditor5-image/src/imageresize/imageresizehandles.js
@@ -56,8 +56,8 @@ export default class ImageResizeHandles extends Plugin {
 		editingView.addObserver( ImageLoadObserver );
 
 		this.listenTo( editingView.document, 'imageLoaded', ( evt, domEvent ) => {
-			// The resizer must be attached only to images loaded by the `ImageInsert` or `ImageUpload` plugins.
-			if ( !domEvent.target.matches( 'figure.image.ck-widget > img' ) ) {
+			// The resizer must be attached only to images loaded by the `ImageInsert`, `ImageUpload` or `LinkImage` plugins.
+			if ( !domEvent.target.matches( 'figure.image.ck-widget > img, figure.image.ck-widget > a > img' ) ) {
 				return;
 			}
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-image/src/imageresize/imageresizehandles.js
```diff
@@ -56,8 +56,8 @@ export default class ImageResizeHandles extends Plugin {
 		editingView.addObserver( ImageLoadObserver );
 
 		this.listenTo( editingView.document, 'imageLoaded', ( evt, domEvent ) => {
-			// The resizer must be attached only to images loaded by the `ImageInsert` or `ImageUpload` plugins.
-			if ( !domEvent.target.matches( 'figure.image.ck-widget > img' ) ) {
+			// The resizer must be attached only to images loaded by the `ImageInsert`, `ImageUpload` or `LinkImage` plugins.
+			if ( !domEvent.target.matches( 'figure.image.ck-widget > img, figure.image.ck-widget > a > img' ) ) {
```
niegowski, line 60: I know that I suggested this solution, but I'm wondering now if we could check if this `img` DOM node has a mapping to view and model elements. WDYT?

psmyrek: Hmm, no idea what kind of consequences this change will have and whether it covers all cases...

niegowski: Hmm seems that `img` element itself has no mapping to model element. Ok let's leave it this way for now.

PR #8812, opened 2021-01-12
The raw content matchers can be updated through `DataProcessor`

### LINKED ISSUE #8789: Pasting HTML embed widget does not paste it's content

1.  https://ckeditor.com/docs/ckeditor5/latest/features/html-embed.html
2.  Copy HTML widget
3.  Paste HTML widget

### ✔️ Expected result

Widget is pasted with its content.

### ❌ Actual result

Content of HTML widget is empty.

## 📃 Other details

*   Browser: …
*   OS: …

---

maxbarnas:

Fix (html-embed): Pasting HTML embed widget from the clipboard will not clear its content anymore. Closes #8789.

Feature (engine): New property `DataController#htmlProcessor` is initialized with `HtmlDataProcessor` and assigned to `DataController#processor` by default.

Internal (clipboard): Plugin now uses `HtmlDataProcessor` instance from the `DataController`.

Internal (editor-balloon, editor-classic, editor-inline, editor-decoupled): The `editor.processor` property does is automatically initialized with `HtmlDataProcessor` now.

---

### Additional information

The `DataController` will always keep `HtmlDataProcessor` instance under `editor.data.htmlProcessor` for reuse. The editor can be processed in many ways, but the clipboard plugin always uses `HtmlDataProcessor`. 

The method `DataController#registerRawContentMatcher` will keep both `data.processor` and `data.htmlProcessor` instances in sync when it comes to matchers.

### DIFF
--- a/packages/ckeditor5-clipboard/tests/clipboard.js
+++ b/packages/ckeditor5-clipboard/tests/clipboard.js
@@ -584,6 +584,14 @@ describe( 'Clipboard feature', () => {
 				);
 			} );
 		} );
+
+		function createDataTransfer( data ) {
+			return {
+				getData( type ) {
+					return data[ type ];
+				}
+			};
+		}
 	} );
 
 	describe( 'clipboard copy/cut pipeline', () => {
@@ -825,11 +833,3 @@ describe( 'Clipboard feature', () => {
 		}
 	} );
 } );
-
-export function createDataTransfer( data ) {
-	return {
-		getData( type ) {
-			return data[ type ];
-		}
-	};
-}
--- a/packages/ckeditor5-clipboard/tests/clipboard.js
+++ b/packages/ckeditor5-clipboard/tests/clipboard.js
@@ -584,14 +584,6 @@ describe( 'Clipboard feature', () => {
 				);
 			} );
 		} );
-
-		function createDataTransfer( data ) {
-			return {
-				getData( type ) {
-					return data[ type ];
-				}
-			};
-		}
 	} );
 
 	describe( 'clipboard copy/cut pipeline', () => {
@@ -833,3 +825,11 @@ describe( 'Clipboard feature', () => {
 		}
 	} );
 } );
+
+export function createDataTransfer( data ) {
+	return {
+		getData( type ) {
+			return data[ type ];
+		}
+	};
+}
--- a/packages/ckeditor5-engine/src/controller/datacontroller.js
+++ b/packages/ckeditor5-engine/src/controller/datacontroller.js
@@ -111,6 +111,7 @@ export default class DataController {
 		/**
 		 * Data processor used specifically for HTML conversion.
 		 *
+		 * @readonly
 		 * @member {module:engine/dataprocessor/htmldataprocessor~HtmlDataProcessor} #htmlProcessor
 		 */
 		this.htmlProcessor = new HtmlDataProcessor( this.viewDocument );
--- a/packages/ckeditor5-engine/src/controller/datacontroller.js
+++ b/packages/ckeditor5-engine/src/controller/datacontroller.js
@@ -452,8 +452,7 @@ export default class DataController {
 	 * be treated as a raw data.
 	 */
 	registerRawContentMatcher( pattern ) {
-		// The `this.processor` can be the same instance as the `this.htmlProcessor`.
-		// Do not register same pattern twice then.
+		// No need to register the pattern if both `htmlProcessor` and `processor` are the same instances.
 		if ( this.processor && this.processor !== this.htmlProcessor ) {
 			this.processor.registerRawContentMatcher( pattern );
 		}
--- a/packages/ckeditor5-engine/src/controller/datacontroller.js
+++ b/packages/ckeditor5-engine/src/controller/datacontroller.js
@@ -109,13 +109,15 @@ export default class DataController {
 		this.stylesProcessor = stylesProcessor;
 
 		/**
-		 * TODO
+		 * Data processor used specifically for HTML conversion.
+		 *
+		 * @member {module:engine/dataprocessor/htmldataprocessor~HtmlDataProcessor} #htmlProcessor
 		 */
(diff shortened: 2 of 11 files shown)


packages/ckeditor5-html-embed/src/htmlembedediting.js
```diff
@@ -95,7 +95,7 @@ export default class HtmlEmbedEditing extends Plugin {
 
 		// Register div.raw-html-embed as a raw content element so all of it's content will be provided
 		// as a view element's custom property while data upcasting.
-		editor.data.processor.registerRawContentMatcher( {
+		editor.data.registerRawContentMatcher( {
```
maxbarnas, line 98: Thinking about changing its name to something that shows it's a special method.

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -431,6 +437,21 @@ export default class DataController {
 		callback( this.stylesProcessor );
 	}
 
+	/**
+	 * TODO
+	 * @param {module:engine/view/matcher~MatcherPattern} pattern Pattern matching all view elements whose content should
+	 * be treated as a raw data.
+	 */
+	registerRawContentMatcher( pattern ) {
```
maxbarnas, line 445: Thinking about changing its name to something that shows it's a special method.

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -114,6 +100,29 @@ export default class DataController {
...
+		 */
+		this.htmlProcessor = new HtmlDataProcessor( this.viewDocument );
+
+		/**
+		 * Data processor used during the conversion.
+		 * Same instance as {@link #htmlProcessor} by default. Can be replaced at run time to handle different format, e.g. XML or Markdown.
+		 *
+		 * @member {module:engine/dataprocessor/dataprocessor~DataProcessor} #processor
+		 */
+		this.processor = this.htmlProcessor;
```
maxbarnas, line 124: [**@Reinmar**](https://github.com/Reinmar), I went with initializing the `processor` by default with `htmlProcessor` and removing initialization from the builds. Is there any reason `processor` was not created this way previously?

Reinmar: The reason was that the DC did not create any DP so it made no sense to create here once if it was to be overridden by e.g. Markdown's DP. 

Now, if we have `this.htmlProcessor` anyway defined above, it's ok to have `this.processor` set to the same instance.

packages/ckeditor5-html-embed/package.json
```diff
@@ -20,6 +20,7 @@
     "@ckeditor/ckeditor5-basic-styles": "^24.0.0",
     "@ckeditor/ckeditor5-editor-classic": "^24.0.0",
     "@ckeditor/ckeditor5-paragraph": "^24.0.0",
+    "@ckeditor/ckeditor5-clipboard": "^24.0.0",
```
Reinmar, line 23: Revert needed.

Reinmar: Sorry, it's dev deps. Nvmnd.

packages/ckeditor5-clipboard/tests/clipboard.js
```diff
@@ -833,3 +825,11 @@ describe( 'Clipboard feature', () => {
...
 	} );
 } );
+
+export function createDataTransfer( data ) {
+	return {
+		getData( type ) {
+			return data[ type ];
+		}
+	};
+}
```
niegowski, line 835: We don't export utils from tests. This is so small code that it could be copied to other tests (and AFAIR we have multiple copies of similar code). If we have some bigger utils that we use in many tests then we store them in the `tests/_utils` directory.

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -114,6 +100,29 @@ export default class DataController {
...
+		 * @member {module:engine/view/stylesmap~StylesProcessor}
+		 */
+		this.stylesProcessor = stylesProcessor;
+
+		/**
+		 * Data processor used specifically for HTML conversion.
+		 *
+		 * @member {module:engine/dataprocessor/htmldataprocessor~HtmlDataProcessor} #htmlProcessor
+		 */
+		this.htmlProcessor = new HtmlDataProcessor( this.viewDocument );
```
niegowski, line 116: This should be marked as read-only.

packages/ckeditor5-engine/tests/controller/datacontroller.js
```diff
@@ -63,6 +59,20 @@ describe( 'DataController', () => {
...
+			const data = new DataController( model, stylesProcessor );
+
+			expect( data.htmlProcessor ).to.be.instanceOf( HtmlDataProcessor );
+		} );
+
+		it( 'should assign #htmlProcessor property to the #processor property', () => {
+			const stylesProcessor = new StylesProcessor();
+			const data = new DataController( model, stylesProcessor );
+
+			expect( data.htmlProcessor ).to.be.equal( data.processor );
```
niegowski, line 74:
```suggestion
			expect( data.htmlProcessor ).to.equal( data.processor );
```

Just for consistency.

niegowski:
I really like this solution 🙂

niegowski:
Also, the merge commit message should include a user-related issue fix description (so in this case `Fix (html-embed):` )

packages/ckeditor5-clipboard/tests/clipboard.js
maxbarnas: Yup, I was wondering whether to copy it or export. Thanks!

maxbarnas:
> Also, the merge commit message should include a user-related issue fix description (so in this case `Fix (html-embed):` )

Thanks for pointing that out. For a good moment I got lost in how to phrase the message. That clears it once and for all.

niegowski (approved):

PR #8891, opened 2021-01-23
Fix 'Cannot read property 'getAttributeKeys' of null'

avalko:
Without this fix we have error:
exception

### DIFF
--- a/packages/ckeditor5-table/src/converters/downcast.js
+++ b/packages/ckeditor5-table/src/converters/downcast.js
@@ -378,6 +378,11 @@ function createViewTableCellElement( tableSlot, tableAttributes, insertPosition,
 
 	conversionApi.mapper.bindElements( tableCell, cellElement );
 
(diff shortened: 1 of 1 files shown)


Reinmar:
Hi! Thanks for the PR. In order to merge it we need test coverage. This means not only covering all lines but also showing (by a test) that this change is needed.

You can read about CKEditor 5's dev env and test env here:

* https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/development-environment.html
* https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/testing-environment.html

Reinmar:
Actually, the first very first thing that we'd need is a scenario in which this bug can be triggered.

mlewand:
Hi, It's been a while since we last heard from you. We're closing this issue for now. Still, feel free provide us requested feedback, so that we can reopen this PR.

PR #8929, opened 2021-01-26
Hide a toolbar tooltip when focusing the editable.

### LINKED ISSUE #8883: Hide a toolbar tooltip when focusing the editable

Recently we introduced a tooltip that shows a feature button in the editor's toolbar.

You can close the tooltip by clicking `x`.

Would be good to hide the tooltip also when the editor is focused (after clicking the editable).

image

pkwasnik:

Fix (docs): Hide a toolbar tooltip when focusing the editable. Closes #8883.

---

### Additional information

### DIFF
--- a/docs/_snippets/assets.js
+++ b/docs/_snippets/assets.js
@@ -85,7 +85,7 @@ window.attachTourBalloon = function( { target, text, tippyOptions } ) {
 		tooltip.hide();
 	} );
 
-	target.addEventListener( 'click', () => {
+	target.offsetParent.addEventListener( 'click', () => {
 		tooltip.hide();
 	} );
 };
--- a/docs/_snippets/assets.js
+++ b/docs/_snippets/assets.js
@@ -21,13 +21,15 @@ import './tour-balloon.css';
  *		// Using a comparison callback to search for an item.
  *		window.attachTourBalloon( {
  *			target: window.findToolbarItem( editor.ui.view.toolbar, item => item.label && item.label === 'Insert HTML' ),
- *			text: 'Tour text to help users discover the feature.'
+ *			text: 'Tour text to help users discover the feature.',
+ *			editor
  *		} );
  *
  *		// Using a toolbar item index.
  *		window.attachTourBalloon( {
  *			target: window.findToolbarItem( editor.ui.view.toolbar, 5 ),
- *			text: 'Tour text to help users discover the feature.'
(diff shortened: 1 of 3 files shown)


docs/_snippets/assets.js
```diff
@@ -85,7 +85,7 @@ window.attachTourBalloon = function( { target, text, tippyOptions } ) {
 		tooltip.hide();
 	} );
 
-	target.addEventListener( 'click', () => {
+	target.offsetParent.addEventListener( 'click', () => {
```
oleq, line 88: This approach is incorrect. 

An offset parent is ([via MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent)):

> a reference to the element which is the closest (nearest in the containment hierarchy) positioned ancestor element. If there is no positioned ancestor element, the nearest ancestor `td`, `th`, `table` will be returned, or the `body` if there are no ancestor table elements either.

so for a toolbar button in:

*   a `ClassicEditor`: this will be the outermost box of the editor. This is why it **looks** like it works because this box also contains the editing root but the tooltip will also disappear when you click any other button of the toolbar or an empty space in the toolbar.
*   an `InlineEditor` or a `BalloonEditor`: this will be the `.ck-balloon-panel` that hosts the toolbar. Focusing the editing root will not fire `click` on this element.
*   a `DecoupledDocumentEditor`: this will be the `div.main__content-inner` of the documentation page, which means clicking basically anywhere in the guide **will hide all tour balloons**.

There's no way to find out which editing root is related to the `target` so you need to pass the editor instance reference to this helper and create `once()` listener such as this to handle this properly

```javascript
editor.ui.view.editable.once( 'change:isFocused', ( evt, name, isFocused ) => {
    if ( isFocused ) {
        tooltip.hide();
    }
} );
```

oleq (changes requested):

docs/_snippets/assets.js
oleq: Edit: Iterating over `editor.editing.view.document.roots` and attaching a listener for each one of them will be safer since some editors such as multi-root may not have `editor.ui.view.editable`.

pkwasnik: I wanted to avoid passing `editor` and "tooltip will also disappear when you click any other button of the toolbar" is a feature IMO. But other arguments are obviously right, I will correct this now.

PR #8954, opened 2021-01-29
Improvements and simplifications in the `PluginCollection`

### LINKED ISSUE #8634: PluginCollection logic might be improved

The plugin collection does multiple checks for required plugins but its logic is mangled.

In #8632 I've pushed some improvements in "soft" requirements checking but this whole logic can be changed to two-way pass:

1.  List all plugins to load by inspecting list of `plugins` and plugins to load from `requires` property.
2.  Check if all plugins are available (the same as we do it now).
3.  Throw an error if plugins are not found.

This is inspired by the idea of having all the changes in the downcast dispatcher (differ) before running conversion (cc @Reinmar ).

Right now, the logic for which plugin is loaded and which are available is a bit mangled. This will allow checking all dependencies (hard or soft), event if deeply nested (like: `ArticlePluginSet` -> `Essentials` -> `Enter`, etc) before trying to load the editor.

This would also allow to properly order plugins (ie like described in #2907.

Additionally, the errors could describe which plugins are missing and which plugins were requiring it. ATM the errors either say which plugins are missing or which soft requirement was not met (on the first error only).

The thrown errors might need to be updated as well: #8680.

psmyrek:

Other (core): Improvements and simplifications in the `PluginCollection` logic. Closes #8634.

Fix (core): Added support in the `PluginCollection` to load dependency plugins using soft requirement, if they are not built-in, but they are available further as dependencies of other plugins. See #8634.

---

### Additional information

Few notes:
* I tried to keep the logic unchanged, that was in the `PluginCollection` so far.
* The `PluginCollection` API has not been changed.
* I added missing plugin names for some plugins, mostly for the `*UI` ones.
* I fixed one minor issue in the docs regarding the `removePlugins` configuration option, which supports strings and functions.
* The `PluginCollection#init()` method will no longer log errors in the console, because they would be doubled (the error is thrown anyway).
* In case of a missing plugin, only one plugin is reported (the first one causing the error) and not all plugins that could be found. It may happen that after adding this one plugin also other missing plugins will be fixed automatically as they are collected globally before starting the plugin instantiation procedure.
* I added a test that does not work in the previous version, but works now, which proves that the previous version did not support all possible cases and the success of the editor initialization depended on the order in which the plugins were defined. The order of plugin dependencies ("normal" and "soft" ones) doesn't matter anymore.


packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -163,35 +162,157 @@ export default class PluginCollection {
...
 	 * and available in the collection.
 	 */
-	init( plugins, removePlugins = [] ) {
+	init( plugins, removedPlugins = [] ) {
+		// Plugin initialization procedure consists of 2 main steps:
+		// 1) collecting all available plugin constructors,
+		// 2) verification whether all required plugins can be instantiated.
+
+		// In the first step, all plugin constructors, available in the provided `plugins` array and inside
+		// plugin's dependencies (from the `requires` array), are recursively collected and added to the existing
```
pomek, line 177:
```suggestion
		// plugin's dependencies (from the `Plugin.requires` array), are recursively collected and added to the existing
```

packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -163,35 +162,157 @@ export default class PluginCollection {
...
-		const pluginConstructors = mapToAvailableConstructors( plugins );
-		const removePluginConstructors = mapToAvailableConstructors( removePlugins );
-		const missingPlugins = getMissingPluginNames( plugins );
+		function isPluginRemoved( plugin, removedPlugins ) {
+			return removedPlugins.some( removedPlugin => {
+				return removedPlugin === plugin ||
+					removedPlugin === getPluginName( plugin ) ||
+					getPluginName( removedPlugin ) === plugin;
+			} );
+		}
```
pomek, line 217: I'd refactor this function a bit:

```suggestion
		function isPluginRemoved( plugin, removedPlugins ) {
			return removedPlugins.some( removedPlugin => {
				if ( removedPlugin === plugin ) {
					return true;
				}
				
				if ( getPluginName( plugin ) === removedPlugin ) {
					return true;
				}

				if ( getPluginName( removedPlugin ) === plugin ) {
					return true;
				}
				
				return false;
			} );
		}
```

packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -163,35 +162,157 @@ export default class PluginCollection {
...
+
+		function getPluginName( plugin ) {
+			return isPluginConstructor( plugin ) ?
+				plugin.pluginName || plugin.name :
```
pomek, line 221: I guess, the `plugin.pluginName` check is enough here. After modification of the code, `plugin.name` could return a string instead of a full name. However, I'm unsure about it.

pomek (approved):
As mentioned but generally, it looks good.

packages/ckeditor5-core/src/plugincollection.js
psmyrek: Done.

packages/ckeditor5-core/src/plugincollection.js
psmyrek: Done.

packages/ckeditor5-core/src/plugincollection.js
psmyrek: The `PluginCollection#get()` method uses both `pluginName` and `name` properties, so I would keep it as it is now to be consistent.

psmyrek:
Hi @Reinmar, would you also like to review this PR?

scofalik (approved):

scofalik:
LGTM. @pomek will you merge it?

pomek:
@scofalik, done.

PR #8955, opened 2021-01-29
Docs: the restoration of the waffle.

godai78:
A pretty image of a delicious waffle has been restored in the documentation after it got accidentally unlinked during some unrelated operations.


AnnaTomanek (approved):
Thanks! TIP: Do not put unrelated changes in a PR as this has a high chance of slipping everyone's attention. Like this one :)

PR #8971, opened 2021-02-02
Changed the way how the DLL modules are added to `window.CKEditor5`

### LINKED ISSUE #8803: The way how DLL pkgs are added to the CKEditor5 global does not work in some cases
image

SelectAll is not required by any other package, so the module itself is not executed until you call `dll()`. Hence, due to how `dll-loader.js` works right now, it's not being registerd in the global object.

pomek:

Internal: Changed the way how the DLL modules are added to the `window.CKEditor5` object. Closes #8803.

---

### Additional information

_Merging as `Internal` because the feature was introduced in the current iteration._


pomek:
The following codes are added at the end of the dll bundle:

Production build:

```js
function(e){const t=()=>{e.CKEditor5=e.CKEditor5||{},e.removeEventListener("DOMContentLoaded",t);const o=["utils","core","engine","ui","cloud-services-core","clipboard","enter","paragraph","select-all","typing","undo","upload","widget"];for(const t of o){const o=t.replace(/-([a-z])/g,(e,t)=>t.toUpperCase());e.CKEditor5[o]=e.CKEditor5.dll(`./src/${t}.js`)}};e.addEventListener("DOMContentLoaded",t)}(window);
```

Development build:

```js
( ( fn, root ) => fn( root ) )( function loadCKEditor5modules( global ) {
	const eventHandler = () => {
		global.CKEditor5 = global.CKEditor5 || {};
		global.removeEventListener( 'DOMContentLoaded', eventHandler );

		const dllPackages = [
			'utils',
			'core',
			'engine',
			'ui',
			'cloud-services-core',
			'clipboard',
			'enter',
			'paragraph',
			'select-all',
			'typing',
			'undo',
			'upload',
			'widget'
		];

		for ( const item of dllPackages ) {
			const windowScope = item.replace( /-([a-z])/g, ( match, p1 ) => p1.toUpperCase() );
			global.CKEditor5[ windowScope ] = global.CKEditor5.dll( `./src/${ item }.js` );
		}
	};

	global.addEventListener( 'DOMContentLoaded', eventHandler );
}, window );
```

psmyrek (changes requested):
The DLL sample test does not work now (`Uncaught TypeError: CKEditor5.core is undefined`).

As I understood your changes correctly, the `window.CKEditor5.<dll-pkg>` can be used after `DOMContentLoaded` has been emited. Due to that fact, all code that relies on DLL build has to be also wrapped by the `DOMContentLoaded` event handler.

My suggestions are:
1. Do not wrap the `loadCKEditor5modules()` body inside the `DOMContentLoaded` event handler, but run it immediately after it has been called.
2. Change the `libraryTarget` from `var` to `window` (to make the suggestion from point 1) work).

pomek:
Thanks, @psmyrek. I pushed a commit with a fix. Could you check it once again?

psmyrek:
The DLL sample test works correctly now, but ES6 module imports are broken, because after importing DLL build (like `import dll from 'ckeditor5/build/ckeditor5-dll.js';`), the `window.CKEditor5.dll()` function is not defined inside `loadCKEditor5modules()` function.

Since we explicitly rely on the fact that the whole DLL core packages are stored in the `window.CKEditor5` object, I think we can also explicitly set the `libraryTarget: 'window'`, rather than `'umd'`. When using `libraryTarget: 'window'` DLLs are working perfectly fine in `<script>` includes and in ES6 module imports.

pomek:
Ok, so let's go with the `window` target.

psmyrek (approved):
It works great now 🎉

PR #8992, opened 2021-02-04
Update angular.md

JamesonNetworks:
Based on comments in https://github.com/ckeditor/ckeditor5-angular/issues/211#issuecomment-644862143 it sounds like the watchdog is already added to the Angular component. It's possible this entire section needs to be removed, however, I wanted to add a note that would prevent future users from trying to add this ContextWatchdog to their own components. FWIW, I also receive an error stating the ContextWatchdog method is not a constructor, which led me to discovering the comment. 

Thanks in advance for you time and everything you all do with CKEditor! Love the editor.

Type (ckeditor/ckeditor5): Adds an info box to ward users away from the watchdog input property.

---

### Additional information

When I tried to implement the watchdog for multiple editors on a single component, I found comments that indicated this is handled for me with the Angular 5+ CKEditor component.

### DIFF
--- a/docs/builds/guides/frameworks/angular.md
+++ b/docs/builds/guides/frameworks/angular.md
@@ -273,6 +273,10 @@ export class MyComponent {
 
 ### `watchdog`
 
+<info-box>
+	The watchdog is built into the Angular Component, so importing the ContextWatchdog is not needed
+</info-box>
+
 An instance of the {@link module:watchdog/contextwatchdog~ContextWatchdog `ContextWatchdog`} class that is responsible for providing the same context to multiple editor instances and restarting the whole structure in case of crashes.
 
 ```ts


JamesonNetworks:
To whoever sees this: Let me know if this is the correct change. I'll sign the CLA agreement and all that assuming this is the way we want to handle it. Otherwise, I'm not planning to jump through the hoops.

ma2ciek:
Hi @JamesonNetworks,

I'm sorry for the late response. 

Unfortunately, your changes are invalid. 

As you can see in [https://github.com/ckeditor/ckeditor5-angular/blob/master/src/ckeditor/ckeditor.component.ts,](https://github.com/ckeditor/ckeditor5-angular/blob/master/src/ckeditor/ckeditor.component.ts,) the component only imports the `EditorWatchdog` so when the component is used only with the editor there's no need to import the watchdog from the app's side.

Things are different when you're using the [https://ckeditor.com/docs/ckeditor5/latest/features/collaboration/context-and-collaboration-features.html.](https://ckeditor.com/docs/ckeditor5/latest/features/collaboration/context-and-collaboration-features.html.) Then you should import the `ContextWatchdog` and pass it to the component as this watchdog will be able to instantiate multiple editor instances connected to the `Context` feature and the `Context` itself.

PR #8998, opened 2021-02-06
#8959 DowncastWriter should handle UIElements consistently while wrapping and inserting

### LINKED ISSUE #1633: Inline widget styling

🐞 Bug report or documentation improvement

## 💻 Version of CKEditor

12.0.0

## 📋 Steps to reproduce

1. Follow the steps from the Inline Widget docs (https://ckeditor.com/docs/ckeditor5/latest/framework/guides/tutorials/implementing-an-inline-widget.html) or (https://ckeditor5.github.io/misc/placeholder-demo/index.html)
2. Try to style all the text

## ✅ Expected result
From my perspective, it would be desired of the inline widget to be treated as part of the text when it comes to styling(bold, italic, font...).

## ❎ Actual result
The text around the inline widget get styled but the widget itself does not.

I was able to get around this issue with changing the createContainerElement with createAttributeElement, adding a random id and priority 20.
And then adding the text using the private _insertChild converting elementToElement. But this does not seam right.

---

### LINKED ISSUE #8959: Inconsistency in down-casting the UIElements

I was checking why putting an `UIElement` inside some text formatted as a code is moving following part of that text and I realized that it's because we split all `AttributeElement`s while inserting any element into the view (this is causing the content movement because a `<code>` element has the padding so it has 2 additional paddings inside after split):

https://github.com/ckeditor/ckeditor5/blob/f9ff247c0d2862da4972b6b451d20707d175be4e/packages/ckeditor5-engine/src/view/downcastwriter.js#L691-L692

On the other hand, if we already have an `UIElement` and we wrap some content with an `AttributeElement` then we allow a `UIElement` to be wrapped with it (also `EmptyElement` and `RawElement` are allowed):

https://github.com/ckeditor/ckeditor5/blob/f9ff247c0d2862da4972b6b451d20707d175be4e/packages/ckeditor5-engine/src/view/downcastwriter.js#L1171-L1176

---

niegowski:

Fix (engine): `DowncastWriter` should handle `UIElement`s consistently while wrapping and inserting. Closes #8959.

Feature (engine): `ContainerElement` could be marked as `isAllowedInsideAttributeElement` and get wrapped with an `AttributeElement`. Closes #1633.

---

### Additional information

This change introduces a view `Element` property `isAllowedInsideAttributeElement` (in the same fashion as `priority` on `AttributeElement`) to specify the element behavior while being wrapped with an `AttributeElement`. Also, it does not break attributes while inserting elements that are not present in the model (`UIElement`, probably should handle `RawElement` in the same way).


packages/ckeditor5-engine/src/view/downcastwriter.js
```diff
@@ -203,10 +203,19 @@ export default class DowncastWriter {
...
+	 * @param {Boolean} [options.isAllowedInAttribute] Whether an element is
+	 * {@link module:engine/view/element~Element#isAllowedInAttribute allowed inside attribute elements}.
 	 * @returns {module:engine/view/containerelement~ContainerElement} Created element.
 	 */
-	createContainerElement( name, attributes ) {
-		return new ContainerElement( this.document, name, attributes );
+	createContainerElement( name, attributes, options = {} ) {
+		const containerElement = new ContainerElement( this.document, name, attributes );
+
+		if ( options.isAllowedInAttribute ) {
```
Reinmar, line 214: isInline, isAllowedInline?

Reinmar:
Out of curiosity: can we make the bold feature work in this case https://ckeditor.com/docs/ckeditor5/latest/framework/guides/tutorials/implementing-an-inline-widget.html#demo? 

I'd like it to generate (i.e. the strong is added outside the placeholder):

```html
<p>xxx <strong><placeholder>text</placeholder></strong></p>
```

I think it comes from: https://github.com/ckeditor/ckeditor5/issues/1633.

niegowski:
> Out of curiosity: can we make the bold feature work in this case https://ckeditor.com/docs/ckeditor5/latest/framework/guides/tutorials/implementing-an-inline-widget.html#demo?
> 
> I'd like it to generate (i.e. the strong is added outside the placeholder):
> 
> ```html
> <p>xxx <strong><placeholder>text</placeholder></strong></p>
> ```
> 
> I think it comes from: #1633.

I updated the PR to properly handle this case. Steps that are required:

*   inline widget schema should add `allowAttributesOf: '$text'`
*   downcast conversion should provide the option to use inline `ContainerElement` (so it could be wrapped by an `AttributeElement`):

```javascript
const widgetElement = writer.createContainerElement( 'placeholder', null, { isInline: true } );
```

*   widget inserting command should provide selection attributes to the inserted model element (to enable downcasting of a model attribute to an `AttributeElement` in the same way as `InputCommand` does for text nodes):

```javascript
const attributes = model.document.selection.getAttributes();

const placeholder = writer.createElement( 'placeholder', {
	...Object.fromEntries( attributes ),
	type: 'placeholder'
} );
```

Also, I noticed that the only inline element that we currently support (`softBreak`) is breaking the `AttributeElement` and I'm not sure if this is the desired behavior.

packages/ckeditor5-engine/src/view/downcastwriter.js
```diff
@@ -685,43 +695,46 @@ export default class DowncastWriter {
...
-			 * @error view-writer-invalid-position-container
-			 */
-			throw new CKEditorError(
-				'view-writer-invalid-position-container',
-				this.document
-			);
-		}
+			// Break attributes on nodes that do exist in the model tree so they can have attributes, other elements
+			// can't have an attribute in model and won't get wrapped with an AttributeElement while down-casted.
+			const breakAttributes = !node.is( 'uiElement' );
```
niegowski, line 704: I'm not sure if this should not include `rawElement` also.

niegowski: But OTOH we could use `RawElement` (instead of `UIElement`) for suggestions in TC.

Reinmar:
Notes from the review: 

---

- Custom property vs private property?
    - Custom props is more for integrations
    - So, let's go with priv prop, just like the priority of attr elements
    - But, does it clone right now? **TODO**
- Do we need this property at all?
    - bold - no difference. Image can be inside <strong>
    - linkHref – is handled in a custom way on  (could be handled by converters)
    - Idea: The first common container element ancestor is the limit of attribute operations.
        - `<attrelem:foo>`
            - `<containerelem:bar>`
                - ... operations on attr elements here are limitted by the `<containerelem:bar>`
    - PK: For me, differentiating inline and block container elements might be useful.
    - SC: Model is controllable. View is generated based on this model and should implement the least possible restriction.
    - PK: For us, restrictions mean problems. For others: could be the other way around as restrictions give directions on how to do things.
    - Position mapping could be a problem. If you'd have `<strong><p>xx</p></strong>` the position after `</p>` will map to something weird in the model.
    - Cases:
        - Backwards compat
        - Future safety
        - View source
        - Weird data structure
- isInline is not a good name for this property
    - Why are UI/Raw/Empty elements "inline"? It's more about that attribute element should wrap this element.
    - isAllowedInAttributeElement.
- The current PR changes the behavior of suggestions
    - Previously: a suggestion markers splits attribute elements. Now: it lands inside.
    - Is it an important breaking change?
        - What if someone extracts content of suggestions by cutting out a piece between two suggestion markers?
        - There's a plugin for post-processing suggestions so we shouldn't need to worry.

Reinmar:
> * But, does it clone right now? **TODO**

If we go with the property – is it automatically copied when you clone an element?

And the second thing – what should be the default value of this property for UI/Raw/Empty elements? "allow in attr elements" seem safe and fine in this case.

niegowski:
> If we go with the property – is it automatically copied when you clone an element?

I missed that part - I'll add it. Also, I should update `isSimilar()`

> And the second thing – what should be the default value of this property for UI/Raw/Empty elements? "allow in attr elements" seem safe and fine in this case.

I think that it should trigger the same behavior as the current implementation to replace `isEmpty || isUI || isRaw` while wrapping nodes. But changing the property name to `allowedInline` or `allowedInAttributeElement` would solve the confusion.

niegowski:
I think that we should use some property to indicate `allowedInline` for the cases when we are mapping a single element in the model to multiple elements in the view. For example an `image` with an `linkHref` attribute is downcasted to `<figure><a></a></figure>` and in this case, we don't want to end up with `<a><figure></figure></a>`

packages/ckeditor5-engine/src/view/rawelement.js
```diff
@@ -47,6 +47,9 @@ export default class RawElement extends Element {
 	constructor( document, name, attrs, children ) {
 		super( document, name, attrs, children );
 
+		// Override the default of a base class.
```
Reinmar, line 50:
```suggestion
		// Override the default of the base class.
```

packages/ckeditor5-engine/src/view/element.js
```diff
@@ -151,6 +160,17 @@ export default class Element extends Node {
 		return this._children.length === 0;
 	}
 
+	/**
+	 * Whether an element is inline and can be wrapped with {@link module:engine/view/attributeelement~AttributeElement}
```
Reinmar, line 164:
```suggestion
	 * Whether the element is inline and can be wrapped with {@link module:engine/view/attributeelement~AttributeElement}
```

packages/ckeditor5-engine/src/view/downcastwriter.js
```diff
@@ -1145,6 +1158,72 @@ export default class DowncastWriter {
 		return new Selection( selectable, placeOrOffset, options );
 	}
 
+	/**
+	 * Inserts a node or nodes at specified position. Takes care about breaking attributes before insertion
```
Reinmar, line 1162:
```suggestion
	 * Inserts a node or nodes at the specified position. Takes care of breaking attributes before insertion
```

Reinmar:
After changing the opinion many times already, I'd go with the most specific `isAllowedInsideAttributeElement`.

If we'll ever have to have a more general `isInline` or `allowInline` we can set `isAllowedInsideAttributeElement` based on this more generic property. Let's just make sure we spot this moment when it's not only about attribute elements.

Also, please make sure to mention that property in `DW#wrap()` and `DW#createAttributeElement()` and `AttributeElement` itself.

Finally, we may consider allowing passing `isAllowedInsideAttributeElement` to createUI/Raw/EmptyElement too. It would make sense to make this controllable, especially for UIElements as they do not map to anything in the model so you cannot control that via model's schema or converters.

niegowski:
Ready for another review round.

Reinmar (changes requested):
I'm missing some tests for the element classes themselves. There are test for create* methods in the DW, but the changes actually affected the element classes and it'd be good to cover that too.

Also, @scofalik could you review changes in the downcast writer? I checked all other files.

niegowski:
> I'm missing some tests for the element classes themselves. There are test for create\* methods in the DW, but the changes actually affected the element classes and it'd be good to cover that too.

I added more tests.

packages/ckeditor5-engine/src/view/downcastwriter.js
```diff
@@ -877,6 +920,16 @@ export default class DowncastWriter {
 	 * Throws {@link module:utils/ckeditorerror~CKEditorError} `view-writer-wrap-nonselection-collapsed-range` when passed range
 	 * is collapsed and different than view selection.
 	 *
+	 * **Note:** Attribute elements by default can wrap {@link module:engine/view/text~Text},
+	 * {@link module:engine/view/emptyelement~EmptyElement}, {@link module:engine/view/uielement~UIElement},
+	 * {@link module:engine/view/rawelement~RawElement} and other attribute elements with higher priority. Other elements while placed
+	 * inside an attribute element will split it (or nest it in case of an `AttributeElement`). This behavior can be modified by changing
+	 * the `isAllowedInsideAttributeElement` option while creating
```
Reinmar, line 927:
```suggestion
	 * the `isAllowedInsideAttributeElement` option while using
```

PR #9011, opened 2021-02-10
Created the `CloudServicesCore` plugin

### LINKED ISSUE #8811: Turn cloud-services-core into a soft requirement

Right now it's part of the DLL. However, it's a bit problematic because it's not a commonly used package.

Some of my thoughts:

> - We need to turn this into a soft requirement.
> - We considered having two DLLs (the base one and a richer, collab-ready one with more packages included), but the fact that packages from DLL must be imported via `ckeditor5` is a deal breaker because that'd mean that in the source files we would need to change imports from `cloud-services-core` to `ckeditor5`-based. This would in turn means that such packages need to be added to `ckeditor5`. While this is more or less fine for `cloud-services-core` which is part of the OSS part, this strategy with having two DLLs would work best if we could use this for commercial packages.
> - Also, if we'd want to use this collab-ready DLL for solving the problem with comments editor using e.g. basic styles, we'd need to also add basic-styles to the DLL (and hence `ckeditor5`. Same for the editor class. That would make it even more messy.
> - Anyway, this would be confusing too. It gets more complex the longer we think about this.
> - It seems that soft requirements, and perhaps some tricks to make them nice to use, is our only options. Soft requirement means a "dynamic requirement" and it's a bit like in the old RPM days when you knew you need to load some lib but the system would not install it automatically for you.

In order to turn it into a soft requirement of other packages:

* We need to check which things from this packages are imported in other packages.
* These things need to be somehow moved to the plugin(s) to make them available to other plugins.
* Code that was directly importing these things previously needs to be updated to use these things via the plugins. These plugins need to be added as soft requirements of code that consumes them.

pomek:

Other (cloud-services-core): All available in the package classes have been moved to the `@ckeditor/ckeditor5-cloud-services` package. Closes #8811.

Feature (cloud-services): Created the `CloudServicesCore` that exposes the base API for communication with CKEditor Cloud Services.

Feature (core): Added the third argument for the `PluginCollection#init()` method that allows replacing built-in plugins with a custom implementation. It's helpful while mocking plugins that are built-in the editor or the Context feature.

MAJOR BREAKING CHANGE (cloud-services-core): The package has been merged into the `@ckeditor/ckeditor5-cloud-services`. Make sure to update the name of the package from where the classes are imported.

---

### Additional information


pomek:
Thanks to that change, we are able to remove our "hacks" in the code

https://github.com/ckeditor/ckeditor5/blob/f41b90ebbb8ed1f110431d86e5cf6b6f71b1c9ce/packages/ckeditor5-easy-image/src/cloudservicesuploadadapter.js#L90-L92

pomek (commented):
Temporary R-, we changed the idea and I need to adjust the PR here.

pomek:
Ready for review.

packages/ckeditor-cloud-services-core/src/cloudservicescore.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @module cloud-services-core/cloudservicescore
+ */
+
+import { ContextPlugin } from 'ckeditor5/src/core';
+import Token from './token/token';
+import UploadGateway from './uploadgateway/uploadgateway';
+
+/**
+ * The `CloudServicesCore` plugin that exposes API for file uploads to CKEditor Cloud Services.
```
Reinmar, line 15:
```suggestion
 * The `CloudServicesCore` plugin exposes the base API for communication with CKEditor Cloud Services.
```

packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -165,10 +165,12 @@ export default class PluginCollection {
 	 * or {@link module:core/plugin~PluginInterface.pluginName plugin names}.
 	 * @param {Array.<String|Function>} [removedPlugins] Names of the plugins or plugin constructors
 	 * that should not be loaded (despite being specified in the `plugins` array).
+	 * @param {Array.<Function>} pluginsToReplace An array of {@link module:core/plugin~PluginInterface plugin constructors}
+	 * that will be used for replacing plugins definitions while instantiating.
```
Reinmar, line 169:
```suggestion
	 * that will be used to replace plugins of the same names that were passed in
	 * `plugins` or that are in their dependency tree.
```

packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -165,10 +165,12 @@ export default class PluginCollection {
 	 * or {@link module:core/plugin~PluginInterface.pluginName plugin names}.
 	 * @param {Array.<String|Function>} [removedPlugins] Names of the plugins or plugin constructors
 	 * that should not be loaded (despite being specified in the `plugins` array).
+	 * @param {Array.<Function>} pluginsToReplace An array of {@link module:core/plugin~PluginInterface plugin constructors}
```
Reinmar, line 168:
```suggestion
	 * @param {Array.<Function>} pluginsSubstitutions An array of {@link module:core/plugin~PluginInterface plugin constructors}
```

Reinmar: Also, I'd rename `removedPlugins` to `pluginsToRemove`.

packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -423,6 +427,83 @@ export default class PluginCollection {
 				return promise.then( plugin[ method ].bind( plugin ) );
 			}, Promise.resolve() );
 		}
+
+		/**
```
Reinmar, line 431: This is an internal function so it should be documented with `//`

packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -423,6 +427,83 @@ export default class PluginCollection {
...
+		 * Replaces plugin constructors with the specified set of plugins. A useful option for replacing built-in plugins while creating
```
Reinmar, line 432: The "useful" part makes sense in `init()`, not here. That's where you might be looking for it.

packages/ckeditor5-core/tests/plugincollection.js
```diff
@@ -553,6 +553,209 @@ describe( 'PluginCollection', () => {
 						.to.deep.equal( [ 'A', 'B', 'Foo', 'C', 'D' ], 'order by returned value' );
 				} );
 		} );
+
+		describe( '#replacePlugins', () => {
```
Reinmar, line 557:
```suggestion
		describe( 'substituting plugins', () => {
```

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -237,8 +237,9 @@ export default class Editor {
 		const plugins = config.get( 'plugins' );
 		const removePlugins = config.get( 'removePlugins' ) || [];
 		const extraPlugins = config.get( 'extraPlugins' ) || [];
+		const replacePlugins = config.get( 'replacePlugins' ) || [];
```
Reinmar, line 240:
```suggestion
		const replacePlugins = config.get( 'substitutePlugins' ) || [];
```

packages/ckeditor5-core/src/plugincollection.js
Reinmar: I'm for `pluginsSubstitutions` as here we're not passing the plugins to be replaced but their replacements and `pluginReplacements` sounds wrong. 

But on the funny note – these words are causing some issues: https://queens-english-society.org/substitute-and-replace/

Reinmar:
Apart from some naming aspects, it looks very good.

pomek:
@Reinmar, thanks for the review. I applied all suggestions.

Reinmar:
I removed a couple more things from the code and this is what's left:



I'd say it looks good, but it also means that we can have a followup to remove this from the last two files. For safety reasons, I'd do that after the release. Can you create a followup?

Reinmar:
I've reviewed the changes. Tests are green, docs are ok. The changes (in all repos) are complete ❤️

One thing before the merge: rewriting the merge message.

pomek:
> One thing before the merge: rewriting the merge message.

Done.

> I'd say it looks good, but it also means that we can have a followup to remove this from the last two files. For safety reasons, I'd do that after the release. Can you create a followup?

We can't remove it now. It must be done after the release.

packages/ckeditor5-core/src/plugincollection.js
```diff
@@ -163,12 +163,19 @@ export default class PluginCollection {
 	 *
 	 * @param {Array.<Function|String>} plugins An array of {@link module:core/plugin~PluginInterface plugin constructors}
 	 * or {@link module:core/plugin~PluginInterface.pluginName plugin names}.
-	 * @param {Array.<String|Function>} [removedPlugins] Names of the plugins or plugin constructors
+	 * @param {Array.<String|Function>} [pluginsToRemove] Names of the plugins or plugin constructors
 	 * that should not be loaded (despite being specified in the `plugins` array).
+	 * @param {Array.<Function>} pluginsSubstitutions An array of {@link module:core/plugin~PluginInterface plugin constructors}
```
pomek, line 168:
```suggestion
	 * @param {Array.<Function>}[pluginsSubstitutions] An array of {@link module:core/plugin~PluginInterface plugin constructors}
```

PR #9023, opened 2021-02-12
BubblingEmitterMixin

### LINKED ISSUE #7437: 2SCM fails when a link is next to an image (widget)

1. `yarn run manual -f ckeditor5`
2. http://localhost:8125/ckeditor5/tests/manual/all-features.html
3. Paste to the console:  
```
editor.setData( '<p><a href="https://cksource.com">https://cksource.com</a></p><figure class="image"><img src="sample.jpg" alt=""></figure>' );
```
4. Click at the end of the link and use the keyboard to put the selection after the link.

### ✔️ Expected result

The selection loses the `linkHref` attribute and I can type after the link.

image

### ❌ Actual result

The image is selected.

image

---

### LINKED ISSUE #8640: Research bubbling model events

There are 3 event types that currently use a hacky way of prioritizing handlers: `enter`, `delete`, `keydown` (for arrow keys). Those listeners often use `priority.high + [some number]` (or similar) to make sure that for example handler for a list item that is inside a blockquote will be triggered before the blockquote handler.

There is an idea to introduce some sort of bubbling events over the model elements. Features could register that their handler is supposed to be executed if a given element name was encountered while bubbling up the model document tree (or custom matcher for widgets etc.)

niegowski:

Feature (engine): Introducing `BubblingEmitterMixin` and `ArrowKeysObserver`. Closes #8640.

Fix (utils): The `EmitterMixin#listenTo()` method is split into listener and emitter parts. The `ObservableMixin` decorated methods reverted to the original method while destroying an observable.

Other (typing): The `TwoStepCaretMovement` is now using bubbling events. Closes #7437.

BREAKING CHANGE: We introduced bubbling of `view.Document` events, similar to how bubbling works in the DOM. That allowed us to reprioritize many listeners that previously had to rely on `priority`. However, it means that existing listeners that use priorities may now be executed in a wrong moment. The listeners to such events should be reviewed in terms of when they should be executed (in what context/element/phase). You can find more information regarding bubbling in the documentation: \[TODO\].

Internal (widget): The enter, delete and arrow key events handling moved to the usage of the bubbling observer.

Internal (block-quote): The enter and delete events handling moved to the usage of the bubbling observer.

Internal (code-block): The enter event handling moved to the usage of the bubbling observer.

Internal (list): The enter and delete events handling moved to the usage of the bubbling observer.

Internal (table): The arrow keys handling moved to the usage of the bubbling observer.

---

### Additional information


Reinmar:
We discussed the current approach with @scofalik and we've got similar feelings:

* The new API looks good.
* We found implementing the extension of view Document's behavior in an observer confusing. This changes the behavior, not adds one. Also, the concept of observers was targetting extending the view Document with more events. While it's possible to achieve more with the API that it offers, it's not a good idea.

IMO, a better approach could be to have the view Document mix a new mixin called `BubblingEmitterMixin`. It would implement a `BubblingEmitter` interface. And this interface would be extending the `Emitter` interface (I hope JSDoc handles such relations). At least – I hope it can be done this way. The code could land directly in `src/view/` or `src/view/observers`. Both have their pros and cons. The latter stands against mixing observers + this extension argument, however, there's a strong relationship between both so I'd be for it.

As for moving this "bubbling emitter util" to utils – I think we can all agree that it's an unnecessaery complication. As long as we use it one place, let's have it specialized for this purpose.

Other than that – I think that the simplifications that you did fully justify this change. Big 👍  in general.

Reinmar:
I left some comments in [https://github.com/cksource/ckeditor5-internal/pull/542#pullrequestreview-590588457.](https://github.com/cksource/ckeditor5-internal/pull/542#pullrequestreview-590588457.) I think we need to stop for a second and design the expected behavior of this mechanism. Since it will be built into an existing, more commonly known mechanism, it needs to be well blended in.

niegowski:
I updated the ADR and the implementation.

scofalik (approved):
API looks fine to me. I am not going deep into implementation. I am not sure if there's a better way to implement that, so I won't be trying to think about something better. One thing that I didn't get yet, and TBH I don't want to spend more time looking for this -- is the information for which element the event is fired available somewhere (`eventInfo`)?

One thought, though. I don't want to complicate things but I have a slight dislike that this is view-only thing and it is an "emitter mixin" that ends up in the engine, not in the core, where this kind of stuff is. I wonder if it would be possible to generalize this solution.

I know that this problem is specific to the view structure but I think it could be generalized to:

1. We have a new property that changes the order of events firing (context).
2. Whenever an event is fired, we call it multiple times with different values of context.
3. So, we need to know what context values we should fire.

The difficult part is point 3. and AFAICS in the code, a lot of the code is for handling that. If we would know what context values should be "fired", we could generalize the rest of the code.

So, how about introducing `BubblingEmitterMixin` or `ContextEmitterMixin` as a general solution. Then, a class mixing that mixin would have to provide a way to evaluate context values (basically lines https://github.com/ckeditor/ckeditor5/pull/9023/files#diff-a8baf73f43f2cbd49cbd55c3ea3bcc79249aef04f2550605191e867767a42c0fR42-R83). For example, maybe as a static method, but it could be something like

```js
this._setContextCallback( ( ... ) => {
	// The capture phase of the event.
	if ( fireListenerFor( eventContexts, '$capture', eventInfo, ...eventArgs ) ) {
		return eventInfo.return;
	}

	// TODO Instead of using this.selection we could pass range in EventInfo.
	const selectionRange = this.selection.getFirstRange();
	const selectedElement = selectionRange ? selectionRange.getContainedElement() : null;
	const isCustomContext = selectedElement ? Boolean( getCustomContext( eventContexts, selectedElement ) ) : false;

	// For the not yet bubbling event trigger for $text node if selection can be there and it's not a custom context selected.
	if ( !isCustomContext && fireListenerFor( eventContexts, '$text', eventInfo, ...eventArgs ) ) {
		return eventInfo.return;
	}

	let node = selectedElement || getDeeperRangeParent( selectionRange );

	while ( node ) {
		// Root node handling.
		if ( node.is( 'rootElement' ) ) {
			if ( fireListenerFor( eventContexts, '$root', eventInfo, ...eventArgs ) ) {
				return eventInfo.return;
			}
		}

		// Element node handling.
		else if ( node.is( 'element' ) ) {
			if ( fireListenerFor( eventContexts, node.name, eventInfo, ...eventArgs ) ) {
				return eventInfo.return;
			}
		}

		// Check custom contexts (i.e., a widget).
		if ( fireListenerFor( eventContexts, node, eventInfo, ...eventArgs ) ) {
			return eventInfo.return;
		}

		node = node.parent;
	}

	// Document context.
	fireListenerFor( eventContexts, '$document', eventInfo, ...eventArgs );
} );
```

Maybe it could be even implement straight in `EmitterMixin`?

---

While writing that, I also had another idea. We already support namespaces. So, my idea was to use namespaces to handle contexts. For example:

```js
this.view.document.on( 'enter:li', ( ... ) => { ... } );
```

Of course, you would still need a mechanism that would fire appropriate events in the appropriate order and it would probably be implemented in view document. Good thing is that the changes would limit to the view. There are some disadvantages to this idea, though, for example, backward compatibility (events registered for `enter` would fire for every event like `enter:li`, `enter:$text` and there would be problem with priorities). Also, we wouldn't have a general solution ;).

niegowski:
> One thing that I didn't get yet, and TBH I don't want to spend more time looking for this -- is the information for which element the event is fired available somewhere (`eventInfo`)?

It's not but there are TODO notes in the code to consider introducing something like it.

scofalik:
Sorry, I didn't include important note regarding the general solution: the "context callback function" would return an array with context names, not fire actual events. The array would be used by the mixin to fire events.

packages/ckeditor5-utils/src/observablemixin.js
```diff
@@ -258,11 +261,33 @@ const ObservableMixin = {
...
+		}
+
+		this[ _decoratedMethods ].push( methodName );
 	}
 };
 
 extend( ObservableMixin, EmitterMixin );
 
+// Override the EmitterMixin stopListening method to be able to clean decorated methods.
+ObservableMixin.stopListening = function( emitter, event, callback ) {
```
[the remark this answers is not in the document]
niegowski: This is a sidetrack that I noticed in the previous attempt at bubbling events. Some tests started to fail because of `ModelTestEditor` is destroying a part of the editor and it causes that all events stop firing (including those for decorated methods so there is no way to call the original method). BTW this can be removed from this PR because it is no longer causing a problem.

niegowski:
Notes from F2F talk with @Reinmar:

*   We don't want to generalize the solution because currently we don't see another usage for it (we could change it in the future)
*   The name is ok because it's inside the view part of the engine code (we already have document, node, element in both view and model parts of the engine)
*   We could use dedicated BubblingEventInfo while triggering the event so we could provide a custom "start position" for the bubbling so it could start bubbling from the click position instead of the current selection
*   We will add eventPhase and currentContext to the BubblingEventInfo (as mentioned in the TODOs in code)

packages/ckeditor5-engine/src/view/observer/arrowkeysobserver.js
```diff
@@ -0,0 +1,56 @@
...
+ * @module engine/view/observer/arrowkeysobserver
+ */
+
+import Observer from './observer';
+import BubblingEventInfo from './bubblingeventinfo';
+
+import { isArrowKeyCode } from '@ckeditor/ckeditor5-utils';
+
+/**
+ * Arrow keys observer introduces the {@link module:engine/view/document~Document#event:arrowKey} event.
```
Reinmar, line 16: Please align it a bit to match other observers' docs.

packages/ckeditor5-engine/tests/view/observer/bubblingeventinfo.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import BubblingEventInfo from '../../../src/view/observer/bubblingeventinfo';
+
+import EventInfo from '@ckeditor/ckeditor5-utils/src/eventinfo';
+
+describe( 'EventInfo', () => {
```
Reinmar, line 10: Update the name :d

packages/ckeditor5-enter/src/enter.js
```diff
@@ -46,7 +46,8 @@ export default class Enter extends Plugin {
 			}
 
 			editor.execute( 'enter' );
+
 			view.scrollToTheSelection();
-		}, { priority: 'low' } );
```
Reinmar, line 50: I'd keep this. We listened on `low` to make sure that if someone didn't pass any priority (and want to override this), that they don't have to do anything else.

Same in: shift enter, delete and perhaps somewhere else.

packages/ckeditor5-enter/src/shiftenter.js
```diff
@@ -68,6 +68,6 @@ export default class ShiftEnter extends Plugin {
 
 			editor.execute( 'shiftEnter' );
 			view.scrollToTheSelection();
-		}, { priority: 'low' } );
```
Reinmar, line 71: Same here.

packages/ckeditor5-utils/src/emittermixin.js
```diff
@@ -638,22 +685,23 @@ function fireDelegatedEvents( destinations, eventInfo, fireArgs ) {
...
-				callbacks.splice( i, 1 );
-				i--;
-			}
-		}
+// Helper for registering event callback on the emitter.
+function addEventListener( listener, emitter, event, callback, options ) {
+	if ( emitter._addEventListener ) {
+		emitter._addEventListener( event, callback, options );
+	} else {
+		// Allow listening on objects that do not implement Emitter interface.
```
Reinmar, line 693: Clarify that the use cases is in the tests.

packages/ckeditor5-utils/src/emittermixin.js
```diff
@@ -638,22 +685,23 @@ function fireDelegatedEvents( destinations, eventInfo, fireArgs ) {
...
+		listener._addEventListener.call( emitter, event, callback, options );
+	}
+}
+
+// Helper for removing event callback from the emitter.
+function removeEventListener( listener, emitter, event, callback ) {
+	if ( emitter._removeEventListener ) {
+		emitter._removeEventListener( event, callback );
+	} else {
+		// Allow listening on objects that do not implement Emitter interface.
```
Reinmar, line 703: Clarify that the use cases is in the tests.

packages/ckeditor5-utils/src/observablemixin.js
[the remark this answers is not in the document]
Reinmar: 1.  Have `x.foo()` decorated
2.  Call `x.stopListening()`
3.  Call `x.foo()`. Problem: nothing happens (the original `foo()` method is not executed)

While not super popular scenario, it's possible. So it'd be good to have it covered. I'd keep this code, but we need to have it explained in the code clearly why this is needed.

Reinmar:
BREAKING CHANGE: We introduced bubbling of `view.Document` events, similar to how bubbling works in the DOM. That allowed us to reprioritize many listeners that previously had to rely on `priority`. However, it means that existing listeners that use priorities may now be executed in a wrong moment. The listeners to such events should be reviewed in terms of when they should be executed (in what context/element/phase). You can find more information regarding bubbling in the documentation: \[TODO\].

niegowski:
@ckeditor/qa-team could you please test it? This is a massive change in event handling. Please focus on the custom enter/delete/delete forward/arrow keys handling in the context of blockquotes, lists, to-do lists, block widgets (widget type around), tables, links (2 step caret movement)

Reinmar:
> @ckeditor/qa-team could you please test it? This is a massive change in event handling. Please focus on the custom enter/delete/delete forward/arrow keys handling in the context of blockquotes, lists, to-do lists, block widgets (widget type around), tables, links (2 step caret movement)

I'd add that it's a lot about nesting. Nesting things inside other things. Like blockquotes in tables, images in block quotes, etc. Those are the most tricky cases. But the PR could've affected also more typical ones with flat structures.

Mgsy:
I've tested it and haven't noticed any weird behaviours. It seems to be fine :+1:

PR #9028, opened 2021-02-14
I/8871 inline images link integratation

### LINKED ISSUE #8871: Integrate inline images with the link feature

pkwasnik:

Internal (link): Integrate inline images with the link feature. Closes #8871.

---

### Additional information


niegowski:
Since we merged #8998 yesterday, I'm not sure if this PR shouldn't be updated to use AttributeElement wrapping the ContainerElement? WDYT @Reinmar ?

oleq:
If that would simplify the solution and make the feature better, I'm for it.

niegowski:
> If that would simplify the solution and make the feature better, I'm for it.

I think it would simplify conversion, this are the changes that are needed to enable links over inline widgets: https://github.com/ckeditor/ckeditor5/compare/v25.0.0...master#diff-439efd180fbb2493acdc56d584f8b2af2ecc5b49c1ca587d876f6a8045b6c511

niegowski:
I updated previous comment with better link

Reinmar:
> Since we merged #8998 yesterday, I'm not sure if this PR shouldn't be updated to use AttributeElement wrapping the ContainerElement? WDYT @Reinmar ?

What will be the outcome for the user? Or is the goal simplifying the implementation? To be able to asses a change we need to understand the "return" from it.

niegowski:
> What will be the outcome for the user? Or is the goal simplifying the implementation? To be able to asses a change we need to understand the "return" from it.

It would be simpler implementation and links would not break on the inline image.

oleq:
> It would be simpler implementation and links would not break on the inline image.

I think we should not increase the technical debt if we can especially that we have the right tool for the job now. @pkwasnik can you please update your PR using the technique linked by Kuba in https://github.com/ckeditor/ckeditor5/pull/9028#issuecomment-784071079?

pkwasnik:
> I think we should not increase the technical debt if we can especially that we have the right tool for the job now. @pkwasnik can you please update your PR using the technique linked by Kuba in #9028 (comment)?

yes, I will do that

oleq:
Closing this one in favor of #9438.

PR #9038, opened 2021-02-16
Placeholder should not be visible if there are multiple elements in the root

### LINKED ISSUE #9009: Editor placeholder is visible with added content

It's possible to add content and the editor placeholder is still visible. Also in some cases placeholder can be affected by styles like heading.

### Case 1 : 
1. Go to http://localhost:8125/ckeditor5/tests/manual/all-features.html on **current master**
2. Remove all content.
3. Place the caret at the first line.
4. Press enter.
5. Start typing.
CKE5-placeholder1

### Case 2:
1. Go to http://localhost:8125/ckeditor5/tests/manual/all-features.html on **current master**
2. Remove all content.
3. Add Heading as the first line.
4. Remove text with backspace - placeholder will have heading 1 styles.
CKE5-placeholder2

### Case 3:
1. Go to http://localhost:8125/ckeditor5/tests/manual/all-features.html on **current master**
2. Remove all content.
3. Add heading as a first line.
4. Add paragraph as a second line.
5. Remove all text by holding backspace.
6. Undo.
CKE5-placeholder3

### ❌ Actual result

Text `Type the content here!` is still visible and it can change styles.

## 📃 Other details

* My browser: Chrome v88
* OS: macOS BigSur 11.1
* First affected CKEditor version: current master branch, it's not affecting released v25.0.0
* Is it caused by this PR? https://github.com/ckeditor/ckeditor5/pull/8867

---

niegowski:

Internal (placeholder): Placeholder should not be visible if there are multiple elements in the root. Closes #9009.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/view/placeholder.js
+++ b/packages/ckeditor5-engine/src/view/placeholder.js
@@ -229,7 +229,10 @@ function updatePlaceholder( writer, element, config ) {
 		wasViewModified = true;
 	}
 
-	if ( needsPlaceholder( hostElement, config.keepOnFocus ) ) {
+	// If the host element is not a direct host then placeholder is needed only when there is only one element.
+	const isOnlyChild = isDirectHost || element.childCount == 1;
+
+	if ( isOnlyChild && needsPlaceholder( hostElement, config.keepOnFocus ) ) {
 		if ( showPlaceholder( writer, hostElement ) ) {
 			wasViewModified = true;
 		}
@@ -248,7 +251,7 @@ function updatePlaceholder( writer, element, config ) {
 // @param {module:engine/view/element~Element} parent
 // @returns {module:engine/view/element~Element|null}
 function getChildPlaceholderHostSubstitute( parent ) {
-	if ( parent.childCount === 1 ) {
+	if ( parent.childCount ) {
 		const firstChild = parent.getChild( 0 );
 
 		if ( firstChild.is( 'element' ) && !firstChild.is( 'uiElement' ) ) {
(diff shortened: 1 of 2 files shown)


niegowski:
@ckeditor/qa-team could you test it?

packages/ckeditor5-engine/src/view/placeholder.js
```diff
@@ -229,7 +229,10 @@ function updatePlaceholder( writer, element, config ) {
 		wasViewModified = true;
 	}
 
-	if ( needsPlaceholder( hostElement, config.keepOnFocus ) ) {
+	// If the host element is not a direct host then placeholder is needed only when there is only one element.
+	const isOnlyChild = isDirectHost || element.childCount == 1;
+
+	if ( isOnlyChild && needsPlaceholder( hostElement, config.keepOnFocus ) ) {
```
mlewand, line 235: Wouldn't the new check logic make more sense in `needsPlaceholder` method? It will require us to pass more variables but currently `needsPlaceholder` is the only place that determines whether placeholder is visible or not, with these changes we have two such places.

mlewand (commented):
The code works fine 👍 I have one question though.

packages/ckeditor5-engine/src/view/placeholder.js
niegowski: I consider `needsPlaceholder` as a check for a single element (it's used in the Title plugin to check individual elements for the need for a placeholder). I would not propagate the `isDirectHost` option deeper because it's only complicating the code without adding any value.

mlewand: Good point on it being used in title plugin I assumed that it's only used in engine. OK, in this case it's good for now but as soon as we have more conditions like that, it make sense to push it to a separate function that also calls `needsPlaceholder`.

mlewand (approved):
LGTM, we can merge it as soon as we have a R+ from @LukaszGudel.

LukaszGudel:
Looking good to me 👍  I did not find any case where the placeholder would have strange behavior

PR #9057, opened 2021-02-20
The keystrokes are no longer conflicting on MacOS

### LINKED ISSUE #5705: Some CKEditor keyboard shortcuts conflict with Mac OS shortcuts

CKEditor uses key shortcuts which conflict with Mac OS shortcuts.  For example for Bold and for Link both shortcuts work: `CMD+B` and `Ctrl+B` for Bold, `CMD+K` and `Ctrl+K` for Link. The problem is that on Mac `Control-K` means "Delete the text between the insertion point and the end of the line or paragraph" and `Control-B` means "Move one character backward" - https://support.apple.com/en-us/HT201236.

The reason is because when meta (Cmd) key is pressed, it is treated as `Ctrl` key by the editor.

### ✔️ Expected result

When `Ctrl+K` or `Ctrl+B` is pressed, Mac OS commands should be executed.
CKEditor commands should only be executed when `Cmd+K` or `Cmd+B` is pressed. 

### ❌ Actual result

When `Ctrl+K`, `Cmd+K` or `Ctrl+B`, `Cmd+B` is pressed, CKEditor commands are executed.
Mac OS commands are not executed at all.

## 📃 Other details

* Browser: Any
* OS: Mac OS
* CKEditor version: 15

---

niegowski:

Fix (utils): The keystrokes are no longer conflicting on macOS. Closes #5705.

Feature (utils): Added forced modifier key (`Ctrl!`) for keystrokes that should not be mapped to <kbd>Command</kbd> on macOS.

Other (engine): The `KeyObserver` should provide information about `metaKey` being pressed.

Other (list): The to-do list item toggle keystroke changed to <kbd>Ctrl</kbd>+<kbd>Enter</kbd> (<kbd>Cmd</kbd>+<kbd>Enter</kbd> on Mac).

Internal (typing): Any keypress while the <kbd>Cmd</kbd> key is pressed is not a typing keystroke.

BREAKING CHANGE: On macOS keystrokes with the <kbd>Ctrl</kbd> modifier will not be handled unless the modifier is registered as the forced one (for example: `Ctrl!+A` will not be translated to `Cmd+A` on macOS).

BREAKING CHANGE (list): The to-do list item toggle keystroke changed to <kbd>Ctrl</kbd>+<kbd>Enter</kbd> (<kbd>Cmd</kbd>+<kbd>Enter</kbd> on Mac).

---

### Additional information


packages/ckeditor5-utils/src/keyboard.js
```diff
@@ -108,22 +114,20 @@ export function parseKeystroke( keystroke ) {
...
-	if ( !env.isMac ) {
-		return keystroke;
-	}
+	let keystrokeCode = parseKeystroke( keystroke );
+
+	const modifiersToGlyphs = Object.entries( env.isMac ? modifiersToGlyphsMac : modifiersToGlyphsNonMac );
+
+	const modifiers = modifiersToGlyphs.reduce( ( modifiers, [ name, glyph ] ) => {
+		if ( ( keystrokeCode & keyCodes[ name ] ) != 0 ) {
+			keystrokeCode &= ~keyCodes[ name ];
```
maxbarnas, line 123: I am wondering if we should add a comment with a short explanation/description of what actually is happening here. It's not totally obvious at first and this construction is not used often in our codebase.

maxbarnas (changes requested):

packages/ckeditor5-utils/src/keyboard.js
maxbarnas: Also, bloody hell, this is sooo clean! 🎉

packages/ckeditor5-enter/src/enterobserver.js
```diff
@@ -23,7 +23,7 @@ export default class EnterObserver extends Observer {
 		const doc = this.document;
 
 		doc.on( 'keydown', ( evt, data ) => {
-			if ( this.isEnabled && data.keyCode == keyCodes.enter ) {
+			if ( this.isEnabled && data.keyCode == keyCodes.enter && !data.ctrlKey && !data.metaKey ) {
```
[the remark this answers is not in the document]
niegowski: It's because `editor.keystrokes` is not stopping the original event so it splits the list item while pressing `Cmd+Enter`. Now I'm wondering if we should change the `KeystrokeHandler` behavior or this code.

packages/ckeditor5-list/src/todolistediting.js
```diff
@@ -113,7 +113,7 @@ export default class TodoListEditing extends Plugin {
 		this.listenTo( editing.view.document, 'keydown', jumpOverCheckmarkOnSideArrowKeyPress( model, editor.locale ) );
 
 		// Toggle check state of selected to-do list items on keystroke.
-		editor.keystrokes.set( 'Ctrl+space', () => editor.execute( 'checkTodoList' ) );
+		editor.keystrokes.set( 'Ctrl+Enter', () => editor.execute( 'checkTodoList' ) );
```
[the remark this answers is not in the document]
niegowski: The original keystroke discussion was here: [https://github.com/ckeditor/ckeditor5/issues/1434#issuecomment-514178310](https://github.com/ckeditor/ckeditor5/issues/1434#issuecomment-514178310)

I decided to change this keystroke to match Notion and avoid conflict with `Cmd+Space`.

packages/ckeditor5-utils/src/keyboard.js
```diff
@@ -97,6 +102,7 @@ export function parseKeystroke( keystroke ) {
 
 	return keystroke
 		.map( key => ( typeof key == 'string' ) ? getCode( key ) : key )
+		.map( key => env.isMac && key == keyCodes.ctrl ? keyCodes.cmd : key )
```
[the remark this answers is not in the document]
niegowski: Yes. Otherwise, we would need to provide some "force" specifier or provide keystone alternative for MacOS.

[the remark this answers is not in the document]
oleq: > Does this line means you cannot assign <kbd>Ctrl+...</kbd> keystroke on a Mac?

I'm worried about existing integrations here. Changing keystrokes in our features to resolve conflicts is a breaking change but disabling keystrokes in 3rd-party integrations is a whole new level. Just to clarify: there will be no way to use <kbd>Ctrl</kbd>+<kbd>whatever</kbd> on Mac after this change?

packages/ckeditor5-utils/src/keyboard.js
```diff
@@ -108,22 +114,20 @@ export function parseKeystroke( keystroke ) {
  * @returns {String} Keystroke text specific for the environment.
  */
 export function getEnvKeystrokeText( keystroke ) {
-	if ( !env.isMac ) {
-		return keystroke;
-	}
+	let keystrokeCode = parseKeystroke( keystroke );
+
+	const modifiersToGlyphs = Object.entries( env.isMac ? modifiersToGlyphsMac : modifiersToGlyphsNonMac );
```
oleq, line 119: Let's cache it outside this helper.

packages/ckeditor5-utils/src/keyboard.js
oleq: Yeah, that's some nice programming porn 👏 But I second the request for comments because unless you run a debugger, it's magic.

packages/ckeditor5-utils/src/keyboard.js
niegowski: It won't be testable then.

packages/ckeditor5-enter/src/enterobserver.js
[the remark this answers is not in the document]
niegowski: I'm changing it, to-do list will use 'keydown' listener so it can be stopped in the context of to-do list.

packages/ckeditor5-utils/src/keyboard.js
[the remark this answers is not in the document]
niegowski: After a talk with @Reinmar we decided that we will introduce `Ctrl!` as a forced `Ctrl` key that won't get mapped to `Cmd` on macOS. This is because some integrators or external features might want to use keystrokes that are not available after mapping to `Cmd` (for example `Cmd+Q`, `Cmd+T`, `Cmd+W`, etc.) but are available as `Ctrl+Q`, `Ctrl+T`, `Ctrl+W`.

niegowski:
Ready for another review round.

packages/ckeditor5-typing/tests/utils/injectunsafekeystrokeshandling.js
```diff
@@ -17,6 +17,11 @@ describe( 'unsafe keystroke handling utils', () => {
 			expect( isNonTypingKeystroke( { keyCode: keyCodes[ 0 ], ctrlKey: true } ), 'Ctrl+0' ).to.be.true;
 		} );
 
+		it( 'should return "true" for any keystroke with the Cmd key', () => {
+			expect( isNonTypingKeystroke( { keyCode: keyCodes.a, metaKey: true } ), 'Ctrl+a' ).to.be.true;
+			expect( isNonTypingKeystroke( { keyCode: keyCodes[ 0 ], metaKey: true } ), 'Ctrl+0' ).to.be.true;
```
maxbarnas, line 22: The `'Ctrl+a'`/`'Ctrl+0'` shouldn't be `'⌘a'`/`'⌘0'`, respectively?

maxbarnas (approved):

packages/ckeditor5-typing/tests/utils/injectunsafekeystrokeshandling.js
maxbarnas, line 22:
```suggestion
			expect( isNonTypingKeystroke( { keyCode: keyCodes.a, metaKey: true } ), '⌘a' ).to.be.true;
			expect( isNonTypingKeystroke( { keyCode: keyCodes[ 0 ], metaKey: true } ), '⌘0' ).to.be.true;
```

Issue #8921, opened 2021-01-25
Text styles not kept when being converted


LangQian:
## 📝 Provide detailed reproduction steps (if any)

1. Load the following HTML into CK Editor
`<span style="font-family:'helvetica';">
    <span style="color:#0000ff;">
        <span style="font-size:16px;font-family:'arial';color:#000000;">
            aaa
        </span>
    </span>
</span>`

2. After CK Editor's processing, the above one was converted to the following one:
`<span style="color:#0000ff">aaa&nbsp;</span>`

### ✔️ Expected result

Styles should be kept after conversion.

### ❌ Actual result

Most styles were lost in the above scenario.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

LangQian:
This is an issue our customers are facing, so it is very urgent for us but we couldn't solve it on our end. If any workaround can be provided, we would really appreciate it!

Mgsy:
Hi! Could you please tell us what is your expected result? Do you expect to keep styles define in this span:

```
<span style="font-size:16px;font-family:'arial';color:#000000;">
```

... and convert it to a single span or keep the structure of nested spans?

LangQian:
@Mgsy Both would work for us, as long as the styles can be kept. In the above example, our customers were seeing black text, but it now becomes blue, which makes them vey confused (since they don't know the underlying HTML looks like).

LangQian:
Hi @Mgsy , do you need more information on this one? I see the 'pending: feedback' label is still there.

Mgsy:
Hi @LangQian, I'm sorry for the late response, we had a busy period. I'll look at your case soon and get back to you.

Mgsy:
@LangQian, thanks for bringing it to our attention. There's an issue with our converters - in the case of nested elements, styles from the deepest one should be applied, while now, those are ignored.

Note: CKEditor 4 handles such case properly and applies styles from the deepest nested element.

jacekbogdanski:
A related issue in text part language feature - https://github.com/ckeditor/ckeditor5/pull/9074#issuecomment-799512201

PR #9074, opened 2021-02-23
Text part language support

### LINKED ISSUE #8989: Add support for setting text part language

Add support for setting text part language, similar to CKEditor 4 [language feature](https://ckeditor.com/cke4/addon/language).

---

jacekbogdanski:

Feature (language): Added language dropdown button to support text part language. Closes #8989.

Other (utils): Added `language.getLanguageDirection` helper function allowing to determine text direction based on language code.

Internal (locale): Language text direction detection moved to the `language` utils module.


packages/ckeditor5-language/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@ckeditor/ckeditor5-language",
+  "version": "25.0.0",
+  "description": "Block quote feature for CKEditor 5.",
```
Reinmar, line 4: Ping ;)

packages/ckeditor5-language/LICENSE.md
```diff
@@ -0,0 +1,17 @@
+Software License Agreement
+==========================
+
+**CKEditor 5 Language Feature** – https://github.com/ckeditor/ckeditor5-language <br>
```
Reinmar, line 4:
```suggestion
**CKEditor 5 Language feature** – https://github.com/ckeditor/ckeditor5-language <br>
```

Reinmar: I know we're not consistent here, but the expected format is "Camel notcase" (however it's called).

packages/ckeditor5-language/README.md
```diff
@@ -0,0 +1,16 @@
+CKEditor 5 language feature
```
Reinmar, line 1: Let's be consistent here – "Language feature".

Also, cc @AnnaTomanek, do you think it makes sense to call it like this?

packages/ckeditor5-language/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "ckeditor",
+    "ckeditor5",
+    "ckeditor 5",
+    "ckeditor5-feature",
+    "ckeditor5-plugin"
+  ],
+  "main": "src/index.js",
+  "dependencies": {
+    "ckeditor5": "^25.0.0",
+    "@ckeditor/ckeditor5-utils": "^25.0.0"
```
Reinmar, line 15: It's not a correct dependency. You can only import things from `ckeditor5` due to DLLs. Can you run the dependency check task to see if it catches this (it should)?

packages/ckeditor5-language/src/language.js
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	static get pluginName() {
+		return 'Language';
+	}
+}
+
+/**
+ * The configuration of the language feature. Introduced by the {@link module:language/languageediting~LanguageEditing} feature.
```
Reinmar, line 42: We need to be super clear here that it's not about the editor interface language. Ideally, let's link to [https://ckeditor.com/docs/ckeditor5/latest/features/ui-language.html](https://ckeditor.com/docs/ckeditor5/latest/features/ui-language.html) in a couple most critical places.

packages/ckeditor5-language/src/utils.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module language/utils
+ */
+
+import { getLanguageDirection } from '@ckeditor/ckeditor5-utils/src/language';
```
Reinmar, line 10: This import is incorrect and I wonder why it wasn't caught by the linter. It should go through `ckeditor5/...`.

packages/ckeditor5-utils/src/language.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module utils/language
+ */
+
+const RTL_LANGUAGE_CODES = [
```
Reinmar, line 10: Are the 3-letter versions in the standard? In the UI we support the 2-letter ones only, AFAIK.

packages/ckeditor5-utils/tests/language.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import { getLanguageDirection } from '../src/language';
+
+describe( 'language', () => {
+	describe( 'getLanguageDirection', () => {
```
Reinmar, line 9:
```suggestion
	describe( 'getLanguageDirection()', () => {
```

packages/ckeditor5-utils/tests/language.js
```diff
@@ -0,0 +1,32 @@
...
+		it( 'determines the language direction', () => {
```
Reinmar, line 10: Avoid one huge tests. Better to have multiple smaller ones so you know which fails.

packages/ckeditor5-utils/tests/language.js
```diff
@@ -0,0 +1,32 @@
...
+			expect( getLanguageDirection( 'en' ) ).to.eql( 'ltr' );
```
Reinmar, line 11:
```suggestion
			expect( getLanguageDirection( 'en' ) ).to.equal( 'ltr' );
```

No idea what's the difference, but we use the longer form everywhere else.

packages/ckeditor5-language/src/language.js
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+/**
+ * The available language options.
+ *
+ * The default value is:
+ *
+ *		const languageConfig = {
+ *			options: [
+ *				{ title: 'Arabic', class: 'ck-language_ar', languageCode: 'ar' },
```
Reinmar, line 72: Where is that class used? I'm not sure the `ck-` prefix makes sense, but it depends on the context.

packages/ckeditor5-language/src/languageediting.js
```diff
@@ -0,0 +1,100 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module block-quote/languageediting
```
maxbarnas: Is this `block-quote` correct here?

packages/ckeditor5-language/src/language.js
Reinmar: I should've read further:

```
 * The `title` and `class` properties will be used by the `language` dropdown to render available options.
```

I don't think anyone will be styling these options in the dropdown so I don't think we need to have specific classes for them. If anything, we can have these classes added automatically.

In the heading feature, the classes are configurable as you're supposed to style the headings in the dropdown to look like headings in the content. And for that you need these classes.

packages/ckeditor5-language/src/language.js
```diff
@@ -0,0 +1,100 @@
...
+ *				{ title: 'French', class: 'ck-language_fr', languageCode: 'fr' },
+ *				{ title: 'Spanish', class: 'ck-language_es', languageCode: 'es' }
+ *			]
+ *		};
+ *
+ * The `title` and `class` properties will be used by the `language` dropdown to render available options.
+ *
+ * The `languageCode` property is used for the lang attribute in ISO 639 format. Language codes can be found
```
Reinmar, line 80: As I mentioned in the other place – so far I think we've been supporting two-letter versions. But if CKE4 allowed using 3-letter ones, then it makes sense to have them covered this way here.

packages/ckeditor5-language/src/language.js
```diff
@@ -0,0 +1,100 @@
...
+ * @module language/language
+ */
+
+import { Plugin } from 'ckeditor5/src/core';
+
+import LanguageEditing from './languageediting';
+import LanguageUI from './languageui';
+
+/**
+ * The language plugin.
```
Reinmar, line 16: This is also a good place to clarify that this is not about UI language but an ability to set the language of parts of the content.

packages/ckeditor5-language/src/language.js
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * See {@link module:core/editor/editorconfig~EditorConfig all editor options}.
+ *
+ * @interface LanguageConfig
+ */
+
+/**
+ * The available language options.
+ *
+ * The default value is:
```
Reinmar, line 68: Is the default from CKE4? It's quite useless, but I suppose it's on purpose – it ensures that the feature works by default, but requires to be reconfigured when integrated for real.

Thus, we don't need to have translations for these language titles. Note: in many other features such strings are covered by language files and we have automatic translations for them. But here, it makes no sense.

This is something that may not be obvious to others (that we don't translate them on purpose), so it'd be good to have a comment about that in the code (next to the default value?). It may be hard to spot if you don't know that you're looking for it, but it's better than nothing.

packages/ckeditor5-language/README.md
AnnaTomanek: "Language" looks fine (as it's consistent with CKE4). Another option would be "text language" as it then would be clear that it's not about the UI language. In WCAG the success criterion is named "Language of Parts", the `lang` attribute itself defines the language of text/element. To make it short, we can go with "language" as the feature name and then use other keywords (language of text, language of content, language of text parts) in the docs.

packages/ckeditor5-language/src/utils.js
```diff
@@ -0,0 +1,64 @@
...
+ * when the user configuration is defined because the editor does not exist yet.
+ *
+ * @param {module:core/editor/editor~Editor} editor
+ * @returns {Array.<module:language/language~LanguageOption>}.
+ */
+export function getLocalizedOptions( editor ) {
+	const t = editor.t;
+
+	return editor.config.get( 'languageList.options' ).map( option => {
+		option.title = t( option.title );
```
niegowski, line 61: I'm not sure if this would work as expected.

packages/ckeditor5-language/src/languagecommand.js
```diff
@@ -0,0 +1,138 @@
...
+ * The language command plugin.
+ *
+ * @extends module:core/command~Command
+ */
+export default class LanguageCommand extends Command {
+	/**
+	 * @param {module:core/editor/editor~Editor} editor
+	 * @param {String} attributeKey Attribute that will be set by the command.
+	 */
+	constructor( editor, attributeKey ) {
```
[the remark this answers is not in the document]
niegowski: My comment depends on the decision from this comment: https://github.com/ckeditor/ckeditor5/pull/9074#discussion_r581257890

packages/ckeditor5-language/src/utils.js
pomek: It will not. `t()` requires a string in order to replace it with a correct translation.

packages/ckeditor5-language/src/languagecommand.js
```diff
@@ -0,0 +1,138 @@
...
+	 * {@link module:engine/model/document~Document#selection} itself (note that typed characters copy attributes from the selection).
+	 * * If the selection is collapsed in an empty node, the command applies the attribute to the parent node of the selection (note
+	 * that the selection inherits all attributes from a node if it is in an empty node).
+	 *
+	 * @fires execute
+	 * @param {Object} [options] Command options.
+	 * @param {String|Boolean} [options.languageCode] Language code to be applied to the model.
+	 * @param {String} [options.textDirection] Language text direction.
+	 */
+	execute( { languageCode, textDirection } = {} ) {
```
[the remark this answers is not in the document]
niegowski: WDYT @Reinmar

packages/ckeditor5-language/README.md
Reinmar: I'm for "text language" and renaming this plugin. I'd prefer to avoid confusion as much as we can.

jacekbogdanski: `text language` would also improve options config. For now, I had to go with `languageList` to avoid duplication with `language`. Not sure if you follow plugin name == config name semantic in the whole codebase, but in most cases, I saw that they are the same.

packages/ckeditor5-language/src/languagecommand.js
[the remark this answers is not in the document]
jacekbogdanski: Maybe just expose `_getValueFromFirstAllowedNode()` and a part of `execute` logic as utils helpers, used in both language and `AttributeCommand`?

packages/ckeditor5-utils/src/language.js
jacekbogdanski: 3-letter version is supported by CKE4.

packages/ckeditor5-language/src/language.js
jacekbogdanski: > Is the default from CKE4? It's quite useless, but I suppose it's on purpose – it ensures that the feature works by default, but requires to be reconfigured when integrated for real.

Yes, and yes :)  that's the purpose of this default configuration in CKE4.

packages/ckeditor5-language/src/languagecommand.js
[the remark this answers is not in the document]
niegowski: Or maybe we could provide those helpers as the model methods similar to `insertContent` and `deleteContent`?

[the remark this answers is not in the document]
niegowski: Also, we have `findAttributeRange` exported from typing package and it's used in some places, maybe typing should provide more such helpers?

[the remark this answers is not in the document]
Reinmar: TBH, IDK what `findAttributeRange()` does in the typing package as it's not very related... I'm afraid whatever's needed in the LanguageCommand doesn't make much sense there too.

Do you have an idea for a better place?

Also, a little copy-paste here is not a big deal. So, if we can reduce the current ~100% duplication by creating 1-2 reasonable utils and putting them in the core or engine packages, then I'd go for that.

packages/ckeditor5-language/src/utils.js
Reinmar: Also, I explained it in another comment that perhaps we don't need this translated: [https://github.com/ckeditor/ckeditor5/pull/9074#discussion_r580850365](https://github.com/ckeditor/ckeditor5/pull/9074#discussion_r580850365.).

packages/ckeditor5-language/README.md
Reinmar: I didn't know we actually used `config.language` already. So thats a must to rename this feature. The name of the property should match the name of the plugin.

packages/ckeditor5-language/src/languagecommand.js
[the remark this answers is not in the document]
niegowski: I think that for now we could go with the copy-paste, but someday we will refactor it to extract some common code fragments to the model and selection methods: #9107

packages/ckeditor5-language/package.json
jacekbogdanski: Do you mean `yarn run check-dependencies` script? If yes, it's not returning any error 🤔 

image

jacekbogdanski: On the other hand, when I changed it into `ckeditor5/src/utils` linter says it's incorrect and asks for `@ckeditor/ckeditor5-utils`: 
image

jacekbogdanski: We've talked about it with @niegowski and he explained to me that the import in `language/src/utils.js` file was invalid in the first place, hence this linter message.

packages/ckeditor5-core/src/editor/editorconfig.jsdoc
```diff
@@ -231,7 +250,23 @@
...
+ *
+ * @member {String} module:core/editor/editorconfig~LanguageConfig#ui
+ */
+
+/**
+ * Allows to use different language of the editor content.
+ *
+ * The language codes are defined in the [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) standard.
+ *
+ * @member {String} module:core/editor/editorconfig~LanguageConfig#content
```
jacekbogdanski, line 269: `textFragmentLanguage` is also supporting ISO 639-2 (3 letters) codes. I'm wondering if we should also extend `language.content` and `language.ui` with ISO 639-2 or rather remove this information from `textFragmentLanguage` and treat ISO 639-2 only as deprecated, compatibility support? I have a feeling that it would be nicer to limit supported codes only to ISO 639-1 as they are more common.

packages/ckeditor5-language/src/textfragmentlanguage.js
```diff
@@ -0,0 +1,88 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	static get pluginName() {
+		return 'TextFragmentLanguage';
+	}
+}
+
+/**
+ * The available {@link module:language/textfragmentlanguage} options allowing setting language of parts of the content.
```
jacekbogdanski, line 51: I went with `language.textFragmentLanguage` option name to match the plugin name, but it would be probably nicer to simply name it `language.textFragment` to avoid duplication. WDYT?

jacekbogdanski (commented):
I didn't cover API docs at `packages/language/docs` yet to avoid some redundant changes if we decide to update code which could be reflected in the documentation.

packages/ckeditor5-language/src/textfragmentlanguagecommand.js
```diff
@@ -0,0 +1,125 @@
...
+			}
+		} );
+	}
+
+	/**
+	 * Returns the attribute value of the first node in the selection that allows the attribute.
+	 * For the collapsed selection returns the selection attribute.
+	 *
+	 * @private
+	 * @returns {Object} The attribute value.
```
niegowski, line 104: Isn't it returning string or false?

packages/ckeditor5-language/src/textfragmentlanguageediting.js
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * @module language/textfragmentlanguageediting
+ */
+
+import { Plugin } from 'ckeditor5/src/core';
+import TextFragmentLanguageCommand from './textfragmentlanguagecommand';
+import { stringifyLanguageAttribute, parseLanguageAttribute } from './utils';
+
+const ATTRIBUTE_KEY = 'language';
```
niegowski, line 14: This is the same as in the command, maybe we should export it so it could be imported in the command?

packages/ckeditor5-language/src/textfragmentlanguageediting.js
```diff
@@ -0,0 +1,103 @@
...
+				value: viewElement => {
+					const languageCode = viewElement.getAttribute( 'lang' );
+					const textDirection = viewElement.getAttribute( 'dir' );
+
+					return stringifyLanguageAttribute( languageCode, textDirection );
+				}
+			},
+			view: {
+				name: 'span',
+				attributes: { lang: /[^]/ }
```
niegowski, line 83: This RegExp is matching one character, maybe it should be `/[^]+/`?

packages/ckeditor5-language/src/textfragmentlanguageediting.js
```diff
@@ -0,0 +1,103 @@
...
+				if ( !attributeValue ) {
+					return;
+				}
+
+				const { languageCode, textDirection } = parseLanguageAttribute( attributeValue );
+
+				return writer.createAttributeElement( 'span', {
+					lang: languageCode,
+					dir: textDirection
+				} );
```
niegowski, line 99: Maybe this AttributeElement should have a lower priority than a link so a link would be wrapped by language span (with default priority it is inside a link and it's split into multiple elements if there is a linkHref inside). I'm not sure which behavior is better in this case. WDYT @Reinmar

packages/ckeditor5-language/src/textfragmentlanguageui.js
```diff
@@ -0,0 +1,119 @@
...
+import { Model, createDropdown, addListToDropdown } from 'ckeditor5/src/ui';
+import { Collection } from 'ckeditor5/src/utils';
+import { stringifyLanguageAttribute } from './utils';
+
+import '../theme/language.css';
+
+/**
+ * The text fragment language UI plugin.
+ *
+ * It introduces the `'language'` button.
```
niegowski, line 20: It's introducing a dropdown, not a button.

packages/ckeditor5-language/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+ * * `languageCode` - The language code used for the lang attribute in ISO 639 format.
+ * * `textDirection` - One of the following values: `rtl` or `ltr`, indicating the reading direction of the language.
+ *
+ * See {@link module:core/editor/editorconfig~LanguageConfig#textFragmentLanguage text fragment config}
+ * for more information about language properties.
+ *
+ * If `textDirection` argument is omitted, it will be automatically detected based on `languageCode`.
+ *
+ * @param {String} languageCode The language code in ISO 639 format.
+ * @param {String} [textDirection] Language text direction. Automatically detected if omitted.
```
niegowski, line 26:
```suggestion
 * @param {'ltr'|'rtl'} [textDirection] Language text direction. Automatically detected if omitted.
```

packages/ckeditor5-language/src/textfragmentlanguage.js
```diff
@@ -0,0 +1,88 @@
...
+ * module:core/editor/editorconfig~LanguageConfig#textFragmentLanguage
+ */
+
+/**
+ * Text fragment language feature option descriptor.
+ *
+ * @typedef {Object} module:language/textfragmentlanguage~TextFragmentLanguageOption
+ * @property {String} title The user-readable title of the option.
+ * @property {String} languageCode The language code in ISO 639 format.
+ * @property {String} [textDirection] The language text direction. Automatically detected if omitted.
```
niegowski, line 87:
```suggestion
 * @property {'ltr'|'rtl'} [textDirection] The language text direction. Automatically detected if omitted.
```

packages/ckeditor5-language/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+ * {@link module:language/utils~stringifyLanguageAttribute stringifyLanguageAttribute} function.
+ *
+ * @param {String} str Attribute value.
+ * @returns {Object} result
+ * @returns {String} result.languageCode The language code in ISO 639 format.
+ * @returns {String} result.textDirection Language text direction.
+ */
+export function parseLanguageAttribute( str ) {
+	const parts = str.split( ':' );
+	return { languageCode: parts[ 0 ], textDirection: parts[ 1 ] };
```
niegowski, line 45:
```suggestion
	const [ languageCode, textDirection ] = str.split( ':' );
	return { languageCode, textDirection };
```

packages/ckeditor5-utils/src/language.js
```diff
@@ -0,0 +1,26 @@
...
+	'he', 'heb', // Hebrew
+	'ku', 'kur', // Kurdish
+	'ug', 'uig' // Uighur, Uyghur
+];
+
+/**
+ * Helps determine whether a language text direction is LTR or RTL.
+ *
+ * @param {String} language The ISO 639 language code.
+ * @returns {String} 'ltr' or 'rtl
```
niegowski, line 22:
```suggestion
 * @returns {'ltr'|'rtl'}
```

packages/ckeditor5-language/src/index.js
```diff
@@ -0,0 +1,18 @@
...
+
+import TextFragmentLanguage from './textfragmentlanguage';
+import TextFragmentLanguageEditing from './textfragmentlanguageediting';
+import TextFragmentLanguageUI from './textfragmentlanguageui';
+
+export default {
+	TextFragmentLanguage,
+	TextFragmentLanguageEditing,
+	TextFragmentLanguageUI
+};
```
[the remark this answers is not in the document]
Reinmar: See #9134. I raised the same question recently. We need to better understand that, but as @psmyrek explains there's some reason for that. Here, follow a feature that's similar in kind - e.g. ckeditor5-image, ckeditor5-highlight.

packages/ckeditor5-language/src/textfragmentlanguageediting.js
Reinmar: IMO, link has a priority – it should be split as rarely as possible. In some sense, I see links as "elements" (inline widgets) and most other elements as text formatting. Language feature defines a property of a piece of content. It has no identity itself – unlike a link that even from a UX perspective behave as a one thing – it reacts to hover.

packages/ckeditor5-language/src/index.js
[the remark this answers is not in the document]
niegowski: :+1:

packages/ckeditor5-language/src/textfragmentlanguageediting.js
niegowski: Ok, let's leave it at the default priority then.

packages/ckeditor5-language/tests/textfragmentlanguageediting.js
```diff
@@ -0,0 +1,128 @@
...
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
+import TextFragmentLanguageEditing from '../src/textfragmentlanguageediting';
+import TextFragmentLanguageCommand from '../src/textfragmentlanguagecommand';
+import VirtualTestEditor from '@ckeditor/ckeditor5-core/tests/_utils/virtualtesteditor';
+import { getData as getModelData, setData as setModelData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
+import { getData as getViewData } from '@ckeditor/ckeditor5-engine/src/dev-utils/view';
```
niegowski, line 11: Please don't mix relative imports with package imports.

packages/ckeditor5-language/tests/textfragmentlanguageediting.js
```diff
@@ -0,0 +1,128 @@
...
+	it( 'should be loaded', () => {
+		expect( editor.plugins.get( TextFragmentLanguageEditing ) ).to.be.instanceOf( TextFragmentLanguageEditing );
+	} );
+
+	it( 'should set proper schema rules', () => {
+		expect( model.schema.checkAttribute( [ '$root', '$block', '$text' ], 'language' ) ).to.be.true;
+		expect( model.schema.checkAttribute( [ '$clipboardHolder', '$text' ], 'language' ) ).to.be.true;
+	} );
+
+	it( 'its attribute is marked with a copOnEnter property', () => {
```
niegowski, line 40:
```suggestion
	it( 'its attribute is marked with a copyOnEnter property', () => {
```

packages/ckeditor5-language/tests/textfragmentlanguageediting.js
```diff
@@ -0,0 +1,128 @@
...
+describe( 'TextFragmentLanguageEditing', () => {
+	let editor, model;
+
+	beforeEach( () => {
+		return VirtualTestEditor
+			.create( {
+				plugins: [ TextFragmentLanguageEditing, Paragraph ]
+			} )
+			.then( newEditor => {
+				editor = newEditor;
```
niegowski, line 22: This editor instance is not destroyed.

packages/ckeditor5-language/tests/textfragmentlanguageediting.js
```diff
@@ -0,0 +1,128 @@
...
+				.create( {
+					plugins: [ TextFragmentLanguageEditing, Paragraph ],
+					language: languageConfig
+				} )
+				.then( newEditor => {
+					editor = newEditor;
+					model = editor.model;
+				} );
+		}
+	} );
```
niegowski, line 126: This would override the editor created before and both the editors would not be destroyed properly.

packages/ckeditor5-language/tests/textfragmentlanguageui.js
```diff
@@ -0,0 +1,144 @@
...
+
+/* globals document */
+
+import ClassicTestEditor from '@ckeditor/ckeditor5-core/tests/_utils/classictesteditor';
+import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
+import TextFragmentLanguageEditing from '../src/textfragmentlanguageediting';
+import TextFragmentLanguageUI from '../src/textfragmentlanguageui';
+import DropdownView from '@ckeditor/ckeditor5-ui/src/dropdown/dropdownview';
+import testUtils from '@ckeditor/ckeditor5-core/tests/_utils/utils';
+import { setData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
```
niegowski, line 14: Please don't mix imports.

packages/ckeditor5-language/src/textfragmentlanguageediting.js
jacekbogdanski: Actually, I would be for changing it to `[\s\S]+`. I know that `/[^]+/` is more concise, but looking into CKE5 codebase, it's not really used anywhere in contrast to the first notation, so it should be more readable.

niegowski: :+1:

packages/ckeditor5-language/src/textfragmentlanguageediting.js
jacekbogdanski: As we talked, I've replaced it with plain string as in most cases we are not using const for attributes.

Reinmar:
CI fails.

packages/ckeditor5-core/src/editor/editorconfig.jsdoc
jacekbogdanski: I've simplified it to ISO 639-1. It's still possible to use other ISO formats since it's just a string that won't conflict with the current implementation, but using a single format makes our life easier.

packages/ckeditor5-language/src/textfragmentlanguage.js
jacekbogdanski: After a second thought, it probably makes a bit more sense to keep a longer format, so the plugin name. Otherwise, it will be hard to tell what feature is actually introducing this option which may affect discoverability.

packages/ckeditor5-language/src/textfragmentlanguage.js
```diff
@@ -0,0 +1,70 @@
...
+
+import { Plugin } from 'ckeditor5/src/core';
+
+import TextFragmentLanguageEditing from './textfragmentlanguageediting';
+import TextFragmentLanguageUI from './textfragmentlanguageui';
+
+/**
+ * @extends module:core/plugin~Plugin
+ */
+export default class TextFragmentLanguage extends Plugin {
```
niegowski, line 18: Missing description.

packages/ckeditor5-language/src/textfragmentlanguage.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * The `languageCode` property is used for the lang attribute in [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) format.
+ *
+ * You can also specify optional `textDirection` property indicating the reading direction of the language.
+ * Correct values are `ltr` and `rtl`. When `textDirection` property is missing, the text fragment language feature will
+ * specify text direction by itself.
+ *
+ * @member {Array.<module:language/textfragmentlanguage~TextFragmentLanguageOption>}
+ * module:core/editor/editorconfig~LanguageConfig#textFragmentLanguage
+ */
```
niegowski, line 61: We should note here that this option is not available by default and the TextFragmentLanguage plugin must be added.

packages/ckeditor5-language/tests/textfragmentlanguagecommand.js
```diff
@@ -0,0 +1,385 @@
...
+
+			expect( command.value ).to.equal( 'ar:rtl' );
+			expect( doc.selection.getAttribute( 'language' ) ).to.equal( 'ar:rtl' );
+		} );
+
+		describe( 'model change event', () => {
+			let spy;
+
+			beforeEach( () => {
+				spy = sinon.spy();
```
niegowski, line 317: We should init sinon sandbox before using it (`testUtils.createSinonSandbox();`)

packages/ckeditor5-language/tests/textfragmentlanguageui.js
```diff
@@ -0,0 +1,145 @@
...
+				] );
+			} );
+		} );
+	} );
+} );
+
+function getListViewItems( listView ) {
+	// Let's drop separator.
+	return listView.items.filter( item => item.children );
+}
```
niegowski, line 145: We should define this helper inside a describe block.

niegowski (commented):
Looks good, only some very minor issues as commented.

Mgsy:
The PR is ready for testing, @FilipTokarski @LukaszGudel, could you please take a look at it?

jacekbogdanski:
I've renamed the feature to "text part language". It shouldn't break anything (especially that tests passed).

FilipTokarski:
There's a little mismatch between content from CKE4 and CKE5. If you create a "nested" text, for example like `Hebrew-French-Hebrew`, CKE4 creates this markup:
```html
<p><span dir="rtl" lang="he">hebrew <span dir="ltr" lang="fr">french</span> hebrew again</span></p>
```
CKE5 creates three separate spans:
```html
<p><span lang="he" dir="rtl">hebrew </span><span lang="fr" dir="ltr">spanish</span><span lang="he" dir="rtl"> hebrew again</span></p>
```
If you now want to get content from 4 and put it inside 5, the inner span will be removed and the whole text remains in Hebrew:
```html
<p><span lang="he" dir="rtl">hebrew french hebrew again</span></p>
```

LukaszGudel:
After testing this PR it's looking good to me. 

Other than the issue reported by Filip I could not find anything buggy related to those changes. BTW, could it be also related to #8921?

There are few issues related to RTL languages or mixing LTR with RTL in one paragraph but same issues can be reproduced in the browser without CKE5, so I guess those are not related to the editor itself.

jacekbogdanski:
@FilipTokarski as @LukaszGudel already wrote the issue with this HTML is caused by https://github.com/ckeditor/ckeditor5/issues/8921.

I've already checked it (using some dirty fix) and it should be fixed with #8921. Basically, element nodes in https://github.com/ckeditor/ckeditor5/blob/d63ec7c3b8ddeecf3a0db9abc41d5ea5830d9c55/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L894-L904 are converted recursively, so every next child for the same range will replace `lang` attribute as there is no check if the attribute has already been set.

FilipTokarski:
I didn't find any other issues, looks ok 👍

niegowski (approved):
LGTM 👍

AnnaTomanek:
Just for future reference, the main changelog entry should not be about adding the dropdown for the feature but about adding the feature itself. Now it sounds like the feature has already existed and we've just added a dropdown that was missing.

Issue #8778, opened 2021-01-08
Invalid link to image.png in the Link docs


psmyrek:
## 📝 Provide a description of requested docs changes

1. Open https://ckeditor.com/docs/ckeditor5/latest/features/link.html#demo-2.
2. Click the `downloadable resource` link in the editor and then the `image.png` link in the baloon.

Error: the 404 page is opened.

pomek:
Fixed with https://github.com/ckeditor/ckeditor5/pull/9122.

PR #9122, opened 2021-02-26
Docs: restoring a missing resource.

godai78:
Closes: #8778

A file that never was, has been uploaded to be. Hence a 404 error is now swiftly avoided.


psmyrek (approved):
Link is working now 🎉

pomek (approved):
LGTM.

Issue #9146, opened 2021-03-02
Add Essentials to the base DLL build


Reinmar:
## 📝 Provide a description of the improvement

So you don't need to list all the essential plugins when instantiating the editor.

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

Reinmar:
Currently, it's a bit odd that Paragraph is part of it, but Essentials not (while all its deps are).

Reinmar:
After we do this, we can simplify the documentation for DLLs a bit (use Essentials instead of Undo,SelectAll, etc. separately).

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity over the last year. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #9147, opened 2021-03-02
Rewrote the DLL builds guide

### LINKED ISSUE #8095: Add a note about jsdoc annotations order that we follow in the project

I could not find in our docs an article why in v5 we’re using jsdoc annotations in a specific order. I mean the following structure:

```
@protected
@params {...} ...
@returns ...
```

Also, would be good to attach more annotations that can describe the function: `@see`, `@fires`, `@chainable`, `@observable`, `@readonly`, `@private`, and `@public`.

As @Reinmar said:

> This is a protected method, introduced in v22.0.0 that takes these params and returns that. It may also fire events like this

I am sure we should write it down anywhere because from time to time we may change the order and would be good to not explain it each time.

### LINKED ISSUE #8905: Update the documentation with info about DLLs

We have a new guide already, but it needs to be reviewed and updated.

Also, we need to think on what DLL-related things should be mentioned and where. Do we change imports? If yes, we need to review tones of docs (especially the framework guides). If not, we need to review whether we didn't change something already and revert that.

### LINKED ISSUE #9094: Review and rewriting the DLL guide

[_http://fake.ckeditor.com:8081/build/docs/ckeditor5/latest/builds/guides/development/dll-builds.html_](http://fake.ckeditor.com:8081/build/docs/ckeditor5/latest/builds/guides/development/dll-builds.html)

Reinmar:

Docs: Rewrote the DLL builds guide. Closes #9094. Closes #8905.

---

### Additional information

*   I didn't read the content once after writing it as I need to leave today. **I didn't even build this guide**.
*   It needs a followup to re-add the missing parts (implementing DLL-compat plugins and similar). I stripped it down now as it's not essential yet.
*   I reported #9146 as I think it's odd it's not there.
*   There's also a PR in cke5-internal.

### DIFF
--- a/docs/builds/guides/development/dll-builds.md
+++ b/docs/builds/guides/development/dll-builds.md
@@ -8,20 +8,20 @@ order: 30
 
 The purpose of a DLL build is to allow adding plugins to an editor build without having to rebuild (recompile) the build itself.
 
-SO far, the two most common integration methods include:
+So far, the two most common integration methods included:
 
 * Using pre-compiled builds. This can be either one of the official builds or a custom build. In this case, adding a plugin requires recompiling the entire build.
 * Integrating the editor from source. In this case, if you want to add a plugin, your application needs to be recompiled.
 
 In some advanced use cases, the list of available plugins cannot be limited &mdash; it should be possible to add plugins without any access to Node.js. In other words, plugins should be built (compiled) separately from the editor's core.
 
-That is where DLL builds come to the rescue.
+This is where the DLL builds come to the rescue.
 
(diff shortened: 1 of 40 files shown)


godai78:
Made a review, haven't built it yet - system-wide problems with yarning at the moment. Will return to the issue once it's fixed.

pomek:
CI fails because of the Chrome update. Tests were passing on Chrome 88.0.4324.192, and started to fail on 89.0.4389.72.

pomek (approved):
LGTM.

PR #9043, opened 2021-02-17
Integrated inline images with responsive images (srcset, sizes)

pkwasnik:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Internal (image): Integrated inline images with responsive images. Closes #8777.

---

### Additional information

_For example – encountered issues, assumptions you had to make, other affected tickets, etc._

PR #9173, opened 2021-03-05
The UX of resizing small images.

### LINKED ISSUE #9162: Image caption placeholder wraps for small images

### LINKED ISSUE #9166: The UX of resizing small images looks broken
It already looks bad for block images (which have 50px `min-height` for `` BTW)

and I can imagine it will look even worse for small inline images (e.g. 16x16px) that will also get resizing after #8777. The resize UI might cover the entire inline image and we should do something about it. For instance, we could display only a single resize handle for very small images. This will require some additional research.

Related to #9165.

magda-chrzescian:

Fix (image): An image should never overflow the widget boundaries while changing its size. Closes #9166.

Fix (image): The size label should be displayed above the image if it doesn't fit inside. See #9166.

Fix (image): An Image caption placeholder text should not wrap. Closes #9162.

---

AFAIR the `SizeView` class was not tested directly at all. I exported it (it is declared as public in the API docs) and wrote the tests covering only the purpose of this issue.

---

**Requires** #9043.


packages/ckeditor5-widget/src/widgetresize/sizeview.js
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+import View from '@ckeditor/ckeditor5-ui/src/view';
+
+/**
+ * A view displaying the proposed new element size during the resizing.
+ *
+ * @extends {module:ui/view~View}
+ */
+
```
oleq, line 17: Extra new line.

packages/ckeditor5-widget/src/widgetresize/sizeview.js
```diff
@@ -0,0 +1,80 @@
...
+	constructor() {
+		/**
+		 * The position of the view defined based on the host size and active handle position.
+		 *
+		 * @private
+		 * @observable
+		 * @member {String} #viewPosition
+		 */
+
+		super();
```
oleq, line 28: `super()` should precede everything else.

packages/ckeditor5-widget/src/widgetresize/sizeview.js
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @module widget/widgetresize/sizeview
+ */
+
+import View from '@ckeditor/ckeditor5-ui/src/view';
+
+/**
+ * A view displaying the proposed new element size during the resizing.
+ *
+ * @extends {module:ui/view~View}
```
oleq, line 15: I'd say there's no point in making this class public. Let's keep it protected.

packages/ckeditor5-widget/src/widgetresize/sizeview.js
```diff
@@ -0,0 +1,80 @@
...
+					display: bind.if( 'isVisible', 'none', visible => !visible )
+				}
+			},
+			children: [ {
+				text: bind.to( 'label' )
+			} ]
+		} );
+	}
+
+	bindToState( options, resizerState ) {
```
oleq, line 50: Missing method docs.

packages/ckeditor5-widget/src/widgetresize/sizeview.js
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @extends {module:ui/view~View}
+ */
+
+export default class SizeView extends View {
+	constructor() {
+		/**
+		 * The position of the view defined based on the host size and active handle position.
+		 *
+		 * @private
```
oleq, line 23: [This should be protected when tested.](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/code-style.html#accessibility)

packages/ckeditor5-widget/src/widgetresize/sizeview.js
```diff
@@ -0,0 +1,80 @@
...
+		 * @observable
+		 * @member {String} #viewPosition
```
oleq, line 25: [If protected then, `_viewPosition`.](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/code-style.html#visibility-levels)

packages/ckeditor5-widget/src/widgetresize/sizeview.js
oleq: Also `@readonly`, no one will ever write to this property.

packages/ckeditor5-widget/tests/widgetresize/sizeview.js
```diff
@@ -0,0 +1,56 @@
...
+
+			state.activeHandlePosition = position;
+			state.update( {
+				handleHostWidth: 200,
+				handleHostHeight: 200
+			} );
+
+			expect( sizeView.viewPosition ).to.equal( position );
+			expect( sizeView.element.classList.contains( `ck-orientation-${ position }` ) ).to.be.true;
+		} );
```
oleq, line 54: Asserting `viewPosition` makes little sense here IMO because it is an internal component state for binding purposes only. The only thing that matters for this component is the class.

packages/ckeditor5-widget/tests/widgetresize/sizeview.js
```diff
@@ -0,0 +1,56 @@
...
+
+		beforeEach( async () => {
+			state = new ResizerState();
+			sizeView = new SizeView();
+
+			sizeView.bindToState( {}, state );
+			sizeView.render();
+		} );
+
+		it( 'should have valid `viewPosition` property and class if the widget width is smaller than 50px', () => {
```
oleq, line 21:
```suggestion
		it( 'should have valid `viewPosition` property and class if the widget width is less than 50px', () => {
```

packages/ckeditor5-widget/tests/widgetresize/sizeview.js
```diff
@@ -0,0 +1,56 @@
...
+			state.update( {
+				handleHostWidth: 49,
+				handleHostHeight: 200
+			} );
+
+			expect( sizeView.viewPosition ).to.equal( 'above-center' );
+			expect( sizeView.element.classList.contains( 'ck-orientation-above-center' ) ).to.be.true;
+		} );
+
+		it( 'should have valid `viewPosition` property and class if the widget height is smaller than 50px', () => {
```
oleq, line 32:
```suggestion
		it( 'should have valid `viewPosition` property and class if the widget height is less than 50px', () => {
```

packages/ckeditor5-widget/tests/widgetresize/sizeview.js
```diff
@@ -0,0 +1,56 @@
...
+			state.update( {
+				handleHostWidth: 200,
+				handleHostHeight: 49
+			} );
+
+			expect( sizeView.viewPosition ).to.equal( 'above-center' );
+			expect( sizeView.element.classList.contains( 'ck-orientation-above-center' ) ).to.be.true;
+		} );
+
+		it( 'should have valid `viewPosition` property and class if the widget width is bigger than 50px/50px', () => {
```
oleq, line 43:
```suggestion
		it( 'should have valid `viewPosition` property and class if the widget dimensions are greater than 50px/50px', () => {
```

packages/ckeditor5-widget/src/widgetresize/sizeview.js
```diff
@@ -0,0 +1,80 @@
...
+		this.setTemplate( {
+			tag: 'div',
+			attributes: {
+				class: [
+					'ck',
+					'ck-size-view',
+					bind.to( 'viewPosition', value => value ? `ck-orientation-${ value }` : '' )
+				],
+				style: {
+					display: bind.if( 'isVisible', 'none', visible => !visible )
```
oleq, line 41: Missing docs for #isVisible. Also, `#_isVisible` should also be protected.

packages/ckeditor5-widget/src/widgetresize/sizeview.js
```diff
@@ -0,0 +1,80 @@
...
+				}
+			},
+			children: [ {
+				text: bind.to( 'label' )
```
oleq, line 45: Missing `#label` docs. This also looks like an internal state for template binding purposes, so let's make it @Protected and `#_label`.

packages/ckeditor5-widget/tests/widgetresize/sizeview.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import SizeView from '../../src/widgetresize/sizeview';
+import ResizerState from '../../src/widgetresize/resizerstate';
+
+describe( 'SizeView', () => {
```
oleq, line 9: If we're testing this view properly now, then some essential tests are missing:

*   template structure assertion,
*   `style.display` tests,
*   children text (label) tests.

packages/ckeditor5-image/theme/image.css
```diff
@@ -61,6 +61,9 @@
 	& .image > figcaption.ck-placeholder::before {
 		padding-left: inherit;
 		padding-right: inherit;
+		white-space: nowrap;
+		overflow: hidden;
+		text-overflow: ellipsis;
```
oleq, line 66: Can you leave a comment as to why we need this with a link to the issue? These things are easily lost and CSS does not come with tests that would explain the code.

oleq (changes requested):
Functionally, things look fine 👍 I got are some concerns regarding the implementation and tests, though.

packages/ckeditor5-widget/src/widgetresize/sizeview.js
magda-chrzescian: It is not used in tests or outside the class actually, so I made it @Private.

packages/ckeditor5-widget/src/widgetresize/sizeview.js
magda-chrzescian: It is not used in tests or outside the class actually, so I made it @Private.

packages/ckeditor5-widget/src/widgetresize/sizeview.js
magda-chrzescian: It is not used in tests anymore (), so I made it @Private.

> Asserting viewPosition makes little sense here IMO because it is an internal component state for binding purposes only. The only thing that matters for this component is the class.

PR #9186, opened 2021-03-09
Overlapping inline images.

### LINKED ISSUE #9108: Inline images overlap when inserted next to each other (styles)

magda-chrzescian:

Fix (image): The selected inline image should not be overlapped by other images next to it. Closes #9108.


packages/ckeditor5-image/theme/image.css
```diff
@@ -34,6 +34,10 @@
 			display: block;
 			max-width: 100%;
 		}
+
+		&.ck-widget_selected {
```
oleq, line 38: It's not a content style but an editing style. Integrators don't need it in their [content (front-end) styles](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/content-styles.html
). It should go under `.ck.ck-editor__editable .image-inline` similar to placeholder styles blow :point\_down:.

oleq (changes requested):

packages/ckeditor5-image/theme/image.css
magda-chrzescian: :sweat_smile:

PR #9189, opened 2021-03-09
Decreased the size of output log from web crawler

### LINKED ISSUE #9062: Improve web crawler output log

Web crawler currently logs a lot of information when it finds an error on tested URL. In case that many errors are found, the log slowly becomes unreadable - important information merges with less important ones.

The idea is to show by default only the important part of the error (message only without the stack trace) and the rest should be hidden. The whole error could be displayed in i.e. `--verbose` flag.

psmyrek:

Other: Decreased the size of output log from web crawler. Closes #9062.

Feature: New CLI switch `--quit` (`-q`) to terminate the web crawler scan as soon as the first error is found.

---

### Additional information

From now on, only the first line of the error message will be displayed. My observations shows, that it is completely sufficient and it helps to quickly find out the cause of the error, especially if the log contains a lot of entries.

I chose **not to add** a "verbose" CLI switch, which would display the entire call stack with an error, as I found it of little use. It is certainly not needed to integrate the web crawler with CI, which we are focusing on now.

Instead, I added a new switch `--quit` (`-q`) so that we can finish scanning as soon as the first error is detected. This CLI switch will shorten the _check → correct → check_ cycle when fixing errors.

Other than that, there are also a few minor visual tweaks in the output log presentation. Some errors are now merged together to take up less space. Only the `host` part from the URL is shown in case of an error related to resource download.

Example:

merged-log-entries

### DIFF
--- a/scripts/web-crawler/index.js
+++ b/scripts/web-crawler/index.js
@@ -11,7 +11,7 @@ const puppeteer = require( 'puppeteer' );
 
 const chalk = require( 'chalk' );
 
-const { getBaseUrl, parseArguments, toArray } = require( './utils' );
+const { getBaseUrl, getFirstLineFromErrorMessage, parseArguments, toArray } = require( './utils' );
 
 const { createSpinner, getProgressHandler } = require( './spinner' );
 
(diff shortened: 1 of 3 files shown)


pomek:
Just one question - do we set a proper exit code if an error has been found? Thanks to that, CI knows that the task failed, and the entire build will be marked as red.

psmyrek:
Yes, already implemented: https://github.com/ckeditor/ckeditor5/blob/master/scripts/web-crawler/index.js#L72

pomek (approved):
LGTM.

PR #9190, opened 2021-03-09
Added docs for DiffItem.

### LINKED ISSUE #9176: Differ's getChanges() is missing API docs on returned type

Currently [https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_differ-Differ.html#function-getChanges](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_differ-Differ.html#function-getChanges) has only information that it returns `Array.<Object>`. That's not enough to use this method.

We should have an interface defined for the returned result.

*   What are the properties of returned object?
*   What values `type` can have and what's the meaning of this?

niegowski:

Docs (engine): Added API docs for `Differ#getChanges()` returned type. Closes #9176.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/model/differ.js
+++ b/packages/ckeditor5-engine/src/model/differ.js
@@ -370,9 +370,11 @@ export default class Differ {
 	 * Calculates the diff between the old model tree state (the state before the first buffered operations since the last {@link #reset}
 	 * call) and the new model tree state (actual one). It should be called after all buffered operations are executed.
 	 *
-	 * The diff set is returned as an array of diff items, each describing a change done on the model. The items are sorted by
-	 * the position on which the change happened. If a position {@link module:engine/model/position~Position#isBefore is before}
-	 * another one, it will be on an earlier index in the diff set.
+	 * The diff set is returned as an array of {@link module:engine/model/differ~DiffItem diff items}, each describing a change done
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -1167,3 +1169,114 @@ function _changesInGraveyardFilter( entry ) {
...
+
+/**
+ * The single diff item for inserted nodes.
+ *
+ * @class DiffItemInsert
+ * @implements module:engine/model/differ~DiffItem
+ */
+
+/**
+ *  The type of diff item.
```
Reinmar, line 1193:
```suggestion
 * The type of diff item.
```

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -1167,3 +1169,114 @@ function _changesInGraveyardFilter( entry ) {
...
+
+/**
+ * The single diff item for attribute change.
+ *
+ * @class DiffItemAttribute
+ * @implements module:engine/model/differ~DiffItem
+ */
+
+/**
+ *  The type of diff item.
```
Reinmar, line 1255:
```suggestion
 * The type of diff item.
```

PR #9212, opened 2021-03-11
I/9136 preserving toolbar in caption

### LINKED ISSUE #9136: Re-think the UX for image caption toggling
# ATM

https://github.com/ckeditor/ckeditor5/pull/9090 implements the following UX:

* 👍 click&type, which is likely the common use-case,
* 👎 the toolbar disappears so to revert adding the caption, you need to re-select the image,
* 👎 there's a lot happening at the same time (toolbar disappears, the caption appears and gets highlighted, the selection moves)

## Reverting an existing caption 

# Other UXs

## Don't move the selection to the caption

* 👍 the toolbar stays in place, click again without moving the mouse to get rid of the caption,
* 👍 the selection does not change and the toolbar stays in place, less distractions,
* 👎 typing in a caption requires another click,

## Move the selection to the caption but keep the toolbar

* 👍 click&type, which is likely the common use-case,
* 👍 the toolbar stays in place, click again without moving the mouse to get rid of the caption,
* 👍 the toolbar stays in place, less distractions,
* 👎 possible collision with the balloon toolbar, although we already do that for the table toolbar
* 👎 requires changes in other features (note the grayed out buttons in the toolbar when the selection is in the caption)

magda-chrzescian:

Other (image): The image toolbar should be visible if the selection is placed inside an image caption. Closes #9136.


packages/ckeditor5-image/src/image/utils.js
```diff
@@ -59,6 +59,26 @@ export function getSelectedImageWidget( selection ) {
...
+ * Returns a image widget editing view element if one is among the selection's ancestors.
+ *
+ * @param {module:engine/view/selection~Selection|module:engine/view/documentselection~DocumentSelection} selection
+ * @returns {module:engine/view/element~Element|null}
+ */
+export function getImageWidgetAncestor( selection ) {
+	let parent = selection.getFirstPosition().parent;
+
+	while ( parent ) {
+		if ( parent.name === 'figure' && parent.getCustomProperty( 'image' ) ) {
```
oleq, line 72: Let's not re-invent the wheel :)

```suggestion
		if ( parent.is( 'element' ) && isImageWidget( parent ) ) {
```

Alternatively `isImageWidget()` starts accepting view node (instead of element) and `parent.is( 'element' )` goes there (+tests).

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -105,6 +107,79 @@ describe( 'image widget utils', () => {
 		} );
 	} );
 
+	describe( 'getImageWidgetAncestor()', () => {
+		let frag, caption;
+
+		describe( 'the selection is inside a caption', () => {
```
oleq, line 113:
```suggestion
		describe( 'when the selection is inside a caption', () => {
```

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -105,6 +107,79 @@ describe( 'image widget utils', () => {
...
+			beforeEach( () => {
+				caption = writer.createContainerElement( 'figcaption' );
+				writer.insert( writer.createPositionAt( element, 1 ), caption );
+				frag = new ViewDocumentFragment( viewDocument, [ element ] );
+			} );
+
+			it( 'should return the widget element if the selection is not collapsed', () => {
+				const text = writer.createText( 'foo' );
+				writer.insert( ViewPosition._createAt( caption, 0 ), text );
```
oleq, line 122: `DowncastWriter.createPositionAt()`.

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -105,6 +107,79 @@ describe( 'image widget utils', () => {
...
 	} );
 
+	describe( 'getImageWidgetAncestor()', () => {
+		let frag, caption;
+
+		describe( 'the selection is inside a caption', () => {
+			beforeEach( () => {
+				caption = writer.createContainerElement( 'figcaption' );
+				writer.insert( writer.createPositionAt( element, 1 ), caption );
+				frag = new ViewDocumentFragment( viewDocument, [ element ] );
```
oleq, line 117: We use writer methods whenever possible and avoid class constructors. It is way safer.

`DowncastWriter.createDocumentFragment()`.

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -105,6 +107,79 @@ describe( 'image widget utils', () => {
...
+			it( 'should return the widget element if the selection is collapsed', () => {
+				const selection = writer.createSelection( caption, 'in' );
+
+				expect( getImageWidgetAncestor( selection ) ).to.equal( element );
+			} );
+		} );
+
+		it( 'should return null if an image is selected', () => {
+			// We need to create a container for the element to be able to create a Range on this element.
+			frag = new ViewDocumentFragment( viewDocument, [ element ] );
```
oleq, line 138: \`DowncastWriter.createDocumentFragment()\`

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -105,6 +107,79 @@ describe( 'image widget utils', () => {
...
+
+			const selection = writer.createSelection( element, 'on' );
+
+			expect( getImageWidgetAncestor( selection ) ).to.be.null;
+		} );
+
+		it( 'should return null if an image is a part of the selection', () => {
+			const notWidgetizedElement = writer.createContainerElement( 'p' );
+
+			frag = new ViewDocumentFragment( viewDocument, [ element, notWidgetizedElement ] );
```
oleq, line 148: `DowncastWriter.createDocumentFragment()`

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -105,6 +107,79 @@ describe( 'image widget utils', () => {
...
+			const selection = writer.createSelection( writer.createRangeIn( frag ) );
+
+			expect( getImageWidgetAncestor( selection ) ).to.be.null;
+		} );
+
+		it( 'should return null if a non-widgetized element is the only element in the selection', () => {
+			const notWidgetizedElement = writer.createContainerElement( 'p' );
+
+			// We need to create a container for the element to be able to create a Range on this element.
+			frag = new ViewDocumentFragment( viewDocument, [ notWidgetizedElement ] );
```
oleq, line 159: `DowncastWriter.createDocumentFragment()`

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -105,6 +107,79 @@ describe( 'image widget utils', () => {
...
+
+			const selection = writer.createSelection( notWidgetizedElement, 'on' );
+
+			expect( getImageWidgetAncestor( selection ) ).to.be.null;
+		} );
+
+		it( 'should return null if the selection is inside a figure element, which is not an image', () => {
+			const innerContainer = writer.createContainerElement( 'p' );
+
+			viewDocument = new ViewDocument( new StylesProcessor() );
```
oleq, line 169: Why is another instance of `ViewDocument` needed here?

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -105,6 +107,79 @@ describe( 'image widget utils', () => {
...
+			writer = new ViewDowncastWriter( viewDocument );
+			element = writer.createContainerElement( 'figure' );
+
+			writer.insert( writer.createPositionAt( element, 1 ), innerContainer );
+
+			frag = new ViewDocumentFragment( viewDocument, [ element ] );
```
oleq, line 175: `DowncastWriter.createDocumentFragment()`

oleq (changes requested):

PR #9216, opened 2021-03-11
Docs: Fixed snippet in Implementing an inline widget guide

### LINKED ISSUE #9215: Mistake in Implementing an inline widget guide

After some recent changes, using our code from [Implementing an inline widget guide](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/tutorials/implementing-an-inline-widget.html#full-source-code) will result in the error after inserting the placeholder element. It's because of the code in the `PlaceholderCommand` that calls `selection.getAttribute()`, while `selection` has not been defined. As I can see, it occurs only in the _Full source code_ snippet.

Mgsy:

Docs: Fixed snippet in Implementing an inline widget guide. Closes #9215.


mlewand (approved):
LGTM, good catch.

It always bugged me that we have these snippets duplicated and thus a risk of issues like that 🙁

PR #9221, opened 2021-03-11
Fix (ckeditor5-core): check that command exists in multicommand execute

fedemp:
`_getFirstEnabledCommand` may return `undefined`, but `execute` does not check if returned value is defined.

### Additional information

Not encountered as a bug, but found will creating Typescript typings.


maxbarnas:
@fedemp, thanks for the contribution! We're sorry it took so long. Is it possible for you to sign the CLA as well?

PR #9249, opened 2021-03-16
Conversion:elementToAttribute: do not overwrite an attribute that is already specified

### LINKED ISSUE #8921: Text styles not kept when being converted

pomek:

Fix (engine): While setting attributes upon upcast conversion, don't override attributes that were already set. The correct behavior is to keep the "information" applied by the deepest nodes in the view tree as, in most cases, the deepest node will have precedence (e.g. inline style applied by the deepest node). Closes #8921.

---

Calling:

```js
editor.setData(`<span style="font-family:'helvetica';"> <span style="color:#0000ff;"> <span style="font-size:16px;font-family:'arial';color:#000000;"> aaa </span> </span> </span>`)
```

in the [all features manual test](http://localhost:8125/ckeditor5/tests/manual/all-features.html) with the following changes:

```diff
diff --git a/tests/manual/all-features.js b/tests/manual/all-features.js
index d40a704ce3..c1b8a76565 100644
--- a/tests/manual/all-features.js
+++ b/tests/manual/all-features.js
@@ -57,6 +57,13 @@ ClassicEditor
             SpecialCharacters, SpecialCharactersEssentials, WordCount,
             ImageUpload, CloudServices
         ],
+        fontSize: {
+            supportAllValues: true,
+            options: [ 10, 12, 14]
+        },
+        fontFamily: {
+            supportAllValues: true
+        },
         toolbar: [
             'heading',
             '|',
```

ends with:

image

So it looks like the PR fixes the issue.

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -923,6 +923,8 @@ function onlyViewNameIsDefined( viewConfig, viewItem ) {
 // Helper function for to-model-attribute converter. Sets model attribute on given range. Checks {@link module:engine/model/schema~Schema}
 // to ensure proper model structure.
 //
+// If any node on the given range has already defined an attribute with the same name, its value will not be updated.
+//
 // @param {module:engine/model/range~Range} modelRange Model range on which attribute should be set.
 // @param {Object} modelAttribute Model attribute to set.
 // @param {module:engine/conversion/upcastdispatcher~UpcastConversionApi} conversionApi Conversion API.
@@ -934,11 +936,17 @@ function setAttributeOn( modelRange, modelAttribute, shallow, conversionApi ) {
 
 	// Set attribute on each item in range according to Schema.
 	for ( const node of Array.from( modelRange.getItems( { shallow } ) ) ) {
-		if ( conversionApi.schema.checkAttribute( node, modelAttribute.key ) ) {
-			conversionApi.writer.setAttribute( modelAttribute.key, modelAttribute.value, node );
-
-			result = true;
+		if ( !conversionApi.schema.checkAttribute( node, modelAttribute.key ) ) {
+			continue;
 		}
+
+		// Do not overwrite the attribute if a node has already specified it. See #8921.
+		if ( node.hasAttribute( modelAttribute.key ) ) {
+			continue;
+		}
+
+		conversionApi.writer.setAttribute( modelAttribute.key, modelAttribute.value, node );
+		result = true;
 	}
 
 	return result;
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -940,7 +940,7 @@ function setAttributeOn( modelRange, modelAttribute, shallow, conversionApi ) {
 			continue;
 		}
 
-		// Do not overwrite the attribute if a node has already specified it. See #8921.
+		// Do not override the attribute if it's already present. See #8921.
 		if ( node.hasAttribute( modelAttribute.key ) ) {
 			continue;
 		}
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -940,13 +940,16 @@ function setAttributeOn( modelRange, modelAttribute, shallow, conversionApi ) {
 			continue;
 		}
 
-		// Do not override the attribute if it's already present. See #8921.
+		// Mark the node as consumed even if the attribute will not be updated because it's in a valid context (schema)
+		// and would be converted if the attribute wouldn't be present. See #8921.
+		result = true;
+
+		// Do not override the attribute if it's already present.
 		if ( node.hasAttribute( modelAttribute.key ) ) {
 			continue;
 		}
 
 		conversionApi.writer.setAttribute( modelAttribute.key, modelAttribute.value, node );
-		result = true;
 	}
 
 	return result;
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -896,12 +896,12 @@ function prepareToAttributeConverter( config, shallow ) {
 			data = Object.assign( data, conversionApi.convertChildren( data.viewItem, data.modelCursor ) );
 		}
 
-		// Set attribute on current `output`. `Schema` is checked inside this helper function.
-		const attributeWasSet = setAttributeOn( data.modelRange, { key: modelKey, value: modelValue }, shallow, conversionApi );
+		// Set attribute on current `output`.
+		setAttributeOn( data.modelRange, { key: modelKey, value: modelValue }, shallow, conversionApi );
 
-		if ( attributeWasSet ) {
-			conversionApi.consumable.consume( data.viewItem, match.match );
-		}
+		// Mark the node as consumed even if the attribute will not be updated because it's in a valid context (schema)
+		// and would be converted if the attribute wouldn't be present. See #9249.
+		conversionApi.consumable.consume( data.viewItem, match.match );
 	};
 }
 
@@ -930,29 +930,20 @@ function onlyViewNameIsDefined( viewConfig, viewItem ) {
 // @param {module:engine/conversion/upcastdispatcher~UpcastConversionApi} conversionApi Conversion API.
 // @param {Boolean} shallow If set to `true` the attribute will be set only on top-level nodes. Otherwise, it will be set
 // on all elements in the range.
-// @returns {Boolean} `true` if attribute was set on at least one node from given `modelRange`.
 function setAttributeOn( modelRange, modelAttribute, shallow, conversionApi ) {
-	let result = false;
-
 	// Set attribute on each item in range according to Schema.
 	for ( const node of Array.from( modelRange.getItems( { shallow } ) ) ) {
 		if ( !conversionApi.schema.checkAttribute( node, modelAttribute.key ) ) {
 			continue;
 		}
 
-		// Mark the node as consumed even if the attribute will not be updated because it's in a valid context (schema)
-		// and would be converted if the attribute wouldn't be present. See #8921.
-		result = true;
-
-		// Do not override the attribute if it's already present.
+		// Do not override the attribute if it's already present. See #8921.
 		if ( node.hasAttribute( modelAttribute.key ) ) {
 			continue;
 		}
 
 		conversionApi.writer.setAttribute( modelAttribute.key, modelAttribute.value, node );
 	}
-
-	return result;
 }
 
 // Helper function for upcasting-to-marker conversion. Takes the config in a format requested by `upcastElementToMarker()`
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -872,7 +872,7 @@ function prepareToAttributeConverter( config, shallow ) {
 		const modelValue = typeof config.model.value == 'function' ?
 			config.model.value( data.viewItem, conversionApi ) : config.model.value;
 
-		// Do not convert if the attribute building function returned a falsy value.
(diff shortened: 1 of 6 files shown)


Mgsy:
PR is ready for testing, @LukaszGudel @FilipTokarski could you please take a look at it?

LukaszGudel:
It's looking good :+1: 

I've also checked reported issues with nested text part language, nested font tags and mixed font and spans tags and everything seems to work correctly.

FilipTokarski:
I didn't find any bugs, looks ok 👍

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -934,11 +936,17 @@ function setAttributeOn( modelRange, modelAttribute, shallow, conversionApi ) {
...
 	// Set attribute on each item in range according to Schema.
 	for ( const node of Array.from( modelRange.getItems( { shallow } ) ) ) {
-		if ( conversionApi.schema.checkAttribute( node, modelAttribute.key ) ) {
-			conversionApi.writer.setAttribute( modelAttribute.key, modelAttribute.value, node );
+		if ( !conversionApi.schema.checkAttribute( node, modelAttribute.key ) ) {
+			continue;
+		}
 
-			result = true;
+		// Do not overwrite the attribute if a node has already specified it. See #8921.
```
Reinmar, line 943:
```suggestion
		// Do not override the attribute if it's already present. See #8921.
```

Reinmar:
Thanks for detailed tests. That makes it easier for me to close it cause it's the type of a change that I'm a bit scared by :D

Reinmar:
While rewriting the changelog entry I realized that this change may have a wider impact than it seems. For inline styles, the innermost style will have the precendence. But it's a bit hard to predict what makes sense for other attributes and other situations (and this helper is used for all "to attribute" converters, so for many scenarios).

I think that the change may still make sense as conflicts on this level (an attribute that is set by two converters) might not have a single solution so we just need to choose one and stick to it (and sticking to innermost makes sense at least for inline styles). But I wonder what @scofalik and @niegowski  think.

niegowski:
I'm not sure how this would go with GHS, maybe we could check if this is not causing problems for GHS attributes? cc @jacekbogdanski

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -934,11 +936,17 @@ function setAttributeOn( modelRange, modelAttribute, shallow, conversionApi ) {
...
-			conversionApi.writer.setAttribute( modelAttribute.key, modelAttribute.value, node );
+		if ( !conversionApi.schema.checkAttribute( node, modelAttribute.key ) ) {
+			continue;
+		}
 
-			result = true;
+		// Do not override the attribute if it's already present. See #8921.
+		if ( node.hasAttribute( modelAttribute.key ) ) {
+			continue;
 		}
```
jacekbogdanski, line 946: Maybe we should set `result = true` before this `if`, so despite that, we didn't really set element attribute, we are still consuming it's allowed by the schema, so upcast helpers ensure converter priority?

jacekbogdanski (commented):
Yes, these changes will break GHS attributes. On `master`, attribute upcast helpers are exhaustive, so:

https://github.com/ckeditor/ckeditor5/blob/76011913a7762b5b77a856a5b80434151e9c8066/packages/ckeditor5-font/src/fontcolor/fontcolorediting.js#L105-L116

Will match every `color` style for every `span` element including element nesting. After the changes, helpers are no longer exhaustive which could be tested (with Font plugin):

```js
editor.conversion.for( 'upcast' ).add( dispatcher => {
	dispatcher.on( 'element:span', ( evt, data, conversionApi ) => {
		const isExhausive = !!conversionApi.consumable.test( data.viewItem, { styles: 'color' } );
		console.log( 'is exhausive', isExhausive );
	}, { priority: 'low' } );
} );
```

The above converter will be able to match green span:
```html
<p><span style="color: green;">foo<span style="color: blue;">bar</span></span></p>
```

However, with the Font feature, we are adding a converter that should match any color, so it seems reasonable that it's consumed even if not applied.

scofalik:
It seems logical that inner elements' attributes have bigger priority than outers. I am still thinking about counter-examples but I am not finding one. Actually, it would all work correctly out of the box if attributes were converted "in a different loop" than elements (so first convert elements, all their children and only then the general attributes like color).

This problem concerns general styles (like color, which can be added anywhere) and also some specific attributes for elements that can be nested (`<li>...<li>..</li>...</li>`). This fix might also fix some problems with converting attributes on list items.

The change is small and I like it.

One thing that came to my mind regarding GHS is scenario like this:

```html
<span style="font-weight:bold;">
	XXX
	<span style="font-weight:normal;">
		YYY
	</span>
	ZZZ
</span>
```

Or even bold -> normal -> bold.

I am afraid that current converters will not handle this properly, unless we are explicitly setting `bold: false` attribute in the model. Which is redundant. But to make it work, we might need to set it and maybe then clear it after upcast is done.

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
scofalik: Makes sense, but TBH, I wonder what would happen if we would consume always, not only when something was actually applied.

TBH, the converter tried to do its job so the attribute was processed. If there is another converter that could do this job better it should be added with a higher priority and overwrite the "more default" converter.

Let's say that we have such scenario:

```html
<span class="X">Foo <Y></Y> Bar</span>
```

And let's say that class `X` can be "set on text" but not on `<Y>`. Even if there is a converter that would handle setting class `X` on `<Y>`, it wouldn't work, cause text is there, so class `X` is consumed (you may say it is consumed for span, text and `Y` element).

If there's a different order of converters, and there is converter that only handles `<Y>`, then you wouldn't have conversion for text.

So, it looks to me that consuming only if something was applied is not needed cause it does not guarantee a better outcome.

Not sure about manual testing but I actually commented out the `if` here:

```javascript
// if ( attributeWasSet ) {
	conversionApi.consumable.consume( data.viewItem, match.match );
// }
```

And all tests passed.

scofalik (changes requested):
I agree with @jacekbogdanski comment about consuming attributes.

pomek:
I've pushed a fix that we were talking about yesterday during the sync. @jacekbogdanski, would you take a look and merge if you have no objections?

jacekbogdanski:
I'm not sure if I messed up something during the sync, but if I remember correctly we want to consume attributes always, even if it's not allowed by the schema, so basically what explained @scofalik in https://github.com/ckeditor/ckeditor5/pull/9249#discussion_r604028825

In that case, we could remove `setAttributeOn` function return statement as it's only used in this single place, and delete `if` statement from https://github.com/ckeditor/ckeditor5/blob/0413f8ee98714b89e55fdac8430bc6a4f807d426/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L902-L904

so attributes are always consumed if matched.

pomek:
@jacekbogdanski, fixed.

Reinmar:
@jacekbogdanski Could you explain here why we wanted to change this (consume every time)? Just for the future reference, because I already forgot the "why" :D

jacekbogdanski:
We are still checking if we can apply the attribute to a node -https://github.com/ckeditor/ckeditor5/blob/3acb05d419c965eb22dbbab63c274b4920b481b6/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L936-L938

but if an attribute is not allowed by data schema, it doesn't mean that converter didn't successfully match the expected MatcherPattern. At this point, we know that converter wanted to consume this attribute, so it should do it even if an attribute is not allowed to write into the model. So the code like:

```js
// Comment out to disallow this attribute.
// editor.model.schema.extend( '$text', allowAttributes: 'fontColor' );

editor.conversion.for( 'upcast' ).elementToAttribute( {
	view: {
		name: 'span',
		styles: {
			'color': /[\s\S]+/
		}
	},
	model: {
		key: 'fontColor',
		value: renderUpcastAttribute( 'color' )
	}

} );

editor.conversion.for( 'upcast' ).add( dispatcher => {
	dispatcher.on( 'element:span', ( evt, data, conversionApi ) => {
		const isExhausive = !!conversionApi.consumable.test( data.viewItem, { styles: 'color' } );
		if ( isExhausive ) throw 'At this point color style should be already consumed!'
	}, { priority: 'low' } );
} );
```

should have a single valid execution path for the lower priority converter - it shouldn't be able to consume attributes already processed by the converter with higher priority.

pomek:
Removing this: https://github.com/ckeditor/ckeditor5/blob/3acb05d419c965eb22dbbab63c274b4920b481b6/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L936-L938

Causes that 12 tests started fails (test command: `yarn run test -f engine/conversion/upcasthelpers.js,font/fontcolor/fontcolorediting.js,highlight/highlightediting.js`)

pomek:
And the bug that we were supposed to fix, appears again.

jacekbogdanski:
@pomek yeah this check is fine, sorry if I was not precise enough with https://github.com/ckeditor/ckeditor5/pull/9249#issuecomment-813935851
I was answering https://github.com/ckeditor/ckeditor5/pull/9249#issuecomment-812919831 request for an explanation for the latest changes.

Reinmar:
Code-wise the PR looks good to me.

cc @ckeditor/qa-team Could you retest this change? The last modification that we made here is rather hard to encounter when testing manually (probably even impossible right now), so I'd rather focus on random tests (e.g. random content from the web) on all features.

Reinmar:
@ckeditor/qa-team Could you test whether it covers #9425 too?

Mgsy:
> [**@ckeditor/qa-team**](https://github.com/orgs/ckeditor/teams/qa-team) Could you test whether it covers [#9425](https://github.com/ckeditor/ckeditor5/issues/9425) too?

This PR changes behaviour described in #9425, however, it doesn't fix the issue - the attribute on the paragraph is lost.

After calling `editor.setData('<div style="text-align: justify;"><p style="text-align: center;">JavaScript</p></div>')` results are as below:

**Changes from this PR**

```html
<p>JavaScript</p>
```

**Master**

```html
<p style="text-align: justify;">JavaScript</p>
```

pomek:
Before introducing this change: 3acb05d419c965eb22dbbab63c274b4920b481b6, everything worked.

```js
editor.setData(`<p style="text-align: justify;"><p style="text-align: center;">JavaScript</p></p>`)

editor.getData()
"<p style="text-align:justify;">&nbsp;</p><p style="text-align:center;">JavaScript</p><p>&nbsp;</p>"
```

pomek:
The `TableCellProperties#horizontalAlignment` converter catches the `style="text-align: *"` attribute. Even, if the schema does not allow for setting the `horizontalAlignment` attribute on the `paragraph` element, the current code will consume the attribute and will not update the `paragraph[alignment]` element properly. Hence, we need a check whether `setAttributeOn` updated the attribute or not.

https://github.com/ckeditor/ckeditor5/blob/a127c971ba23e4e6ed58d83327db617d85063b24/packages/ckeditor5-table/src/tablecellproperties/tablecellpropertiesediting.js#L120-L143

jacekbogdanski:
The code looks good. The recent changes seem to be fixing https://github.com/ckeditor/ckeditor5/issues/9425. I've also added a unit test confirming that it fixed `textpartlanguage` feature.

@ckeditor/qa-team could you take a look if everything fine? I know it's 3rd time already but the code changed again since your testing. I've already tested if it fixes https://github.com/ckeditor/ckeditor5/issues/9425 but it would be good to double-check it also.

FilipTokarski:
I didn't find anything new, looks fine 👍

LukaszGudel:
Same here, looking good to me 👍

jacekbogdanski (approved):
LGTM 🎉

PR #9251, opened 2021-03-16
Added attributes support in default markers highlight downcast conversion for widgets

### LINKED ISSUE #8362: Marker highlight for widgets should support attributes

At the moment there is a difference in how `markerToHighlight` is handled for texts and elements (widgets). Let me start by reminding that as `markerToHighlight` `view` parameter we set a view descriptor. In the text highlighting, from that descriptor, we create an attribute element, so all descriptor options are supported (classes, attributes, styles).

However, when it comes to widget highlighting, it only supports adding class to the widget view element. We should also enable attributes and styles.

scofalik:

Feature (widget): Added attributes support in default markers highlight downcast conversion for widgets. Closes #8362.


scofalik:
I've also expanded a test in CKE5 that checks marker to highlight conversion, so it is in line with a changed example in docs.

pomek (approved):
LGTM.

PR #9295, opened 2021-03-18
`DataController#toView()` should have a default value for the `options` parameter

### LINKED ISSUE #9293: DataController#toView() should have default value for options parameter

Some time ago we started to pass `options` parameter (available in `DataController#get()` and `#toView()` methods. While in `get()` it received default value (`{}`), it does not have a default value in `toView()`. Sometimes we use `toView()` directly and when doing this we have conversion errors in converters that check `conversionApi.options` object. One such place is clipboard handling.

Same goes for `stringify()`.

scofalik:

Fix (engine): `DataController#toView()` should have a default value for the `options` parameter. Closes #9293.

### DIFF
--- a/packages/ckeditor5-engine/src/controller/datacontroller.js
+++ b/packages/ckeditor5-engine/src/controller/datacontroller.js
@@ -205,7 +205,7 @@ export default class DataController {
 	 *
 	 * @param {module:engine/model/element~Element|module:engine/model/documentfragment~DocumentFragment} modelElementOrFragment
 	 * Element whose content will be stringified.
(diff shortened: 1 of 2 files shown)


pomek (approved):
LGTM.

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -208,7 +208,7 @@ export default class DataController {
 	 * @param {Object} [options] Additional configuration passed to the conversion process.
```
pomek, line 208:
```suggestion
	 * @param {Object} [options={}] Additional configuration passed to the conversion process.
```

PR #9319, opened 2021-03-22
Added retries in crawler in case of an error

psmyrek:

Internal: Added retries in crawler in case of an error. Closes https://github.com/cksource/ckeditor5-internal/issues/687.

---

### Additional information

From now on, the crawler will try to open the page up to 3 times when it encounters an error to eliminate various random causes of these errors.

All NPM badges in the documentation (except the API, which is not scanned by default by the crawler) have been converted to `` tags. The `loading="lazy"` attribute has been added, so these resources do not delay the `load` event anymore.

Additionally, request failures from `badge.fury.io` domain are now ignored.


pomek (approved):
LGMT.

PR #9327, opened 2021-03-22
Removed `ckeditor5` dep from ckeditor5-watchdog

### LINKED ISSUE #9315: Watchdog should not import CKEditor 5 utils as this leads to package duplication problem in integrations
In other words, we should get rid of this line: [https://github.com/ckeditor/ckeditor5/blob/a127c971ba23e4e6ed58d83327db617d85063b24/packages/ckeditor5-watchdog/src/contextwatchdog.js#L12](https://github.com/ckeditor/ckeditor5/blob/a127c971ba23e4e6ed58d83327db617d85063b24/packages/ckeditor5-watchdog/src/contextwatchdog.js#L12)

And write the \`toArray()\` utility for this package independently.

The current situation leads to the package duplication problem because the watchdog is bundled in the integration library independently from the CKEditor 5 bundle.

pomek:

Fix (watchdog): Removed import from the `ckeditor5` package. Closes #9315.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-watchdog/package.json
+++ b/packages/ckeditor5-watchdog/package.json
@@ -10,7 +10,6 @@
   ],
   "main": "src/index.js",
   "dependencies": {
-    "ckeditor5": "^26.0.0",
     "lodash-es": "^4.17.15"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-watchdog/package.json
```diff
@@ -20,6 +19,7 @@
     "@ckeditor/ckeditor5-paragraph": "^26.0.0",
     "@ckeditor/ckeditor5-theme-lark": "^26.0.0",
     "@ckeditor/ckeditor5-utils": "^26.0.0",
+    "ckeditor5": "^26.0.0",
```
ma2ciek, line 22: I'm not sure if we use it in tests at all. But I'm fine with it.

ma2ciek:
Thanks @pomek for handling this.

packages/ckeditor5-watchdog/package.json
pomek: DLL, webpack cfg.

PR #9337, opened 2021-03-23
config.initialData was skipped when set to an empty string and DOM data was used instead

### LINKED ISSUE #8974: Empty `config.initialData` is overwritten with data from DOM

To reproduce the error you need DOM element with some data and then create an editor on that DOM element and use `initialData: ''` in config.

In such configuration, I'd expect that the editor will be empty, but it actually is inited with the data from the DOM. That's because this code:

```js
const initialData = config.initialData || getInitialData( sourceElementOrData );

return editor.data.init( initialData );
```

Which is incorrect. This code treats empty and unset initial data the same way.

scofalik:

Fixed: `config.initialData` was skipped when set to an empty string and DOM data was used instead. Closes #8974.

---

### Additional information

Alternative commit message:

Fixed: Editor was not initialized with empty data for `config.initialData` set to an empty string.

### DIFF
--- a/packages/ckeditor5-editor-balloon/src/ballooneditor.js
+++ b/packages/ckeditor5-editor-balloon/src/ballooneditor.js
@@ -213,7 +213,7 @@ export default class BalloonEditor extends Editor {
 							throw new CKEditorError( 'editor-create-initial-data', null );
 						}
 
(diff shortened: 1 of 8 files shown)


pomek (approved):
LGTM.

Issue #9345, opened 2021-03-23
Block filler recognition is incorrect when converting from DOM to view


scofalik:
## 📝 Provide detailed reproduction steps (if any)

Filler element `&nbsp;` is not correctly removed when it was placed after `<br />` and the data is loaded.

1.  `editor.setData( '<p>A<br></p>' )` or prepare such content in the editor.
2.  `editor.getData()` -> `'<p>A<br>&nbsp;</p>'` -> this is fine, filler is needed.
3.  `editor.setData( '<p>A<br>&nbsp;</p>' )`

### ✔️  Expected result

There is no space in the editor (in the model).

### ❌  Actual result

The filler `&nbsp;` has not been removed and has been added as a space to the model.

### What is wrong

Block filler element is not recognised properly. We have a mechanism that should remove this `&nbsp;` when it was converted from DOM to view but the mechanism didn't work.

We should start from looking at the method that adds the filler. For a container element it is:

```javascript
export function getFillerOffset() {
	const children = [ ...this.getChildren() ];
	const lastChild = children[ this.childCount - 1 ];

	// Block filler is required after a `<br>` if it's the last element in its container. See #1422.
	if ( lastChild && lastChild.is( 'element', 'br' ) ) {
		return this.childCount;
	}

	for ( const child of children ) {
		// If there's any non-UI element – don't render the bogus.
		if ( !child.is( 'uiElement' ) ) {
			return null;
		}
	}

	// If there are only UI elements – render the bogus at the end of the element.
	return this.childCount;
}
```

As you can see, there is some logic here, but tl;dr - the filler is added at the end of the element if there are no or only UI elements in the container or if there is `<br>` at the end of the container.

Actually, different types of elements have different rules for that. And AFAIK, it is possible to set a custom behavior.

The logic presented above is not reflected in the function that tries to recognise and remove fillers. It looks like this:

```javascript
function isNbspBlockFiller( domNode, blockElements ) {
	const isNBSP = isText( domNode ) && domNode.data == '\u00A0';

	return isNBSP && hasBlockParent( domNode, blockElements ) && domNode.parentNode.childNodes.length === 1;
}
```

As you can see, the logic here is simpler, it just checks if this is an `&nbsp;` in an empty element. It is not compatible with `getFillerOffset()` for container elements and it would get outdated again if `getFillerOffset()` changes.

### Solutions

[...]

PR #9347, opened 2021-03-23
 Extra space was added after a soft-break when data was loaded if the soft-break was the last element in its parent

### LINKED ISSUE #9346: Remove block filler &nbsp; inserted after <br> during DOM->view conversion

This is a ticket to provide "quick solution" as described in #9345. Please read that ticket for explanation.

scofalik:

Fixed (engine): Extra space was added after a soft-break when data was loaded if the soft-break was the last element in its parent. Closes #9346.

---

### Additional information

I decided to go with commit message that explains the benefit to the end-user. However, I am not sure if it should be "soft-break", "softbreak" or "soft break" :D.


scofalik:
Unfortunately, marker element is a UI element, and `&nbsp;` is added in such case. This means that we should probably implement the "good solution" as described in #9345.

PR #9352, opened 2021-03-24
I/9117 inline images in tables

### LINKED ISSUE #9117: Integrate inline images with table feature

### LINKED ISSUE #9177: Markdown output results in different content when fed back into the editor
On https://ckeditor.com/docs/ckeditor5/latest/features/markdown.html, which I suppose is v26.0.0 right now, try the following:

- Click Bold and Inline Code, and type hello
- You should see the output be \`\*\*hello\*\*\`
- Then in the console take that output and pass it to editor.setData
- The contents of the editor will not be the same as before

Seems to happen with Italic or Strikethrough as well- basically any inline style going in an Inline Code block.

magda-chrzescian:

Fix (image): An `img` element nested in a table should be upcasted to the `imageInline` model element. Closes #9117.

Fix (table): An inline image should be always positioned and resized relative to the whole width of the table cell. See #9117.

Fix (image): A table cell should adjust its size to an inline image in its original size. See #9117.


packages/ckeditor5-table/theme/table.css
```diff
@@ -33,6 +33,12 @@
 			border: 1px solid hsl(0, 0%, 75%);
 		}
 
+		/* Inline HTML elements nested in the span should always be dimensioned in relation to the whole cell width.
+		See https://github.com/ckeditor/ckeditor5/issues/9117. */
+		& td > span {
```
oleq, line 38: What about th used in headers?

packages/ckeditor5-image/theme/imageresize.css
```diff
@@ -24,6 +24,14 @@
 	}
 }
 
+.ck.ck-editor__editable {
+	/* The resized inline image nested in the table should respect its parent size.
+	See https://github.com/ckeditor/ckeditor5/issues/9117. */
+	& td .image-inline.image_resized img {
```
oleq, line 30: What about th used in theaders?

packages/ckeditor5-image/theme/image.css
```diff
@@ -78,4 +78,10 @@
 	& .image-inline.ck-widget_selected {
 		z-index: 1;
 	}
+
+	/* The inline image nested in the table should have its original size if not resized.
+	See https://github.com/ckeditor/ckeditor5/issues/9117. */
+	& td .image-inline img {
```
oleq, line 84: What about th used in headers?

packages/ckeditor5-image/tests/manual/resizing-images-in-tables.js
```diff
@@ -0,0 +1,101 @@
...
+		} );
+
+	window.editor = editor;
+
+	editor.plugins.get( 'FileRepository' ).createUploadAdapter = loader => {
+		const adapterMock = new UploadAdapterMock( loader );
+
+		return adapterMock;
+	};
+} )();
```
oleq, line 101:
```javascript
async function createEditor( resizeUnit ) {
    // ...

    return editor;
}

window.editorPx = await createEditor( 'px' );
window.editorPercent = await createEditor( '%' );

CKEditorInspector.attach( {
	px: window.editorPx,
	percent: window.editorPercent
} );
```

looks like both editors share the same config except for the `image.resizeUnit`. Also, in your code the inspector attaches by accident, check out the [code responsible for that](https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-tests/lib/utils/manual-tests/attachinspector.js#L15-L24).

packages/ckeditor5-image/tests/manual/resizing-images-in-tables.js
```diff
@@ -0,0 +1,101 @@
...
+					'imageStyle:wrapText',
+					'imageStyle:breakText',
+					'|',
+					'imageTextAlternative', '|',
+					'resizeImage'
+				]
+			}
+		} );
+
+	window.editor = editor;
```
oleq, line 94: This overrides the reference created in L50.

oleq (changes requested):
The functionality looks 👍 Got some concerns regarding the implementation, though.

packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -599,4 +605,185 @@ describe( 'image widget utils', () => {
...
+						return false;
+					} else {
+						return true;
+					}
+				} );
+			} );
+
+			afterEach( () => {
+				expect( getImageTypeMatcher( 'image', editor ) ).to.eql( returnValue );
+				expect( getImageTypeMatcher( 'imageInline', editor ) ).to.eql( returnValue );
```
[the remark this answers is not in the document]
magda-chrzescian: I thought I saw something similar in the link feature, but I guess I was wrong :( So it doesn't seem to be good practice and I'm scared too now.

PR #9357, opened 2021-03-25
Introduced "markedNbsp" block filler mode

### LINKED ISSUE #9345: Block filler recognition is incorrect when converting from DOM to view

scofalik:

Feature (engine): Introduced new block filler mode "markedNbsp" in `DomConverter`, in which `<span data-cke-filler="true">&nbsp;</span>` are inserted, to prevent leaking extra space characters into the data.

Feature (engine): Introduced `useFillerType()` in `HtmlDataProcessor` and `XmlDataProcessor` to switch between using marked and regular nbsp block fillers. Closes #9345.

MINOR BREAKING CHANGE (engine): Added new method `useFillerType()` in `DataProcessor` interface. Classes that base on this interface should implement `useFillerType()` to avoid errors.

### DIFF
--- a/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
+++ b/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
@@ -94,8 +94,8 @@ export default class HtmlDataProcessor {
 	}
 
 	/**
-	 * If the processor is set to use marked filler, it will insert marked nbsp fillers (`<span data-cke-filler="true">&nbsp;</span>`),
-	 * instead of regular nbsp characters (`&nbsp;`).
+	 * If the processor is set to use marked fillers, it will insert nbsp fillers wrapped in spans
+	 * (`<span data-cke-filler="true">&nbsp;</span>`), instead of regular nbsp characters (`&nbsp;`).
 	 *
 	 * This mode allows for more precise handling of block fillers (so they don't leak into editor content) but bloats the editor
 	 * data with additional markup.
@@ -104,7 +104,7 @@ export default class HtmlDataProcessor {
 	 *
 	 * @param {Boolean} value Whether to use (`true`) or not (`false`) marked nbsp fillers.
 	 */
-	useMarkedFiller( value ) {
+	useMarkedFillers( value ) {
 		this._domConverter.blockFillerMode = value ? 'markedNbsp' : 'nbsp';
 	}
 
--- a/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
+++ b/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
@@ -102,10 +102,10 @@ export default class HtmlDataProcessor {
 	 *
 	 * This mode may be required by some features and will be turned on by them automatically.
 	 *
-	 * @param {Boolean} value Whether to use (`true`) or not (`false`) marked nbsp fillers.
+	 * @param {'default'|'marked'} type Whether to use default or marked nbsp block fillers.
 	 */
-	useMarkedFillers( value ) {
-		this._domConverter.blockFillerMode = value ? 'markedNbsp' : 'nbsp';
+	useFillerType( type ) {
+		this._domConverter.blockFillerMode = type == 'marked' ? 'markedNbsp' : 'nbsp';
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
```diff
@@ -93,6 +93,21 @@ export default class HtmlDataProcessor {
...
+	/**
+	 * If the processor is set to use marked filler, it will insert marked nbsp fillers (`<span data-cke-filler="true">&nbsp;</span>`),
+	 * instead of regular nbsp characters (`&nbsp;`).
+	 *
+	 * This mode allows for more precise handling of block fillers (so they don't leak into editor content) but bloats the editor
+	 * data with additional markup.
+	 *
+	 * This mode may be required by some features and will be turned on by them automatically.
+	 *
+	 * @type {Boolean} value Whether to use (`true`) or not (`false`) marked nbsp fillers.
```
niegowski, line 105:
```suggestion
	 * @param {Boolean} value Whether to use (`true`) or not (`false`) marked nbsp fillers.
```

packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
```diff
@@ -93,6 +93,21 @@ export default class HtmlDataProcessor {
...
+	 */
+	useMarkedFiller( value ) {
+		this._domConverter.blockFillerMode = value ? 'markedNbsp' : 'nbsp';
+	}
```
niegowski, line 109: `DomConverter` has a third (default) option `'br'` so maybe this should allow passing any `blockFillerMode` to the `DomConverter`?

scofalik: `HtmlDataProcessor` by default uses `nbsp` and does not have any way to change it to `br`, so this is why I don't want to introduce such possibility.

niegowski (approved):
👍  LGTM

PR #9362, opened 2021-03-26
Enabled support for nested tables

### LINKED ISSUE #3232: Support for nested tables
Add support for nesting one table inside another.

This feature wasn't on our roadmap yet, but let's gather a feedback if this is a common requirement.

**NOTE**: It is possible to enable support for nested tables. See https://github.com/ckeditor/ckeditor5-table/issues/121#issuecomment-528251188 for how to do that.

--- 

If you'd like this to be officially supported add 👍 to this ticket.

psmyrek:

Feature (table): Enabled support for nested tables. Closes #3232.

---

### Additional information

Support for nested tables in CF has a dedicated follow-up: https://github.com/cksource/collaboration-features/issues/3891.
After a brief check, I wasn't able to blow up the editor in CF samples.


Mgsy:
PR is ready for testing @ckeditor/qa-team.

FilipTokarski:
Seems to be fine, I didn't find any bugs 👍

packages/ckeditor5-list/tests/listediting.js
```diff
@@ -1646,7 +1646,15 @@ describe( 'ListEditing', () => {
 										'</ul>' +
 										'</li>' +
 										'</ul>' +
-										'<p>e</p>' +
```
Reinmar, line 1649: Why did you decide to modify this test?

psmyrek: This test used nested tables so I didn't want to change the scope of this test. I just decided to update it to the changes that my PR introduces. Is it wrong?

packages/ckeditor5-table/tests/manual/table.html
```diff
@@ -201,4 +201,38 @@
 		</tr>
 		</tbody>
 	</table>
+
+	<p>Nested tables</p>
```
Reinmar, line 205: The same change would be good in the `tableproperties.html` manual test. Table props is a feature that I'd be worried may not work well with nested tables.

psmyrek: Nested and styled table has been added to `tableproperties.html` test.

packages/ckeditor5-list/tests/listediting.js
Reinmar: No, it's completely fine. I didn't understand the change at first, but it makes perfect sense.

packages/ckeditor5-table/tests/manual/table.html
```diff
@@ -201,4 +201,38 @@
...
+	<table>
+		<tr>
+			<td></td>
+			<td></td>
+			<td></td>
+		</tr>
+		<tr>
+			<td></td>
+			<td>
+				Paragraph before table
```
Reinmar, line 216: This should be a `<p>`, if we want to have here content that would normally be created by an editor.

Also, all empty `<td>`s should be filled with `&nbsp` too.

packages/ckeditor5-table/tests/manual/tableproperties.html
```diff
@@ -23,6 +23,33 @@
...
+											<td style="border:2px dashed black;background-color: lightblue"><p>Column Y</p></td>
+											<td style="background: palevioletred"><p>Column Z</p></td>
+										</tr>
+									</thead>
+									<tbody>
+										<tr>
+											<td style="border-color: deeppink deepskyblue;border-style: solid;border-width: 1px;border-bottom-width: 2px;border-right-style: dotted;">
+												<p>X1</p>
+											</td>
+											<td style="border: 1px solid blue;"><p>Y1</p></td>
```
Reinmar, line 41: Spaces or no after `:`? IDK if you copied this piece from somewhere else, but let's try to be consistent (in L45-47 you have no space). Let's use the same format that the editor uses when outputting content.

Reinmar: Hmmm, OTOH, a part of table props' logic is how it parses inline styles, so that may be the reason for lack of consequence. So I take the above back – let's keep it this way.

packages/ckeditor5-table/tests/manual/table.html
psmyrek: Done.

Reinmar:
I reported #9420 for d&d's behavior.

Reinmar:
I added a manual test because we may need to be able to test this from time to time. I also added docs for how to disable nesting.

PR #9371, opened 2021-03-26
Add support for the `config.forceValue` option in `ListCommand`

ma2ciek:

Internal: Introduced the `config.forceValue` option to `ListCommand` that allows overriding the default toggle-like behavior to allow more predictable state changes during applying track changes suggestions.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/src/listcommand.js
+++ b/packages/ckeditor5-list/src/listcommand.js
@@ -57,7 +57,7 @@ export default class ListCommand extends Command {
 	 * @param {Object} [options] Command options.
 	 * @param {Boolean} [options.forceValue] If set, it will force the command behavior. If `true`, the command will try to convert the
 	 * selected items and potentially the neighbor elements to the proper list items.
-	 * If not set, the command will look for its current value to decide what it should do.
+	 * If not set, the command will toggle selected elements to list items or paragraphs, depending on the selection.
 	 */
 	execute( options = {} ) {
 		const model = this.editor.model;
--- a/packages/ckeditor5-list/src/listcommand.js
+++ b/packages/ckeditor5-list/src/listcommand.js
@@ -51,18 +51,23 @@ export default class ListCommand extends Command {
 	}
 
 	/**
-	 * Executes the command.
+	 * Executes the list command.
 	 *
-	 * @protected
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-list/src/listcommand.js
```diff
@@ -51,18 +51,23 @@ export default class ListCommand extends Command {
...
 
 	/**
-	 * Executes the command.
+	 * Executes the list command.
 	 *
-	 * @protected
+	 * @fires execute
+	 * @param {Object} [options] Command options.
+	 * @param {Boolean} [options.forceValue] If set, it will force the command behavior. If `true`, the command will try to convert the
+	 * selected items and potentially the neighbor elements to the proper list items.
```
scofalik, line 59: "If set to `false` it will convert selected elements to paragraphs."

packages/ckeditor5-list/src/listcommand.js
```diff
@@ -51,18 +51,23 @@ export default class ListCommand extends Command {
...
+	 * If not set, the command will look for its current value to decide what it should do.
```
scofalik, line 60:
```suggestion
	 * If not set, the command will toggle selected elements to list items or paragraphs, depending on the selection.
```

packages/ckeditor5-list/tests/listcommand.js
```diff
@@ -133,6 +133,33 @@ describe( 'ListCommand', () => {
 				} );
 			} );
 
+			describe( 'options.forceValue', () => {
```
scofalik, line 136: I'd make tests showing that lisItem with forceValue true does not change anything, and similarly for paragraphs and false. AFAICS, these tests describe "normal" behavior.

scofalik (changes requested):

packages/ckeditor5-list/tests/listcommand.js
ma2ciek: There are two assertions for each test to check both scenarios. But I can simplify these tests and split the 4 assertions into 4 tests.

scofalik: :facepalm:

Issue #2737, opened 2018-08-23
Implement a web component to handle <oembed url="">


Reinmar:
In the semantic mode (`config.media.semanticDataOutput`) the editor produces `<oembed>` tags. Maybe we can include a simple WC in the docs for handling those with Embedly/Iframely? Or is it pointless because Embedly/Iframely's JS libs allow replacing these elements automatically?

tony:
I was caught off guard by this - I didn't realize I needed to also have a separate rendering system.

I thought if YouTube showed inside the Editor, things would just work at the get go.

tony:
One more thing: A web component would need to have a hyphen, https://github.com/WICG/webcomponents/issues/658

As of a127c97, we use `<oembed>`:

https://github.com/ckeditor/ckeditor5/blob/b278fde89d1eb635be7e4e3a57d8dba2bd0f98a6/packages/ckeditor5-media-embed/src/mediaregistry.js#L236

So we need to switch to something with a hyphen, e.g. `<o-embed>`, which should work

tony:
Here's an `<o-embed>` web component (proof of concept w/ YouTube):

License MIT, TypeScript:

- https://github.com/tony/oembed-component
- https://oembed-component.git-pull.com/
- https://www.npmjs.com/package/react
- https://codepen.io/attachment/pen/bGBZGEv

Right now you can have this working by replace `'oembed'` with `o-embed` in two spots in the media-embed module. Then import this module wherever you render `<o-embed>` tags.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #9375, opened 2021-03-26
MediaEmbed: Set custom element name

### LINKED ISSUE #9373: MediaEmbed: Rename <oembed> to <o-embed> for custom elements compatibility
I propose changing the name of `<oembed>` to write `<o-embed>` for forward compatibility with [custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) per #2737

[Media embed](https://ckeditor.com/docs/ckeditor5/latest/features/media-embed.html) saves `<oembed>`, but in the WhatWG specification requires a dash/hyphen ([1](https://html.spec.whatwg.org/#valid-custom-element-name), [2](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#high-level_view)):

> These requirements ensure a number of goals for valid custom element names:
> 
> They start with an ASCII lower alpha, ensuring that the HTML parser will treat them as tags instead of as text.
> 
> They do not contain any ASCII upper alphas, ensuring that the user agent can always treat HTML elements ASCII-case-insensitively.
> 
> **They contain a hyphen, used for namespacing and to ensure forward compatibility (since no elements will be added to HTML, SVG, or MathML with hyphen-containing local names in the future).**
> 
> They can always be created with createElement() and createElementNS(), which have restrictions that go beyond the parser's.
> 
> Apart from these restrictions, a large variety of names is allowed, to give maximum flexibility for use cases like <math-α> or <emotion-😍>.

This will permit systems like [LitElement](https://lit-element.polymer-project.org/), [Stencil](https://stenciljs.com/) and [others](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#libraries) to fill support for embeds.

## 📃 Other details

* Browser: Any
* OS: Any
* CKEditor version: 26
* Installed CKEditor plugins: [`MediaEmbed`](https://ckeditor.com/docs/ckeditor5/latest/features/media-embed.html)

## Ideas:
That change should be backward compatible so it doesn't break existing HTML.

As a first step, prepare `MediaEmbed` to handle different tags:

- An option such as `preferredElementName` can be set,
- An option like `elementNames` for tags that should be detected as media embeds

Later:
- An option like `migrateToPreferredTagOnChange` can be turned on to implicitly update existing tags to the preferred tags

```js
  mediaEmbed: {
    // default: ['oembed', 'o-embed']
    elementNames: ['oembed'],
    // default: 'oembed', in a later version, `o-embed` can be the default
    preferredElementName: `o-embed`,
    // default: false
    migrateToPreferredTagOnChange: true,
  }
```

---

tony:
See #9373 for details.

`<oembed>`  is an invalid name for custom elements ([WhatWG](https://html.spec.whatwg.org/#valid-custom-element-name), [MDN](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#high-level_view)).

Allow setting element name (especially to `<o-embed>`). Retain compatibility with existing tags.

Type: Message. Closes #9373

### Checklist

`MediaEmbed`
- [x] Add `preferredElementName` option (default: `'oembed'`)
- [x] Add `elementNames` option (default: `['oembed', 'o-embed']`)
- [x] Preserves existing behavior
  - [x] `elementNames` option: Default includes `'oembed'`: `['oembed', 'o-embed']`
  - [x] `preferredElementName` option: Default is `'oembed'`
- [x] Basic docs (Internal API)
- [x] Basic docs (Markdown docs) 
- [x] Tests

---

### Additional information

Also related: #2737 (Implement a web component to handle `<oembed url="...">`)

I have an `<o-embed>` custom element polyfill here: [Github](https://github.com/social-embed/social-embed
), [NPM](https://www.npmjs.com/package/@social-embed/wc) [CodeSandbox](https://codepen.io/attachment/pen/poRRwdy)

### DIFF
--- a/packages/ckeditor5-media-embed/docs/features/media-embed.md
+++ b/packages/ckeditor5-media-embed/docs/features/media-embed.md
@@ -101,6 +101,30 @@ By default, the media embed feature outputs semantic `<oembed url="...">` tags f
 </figure>
 ```
 
+Further customization of semantic data output can be done through the {@link module:media-embed/mediaembed~MediaEmbedConfig#preferredElementName `config.mediaEmbed.preferredElementName`} and {@link module:media-embed/mediaembed~MediaEmbedConfig#elementNames `config.mediaEmbed.elementNames`} options. As an example, if `preferredElementName` is set to `o-embed`:
+
+```html
+<figure class="media">
+	<o-embed url="https://media-url"></oembed>
+</figure>
+```
+
+And further, to be backward compatible with legacy semantic elements (any element using the `url` attribute, these can be passed via `elementNames`: `['oembed', 'o-embed']`. If there is a semantic tag for `<mytag url="..."></mytag>`, you can set `['oembed', 'o-embed', 'mytag']`. To remove support for tags, omit the tag name. To skip handling of `<oembed>` tags, `['oembed']`:
+
+```js
+ClassicEditor
+	.create( document.querySelector( '#editor' ), {
+		plugins: [ MediaEmbed, ... ],,
+		toolbar: [ 'mediaEmbed', ... ]
+		mediaEmbed: {
+			elementNames: ['oembed']
+		}
+	} )
+	.then( ... )
+	.catch( ... );
+```
+
+
 #### Including previews in data
 
 Optionally, by setting `mediaEmbed.previewsInData` to `true` you can configure the media embed feature to output media in the same way they look in the editor. So if the media element is "previewable", the media preview (HTML) is saved to the database:
--- a/packages/ckeditor5-media-embed/src/mediaembed.js
+++ b/packages/ckeditor5-media-embed/src/mediaembed.js
@@ -237,6 +237,44 @@ export default class MediaEmbed extends Plugin {
  */
 
 /**
+ * Customizing semantic element name.
+ *
+ * When `oembed` (default), the feature produces "semantic" data with tag `<oembed>`:
+ *
+ *		<figure class="media">
+ *			<oembed url="https://url"></oembed>
+ *		</figure>
+ *
+ * It can also be set to other element names, for instance, `o-embed` will produce:
+ *
+ *		<figure class="media">
+ *			<o-embed url="https://url"></oembed>
+ *		</figure>
+ *
+ * @member {String} [module:media-embed/mediaembed~MediaEmbedConfig#preferredElementName]
+ */
+
+/**
+ * Supporting legacy semantic element names.
+ *
+ * When `['oembed', 'o-embed']` (default), the feature renders "semantic" data for content with
+ * `<oembed>` an `<o-embed>` tags:
+ *
+ *		<figure class="media">
+ *			<oembed url="https://url"></oembed>
+ *		</figure>
+ *
+ *		<figure class="media">
+ *			<o-embed url="https://url"></oembed>
+ *		</figure>
+ *
+ * By default, the feature will render new media embeds via the option
+ * {@link module:media-embed/mediaembed~MediaEmbedConfig#preferredElementName `config.mediaEmbed.preferredElementName`}
+ *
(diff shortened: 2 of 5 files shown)


Reinmar:
Wow, this is super complete :) Thank you :heart\_eyes: 

I'll check who could review this on our side.

As for the solution – initially, I wasn't sure whether making this configurable is the right direction. If this feature's output is configurable this way then why not every other's? We need to be cautious in such cases.

However, I think we should ask ourselves why this particular feature got this PR and no other. I think the answer is that it uses a non-standard HTML output and hence there's no convention we're able to follow (BTW, I wouldn't go with `o-embed` too as why would the `o` be separated from the rest of the name? what about features that are undividable single-word?). In such case, making the output configurable makes more sense, even though it's always modifiable via custom converters.

One thing that I'd consider simplifying in the PR is whether exposing `elementNames` really makes sense. What if the feature only handles `oembed` + what it was configured to use in its output (config.mediaEmbed.elementName)? It's potentially less future-proof but also simpler.

Reinmar:
Wow, this is super complete :) Thank you :heart\_eyes: 

I'll check who could review this on our side.

As for the solution – initially, I wasn't sure whether making this configurable is the right direction. If this feature's output is configurable this way then why not every other's? We need to be cautious in such cases.

However, I think we should ask ourselves why this particular feature got this PR and no other. I think the answer is that it uses a non-standard HTML output and hence there's no convention we're able to follow (BTW, I wouldn't go with `o-embed` too as why would the `o` be separated from the rest of the name? what about features that are undividable single-word?). In such case, making the output configurable makes more sense, even though it's always modifiable via custom converters.

One thing that I'd consider simplifying in the PR is whether exposing `elementNames` really makes sense. What if the feature only handles `oembed` + what it was configured to use in its output (config.mediaEmbed.elementName)? It's potentially less future-proof but also simpler.

tony:
> What if the feature only handles oembed ...

`oembed` is not a valid element name ([WhatWG](https://html.spec.whatwg.org/#valid-custom-element-name), [MDN](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#high-level_view)). In my opinion, CKEditor's MediaEmbed is rendering invalid HTML.

If you try to build a web component with `oembed` instead of `o-embed` (or something hyphenated), here's what most linters will show:

```
./src/OEmbedElement.ts

    The tag name 'oembed' is not a valid custom element name. Remember that a hyphen (-) is required.
    26:  @customElement('oembed')
    no-invalid-tag-name
```

That's why my PR is opinionated toward phasing it out, as it's introducing URL that wouldn't work with custom elements.

> One thing that I'd consider simplifying in the PR is whether exposing elementNames really makes sense. What if the feature only handles oembed + what it was configured to use in its output (config.mediaEmbed.elementName)? It's potentially less future-proof but also simpler.

I will make a secondary PR without a public `elementNames` where `oembed` is implied.

The downside is `oembed` can't be omitted or phased out without a custom build. It's an edgecase, but I wanted the config to be able to do it without needing a specialized build/fork. From a user standpoint: I'm flexible of course! Generally I think making things available via config means not requiring a rebuild of the editor. That's a win for the user - at the expense of complication / (if taken too far, bloat) in the code itself.


If there's any renaming/bikeshedding - I'm happy to make adjustments! (e.g. renamed `preferredElementName` -> `elementName`)
IfIf there's

psmyrek:
I also think that solution for this issue could be simplified to a form, where media embed feature supports the existing `<oembed>` element plus the configurable one via `config.mediaEmbed.elementName`. I think there is no need to expose the list of supported view elements.

So lets move to your 2<sup>nd</sup> PR: https://github.com/ckeditor/ckeditor5/pull/9418. I'm closing this one.

tony:
This PR continues at #9418

PR #9376, opened 2021-03-26
Fix crash in renderer

bendemboski:
This loop was doing unsafe removals of DOM nodes while iterating over indices that point into a live `NodeList`, and was causing a crash in certain cases.

The error is:

```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at Function.from (<anonymous>)
    at Renderer._updateAttrs (webpack:///./packages/ckeditor5-engine/src/view/renderer.js?:19:11593)
    at Renderer.render (webpack:///./packages/ckeditor5-engine/src/view/renderer.js?:19:2390)
```

and it can happen when this loop ends up setting up an incorrect mapping from a view element to a text node, and then `_updateAttrs()` is called on that text node.

The factors that I've identified that will cause this to reproduce are:
* Enough nodes to make `_diffNodeLists()` use the fastDiff implementation
* A render involving at least two changes (e.g. removals) separated other nodes

### DIFF
--- a/packages/ckeditor5-engine/src/view/renderer.js
+++ b/packages/ckeditor5-engine/src/view/renderer.js
@@ -268,6 +268,12 @@ export default class Renderer {
 
 		if ( actions.indexOf( 'replace' ) !== -1 ) {
 			const counter = { equal: 0, insert: 0, delete: 0 };
+			// Removing nodes from the DOM as we iterate can cause `actualDomChildren`
+			// (which is a live-updating `NodeList`) to get out of sync with the
+			// indices that we compute as we iterate over `actions`, producing
+			// incorrect element mappings. So collect all the DOM nodes that need to
+			// be removed and remove them after the iteration is complete.
+			const toRemove = [];
 
 			for ( const action of actions ) {
 				if ( action === 'replace' ) {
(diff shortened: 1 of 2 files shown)


bendemboski:
As an aside, it seems to me that there's extra complexity/instability here because of the diff/fastDiff behavior. I can understand how with use cases like generating operations, it could make sense to try to use a more expensive algorithm that yields more precise results, but sacrifice that precision for speed when the size of the data set is large enough. There are multiple sets of operations that can describe the same change, but some might be qualitatively better or worse in terms of how clearly they describe the change, so it makes sense to aim for the better ones but settle for the worse ones when it will hurt performance.

But that doesn't seem to really make sense for the renderer -- there's only one correct result the renderer can produce, with no qualitative better/worse variants. Therefore for this all to work the renderer _must_ produce the correct result regardless of which diff algorithm is run, so why not always use the fast diff algorithm in the renderer? That might produce some very modest additional performance gains, but more importantly, would simplify the test matrix a lot, because you wouldn't have to worry about every single combination of fast diff vs slow diff across all the places in the renderer that do a diff.

I did some playing around, and if I hacked the diff algorithm to always use the fast diff, or to always use the slow diff, this bug did not reproduce. It looked to me like what was required was for the diff called via [this](https://github.com/ckeditor/ckeditor5/blob/3e276523c5c18bf6f29d3f7ced8d98b31518c511/packages/ckeditor5-engine/src/view/renderer.js#L266) method call to run the fast version, but then one or possibly both of the diffs called via the `_findReplaceActions()` call on the next line to use the slow version. Testing those various combinations seems like a headache, so I'm wondering why not use the fastDiff everywhere in the renderer?

Reinmar:
Hi @bendemboski! I asked @niegowski to review your PR as his probably the best person to go into Render's details :)

niegowski:
Thanks for your contribution!

PR #9386, opened 2021-03-30
Page break: make upcast converter less strict

### LINKED ISSUE #9385: Page break upcast converter is too strict

At the moment, the upcast converter is looking for:

1.  An empty element that has `page-break-before` or `page-break-after` style, or
2.  an element with `page-break-before` or `page-break-after` style which has exactly one child - `<span>` with `display:none` which has exactly one child - text node with a single space.

I don't understand why it was created this way. This is very strict, in my opinion. Also, it works only because DomConverter is kind-of bugged and leaks the space on upcasting. The space that is expected in the view is actually a block filler. When block fillers are properly removed (#9357) the page break stops working.

In my opinion, the upcast converter should simply look for an element with `page-break-before` or `page-break-after`. Maybe there are some exceptions which I am not aware of (paste from Word?).

However, for "let's be safe" reasons I can softer upcast converter requirements only by a little. So we will only drop the requirement for having exactly one-space-text-node.

scofalik:

Internal (page-break): Upcast converter will be less strict when checking if a view element should be converted to the model. Closes #9385.

### DIFF
--- a/packages/ckeditor5-page-break/src/pagebreakediting.js
+++ b/packages/ckeditor5-page-break/src/pagebreakediting.js
@@ -99,7 +99,7 @@ export default class PageBreakEditing extends Plugin {
 					if ( element.childCount == 1 ) {
 						const viewSpan = element.getChild( 0 );
 
-						// The child must be the "span" element that is not displayed and has a space inside.
+						// The child must be the "span" element that is not displayed.
 						if ( !viewSpan.is( 'element', 'span' ) || viewSpan.getStyle( 'display' ) != 'none' ) {
 							return;
 						}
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-page-break/src/pagebreakediting.js
```diff
@@ -100,13 +100,7 @@ export default class PageBreakEditing extends Plugin {
 						const viewSpan = element.getChild( 0 );
 
 						// The child must be the "span" element that is not displayed and has a space inside.
```
mlewand, line 102:
```suggestion
						// The child must be the "span" element that is not displayed.
```

mlewand (approved):
LGTM

`<* style="page-break-after: always"><span style="display: none"></span></*>` is unique enough that we don't have any further checks for children of the inner span.

PR #9390, opened 2021-03-30
Introduce the uploadComplete event in ImageUploadEditing

### LINKED ISSUE #5204: Support setting data attributes on image from upload adapters
Current `upload()` return
```
        return {
            default: 'image src',
            360: '...',
            600: '...', // etc.
        };
```

We currently have a situation where I would like to save the original path of the uploaded image, but the images have to be served through a CDN.

It would be nice if I could save the original path as a data attribute. I've had a look at https://github.com/ckeditor/ckeditor5-image/blob/8fa4acf7850e958526d0f049899f88cb91282a74/src/imageupload/imageuploadediting.js to find the relevant code and it comes down to the `_parseAndSetSrcsetAttributeOnImage` function which sets the srcset attributes for numeric keys.

If we could have support another `attributes` property where this information could be passed through, allowing the Custom Upload Adapter to supply any number of attributes to save, that would be very handy. Could be used to set the `sizes` attribute, for example, which this plugin currently does not do. 

Proposed `upload()` return
```
        return {
            default: 'image src',
            360: '...',
            600: '...', // etc.

            attributes: { 
               'data-original-path': 'image path',
            },
        };
```

My current work around is to use downcast/upcast to parse the `src` and reverse engineer the path that way, which is slightly annoying but not impossible, I suppose. Just feels easier to get it from the horse's mouth.

Thank you very much!

ma2ciek:

Feature (image): Introduced the `uploadComplete` event in `ImageUploadEditing` that allows customizing the image element (e.g. setting custom attributes) based on data retrieved from the upload adapter. Closes #5204.

Feature (upload): Upload adapters' `upload` method can now return a promise that resolves to an object with additional properties along with the `urls` array. See more in #5204.

MINOR BREAKING CHANGE (upload): The `SimpleUploadAdapter#upload()` returns now a promise that resolves to an object with normalized data including the `urls` array, which was only returned before. This may affect all integrations depending on the `SimpleUploadAdapter` uploading mechanism.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-image/src/imageupload/imageuploadediting.js
+++ b/packages/ckeditor5-image/src/imageupload/imageuploadediting.js
@@ -282,7 +282,7 @@ export default class ImageUploadEditing extends Plugin {
 					 *
 					 * 		const imageUploadEditing = editor.plugins.get( 'ImageUploadEditing );
 					 *
-					 * 		imageUploadEditing.on( 'uploadComplete', evt, { data, imageElement } => {
+					 * 		imageUploadEditing.on( 'uploadComplete', ( evt, { data, imageElement } ) => {
 					 * 			editor.model.change( writer => {
 					 * 				writer.setAttribute( 'someAttribute', 'foo', imageElement );
 					 * 			} );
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-image/src/imageupload/imageuploadediting.js
```diff
@@ -192,6 +195,13 @@ export default class ImageUploadEditing extends Plugin {
 				}
 			}
 		} );
+
+		// Set the default handler for feeding the image element with `src` and `srcset` attributes.
+		this.on( 'uploadComplete', ( evt, { imageElement, writer, data } ) => {
```
niegowski, line 200: I think that we should not pass `writer` in the event params, we could here a separate change block - those are `'transparent'` because it's sth that happened "in the background" so it should not be undoable.

packages/ckeditor5-image/src/imageupload/imageuploadediting.js
```diff
@@ -192,6 +195,13 @@ export default class ImageUploadEditing extends Plugin {
...
+			writer.setAttribute( 'src', data.default, imageElement );
+
```
niegowski, line 202: This empty line does not seem to be needed.

packages/ckeditor5-image/tests/imageupload/imageuploadediting.js
```diff
@@ -68,7 +69,7 @@ describe( 'ImageUploadEditing', () => {
 				viewDocument = view.document;
 
 				// Stub `view.scrollToTheSelection` as it will fail on VirtualTestEditor without DOM.
-				sinon.stub( view, 'scrollToTheSelection' ).callsFake( () => {} );
+				sinon.stub( view, 'scrollToTheSelection' ).callsFake( () => { } );
```
niegowski, line 72:
```suggestion
				sinon.stub( view, 'scrollToTheSelection' ).callsFake( () => {} );
```

niegowski (changes requested):

packages/ckeditor5-image/src/imageupload/imageuploadediting.js
ma2ciek: Tell me if I'm wrong, but then the integrator would need to know to wrap the code in the `transparent` change block. With the writer from the `transparent` change block, he doesn't need to know about it. But I'm fine with both approaches.

ma2ciek:
@niegowski, could you check it again?

PR #9392, opened 2021-03-31
Added support in crawler to detect not responding pages

psmyrek:

Internal: Added support in crawler to detect not responding pages. See https://github.com/cksource/ckeditor5-internal/issues/687.

Internal: All navigation timeouts on framework integration docs are now ignored. See https://github.com/cksource/ckeditor5-internal/issues/687.

---

### Additional information

The crawler waits for the `load` event to consider that the navigation was successful. Unfortunately, the `load` event may not be fired, when the page is still not fully loaded, which leads to a navigation timeout.

Now, the crawler is able to detect, whether the navigation error was caused by the page not loading at all (i.e. due to an infinite loop in the script), or whether the page is essentially up and running, but it is still waiting for some resources that won't interfere with page testing.

This PR ignores all navigation timeouts on framework integration pages in the docs, because external resources that are used there sometimes do not respond.

### Hint for reviewer

As I was unable to reproduce the case related to the lack of response to external resources in our docs, I came up with an easier way to simulate such a case. It is enough to add a special external resource to the page, which responds very slowly:

```html
<img src="https://en.js.cx/clipart/train.gif?speed=0&cache=0">
```

and then:
* run the crawler to make sure that navigation timeout is detected properly,
* add an exception:

  ```html
  <meta name="x-cke-crawler-ignore-patterns" content='{"navigation-error": "timeout"}'>
  ```
  and the crawler will ignore such an error.


scripts/web-crawler/index.js
```diff
@@ -539,6 +548,19 @@ function isNavigationRequest( request ) {
...
+/**
+ * Checks, if the page is not hung by trying to evaluate a function within the page context in defined time.
+ *
+ * @param {Object} page The page instance from Puppeteer.
+ * @returns {Promise.<Boolean>}
+ */
+async function isPageResponding( page ) {
+	return Promise.race( [
+		page.title(),
+		new Promise( ( resolve, reject ) => setTimeout( () => reject(), DEFAULT_RESPONSIVENESS_CHECK_TIMEOUT ) )
```
pomek, line 560:
```suggestion
		setTimeout( () => Promise.reject(), DEFAULT_RESPONSIVENESS_CHECK_TIMEOUT )
```

or (_magic version_) 

```suggestion
		setTimeout( Promise.reject.bind( Promise ), DEFAULT_RESPONSIVENESS_CHECK_TIMEOUT )
```


will do the same.

pomek (approved):
LGTM.

pomek:
I was able to reproduce the issue using your hints. Thanks for the note.

scripts/web-crawler/index.js
pomek: Yeah, I've just noticed it will not work because `Promise.race` expects an array of promises. Sorry for the confusion.

PR #9424, opened 2021-04-06
Docs: fixing toolbar tooltip for restricted editing.

### LINKED ISSUE #9280: Feature tour balloon is in the wrong place after switching to restricted mode
## Steps to reproduce
1. Go to https://ckeditor.com/docs/ckeditor5/latest/features/restricted-editing.html
2. Switch demo to restricted editing mode

## Actual behavior:
Balloon stays in wrong place, there's a warning in the console:
```
assets.js:4 [attachTourBalloon] The target DOM node for the feature tour balloon does not exist. 
{text: "Click to switch restricted editing mode."}
```

godai78:

Docs. Closes #9280.

---

### Additional information

Moving the toolbar tooltip to stay in place when switching between modes in restricted editing demo.

### DIFF
--- a/packages/ckeditor5-restricted-editing/docs/_snippets/features/restricted-editing.js
+++ b/packages/ckeditor5-restricted-editing/docs/_snippets/features/restricted-editing.js
@@ -39,9 +39,9 @@ async function startStandardEditingMode() {
 	await reloadEditor( {
 		removePlugins: [ 'RestrictedEditingMode' ],
 		toolbar: [
-			'heading', '|', 'bold', 'italic', 'link', '|',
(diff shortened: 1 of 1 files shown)


FilipTokarski:
Balloon works fine now 👍  I see that the warning is still there but that's not critical and I suppose it would require more time to fix, am I right?

godai78:
> I suppose it would require more time to fix

Yes. And - to be honest - more than I know about it, we would need a helping hand for that. But as a dirty hotfix, we may decide to merge this and address this problem in the future.

FilipTokarski (approved):

AnnaTomanek:
Added the estimate and milestone to the PR and left the original issue unassigned in the backlog.

PR #9428, opened 2021-04-07
Add `allowChildren` schema item definition property

### LINKED ISSUE #9261: Extend data schema item definition with allowChildren property

Currently, to allow additional model children, we have to go with 2 steps:

1. Register new model schema definition:
```js
editor.model.schema.register( 'custom-model', {
  allowIn: '$root'
} );
```
2. Extend existing model with `allowIn` property:
```js
editor.model.schema.extend( '$text', {
  allowIn: 'custom-model'
} );
```

This code is so often used (especially for `$text` model) that we could improve it by extending current [SchemaItemDefinition](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-SchemaItemDefinition.html) with additional member `allowChildren` which would simplify above code to:

```js
editor.model.schema.register( 'custom-model', {
  allowIn: '$root',
  allowChildren: '$text'
} );
```
Of course, it can be used for other model elements, the feature is not restricted to `$text` node only:

```js
editor.model.schema.register( 'custom-model', {
  allowIn: '$root',
  allowChildren: [ '$text', 'listItem' ]
} );
```

The new property should make sure if the children element is already registered or throw an error otherwise.

---

jacekbogdanski:

Features (engine): Introduced the `SchemaItemDefinition#allowChildren` property simplifying defining which other items are allowed inside this schema item definition. Closes #9261.

---

#### Missing definition

Some time ago we've discussed that in case of missing parent schema item definition, this feature should throw an error. However, looking at how we are currently handling such cases, it seems that it should rather ensure that the feature works correctly even if the item defined in `allowChildren` does not exist. As an example, take a look at `allowIn` unit tests:

https://github.com/ckeditor/ckeditor5/blob/c9ddd79b441fb358d7c1a65b1aa44a71dd29cafd/packages/ckeditor5-engine/tests/model/schema.js#L2744-L2751

Thanks to that we can still register definitions lazily without specific order:

```js
// We are registering paragraph before $root and without $text but this code will still work.
schema.register( 'paragraph', {
  isBlock: true,
  allowChildren: '$text'
} );

schema.register( '$root', {
  allowChildren: 'paragraph'
} );
```

We could still keep this behavior (lazy loading) by throwing errors during compilation, not definition registration. However, it will work differently than the current approach for other properties, so it probably makes more sense to avoid throwing errors in case of missing definitions and just skip it.

#### Refactoring

In some places, we are using the approach of extending the existing definition (mostly for `$text` item). As an example:

https://github.com/ckeditor/ckeditor5/blob/c9ddd79b441fb358d7c1a65b1aa44a71dd29cafd/packages/ckeditor5-engine/src/model/model.js#L107-L111

I'm wondering if it makes sense to refactor such places, as it could be rewritten to:

```js
this.schema.register( '$clipboardHolder', {
	allowContentOf: '$root',
	allowChildren: '$text',
	isLimit: true
} );
```
I didn't want to do it before the first review iteration to avoid unnecessary changes for now, but I suppose it would be nice to improve the code if we already have a bit nicer API for that. WDYT?


packages/ckeditor5-engine/tests/model/schema.js
```diff
@@ -2411,6 +2411,173 @@ describe( 'Schema', () => {
...
+				} );
+
+				schema.register( 'paragraph' );
+
+				expect( schema.getDefinition( '$root' ) ).to.deep.equal( {
+					allowAttributes: [],
+					allowIn: [],
+					name: '$root'
+				} );
+			} );
```
niegowski, line 2510: Maybe it could be useful to be able to see what elements might be children for some element?

niegowski (commented):
As commented below. I would also update the usages in our features. The guides should also be updated.

packages/ckeditor5-engine/tests/model/schema.js
jacekbogdanski: Seeing all allowed children for the specific definition looks very useful for _at least_ debugging, even if we don't have any real case scenario for now 👍🏻

jacekbogdanski:
I've added `allowChildren` to a compiled schema item definition, refactored code to use new property where it makes sense (didn't cover every possible unit test as in some cases it's more readable when actually the code is using a longer `extend` approach) and updated guide.

packages/ckeditor5-engine/docs/framework/guides/deep-dive/schema.md
```diff
@@ -17,7 +17,7 @@ Elements and attributes are checked by features separately by using the {@link m
 
 ## Defining allowed structures
 
-When a feature introduces a model element, it should register it in the schema. Besides defining that such an element may exist in the model, the feature also needs to define where this element can be placed:
+When a feature introduces a model element, it should register it in the schema. Besides defining that such an element may exist in the model, the feature also needs to define where this element can be placed. This information is provided by {@link module:engine/model/schema~SchemaItemDefinition#allowIn} property of the {@link module:engine/model/schema~SchemaItemDefinition}:
```
godai78, line 20:
```suggestion
When a feature introduces a model element, it should register it in the schema. Besides defining that such an element may exist in the model, the feature also needs to define where this element can be placed. This information is provided by the {@link module:engine/model/schema~SchemaItemDefinition#allowIn} property of the {@link module:engine/model/schema~SchemaItemDefinition}:
```

packages/ckeditor5-engine/docs/framework/guides/deep-dive/schema.md
```diff
@@ -37,14 +37,39 @@ In other words, this would be correct:
...
-```js
+```xml
 <$root>
 	<foo>
 		<myElement></myElement>
 	</foo>
 </$root>
 ```
 
+To declare what nodes could be inside the registered element, the {@link module:engine/model/schema~SchemaItemDefinition#allowChildren} property could be used:
```
godai78, line 48:
```suggestion
To declare which nodes could be inside the registered element, the {@link module:engine/model/schema~SchemaItemDefinition#allowChildren} property could be used:
```

packages/ckeditor5-engine/docs/framework/guides/deep-dive/schema.md
```diff
@@ -37,14 +37,39 @@ In other words, this would be correct:
...
+
+```xml
+<$root>
+	<myElement>
+		foobar
+	</myElement>
+</$root>
+```
+
+Both `{@link module:engine/model/schema~SchemaItemDefinition#allowIn}` and `{@link module:engine/model/schema~SchemaItemDefinition#allowChildren}` properties can be also inherited from other `SchemaItemDefinition` items.
```
godai78, line 67:
```suggestion
Both the `{@link module:engine/model/schema~SchemaItemDefinition#allowIn}` and `{@link module:engine/model/schema~SchemaItemDefinition#allowChildren}` properties can be also inherited from other `SchemaItemDefinition` items. 
```

godai78 (changes requested):

packages/ckeditor5-engine/src/model/schema.js
```diff
@@ -1654,6 +1693,16 @@ function cleanUpAllowIn( compiledDefinitions, itemName ) {
 	itemRule.allowIn = Array.from( new Set( existingItems ) );
 }
 
+// Restore allowChildren items based on allowIn.
+function cleanUpAllowChildren( compiledDefinitions, itemName ) {
```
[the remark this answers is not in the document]
jacekbogdanski: I was wondering about `setupAllowChildren` and `restoreAllowChildren` - I like `restore` a bit less but `setup` would be too close to `compile` and hard to distinguish two methods.

niegowski (approved):
👍  LGTM

godai78 (approved):
godai's seal of approval

PR #9440, opened 2021-04-08
Preserve markers on upcast.

### LINKED ISSUE #9402: Codeblock upcast converter prevents from loading markers inside codeblock

1.  Open an editor with comments or track changes feature.
2.  Create a code block.
3.  Create a comment or suggestion inside the code block.
4.  Save & load the editor data (e.g. call `editor.setData( editor.getData() )` ).

### ✔️ Expected result

Markers are retained.

### ❌ Actual result

Markers are gone.

## 📃 Other details

Additionally, it is impossible to upcast any "general" attributes that are set on `<code>` element /cc @Mgsy.

Both problem happen because on upcast, converter for `<pre>` takes care of everything and it consumes `<code>` element and does not fire conversion for its children. It is different than in image, where there is a converter for `<figure>` but it fires conversion for `` and `` itself has its own converter.

Additionally, the code block content is also converted in a custom way, to remove HTML tags. This is another blocker that prevents proper marker upcasting.

AFAIU (not tested), HTML coming from outside of the editor, which does not have `<pre>` will not be properly upcasted as well.

The solution here is to:

*   Provide converter for `<code>` that will mostly do what converter for `<pre>` is doing at the moment. Converter for `<pre>` should actually do almost nothing - see converter for `<figure>` in image.
*   Do not provide special conversion for text. If you want to "remove HTML", simply do that on model level with `Schema`. Actually, AFAICS, this is already done like that so I don't know why we have additional cleaning in the upcast converter. Maybe it is not needed?

maxbarnas:

Fix (code-block): Markers are now preserved on the upcast. Closes #9402.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-code-block/src/converters.js
+++ b/packages/ckeditor5-code-block/src/converters.js
@@ -148,7 +148,7 @@ export function dataViewToModelCodeBlockInsertion( editingView, languageDefs ) {
 			return;
 		}
 
-		// In case of nested `<pre>` we don't want to convert to another code block.
+		// In case of nested code blocks we don't want to convert to another code block.
 		if ( data.modelCursor.findAncestor( 'codeBlock' ) ) {
 			return;
 		}
--- a/packages/ckeditor5-code-block/src/converters.js
+++ b/packages/ckeditor5-code-block/src/converters.js
@@ -7,10 +7,7 @@
  * @module code-block/converters
  */
 
-import {
-	rawSnippetTextToModelDocumentFragment,
-	getPropertyAssociation
-} from './utils';
+import { getPropertyAssociation } from './utils';
 
 /**
  * A model-to-view (both editing and data) converter for the `codeBlock` element.
@@ -120,11 +117,11 @@ export function modelToDataViewSoftBreakInsertion( model ) {
  *
  * Sample input:
  *
- *		<pre><code class="language-javascript">foo();\nbar();</code></pre>
+ *		<pre><code class="language-javascript">foo();bar();</code></pre>
  *
  * Sample output:
  *
- *		<codeBlock language="javascript">foo();<softBreak></softBreak>bar();</codeBlock>
+ *		<codeBlock language="javascript">foo();bar();</codeBlock>
  *
  * @param {module:engine/view/view~View} editingView
  * @param {Array.<module:code-block/codeblock~CodeBlockLanguageDefinition>} languageDefs The normalized language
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-code-block/src/converters.js
```diff
@@ -144,21 +141,26 @@ export function dataViewToModelCodeBlockInsertion( editingView, languageDefs ) {
...
-		const viewChild = viewItem.getChild( 0 );
+		const viewCodeElement = data.viewItem;
+		const viewPreElement = viewCodeElement.parent;
+
+		if ( !viewPreElement || !viewPreElement.is( 'element', 'pre' ) ) {
+			return;
+		}
 
-		if ( !viewChild || !viewChild.is( 'element', 'code' ) ) {
+		// In case of nested `<pre>` we don't want to convert to another code block.
```
scofalik, line 151:
```suggestion
		// In case of nested code blocks we don't want to convert to another code block.
```

packages/ckeditor5-code-block/tests/codeblockediting.js
```diff
@@ -734,6 +735,26 @@ describe( 'CodeBlockEditing', () => {
 					return editor.destroy();
 				} );
 		} );
+
+		it( 'should convert markers inside pre > code', () => {
```
scofalik, line 739: Do we need this test? It looks the same as the test below (only it uses `markerToElement`).

packages/ckeditor5-code-block/tests/codeblockediting.js
```diff
@@ -948,7 +1013,53 @@ describe( 'CodeBlockEditing', () => {
...
+
+			const marker = model.markers.get( 'group:foo:id' );
+
+			expect( marker.getStart().path ).to.deep.equal( [ 0, 0 ] );
+			expect( marker.getEnd().path ).to.deep.equal( [ 0, 3 ] );
+
+			expect( getModelData( model ) ).to.equal( '<codeBlock language="plaintext">[]Bar</codeBlock>' );
+		} );
+
+		it( 'should preserve markers inside pre > code', () => {
```
scofalik, line 1041:
```suggestion
		it( 'should preserve markers on a code block', () => {
```

scofalik:
LGTM, see those two small comments I've made.

packages/ckeditor5-code-block/tests/codeblockediting.js
maxbarnas: Wanted to cover all pipelines as other tests do. For editing downcast the right method is the `markerToElement`, but for data downcast it's `markerToData`. From my point of view it was a totally different case but I might be wrong here.

Issue #7316, opened 2020-05-27
Upcast/DowncastWriters insert node methods are a mess


Reinmar:
### DowncastWriter

Right now you need to use `insert()` with `createPosition()`. There's no other method. That's super irritating if you want to create a structure of a couple elements.

### UpcastWriter

Same here... but the API is different. The APIs should be identical between both writers.

## Direction

I think that these writers should be identical to the model writer in as many details as possible. Not everything can be ported and the other problem will be backwards compat. But perhaps we can find a way to ensure that nothing breaks.

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

Reinmar:
cc @scofalik @jodator

jodator:
What always bugged me is:

1.  `modelWriter.insert( item, itemOrPosition, offset )`
2.  `viewWriter.insert( postion, nodeOrNodes )` (reversed, multiple items)

Also, the `UpcastWriter` is a poor brother of `DowncastWriter` - we left them alone.

But I think the problem is different IMO - why we have both? Do they differ that much that we have to have two similar implementations? Maybe there's a base `ViewWriter` for both? AFAICS their API is only different for

1.  creating elements (bare element in upcast vs attribute, container, etc in downcast)
2.  setting selection (downcast writer)
3.  manipulating view - move, break attributes, wrap, etc (downcast writer)

Other methods in both writes are just copied code.

scofalik:
At some point I was making those writers the same and AFAIR there was always some kind of inconsistency somewhere. I don't remember where, though.

Reinmar:
OMG, another issue:

```
for ( const child of viewContentFragment.getChildren() ) {
	viewWriter.insert( viewWriter.createPositionAt( viewContainer, 'end' ), child );
}
```

You have to go through the children manually.

jodator:
Would be nice to work this after unified enviroment setup.

jodator:
> You have to go through the children manually.

I've stumbled upon this again - writing from top of my head:

```javascript
writer.insert( childView, complexInfoBoxView, 'end' );
```

:point\_up: Using `model` writer `insert()` signature for a downcast writer...

niegowski:
There is also a nasty case of the model to view position mapping that in data pipeline is using the editing view to be able to create view positions and ranges: #9466

CKEditorBot:
[...]

PR #9445, opened 2021-04-09
The widget type around feature should work in the "Custom element conversion guide"

### LINKED ISSUE #9444: WTA does not work in the Custom element conversion guide

https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/conversion/custom-element-conversion.html#demo

### ✔️ Expected result

WTA is working and snippets are using the model to view position mapping instead of a model element mapped to multiple view elements.

### ❌ Actual result

The WTA is not working and the guide is not describing the model to view position mapping.

## 📃 Other details

*   Browser: …
*   OS: …

---

niegowski:

Docs (engine): The widget type around feature should work in the "Custom element conversion guide". Closes #9444.

---

### Additional information


packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
```diff
@@ -298,7 +363,12 @@ class InfoBox {
 			.add( dispatcher => dispatcher.on( 'insert:infoBox', editingDowncastConverter ) );
 		editor.conversion.for( 'dataDowncast' )
 			.add( dispatcher => dispatcher.on( 'insert:infoBox', dataDowncastConverter ) );
-	}
+
+        // Model to view position mapper is needed since the model <infoBox> content needs to end up in the inner
+        // <div class="info-box-content">.
+        editor.editing.mapper.on( 'modelToViewPosition', createModelToViewPositionMapper( editor.editing.view ) );
+        editor.data.mapper.on( 'modelToViewPosition', createModelToViewPositionMapper( editor.editing.view ) );
```
Reinmar, line 370: Passing `editor.editing.view`? I didn't check how it's used there but it's smelly.

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
```diff
@@ -274,6 +270,75 @@ editor.conversion.for( 'dataDowncast' )
 	.add( dispatcher => dispatcher.on( 'insert:infoBox', dataDowncastConverter ) );
 ```
 
+### The model to view position mapping
```
Reinmar, line 273: Don't we have to work on reverse direction mapping too? Or does it work out of the box? If so, it'd be good to mention that at the end of this section (e.g. in an info-box ... the Umberto's info-box :D)

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
niegowski: Yeah, I was talking with you about it some time ago (while refactoring the to-do lists conversion). It's used there only to be able to create a view range.

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
niegowski: The view to model works ok because it uses `findMappedViewAncestor()`. Can you add it?

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
Reinmar: Do we have a ticket for that? If yes, could you reference it from #7316? If not, please open it.

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
```diff
@@ -274,6 +270,75 @@ editor.conversion.for( 'dataDowncast' )
...
+
+After adding the above downcast converters we would end up with the following view structure:
+
+```
```
Reinmar, line 277: It'd help to show (maybe in a two-column table) what model translates to what view.

Also, what is the `[]` below for?

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
```diff
@@ -274,6 +270,75 @@ editor.conversion.for( 'dataDowncast' )
 	.add( dispatcher => dispatcher.on( 'insert:infoBox', dataDowncastConverter ) );
 ```
 
+### The model to view position mapping
+
+After adding the above downcast converters we would end up with the following view structure:
```
Reinmar, line 275:
```suggestion
The downcast converters shown in the previous section will not yet work correctly. This is how a given model would look after being downcasted:
```

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
Reinmar, line 273:
```suggestion
Due to the fact that the info box's view structure is much more complex than its model structure we need to take care of one additional aspect to make the  converters work &mdash; position mapping.

### The model to view position mapping
```

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
```diff
@@ -274,6 +270,75 @@ editor.conversion.for( 'dataDowncast' )
...
+<div class="info-box info-box-info">
+    []<p>
+        Foobar
+    </p>
+    <div class="info-box-title">Info</div>
+    <div class="info-box-content"></div>
+</div>
+```
+
+This is not what we need. We expect the content of the `<infoBox>` to be inside the `<div class="info-box-content">`. We have the downcast conversion for the `<infoBox>` and its own structure, but we need to specify where its content should land in the view. By the default it would be converted as a direct child of `<div class="info-box">` (as you can see in the above snippet) but it should go into the `<div class="info-box-content">`. To achieve this, we need to register a callback for the {@link module:engine/conversion/mapper~Mapper#event:modelToViewPosition `Mapper#modelToViewPosition` event}, so the positions inside the model `<infoBox>` element would map to the positions inside the `<div class="info-box-content">` view element.
```
Reinmar, line 287:
```suggestion
As you can see, this is not a correct view structure. The content of the model's `<infoBox>` element ended up directly inside the outer `<div>`. We expect the the `<infoBox>`'s content to be inside the `<div class="info-box-content">`.

We defined downcast conversion for `<infoBox>` itself, but we need to specify where its content should land in its view structure. By default, it is converted as direct children of `<div class="info-box">` (as you can see in the above snippet) but it should go into the `<div class="info-box-content">`. To achieve this, we need to register a callback for the {@link module:engine/conversion/mapper~Mapper#event:modelToViewPosition `Mapper#modelToViewPosition` event}, so positions inside the model `<infoBox>` element would map to positions inside the `<div class="info-box-content">` view element.
```

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
```diff
@@ -274,6 +270,75 @@ editor.conversion.for( 'dataDowncast' )
...
+
+```
```
Reinmar, line 289: If you'll convert the above snippet into a table, this one could be converted too. In any case, be consistent.

Also, I don't feel `[]` helps here.

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
```diff
@@ -274,6 +270,75 @@ editor.conversion.for( 'dataDowncast' )
...
+                                          <div class="info-box-title">Info</div>
+                                          <div class="info-box-content">
+    []<paragraph>               ->            []<p>
+        Foobar                  ->                Foobar
+    </paragraph>                ->            </p>
+                                          </div>
+</infoBox>                      ->    </div>
+```
+
+The callback that we need:
```
Reinmar, line 300:
```suggestion
Such mapping can be achieved by registering this callback to {@link XYZ} event:
```

packages/ckeditor5-engine/docs/_snippets/framework/extending-content-custom-element-converter.js
```diff
@@ -123,19 +128,51 @@ function insertViewElements( data, conversionApi, infoBox, infoBoxTitle, infoBox
...
-	// a part of the default binding.
-	conversionApi.mapper.bindElements( data.item, infoBoxContent );
 
 	conversionApi.writer.insert(
 		conversionApi.mapper.toViewPosition( data.range.start ),
 		infoBox
 	);
 }
 
+function createModelToViewPositionMapper( view ) {
```
Reinmar, line 140: The code here should be an exact copy of the code in the guide itself (comments are missing) for easier comparison and maintenance.

packages/ckeditor5-engine/docs/_snippets/framework/extending-content-custom-element-converter.js
```diff
@@ -123,19 +128,51 @@ function insertViewElements( data, conversionApi, infoBox, infoBoxTitle, infoBox
...
+
+		const viewElement = data.mapper.toViewElement( parent );
+		const viewContentElement = findContentViewElement( view, viewElement );
+
+		data.viewPosition = data.mapper.findPositionIn( viewContentElement, modelPosition.offset );
+	};
+}
+
+function findContentViewElement( editingView, viewElement ) {
+	for ( const value of editingView.createRangeIn( viewElement ) ) {
```
Reinmar, line 157: It's sad that we don't have an easier API to find a node inside another node :( Also worth reporting in the domain:dx, squad:dx (although, we may have it reported already).

Reinmar:
I reviewed the guide in terms of readability. The code itself looks good too, but I didn't verify whether it works. So, cc @ckeditor/qa-team

LukaszGudel:
I've checked demo editor on this guide and it's looking good for me. I am able to type around those widgets using buttons, horizontal caret or keyboard shortcuts. I've also checked some nested structures with boxes and tables and it also works correctly :+1:

packages/ckeditor5-engine/docs/_snippets/framework/extending-content-custom-element-converter.js
niegowski: #9465

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
niegowski: I have no idea how to represent it as a table. I guess we could go with the proposed snippet and tune it later.

packages/ckeditor5-engine/docs/framework/guides/deep-dive/custom-element-conversion.md
niegowski: #9466

niegowski:
Extracted to #9467 to merge directly to `stable`.

PR #9459, opened 2021-04-13
I/9213 enabling caption toolbar

### LINKED ISSUE #9213: Disabled buttons in the image toolbar.
After resolving the #9136, the image toolbar remains visible when the selection is placed inside the image caption. Still, all the buttons (except the caption button) in the toolbar are disabled.

It makes sense to enable the commands bound to these buttons to make the toolbar fully functional.

magda-chrzescian:

Fix (image): The buttons in the image toolbar should be enabled when the selection is in a caption. Closes #9213.
Fix (image): Dropdowns in the image toolbar should be disabled when all of the nested buttons are disabled. See #9213.

---

We actually don't need the [getSelectedImageWidget](https://ckeditor.com/docs/ckeditor5/latest/api/module_image_image_utils.html#function-getSelectedImageWidget) function anymore. Whenever we are looking for an image on which to execute a command, we are also looking for the image ancestor of the selected caption. So I replaced the `getSelectedImageWidget` with `getSelectedOrAncestorImageWidget`.

It was a public function, so I'm not sure if it's ok to change its name and behavior. But because of the changes in managing the toolbar state, our integrator probably also should start using the `getSelectedOrAncestorImageWidget` function.

BTW, should such renaming be considered a breaking change?


packages/ckeditor5-image/tests/image/utils.js
```diff
@@ -890,4 +892,71 @@ describe( 'image widget utils', () => {
...
+			expect( getImageModelElementAncestor( model.document.selection ) ).to.be.null;
+		} );
+
+		it( 'should return an imageInline element if it is selected', () => {
+			setModelData( model, '<paragraph>[<imageInline></imageInline>]</paragraph>' );
+
+			const image = getImageModelElementAncestor( model.document.selection );
+
+			expect( image ).to.be.instanceOf( ModelElement );
+			expect( image.name ).to.equal( 'imageInline' );
```
oleq, line 932: `expect( image.is( 'element', 'imageInline' ) ).to.be.true` is enough to check this. The same applies to all tests that follow :point\_down:

packages/ckeditor5-image/src/imagestyle/imagestyleui.js
```diff
@@ -150,6 +150,9 @@ export default class ImageStyleUI extends Plugin {
 				}
 			} );
 
+			dropdownView.bind( 'isEnabled' )
+				.toMany( buttonViews, 'isEnabled', ( ...areEnabled ) => !!areEnabled.find( isEnabled => isEnabled ) );
```
oleq, line 154:
```suggestion
				.toMany( buttonViews, 'isEnabled', ( ...areEnabled ) => areEnabled.some( isEnabled => isEnabled ) );
```

packages/ckeditor5-image/tests/imagestyle/imagestyleui.js
```diff
@@ -440,6 +440,46 @@ describe( 'ImageStyleUI', () => {
...
+				const activeButton = items.first;
+
+				activeButton.isEnabled = true;
+
+				return { ...group, activeButton };
+			} );
+
+			for ( const { buttonView } of groups ) {
+				expect( buttonView.isEnabled ).to.be.true;
+			}
```
oleq, line 463: This could be way simpler (why `map()` and return `activeButton`?):

```suggestion
			for ( const group of groups ) {
				const groupItems = group.view.toolbarView.items;

				for ( const item of groupItems ) {
					item.isEnabled = false;
				}

				groupItems.first.isEnabled = true;
			}

			for ( const { buttonView, config } of groups ) {
				expect( buttonView.isEnabled, `Failing group name: "${ config.name }"` ).to.be.true;
			}
```

packages/ckeditor5-image/tests/imagestyle/imagestyleui.js
```diff
@@ -440,6 +440,46 @@ describe( 'ImageStyleUI', () => {
...
+
+					return item;
+				} );
+
+				return group;
+			} );
+
+			for ( const { buttonView } of groups ) {
+				expect( buttonView.isEnabled ).to.be.false;
+			}
```
oleq, line 481:
```suggestion
			for ( const group of groups ) {
				const groupItems = group.view.toolbarView.items;

				for ( const item of groupItems ) {
					item.isEnabled = false;
				}
			}

			for ( const { buttonView, config } of groups ) {
				expect( buttonView.isEnabled, `Failing group name: "${ config.name }"` ).to.be.false;
			}
```

oleq (changes requested):
Looks good 👍

My biggest concern, though, is why was the text alternative button left disabled when the selection is the caption? Is there any reason for that?

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -79,6 +79,19 @@ export function getImageWidgetAncestor( selection ) {
...
+ * Returns a image model element if one is among the selection's ancestors.
+ *
+ * @param {module:engine/model/selection~Selection|module:engine/model/documentselection~DocumentSelection} selection
+ * @returns {module:engine/model/element~Element|null}
+ */
+
+export function getImageModelElementAncestor( selection ) {
+	const selectedElement = selection.getSelectedElement();
+
+	return isImage( selectedElement ) ? selectedElement : selection.getFirstPosition().findAncestor( 'image' );
```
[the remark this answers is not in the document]
magda-chrzescian: It seems to me that the model element `findAncestor` method accepts the element name only: https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_element-Element.html#function-findAncestor. Unless I'm missing something.

packages/ckeditor5-image/tests/imagestyle/imagestyleui.js
magda-chrzescian: I didn't want to mutate the `groups` and mess up the following tests. I missed the fact that each of these tests has its own `beforeEach` and thus separate `groups` array. Thanks for the catch :)

magda-chrzescian:
> My biggest concern, though, is why was the text alternative button left disabled when the selection is the caption? Is there any reason for that?

Yes, my scatterbraining. I fixed it for the text alternative and image resize button.
I kept the imageUpload button the way it was (disabled). Theoretically, it can be placed in the image toolbar, but having it enabled with the caret inside the caption would suggest that it is possible to insert the image in there.

packages/ckeditor5-image/src/image/utils.js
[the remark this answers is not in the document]
oleq: You're right. I must've mistaken it with some other helper. Sorry for the confusion.

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -65,7 +65,7 @@ export function getSelectedImageWidgetOrAncestor( selection ) {
  * @param {module:engine/view/selection~Selection|module:engine/view/documentselection~DocumentSelection} selection
  * @returns {module:engine/view/element~Element|null}
  */
-export function getSelectedImageWidgetAncestor( selection ) {
+export function getSelectionAncestorImageWidget( selection ) {
```
oleq, line 68: Looks like this helper is only used internally by `getSelectedOrAncestorImageWidget()`. I think we should merge them.

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -86,7 +86,7 @@ export function getSelectedImageWidgetAncestor( selection ) {
  * @returns {module:engine/model/element~Element|null}
  */
 
-export function getSelectedImageElementOrAncestor( selection ) {
+export function getSelectedOrAncestorImageElement( selection ) {
```
oleq, line 89: What about `getClosestSelectedImageElement()`?

packages/ckeditor5-image/src/image/utils.js
```diff
@@ -49,14 +49,14 @@ export function isImageWidget( viewElement ) {
  * @param {module:engine/view/selection~Selection|module:engine/view/documentselection~DocumentSelection} selection
  * @returns {module:engine/view/element~Element|null}
  */
-export function getSelectedImageWidgetOrAncestor( selection ) {
+export function getSelectedOrAncestorImageWidget( selection ) {
```
oleq, line 52: What about `getClosestSelectedImageWidget()`?

packages/ckeditor5-image/src/image/utils.js
magda-chrzescian: I'm not sure, the image element returned by this function doesn't have to be formally selected (`<image><caption>[Foo]</caption></image>`), so wouldn't this name be misleading?
However, I don't like the name above either. I was thinking about something like `getCorrelatedImageElement` - a bit enigmatic, but still understandable IMO.

magda-chrzescian:
@oleq could you also refer to the following?

> We actually don't need the getSelectedImageWidget function anymore. Whenever we are looking for an image on which to execute a command, we are also looking for the image ancestor of the selected caption. So I replaced the getSelectedImageWidget with getSelectedOrAncestorImageWidget.
> 
> It was a public function, so I'm not sure if it's ok to change its name and behavior. But because of the changes in managing the toolbar state, our integrator probably also should start using the getSelectedOrAncestorImageWidget function.
> 
> BTW, should such renaming be considered a breaking change?

packages/ckeditor5-image/src/image/utils.js
oleq: Maybe then simply `getClosestImageElement()` (`getClosestImageWidget()`)?

oleq:
> We actually don't need the getSelectedImageWidget function anymore. Whenever we are looking for an image on which to execute a command, we are also looking for the image ancestor of the selected caption. So I replaced the getSelectedImageWidget with getSelectedOrAncestorImageWidget.
> 
> It was a public function, so I'm not sure if it's ok to change its name and behavior. But because of the changes in managing the toolbar state, our integrator probably also should start using the getSelectedOrAncestorImageWidget function.
> 
> BTW, should such renaming be considered a breaking change?

Since we're doing tons of breaking changes in this feature branch, I see nothing wrong with removing unused helpers and simplifying the names of the confusing ones.

packages/ckeditor5-image/src/image/utils.js
magda-chrzescian: I liked the previous version better :) Changed to `getClosestSelectedImageElement`.

PR #9461, opened 2021-04-13
Added max width to the left and right aligned images, added border co…

### LINKED ISSUE #9342: Add sensible max-width to left/right block styles

ATM if an image is larger than the editing root, applying one of "left/right aligned image" styles does not change anything visually. With `max-width: 90%` there would be at least some feedback to the user confirming the successful application of the style.

---

magda-chrzescian:

Other (image): The side-aligned images should always have some `min-width` property to not take the whole editor width. Closes #9342.

---

### Additional information

Besides that, the floating image coming directly after a paragraph should have its top margin collapsed with the bottom margin of the paragraph.


packages/ckeditor5-image/theme/imagestyle.css
```diff
@@ -48,6 +55,13 @@
 		margin-right: auto;
 	}
 
+	/* Simulates margin collapsing with the preceding paragraph, which does not work for the floating elements. */
+	& p + .image-style-align-left,
+	& p + .image-style-align-right
+	& p + .image-style-side, {
```
oleq, line 61: This won't transpile to correct CSS.

oleq: Why does it not apply to `image-style-block-*` BTW? I'm trying to understand the use case.

oleq (changes requested):
Minor issue found + a question.

packages/ckeditor5-image/theme/imagestyle.css
magda-chrzescian: It took me a while to catch this comma.

It doesn't have to be applied to the block images because they don't have the `float` property, so their margins collapse properly with the preceding paragraph's margins.
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing

PR #9473, opened 2021-04-14
Add plugins metadata to packages

### LINKED ISSUE #9400: Move plugins metadata from OB to packages
### Scope of this ticket:

*   Make a final proposal for the metadata format and verify whether putting that into package.json is a good idea (taking into account that we'll have quite a lot of data there).
*   Move the data from OB to each package and update it to the new format (excludes HTML format - it'll be covered separately).
*   We don't need to scan existing plugins currently. We can base on what's available in OB right now.

---

\[from meeting notes\]

### Which plugins do we cover?

*   Only the glue code. No reason to cover \*Editing, \*UI
*   We need to cover soft-deps of other plugins (e.g. CloudServices)

### Plugins data

*   Essential data
    *   plugin name
    *   human-readable plugin name
    *   description
    *   link to a feature guide
    *   path to the plugin
    *   additional toolbars
    *   requires (soft and other weird requirements)
    *   UI components:
        *   name
        *   path to SVG
            *   what about plugins that use icons from other package?
            *   relative if local
            *   or npm's path: `@ckeditor/ckeditor5-core/theme/icons/...`
        *   type (button, dropdown, split button)
        *   additional info (e.g. dropdown config - with or without text/icon?)
        *   target toolbar
            *   if defined, OB adds it automatically to this toolbar
        *   if not, the user can
*   Online builder data
    *   metadata of all packages + source SVG files
    *   exclusions
        *   what about exclusion groups? e.g. upload adapters
        *   gets more complicated e.g. in the case of RTE plugins
    *   which sample
    *   whether to show a certain plugin (e.g. Paragraph, Essentials)
    *   "requiresServer"
    *   private → premium

### The default configuration of builds

*   duplicates the src/ckeditor.js file of each build
*   might be reasonable to write a parser to exclude `builtinPlugins` and `defaultConfig`

### LINKED ISSUE #9401: For each major plugin specify the HTML output and add it to the package matadata
Extracted from #6642. 

### Scope

*   Try to document a couple of more "interesting" plugins with the proposed (or similar) format.
*   Review the outcome of the above.
*   Once we have the final decision, document other features.

### Notes

Before documenting every plugin output, let's do it for a few to check if the proposal works.

> HTML output
> 
> *   To not repeat ourselves - plugins are configurable and so is their output. Also, plugins influence one another (e.g. alignment + block features).
> *   Conclusions:
>     *   It'd be enough if CKE5 documented in its docs (in one place) all the possible outputs of all the features
>     *   CKE5 could try to standardize where this information can be written down by 3rd party plugin authors

Current proposal:

```javascript
"ckeditor": {
    "plugins": [
      { 
        "name": "Alignment",
        "output": [
          {
            "element": "$block",
            "attributes": [
              {
                "name": "style",
                "description": "For default config.", // Consider renaming to "comments"

                // Possible future extension - list values/particular styles
              },
              {
                "name": "class",
                "description": "For config.alignment.classes", // Consider renaming to "comments"

                // Possible future extension - list particular classes
              }
            ]
          }
        ]
      },
      { 
        "name": "Image",
        "output": [
          {
            "element": "figure",
            "attributes": [
              {
                "name": "class",
                "value": "image"
              }
            ]
          },
          {
            "element": "img",
            "attributes": [ "src", "srcset", "sizes" ]
          }
        ]
      },
```

ma2ciek:

Internal: Add plugins metadata. Closes #9400. Closes #9401.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-adapter-ckfinder/ckeditor5-metadata.json
+++ b/packages/ckeditor5-adapter-ckfinder/ckeditor5-metadata.json
@@ -1,0 +1,11 @@
+{
+	"plugins": [
+		{
+			"name": "CKFinder Upload Adapter",
+			"className": "CKFinderUploadAdapter",
+			"path": "src/uploadadapter.js",
+			"description": "This package implements a CKEditor 5 upload adapter compatible with the CKFinder file manager and uploader's server–side connector.",
+			"docs": "features/image-upload/image-upload.html#ckfinder"
+		}
+	]
+}
--- a/packages/ckeditor5-alignment/ckeditor5-metadata.json
+++ b/packages/ckeditor5-alignment/ckeditor5-metadata.json
@@ -22,7 +22,7 @@
 				{
 					"elements": "$block",
 					"classes": "*",
-					"_comment": "If class names are defined in `config.alignment.options`, classes instead of inline styles are used for alignment."
+					"_comment": "If class names are defined in `config.alignment.options`, then these classes are used for alignment instead of inline styles."
 				}
 			]
 		}
--- a/packages/ckeditor5-alignment/ckeditor5-metadata.json
+++ b/packages/ckeditor5-alignment/ckeditor5-metadata.json
@@ -12,6 +12,18 @@
 					"name": "alignment",
 					"pathToIcon": "TODO"
 				}
+			],
+			"htmlOutput": [
+				{
+					"elements": "$block",
+					"styles": "text-align",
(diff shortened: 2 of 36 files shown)


psmyrek (approved):
LGTM

ma2ciek:
I improved some comments. The whole specification of the HTML output looks good to me.

ma2ciek:
I'll merge this PR to the main feature branch (`i/9400`).

PR #9488, opened 2021-04-16
Docs: fixed code block in Markdown output demo.

### LINKED ISSUE #9487: Code block broken in Markdown guide demo
It can be reproduced in the docs on production: https://ckeditor.com/docs/ckeditor5/latest/features/markdown.html

Output:

godai78:

Docs (markdown-gfm): Fixed demo display error. Closes #9487.

### DIFF
--- a/packages/ckeditor5-markdown-gfm/docs/_snippets/features/markdown.html
+++ b/packages/ckeditor5-markdown-gfm/docs/_snippets/features/markdown.html
@@ -28,6 +28,7 @@
 		<ul>
 			<li>Numbered lists</li>
 			<li>Bulleted lists</li>
+			<li>To-do lists</li>
 		</ul>
 	</li>
 </ul>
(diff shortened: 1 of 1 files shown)


godai78:
We do, in the code blocks demo here: [https://ckeditor.com/docs/ckeditor5/latest/features/code-blocks.html](https://ckeditor.com/docs/ckeditor5/latest/features/code-blocks.html)

The answer to "why" is "because it works this way" as opposed to `<div>`, which does not.

AnnaTomanek:
Makes sense, thanks!

AnnaTomanek (approved):

Mgsy:
Indeed, the snippet doesn't work if a source element is `<div>`. However, changing the snippet to markdown, instead of HTML, fixes the problem and `<div>` could be used:

````markdown
```css
p {
	text-align: center;
	color: red;
}
```
````

I guess that's a bug in the Markdown data processor.

Mgsy:
IMO that's fine to leave it with `<textarea>`, as there is a bug in the Markdown processor that can't parse this `<pre><code></code></pre>` structure properly. I'll report a follow-up.

PR #9494, opened 2021-04-16
#9399: Removed the Image plugin dependency from the EasyImage plugin

### LINKED ISSUE #9399: EasyImage should not depend on Image (after introducing ImageBlock and ImageInline)

When you use `EasyImage` and load either `ImageBlock` or `ImageInline` (but not `Image`) there is a soft requirement error in the console because `Image` is missing.

The code of the `EasyImage` plugin does not use anything from `@ckeditor/ckeditor5-image` so this is a virtual requirement. Most likely we made `Image` the requirement of `EasyImage` to improve DX so developers don't have to include both if they want to use our service. The intention was good but it backfired when the API has changed.

The solution is getting rid of the requirement (breaking change) and mentioning in the migration guide that `EasyImage` must be used together with `Image`.

---

oleq:

Other (easy-image): Removed the Image plugin dependency from the EasyImage plugin. Closes #9399.

---

The migration guide was updated in the dedicated branch a37e166.


oleq:
As it turned out, the `ImageUploadEditing` also required some changes, namely moving schema extension to `afterInit()`. Previously, when `EasyImage` required `Image`, it guaranteed that `Image` loads `Image(Block|Inline)Editing` first and they, in turn, register their model elements in the schema. The `ImageUpload` (also required by `EasyImage`) would always init after the model elements in schema were defined.

Now, with the requirements loosened there's nothing that would guarantee that `Image(Block|Inline)Editing` initializes before `ImageUpload` so the latter can safely extend the schema. Hence, changes in `ImageUploadEditing` to postpone the schema extension.

niegowski (approved):

PR #9511, opened 2021-04-19
documentselection: fix jsdoc for getters

fedemp:


pomek:
Again, thank you, @fedemp.

PR #9516, opened 2021-04-19
utils: fix return type for getBoxSidesShorthandValue

fedemp:


packages/ckeditor5-engine/src/view/styles/utils.js
```diff
@@ -220,7 +220,7 @@ export function getBoxSidesValueReducer( styleShorthand ) {
  *		// will return '1px 1px 2px'
  *
  * @param {module:engine/view/stylesmap~BoxSides} styleShorthand
- * @returns {Function}
+ * @returns {string}
```
mlewand, line 223:
```suggestion
 * @returns {String}
```

mlewand (approved):
LGTM

PR #9518, opened 2021-04-20
Docs: changelog 27.1.0

godai78:

Docs: changelog review.


godai78:
We may reconsider linking to migration guide and link to docs section directly instead.

pomek:
Will we have the migration guide to` 27.1.0`?

godai78:
Yes, it is already in the docs. It mentions disallowing back the nesting of tables and block quotes.

pomek (approved):
LGTM.

AnnaTomanek:
As for migration guides, I've been thinking we could actually change their structure a bit and have them as "Migration to v27.X" or something to save the space because normally, there will be no migration guide for any minor version or if we have it, it's gonna be very short (because if minor, there shouldn't be that many things to describe). WDYT?

godai78:
Like, putting all those (two) guides together? Sure.

AnnaTomanek (approved):

pomek (approved):
LGTM.

PR #9525, opened 2021-04-20
model/range fix @returns for fromJSON

fedemp:


pomek (approved):
LGTM.

pomek:
Thanks, @fedemp.

PR #9526, opened 2021-04-20
#9524: Selected inline widgets wrapped in an attribute in the view should create a fake selection

### LINKED ISSUE #9521: Linked image resize handlers do not refresh when converting block image to inline

---

### LINKED ISSUE #9524: Selected inline widgets wrapped in an attribute in the view should create fake selection
A generalization of https://github.com/ckeditor/ckeditor5/issues/9521.

ATM when a linked inline image gets selected, the selection is `[<a><span class="ck-widget"></span></a>]`. This is how engine works, selecting a linked text also results in `[<a>text</a>]`. 

This causes numerous issues with the logic depending on view selection `getSelectedElement()` (where `<a>` is returned instead of the widget). The main issue is that the selection is not fake, which affects accessibility and who-know-what-else – the widget plugin does not recognize as `[<a><span class="ck-widget"></span></a>]` a selected widget.

oleq:

Fix (widget): Selected inline widgets wrapped in an attribute in the view should create a fake selection. Closes #9524. Closes #9521.

### DIFF
--- a/packages/ckeditor5-widget/tests/widget.js
+++ b/packages/ckeditor5-widget/tests/widget.js
@@ -59,7 +59,7 @@ describe( 'Widget', () => {
 				} );
 				model.schema.extend( '$text', {
 					allowIn: [ 'nested', 'editable' ],
-					allowAttributes: [ 'attr' ]
+					allowAttributes: [ 'attr', 'bttr' ]
 				} );
 				model.schema.register( 'editable', {
 					allowIn: [ 'widget', '$root' ]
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-widget/tests/widget.js
```diff
@@ -228,6 +236,51 @@ describe( 'Widget', () => {
 		expect( viewDocument.selection.isFake ).to.be.true;
 	} );
 
+	it( 'should apply fake view selection when an inline widget is surrounded by an attribute element', () => {
```
niegowski, line 239: Please add test for deeper nested `AttributeElements`.

oleq: There you go 6fe4964.

niegowski (approved):
👍  LGTM

PR #9546, opened 2021-04-22
model/range.js add @returns to getWalker

fedemp:


mlewand (approved):
Nice catch, thanks for that! 👍

PR #9553, opened 2021-04-22
jsdoc model/liveposition fromPosition should return LivePosition

fedemp:


pomek (approved):
LGTM. Let's wait for CI.

pomek:
Thanks, @fedemp.

PR #9588, opened 2021-04-28
Update htmlembed.css

### LINKED ISSUE #9486: HTML embed preview text centered
Why does `.raw-html-embed__preview-content` have `text-align: center`? Click [here](https://github.com/ckeditor/ckeditor5/blob/bce8267e16fccb25448b4c68acc3bf54336aa087/packages/ckeditor5-theme-lark/theme/ckeditor5-html-embed/htmlembed.css#L145) to see.

nfplee:
Other (html-embed): The preview content for the HTML Embed plugin should not be centered. Closes #9486.

MINOR BREAKING CHANGE (html-embed): The preview content will not be centered anymore.

### DIFF
--- a/packages/ckeditor5-theme-lark/theme/ckeditor5-html-embed/htmlembed.css
+++ b/packages/ckeditor5-theme-lark/theme/ckeditor5-html-embed/htmlembed.css
@@ -142,7 +142,6 @@
 
 	& .raw-html-embed__preview-content {
 		box-sizing: border-box;
-		text-align: center;
 		background-color: var(--ck-color-base-foreground);
 
 		& > * {


Reinmar:
Hi and thanks! We need to verify whether it doesn't actually serve some unobvious reason, but git blame points to a bigger refactor, so it's hard to tell.

Please keep in mind that you can always override it with your own stylesheet.

nfplee:
> Hi and thanks! We need to verify whether it doesn't actually serve some unobvious reason, but git blame points to a bigger refactor, so it's hard to tell.
> 
> Please keep in mind that you can always override it with your own stylesheet.

Thanks, that's what I'm doing for now but I don't get why it's applied. It's not much of a preview if it's being displayed differently to what the user implied.

pomek:
The `text-align:center` was introduced in this commit – https://github.com/ckeditor/ckeditor5/commit/6fd46caadfb1eef726a49e358ebb60c127caba91 – by me. I do not recall any reason why I did it. I'd say it was a temporary solution and I forgot to revert the change.

Reinmar:
@pomek Could you check whether the docs look good after removing this line? If not, let's add this rule there (in the docs only). If yes, let's maybe merge this (with a minor BC). Finally, if you'll still have any doubts, let's drop the change as there's nothing particularly wrong with that style.

pomek:
@Reinmar, this change affects a manual test only. The documentation remains untouched. I'll merge it.

PR #9590, opened 2021-04-29
The table should not save the caption in the model attribute

### LINKED ISSUE #9543: The table should not save the caption in the model attribute to prevent issues in collaboration

Extracted from #9368. The solution should be based on #9503.

---

maxbarnas:

Fix (table): The table should not save the caption in the model attribute to prevent issues in collaboration. Closes #9543.

---

### Additional information


packages/ckeditor5-table/tests/tablecaption/toggletablecaptioncommand.js
```diff
@@ -237,6 +276,22 @@ describe( 'ToggleTableCaptionCommand', () => {
...
+			setData( model,
+				'<table>' +
+					'<tableRow>' +
+						'<tableCell>' +
+							'<paragraph></paragraph>' +
+						'</tableCell>' +
+					'</tableRow>' +
+					'<caption></caption>' +
+				'</table>'
+			);
```
niegowski: This would override the model and the original issue would not happen. You should modify the model by removing the caption content.

PR #9598, opened 2021-04-30
I/9568 resizing linked images

### LINKED ISSUE #9568: Linking image results in resize value NaN%
## Steps to reproduce:
1. Upload an image as inline
2. Add a link on the image
3. Try to resize it

## Result:
You cannot resize it, size value is `NaN%`, you also cannot drag it anywhere in Firefox. Sometimes it's hard to reproduce it, but with those steps I got it every time.

## Screencast:

https://user-images.githubusercontent.com/34380544/115735574-e0f68a00-a38a-11eb-9272-0695056f6956.mp4

magda-chrzescian:

Fix (image): It should be possible to resize a linked image with the image handles. Closes #9568.
Fix (image): It should be possible to resize an image nested in a to-do list. See #9568.

---

### Additional information


oleq (changes requested):
Looks great 👍 

We're missing a "real-life-scenario test", though. Well, as real as `resizerMouseSimulator.dragTo()` will allow. 

Tests you added check what `._getResizeHost()` would return (which is fine). Still, if the mechanics of the feature changes and `_getResizeHost` logic is gone in the future (and its tests...) the code will be vulnerable again. A real-life (user) scenario test will prevent that because it is immune to the changes under the hood.

PR #9604, opened 2021-04-30
Hide toolbars when starts dragging the content

### LINKED ISSUE #9566: Image toolbar overlapping drop cursor
Image toolbar being visible all the time can become a bit problematic with inline images, especially when you for example want to drag an image into the line above the current one. Screencast contains dragging a table as a comparison:

https://user-images.githubusercontent.com/34380544/115722931-c0750280-a37f-11eb-8c1a-20deaf43aa4f.mp4

pomek:

Fix (clipboard): Hide all toolbars when the user starts dragging the widget and show them when the drag ends. Closes #9566.

---

https://user-images.githubusercontent.com/2270764/116685778-2f231300-a9b3-11eb-9fec-69a5d7ca89b6.mov


FilipTokarski:
Works fine now👌

niegowski:
There is a problem in Chrome and Firefox that the `dragend` event is not fired if the original node got removed from the document while dragging. This works ok for widgets because they are not modified, but if you drag out some text content to some other editable area the event is not fired ever and the toolbars stay disabled until you start dragging some widget again:

pomek:
It does not sound good. However, it's easy to fix - let's disable the `WidgetToolbarRepository` plugin only if the widget element is dragged.

pomek:
Should be good now.

niegowski:
I feel like it should be the other way around. The DragDrop plugin should disable the widget toolbar.

pomek:
Even fewer changes.

niegowski (approved):

PR #9616, opened 2021-05-04
Executing the min size of the block image properly.

### LINKED ISSUE #9300: Block images smaller than 50px render incorrectly in table cells
A regression after https://github.com/ckeditor/ckeditor5/pull/9173.

magda-chrzescian:

Fix (image): The block image smaller than 50px should be resized to the minimum size. Closes #9300 .


oleq (changes requested):
It looks OK but it breaks the resizer logic



I guess we'd need to replicate the same 10% logic on the JS side.

magda-chrzescian:
@oleq It actually didn't break it, this is the way it used to work. At least on the inline images branch, even stranger things happen on master ;) 
image
It escalates when you set the block image size to a very little value and then convert it to the inline image - it doesn't have the `min-size`, so the image size changes significantly.
I was about to add a separate issue to address this problem, but I guess I could fix it in this PR. WDYT?

oleq:
> I was about to add a separate issue to address this problem, but I guess I could fix it in this PR. WDYT?

I see no good reason to postpone the fix you proposed and mix it up with anything else. Let's have another issue with that. Can you create one?

PR #9624, opened 2021-05-04
Change marker-to-data logic to improve conversion inside document fragment

### LINKED ISSUE #9622: Change marker-to-data logic to improve conversion inside document fragment

Marker-to-data conversion converts marker boundaries to either an element or and attribute on an element. The decision is based on whether text is allowed at given place. So, for example, we have:

```
<p>Foo<comment-start name="..."></comment-start>bar</p><img data-comment-end-after="..." ... />
```

The problem is in that document fragment allows everything inside, to enable both conversion of bigger structures with multiple elements on top-level as well as simply a fragment of a text.

However, since document fragment allows text, markers are always converted to elements. This is problematic for two reasons:

1.  For the same model, the output is simply different than what `editor.getData()` returns, which may be problematic.
2.  There is at least one specific problem: the marker element is placed inside `<ul>` but upon upcast, all non-allowed elements are cleared from `<ul>`. (BTW this mechanism most probably is not needed and I will create additional issue for that).

scofalik:

Other (engine): In marker-to-data conversion, attributes for marker boundaries will be used every time the marker starts or ends before or after a model element, instead only where text is not allowed by model schema. Closes #9622.

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/downcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/downcasthelpers.js
@@ -405,9 +405,8 @@ export default class DowncastHelpers extends ConversionHelpers {
 	 *
 	 * This conversion creates a representation for model marker boundaries in the view:
 	 *
-	 * * If the marker boundary is at a position where text nodes are allowed, then a view element with the specified tag name
-	 * and `name` attribute is added at this position.
-	 * * In other cases, a specified attribute is set on a view element that is before or after the marker boundary.
+	 * * If the marker boundary is before or after a model element, a view attribute is set on a corresponding view element.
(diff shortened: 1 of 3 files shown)


niegowski (approved):

scofalik:
Or it could be

```
[<elementAfter>...</elementAfter> -> <elementAfter data-group-start-before="...">...</elementAfter>
<parent>]<elementAfter> -> <parent>]<elementAfter data-group-end-before="...">
<elementBefore>...</elementBefore>] -> <elementBefore data-group-end-after="...">...</elementBefore>
</elementBefore>[</parent> -> </elementBefore data-group-start-after="..."></parent>
```

PR #9629, opened 2021-05-05
Script that uses package metadata files to generate "Features overview" page content

### LINKED ISSUE #9414: Implement a generator for "Features output" documentation page
Extracted from #6642.

This generator should scan the metadata and try to print them out in a human-readable form in the docs.

We should not forget about 3rd paty plugins currently covered in our docs.

As for the integration with Umberto - see how error codes (also generated automatically) are embedded.

To be decided:

*   A la content styles (once every release) - more performant
*   or a la error pages (dynamically on every build) - slower

psmyrek:

Internal: Added script that uses package metadata files to generate "Features overview" page content. Closes #9414.

---

### Additional information

To build the features overview, run `yarn docs:features-overview`. Packages are sorted alphabetically in the output table.
The generated features overview output is stored in `Builds → Guides → Integration → Features overview` page.

features-overview


psmyrek:
CI is failing, because it does not contain the `external` folder, so built docs do not have features from INT or CF. Links generated by the script from this PR, that point to plugins from INT or CF, are invalid from Umberto point of view 🤔

ma2ciek:
Overall it looks quite good to me. My major concern is the lack of space for the default 760px width of the table content:





However, I'm not sure how we could achieve more space here with these three columns.

Also, I'm still not convinced about duplicating the plugin name. It's hard to determine which link is which without some explicit information. It just looks a bit strange to me.

> `<... class`

Should not we go with `<*` syntax? As the `*` is also used in other places? 



This is partially true, partially not. As this plugin is required internally the `Comments` plugin, we could define the output here as well.

psmyrek:
@ma2ciek, thanks for comments 👍🏻 

I also have doubts about the width of this third, most important column, because it looks just poor on small monitors.

Regarding the `<… class` or `<… style` syntax: I was wondering how to indicate, that a given class or inline style only applies to HTML elements, that are listed above **in the same cell**. Replacing the `…` with `*` is IMHO inappropriate, because `*` means **any value**, but `…` in this context means the name of the HTML element from that particular table cell.

I forgot to include HTML output from packages from CF and INT. Now these packages are also included in this PR.

ma2ciek:
> Regarding the `<… class` or `<… style` syntax: I was wondering how to indicate, that a given class or inline style only applies to HTML elements, that are listed above **in the same cell**. Replacing the `…` with `*` is IMHO inappropriate, because `*` means **any value**, but `…` in this context means the name of the HTML element from that particular table cell. Maybe this should be explicitly described in this guide.

Ok, so a description for the `...`  syntax would be helpful in this case. 

> I forgot to include HTML output from packages from CF and INT. Now these packages are also included in this PR.

pomek:
> Replacing the … with * is IMHO inappropriate, because * means any value, but … in this context means the name of the HTML element from that particular table cell. Maybe this should be explicitly described in this guide.

I'd expect that `<* ...>` means any of the existing HTML elements. But if we will explain that `...` means somewhere below the table, it also would be good.

psmyrek:
Added description for the ellipsis `…` character in the `<… class` syntax.

psmyrek:
> 
> 
> This is partially true, partially not. As this plugin is required internally the `Comments` plugin, we could define the output here as well.

I thought that if a given plugin does not generate or alter HTML elements by itself, we do not define its output in this case. HTML output for `Comments` plugin is already defined.

pomek:
A question from the release point of view – when should this guide be updated? Before preparing TP or before releasing packages? Content styles are updated directly before calling `npm publish`.

psmyrek:
> 
> 
> A question from the release point of view – when should this guide be updated? Before preparing TP or before releasing packages? Content styles are updated directly before calling `npm publish`.

IMHO before preparing TP to give others the opportunity to briefly check, if generated features overview for upcoming release contains what is expected there.

Since the links generated by this script will not be validated by Umberto (to make CI pass), I wonder if there should be a note, that after generating new features overview, docs should be checked by the web crawler locally. Crawler by default skips API links, so to make it scan all links in the docs, the following command can be used: `yarn docs:verify -e /assets/` (on Windows: `yarn docs:verify -e //assets//`) to override the default exclusion list. Scanning all links in the docs takes around 15 minutes on my machine.

ma2ciek:
I guess that we should also include the `WProofreader`, `MathType` and `ChemType` plugins here. It's a bit tricky as these are 3rd-party plugins. And the list of 3-party plugins can grow with time.

pomek:
One of the step when preparing TP is:

image

If we prepare the Features overview guide, it will be scanned by the crawler. Is it fine?

psmyrek:
> If we prepare the Features overview guide, it will be scanned by the crawler. Is it fine?

I'm not sure. The `yarn run docs:verify` will scan links to features, but crawler will skip links to API docs. I wonder if we should override the default exclusions and scan also all API links.

scripts/docs/build-features-overview.js
```diff
@@ -0,0 +1,456 @@
...
+ * @param {String} [options.prefix] A string to add as a prefix to provided value. Empty string by default.
+ * @param {String} [options.suffix] A string to add as a suffix to provided value. Empty string by default.
+ * @returns {Function}
+ */
+function wrapBy( { prefix = '', suffix = '' } = {} ) {
+	return item => `${ prefix }${ item }${ suffix }`;
+}
+
+/**
+ * @typedef {Object.<String, String|Array.<String>>} HtmlOutput
```
ma2ciek, line 420: I think that you can specify the base type to just `Object` in all these typedefs. `Object.<String, String>` and similar are a bit misleading as they suggest that the object behaves like a key-value dictionary, while it only accepts known properties.

psmyrek: Good idea 👍🏻

ma2ciek:
As the `MathType` and `Wproofreader` are installed locally (we use them for building docs, manual tests, etc.) we can add HTML output for these features as well as this should be quite easy to achieve. Perhaps we could keep a list of all 3rd-party plugins in a JSON file.

scripts/docs/build-features-overview.js
```diff
@@ -0,0 +1,456 @@
...
+function prepareApiLink( packageName, plugin ) {
+	const shortPackageName = packageName.replace( /^ckeditor5-/g, '' );
+
+	const packagePath = plugin.path
+		.replace( /(^src\/)|(\.js$)/g, '' )
+		.replace( /\//g, '_' );
+
+	const link = `../../../api/module_${ shortPackageName }_${ packagePath }-${ plugin.className }.html`;
+
+	return `<a href="${ link }" data-skip-validation><code class="nowrap">${ plugin.className }</code></a>`;
```
ma2ciek, line 217: Ideally, we could skip validation only for CF and internal packages.

psmyrek: Done. The `data-skip-validation` attribute is now added only to links from `external` folder.

psmyrek:
I added `ckeditor5-metadata.json` for 3<sup>rd</sup> party packages: `MathType` and `WProofreader`. I did not provide the full list of [MathML](https://developer.mozilla.org/en-US/docs/Web/MathML) elements and attributes (there are dozens of them), but a link where to find them instead.

Also I removed the `white-space: nowrap` from package name and package class name, to force the "HTML output" column a little wider on small screens.

scripts/docs/features-overview/build-features-overview.js
```diff
@@ -0,0 +1,491 @@
...
+
+/* eslint-env node */
+
+const fs = require( 'fs' );
+const path = require( 'path' );
+const glob = require( 'glob' );
+const chalk = require( 'chalk' );
+
+const DESTINATION_DOCS_PATH = 'docs/builds/guides/integration/features-overview.md';
+const PACKAGE_METADATA_RELATIVE_PATHS = [ 'packages', 'external/*/packages', 'scripts/docs/features-overview/3rd-party-packages' ];
```
ma2ciek, line 14: We need to primarily try to get the 3rd-party packages from node\_modules, as they are installed locally and they will provide their metadata file there in the future.

And then we should fall back to the 3rd-party-packages data that we maintain locally (and hopefully we will be able to remove them soon).

pomek:
One thing – the script prepares the content, but it does not commit it. It would be good to note it at the beginning of the `*.js` file or... add the commit option. I guess we can copy the code from the content-styles generator.

https://github.com/ckeditor/ckeditor5/blob/f6e48bb1495e276456f317a3df79d986bb9415a8/scripts/docs/build-content-styles.js#L235-L242

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+	console.log( `✨ ${ chalk.green( `The features HTML output from ${ numberOfPackages } packages has been generated successfully.` ) }` );
+
+	const shouldCommitChanges = process.argv.includes( '--commit' );
+
+	if ( shouldCommitChanges ) {
+		commitChanges();
+	}
+} catch ( error ) {
+	console.log( `❌ ${ chalk.red( 'An error occurred during parsing a package metadata file.' ) }` );
+	console.log( error );
```
ma2ciek, line 29: You miss here exiting with non-0 error code. Now, this script would pass even after the failure.

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+ * @returns {Number}
+ */
+function parseMetadataFiles() {
+	const parsedFiles = parseFiles()
+		.map( packageMetadata => {
+			const outputRows = packageMetadata.plugins
+				.map( plugin => {
+					const numberOfRowsPerPlugin = plugin.htmlOutput.length;
+
+					const pluginNameRowspan = numberOfRowsPerPlugin > 1 ?
```
ma2ciek, line 76: `cellRowspan`? I think that this variable is not connected with the plugin name at all.

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+ * ┃ last plugin  │ output #1 for the last plugin  ┃
+ * ┃              ├────────────────────────────────┨
+ * ┃              ┄                                ┄
+ * ┃              ├────────────────────────────────┨
+ * ┃              │ output #N for the last plugin  ┃
+ * ┗━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
+ *
+ * Generated table is preceded by the package name as a heading and the link to a source package metadata file on GitHub.
+ *
+ * @returns {Number}
```
ma2ciek, line 67: I think that it would be better if this function would return the HTML output markup (or the markup and parsed files).

Maybe it would be good to rename it a bit to `generateHtmlOutputMarkup`? WDYT? For me the `parseMetadataFiles` is a bit misleading as this function does much more and parsing the metadata files is not actually the main job here.

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+ * @property {String} name Plugin name.
+ * @property {String} className Plugin class name.
+ * @property {String} docs An absolute or relative URL to the plugin's documentation.
+ * @property {String} path A path to the file, relative to the metadata file, that exports the plugin.
+ * @property {HtmlOutput} htmlOutput An array of objects, that defines all possible HTML elements which can be created by a given plugin.
+ */
+
+/**
+ * @typedef {Object} ParsedPlugin
+ * @property {String} name Plugin name.
```
ma2ciek, line 545: This actually isn't a plugin name. But for me, it's quite hard to define such property names better :sweat\_smile:

ma2ciek: But the `pluginNameMarkup` would be a bit better IMO. And I'd use `htmlOutputMarkup` as well as this is the markup representation of the plugins output.

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+ */
+
+/**
+ * @typedef {Object} Package
+ * @property {String} name Package name.
+ * @property {String} isExternal Determines, if a given package is considered as external one in the context of the CKEditor 5. If package
+ * is not created by CKSource, then isExternal = false. Otherwise, it informs if it belongs to a CKEditor 5 repo (isExternal = false), or if
+ * it comes from an external folder: from Collaboration Features or Internal (isExternal = true).
+ * @property {String} isThirdParty Determines whether a given package has been created outside CKSource. A third-party package is not
+ * considered as external one.
```
ma2ciek, line 558:
```suggestion
 * @property {String} isExternal Determines if a given package comes from a CKEditor 5 external project like Collaboration Features or CKEditor 5 Internal.
 * It is set to `false` for 3rd-party packages.
 * @property {String} isThirdParty Determines whether a given package has been created outside the CKEditor 5 ecosystem. A third-party package is not
 * considered as the external one.
```

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+
+try {
+	const numberOfPackages = parseMetadataFiles();
+
+	console.log( `✨ ${ chalk.green( `The features HTML output from ${ numberOfPackages } packages has been generated successfully.` ) }` );
+
+	const shouldCommitChanges = process.argv.includes( '--commit' );
+
+	if ( shouldCommitChanges ) {
+		commitChanges();
```
ma2ciek, line 25: Maybe we could assert if the repository is clean before generating the HTML output page when the `commit` option is set to `true`?

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+const chalk = require( 'chalk' );
+const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
+
+const DESTINATION_DOCS_PATH = 'docs/builds/guides/integration/features-html-output-overview.md';
+const THIRD_PARTY_PACKAGES_LOCAL_DIR = 'scripts/docs/features-html-output/third-party-packages';
+
+try {
+	const numberOfPackages = parseMetadataFiles();
+
+	console.log( `✨ ${ chalk.green( `The features HTML output from ${ numberOfPackages } packages has been generated successfully.` ) }` );
```
ma2ciek, line 20:
```suggestion
	console.log( `✨ ${ chalk.green( `The features HTML output page has been generated successfully for ${ numberOfPackages } packages.` ) }` );
```

scripts/docs/features-html-output/build-features-html-output.js
ma2ciek: the `parsedFiles` could be also returned here.

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+	const plugins = preparePlugins( packageData, metadata.plugins );
+
+	return {
+		packageName,
+		plugins
+	};
+}
+
+/**
+ * Parses all plugins from package metadata file.
```
ma2ciek, line 213:
```suggestion
 * Parses all plugins from package metadata file and generates the HTML output markup for each plugin.
```

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+ *
+ * @param {Package} packageData Package properties.
+ * @param {Array.<Plugin>} plugins Plugins to parse.
+ * @returns {Array.<ParsedPlugin>}
+ */
+function preparePlugins( packageData, plugins = [] ) {
+	return plugins
+		.map( plugin => {
+			const pluginName = prepareFeatureLink( packageData, plugin );
```
ma2ciek, line 222: `pluginNameLink`?

scripts/docs/features-html-output/build-features-html-output.js
```diff
@@ -0,0 +1,571 @@
...
+
+			const pluginClassName = prepareApiLink( packageData, plugin );
```
ma2ciek, line 224: `pluginClassNameLink`?

ma2ciek (commented):
I reported some minor issues, but overall it looks pretty good 👍

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Done.

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Changed to `createHtmlOutputMarkup()` to be aligned with naming to other `create...()` functions.

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Done.

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Done.

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Done.

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Done.

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Done.

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Done.

scripts/docs/features-html-output/build-features-html-output.js
psmyrek: Done.

psmyrek:
Changes have been added. Feel free to change the other names as you wish 🙂

ma2ciek:
LGTM :+1:

PR #9649, opened 2021-05-07
Delete content should not exclude block widget at the end of the deletion range

niegowski:

Fix (engine): The `model.deleteContent()` should not exclude a block widget at the end of the deletion range.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/model/utils/deletecontent.js
+++ b/packages/ckeditor5-engine/src/model/utils/deletecontent.js
@@ -148,7 +148,13 @@ function getLivePositionsForSelectedBlocks( range ) {
 			// This is how modifySelection works and here we are making use of it.
 			model.modifySelection( selection, { direction: 'backward' } );
 
-			endPosition = selection.getLastPosition();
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-engine/src/model/utils/deletecontent.js
```diff
@@ -148,7 +148,13 @@ function getLivePositionsForSelectedBlocks( range ) {
 			// This is how modifySelection works and here we are making use of it.
```
scofalik, line 148: This, earlier and later parts of code would greatly benefit from comments showing what is going on / what is described as model (wrote in HTML-fashion).

scofalik: For example, I thought that it will be like this:

Before modify selection:

```
<paragraph>A[</paragraph><image></image><paragraph>]B</paragraph>
```

After:

```
<paragraph>A[]</paragraph><image></image><paragraph>B</paragraph>
```

PR #9674, opened 2021-05-11
Fix (engine): Don't strip non-Javascript `<script>` tags from HTML when converting to DOM

### LINKED ISSUE #9659: Lone HTML <script> tags are stripped on paste

I discovered this problem while debugging https://github.com/isaul32/ckeditor5-math/issues/35. The plugin makes it easy to observe the incorrect behavior:
1. Open https://jsfiddle.net/isaul32/qktj9h7x/
2. Attempt to copy and paste the first or second equation, with or without surrounding text. The operation will succeed
3. Attempt to copy and paste the third or fourth equation, along with surrounding text. The operation will succeed.
4. Attempt to copy and paste the third or fourth equation, without surrounding text. Nothing will be pasted.

The issue is that when content is pasted, [`_setupPasteDrop()`
](https://github.com/ckeditor/ckeditor5/blob/v27.1.0/packages/ckeditor5-clipboard/src/clipboardpipeline.js#L148) from [`ClipboardPipeline`](https://github.com/ckeditor/ckeditor5/blob/v27.1.0/packages/ckeditor5-clipboard/src/clipboardpipeline.js)  processes the content with [`HtmlDataProcessor`](https://github.com/ckeditor/ckeditor5/blob/v27.1.0/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js)'s [`toView()`](https://github.com/ckeditor/ckeditor5/blob/6e44d34449a95c93c01170a079bdd3c75c3eab5f/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js#L74), which itself calls [`_toDom()`](https://github.com/ckeditor/ckeditor5/blob/6e44d34449a95c93c01170a079bdd3c75c3eab5f/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js#L119), which processes the pasted HTML via a `DOMParser`:
```
_toDom( data ) {
	const document = this._domParser.parseFromString( data, 'text/html' );
	const fragment = document.createDocumentFragment();
	const nodes = document.body.childNodes;

	while ( nodes.length > 0 ) {
		fragment.appendChild( nodes[ 0 ] );
	}

	return fragment;
}
```

`_toDom()` only retains nodes in the `body` of `DOMParser.parseFromString()`'s response. However, when DOMParser encounters certain tags, like `<script>`, that are alone and not surrounded by text, it places these tags in the `<head>` and not the `<body>`. Therefore, if plugins (like the aforementioned math plugin) use `<script>` tags to store view elements, those elements can be stripped on paste.

### ✔️ Expected result

All HTML tags are retained when content is pasted into CKEditor.

### ❌ Actual result

Lone `<script>` tags are stripped upon pasting into CKEditor.

## 📃 Other details

* Browser: Chromium, Firefox
* OS: Linux
* First affected CKEditor version: ?
* Installed CKEditor plugins: Clipboard, https://github.com/isaul32/ckeditor5-math (for example)

---

Jules-Bertholet:
Closes #9659


Jules-Bertholet:
There seems to be an issue with the manual tests, `sh ./scripts/check-manual-tests.sh -r ckeditor5 -f ckeditor5` fails on `master` as well

Jules-Bertholet:
https://user-images.githubusercontent.com/79955405/117917774-cfccd900-b2b7-11eb-93bb-960c2ecd7b0b.mp4

Video demonstrating that the issue is fixed

PR #9675, opened 2021-05-11
The image style command should be augmentable

niegowski:

Internal (image): Added post-fixer for the `imageStyle` attribute to ensure a valid value for a given image type.

---

### Additional information


magda-chrzescian (approved):
Looks good, works fine :) I assume it will be tested with CF separately, so I tested it only in cke5 without CF context.

PR #9692, opened 2021-05-13
i/5585: Added support for RTL toolbar keyboard navigation

### LINKED ISSUE #5585: [A11y][RTL] Toolbar navigation with the keyboard works in the wrong direction

1. Open a classic editor in Arabic,
2. (Fn+)Alt+F10 to focus the toolbar,
3. Press the left arrow

### ✔️ Expected result

* Focus moves (visually) left, like, for instance in v4.

### ❌ Actual result

* Focus moves right because the DOM order is used not the visual order.

## 📃 Other details

master

---

cc @mlewand

mateuszzagorski:

Fix: Added support for handling toolbar keyboard navigation with **RTL** editor.  
  
Additionally added 1 editor instance to manual tests - to enable testing the scenario where there is **RTL** content with **LTR UI** (similar to the examples in the docs).  
  
Closes #5585.

---


mlewand (changes requested):
When we discussed it initially I thought that reversing up/down arrow handling should be changed too. But during the review I realized that up and down arrows should remain unchanged - changing from LTR to RTL language means that only handling of **horizontal** axis should be mirrored.

Up and down arrows are vertical axis. LTR and RTL languages "see" up and down the same way (you don't read text from bottom to top).

So let's make `isRtl` flag to affect only left / right arrow.

mlewand (approved):
LGTM 👍

PR #9694, opened 2021-05-13
Improve engine view matcher with new pattern syntax

### LINKED ISSUE #9872: Improve Matcher to allow matching attribute keys by regexp and unify API

[MatcherPattern](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_matcher-MatcherPattern.html) should be extended with an option to match attribute key using RegExp. We should also unify patterns between attributes, styles, and classes, so the API is easier to use. Proposed changes using `attributes` example (they will:

```js
// Match view element with any attribute value.
const pattern = {
	name: 'p',
	attributes: true
};

// Match view element which has matching attributes (String).
const pattern = {
	name: 'figure',
	attributes: 'title' // Match title attribute (can be empty).
};

// Match view element which has matching attributes (RegExp).
const pattern = {
	name: 'figure',
	attributes: /^data-.*$/ // Match attributes starting with `data-` e.g. `data-foo` with any value (can be empty).
};

// Match view element which has matching attributes (Object).
const pattern = {
	name: 'figure',
	attributes: {
		title: 'foobar',           // Match `title` attribute with 'foobar' value.
		alt: true,                 // Match `alt` attribute with any value (can be empty).
		'data-type': /^(jpg|png)$/ // Match `data-type` attribute with `jpg` or `png` value.
	}
};

// Match view element which has matching attributes (Array).
const pattern = {
	name: 'figure',
	attributes: [
		'title',    // Match `title` attribute (can be empty).
		/^data-*$/, // Match attributes starting with `data-` e.g. `data-foo` with any value (can be empty).
	]
};

// Match view element which has matching attributes (key-value-pairs).
const pattern = {
	name: 'input',
	attributes: [
		{
			key: 'type',                     // Match `type` as an attribute key.
			value: /^(text|number|date)$/ }, // Match `text`, `number` or `date` values.
		{
			key: /^data-.*$/,                // Match attributes starting with `data-` e.g. `data-foo`.
			value: true                      // Match any value (can be empty).
		}
	]
};

```

Some of the above patterns are already working but are not unified between attributes, styles, and classes. This improvement should also cover these cases.

---

maxbarnas:

Feature (engine): Improved engine view matcher with new pattern syntax allowing to match attribute keys using regular expressions. Unified pattern syntax between attributes, styles, and classes. Closes #9872.

Feature (engine): Added special `expand` option to `StylesMap.getStyleNames()` and view `Element.getStyleNames()` methods allowing to expand shorthand style properties.

---

### Additional information

* Adding missing test suite to `StylesProcessor` will be handled in the follow-up: https://github.com/ckeditor/ckeditor5/issues/9811.
* Exclusion of `style` and `class` attributes is reverted back to original form. There is a follow-up, where we'll decide how to approach this: https://github.com/ckeditor/ckeditor5/issues/9813.

### DIFF
--- a/packages/ckeditor5-engine/src/view/matcher.js
+++ b/packages/ckeditor5-engine/src/view/matcher.js
@@ -462,7 +462,8 @@ function matchStyles( patterns, element ) {
  *		const pattern = /^p/;
  *
  * If `MatcherPattern` is given as an `Object`, all the object's properties will be matched with view element properties.
- * Let's examine available pattern options:
+ * If the view element won't meet all of the object's pattern properties, the match won't happen.
+ * Available `Object` matching properties:
  *
  * Matching view element:
  *
@@ -515,7 +516,7 @@ function matchStyles( patterns, element ) {
  *			attributes: [
  *				'title',    // Match `title` attribute (can be empty).
  *				/^data-*$/, // Match attributes starting with `data-` e.g. `data-foo` with any value (can be empty).
-*			]
+ *			]
  *		};
  *
  *		// Match view element which has matching attributes (key-value pairs).
@@ -528,10 +529,10 @@ function matchStyles( patterns, element ) {
  *				{
  *					key: /^data-.*$/,                // Match attributes starting with `data-` e.g. `data-foo`.
  *					value: true                      // Match any value (can be empty).
-*				}
-*			]
+ *				}
+ *			]
  *		};
-*
+ *
  * Matching view element styles:
  *
  *		// Match view element with any style.
@@ -539,7 +540,7 @@ function matchStyles( patterns, element ) {
  *			name: 'p',
  *			styles: true
  *		};
-*
+ *
  *		// Match view element which has matching styles (String).
  *		const pattern = {
  *			name: 'p',
@@ -568,7 +569,7 @@ function matchStyles( patterns, element ) {
  *			attributes: [
  *				'color',      // Match `color` with any value.
  *				/^border.*$/, // Match all border properties.
-*			]
+ *			]
  *		};
  *
  *		// Match view element which has matching styles (key-value pairs).
@@ -581,8 +582,8 @@ function matchStyles( patterns, element ) {
  *				{
  *					key: /^border.*$/, // Match any border style.
  *					value: true        // Match any value.
-*				}
-*			]
+ *				}
+ *			]
  *		};
  *
  * Matching view element classes:
@@ -592,7 +593,7 @@ function matchStyles( patterns, element ) {
  *			name: 'p',
  *			classes: true
  *		};
-*
+ *
  *		// Match view element which has matching class (String).
  *		const pattern = {
  *			name: 'p',
@@ -633,8 +634,8 @@ function matchStyles( patterns, element ) {
  *				{
  *					key: /^image-side-(left|right)$/, // Match `image-side-left` or `image-side-right` class.
  *					value: true
-*				}
-*			]
+ *				}
+ *			]
  *		};
  *
  * Pattern can combine multiple properties allowing for more complex view element matching:
--- a/packages/ckeditor5-engine/src/view/matcher.js
+++ b/packages/ckeditor5-engine/src/view/matcher.js
@@ -335,7 +335,8 @@ function matchPatterns( patterns, attributes ) {
 		} );
 	} );
 
-	if ( !match.length ) {
+	// Return null when there was no matches or we didn't match all patterns.
+	if ( !match.length || match.length != patterns.length ) {
 		return null;
 	}
 
--- a/packages/ckeditor5-engine/src/view/matcher.js
+++ b/packages/ckeditor5-engine/src/view/matcher.js
@@ -322,29 +322,27 @@ function matchPatterns( patterns, attributes ) {
 		}
 	}
 
-	patterns = normalizePatterns( patterns );
+	patterns = normalizePatterns( patterns ) || [];
 
-	for ( const pattern in patterns ) {
-		const { key: patternKey, value: patternVal } = patterns[ pattern ];
-
+	patterns.forEach( ( { key: patternKey, value: patternVal } ) => {
 		attributes.forEach( ( [ attributeKey, attributeValue ] ) => {
 			if (
-				( patternKey instanceof RegExp && patternKey.test( attributeKey ) ) ||
-				patternKey === attributeKey
+				patternKey === attributeKey ||
+				( patternKey instanceof RegExp && patternKey.test( attributeKey ) )
 			) {
-				if ( patternVal === true ) {
-					match.push( attributeKey );
-				} else if ( patternVal === attributeValue ) {
-					match.push( attributeKey );
-				} else if ( patternVal instanceof RegExp && patternVal.test( attributeValue ) ) {
+				if (
+					patternVal === true ||
+					patternVal === attributeValue ||
+					( patternVal instanceof RegExp && patternVal.test( attributeValue ) )
+				) {
 					match.push( attributeKey );
 				}
 			}
 		} );
+	} );
 
-		if ( !match.length ) {
-			return null;
-		}
+	if ( !match.length ) {
+		return null;
 	}
 
 	return match;
@@ -364,7 +362,7 @@ function matchPatterns( patterns, attributes ) {
 //		value: /https:.*/
 //	}
 // @param {Object|Array} patterns
-// @returns {Array|null} Returns an array of objects or null if patterns were not in a form of expected input.
+// @returns {Array|null} Returns an array of objects or null if provided patterns were not in an expected form.
 function normalizePatterns( patterns ) {
 	if ( Array.isArray( patterns ) ) {
 		return patterns.map(
@@ -382,9 +380,9 @@ function normalizePatterns( patterns ) {
 		return Object.entries( patterns ).map(
 			( [ patternKey, patternValue ] ) => ( { key: patternKey, value: patternValue } )
 		);
-	} else {
-		return null;
 	}
+
+	return null;
 }
 
 // Checks if attributes of provided element can be matched against provided patterns.
@@ -406,10 +404,10 @@ function matchAttributes( patterns, element ) {
 // @param {module:engine/view/element~Element} element Element which classes will be tested.
 // @returns {Array|null} Returns array with matched class names or `null` if no classes were matched.
 function matchClasses( patterns, element ) {
-	const classNames = Array.from( element.getClassNames() );
-	const attributes = classNames.map( className => [ className ] );
+	const classes = Array.from( element.getClassNames() )
+		.map( className => [ className ] );
 
-	return matchPatterns( patterns, attributes );
+	return matchPatterns( patterns, classes );
 }
 
 // Checks if styles of provided element can be matched against provided patterns.
@@ -419,10 +417,10 @@ function matchClasses( patterns, element ) {
 // @param {module:engine/view/element~Element} element Element which styles will be tested.
 // @returns {Array|null} Returns array with matched style names or `null` if no styles were matched.
 function matchStyles( patterns, element ) {
-	const styleNames = element.getStyleNames();
-	const attributes = styleNames.map( style => [ style, element.getStyle( style ) ] );
+	const styles = element.getStyleNames()
+		.map( style => [ style, element.getStyle( style ) ] );
 
-	return matchPatterns( patterns, attributes );
+	return matchPatterns( patterns, styles );
 }
 
 /**
--- a/packages/ckeditor5-engine/src/view/matcher.js
+++ b/packages/ckeditor5-engine/src/view/matcher.js
@@ -7,8 +7,8 @@
  * @module engine/view/matcher
  */
 
-import { logWarning } from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 import { isPlainObject } from 'lodash-es';
+import { logWarning } from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 
 /**
  * View matcher class.
--- a/packages/ckeditor5-engine/src/view/matcher.js
+++ b/packages/ckeditor5-engine/src/view/matcher.js
@@ -382,7 +382,13 @@ function isValueMatched( patternValue, itemKey, getter ) {
 // @param {module:engine/view/element~Element} element Element which attributes will be tested.
 // @returns {Array|null} Returns array with matched attribute names or `null` if no attributes were matched.
 function matchAttributes( patterns, element ) {
-	return matchPatterns( patterns, element.getAttributeKeys(), key => element.getAttribute( key ) );
+	// Both `class` and `style` attributes are handled in `matchClasses()` and `matchStyles()` respectively.
+	const attributesToExclude = new Set( [ 'class', 'style' ] );
+	const attributeKeys = [ ...element.getAttributeKeys() ].filter(
+		key => !attributesToExclude.has( key )
+	);
+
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -242,71 +244,183 @@ function matchName( pattern, name ) {
...
+		);
+	}
+
+	return null;
+}
+
+// @param {String|RegExp} patternKey A pattern representing attribute key we want to match.
+// @param {String} attributeKey An actual attribute key (e.g. `'src'`, `'background-color'`, `'ck-widget'`) we're testing against pattern.
+// @returns {Boolean}
+function isAttributeKeyMatched( patternKey, attributeKey ) {
```
jacekbogdanski, line 386: Maybe just `matchAttributeKey`?

packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -242,71 +244,183 @@ function matchName( pattern, name ) {
...
+	return patternKey === attributeKey ||
+		( patternKey instanceof RegExp && patternKey.test( attributeKey ) );
+}
+
+// @param {String|RegExp} patternValue A pattern representing attribute value we want to match.
+// @param {String} attributeValue An actual attribute value (e.g. `'http://example.com/'`, `'red'`) we're testing against pattern.
+// @returns {Boolean}
+function isAttributeValueMatched( patternValue, attributeValue ) {
```
jacekbogdanski, line 394: `matchAttributeValue`?

packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -316,31 +430,37 @@ function matchClasses( patterns, element ) {
...
 function matchStyles( patterns, element ) {
-	const match = [];
+	const styles = [];
 
-	for ( const name in patterns ) {
-		const pattern = patterns[ name ];
+	element.getStyleNames()
+		.forEach( style => {
+			// Add the direct style (might be a shorthand).
+			styles.push( [ style, element.getStyle( style ) ] );
```
jacekbogdanski, line 438: It was actually surprising for me that we don't have a method of returning styles similar to `element.getAttributes()`. But I see that actually `getStyleNames` is using private `_getStyleEntries` method:
https://github.com/ckeditor/ckeditor5/blob/a8b34f05482d25ac6ecbfa60b073e9799c6412e3/packages/ckeditor5-engine/src/view/stylesmap.js#L357-L365
Maybe we should add `element.getStyles()` method instead?

packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -242,71 +244,183 @@ function matchName( pattern, name ) {
...
+	if ( Array.isArray( patterns ) ) {
+		return patterns.map(
+			pattern => {
+				// eslint-disable-next-line dot-notation
+				if ( isPlainObject( pattern ) && pattern[ 'key' ] && pattern[ 'value' ] ) {
+					return pattern;
+				}
+
+				// Assume the pattern is either String or RegExp.
+				return { key: pattern, value: true };
```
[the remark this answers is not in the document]
jacekbogdanski: It would be also good to have some test coverage for that.

packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -242,71 +244,183 @@ function matchName( pattern, name ) {
...
+}
+
+// Checks if attributes of provided element can be matched against provided patterns.
+//
+// @param {Object} patterns Object with information about attributes to match. Each key of the object will be
+// used as attribute name. Value of each key can be a string or regular expression to match against attribute value.
+// @param {module:engine/view/element~Element} element Element which attributes will be tested.
+// @returns {Array|null} Returns array with matched attribute names or `null` if no attributes were matched.
+function matchAttributes( patterns, element ) {
+	// TODO: The `getAttributes()` returns `class` and `style` attribute if present. Should we exclude them when querying attributes?
```
jacekbogdanski, line 407: I suppose that for now it should return all attributes - it's a bit counterintuitive because we are separating `classes`, `styles` and `attributes` in MatcherPattern, but on the other hand we can request `style` or `class` attribute in old API:
```js
const matcherPattern = { attributes: { style: true, class: true } };
```

jacekbogdanski (commented):
Looks pretty good 👍🏻 

TBC

packages/ckeditor5-engine/src/view/matcher.js
jacekbogdanski: After small talk let's go with the current naming as it better indicates that we are expecting boolean value.

packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -242,40 +239,156 @@ function matchName( pattern, name ) {
...
+	return patternKey === true ||
+		patternKey === itemKey ||
+		patternKey instanceof RegExp && patternKey.test( itemKey );
+}
+
+// @param {String|RegExp} patternValue A pattern representing a value we want to match.
+// TODO param
+// @param {String} itemValue An actual item value (e.g. `'http://example.com/'`, `'red'`) we're testing against pattern.
+// @returns {Boolean}
+function isValueMatched( patternValue, itemKey, getter ) {
```
jacekbogdanski, line 368: `getter` is not documented. Also, maybe it could be renamed to `attributeGetter` or some other more verbose name?

packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -394,6 +462,13 @@ function matchStyles( patterns, element ) {
  *			}
  *		};
  *
+ *		// Match view element which has many custom data attributes.
+ *		const pattern = {
+ *			attributes: [
+ *				{ key: /data-.+/, value: true }
+ *			]
+ *		}
+ *
```
jacekbogdanski, line 471: Docs should be also extended with other new options, like `attributes: true` etc

packages/ckeditor5-engine/tests/view/matcher.js
```diff
@@ -83,7 +90,141 @@ describe( 'Matcher', () => {
...
+		it( 'should match nothing if pattern is incorrect', () => {
+			const pattern = {
+				// A stub for a non-plain object.
+				attributes: 1
+			};
+			const matcher = new Matcher( pattern );
+			const el1 = new Element( document, 'p', { 'data-foo': 'foo', 'data-bar': 'bar', title: 'other' } );
+
+			expect( matcher.match( el1 ) ).to.be.null;
+		} );
```
jacekbogdanski, line 225: It would be also good to add a similar test for incorrect key/value objects, like missing key/value, etc.

packages/ckeditor5-engine/tests/view/matcher.js
```diff
@@ -176,6 +317,25 @@ describe( 'Matcher', () => {
 			expect( new Matcher( { classes: 'baz' } ).match( el1 ) ).to.be.null;
 		} );
 
+		it( 'should match element class names using an array', () => {
+			const pattern = { classes: [ 'foobar', 'foobaz' ] };
+			const matcher = new Matcher( pattern );
+			const el1 = new Element( document, 'p', { class: 'foobar foobaz' } );
+
+			const el2 = new Element( document, 'p', { class: 'foobaz'	} );
```
jacekbogdanski, line 325: Something wrong with spacing in the object.

packages/ckeditor5-engine/tests/view/matcher.js
```diff
@@ -176,6 +317,25 @@ describe( 'Matcher', () => {
...
+			const el3 = new Element( document, 'p', { class: 'qux'	} );
+
+			const result = matcher.match( el1 );
+			expect( result ).to.be.an( 'object' );
+			expect( result ).to.have.property( 'element' ).that.equal( el1 );
+			expect( result ).to.have.property( 'pattern' ).that.equal( pattern );
+			expect( result ).to.have.property( 'match' ).that.has.property( 'classes' ).that.is.an( 'array' );
+			expect( result.match.classes[ 0 ] ).equal( 'foobar' );
```
jacekbogdanski, line 333: Shouldn't we also check for array size and `foobaz` class?

packages/ckeditor5-engine/tests/view/matcher.js
```diff
@@ -176,6 +317,25 @@ describe( 'Matcher', () => {
 			expect( new Matcher( { classes: 'baz' } ).match( el1 ) ).to.be.null;
 		} );
 
+		it( 'should match element class names using an array', () => {
+			const pattern = { classes: [ 'foobar', 'foobaz' ] };
+			const matcher = new Matcher( pattern );
+			const el1 = new Element( document, 'p', { class: 'foobar foobaz' } );
```
jacekbogdanski, line 323: Maybe just `foo` and `bar`? Nothing important thou, it would be just a bit more readable when reading expected results.

jacekbogdanski (commented):
Seems like we are also missing unit tests for regexp pattern full properties, like:
```js
const match = {
  attributes: /^data-.*$/
}
```

packages/ckeditor5-engine/src/view/matcher.js
maxbarnas: Good catch! Thanks!

packages/ckeditor5-engine/tests/view/stylesmap.js
```diff
@@ -270,5 +270,17 @@ describe( 'StylesMap', () => {
...
+			stylesMap.setTo( 'margin: 1px' );
+
+			expect( stylesMap.getStyleNames( true ) ).to.deep.equal( [
+				'margin', 'margin-top', 'margin-right', 'margin-bottom', 'margin-left'
+			] );
+
+			stylesMap.setTo( 'margin-top: 1px' );
+
+			expect( stylesMap.getStyleNames( true ) ).to.deep.equal( [ 'margin', 'margin-top' ] );
+		} );
```
jacekbogdanski, line 284: It would be also good to have the sad path covered checking out if `getStylesNames` is not returning all styles from extractors. Like
```js
		it( 'should output full names for known style names - deep = true', () => {
		        addBorderRules( stylesProcessor ); // Let's add this line to ensure that only matching extractors are used to expand style names.

			stylesMap.setTo( 'margin: 1px' );

			expect( stylesMap.getStyleNames( true ) ).to.deep.equal( [
				'margin', 'margin-top', 'margin-right', 'margin-bottom', 'margin-left'
			] );

			stylesMap.setTo( 'margin-top: 1px' );

			expect( stylesMap.getStyleNames( true ) ).to.deep.equal( [ 'margin', 'margin-top' ] );
		} );
```

It may seem a bit redundant, but looking at the code:
https://github.com/ckeditor/ckeditor5/blob/0c9267f8c5e16aaa75342ab9d973b8fae5ccf2a4/packages/ckeditor5-engine/src/view/stylesmap.js#L585-L593
It's not totally obvious since we are operating on all registered CSS properties.

packages/ckeditor5-engine/tests/view/matcher.js
```diff
@@ -83,7 +92,141 @@ describe( 'Matcher', () => {
...
 		} );
 
-		it( 'should match element attributes', () => {
+		it( 'should match all element attributes', () => {
+			const pattern = {
+				attributes: true
+			};
+			const matcher = new Matcher( pattern );
+			const el1 = new Element( document, 'p', { title: 'foobar' } );
+			const el2 = new Element( document, 'p', { title: ''	} );
```
jacekbogdanski, line 101: Maybe we could add another attribute to make sure that it's able to match more than one?

packages/ckeditor5-engine/tests/view/matcher.js
```diff
@@ -246,6 +550,65 @@ describe( 'Matcher', () => {
...
+			expect( result ).to.have.property( 'match' ).that.has.property( 'styles' ).that.is.an( 'array' );
+			expect( result.match.styles ).to.deep.equal( [
+				'border-color',
+				'border-style',
+				'border-width',
+				'border-top',
+				'border-top-color',
+				'border-top-style',
+				'border-top-width'
+			] );
```
jacekbogdanski, line 577: I'm wondering if currently we are not matching too much, but I will explain my doubts in a summary comment.

packages/ckeditor5-engine/tests/view/matcher.js
```diff
@@ -246,6 +550,65 @@ describe( 'Matcher', () => {
...
+			const el1 = new Element( document, 'p', { style: 'border-top: 1px solid blue' } );
+
+			matcher.match( el1 );
+
+			sinon.assert.calledOnceWithMatch(
+				warnStub,
+				'matcher-patterns-pattern-missing-key-or-value',
+				pattern.styles[ 0 ]
+			);
+		} );
```
jacekbogdanski, line 610: It's only missing `key`, we would need a separate test for `value`.

packages/ckeditor5-engine/tests/view/matcher.js
jacekbogdanski: I started writing my doubts and find out that I missed the point that we will be passing more strict patterns this match will be much more restricted (like using only `border-top` will only give us `border-top` property, not all of them) so it seems to be fine 👍🏻

jacekbogdanski:
@maxbarnas I've added a test in https://github.com/ckeditor/ckeditor5/pull/9694/commits/526d5d7f939f17921cdf21aaabcb16419e33e2be verifying if matcher returns the unique result for similar patterns. It wasn't the case previously as it was not possible to match the attribute key multiple times, but now this test fails.

packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -284,29 +398,8 @@ function matchAttributes( patterns, element ) {
...
-		} else if ( element.hasClass( pattern ) ) {
-			match.push( pattern );
-		} else {
-			return null;
-		}
-	}
-
-	return match;
+	// We don't need `getter` here because patterns for classes are always normalized to `[ className, true ]`.
+	return matchPatterns( patterns, element.getClassNames() );
```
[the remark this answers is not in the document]
maxbarnas: By the check for `true` I assume the condition in `isValueMatched()` function. I can't remove it as it is a check for a pattern value, not item value. While the getter here provides value.

packages/ckeditor5-engine/src/view/matcher.js
```diff
@@ -7,6 +7,9 @@
  * @module engine/view/matcher
  */
 
+import { logWarning } from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
```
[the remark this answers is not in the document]
maxbarnas: The `ckeditor5-engine` package is reexported in `ckeditor5` import package. Seems like it would cause circular dependency - that is my assumption, as the precommit hook stops the commit with an error without detailed description.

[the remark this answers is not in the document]
jacekbogdanski: Indeed I see that similar imports in engine are done via your approach - https://github.com/ckeditor/ckeditor5/blob/3b85ab92a2366983b2e29d3a4111c7a7139194a5/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L10

jacekbogdanski:
@maxbarnas I've rewritten a bit of documentation for new matcher options. It's more organized now so should be more readable for users. Could you take a look at these changes if everything is ok? https://github.com/ckeditor/ckeditor5/pull/9694/commits/84f971f9d792c797186baf7340fd6b7371d74165

jacekbogdanski (approved):
🎉

PR #9728, opened 2021-05-19
#9684: Fixed remove format plugin issue with image resizing

### LINKED ISSUE #9684: Image size can be reset with remove format plugin

1. Resize an image
2. Select it
3. Click `remove format`

Not sure if bug or rather an improvement proposal, but this may become annoying with inline images. I can imagine a situation when someone applies for example `bold` to some bigger selection, containing inline image. Then the user selects even bigger part of content and hits `remove format` to get rid of `bold`, but it also resets the size of the image. In such case I would assume that only the text formatting will be removed and that image size should be a separate thing here.   
  
Example:

https://user-images.githubusercontent.com/34380544/117978302-03e4e000-b332-11eb-86ba-766663ea0315.mp4

---

mateuszzagorski:

MINOR BREAKING CHANGE (image): Resized images' size will no longer reset by remove format feature. The `width` attribute in model schema is no longer marked as with `isFormatting` property. Closes #9684.

---

Those changes were needed in order for the remove-format plugin to work as intended.


mateuszzagorski:
I've resolved the existing conflicts.  
  
However, these needs to be reviewed further as previous changes on this branch, after merging them with the latest master are not working in an expected way - so the issue is not resolved at this stage:  
\- Image width attribute will currently be removed when "remove-format" is applied.   
  
Looking into it.

packages/ckeditor5-remove-format/tests/integration.js
```diff
@@ -65,6 +66,19 @@ describe( 'RemoveFormat', () => {
 					'<imageBlock src="assets/sample.png"><caption>caption</caption></imageBlock><paragraph>bar]</paragraph>' );
 		} );
 
+		it( 'does not remove image width attribute', () => {
+			setModelData( model, '<paragraph>foo[<$text bold="true">foo</$text></paragraph>' +
```
mlewand, line 70:
```suggestion
			// (#9684)
			setModelData( model, '<paragraph>foo[<$text bold="true">foo</$text></paragraph>' +
```

It's a good idea to reference the issue that made to make this decision.

packages/ckeditor5-remove-format/tests/integration.js
```diff
@@ -9,6 +9,7 @@ import RemoveFormat from '../src/removeformat';
 import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
 import Heading from '@ckeditor/ckeditor5-heading/src/heading';
 import Image from '@ckeditor/ckeditor5-image/src/image';
+import ImageResizeEditing from '@ckeditor/ckeditor5-image/src/imageresize/imageresizeediting';
```
mlewand, line 12:
```suggestion
import ImageResize from '@ckeditor/ckeditor5-image/src/imageresize';
```

Let's just add full featured image resize editing plugin like you did in another manual test.

packages/ckeditor5-remove-format/tests/integration.js
```diff
@@ -29,7 +30,7 @@ describe( 'RemoveFormat', () => {
 
 		return ClassicTestEditor
 			.create( element, {
-				plugins: [ Paragraph, Heading, Image, Bold, Underline, RemoveFormat, Image, ImageCaption, Link ]
+				plugins: [ Paragraph, Heading, Image, Bold, Underline, RemoveFormat, Image, ImageCaption, ImageResizeEditing, Link ]
```
mlewand, line 33:
```suggestion
				plugins: [ Paragraph, Heading, Image, Bold, Underline, RemoveFormat, Image, ImageCaption, ImageResize, Link ]
```

packages/ckeditor5-remove-format/tests/manual/removeformat.md
```diff
@@ -1,10 +1,12 @@
 # Remove Format
 
+Select the image and resize it.
```
mlewand, line 3: It renders like that:



I'd prefer it to be a list of steps to be more readable, like that:

> 1.  Select the image and resize it.
> 2.  Select the content and press the "Remove Format" button in the toolbar.

BUT, I think we can even make it simpler 🙂 Wouldn't it be better for testers if instead of having to resize the image every single time they would already have it resized?

All you need to do is to edit `removeformat.html` file and add `width` style and `image_resized` class to the figure.

```css
<figure class="image image_resized" style="width:18%;"><img src="assets/sample.png" alt="Autumn fields"><figcaption>Autumn fields by Aleksander Nowodziński</figcaption></figure>
```

With such change the image is already resized when the manual test is open 👌 Less steps = more time for other things.

packages/ckeditor5-remove-format/tests/integration.js
```diff
@@ -65,6 +66,19 @@ describe( 'RemoveFormat', () => {
 					'<imageBlock src="assets/sample.png"><caption>caption</caption></imageBlock><paragraph>bar]</paragraph>' );
 		} );
 
+		it( 'does not remove image width attribute', () => {
+			setModelData( model, '<paragraph>foo[<$text bold="true">foo</$text></paragraph>' +
+				'<imageBlock src="assets/sample.png" width="50%"><caption>caption</caption></imageBlock>' +
+				'<paragraph>bar]</paragraph>' );
```
mlewand, line 72: This unit test can be further simplified. The only goal of your test is to check whether width attribute is removed from a block image. You don't care about paragraphs.

So let's just remove them. Simplified your test can look like that:

```
	it( 'does not remove image width attribute', () => {
		setModelData( model, '[<imageBlock src="assets/sample.png" width="50%"><caption>caption</caption></imageBlock>]' );

		editor.execute( 'removeFormat' );

		expect( getModelData( model ) )
			.to.equal( '[<imageBlock src="assets/sample.png" width="50%"><caption>caption</caption></imageBlock>]' );
	} );
```

Note that after doing this the test no longer **has no longer multiline range**, while it is contained in a [group that tests multiline ranges](https://github.com/ckeditor/ckeditor5/blob/9557aea925b968cfd3ba82810b1ef09e96c03164/packages/ckeditor5-remove-format/tests/integration.js#L47). So you'll have to move it outside of this group (IMHO it make sense to put it directly in [RemoveFormat test suite](https://github.com/ckeditor/ckeditor5/blob/9557aea925b968cfd3ba82810b1ef09e96c03164/packages/ckeditor5-remove-format/tests/integration.js#L24).

mlewand (changes requested):
The fix looks fine 👍 There are couple of minor things to correct in tests.

> Fix: Images won't be automatically resized anymore when using the remove format plugin. 

Your commit message should include the package that is affected by the change. In this case that's image package.

packages/ckeditor5-remove-format/tests/manual/removeformat.md
```diff
@@ -6,5 +6,6 @@ Select the content and press the "Remove Format" button in the toolbar.
 
 * Bold, emphasize and underline should be removed.
 * Links **must not** be removed.
+* Image **must not** change its size.
```
mlewand, line 9: Good catch that the description should be changed from "Image \*\*must not\*\* return to its original size." after changing the behavior to "Image \*\*must not\*\* change its size." 👍

packages/ckeditor5-remove-format/tests/manual/removeformat.html
```diff
@@ -9,9 +9,9 @@
 		.
 	</p>
 
-	<figure class="image">
+	<figure class="image image_resized" style="width:18%;">
 		<img src="assets/sample.png" alt="Autumn fields">
-		<figcaption>Autumn <em>fields</em> by Aleksander Nowodzi&#x144;ski</figcaption>
+		<figcaption>Autumn fields by Aleksander Nowodziński</figcaption>
```
mlewand, line 14: I don't know what's the reason for removing `em` formatting in this particular line. I'll revert this (but I get substitution for a regular letter - improved code readability).

mlewand (approved):
Perfect, looks good 👍

packages/ckeditor5-remove-format/tests/manual/removeformat.html
mlewand: Done in 8a6cfea

PR #9736, opened 2021-05-20
Source editing feature for the classic editor

### LINKED ISSUE #9647: Implement the view source feature for the classic editor

Extracted from #592.

This feature request is about introducing a feature for the classic editor that will allow switching to the source editing mode.

---

psmyrek:

Feature (source-editing): Source editing feature for the classic editor. Closes #9647.

Feature (engine): Introduced `batchType` option in `editor.data.set()` which can be used to preserve the undo/redo steps and to add an additional item to the undo stack. Note that it will still replace the whole content and should not be used with real-time collaboration.


pomek:
Two things:

1. Please, add the `CHANGELOG.md` file with the following content:
```md
Changelog
=========

All changes in the package are documented in the main repository. See: https://github.com/ckeditor/ckeditor5/blob/master/CHANGELOG.md.
```
2. Did you check how the plugin behaves in the read-only mode?

psmyrek:
I forgot about the read-only mode after being told not to use commands, but the plain function bound to the button's `execute` event instead. I will add support for read-only mode.

harleenarora:
How to get the "view source" plugin all files as zip and any other way?

psmyrek:
The `SourceEditing` plugin is almost finished. There are a few things to do:
- create and use new icon for this feature,
- handle undo in the `DataController#set()` method,
- integration with the Collaboration Features.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,276 @@
...
+		const editor = this.editor;
+		const editingView = editor.editing.view;
+
+		// It is not needed to iterate through all editing roots, as currently the plugin supports only the Classic Editor with single
+		// main root, but this code may help understand and use this feature in external integrations.
+		for ( const [ rootName, domRootElement ] of editingView.domRoots ) {
+			const data = editor.data.get( { rootName } );
+
+			// Remove the current data from the editing root, as the data will be overwritten anyway upon exit from the source editing mode.
+			editor.data.set( { [ rootName ]: '' } );
```
ma2ciek, line 170: What is the actual purpose of this line?

I'm not sure if it is a good idea, it will create another undo step...

packages/ckeditor5-source-editing/docs/features/source-editing.md
```diff
@@ -0,0 +1,73 @@
...
+category: features
+menu-title: Source editing
+modified_at: 2021-05-28
+---
+
+# Source editing
+
+The {@link module:source-editing/sourceediting~SourceEditing} feature provides the ability for viewing and editing the source of the document. The source editing plugin is a low-level document editing interface, while all the buttons and dropdowns located in a editor's toolbar are the high-level ones.
+
+The changes made to the document source will be applied to the editor's {@link framework/guides/architecture/editing-engine data model} only, if the editor understands (via loaded plugins) the given syntax. You will loose all changes done in the source editing mode unless they are compatible with what the initialized features understand. For example, if editor does not have a {@link features/horizontal-line horizontal line} plugin loaded, the `<hr>` tag added in the document source will be removed upon exit from the source editing mode.
```
ma2ciek, line 11: > You will loose all changes done in the source editing mode unless they are compatible with what the initialized features understand.

I'd rephrase it more to "You will lose all changes that editor features can't understand" or something similar.

packages/ckeditor5-source-editing/docs/features/source-editing.md
```diff
@@ -0,0 +1,73 @@
...
+
+<info-box>
+	Currently, the source editing mode is supported only for the {@link examples/builds/classic-editor classic editor}.
```
ma2ciek, line 14: I'm not sure if this is the best place to link. However, AFAIK we don't have a guide describing all editor types :/ So maybe there's no better place ATM.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Yeah, I was wondering how to solve 2 issues:

1. After opening the source editing mode, other buttons in the toolbar become disabled, but they can be "on", if the current selection's attribute match them.
2. In Firefox only, while the table (or other) contextual balloon is displayed and source editing mode is opened, the contextual balloon is re-positioned to the top left edge of the viewport. Sometimes (quite randomly), also the generic Firefox `NS_ERROR_FAILURE` is thrown in this case.

In case of the first issue, it could be quite easily (I think) solve by something like:

```js
model.change( writer => {
    writer.setSelection( null );
    writer.removeSelectionAttribute( model.document.selection.getAttributeKeys() );
} );
```

In case of the second issue, it should be investigated why the balloon is not hidden in Firefox. Quick solution would be to check, if the `ContextualBalloon` is loaded in the editor, and if yes then call `balloon.view.hide()`. I don't know what the consequences will be, to be honest.

Both solutions add some additional logic, and IMHO both issues can be easily solved by removing the current data from the editing root by calling the `editor.data.set()` with an empty string. And important thing is that this one `editor.data.set()` call **should not add undo step** - I hope this will be possible to achieve after the `DataController#set()` method will handle undo.

To sum up, this one line was my first idea to solve both issues at once.

psmyrek: After thinking about this problem over the weekend, I changed my solution and now the editor data is no longer deleted.

packages/ckeditor5-source-editing/docs/features/source-editing.md
psmyrek: Done.

packages/ckeditor5-source-editing/docs/features/source-editing.md
psmyrek: So I'm leaving it as it is now.

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -350,8 +350,13 @@ export default class DataController {
 	 * @fires set
 	 * @param {String|Object.<String,String>} data Input data as a string or an object containing `rootName` - `data`
 	 * pairs to set data on multiple roots at once.
+	 * @param {Object} [options]
```
pomek, line 353:
```suggestion
	 * @param {Object} [options={}]
```

As it defines the default value.

packages/ckeditor5-undo/src/basecommand.js
```diff
@@ -42,7 +42,26 @@ export default class BaseCommand extends Command {
...
+		// Set the transparent batch for the `editor.data.set()` call if the
+		// batch type is not set and the `options.clearUndoStack` is falsy.
+		this.listenTo( editor.data, 'set', ( evt, data ) => {
+			const options = data[ 1 ] || {};
+
+			if ( options.batchType ) {
+				return;
+			}
+
+			options.batchType = options.supportUndo ? 'default' : 'transparent';
```
pomek, line 54: TBH, the `supportUndo` option does not sound good in the `Undo/Redo` context. I'd rename the property for something like `disableClearingStacks` or something that will explain what it does .

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -386,7 +391,11 @@ export default class DataController {
 				writer.remove( writer.createRangeIn( modelRoot ) );
 				writer.insert( this.parse( newData[ rootName ], modelRoot ), modelRoot, 0 );
 			}
-		} );
+		};
+
+		const batchType = options.batchType || 'default';
+
+		this.model.enqueueChange( batchType, setNewContent );
```
pomek, line 398: The `setNewContent` callback does not have to be assigned to the variable.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,307 @@
...
+ */
+
+import { Plugin, PendingActions } from 'ckeditor5/src/core';
+import { ButtonView } from 'ckeditor5/src/ui';
+import { createElement, ElementReplacer } from 'ckeditor5/src/utils';
+
+import '../theme/sourceediting.css';
+
+// TODO: create icon
+// import sourceEditingIcon from '../theme/icons/sourceediting.svg';
```
pomek, line 17: Make sure to remove these comments before the merge.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,307 @@
...
+			// - there is a pending action.
+			//
+			// Pending action may change the model, so viewing and/or editing the document source should be prevented until the model is
+			// finally set.
+			buttonView.bind( 'isEnabled' ).to(
+				this, 'isEnabled',
+				editor, 'isReadOnly',
+				editor.plugins.get( PendingActions ), 'hasAny',
+				( isEnabled, isEditorReadOnly, hasAnyPendingActions ) => isEnabled && !isEditorReadOnly && !hasAnyPendingActions
+			);
```
pomek, line 109:
```suggestion
			// The button should be disabled if one of the following condition is met.
			buttonView.bind( 'isEnabled' ).to(
				this, 'isEnabled',
				editor, 'isReadOnly',
				editor.plugins.get( PendingActions ), 'hasAny',
				( isEnabled, isEditorReadOnly, hasAnyPendingActions ) => {
					// (1) The plugin is disabled itself.
					if ( !isEnabled ) {
						return false;
					}

					// (2) The editor is in read-only mode.
					if ( isEditorReadOnly ) {
						return false;
					}

					// (3) Any pending action is scheduled. It may change the model, so modifying the document source 
					// should be prevented until the model is finally set.
					if ( hasAnyPendingActions ) {
						return false;
					}

					return true;
				}
			);
```

WDYT?

pomek (changes requested):
I did not find a test that checks whether changes introduced by the plugin can be reverted/restored (undo/redo).

packages/ckeditor5-undo/src/basecommand.js
ma2ciek: The problem with the `disableClearingStacks` is that we need to do two things here. First, we need to preserve stacks, and second, we need to set the batch type.

But this option makes sense from the end-user point of view when someone calls `edtior.set.data( { supportUndo: true } )`. Or `editor.setData( { supportUndo: true } )` 

You don't want to write `editor.data.set( { disableClearingStacks: true } )` as you would have no idea what it is about.

ma2ciek: The problem with the `disableClearingStacks` is that we need to do two things here. First, we need to preserve stacks, and second, we need to set the batch type.

But this option makes sense from the end-user point of view when someone calls `edtior.set.data( { supportUndo: true } )`. Or `editor.setData( { supportUndo: true } )` 

You don't want to write `editor.data.set( { disableClearingStacks: true } )` as you would have no idea what it is about.

pomek: I understood, but still, it does not sound good: `editor.setData( data, { supportUndo: true } )` looks like the Undo feature is not supported if the option is not specified. 

What about the option called `createUndoStep: true`, which is false by default? This explicitly says what we want to achieve – "set data and create an undo step for this action".

ma2ciek: IMO we could either have `supportUndo` or both - `createUndoStep` + `preserveUndoHistory`. But TBH, setting only one from this pair doesn't make sense to me :smile: .

psmyrek: What about:

1. `editor.data.set( { withUndo: true } )`: generic flag name to toggle an `Undo` feature.
2. `editor.data.set( { allowUndo: true } )`: as above.
3. `editor.data.set( { addUndoStep: true } )`: As @ma2ciek explained, the `Undo` feature consists of two parts (creating an undo step and preserving undo history), but IMHO from end-user perspective it would be sufficient to mention here only the "step" part in the flag name. Moreover, the "undo step" phrase is quite commonly used in our docs so far in the context of "undoing" an operation.

I vote for 2) or 3) 😄

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Good idea, thanks.

packages/ckeditor5-undo/src/basecommand.js
pomek: (2) sounds the best from all proposals to me.

ma2ciek: 132f9f9#diff-cfbce7194f50d3984cd54ed91d7d0fc7f47dd801cc80e6487154e3ae0c0fd80cR393 :D :joy:  

For me, all these names lose some details, but I can't find a better one :)

niegowski: what about `batchType` that we talked about? it could default to unset (‘default’) and get overridden by undo plugin to `transparent` if not set to any other value.

psmyrek:
The work progress on the icon can be followed in https://github.com/ckeditor/ckeditor5/issues/9778.

psmyrek:
Adding this package to the `README.md` and `package.json` files will be handled in https://github.com/ckeditor/ckeditor5/issues/9803.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,319 @@
...
+			return;
+		}
+
+		for ( const [ , domSourceEditingElementWrapper ] of this._replacedRoots ) {
+			domSourceEditingElementWrapper.querySelector( 'textarea' ).readOnly = isReadOnly;
+		}
+	}
+
+	/**
+	 * Checks, if the editor's editable belongs to the editor's DOM tree.
```
ma2ciek, line 309: I'd move it to the private comment inside the method. I'd tell more here what means `true` and `false` from the `_isAllowedToHandleSourceEditingMode`() point of view.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,320 @@
...
+	constructor( editor ) {
+		super( editor );
+
+		/**
+		 * Flag indicating whether the document source mode is active.
+		 *
+		 * @observable
+		 * @member {Boolean}
+		 */
+		this.set( 'isSourceEditingMode', false );
```
ma2ciek, line 58: As this is an observable property, do we need to maintain also the `sourceEditing` event? IMO we can handle all cases via changing this property and handling its value changes.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,320 @@
...
+				this.isSourceEditingMode = !this.isSourceEditingMode;
+
+				/**
+				 * Fired whenever the source editing mode is toggled.
+				 *
+				 * @event sourceEditing
+				 * @param {Object} data Additional information about the event.
+				 * @param {Boolean} data.isSourceEditingMode Flag indicating whether the document source mode is active.
+				 */
+				this.fire( 'sourceEditing', { isSourceEditingMode: this.isSourceEditingMode } );
```
ma2ciek, line 134: As above.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,320 @@
...
+			} );
+
+			return buttonView;
+		} );
+
+		// Currently, plugin handles the source editing mode by itself only for the Classic Editor. To use this plugin with other
+		// integrations, listen to `sourceEditing` event and act accordingly.
+		if ( this._isAllowedToHandleSourceEditingMode() ) {
+			this.on( 'sourceEditing', ( evt, { isSourceEditingMode } ) => {
```
ma2ciek, line 143:
```
this.on( 'change:isSourceEditingMode', ( evt, name, isSourceEditingMode ) => {

} 
```

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Done.
Please check, if it is better now 🙂

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Changed as proposed. Plugin now listen to the `change:isSourceEditingMode`.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Custom `sourceEditing` is no longer fired.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Done.

godai78 (approved):
Looks like everything is covered nicely.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Icon is added.

psmyrek:
It looks like all pending change requests have been done, so it is ready for final (I hope) code review 😅

psmyrek:
I've just added a warning in the console, if CF plugin is loaded. This is the basic integration with CF for now.

packages/ckeditor5-source-editing/docs/features/source-editing.md
```diff
@@ -11,7 +11,7 @@ The {@link module:source-editing/sourceediting~SourceEditing} feature provides t
 The changes made to the document source will be applied to the editor's {@link framework/guides/architecture/editing-engine data model} only, if the editor understands (via loaded plugins) the given syntax. You will lose all changes that the editor features cannot understand. For example, if the editor does not have a {@link features/horizontal-line horizontal line} plugin loaded, the `<hr>` tag added in the document source will be removed upon exit from the source editing mode.
 
 <info-box>
-	Currently, the source editing mode is supported only for the {@link examples/builds/classic-editor classic editor}.
+	Currently, the source editing mode is supported only for the {@link examples/builds/classic-editor classic editor}. The source editing feature may not work with [CKEditor 5 collaboration features](https://ckeditor.com/docs/ckeditor5/latest/features/collaboration/collaboration.html).
```
ma2ciek, line 14: `may not work` -> `does not work`? `supported for` -> `supported by`?

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -146,8 +148,30 @@ export default class SourceEditing extends Plugin {
...
-	 *  the document source from the corresponding root.
+	 * @inheritDoc
+	 */
+	afterInit() {
+		const editor = this.editor;
+
+		const pluginNamesToWarn = [
+			'RealTimeCollaborativeEditing',
+			'CommentsEditing',
+			'TrackChangesEditing'
```
ma2ciek, line 159: Let's add the `RevisionHistory` as well :)

packages/ckeditor5-source-editing/docs/features/source-editing.md
godai78: I'd say "supported in the classic editor".  
  
As for "may" it depends - if it never works, then yes, "does not work".

psmyrek: I would leave the first suggestion ("_may not work_") as it is now, because in some cases `SourceEditing` will work with CF plugins.

I've changed the second one ("_supported only for_" -> "_supported only by_") as you proposed.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Done, thanks.

packages/ckeditor5-source-editing/docs/features/source-editing.md
psmyrek: Now I noticed a comment from @godai78, sorry 🙂 I changed this sentence as proposed: "_Currently, the source editing mode is supported in the classic editor._"

packages/ckeditor5-source-editing/README.md
```diff
@@ -1,4 +1,20 @@
 CKEditor 5 source editing feature
-=================================
+========================================
```
pomek, line 2: I'd revert the change. I guess the number of `=` should be equal to the length of the title string.

packages/ckeditor5-source-editing/README.md
```diff
@@ -1,4 +1,20 @@
...
 
-This is an initial package for development purposes. It does not contain code yet.
+[![npm version](https://badge.fury.io/js/%40ckeditor%2Fckeditor5-source-editing.svg)](https://www.npmjs.com/package/@ckeditor/ckeditor5-source-editing)
+[![Dependency Status](https://david-dm.org/ckeditor/ckeditor5-source-editing/status.svg)](https://david-dm.org/ckeditor/ckeditor5-source-editing)
+[![devDependency Status](https://david-dm.org/ckeditor/ckeditor5-source-editing/dev-status.svg)](https://david-dm.org/ckeditor/ckeditor5-source-editing?type=dev)
```
pomek, line 6: Recently we removed these badges. Please, replace those with the following.

```suggestion
[![Coverage Status](https://coveralls.io/repos/github/ckeditor/ckeditor5/badge.svg?branch=master)](https://coveralls.io/github/ckeditor/ckeditor5?branch=master)
[![Build Status](https://travis-ci.com/ckeditor/ckeditor5.svg?branch=master)](https://travis-ci.com/ckeditor/ckeditor5)
```

packages/ckeditor5-source-editing/docs/features/source-editing.md
```diff
@@ -0,0 +1,73 @@
...
+
+There is other source-related CKEditor 5 feature you may want to check:
+
+* {@link features/html-embed HTML embed} &ndash; Allows embedding an arbitrary HTML snippet in the editor.
+
+## Installation
+
+To add this feature to your rich-text editor, install the [`@ckeditor/ckeditor5-source-editing`](https://www.npmjs.com/package/@ckeditor/ckeditor5-source-editing) package:
+
+```plaintext
```
pomek, line 39:
```suggestion
```bash
```

WDYT?

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,337 @@
...
+	 */
+	init() {
+		const editor = this.editor;
+		const t = editor.t;
+
+		editor.ui.componentFactory.add( 'sourceEditing', locale => {
+			const buttonView = new ButtonView( locale );
+
+			buttonView.set( {
+				label: t( 'Source' ),
```
pomek, line 89:
```suggestion
				label: t( 'View source' ),
```
WDYT? I am afraid that the `Source` is not explicit enough.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -0,0 +1,337 @@
...
+		const editor = this.editor;
+
+		const pluginNamesToWarn = [
+			'RealTimeCollaborativeEditing',
+			'CommentsEditing',
+			'TrackChangesEditing',
+			'RevisionHistory'
+		];
+
+		// Currently, the basic integration with Collaboration Features is to display a warning in the console.
```
pomek, line 163: I'd add a section about the plugin limitation in the documentation. As for now, the code could be translated into words and copied to the feature guide.

packages/ckeditor5-source-editing/theme/sourceediting.css
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+@import "../../ckeditor5-theme-lark/theme/mixins/_rounded.css";
+@import "../../ckeditor5-theme-lark/theme/mixins/_focus.css";
+@import "../../ckeditor5-theme-lark/theme/mixins/_shadow.css";
```
pomek, line 8: Shouldn't it be `@ckeditor/ckeditor5-theme-lark`?

packages/ckeditor5-source-editing/package.json
```diff
@@ -9,6 +9,24 @@
...
+    "@ckeditor/ckeditor5-core": "^27.1.0",
+    "@ckeditor/ckeditor5-dev-utils": "^25.0.0",
+    "@ckeditor/ckeditor5-editor-classic": "^27.1.0",
+    "@ckeditor/ckeditor5-engine": "^27.1.0",
+    "@ckeditor/ckeditor5-essentials": "^27.1.0",
+    "@ckeditor/ckeditor5-paragraph": "^27.1.0",
+    "@ckeditor/ckeditor5-table": "^27.1.0",
+    "@ckeditor/ckeditor5-theme-lark": "^27.1.0",
+    "@ckeditor/ckeditor5-ui": "^27.1.0",
+    "@ckeditor/ckeditor5-utils": "^27.1.0",
```
pomek, line 26: As we released the new version today (`v28.0.0`), CI prints the duplicated packages error. Please, align versions.

packages/ckeditor5-source-editing/README.md
psmyrek: Done.

packages/ckeditor5-source-editing/README.md
psmyrek: Done.

packages/ckeditor5-source-editing/docs/features/source-editing.md
psmyrek: Changed as proposed.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: I would leave it as it is now (with the "Source" label), because:
* this label was proposed by our UI/UX team,
* we often lack space in the toolbar so IMHO a shorter label is a better label,
* we will be consistent with the source area plugin from CKE4, where there is also "Source" label: https://github.com/ckeditor/ckeditor4/blob/major/plugins/sourcearea/lang/en.js

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Currently, the feature guide contains a warning, that this plugin will work out-of-the-box only for classic editor and it may not work with CF. IMHO this is sufficient for a feature guide. Integrator will be also notified in the browser's console in case his/her editor setup contains specific plugins from CF. I wouldn't add these CF plugin names to the feature guide as it would be too specific/detailed.

packages/ckeditor5-source-editing/theme/sourceediting.css
psmyrek: Import path changed. Moved `@ckeditor/ckeditor5-theme-lark` from `devDependencies` to `dependencies`.

packages/ckeditor5-source-editing/package.json
psmyrek: Done, thanks.

hatpick:
@psmyrek can this plugin be used with classic editor?

psmyrek:
@hatpick Yes. This plugin currently only works with classic editor.

hatpick:
@psmyrek thanks for the reply, https://www.npmjs.com/package/@ckeditor/ckeditor5-source-editing is at v0.0.1 and it's empty, how can I add it to my classic editor?

psmyrek:
The source editing feature is not officially published yet. It will be available on npm when next version of CKEditor (v29.0.0) is released.

PR #9742, opened 2021-05-21
i/5797: Attribute element's priority now allowing value of 0 correctly

### LINKED ISSUE #5797: Attribute element's priority 0 is treated as "no priority set"
This works fine:

```js
const span = viewWriter.createAttributeElement( 'span', { class: 'foo' }, { priority: 1 } );
```

This has no effect on the order of attribute elements:

```js
const span = viewWriter.createAttributeElement( 'span', { class: 'foo' }, { priority: 0 } );
```

mateuszzagorski:

Fix (engine): Updated `downcastwriter` to allow setting up attribute element's priority to 0. Closes #5797.


packages/ckeditor5-engine/tests/view/downcastwriter/writer.js
```diff
@@ -122,6 +122,17 @@ describe( 'DowncastWriter', () => {
 			expect( element.id ).to.equal( 'bar' );
 			assertElementAttributes( element, attributes );
 		} );
+
+		it( 'should not treat priority 0 as no priority set', () => {
```
mlewand, line 126:
```suggestion
		it( 'should pass priority 0', () => {
```

You use negation twice in this sentence. You can simplify it down to something like above.

packages/ckeditor5-engine/src/view/downcastwriter.js
```diff
@@ -191,7 +191,7 @@ export default class DowncastWriter {
 	createAttributeElement( name, attributes, options = {} ) {
 		const attributeElement = new AttributeElement( this.document, name, attributes );
 
-		if ( options.priority ) {
+		if ( options.priority !== undefined ) {
```
mlewand, line 194:
```suggestion
		if ( typeof options.priority === 'number' ) {
```

Given that `options.priority` is documented as a number let's be a little more pedantic about what we take here. priority=true or priority='foo' doesn't make much sense here.

packages/ckeditor5-engine/tests/view/downcastwriter/writer.js
```diff
@@ -122,6 +122,17 @@ describe( 'DowncastWriter', () => {
...
+			const element = writer.createAttributeElement( 'foo', attributes, { priority: 0, id: 'bar' } );
+
+			expect( element.is( 'attributeElement' ) ).to.be.true;
```
mlewand, line 129: You can skip checking anything else than `element.priority` here. Other properties are handled by relevant assertion, your point here is to check that priority has been set to `0`.

mlewand (changes requested):
There's another method to create attribute element in https://github.com/ckeditor/ckeditor5/blob/d1215a1c0178efc7e61e7bb978253f2f92c98f20/packages/ckeditor5-engine/src/conversion/downcasthelpers.js#L587 that has not been covered.

Other than that, few minor suggestions.

mlewand (approved):
LGMT

PR #9757, opened 2021-05-26
Docs: table feature update.

godai78:

Docs: updating table feature guide.


AnnaTomanek:
Pushed some updates.

How about we replace the demo of default table and table cell styles with something better (meaningful)? We were supposed to only show demos that look good and make sense. This one is a purely dev sample, not something for the customers to see.

godai78:
I think we can come up with something, like the spacecraft table we used for some other purposes.

godai78:
The very first demo does have a tooltip, as this is a general table demo. Also, the nested tables demo has a tooltip because this is the table insertion demo. The other ones I deem need these not, as they aim at presentation more than interaction.

packages/ckeditor5-table/docs/_snippets/features/table-default-properties.html
```diff
@@ -1,8 +1,8 @@
 <style id="table-default-properties-styles">
     .live-snippet .table-default-properties-container .ck.ck-content .table {
-        float: right;
-        width: 500px;
-        height: 250px;
+        float: left;
+        width: 550px;
+        height: 450px;
     }
```
pomek, line 6: This change must be synchronized with the editor's configuration. See https://github.com/ckeditor/ckeditor5/blob/4cc1422c32dfb6a3ef2e1da0db0752c4795c96c5/packages/ckeditor5-table/docs/_snippets/features/table-default-properties.js#L14-L33

packages/ckeditor5-table/docs/_snippets/features/table-default-properties.html
```diff
@@ -20,16 +20,32 @@
...
+                    <tr>
+                        <td colspan="2"><strong>Animal astronauts by country of origin</strong></td>
+                    </tr>
+                    <tr>
+                        <td>
+                            <p><strong>American</strong>: fruit flies, primates first flew in the 1940s.</p>
+                            <p>The most memorable was <i>Ham</i>, a chimpanzee.</p>
+                        </td>
+                        <td>
+                            <p><strong>Soviet</strong>: mostly dogs, sent to space since 1950s.</p>
```
AnnaTomanek, line 37:
```suggestion
                            <p><strong>Soviet</strong>: mostly dogs, sent to space since the 1950s.</p>
```

packages/ckeditor5-table/docs/_snippets/features/table-default-properties.html
```diff
@@ -20,16 +20,32 @@
 
 <div class="table-default-properties-container">
     <div id="snippet-table-default-properties">
+    <p>Before people could leave the cradle of humanity and enter the cosmic void, tests were required. A number of animals risked and sometimes lost their lives to open the gates for their owners to be able to reach the stars.</p>
+    <p>The animal spaceflights started in the late 1940s. German A4 rockets were used for the initial launches, later on each country developed their own carriers. Even though these simian, canine and rodent space farers were the most important test subjects during the first stage of the space race, before launching actual human cosmonauts, the animal spaceflight is very much a thing even today. In 2018 a rodent crew arrived at International Space Station!</p>
```
AnnaTomanek, line 24:
```suggestion
    <p>The animal spaceflights started in the late 1940s. German A4 rockets were used for the initial launches, later on each country developed their own carriers. Even though these simian, canine and rodent space farers were the most important test subjects during the first stage of the space race, before launching actual human astronauts, the animal spaceflight is very much a thing even today. In 2018 a rodent crew arrived at International Space Station!</p>
```

godai78: Both are correct, by the way. Andcosmonaut is universally used for East, so Laika was a cosmonaut ;)

packages/ckeditor5-table/docs/_snippets/features/table-default-properties.html
godai78: Addressed in the commit.

magda-chrzescian:
> Put the caret anywhere inside the table and click the **"Table properties"** button in the toolbar to open a pop–up with multiple options that will allow you to shape the look of the entire table. If you click the **"Cell properties"** button, a similar interface will appear with styling options for individual table cells.

It is hard to recognize at the first sight, in which toolbar we should look for the buttons (We could call the toolbar "balloon/contextual".
It would be extremely helpful for the user if we could place the button icons next to their names ("Table properties", "Cell properties") in the text.

magda-chrzescian:
I also have a suggestion regarding the table caption button.

image

Since it concerns the whole table, not the cells, it should be separated with the divider from the rest of the buttons and maybe placed at the beginning of the toolbar.

godai78:
> It is hard to recognize at the first sight, in which toolbar we should look for the buttons (We could call the toolbar "balloon/contextual".

We usually call it "table toolbar" like "image toolbar" and so on. We may want to add this here.

godai78:
So this happened and we lost border in one of the demos:

magda-chrzescian:
> So this happened and we lost border in one of the demos. If that's on purpose, then I'm not exactly sure about the looks.

This is on purpose, I was out of ideas what to do to make these colours look ok. Let's have a third opinion about that and decide if we want to fix the borders. @AnnaTomanek WDYT?

godai78:
> This is on purpose

Yes, I only realised this after writing the comment. I'm not against it, it's just that I was startled at first.

magda-chrzescian:
> > This is on purpose
> 
> Yes, I only realised this after writing the comment. I'm not against it, it's just that I was startled at first.

We don't want our clients to be startled :( they should be impressed ;) However, I think we're just got used to our gray table borders.

AnnaTomanek:
> It is hard to recognize at the first sight, in which toolbar we should look for the buttons (We could call the toolbar "balloon/contextual".  
> It would be extremely helpful for the user if we could place the button icons next to their names ("Table properties", "Cell properties") in the text.

We actually have this functionality in [CKE4 docs](https://ckeditor.com/docs/ckeditor4/latest/features/index.html), see here:



This was a pre-existing feature going back to the previous version of the docs - we had a script that added these on the fly out of the sprites file for the editor toolbar/icons. I'm not sure how it was ported to Umberto (if this is still a script or just linked images) but I agree it's great - it's useful also when the reader uses a localized editor version and does not really see the same name that we are referring to. @pomek do you think we could implement it in some smart way? Can we have a follow-up issue for this?

godai78:
Would be splendid.

Issue #9779, opened 2021-05-28
upcastAttributeToMarker converter overrides the conversion data object even when there is no conversion match


ma2ciek:
This leads to a situation when adding a converter with the `editor.conversion.dataToMarker` upcast helper changes can change the conversion for some elements even if they do not match the view specified by the converter.

This behavior crashes the editor when the `Table` and either `Track Changes` or `Comments` are added to the editor (there're some additional prerequisites to reproduce the error though).

We should rethink if overriding the `data` properties for the whole element conversion in the  [https://github.com/ckeditor/ckeditor5/blob/523f88cfc55a920d931c784ab3469e94f749557e/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L690](https://github.com/ckeditor/ckeditor5/blob/523f88cfc55a920d931c784ab3469e94f749557e/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L690) is necessary. Currently adding any marker converter may change any element's upcast conversion - what's worse - it can be a race condition depending on the order of plugins in the editor.

As a temporary solution, we can check if any element can be consumed and return early for conversion calls for non-matching view elements - before the [https://github.com/ckeditor/ckeditor5/blob/523f88cfc55a920d931c784ab3469e94f749557e/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L685-L707](https://github.com/ckeditor/ckeditor5/blob/523f88cfc55a920d931c784ab3469e94f749557e/packages/ckeditor5-engine/src/conversion/upcasthelpers.js#L685-L707) lines.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity over the last year. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #9780, opened 2021-05-28
Added checks for the upcast attribute to marker converter before the view consumption

ma2ciek:

Fix (engine): Added checks for the upcast attribute to marker converter before changing the data and consuming view elements. Part of the #9779.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/tests/table-integration.js
+++ b/packages/ckeditor5-table/tests/table-integration.js
@@ -187,11 +187,13 @@ describe( 'Table feature – integration', () => {
 		} );
 
 		it( 'should not make the Model#hasContent() method return "true" when an empty table cell is selected', () => {
-			setModelData( editor.model, '<table>' +
-				'<tableRow>' +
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-table/tests/table-integration.js
```diff
@@ -187,13 +187,41 @@ describe( 'Table feature – integration', () => {
...
+					'</tableRow>' +
+				'</table>'
+			) );
 
 			expect( editor.model.hasContent( editor.model.document.selection.getFirstRange() ) ).to.be.false;
 		} );
 	} );
 } );
+
+describe( 'Table feature – integration #2', () => {
```
pomek, line 203: Not the best name for describing the test case. What is `#2`? I'd describe what it integrates.

scofalik:
Solution LGTM. Tests are a bit meh but this kind of stuff is quite difficult to test. I don't want to block that PR because of that.

PR #9812, opened 2021-06-02
Replaced badges in README.md for all packages.

### LINKED ISSUE #8938: Remove or replace deps and devDeps badges
I suppose david reads package.json from the old repo. IDK, maybe there's a way to make it read from https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/package.json. If not, we can consider a different badge system or remove it at all.

Side quest: we can consider adding a CI badge to all packages. We removed these once we merged the project into one repo. I think it's fine to have the same travis badge in all packages. To me, it looks "nice" if the project has badges. The next one to consider would be something for the CC.

image

image

przemyslaw-zan:

Docs: Replaced badges in README.md for all packages. Closes #8938.

---

### Additional information


pomek (approved):
LGTM.

PR #9822, opened 2021-06-04
Replaced GH Action for triggering Travis with custom script

### LINKED ISSUE #9336: Replace GH Action for triggering Travis with custom script

After removing the merging branches action, we still have one more that could be replaced with a solution that we use in private repositories.

przemyslaw-zan:

Internal: Replaced GH Action for triggering Travis with custom script. Closes #9336.

---

### Additional information


ma2ciek:
I would ensure first that the [scripts/trigger-ckeditor5-continuous-integration.sh](https://github.com/ckeditor/ckeditor5/pull/9822/files#diff-9b3441ac7cf51fb668e2a8602b4d6d78cc90bc45c0a542f10d9429e28b704bfe) file is executable by using `chmod +x`  and committing this file again.

scripts/trigger-ckeditor5-continuous-integration.sh
```diff
@@ -0,0 +1,32 @@
...
+  exit 0
+fi
+
+# Trigger the integration build only when checking the main branch in the repository.
+if [[ $TRAVIS_BRANCH != "master" ]]
+then
+  exit 0
+fi
+
+REPOSITORY="cksource/ckeditor5"
```
pomek, line 20:
```suggestion
REPOSITORY="ckeditor/ckeditor5"
```

.travis.yml
```diff
@@ -16,6 +16,7 @@ install:
 - yarn install
 script:
 - ./scripts/ci/travis-check.js
+- ./scripts/trigger-ckeditor5-continuous-integration.sh
```
pomek, line 19: Let's move the file to the `./scripts/ci/` directory.

pomek (changes requested):
As mentioned.

pomek:
Please, remove this file – https://github.com/ckeditor/ckeditor5/blob/master/.github/workflows/ckeditor5-integrations.yml.

.travis.yml
```diff
@@ -16,6 +16,7 @@ install:
 - yarn install
 script:
 - ./scripts/ci/travis-check.js
+- ./scripts/ci/trigger-ckeditor5-continuous-integration.sh
```
pomek, line 19: I'm wondering which build should trigger the script. We have three options:

1. Tests/Coverage
2. Validation (Linters/Manual tests)
3. Both.

The last option (3) will duplicate the job because, for a single commit, we will execute the same action twice.

The Validation step requires slightly much more time for checking everything so that I will go with the (1) option. 

Also, the Validation step may wait in the queue if we will have 3 builds (for different commits, e.g., PRs) at the same time.

pomek, line 19:
```suggestion
- if [[ $TRAVIS_JOB_TYPE == "Tests" && $TRAVIS_TEST_RESULT -eq 0 ]]; then ./scripts/ci/trigger-ckeditor5-continuous-integration.sh; fi
```

pomek (approved):
LGTM.

PR #9840, opened 2021-06-09
Added the option to remember the language when creating a new code block.

### LINKED ISSUE #8722: Remember the language when creating a new code block

If I am using CKEditor to write an article, most of the time, when creating a new code block, I want to use the same code block language that I used the previous time. Otherwise I have to correct every single code block from "Plain text" to whatever programming language I am writing about.

It would be great to have this as a (perhaps opt-out?) feature.

Notion does this and it's been pretty intuitive.

---

przemyslaw-zan:

Feature (code-block): When inserting a new code block, instead of applying the default language (the first in the dropdown view), the feature re-use the previously inserted language. Closes #8722.

### DIFF
--- a/packages/ckeditor5-code-block/src/codeblockcommand.js
+++ b/packages/ckeditor5-code-block/src/codeblockcommand.js
@@ -55,10 +55,11 @@ export default class CodeBlockCommand extends Command {
 	 *
 	 * @fires execute
 	 * @param {Object} [options] Command options.
-	 * @param {String|null} [options.language] TODO.
+	 * @param {String|null} [options.language] Language picked from dropdown.
 	 * @param {Boolean} [options.forceValue] If set, it will force the command behavior. If `true`, the command will apply a code block,
 	 * otherwise the command will remove the code block. If not set, the command will act basing on its current value.
-	 * @param {Boolean} [options.usePreviousLanguageChoice=false] TODO.
+	 * @param {Boolean} [options.usePreviousLanguageChoice=true] Defines whether or not button will create code block with the same
+	 * language as the previous code block.
 	 */
 	execute( options = {} ) {
 		const editor = this.editor;
--- a/packages/ckeditor5-code-block/src/codeblockcommand.js
+++ b/packages/ckeditor5-code-block/src/codeblockcommand.js
@@ -53,7 +53,10 @@ export default class CodeBlockCommand extends Command {
 
 		const blocks = Array.from( selection.getSelectedBlocks() );
 		const value = ( options.forceValue === undefined ) ? !this.value : options.forceValue;
-		const language = options.language || firstLanguageInConfig.language;
+		let language = options.language || firstLanguageInConfig.language;
+		if ( options.usePreviousLanguageChoice && options.lastLanguage ) {
+			language = options.lastLanguage;
+		}
 
 		model.change( writer => {
 			if ( value ) {
--- a/packages/ckeditor5-code-block/src/codeblockcommand.js
+++ b/packages/ckeditor5-code-block/src/codeblockcommand.js
@@ -25,7 +25,8 @@ export default class CodeBlockCommand extends Command {
 		super( editor );
 
 		/**
-		 * Remebers what language the last code block was created in.
+		 * Contains the last used language.
+
 		 * @protected
 		 * @type {String|null}
 		 */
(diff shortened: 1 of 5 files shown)


przemyslaw-zan:
A few tests are missing. I’ll add them, and once it is done, I’ll let you know.

packages/ckeditor5-code-block/src/codeblockcommand.js
```diff
@@ -41,8 +55,11 @@ export default class CodeBlockCommand extends Command {
 	 *
 	 * @fires execute
 	 * @param {Object} [options] Command options.
+	 * @param {String|null} [options.language] Language picked from dropdown.
```
pomek, line 58:
```suggestion
	 * @param {String} [options.language] The code block language.
```

1. When setting `options.language` to `null`, the command will use the default language. So it does not have to be specified. So, for example, `{String}` will be enough.
	* `[` and `]` marks the argument as optional, so if an integrator does not specify the value, the command will still work.
1. The language option does not have to come from the dropdown. The integrator can just execute the command directly: `editor.execute( 'codeBlock', { language: 'php' } )`. 
    * We do not have to focus on where the value comes from but what it does.

packages/ckeditor5-code-block/src/codeblockcommand.js
```diff
@@ -41,8 +55,11 @@ export default class CodeBlockCommand extends Command {
...
 	 * @param {Boolean} [options.forceValue] If set, it will force the command behavior. If `true`, the command will apply a code block,
 	 * otherwise the command will remove the code block. If not set, the command will act basing on its current value.
+	 * @param {Boolean} [options.usePreviousLanguageChoice=true] Defines whether or not button will create code block with the same
```
pomek, line 61:
```suggestion
	 * @param {Boolean} [options.usePreviousLanguageChoice] If set on `true` and the `options.language` is not specified, the command will apply the previous language (if the command was already executed) when inserting the `codeBlock` element.
```

1. `options.usePreviousLanguageChoice` is marked as optional (`[]`). If not specified, it's a false value (`undefined`).
1. The command itself does not know about the UI, so it does not know what the "button" is.

packages/ckeditor5-code-block/src/codeblockcommand.js
```diff
@@ -166,3 +185,26 @@ function canBeCodeBlock( schema, element ) {
...
+// @param {String} defaultLanguage
+// @return {String}
+function getLanguage( options, lastLanguage, defaultLanguage ) {
+	if ( options.usePreviousLanguageChoice && lastLanguage ) {
+		return lastLanguage;
+	}
+
+	if ( options.language ) {
+		return options.language;
+	}
```
pomek, line 207:
```suggestion
	if ( options.language ) {
		return options.language;
	}

	if ( options.usePreviousLanguageChoice && lastLanguage ) {
		return lastLanguage;
	}
```

I'd swap the order of these conditions. If the `options.language` is specified, it should be used. Please, do not update the function docs.

pomek: One more thing, make sure that the change is covered by a test (in the `CodeBlockCommand` file).

packages/ckeditor5-code-block/tests/codeblockcommand.js
```diff
@@ -276,6 +276,44 @@ describe( 'CodeBlockCommand', () => {
...
+			} );
+
+			it( 'it should apply the default language when the last language is not set yet', () => {
+				setModelData( model, '<paragraph>fo[]o</paragraph>' );
+
+				command.execute( { usePreviousLanguageChoice: true } );
+
+				expect( getModelData( model ) ).to.equal( '<codeBlock language="plaintext">fo[]o</codeBlock>' );
+			} );
+		} );
```
pomek, line 316: We need one more test that will cover a change described in the `getLanguage()` function.

```js
# Mock the "last language"
command._lastLanguage = 'css';

# Then execute the command, but with the specified language:
command.execute( { language: 'php', usePreviousLanguageChoice: true } );

# Expecting that the codeBlock will use `php` as the language.
```

I agree these options are mutually exclusive, but we can handle it and use the provided language.

pomek (changes requested):
As mentioned.

packages/ckeditor5-code-block/src/codeblockcommand.js
```diff
@@ -41,8 +56,12 @@ export default class CodeBlockCommand extends Command {
 	 *
 	 * @fires execute
 	 * @param {Object} [options] Command options.
+	 * @param {String} [options.language] The code block language.
 	 * @param {Boolean} [options.forceValue] If set, it will force the command behavior. If `true`, the command will apply a code block,
 	 * otherwise the command will remove the code block. If not set, the command will act basing on its current value.
+	 * @param {Boolean} [options.usePreviousLanguageChoice] If set on `true` and the `options.language` is not specified, the command
+	 * will apply the previous language (if the command was already executed) when inserting the `codeBlock` element.
+	 * language as the previous code block.
```
pomek, line 64:
```suggestion
```

pomek (approved):
LGTM.

PR #9860, opened 2021-06-13
Removed the unnecessary resetting of the last element's bottom margin.

### LINKED ISSUE #9825: The demo content in the documentation has fixed margin bottom for the last-of-type elements.

The following rule: https://github.com/ckeditor/ckeditor5/blob/e58ff6a8d4741a81d9a879770788c14f16f1849d/docs/assets/snippet-styles.css#L32-L34
forces the sie of the bottom margin of every last-of-type element in the editor content. It breaks the behavior of content with custom styling (has to be overwritten with `!important` at this point.

It was introduced in https://github.com/ckeditor/ckeditor5/pull/8550.

This is an example of such behavior: there is no margin between the image and the next paragraph, despite it is directly set with the CSS (as it is the `last-of-type` element in the editor content):

image

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

magda-chrzescian:

Docs (image): The last element of the editor content should not have the forced `margin: 0` CSS rule. Closes #9825.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-image/theme/image.css
+++ b/packages/ckeditor5-image/theme/image.css
@@ -10,7 +10,7 @@
 		text-align: center;
 
 		/* Make sure there is some space between the content and the image. Center image by default. */
-		margin: 1em auto;
+		margin: var(--ck-spacing-large) auto;
 
 		/* Make sure the caption will be displayed properly (See: https://github.com/ckeditor/ckeditor5/issues/1870). */
 		min-width: 50px;
--- a/packages/ckeditor5-image/theme/image.css
+++ b/packages/ckeditor5-image/theme/image.css
@@ -10,7 +10,7 @@
 		text-align: center;
 
 		/* Make sure there is some space between the content and the image. Center image by default. */
-		margin: var(--ck-spacing-large) auto;
+		margin: 0.9em auto;
 
 		/* Make sure the caption will be displayed properly (See: https://github.com/ckeditor/ckeditor5/issues/1870). */
 		min-width: 50px;
--- a/packages/ckeditor5-image/theme/image.css
+++ b/packages/ckeditor5-image/theme/image.css
@@ -10,6 +10,8 @@
 		text-align: center;
 
 		/* Make sure there is some space between the content and the image. Center image by default. */
+		/* The first value should be equal to --ck-spacing-large variable if used in the editor context
+	 	to avoid the content jumping (See https://github.com/ckeditor/ckeditor5/issues/9825). */
 		margin: 0.9em auto;
 
 		/* Make sure the caption will be displayed properly (See: https://github.com/ckeditor/ckeditor5/issues/1870). */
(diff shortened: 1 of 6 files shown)


oleq:
I just noticed that both this branch and nightly (master) suffer from the content jumping problem which does not exist in stable [https://ckeditor.com/docs/ckeditor5/latest/examples/builds/inline-editor.html](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/inline-editor.html)



Can you check this out?

oleq (commented):
Let's first figure out what's going on in https://github.com/ckeditor/ckeditor5/pull/9860#issuecomment-860587971

oleq (approved):
Guess it's alright :)

magda-chrzescian:
@oleq There was also a problem with jumping content when a widget was the last element in the editable. It also occurred on `stable`, since it wasn't related to the Umberto styles, but inconsistencies between the default widgets margins and the margins applied to the first/last elements.

packages/ckeditor5-image/theme/image.css
```diff
@@ -10,7 +10,7 @@
 		text-align: center;
 
 		/* Make sure there is some space between the content and the image. Center image by default. */
-		margin: 1em auto;
```
oleq, line 13: We can't use `var(--ck-spacing-large)` in [content styles](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/content-styles.html) because they could be used in the front-end where CKEditor is not loaded at all and there are no custom property definitions :(

Same with other custom properties such as colors, fonts, etc.

packages/ckeditor5-table/theme/table.css
```diff
@@ -5,7 +5,7 @@
 
 .ck-content .table {
 	/* Give the table widget some air and center it horizontally */
-	margin: 1em auto;
```
oleq, line 8: Same as `.ck-content .image`.

packages/ckeditor5-media-embed/theme/mediaembed.css
```diff
@@ -9,7 +9,7 @@
 	clear: both;
 
 	/* Make sure there is some space between the content and the media. */
-	margin: 1em 0;
```
oleq, line 12: Same as `.ck-content .image`.

oleq (changes requested):

packages/ckeditor5-image/theme/image.css
magda-chrzescian: You're right, thanks.
I guess we should just stay with the fixed values (committed).
The other way was to override the content rule in the internal editor CSS stylesheet, but I guess the differences is displaying content in the editor and on the webpage may be confusing.

magda-chrzescian: Something else came to my mind.
We can also use the fallback value for the variable:
`margin: var(--ck-spacing-large, 0.9em) auto`.
This way it will work properly in the editor and outside of it. But I'm not sure if we gain anything from it. It introduces some external variables in content styles (another thing the integrator has to understand) and doesn't say more than a good comment about why this code looks like this.

Issue #9857, opened 2021-06-13
Source editing should format the source before showing it to the user


Reinmar:
## 📝 Provide a description of the improvement

Currently, it displays like this:

![](https://user-images.githubusercontent.com/156149/121806689-2af44180-cc51-11eb-91e7-a027b922986d.png)

Which makes it very hard to read and edit.

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

Reinmar:
IMO, this is a blocker for the GHS and source editing releases.

PR #9864, opened 2021-06-14
[GHS] First draft of the docs

### LINKED ISSUE #9854: [GHS] Feature guide for General HTML support
I think a feature guide would be a good starting point. It needs to be well thought how and where we link to it, but I'll skip this for now.

Important: GHS will be in a "beta" for now, so it actually makes sense to avoid linking to it from too many places. We want people to be able to play with it, but not to focus too broad audience on it.

Reinmar:

Docs (html-support): Feature guide for the GHS feature. Closes #9854.

---

### Additional information

*   This is a draft. It needs more work – mainly review, clarifications and synchronization with the works from @jacekbogdanski and @maxbarnas.
*   Please review the format of the `allowed` and `disallowed` rules. I wrote it from the top of my head. One thing that I don't want to cover there is the `{key, value}` notation. It can be shown in the examples below, but not in the notation, to keep it simpler.
*   I tried to come up with a reasonable demo, but it wasn't simple. I'm sure it can be done better, but I didn't want to spend too much time on it now. Also, I could not make it work due to the recent changes in GHS itself, so I didn't test it and it for sure needs updates and review.
*   When updating the demo, make sure to update also the snippet in the MD file, not only in the JS file.
*   I found #9857 to be a blocker for this guide. This is - it can be done later but must be done before the release.
*   We need to point people to GH, so e.g. when we mention that extending output of the table feature is not possible yet, we should link to an existing ticket on GH. Hence, these tickets need to be opened.
*   The entire guide and snippet should be moved to `packages/html-support/docs/`. I was working on a `master` so I couldn't push them to the correct place.
*   And once again – it's really a draft :D I wanted to write anything quickly, without giving it that much thought to not block the rest of the development. I'm sure it requires thorough review later on – perhaps once we'll be stabilizing this.


Reinmar:
There's also a discussion in #9854 regarding the name of the feature. I also pointed out its close relation to the source editing feature. In a longer term, a decision for the direction of these guides from the CCX squad may be necessary. Although, for now, taken that GHS is experimental, having a separate guide for it seems to be necessary anyway.

godai78:
We should preferably not describe the feature within the demo, but showcase this instead. I will need some time, however, to come up with a nice demo for that.

jacekbogdanski:
I've corrected technical issues like a base branch, config options, API links etc. The demo should start correctly now.

psmyrek:
I've added missing `@ckeditor/ckeditor5-html-support` dependency in main `package.json` because this prevents building the documentation.

godai78 (approved):

godai78:
Cool.

packages/ckeditor5-html-support/docs/features/general-html-support.md
```diff
@@ -0,0 +1,212 @@
+---
+category: features
+modified_at: 2021-06-13
+updated_at: 2021-06-14
+---
+
+# General HTML support
+
+The general HTML support ("GHS") feature allows the user to easily enable HTML features that are not explicitly supported by any other dedicated CKEditor 5 plugins.
```
jacekbogdanski, line 9:
```suggestion
The general HTML support ("GHS") feature allows developers to easily enable HTML features that are not explicitly supported by any other dedicated CKEditor 5 plugins.
```
The feature integrator is a developer, not a user.

packages/ckeditor5-html-support/docs/features/general-html-support.md
```diff
@@ -0,0 +1,212 @@
...
+
+Some examples of HTML features that can be easily enabled using general HTML support include:
```
jacekbogdanski, line 11:
```suggestion
Some examples of HTML features that can be easily enabled using General HTML Support include:
```
Since we are using GHS abbreviation everywhere in the guide, I think it makes sense to use the upper case feature name.

packages/ckeditor5-html-support/docs/features/general-html-support.md
```diff
@@ -0,0 +1,212 @@
...
+
+<info-box>
+	The General HTML support feature is **experimental and not yet production-ready**.
+
+	Follow the ["Stabilize and release a production-ready General HTML support feature"](https://github.com/ckeditor/ckeditor5/issues/9856) issue for more updates and related issues.
+</info-box>
+
+## Demo
+
+The general HTML support feature is configured via the `config.htmlSupport` property. In it, you need to list the HTML features that should be handled by GHS.
```
jacekbogdanski, line 30:
```suggestion
The General HTML Support feature is configured via the `config.htmlSupport` property. In it, you need to list the HTML features that should be handled by GHS.
```

packages/ckeditor5-html-support/docs/features/general-html-support.md
```diff
@@ -0,0 +1,212 @@
...
+
+For instance, the dedicated {@link features/basic-styles#available-text-styles bold} feature offers a toolbar button used to make the selected text bold. Together with the {@link features/autoformat autoformatting feature}, it also allows for applying bold style to content by typing a Markdown shortcode (`**foo**`) in the editor. The {@link features/headings headings} feature offers a dropdown from which the user can choose a heading level and ensures that pressing <kbd>Enter</kbd> at the end of a heading creates a new paragraph (and not another heading).
+
+The general HTML support does not offer any UI for the enabled features and takes only basic semantics of a given feature into account. If you enable support for `<div>`s via GHS, the user will not be able to create `<div>`s from the editor UI. The GHS will know that a `<div>` is a container element, so it can wrap other blocks (like paragraphs) but cannot be used inline (next to e.g. a `<strong>` element). There is no more, however.
```
jacekbogdanski, line 86:
```suggestion
The General HTML Support does not offer any UI for the enabled features and takes only the basic semantics of a given feature into account. If you enable support for `<div>`s via GHS, the user will not be able to create `<div>`s from the editor UI. The GHS will know that a `<div>` is a container element, so it can wrap other blocks (like paragraphs) but cannot be used inline (next to e.g. a `<strong>` element). There is no more, however.
```

packages/ckeditor5-html-support/docs/features/general-html-support.md
```diff
@@ -0,0 +1,212 @@
...
+		// Adds support for style="color: *" to the already supported
+		// <p> and <h2-h4> elements.
+		{
+			name: /^(p|h[2-4])$/',
+			styles: { 'color': true }
+		},
+}
+```
+
+The general HTML support feature distinguishes several content types that it treats a bit differently:
```
jacekbogdanski, line 184:
```suggestion
The General HTML Support feature distinguishes several content types that it treats a bit differently:
```

jacekbogdanski (approved):
🎉

Issue #9797, opened 2021-05-31
Source editing: map the cursor position from the editing view to the "offset" in the source view


psmyrek:
## 📝 Provide a description of the new feature

A follow-up to https://github.com/ckeditor/ckeditor5/pull/9736.

Currently, after switching to source editing mode, `<textarea>` containing the document source is focused and the cursor is placed at the end of the content.

Proposal: map the cursor position from the editing view to the "offset" in the source view.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity over the last year. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #9875, opened 2021-06-14
Added document content formatting before entering the source mode

### LINKED ISSUE #9857: Source editing should format the source before showing it to the user

psmyrek:

Internal: Added document content formatting before entering the source mode. Closes #9857.

### DIFF
--- a/packages/ckeditor5-source-editing/src/sourceediting.js
+++ b/packages/ckeditor5-source-editing/src/sourceediting.js
@@ -198,7 +198,7 @@ export default class SourceEditing extends Plugin {
 		// It is not needed to iterate through all editing roots, as currently the plugin supports only the Classic Editor with a single
 		// main root, but this code may help understand and use this feature in external integrations.
 		for ( const [ rootName, domRootElement ] of editingView.domRoots ) {
-			const data = editor.data.get( { rootName } );
+			const data = formatSource( editor.data.get( { rootName } ) );
 
 			const domSourceEditingElementTextarea = createElement( domRootElement.ownerDocument, 'textarea', { rows: '1' } );
 
@@ -241,7 +241,7 @@ export default class SourceEditing extends Plugin {
 		const data = {};
 
 		for ( const [ rootName, domSourceEditingElementWrapper ] of this._replacedRoots ) {
-			const oldData = editor.data.get( { rootName } );
+			const oldData = formatSource( editor.data.get( { rootName } ) );
 			const newData = domSourceEditingElementWrapper.dataset.value;
 
 			// Do not set the data unless some changes have been made in the meantime.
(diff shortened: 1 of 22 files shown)


packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -335,3 +335,105 @@ export default class SourceEditing extends Plugin {
...
+		{ name: 'li', isVoid: false },
+		{ name: 'ol', isVoid: false },
+		{ name: 'p', isVoid: false },
+		{ name: 'table', isVoid: false },
+		{ name: 'tbody', isVoid: false },
+		{ name: 'td', isVoid: false },
+		{ name: 'th', isVoid: false },
+		{ name: 'thead', isVoid: false },
+		{ name: 'tr', isVoid: false },
+		{ name: 'ul', isVoid: false }
```
ma2ciek, line 366: Defining an array of all elements that should be formatted is quite risky to me. Note that e.g. `` is missing here. And this array should be actively maintained in the future. 

I'd rather collect only all possible tags that are empty and handle the rest tags in a more generic way.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -335,3 +335,105 @@ export default class SourceEditing extends Plugin {
...
+
+	const elementNamesToFormat = elementsToFormat.map( element => element.name ).join( '|' );
+
+	const lines = input
+		// Add new line before `<tag>` or `</tag>`, but only if it is not already preceded by a new line (negative lookbehind).
+		.replace( new RegExp( `(?<!\n)</?(${ elementNamesToFormat })( .*?)?>`, 'g' ), '\n$&' )
+		// Add new line after `<tag>` or `</tag>`, but only if it is not already followed by a new line (negative lookahead).
+		.replace( new RegExp( `</?(${ elementNamesToFormat })( .*?)?>(?!\n)`, 'g' ), '$&\n' )
+		// Divide input string into lines, which start with either an opening tag, a closing tag, or just a text.
+		.split( '\n' );
```
ma2ciek, line 377: I think that this may be slow for large documents with the `elementNamesToFormat` being an alternation of many strings. But I might be wrong.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -241,7 +241,7 @@ export default class SourceEditing extends Plugin {
 		const data = {};
 
 		for ( const [ rootName, domSourceEditingElementWrapper ] of this._replacedRoots ) {
-			const oldData = editor.data.get( { rootName } );
+			const oldData = formatSource( editor.data.get( { rootName } ) );
```
ma2ciek, line 244: I missed it previously when I wrote it :sweat\_smile: But perhaps it would be good to store the roots data to not read them twice (`editor.data.get()` calls can be quite expensive). I'll create an issue for that.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -241,7 +241,7 @@ export default class SourceEditing extends Plugin {
...
 			const newData = domSourceEditingElementWrapper.dataset.value;
```
ma2ciek, line 245: We could use the formatter here as well. I'm not sure if the current one would work for the already converted data.

packages/ckeditor5-source-editing/src/sourceediting.js
ma2ciek: Maybe we could extend it with at least all defined block elements - https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements.

ma2ciek:
I'd add at least a few tests for the beautifier.

*   nested block elements
*   nested block + inline elements
*   elements + text

Perhaps it would be good to extract this utility to a separate file in order to write make this part of the code testable without the overhead of creating the editor.

packages/ckeditor5-source-editing/src/sourceediting.js
ma2ciek: `isVoid` -> `isEmpty` ? I guess this is the proper naming for these elements.

psmyrek: I will add all block elements to this list.

> `isVoid` -> `isEmpty` ? I guess this is the proper naming for these elements.

According to W3C naming convention such elements are "void": https://html.spec.whatwg.org/multipage/syntax.html#void-elements. The term "empty element" is used to describe an element that contains no content. Elements of any type can be empty, depending only on their content model (i.e. `<span></span>` is an empty element). So IMHO we should keep the `isVoid` property as it is now.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: I've tested the performance of `formatSource()` function with generated random HTML content (paragraphs, tables, lists, etc.):

| HTML content size | Execution time |
| --- | --- |
| 100 KB | ~100 ms |
| 200 KB | ~190 ms |
| 500 KB | ~460 ms |
| 1 MB | ~900 ms |

IMHO these results are good enough.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Good idea. You don't need to create a separate issue for that, I will add this simple improvement in this PR 🙂

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Unfortunately, using a formatter twice (for already converted data) will not work here. Moreover, preparing the reverse operation for formatting (de-formatting? 😅) would be non-trivial and error-prone, so I think we should just compare the initial (formatted) data with the current one, whatever it is at the time of exiting from the source editing mode.

packages/ckeditor5-source-editing/src/sourceediting.js
ma2ciek: I took this name from [https://developer.mozilla.org/en-US/docs/Glossary/Empty_element.](https://developer.mozilla.org/en-US/docs/Glossary/Empty_element.) But we can use void if this is a good name for these elements. My main concern here is rather https://github.com/ckeditor/ckeditor5/pull/9875#discussion_r650951288.

pomek: One more option to consider: `isSelfClosing`.

psmyrek: A self-closing tag ends with `/>` which is not true for void elements. Our editor produces `<br>` but not `<br />`.

packages/ckeditor5-source-editing/src/sourceediting.js
ma2ciek: In this case it would be good to add a proper comment to the formatting function :)

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: > Maybe we could extend it with at least all defined block elements - https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements.

Because we shouldn't format all elements, we have to define somehow list of elements, that will be formatted. I don't know how it could have been done otherwise than introducing an array of them in the code.

psmyrek:
I extracted the source code formatter to `./utils/formatsource.js` and I've added tests for it.

Also I've added a check if the document source is HTML and only in this case the formatter is used.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -12,6 +12,7 @@
 import { Plugin, PendingActions } from 'ckeditor5/src/core';
 import { ButtonView } from 'ckeditor5/src/ui';
 import { createElement, ElementReplacer } from 'ckeditor5/src/utils';
+import { formatSource } from './utils/formatsource';
```
ma2ciek, line 15: What do you think about renaming it to `formatHtml`? This would allow moving this function someday to the `ckeditor5-utils` when there would be a need to use it elsewhere.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -269,14 +282,17 @@ export default class SourceEditing extends Plugin {
...
 	 * @private
 	 */
 	_focusSourceEditing() {
 		const [ domSourceEditingElementWrapper ] = this._replacedRoots.values();
 
-		domSourceEditingElementWrapper.querySelector( 'textarea' ).focus();
+		const textarea = domSourceEditingElementWrapper.querySelector( 'textarea' );
+
+		textarea.focus();
+		textarea.selectionEnd = 0;
```
ma2ciek, line 295: Is there a ticket for it? I guess that this PR fixes something else as well.

psmyrek: I have not found a ticket, but IMHO it would be cool if switching between modes set the cursor in the same way: at the beginning of the content. I think that such a small improvement does not need to be extracted to a separate issue.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Done.

packages/ckeditor5-source-editing/src/sourceediting.js
ma2ciek: Currently, it takes about half of the time of `editor.data.get()`:

```
Editor.getData() 109
formatSource() 55

Editor.getData() 85
formatSource() 55
```

packages/ckeditor5-source-editing/src/sourceediting.js
ma2ciek, line 295: With this change there's a problem as the textarea can be quite big, the window is scrolled to the bottom, but the selection is placed on top of the textarea.

packages/ckeditor5-source-editing/src/sourceediting.js
ma2ciek: I'm not sure if it's OK or NOK but it can be easily improved later if this would be a problem.

packages/ckeditor5-source-editing/src/sourceediting.js
```diff
@@ -198,7 +207,7 @@ export default class SourceEditing extends Plugin {
 		// It is not needed to iterate through all editing roots, as currently the plugin supports only the Classic Editor with a single
 		// main root, but this code may help understand and use this feature in external integrations.
 		for ( const [ rootName, domRootElement ] of editingView.domRoots ) {
-			const data = editor.data.get( { rootName } );
+			const data = formatSource( editor.data.get( { rootName } ) );
```
ma2ciek, line 210: I was thinking if the formatter should be configurable but we agreed with @psmyrek that this configurability can be added in the future.

packages/ckeditor5-source-editing/src/sourceediting.js
psmyrek: Only in Chrome. In Firefox it works as expected.
But still this should be fixed.

psmyrek: I found some hacky ways to prevent Chrome from scrolling to the bottom, but then I decided to remove this behavior at all. This behavior might be caused by an old issue in Chrome, that the text selection or the text cursor is not made visible after the selection is changed from JavaScript code (https://bugs.chromium.org/p/chromium/issues/detail?id=331233).

We have already an issue to map the cursor position from the editing view to the "offset" in the source view (https://github.com/ckeditor/ckeditor5/issues/9797).

magda-chrzescian (approved):
Looks good to me :)

PR #9889, opened 2021-06-16
Fixed the selection issues

### LINKED ISSUE #9513: Nested marker highlight breaks mouse text selection

*   [http://fake.ckeditor.com:8125/ckeditor5-engine/tests/manual/markers.html](http://fake.ckeditor.com:8125/ckeditor5-engine/tests/manual/markers.html)
*   Set markers so that they overlap:

*   try selecting text (with the mouse) starting inside the nested marker (do it quickly, don't keep the mouse button pressed in one place):

### ✔️ Expected result

Selection is not collapsed and where the mouse button was pressed and ends where it was released.

### ❌ Actual result

Selection is collapsed at the end of the expected selection range.

---

What I discovered while debugging it:

*   The `selectionchange` DOM event is not providing a proper range, it's already collapsed at this point
*   The DOM structure is split in a crazy way if the caret is inside the nested highlight:

While with caret outside the marker looks like this:

I suppose that the problem might be caused by the replacement of the DOM nodes on the first selection change that causes this issue (the original node is removed and replaced with other nodes).

## 📃 Other details

*   Browser: Chrome

---

niegowski:

Fix (engine): Markers should not be split in view on the caret position. Closes #9513.

Fix (engine): `FocusObserver` should not force the view to render in random moments. See #9513.

---

### Additional information

This PR is fixing the two above cases. It's not handling the DOM structure changing while selecting that was also discussed in the issue.

### DIFF
--- a/packages/ckeditor5-engine/src/view/observer/focusobserver.js
+++ b/packages/ckeditor5-engine/src/view/observer/focusobserver.js
@@ -7,8 +7,6 @@
  * @module engine/view/observer/focusobserver
  */
 
-/* globals setTimeout, clearTimeout */
-
 import DomEventObserver from './domeventobserver';
 
 /**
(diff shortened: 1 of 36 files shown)


niegowski:
@ckeditor/qa-team could you play with it and check if it helps? It doesn't solve all issues but should help in cases when the marker is highlighted and then selecting inside it.

Mgsy:
I've tested the solution and seems to work fine, I couldn't find any issues :tada: .

packages/ckeditor5-engine/src/view/observer/focusobserver.js
```diff
@@ -37,7 +37,10 @@ export default class FocusObserver extends DomEventObserver {
 			// overwrite new DOM selection with selection from the view.
 			// See https://github.com/ckeditor/ckeditor5-engine/issues/795 for more details.
 			// Long timeout is needed to solve #676 and https://github.com/ckeditor/ckeditor5-engine/issues/1157 issues.
-			this._renderTimeoutId = setTimeout( () => view.forceRender(), 50 );
+			//
+			// Changing document.isFocused marks view as changed since last rendering, it would be best if view got rendered
+			// on the next selectionChange event, but as a fallback let's trigger rendering in 50ms.
```
scofalik, line 42:
```suggestion
			// Using `view.change()` instead of `view.forceRender()` to prevent double rendering
			// in a situation where `selectionchange` already caused selection change.
```

scofalik: Is this comment true and better?

scofalik (changes requested):

PR #9891, opened 2021-06-16
[GHS] Support for div-like elements

### LINKED ISSUE #10085: [GHS] Support for div-like elements

Div-like elements can behave both as sectioning element and block element:

*   block content - allows only text and phrasing content, similar to paragraph

```
<div>
	Hello <strong>World!</strong>
</div>
```

*   sectioning content - defines section scope, similar to the article, section, etc. Can't directly include text and phrasing content

```
<div>
	<h2>Title</h2>
	<p>Hello, World!</p>
</div>
```

Similar features: `blockquote`

We should distinguish this double nature of div-like elements by separating them in the model schema and downcasting into a single element. It will require some smart API when registering data schema definition also.

jacekbogdanski:

Feature (html-support): Added support for elements that can behave both as a paragraph and sectioning element depending on the content context. Closes #10085.

---

### Additional information

I've slightly reversed the initial approach to treat special elements as paragraph-like, not div-like, so the feature has a more broad context.

I went with `asParagraph` name for the paragraph-like model, but I suppose it should be something less boolean-ish, maybe just `textModel`?


niegowski (commented):
Partial review.

jacekbogdanski:
Rebased onto the latest `master`.

jacekbogdanski:
I've simplified the whole implementation to use the view element as a source of the information on which model should be used in the upcast conversion. For now, I went with `DualContentModelElementSupport` name for the class name and `textModel` for model property in schema definition as the naming looked the most understandable for me. I've also changed the name of `CodeBlockHtmlSupport` to `CodeBlockElementSupport` which also seems easier to look at.

jacekbogdanski:
image
I will fix CI in a moment.

packages/ckeditor5-html-support/src/integrations/dualcontent.js
```diff
@@ -0,0 +1,134 @@
...
+
+	init() {
+		const dataFilter = this.editor.plugins.get( DataFilter );
+
+		dataFilter.on( 'register', ( evt, definition ) => {
+			const editor = this.editor;
+			const schema = editor.model.schema;
+			const conversion = editor.conversion;
+
+			if ( !definition.textModel ) {
```
jacekbogdanski, line 52: Let's rename it to `paragraphLikeModel`.

PR #9897, opened 2021-06-17
Fire `addMarker` event in data pipeline if marker is at element boundary

### LINKED ISSUE #8485: addMarker event is not called in data pipeline when range is at start of the root element

1. Open https://codepen.io/jacekbogdanski/pen/XWKwRjB?editors=1010
2. Open developer tools console.
3. Refresh the page.

### ✔️ Expected result

Debugger statement for `addMarker` event should stop script execution.

### ❌ Actual result

Nothing happens.

The issue seems to be caused by the https://github.com/ckeditor/ckeditor5/blob/eee94cef569a159ba1e59482d571e2413dde575c/packages/ckeditor5-engine/src/controller/datacontroller.js#L491-L510 function executed in https://github.com/ckeditor/ckeditor5/blob/eee94cef569a159ba1e59482d571e2413dde575c/packages/ckeditor5-engine/src/controller/datacontroller.js#L240-L248

It looks like when the marker range is at the start or end of the root it won't be contained in the root element range and the following code will fail to process markers.

---

psmyrek:

Fix (engine): Fire `addMarker` event in data pipeline if collapsed marker is at element boundary. Closes #8485.


packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -250,7 +250,7 @@ export default class DataController {
 		// Convert markers.
 		// For document fragment, simply take the markers assigned to this document fragment.
 		// For model root, all markers in that root will be taken.
-		// For model element, we need to check which markers are intersecting with this element and relatively modify the markers' ranges.
+		// For model element, we need to check which markers are contained in this element and relatively modify the markers' ranges.
```
niegowski, line 253: This comment change is not entirely true - markers that are partly in that element are also included.

packages/ckeditor5-engine/src/controller/datacontroller.js
```diff
@@ -541,10 +542,16 @@ function _getMarkersRelativeToElement( element ) {
 	const elementRange = ModelRange._createIn( element );
 
 	for ( const marker of doc.model.markers ) {
-		const intersection = elementRange.getIntersection( marker.getRange() );
+		const markerRange = marker.getRange();
 
-		if ( intersection ) {
-			result.push( [ marker.name, intersection ] );
+		const isMarkerCollapsedAtElementBoundary = markerRange.isCollapsed && ( markerRange.start.isAtStart || markerRange.start.isAtEnd );
```
niegowski, line 547:
```suggestion
		const isMarkerCollapsedAtElementBoundary = markerRange.isCollapsed && ( markerRange.start.isEqual( elementRange.start) || markerRange.start.isEqual( elemenetRange.end ) );
```

packages/ckeditor5-engine/tests/controller/datacontroller.js
```diff
@@ -671,6 +671,78 @@ describe( 'DataController', () => {
...
+				writer.addMarker( 'fooMarkerAtElementEnd', {
+					range: writer.createRange(
+						writer.createPositionFromPath( root, [ 1 ] ),
+						writer.createPositionFromPath( root, [ 1 ] )
+					),
+					usingOperation: true
+				} );
+			} );
+
+			data.toView( root );
```
niegowski, line 702: Let's check how this looks in the view as in other tests it's done.

packages/ckeditor5-engine/src/controller/datacontroller.js
niegowski: It's worth adding tests for the previous implementation - with a collapsed marker that is at the start or end of some other element (not included in the checked element) and for a collapsed marker at the start or end inside some deeper nested element.

niegowski (changes requested):

packages/ckeditor5-engine/src/controller/datacontroller.js
psmyrek: Added.

packages/ckeditor5-engine/tests/controller/datacontroller.js
psmyrek: I rewrote the tests to make them look similar to others. I didn't use the `markerToHighlight` conversion, because it seems that it does not allow creating a collapsed marker. Instead, I'm using now `markerToData` helper.

niegowski (approved):

PR #9902, opened 2021-06-17
Bugfix for mention plugin, trivial, mentionId input validation

platowski:
Type (ckeditor5-mention): Added missing validation check, so non-string input won't trigger **typeError**, but let the proper exception be thrown.

---

### Additional information

As per: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/contributing.html
I skipped creating a separate GH issue for such a trivial change.

### DIFF
--- a/packages/ckeditor5-mention/src/mentioncommand.js
+++ b/packages/ckeditor5-mention/src/mentioncommand.js
@@ -104,7 +104,7 @@ export default class MentionCommand extends Command {
 			);
 		}
 
-		if ( mentionID.charAt( 0 ) != options.marker ) {
+		if ( typeof mentionID !== 'string' || mentionID.charAt( 0 ) != options.marker ) {
 			/**
 			 * The feed item ID must start with the marker character.
 			 *


Mgsy:
Hello @platowski! Thank you for the pull request and sorry for the late response - we are in the middle of the release and we had a busy period. 

We are always happy to review changes proposed by external contributors, however, in order to do that, at first we should understand why the proposed change is needed :) It would be appreciated if you could create a new issue with a description of the problem and provide a scenario when the current solution fails. Thanks!

platowski:
@Mgsy thanks for the reply and no worries, I guess everyone has an end of the quarter now. 
Regarding your comment. I stated in the description that in the linked procedure there is a statement that for trivial changes creating an issue is not required. 
I'll allow myself to reword the original description first:
Current code of mention plugin lacks input validation. If you provide non-string input i.e. integer 1 it will explode with typeError that shows nothing but a stack trace. Instead, it should throw CKEditorError (see next line of code, after comments). 
Error is as trivial as getting [0] element from the array without checking if it's not undefined first.
That is why I skipped creating an issue. :)
Anyway, if you still think that requires a separate issue with an explanation I'm happy to do it.

Reinmar:
> Current code of mention plugin lacks input validation. If you provide non-string input i.e. integer 1 it will explode with typeError that shows nothing but a stack trace.

A generic type error + a stack trace may indeed not be easy to understand, especially if someone doesn't have source maps configured, The error that will be thrown now, in my opinion, is not much better either as it's focused on the first character. Also, the actual value is not logged.

The easiest option would be to update its documentation (to make the error a bit more generic like saying "it must be a string that starts with") and adding the actual value as the third param of CKEditorError (end documenting what it is).

Finally, we require to cover every change in the code with a test. The idea is – someone may mistakenly remove your addition in the future when rewriting this code and the test should catch that.

PR #9909, opened 2021-06-18
[GHS] Add support for more elements

jacekbogdanski:
### Additional information

I've added support for most elements supported by CKE4. Note that some elements are duplicated with CKE5 features - it's done on purpose, so e.g. we can still enable headings using GHS, even if we are not using the dedicated Headings feature.

I will create a follow-up ticket for adding unit tests for all HTML elements, as for now, we are just using chosen elements for integration testing.


jacekbogdanski:
TODO we should rename `$htmlBlock` into something else, as `$block` element is used to describe `paragraph`. In the case of GHS, `$htmlBlock` describes the sectioning element. We could go with e.g. `$htmlSection` or `$htmlContainer` to better distinguish it from `$block`.

packages/ckeditor5-html-support/tests/manual/features.html
```diff
@@ -0,0 +1,123 @@
...
+					</tr>
+				</tbody>
+			</table>
+
+			<figure class="media" data-foo><oembed data-bar url="https://www.youtube.com/watch?v=ZVv7UMQPEWk"></oembed></figure>
+
+			<div style="border: 1px solid blue" class="raw-html-embed">HTML snippet</div>
+
+			<div style="border: 1px solid blue" class="page-break" style="page-break-after:always;"><span style="display:none;">&nbsp;</span></div>
+		</div>
```
jacekbogdanski, line 113: We are missing schema definitions for the Page Break and HTML Snippet features. I thought it will require some special integration but I did some quick check and it seems like it may work out of the box. I will check out it next week.

maxbarnas:
Looks good to me. Manual tests are working fine. As I mentioned offline, I am not feeling confident enough to confirm the current schema though. It would be great to have @Reinmar look at the `schemadefinitions.js`.

packages/ckeditor5-html-support/src/schemadefinitions.js
```diff
@@ -41,8 +41,7 @@
 // * title
 // * head
 // * meta
-// * link
-// * etc...
+// * link * etc...
```
maxbarnas, line 44: Are those line merged on purpose?

packages/ckeditor5-html-support/tests/datafilter.js
```diff
@@ -317,6 +317,34 @@ describe( 'DataFilter', () => {
...
+				.getChild( 0 ); // <input>
+
+			expect( input.getAttribute( 'type' ) ).to.equal( 'number' );
+		} );
+
+		it( 'should consume htmlAttributes attribute (editing downcast)', () => {
+			editor.conversion.for( 'editingDowncast' ).add( dispatcher => {
+				dispatcher.on( 'attribute:htmlAttributes:htmlInput', ( evt, data, conversionApi ) => {
+					conversionApi.consumable.consume( data.item, evt.name );
+					expect( conversionApi.consumable.test( data.item, evt.name ) ).to.be.true;
```
maxbarnas, line 338: Shouldn't it be equal to false?

maxbarnas (changes requested):

packages/ckeditor5-html-support/src/schemadefinitions.js
jacekbogdanski: No, I will correct it.

packages/ckeditor5-html-support/tests/datafilter.js
jacekbogdanski: Some leftover from testing, sorry

PR #9927, opened 2021-06-21
Move leading comment nodes into the document fragment

### LINKED ISSUE #9861: First comment node is missing after calling `DOMParser.parseFromString()`

This issue has been found during working on HTML comments.

The DOM string parsed by [`HtmlDataProcessor#_toDom()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_dataprocessor_htmldataprocessor-HtmlDataProcessor.html#function-_toDom) does not contain the first comment node, if it is the first node in the DOM string.

This issue is caused because the return value from:
```js
new DOMParser().parseFromString( '<!--COMMENT 1--><p>PARAGRAPH 1</p><!--COMMENT 2-->', 'text/html' ).body.childNodes
```
is a two-elements list: `NodeList [ p, <!--COMMENT 2--> ]` without the first comment node `<!--COMMENT 1-->`. The first comment is somehow hoisted and inserted before the newly created HTML document:

This issue could be fixed by wrapping the DOM string with a `<body>` or even better with a `<div>`, just like it is done for [`BasicHtmlWriter#getHtml()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_dataprocessor_basichtmlwriter-BasicHtmlWriter.html#function-getHtml)

psmyrek:

Other (engine): Fixed parsing leading HTML comments by `HtmlDataProcessor.toView()`. Closes #9861.

### DIFF
--- a/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
+++ b/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
@@ -7,7 +7,7 @@
  * @module engine/dataprocessor/htmldataprocessor
  */
 
-/* globals document, DOMParser */
+/* globals document, DOMParser, Node */
 
 import BasicHtmlWriter from './basichtmlwriter';
 import DomConverter from '../view/domconverter';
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
```diff
@@ -119,10 +119,36 @@ export default class HtmlDataProcessor {
...
+
+		while ( !documentChildNode.isSameNode( document.documentElement ) ) {
+			const node = documentChildNode;
+
+			documentChildNode = documentChildNode.nextSibling;
+
+			/* istanbul ignore else */
+			if ( node.nodeType == Node.COMMENT_NODE ) {
+				fragment.appendChild( node );
+			}
```
psmyrek, line 145: Probably only comment nodes can be added directly to the Document object, before the `<html>` node. I have not been able to create any HTML string that would insert a node other than a comment at the beginning of the Document object, so the `/* istanbul ignore else */` is here to keep the 100% CC.

Anyway, this `if` condition here is just to be sure.

ma2ciek: Maybe it could be moved to the comment.

psmyrek: Comment has been added.

I also replaced the `Document` with `HTMLDocument` in comments, as the latter name [is used in the specification](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString#return_value).

ma2ciek (approved):
LGTM. @niegowski, maybe you could take a look as well.

niegowski (approved):
👍  LGTM

niegowski:
But the merge commit message should indicate that it fixes parsing the leading HTML comments.

psmyrek:
Commit message is updated. Is it better now? 🙂

niegowski:
For the commit message, I think there are too many internal details. I'd go just with:

Fixed parsing leading HTML comments with `HtmlDataProcessor.toView()`.

PR #9931, opened 2021-06-21
Expand DomConverter.blockElements with missing elements

### LINKED ISSUE #3920: Why is DomConverter.blockElements an array?
A set would make more sense.

### LINKED ISSUE #7863: Possible bug in handling newline element in blockqoute

Found here: #7859

Use `HtmlDataProcessor.toView()`  with (notice the `\n` after `</pre>`:

```
"<blockquote>\n<p>Example 1:</p>\n<pre><code>code 1</code></pre>\n<p>Example 2:</p>\n<pre><code>code 2</code></pre>\n</blockquote>\n"
```

This is converted to the view as (notice the `</pre> </blockquote>`)

```
"<blockquote><p>Example 1:</p><pre><code>code 1</code></pre><p>Example 2:</p><pre><code>code 2</code></pre> </blockquote>"
```

It is represented in the view as `Text` element with `" "` (space):

### ✔️ Expected result

The `" "` should not be inserted here.

## 📃 Other details

I'd suspect recent changes in handling paragraphable elements. But that's a wild guess.

---

### LINKED ISSUE #9801: Complete DomConverter.blockElements array

DomConverter.blockElements doesn't include every possible block element. We should extend it to cover additional blocks like `pre`, `section`, `article` etc, so basically, every HTML tag considered a block/flow element. The change is required for full HTML support to correctly handle trailing spaces.

https://github.com/ckeditor/ckeditor5/blob/79454f872b26db6fafa56aa63dd5f4e5a4ec6477/packages/ckeditor5-engine/src/view/domconverter.js#L90

Also, we could cover https://github.com/ckeditor/ckeditor5/issues/3920

---

jacekbogdanski:

Fix (engine): `DomConverter.blockElements` is missing some of the HTML block element names. Closes #9801. Closes #7863.

### DIFF
--- a/packages/ckeditor5-code-block/tests/codeblockediting.js
+++ b/packages/ckeditor5-code-block/tests/codeblockediting.js
@@ -1135,10 +1135,8 @@ describe( 'CodeBlockEditing', () => {
 			editor.setData( `<pre><code>foo</code></pre>
 				<pre><code>bar</code></pre>` );
 
-			// Note: The empty <paragraph> in between should not be here. It's a conversion/auto–paragraphing bug.
 			expect( getModelData( model ) ).to.equal(
 				'<codeBlock language="plaintext">[]foo</codeBlock>' +
-				'<paragraph> </paragraph>' +
 				'<codeBlock language="plaintext">bar</codeBlock>' );
 		} );
 
--- a/packages/ckeditor5-engine/src/view/domconverter.js
+++ b/packages/ckeditor5-engine/src/view/domconverter.js
@@ -1304,7 +1304,7 @@ export default class DomConverter {
 // Used to check if given native `Element` or `Text` node has parent with tag name from `types` array.
 //
 // @param {Node} node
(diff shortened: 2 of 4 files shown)


maxbarnas:
All good, restarted the build job. Will wait for it to finish before merging.

Reinmar:
> Other (engine): Changed `DomConverter.blockElements` type to `Set.<String>`. Closes #3920.

This sounds like a breaking change. This is a part of a public API and we recommended adding elements to that array.

Secondly, the ticket that I opened in 2016 was opened when this thing was added in the first place and it made more sense then. Changing this now, without a good reason has a dubious ROI in my opinion. The cost here is in the fact that it's a breaking change.

jacekbogdanski:
@Reinmar thanks, I've reverted the type change. I've also changed the merge message into a fix, as it seems to be fixing some cases in `code-block` and `markdown-gfm` plugins.

Issue #9339, opened 2021-03-23
Document image style migration scenarios and default styles


oleq:
## 📝 Provide a description of requested docs changes

*   Let's keep in as a separate guide in `i/2052-inline-images`
*   Later we may decide whether to keep it in [https://ckeditor.com/docs/ckeditor5/latest/features/image.html](https://ckeditor.com/docs/ckeditor5/latest/features/image.html) (or close to it because Bartek started rewriting and breaking it down).

magda-chrzescian:
Some guidelines agreed at the meeting:

- Way to think about it – someone has a certain setup (preset + configuration) and upgrades to the new version where we changed the default settings. There must be a way to:
    - Ensure that the content previously created will be maintained.
    - Which, in the simplest scenario, boils down to configuring the new version of the editor in the same way it was configured previously.
    - However, an additional thing to consider is – what if someone wants the same setup they had, but also use the new inline images.
- Scenarios to cover:
    - Article (Classic/Balloon/Inline) → new version (new preset)
        - Must be able to set the styles to side and full and remove inline images (assuming it's added to the build in the new version)
        - Bonus: Remember that side-image must apply `max-width:50%`
        - Bonus: How to remove `InlineImage` from a build? May require re-building.
    - Article → new version + inline images (new preset)
        - The same block styles (side and full) + something reasonable for inline images
    - Document → new version
        - Document has left, full, right currently
        - It must be able to set the styles to left, full, right and remove inline images
    - Document → new version + inline images
        - Document has left, full, right currently
        - Two options here:
            - Will the new default configuration work?
            - Or does it have to be adjusted?
    - Custom styles → new version + ?
        - Are there any tricky scenarios that were possible to achieve in the previous version of the feature but will not be easy to do now?

magda-chrzescian:
It may be worth mentioning that we are slightly modifying the content styles by removing `clear: both` rule from the `image` class (#9183).

magda-chrzescian:
A few words about renaming the public functions.

> We actually don't need the getSelectedImageWidget function anymore. Whenever we are looking for an image on which to execute a command, we are also looking for the image ancestor of the selected caption. So I replaced the getSelectedImageWidget with getSelectedOrAncestorImageWidget.
> 
> It was a public function, so I'm not sure if it's ok to change its name and behavior. But because of the changes in managing the toolbar state, our integrator probably also should start using the getSelectedOrAncestorImageWidget function.
> 
[...]

PR #9952, opened 2021-06-25
Docs: migration to 29 language review.

godai78:

Docs: Migration to v29.0.0 user guide.

Language review as requested in https://github.com/ckeditor/ckeditor5/issues/9339#issuecomment-868324817

### DIFF
--- a/docs/builds/guides/migration/migration-to-29.md
+++ b/docs/builds/guides/migration/migration-to-29.md
@@ -106,7 +106,7 @@ Since the appearance of the image in the document depends on the image type (blo
 	* In the {@link module:image/imagestyle/utils image style utils} module:
 		* the `defaultIcons` were renamed to {@link module:image/imagestyle/utils~DEFAULT_ICONS},
 		* the `defaultStyles` were renamed to {@link module:image/imagestyle/utils~DEFAULT_OPTIONS},
-		* the `normalizeImageStyles` function was removed from the public API.
+		* the `normalizeImageStyles()` function was removed from the public API.
 	* The `ImageStyleCommand#defaultStyle` and `ImageStyleCommand#styles` were removed from the public API.
 
 ## Image toolbar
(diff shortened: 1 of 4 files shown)


godai78:
Still missing this:

> Read more about the \[logic controlling the image type while inserting/pasting/dropping\](TODO: link).

docs/builds/guides/migration/migration-to-29.md
```diff
@@ -39,77 +40,78 @@ Since the appearance of the image in the document depends on the image type (blo
...
+	* In the {@link module:image/imagestyle/utils image style utils} module:
+		* the `defaultIcons` were renamed to {@link module:image/imagestyle/utils~DEFAULT_ICONS},
+		* the `defaultStyles` were renamed to {@link module:image/imagestyle/utils~DEFAULT_OPTIONS},
+		* the `normalizeImageStyles()` function was removed from the public API.
+	* The `ImageStyleCommand#defaultStyle` and `ImageStyleCommand#styles` were removed from the public API.
 
 ## Image toolbar
 
-Due to the changes mentioned above, the {@link module:image/image~ImageConfig#toolbar image toolbar} became crucial in terms of providing the user with proper interaction with images in terms of managing the image type and caption. Thus, it is recommended to use one of the following configurations as the minimum set-up for the image toolbar:
+Due to the changes mentioned above, the {@link module:image/image~ImageConfig#toolbar image toolbar} became crucial in terms of providing tools for proper interaction with images in terms of managing the image type and caption. Thus, it is recommended to use one of the following configurations as the minimum set-up for the image toolbar:
```
magda-chrzescian, line 114: > is unclear to me. Do I need to configure this in v29.0.0+ or the feature will not work? It's more an announcement than migration to me. I think we need to think this over.

Well, that's complicated 🤦 . It's going to work - it is still possible to insert an inline image in the middle of the paragraph and a block image in the empty paragraph. It is even possible to change the image type by dragging, but only if it has no caption. 

The caption won't work without the toolbar, but this one is handled in the caption section. Maybe the rest of the information should be placed in the image styles section. But IMO it's still about the migration. The other day it was completely fine to have images in the editor without the `ImageStyle` and `ImageToolbar` plugins, but since the inline images, it just gets handicapped, although it works in the basics. As an integrator, where should I get information about that? I'm pretty sure I wouldn't read the whole images guide, especially if I'm just upgrading the version.

godai78: We try incorporating this explanation into the guide. Even if as something short, like "The image feature will still work if no changes are made to the contextual toolbar configuration. Some of the functions, however, will not work as expected (like captions) and some will only available with limited interaction (e.g. by dragging the image around, while there is a ready toolbar item solution).  
  
How 'bout it?

godai78: On second thought, we may also add a warning info-box in the guides but keep this here as well.

PR #9957, opened 2021-06-25
Ck/9895 image plugin metadata

### LINKED ISSUE #9895: Updating the image plugin metadata.

We should update the image plugin metadata. It requires changes in the `UIComponents` and possibly some other sections.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

magda-chrzescian:

Internal (Image): The metadata of the image-related plugins should be up to date. Closes #9895.

---

### Additional information

Because Online Builder generates the image toolbar based on the plugin metadata `uiComponents`, the `ImageStyle` plugin metadata provides the `inline`, `block`, and `side` buttons, which are correct for most of the editor builds.

However, this is not correct for the document build and we should provide the possibility to customize the default toolbars for the builds in the future.


packages/ckeditor5-image/ckeditor5-metadata.json
```diff
@@ -7,7 +7,70 @@
 			"docs": "features/image.html#inserting-images-via-pasting-url-into-editor",
```
psmyrek, line 7: The docs have been rewritten and the links have changed. I know that there are some new redirects added in the Umberto, but IMHO it would be nice to use valid links in the metadata files.

This concerns also other plugins in the `ckeditor5-image` package.

packages/ckeditor5-image/ckeditor5-metadata.json
```diff
@@ -7,7 +7,70 @@
...
+					"ImageBlock",
+					"ImageInline"
+				]
+			]
+		},
+		{
+			"name": "Block image",
+			"className": "ImageBlock",
+			"description": "Enables support for inserting images between the blocks of text, like paragraphs or block quotes. This functionality is also provided by the main Image plugin. Note: The image feature is introduced in a granular form, implemented by separate plugins responsible for such aspects as image caption, toolbar, resize, upload or styles.",
+			"path": "src/imageblock.js",
```
psmyrek, line 20: Missing "docs" property, so the `ImageBlock` plugin will not have a "Documentation" link on the bottom of the tile in Online Builder:

packages/ckeditor5-image/ckeditor5-metadata.json
```diff
@@ -7,7 +7,70 @@
...
+						"width"
+					]
+				}
+			]
+		},
+		{
+			"name": "Inline image",
+			"className": "ImageInline",
+			"description": "Enables support for inserting images inside the blocks of text, like paragraphs, header or block quotes. This functionality is also provided by the main Image plugin. Note: The image feature is introduced in a granular form, implemented by separate plugins responsible for such aspects as image toolbar, resize, upload or styles.",
+			"path": "src/imageinline.js",
```
psmyrek, line 52: Missing "docs" property, so the `ImageInline` plugin will not have a "Documentation" link in Online Builder.

packages/ckeditor5-image/ckeditor5-metadata.json
```diff
@@ -7,7 +7,70 @@
...
+					"elements": "img",
+					"attributes": [
+						"alt",
+						"sizes",
+						"src",
+						"srcset",
+						"width"
+					]
+				}
+			]
```
psmyrek, line 46: Since the new `ImageBlock` and `ImageInline` plugins are the ones that create HTML elements (`<figure>`, ``) now, then I wonder if the "old" `Image` plugin should still have the `htmlOutput` and `uiComponents` arrays here, as it does not create those things anymore.

oleq: I tried to find a similar case where there's a glue plugin and individual features underneath. I think it mainly depends on how this data will be consumed.

ma2ciek: > I tried to find a similar case where there's a glue plugin and individual features underneath. I think it mainly depends on how this data will be consumed.

It can be consumed either by some tools like Online Builder, either by developers.

For me, we could try to generate the `Image`'s `htmlOutput` and `uiComponents` dynamically (in the OB and HTML output guide) as we know that it consists of `ImageBlock` and `ImageInline`. So I'd leave these fields empty as we can generate them dynamically (of course this behavior would have to be documented).

oleq:
I addressed all the issues. Now I guess this needs a final review and a merge.

psmyrek (approved):
LGTM 🎉

packages/ckeditor5-image/ckeditor5-metadata.json
```diff
@@ -7,15 +7,18 @@
 			"docs": "features/images/image-upload/images-inserting.html#inserting-images-via-pasting-url-into-editor",
 			"path": "src/autoimage.js",
 			"requires": [
-				"Image"
+				[
+					"ImageBlock",
+					"ImageInline"
+				]
```
ma2ciek, line 13: @psmyrek, AFAIS it's an undocumented feature now. Could you add a comment there?

psmyrek: Added.
I also updated the Features’ HTML output page in the docs

PR #9968, opened 2021-06-28
[GHS] Existing features shouldn't break attribute filtration

jacekbogdanski:

Fix (html-support): Existing features no longer breaks GHS attributes filtering.

### DIFF
--- a/packages/ckeditor5-html-support/src/datafilter.js
+++ b/packages/ckeditor5-html-support/src/datafilter.js
@@ -26,7 +26,7 @@ import {
 	viewToModelBlockAttributeConverter,
 	modelToViewBlockAttributeConverter
 } from './converters';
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-html-support/src/datafilter.js
```diff
@@ -521,6 +523,28 @@ function consumeAttributeMatches( viewElement, { consumable }, matcher ) {
 	return consumedMatches;
 }
 
+// Removes attributes from the given match that where already consumed by other converters.
```
maxbarnas, line 526:
```suggestion
// Removes attributes from the given match that were already consumed by other converters.
```

maxbarnas (changes requested):

PR #9972, opened 2021-06-29
I/9959: UI/UX improvements to find and replace feature

### LINKED ISSUE #9959: UI/UX improvements of the find and replace feature.
Since I won't be able to review the functionality on Monday, I'll check it now, so just ignore my comments in case it points at some already known issue.

1. The visual appearance of the split button. It has probably unnecessary class `ck-button_with-text`.
How it looks like:
image

How it should look like:
image

2. The text in buttons is misplaced vertically (too high). It probably would be better to set the fixed height for the button and align the text vertically with the `line height` = button height or with the flex `align-items` property.
image

3. Checkbox labels should be prepared for longer texts. It should have the `max-size`, `overflow` and `text-overflow` properties, or optionally should wrap the text if the max-width is exceeded.
image

4. If we wanted to stay with the arrows and skip the `find` button in the MVP, the second arrow should point to the right, as it's a symbol for something like "run"/"play", an arrow pointing down is a bit unclear about the purpose. The first arrow should be rotated accordingly.

5. The tooltip should not be capitalized. It can be set on the buttons only using the `text-transform: uppercase` CSS rule.

6. The arrows are a bit too big (scale set 0.9 looks fine IMO, but @oleq may have some concerns about the sharpness of the icons(?)). They are also misplaced vertically, but it's hard to define the exact values they should be moved, as I'm not sure if we are going to change their rotation (as mentioned in 4). Again, please consult it with @oleq if needed.

7. Let's keep the input and checkbox border colors consistent (the lighter gray).
image

8. Clicking the label should result in checking/unchecking. Moreover, the cursor should be set to the pointer when hovering over the checkbox/label.

9. Let's make the margin top of this pretty selector `.ck .ck-find-and-replace-form .ck-find-form__wrapper.ck-responsive-form .ck-find-checkboxes .ck-find-checkboxes__box:first-child` equal to `ck-spacing-large` instead of the `ck-spacing-unit`. (BTW, to set the size to `ck-spacing-unit`, the `ck-spacing-standard` variable should be used).

mateuszzagorski:

Fix (find-and-replace): As per issue #9959, I've applied various changes to the code to fix UI/UX problems with the find-and-replace feature.  
As of this moment, checkboxes styling (point 7) and the position of the prev/next arrows (point 6) are still missing. Closes #9959.

### DIFF
--- a/packages/ckeditor5-find-and-replace/src/findandreplaceui.js
+++ b/packages/ckeditor5-find-and-replace/src/findandreplaceui.js
@@ -172,7 +172,6 @@ export default class FindAndReplaceUI extends Plugin {
 		// Configure dropdown's button properties:
 		buttonView.set( {
 			icon,
-			withText: true,
 			tooltip: t( 'Find and replace' )
 		} );
 
--- a/packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
+++ b/packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
@@ -105,16 +105,12 @@
 				border-radius: 5px 0 0 5px;
 
 				& svg {
-					transform: rotate(90deg);
+					transform: rotate(180deg);
 				}
 			}
 
 			&.ck-button-next {
 				border-radius: 0 5px 5px 0;
-
-				& svg {
-					transform: rotate(270deg);
-				}
 			}
 		}
 	}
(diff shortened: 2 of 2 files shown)


LukaszGudel (approved):

mlewand (approved):
LGTM

PR #9977, opened 2021-06-29
Fixed find and replace commands integration with readonly editor

### LINKED ISSUE #9975: It should be possible to find text in readonly editor (but not to replace it)
Based on https://github.com/ckeditor/ckeditor5/issues/9964#issuecomment-869651757

> 1.  Switch to source mode
> 
> **Expected:** find & replace toolbar dropdown is inactive.
> 
> **Actual:** find & replace toolbar dropdown is active, you can open it and type some text, but nothing happens anyway.

Some more conversation from outside GH (we changed expected a bit):

> m.lewandowski:spiral\_calendar\_pad:  8 minutes ago  
> We were discussing that find and replace should be integrated with readonly mode. On sync we were talking that button should be disabled - but find feature still makes a perfect sense in readonly mode. It's only the replace feature that has no sense for readonly editor. Anything agains this? @filip Tokarski @Łukasz Gudel
> 
> Filip Tokarski  6 minutes ago  
> You're right :+1:
> 
> Łukasz Gudel  5 minutes ago  
> Find only in read-only mode seems to be useful, but if it’s find only then user can also use native browser find tool, as those can find words in our editor as well
> 
> m.lewandowski:spiral\_calendar\_pad:  4 minutes ago  
> @Łukasz Gudel you're right but they will be searching in entire website content, so if there are multiple editors this would be searching across all the editors.
> 
> m.lewandowski:spiral\_calendar\_pad:  3 minutes ago  
> Plus native browser doesn't have some features that we have - match whole words and case sensitive search.

mlewand:

Internal (find-and-replace): Fixed find and replace commands integration with readonly editor. Closes #9975.

### DIFF
--- a/packages/ckeditor5-find-and-replace/src/findandreplace.js
+++ b/packages/ckeditor5-find-and-replace/src/findandreplace.js
@@ -104,10 +104,25 @@ export default class FindAndReplace extends Plugin {
 			// We need to wait for UI ready to have the toolbar dropdown available. Otherwise findAndReplace component
 			// is registered but not yet constructed.
 			this.listenTo( this.editor.ui, 'ready', () => {
+				const formView = ui.formView;
 				// If editor doesn't contain findAndReplace button then there's no ui#formView property.
-				if ( ui.formView ) {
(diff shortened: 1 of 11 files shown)


mlewand:
@ckeditor/qa-team PTAL

FilipTokarski:
It works fine now 👍

PR #9979, opened 2021-06-29
Docs: Added section about allowing all elements and attributes in GHS feature guide

### LINKED ISSUE #9976: [GHS] Add the configuration option for disabling content filtering to GHS docs

It would be nice to add a small section about disabling content filtering (like `allowedContent: true`) to the General HTML Support feature guide.

Mgsy:

Docs: Added section about allowing all elements and attributes in GHS feature guide. Closes #9976.

### DIFF
--- a/packages/ckeditor5-html-support/docs/features/general-html-support.md
+++ b/packages/ckeditor5-html-support/docs/features/general-html-support.md
@@ -214,6 +214,21 @@ The General HTML Support feature distinguishes several content types, each treat
 
 The enabled elements will not just be available "anywhere" in the content, as they still need to adhere to certain rules derived from the HTML schema and from common sense. Also, the behavior of specific types of elements in the editing area will be different. For instance, the object elements will only be selectable as a whole, and the inline elements will work the same as other formatting features supported by CKEditor 5 (e.g. bold, italic) do.
 
+### Enabling all elements and attributes
+
+In some cases, it might be desired to disable content filtering, so all elements and attributes will be allowed. It could be done with a special configuration:
+
+```js
+{
+	name: /.*/,
+	attributes: true,
+	classes: true,
+	styles: true
+}
+```
+
+The above configuration will work similarly to `allowedContent: true` option from CKEditor 4.
+
 ## Known issues
 
 It is possible to add support for arbitrary styles, classes and other attributes to existing CKEditor 5 features (such as paragraphs, headings, list items, etc.).


packages/ckeditor5-html-support/docs/features/general-html-support.md
```diff
@@ -214,6 +214,21 @@ The General HTML Support feature distinguishes several content types, each treat
 
 The enabled elements will not just be available "anywhere" in the content, as they still need to adhere to certain rules derived from the HTML schema and from common sense. Also, the behavior of specific types of elements in the editing area will be different. For instance, the object elements will only be selectable as a whole, and the inline elements will work the same as other formatting features supported by CKEditor 5 (e.g. bold, italic) do.
 
+### Enabling all elements and attributes
+
+In some cases, it might be desired to disable content filtering, so all elements and attributes will be allowed. It could be done with a special configuration:
```
jacekbogdanski, line 219:
```suggestion
### Enabling all HTML features

It might be desired to enable all HTML features in some cases, so all elements and attributes will be allowed by the editor. It could be done with a special configuration:
```

It's a bit backward what we have in CKE4 and ACF, so in the case of CKE5, we are allowing for any HTML feature, not disabling filtration which is disabled by default.

packages/ckeditor5-html-support/docs/features/general-html-support.md
```diff
@@ -214,6 +214,21 @@ The General HTML Support feature distinguishes several content types, each treat
...
+
+```js
+{
+	name: /.*/,
+	attributes: true,
+	classes: true,
+	styles: true
+}
```
jacekbogdanski, line 227:
```suggestion
htmlSupport: {
	allow: [
		{
			attributes: true,
			classes: true,
			styles: true
		}
	]
}
```
To keep it consistent with other samples.

packages/ckeditor5-html-support/docs/features/general-html-support.md
```diff
@@ -214,6 +214,21 @@ The General HTML Support feature distinguishes several content types, each treat
...
+```
+
+The above configuration will work similarly to `allowedContent: true` option from CKEditor 4.
```
jacekbogdanski, line 230: Do you think it makes sense to link to https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-allowedContent when we are mentioning the CKE4 feature? Not sure if we want to do it from CKE5 docs but it would improve navigation for someone migrating content from CKE4.

jacekbogdanski (changes requested):

packages/ckeditor5-html-support/docs/features/general-html-support.md
Mgsy: Yeah, I was thinking about it and wasn't sure if it makes sense to do that. I'll add a link :+1:

jacekbogdanski (approved):
🎉

PR #9993, opened 2021-06-30
Move the context for the 'Show more items' message to 'ckeditor5-core'.

### LINKED ISSUE #9991: Move the `Show more items` message context to the `ckeditor5-core` so CF and ckeditor5-ui can use it
As in the title :point\_up:

ma2ciek:

Internal: Moved the context for the `Show more items` message to `ckeditor5-core`. Closes #9991.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-core/lang/contexts.json
+++ b/packages/ckeditor5-core/lang/contexts.json
@@ -1,6 +1,7 @@
 {
-	"Save": "Label for the Save button.",
 	"Cancel": "Label for the Cancel button.",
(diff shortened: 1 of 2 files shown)


pomek (approved):
LGTM.

pomek:
Is that all? I mean, don't we have to copy existing translations between files?

ma2ciek:
> Is that all? I mean, don't we have to copy existing translations between files?

The download from Tx overwrites everything.

After uploading these translations we should check on Tx if the translations for the `Show more items` can be restored in the new package. AFAIR it worked this way.

PR #10047, opened 2021-07-02
Find and replace: Input indication when no matches found

### LINKED ISSUE #10014: The find and replace view does not indicate that no matches were found

[http://fake.ckeditor.com:8081/build/docs/ckeditor5/latest/features/find-and-replace.html](http://fake.ckeditor.com:8081/build/docs/ckeditor5/latest/features/find-and-replace.html)

if you search for anything that has no matches there is no UI response of any kind. It would be nice to see `0 of 0` in the input view as for found matches. 

## 📃 Other details

*   Browser: …
*   OS: …
*   CKEditor version: …

---

mateuszzagorski:

Fix (find-and-replace): The find and replace view should now correctly indicate when no results were found. Closes #10014.


packages/ckeditor5-find-and-replace/src/findandreplaceui.js
```diff
@@ -140,8 +140,8 @@ export default class FindAndReplaceUI extends Plugin {
 		} );
 
 		this.bind( 'highlightOffset' ).to( state, 'highlightedResult', highlightedResult => {
-			if ( !highlightedResult ) {
-				return null;
+			if ( !highlightedResult && state.results.length === 0 ) {
+				return 0;
```
mlewand, line 144: These proposed 2 lines, although it gives expecting results for the end user, are wrong from the logic perspective.

`highlightOffset` is a 0-based index of currently highlighted item. If you assign 0 to it when there are no results or there is no highlighted result you're acting as if there was at least one search result available and the first search result was highlighted. That is not true.

The easiest fix would be to fix it by binding form's `isCounterHidden` observable to state's `searchText` value. If state has non-empty value in `searchText` it means that searching was executed and we can show the counter.

However, the more proper solution is more complex: as I started to look through the code I understood that [I initially messed the `isSearching` observable in the UI plugin](https://github.com/ckeditor/ckeditor5/blob/85f658bf3471c609fbd8644d6a25a425804fb573/packages/ckeditor5-find-and-replace/src/findandreplaceui.js#L103-L116). What I implemented there should be called `isDirty` property - as it tells whether the state differs from what user entered into the UI. `isSearching` should be much simpler. It should be pretty much what I wrote above.

The last idea is the proper fix for it and we should take the occasion to do it as it will make the plugin more intuitive.

mlewand (changes requested):
The tests are missing.

packages/ckeditor5-find-and-replace/src/findandreplaceui.js
mateuszzagorski: As per your suggestion, the `isDirty` property has been introduced now. Also, a lot of changes to the code to improve code readability, usability etc.

mlewand (approved):

mlewand:
Review misclick I wanted to accept PR in a different tab :slightly\_smiling\_face:

packages/ckeditor5-find-and-replace/src/findandreplacestate.js
```diff
@@ -0,0 +1,127 @@
...
+		/**
+		 * A collection of results.
+		 *
+		 * @private
+		 * @member {module:utils/collection~Collection}
+		 */
+		this.set( 'results', new Collection() );
+
+		/**
+		 * Currently highlighted search result in {@link #matchCount matched results}.
```
mlewand, line 30: That's in the #results property rather than #matchCount.

packages/ckeditor5-find-and-replace/src/findandreplaceui.js
mlewand: That's right we introduced the `isDirty` property but the change raised by me in this comment was not reverted. It should be reverted because otherwise offset can have misleading value.

packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js
```diff
@@ -25,7 +25,7 @@ import '@ckeditor/ckeditor5-ui/theme/components/responsive-form/responsiveform.c
 import '../../theme/findandreplaceform.css';
 // eslint-disable-next-line ckeditor5-rules/ckeditor-imports
 import findArrowIcon from '@ckeditor/ckeditor5-ui/theme/icons/dropdown-arrow.svg';
-import CheckboxView from '../ui/checkboxview';
```
mlewand, line 28: Good catch that it was using the same directory 👍

packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js
```diff
@@ -47,190 +47,133 @@ export default class FindAndReplaceFormView extends View {
...
+		  *
+		  * @readonly
+		  * @observable
+		  * @member {String} #replaceText
+		  */
+		this.set( 'replaceText', '' );
 
 		/**
-		 * Whether the search results counter should be visible.
+		 * Indicates if searched text has been changed in form.
```
mlewand, line 71: This property should not be limited to sniff changes in searched text. It should also sniff changes in match case, whole words options.

packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js
```diff
@@ -47,190 +47,133 @@ export default class FindAndReplaceFormView extends View {
...
+		 * @readonly
+		 * @observable
+		 * @member {Number} #matchCount
 		 */
-		this.findView = this._createFindView();
+		this.set( 'matchCount', null );
 
 		/**
-		 * Stores gathered views related to replace functionality of the feature.
+		 * The offset of currently highlighted search result in {@link #matchCount matched results}.
```
mlewand, line 107: Or `null` if there's no highlighted result.

packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js
```diff
@@ -47,190 +47,133 @@ export default class FindAndReplaceFormView extends View {
...
-		 * @member {module:utils/keystrokehandler~KeystrokeHandler}
+		 * @observable
+	 	 * @param {Function} t Used to translate message to the uiLanguage.
 		 */
-		this.keystrokes = new KeystrokeHandler();
+		this._assignViews( t );
 
 		/**
-		 * A collection of views that can be focused in the form.
+	 	 * Initialization of focusables elements of the form.
```
mlewand, line 136: I believe I'm seeing a space introduced inbetween tabs which means a mixed tabs and spaces.

packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js
```diff
@@ -47,190 +47,133 @@ export default class FindAndReplaceFormView extends View {
...
-		 * @member {module:utils/focustracker~FocusTracker}
+		 * @observable
+		 * @member {Boolean} #isCounterHidden
 		 */
-		this.focusTracker = new FocusTracker();
+		this.set( 'isCounterHidden', true );
 
 		/**
-		 * An instance of the {@link module:utils/keystrokehandler~KeystrokeHandler}.
+		 * Assigning form elements views.
```
mlewand, line 126: This is a function call, this is not an observable property assignment.

You want to annotate only properties assignments that way. Functions have their own JSDoc before its definition.

The same applies to `_initFocusables` call.

packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js
```diff
@@ -47,190 +47,133 @@ export default class FindAndReplaceFormView extends View {
...
 
 		this.bind( 'isCounterHidden' ).to( this, 'matchCount', this, 'highlightOffset', ( matchCount, highlightOffset ) => {
-			return matchCount === null || matchCount === 0 ||
-				highlightOffset === null || highlightOffset === 0;
+			return matchCount === null || highlightOffset === null;
 		} );
 
+		this.bind( 'isSearching' ).to( this );
+
+		if ( state ) {
```
mlewand, line 167: State is mandatory, there can be no situation when state is not passed.

So this if is not needed.

mlewand: Also this means that you won't have to do two `isSearching` bindings and later unbinding. Instead you'll bind `isSearching` just once.

packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js
```diff
@@ -337,6 +285,162 @@ export default class FindAndReplaceFormView extends View {
 		}, { priority: 'high' } );
 	}
 
+	/**
+	 * Initialization of focusables elements of the form.
+	 *
+	 * @private
+	 */
+	_initFocusables() {
```
mlewand, line 293: Good that you made the function a private one.

packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js
```diff
@@ -337,6 +285,162 @@ export default class FindAndReplaceFormView extends View {
 		}, { priority: 'high' } );
 	}
 
+	/**
+	 * Initialization of focusables elements of the form.
+	 *
+	 * @private
```
mlewand, line 291: The order of functions in class should be:

*   public
*   protected
*   private

Both \_assignViews and \_initFocusables occurrs before focus (which is a public method).

mlewand (changes requested):
Docs are throwing errors as reported by our CI:

```
JSDoc started...
Incorrect param type: module:find-and-replace/findandreplaceediting~FindAndReplaceState
	at /home/travis/build/ckeditor/ckeditor5/packages/ckeditor5-find-and-replace/src/findandreplaceui.js:109
Incorrect link: module:find-and-replace/findandreplacestate~FindAndReplaceState#matchCount
	at /home/travis/build/ckeditor/ckeditor5/packages/ckeditor5-find-and-replace/src/findandreplacestate.js:29
Validation Summary:
* Validator found 2 errors.
FATAL: Since the process is being executed in strict or checking mode, aborting due to errors.
An error was thrown by JSDoc:
Error: Error while executing node /home/travis/build/ckeditor/ckeditor5/node_modules/jsdoc/jsdoc.js -c /tmp/tmp-10812-NjXUdA4dj6Di: Incorrect param type: module:find-and-replace/findandreplaceediting~FindAndReplaceState
	at /home/travis/build/ckeditor/ckeditor5/packages/ckeditor5-find-and-replace/src/findandreplaceui.js:109
Incorrect link: module:find-and-replace/findandreplacestate~FindAndReplaceState#matchCount
	at /home/travis/build/ckeditor/ckeditor5/packages/ckeditor5-find-and-replace/src/findandreplacestate.js:29
Validation Summary:
* Validator found 2 errors.
FATAL: Since the process is being executed in strict or checking mode, aborting due to errors.
    at Object.shExec (/home/travis/build/ckeditor/ckeditor5/node_modules/@ckeditor/ckeditor5-dev-utils/lib/tools.js:42:10)
    at Object.build (/home/travis/build/ckeditor/ckeditor5/node_modules/@ckeditor/ckeditor5-dev-docs/lib/docs/index.js:94:9)
error Command failed with exit code 1.
```

mlewand:
I also fixed couple of issues while reviewing in ff5e575df5

mlewand (changes requested):
I can see that review notes have been addressed but while doing this this PR is no longer resolving #10014.

I believe that when you reverted code for https://github.com/ckeditor/ckeditor5/pull/10047#discussion_r668491773 you didn't adjust the `isCounterHidden` observable property implementation.

This also shows that there are no proper test coverage as the tests are green despite issue not being fixed.

mlewand:
Fixed [my findings](https://github.com/ckeditor/ckeditor5/pull/10047#pullrequestreview-705041089) in b054609.

mlewand (approved):
After b054609a72 it's fine.

PR #10076, opened 2021-07-05
Fixed autoformat not remembering previous language choice for codeblock

### LINKED ISSUE #10005: The code block does not remember the language

1.  Insert C++ code block
2.  Insert another code block via the Autorformatting feature shortcode.
3.  A new code block is created

### ✔️ Expected result

_The new code block should_ be a C++ block.

### ❌ Actual result

It is a plain text code bock instead as pictured:

### ❓ Possible solution

## 📃 Other details

*   Browser: Safari 14
*   OS: MacOS Catalina
*   First affected CKEditor version: pre v29.x release
*   Installed CKEditor plugins: default for the Code blocks feature demo.  

    also on:

*   Browser: Chrome 91.0.4472.114  
*   OS: MacOS Catalina  
*   First affected CKEditor version: pre v29.x release  
*   Installed CKEditor plugins: default in the image feature guide (namely image styles guide)

---

przemyslaw-zan:
Fix (autoformat): Autoformat will apply the previous language choice for the code block feature. Closes #10005.

### DIFF
--- a/packages/ckeditor5-autoformat/src/autoformat.js
+++ b/packages/ckeditor5-autoformat/src/autoformat.js
@@ -178,7 +178,11 @@ export default class Autoformat extends Plugin {
 	 */
 	_addCodeBlockAutoformats() {
 		if ( this.editor.commands.get( 'codeBlock' ) ) {
-			blockAutoformatEditing( this.editor, this, /^```$/, 'codeBlock' );
+			blockAutoformatEditing( this.editor, this, /^```$/, () => {
+				this.editor.execute( 'codeBlock', {
+					usePreviousLanguageChoice: true
+				} );
+			} );
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pomek (changes requested):
Unfortunately, the proposed change isn't correct. A few tests related to the integration of the code block feature with lists do not pass.

```js
_addCodeBlockAutoformats() {
	const editor = this.editor;
	const selection = editor.model.document.selection;

	if ( editor.commands.get( 'codeBlock' ) ) {
		blockAutoformatEditing( editor, this, /^```$/, () => {
			// Do not execute the `codeBlock` command if a selection is inside a list item.
			if ( selection.getFirstPosition().parent.is( 'element', 'listItem' ) ) {
				return false;
			}

			this.editor.execute( 'codeBlock', {
				usePreviousLanguageChoice: true
			} );
		} );
	}
}
```

I checked this, and it solves the original problem and does not break existing features. 

We need a test for this change. Put the selection into an empty paragraph, then execute the `codeBlock` command with some language. Then, execute the command again (it should remove the code block element). Then, type \`\`\` and check whether inserted code block element uses the previous language. The model check will be enough.

psmyrek (approved):
LGTM 🎉

Issue #10048, opened 2021-07-02
Add editor icons to guides


godai78:
Add editor toolbar icons to user guides.

Guides that could use toolbar icons:

*   [x] basic text styles (plenty)
*   [x] block indentation
*   [x] block quote
*   [x] code blocks
*   [x] export to PDF
*   [x] export to Word
*   [x] find and replace
*   [x] font (plenty)
*   [x] highlight
*   [x] HR
*   [x] HTML embed
*   [x] link
*   [x] lists (both guides)
*   [x] math / chem formulas
*   [x] media embed
*   [x] page break
*   [x] pagination
*   [x] remove formatting
*   [x] restricted editing
*   [x] revision history
*   [x] select all
*   [x] special characters
*   [x] spelling and grammar checking
*   [x] tables (goddamn plenty of those little pesky bastards)
*   [x] text alignment

Collaboration features:

*   [x] comments
*   [x] track changes

Post-release new guides:

*   [x] images (plenty of those, really plenty)
*   [x] source editing

godai78:
That's it for now, more after code release.

PR #10077, opened 2021-07-05
Add editor icons to guides #04

godai78:
Docs: Adding toolbar items icons to user guides.

Part of: [#10048](https://github.com/ckeditor/ckeditor5/issues/10048) 

This PR covers the following guides:

*   link
*   media embed
*   page break
*   math / chem formulas
*   lists (all 3 of these)

---

### **Additional information**

One reviewer should be enough from the technical point of view, so the first one is the winner and we merge, but the math/chem icons need more love before merging.

### DIFF
--- a/docs/_snippets/features/mathtype.html
+++ b/docs/_snippets/features/mathtype.html
@@ -1,3 +1,10 @@
+<style>
+	.editor-icon__image[alt="MathType"],
+	.editor-icon__image[alt="ChemType"] {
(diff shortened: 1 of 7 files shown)


FilipTokarski:
I understand we're leaving those green and red MathType icons for now?

godai78:
There is a proposal for these to be grayscale (for now), so I guess we'll merge it after the changes and work on it later.

pomek:
image

PR #10080, opened 2021-07-05
Find and replace: Source mode integration added, manual test updated

### LINKED ISSUE #10001: Integrate find and replace plugin with source mode

Based on https://github.com/ckeditor/ckeditor5/issues/9964#issuecomment-869651757

> 1. Switch to source mode
> 
> **Expected:** find & replace toolbar dropdown is inactive.
> 
> **Actual:** find & replace toolbar dropdown is active, you can open it and type some text, but nothing happens anyway.

---

mateuszzagorski:

Fix (find-and-replace): The dropdown button should now be disabled when the source mode is active. Closes #10001.

### DIFF
--- a/packages/ckeditor5-find-and-replace/src/findandreplaceui.js
+++ b/packages/ckeditor5-find-and-replace/src/findandreplaceui.js
@@ -122,6 +122,9 @@ export default class FindAndReplaceUI extends Plugin {
 				cancelEvent();
 			} );
 
+			// Dropdown won't be enabled when find command is disabled.
+			dropdown.bind( 'isEnabled' ).to( editor.commands.get( 'find' ) );
+
 			return dropdown;
 		} );
 	}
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-find-and-replace/src/findandreplaceui.js
```diff
@@ -122,6 +122,9 @@ export default class FindAndReplaceUI extends Plugin {
 				cancelEvent();
 			} );
 
+			// Dropdown won't be enabled when find command is disabled.
```
mlewand, line 125: This comment describes obvious thing - it basically reads the code one line below which makes it redundant. If anything we should say **why** we are doing this binding. You could do one sentence explanation and add a reference number to issue #10001 - this will give the most insights.

mlewand (changes requested):
Unit tests are missing. Other than that looks ok.

mlewand (commented):
Looking good 👍

PR #10087, opened 2021-07-06
Remove `style` and `class` attributes from Matcher `attributes` match

### LINKED ISSUE #9813: Normalize further patterns passed to Matcher
As mentioned in https://github.com/cksource/ckeditor5-internal/issues/718#issuecomment-852362126, we came across a conflicting case while trying to limit the scope of attributes filtering.

We would like to normalize patterns of cases such as the one in [LinkImageEditing](https://github.com/ckeditor/ckeditor5/blob/8024a9169fb4aa1cbc89078c1ba18560b68ccf15/packages/ckeditor5-link/src/linkimageediting.js#L253-L258) by essentially making this:

```js
const pattern = {
	attributes: {
		rel: 'no-follow',
		class: 'foo-class'
	}
}
```

into 

```js
const pattern = {
	attributes: {
		rel: 'no-follow'
	},
	classes: [ 'foo-class' ]
}
```
	

Things to consider:
- Pattern `attributes: true` makes the normalization problematic. Should we match styles and classes in such case?
- Methods like `element.getAttribute()`, `element._setAttribute()` etc. do work with `style` and `class` attributes.

---

jacekbogdanski:

Fix (engine): Matcher shouldn't match style and class attributes with object "attributes" pattern. Closes #9813.

Fix (link): Link decorators should use `classes` and `styles` properties instead of directly matching `style` and `class` HTML attributes . Closes #9813.

### DIFF
--- a/docs/builds/guides/migration/migration-to-30.md
+++ b/docs/builds/guides/migration/migration-to-30.md
@@ -14,7 +14,7 @@ Listed below are the most important changes that require your attention when upg
 
 ### Matcher pattern API change
 
-Starting from v30.0.0, the {@link module:engine/view/matcher~Matcher} feature deprecated matching `style` and `class` HTML attributes using `attributes` key->value object pattern.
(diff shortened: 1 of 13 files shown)


maxbarnas:
LGTM! 🎉

PR #10088, opened 2021-07-06
Preventing CI from running twice per PR

przemyslaw-zan:
Internal: Limited CI to building only for the following branches: `#master`, `#stable`, and `#release`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ dist: xenial
 addons:
   chrome: stable
 language: node_js
+branches:
+  only:
+  - master
+  - stable
+  - release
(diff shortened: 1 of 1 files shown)


pomek:
We need to take care about tags:

> Note that safelisting also prevents tagged commits from being built. If you consistently tag your builds in the format v1.3 you can safelist them all with regular expressions, for example /^v\d+\.\d+(\.\d+)?(-\S*)?$/.

`^v\d+\.\d+\.\d+$` should be enough in our case.

pomek (approved):
LGTM.

PR #10093, opened 2021-07-07
Added ckeditor5-metadata.json to package.json

### LINKED ISSUE #10004: `ckeditor5-metadata.json` files are not published on NPM
### ❌ Actual result

The file is published only for CF packages, it is not published in `CKEditor 5` and `ckeditor5-internal` due to the fact that this file is not added to the `files` array in the `package.json`.

### ❓ Possible solution

Add `ckeditor5-metadata.json` file to the `files` array in `package.json` array.

## 📃 Other details

Currently, the Online builder can work on metadata files that are taken from GH and saved locally, but for the 3rd-party integrators, these files won't be available. Also it's planned to use the NPM-based versions in the OB as well. 

---

przemyslaw-zan:
Other: The `ckeditor5-metadata.json` file will be published on npm along with the package's code. Closes ckeditor/ckeditor5#10004.

### DIFF
--- a/package.json
+++ b/package.json
@@ -210,8 +210,5 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
-  },
-  "files": [
-    "ckeditor5-metadata.json"
-  ]
+  }
 }
--- a/package.json
+++ b/package.json
@@ -210,5 +210,8 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
-  }
+  },
+  "files": [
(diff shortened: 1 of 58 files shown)


pomek (changes requested):
The `ckeditor5` package does not contain the `ckeditor5-metadata.json` file. However, almost all `ckeditor5-*` packages do. 
The ticket is about adding the file to CKEditor 5 packages.

pomek (approved):
LGTM.

PR #10097, opened 2021-07-07
Find and replace: Label won't animate on view opening

### LINKED ISSUE #10008: The find and replace view animates "find in text" label on view opening

1.  http://fake.ckeditor.com:8081/build/docs/ckeditor5/latest/features/find-and-replace.html
2.  Click the "find and replace" toolbar button

### ✔️ Expected result

The input field label should not animate as in other features (for example table properties).

### ❌ Actual result

It's animating (slow motion recording):

### ❓ Possible solution

## 📃 Other details

*   Browser: …
*   OS: …

---

mateuszzagorski:

Fix (find-and-replace):  Label animation on view opening has been removed. Closes #10008.

Tests (find-and-replace): Added missing tests for the dropdown's button.

### DIFF
--- a/packages/ckeditor5-find-and-replace/src/findandreplaceui.js
+++ b/packages/ckeditor5-find-and-replace/src/findandreplaceui.js
@@ -184,9 +184,16 @@ export default class FindAndReplaceUI extends Plugin {
 		buttonView.actionView.delegate( 'execute' ).to( buttonView.arrowView );
 
 		// Each time a dropdown is opened, the search text field should get focused.
+		// Note: Use the low priority to make sure the following listener starts working after the
+		// default action of the drop-down is executed (i.e. the panel showed up). Otherwise, the
+		// invisible form/input cannot be focused/selected.
 		buttonView.on( 'open', () => {
+			formView.disableCssTransitions();
+
 			formView.findInputView.fieldView.select();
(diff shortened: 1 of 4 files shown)


mlewand (approved):
Nice one. I didn't even know we have this `injectCssTransitionDisabler` thing 😄

I see that we have this implemented this way in other places in our codebase.

If I were to implement this what I would do is I'd place it in [FindAndReplaceFormView#focus()](https://github.com/ckeditor/ckeditor5/blob/514bf846642b058c4764f69ad42800cdeefe5457/packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.js#L343-L345) method. I would add a `disableCssTransitions = false` parameter and used it in this line: https://github.com/ckeditor/ckeditor5/blob/514bf846642b058c4764f69ad42800cdeefe5457/packages/ckeditor5-find-and-replace/src/findandreplaceui.js#L193.

This would mean that disable CSS logic would go to FindAndReplaceFormView class. In this PR we're doing some internal things on form view from FindAndReplaceUI class.

But this proposal works, it's in line with how we are doing this in other places so that's fine 👍

mlewand:
Remember about referencing the issue though, it's very helpful in future when someone is wondering why a given thing was added: fcd3665

PR #10099, opened 2021-07-07
Find and replace: toolbar button is now using DropdownButtonView

### LINKED ISSUE #10007: The find and replace toolbar button is using a split button instead of dropdown button

1.  http://fake.ckeditor.com:8081/build/docs/ckeditor5/latest/features/find-and-replace.html

### ✔️ Expected result

The find and replace button is using a `DropdownButtonView` same as `insertTable` button.

### ❌ Actual result

It's using a `SplitButtonView` and action for both sides is the same.

### ❓ Possible solution

## 📃 Other details

*   Browser: …
*   OS: …

---

mateuszzagorski:

Fix (find-and-replace): Find and replace toolbar button now uses `DropdownButtonView`. Closes #10007.

### DIFF
--- a/packages/ckeditor5-find-and-replace/tests/findandreplace.js
+++ b/packages/ckeditor5-find-and-replace/tests/findandreplace.js
@@ -124,7 +124,7 @@ describe( 'FindAndReplace', () => {
 						item.buttonView && item.buttonView.tooltip == 'Find and replace'
 					)[ 0 ];
 
-				itemView.buttonView.arrowView.fire( 'execute' );
+				itemView.buttonView.fire( 'execute' );
 
 				expect( findAndReplaceUI.formView.findButtonView.isEnabled, 'findButtonView' ).to.be.false;
 				expect( findAndReplaceUI.formView.replaceAllButtonView.isEnabled, 'replaceAllButtonView' ).to.be.false;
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-find-and-replace/tests/findandreplace.js
```diff
@@ -116,16 +116,6 @@ describe( 'FindAndReplace', () => {
...
 			it( 'has a proper initial state', () => {
-				// "Initial state" mock.
-				editor.setData( LONG_TEXT );
-
-				const itemView = Array.from( editor.ui.view.toolbar.items )
-					.filter( item =>
-						item.buttonView && item.buttonView.tooltip == 'Find and replace'
-					)[ 0 ];
-
-				itemView.buttonView.arrowView.fire( 'execute' );
```
mlewand, line 127: Why was this preparation logic removed?

This test was meant to test initial state after the dropdown is shown. With this change we check just state after formView was constructed but not really shown.

mlewand: This question is valid for other instances like that.

mlewand (commented):
The fix remains valid however I have questions regarding changes in unit tests.

packages/ckeditor5-find-and-replace/tests/findandreplace.js
mateuszzagorski: You are right, that logic should've stayed there. I've reverted the changes made before.

mlewand (approved):

PR #10104, opened 2021-07-08
[GHS] Table feature integration

### LINKED ISSUE #9914: [GHS] Applying attributes to existing features - table

Due to the complex model structure of the [Table](https://ckeditor.com/docs/ckeditor5/latest/features/table.html) feature, it's not possible to easily extend it with additional attributes. We will need to create separate integration for it.

---

jacekbogdanski:

Feature (html-support): Added General HTML Support integration for Tables feature. Closes #9914.

---

### Additional information

No support for `tfooter` as it's not supported by Table feature itself. `caption`/`figcaption` elements are supported by default GHS itself and don't require any additional integration.

The class name for integration (`TableElementSupport`) doesn't follow the same naming pattern as `CodeBlockHtmlSupport` because it's something we are currently changing in https://github.com/ckeditor/ckeditor5/pull/9891 - although I have some small doubts if actually `*HtmlSupport` is not a better name as we are not providing simple element support (which is already done by GHS) but existing feature integration.

### DIFF
--- a/packages/ckeditor5-html-support/src/integrations/table.js
+++ b/packages/ckeditor5-html-support/src/integrations/table.js
@@ -153,5 +153,4 @@ function getDescendantElement( conversionApi, containerElement, elementName ) {
 			return item;
 		}
 	}
-	return null;
 }
--- a/packages/ckeditor5-html-support/src/integrations/table.js
+++ b/packages/ckeditor5-html-support/src/integrations/table.js
@@ -1,0 +1,157 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module html-support/integrations/table
+ */
+
+import { Plugin } from 'ckeditor5/src/core';
+import { disallowedAttributesConverter } from '../converters';
+import { setViewAttributes } from '../conversionutils.js';
+
+import DataFilter from '../datafilter';
+
+/**
+ * Provides the General HTML Support integration with {@link module:table/table~Table Table} feature.
+ *
+ * @extends module:core/plugin~Plugin
+ */
+export default class TableElementSupport extends Plugin {
+	static get requires() {
+		return [ DataFilter ];
+	}
+
+	init() {
+		const editor = this.editor;
+
+		if ( !editor.plugins.has( 'TableEditing' ) ) {
+			return;
+		}
+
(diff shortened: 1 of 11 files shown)


packages/ckeditor5-html-support/src/integrations/table.js
```diff
@@ -0,0 +1,164 @@
...
+				if ( childNode.is( 'element', 'thead' ) ) {
+					preserveElementAttributes( childNode, 'htmlTheadAttributes' );
+				}
+
+				if ( childNode.is( 'element', 'tbody' ) ) {
+					preserveElementAttributes( childNode, 'htmlTbodyAttributes' );
+				}
+			}
+
+			function preserveElementAttributes( viewElement, attributeName ) {
```
maxbarnas, line 92: Maybe we can move this function directly to the body of `viewToModelTableAttributeConverter()`? Otherwise it'll be created every time the converter for `element:table` is called.

packages/ckeditor5-html-support/src/integrations/table.js
```diff
@@ -0,0 +1,164 @@
...
+// @private
+// @returns {Function} Returns a conversion callback.
+function modelToViewTableAttributeConverter() {
+	return dispatcher => {
+		addAttributeConversionDispatcherHandler( 'table', 'htmlAttributes' );
+		addAttributeConversionDispatcherHandler( 'figure', 'htmlFigureAttributes' );
+		addAttributeConversionDispatcherHandler( 'thead', 'htmlTheadAttributes' );
+		addAttributeConversionDispatcherHandler( 'tbody', 'htmlTbodyAttributes' );
+
+		function addAttributeConversionDispatcherHandler( elementName, attributeName ) {
```
maxbarnas, line 115: Same thing as in line 92.

maxbarnas (changes requested):
Two minor things to change, otherwise LGTM.

packages/ckeditor5-html-support/src/integrations/table.js
jacekbogdanski: Extracting these functions outside the scope of the dispatcher function will require passing multiple (~5) parameters which looks ugly. We are looking into micro pre-optimization here, so I would prefer to keep it more readable instead.

packages/ckeditor5-html-support/src/integrations/table.js
jacekbogdanski: https://github.com/ckeditor/ckeditor5/pull/10104#discussion_r669675048

PR #10174, opened 2021-07-16
JSDoc: Add DOMRect to getOptimalPosition and Rect

fedemp:
The constructor for `Rect` shows this example:

```js
document.body.getClientRects().item( 0 )
```

The return type in every browser but IE is [`DOMRect`](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) so I'm adding it to
the possible values of the `Rect` constructor. I'm also adding `Window`
cause `Rect` can handle it.

For the `target` property of `Options` interface, I'm adding the same
types cause that property is used to create a new `Rect`.

### DIFF
--- a/packages/ckeditor5-utils/src/dom/position.js
+++ b/packages/ckeditor5-utils/src/dom/position.js
@@ -377,7 +377,7 @@ function getAbsoluteRectCoordinates( { left, top } ) {
 /**
  * Target with respect to which the `element` is to be positioned.
  *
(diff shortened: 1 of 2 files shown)


oleq (approved):
Looks good to me! 

We need to allow `DOMRect` in our docs validator, though first https://github.com/ckeditor/ckeditor5-dev/pull/717, so this will have to wait before we can merge it.

PR #10177, opened 2021-07-19
Feature (table): Option in table config to set rows/columns as headings by default

### LINKED ISSUE #10039: It should be possible to enforce tables to contain a header row by default

I've had a discussion with two independent CKEditor 5 users who would like the tables to be inserted with a header row and a regular row by default.

That's something that should definitely go behind an opt-in configuration option.

---

dawidurbanski:

Feature (table): Option in table config to set rows/columns as headings by default. Closes #10039.

---

### Additional information

To test run `yarn manual --files=table/manual/tableautoheadings.js`

### DIFF
--- a/packages/ckeditor5-table/src/commands/inserttablecommand.js
+++ b/packages/ckeditor5-table/src/commands/inserttablecommand.js
@@ -49,8 +49,17 @@ export default class InsertTableCommand extends Command {
 		const model = this.editor.model;
 		const selection = model.document.selection;
 		const tableUtils = this.editor.plugins.get( 'TableUtils' );
+		const config = this.editor.config.get( 'table' );
 
 		const insertionRange = findOptimalInsertionRange( selection, model );
+
+		if ( config.autoHeading && config.autoHeading.rows ) {
+			options.headingRows = config.autoHeading.rows;
+		}
+
+		if ( config.autoHeading && config.autoHeading.columns ) {
+			options.headingColumns = config.autoHeading.columns;
+		}
 
 		model.change( writer => {
 			const table = tableUtils.createTable( writer, options );
--- a/packages/ckeditor5-table/src/commands/inserttablecommand.js
+++ b/packages/ckeditor5-table/src/commands/inserttablecommand.js
@@ -41,10 +41,12 @@ export default class InsertTableCommand extends Command {
 	 * @param {Object} options
 	 * @param {Number} [options.rows=2] The number of rows to create in the inserted table.
 	 * @param {Number} [options.columns=2] The number of columns to create in the inserted table.
-	 * @param {Number} [options.headingRows=0] The number of heading rows.
-	 * If not provided it will default to {@link module:table/table~TableConfig#defaultHeadings `defaultHeadings.rows`} table config.
-	 * @param {Number} [options.headingColumns=0] The number of heading columns.
-	 * If not provided it will default to {@link module:table/table~TableConfig#defaultHeadings `defaultHeadings.columns`} table config.
+	 * @param {Number} [options.headingRows] The number of heading rows.
+	 * If not provided it will default to {@link module:table/table~TableConfig#defaultHeadings `tableConfig.defaultHeadings.rows`}
+	 * table config.
+	 * @param {Number} [options.headingColumns] The number of heading columns.
+	 * If not provided it will default to {@link module:table/table~TableConfig#defaultHeadings `tableConfig.defaultHeadings.columns`}
+	 * table config.
 	 * @fires execute
 	 */
 	execute( options = {} ) {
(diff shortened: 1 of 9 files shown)


Reinmar:
The `autoHeading` option name doesn't work for me – it seems too ambiguous. "Auto" does not specify when that mechanism works. I'm thinking more like `defaultHeadings` because "default" means "at the beginning".

dawidurbanski:
> The `autoHeading` option name doesn't work for me – it seems too ambiguous. "Auto" does not specify when that mechanism works. I'm thinking more like `defaultHeadings` because "default" means "at the beginning".

Renamed. Thanks.

packages/ckeditor5-table/tests/manual/tableautoheadings.md
```diff
@@ -0,0 +1,24 @@
+### Editor 1: Auto insert header in the first row
+
+1. Insert table of chosen size
+2. First row should be marked as header by default
+3. You should be able to make it back to a regular row
```
niegowski, line 5:
```suggestion
1. Insert table of chosen size.
1. First row should be marked as header by default.
1. You should be able to make it back to a regular row.
```

niegowski: Please apply those changes to other places in this file.

packages/ckeditor5-table/src/commands/inserttablecommand.js
```diff
@@ -49,9 +49,18 @@ export default class InsertTableCommand extends Command {
...
 
 		const insertionRange = findOptimalInsertionRange( selection, model );
 
+		if ( config.defaultHeadings.rows ) {
+			options.headingRows = config.defaultHeadings.rows;
+		}
+
+		if ( config.defaultHeadings.columns ) {
+			options.headingColumns = config.defaultHeadings.columns;
+		}
```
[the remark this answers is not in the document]
niegowski: Heading rows should not be greater than available rows/columns.

packages/ckeditor5-table/tests/commands/inserttablecommand.js
```diff
@@ -194,5 +194,106 @@ describe( 'InsertTableCommand', () => {
 				);
 			} );
 		} );
+
+		describe( 'auto headings', () => {
+			it( 'should have first row as a heading by default', async () => {
+				const _editor = await ModelTestEditor
```
niegowski, line 200: Please don't use underscore for local variables.

niegowski (changes requested):

packages/ckeditor5-table/src/commands/inserttablecommand.js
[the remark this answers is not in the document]
niegowski: Please document this default value (from config) in the method JSDoc.

packages/ckeditor5-table/src/commands/inserttablecommand.js
```diff
@@ -49,9 +49,21 @@ export default class InsertTableCommand extends Command {
...
 		const tableUtils = this.editor.plugins.get( 'TableUtils' );
+		const config = this.editor.config.get( 'table' );
 
 		const insertionRange = findOptimalInsertionRange( selection, model );
 
+		const defaultRows = config.defaultHeadings.rows;
+		const defaultColumns = config.defaultHeadings.columns;
+
+		if ( options.headingRows === undefined && defaultRows ) {
+			options.headingRows = Math.min( defaultRows, options.rows );
```
niegowski, line 60: In here you don't know what are the default values for rows and columns, those could be undefined here. Please move clamping to the `TableUtils#createTable()` because it's where those default rows/columns are resolved.

packages/ckeditor5-table/tests/commands/inserttablecommand.js
```diff
@@ -194,5 +194,156 @@ describe( 'InsertTableCommand', () => {
 				);
 			} );
 		} );
+
+		describe( 'auto headings', () => {
+			it( 'should have first row as a heading by default', async () => {
+				editor = await ModelTestEditor
```
niegowski, line 200:
```suggestion
				const editor = await ModelTestEditor
```

niegowski: You should not override the editor from `beforeEach` because it won't be properly destroyed. Also this editor instance should be destroyed locally.

niegowski (changes requested):

godai78:
@niegowski - can you suggest a mention in the documentation about it?

niegowski:
Maybe we should mention it here: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/table.html#basic-table-features-2

godai78:
Yes, I believe this is the right place. Can you come up with the mention for that?

niegowski:
@dawidurbanski could you create an issue (and PR) for that?

dawidurbanski:
Sure. Ticket created (#10292). I'll take care of it.

godai78:
Thank you! Call me for review after :)

PR #10186, opened 2021-07-20
Docs: explicitly require Image plugin in plugin tutorial.

### LINKED ISSUE #10183: InsertImage in Creating a simple plugin guide should require Image plugin

`InsertImage` plugin presented in docs snippet doesn't make sense without the base `Image` plugin.

I know that we explicitly include it to editor config:

```
ClassicEditor
        plugins: [ (...) Image, InsertImage, (...) ],
        toolbar: [ (...) ]
    } )
```

but it should be a explicit dependency to the plugin as without it there's no conversion for image element.

The case we had was that our dev while reading docs just took the plugin code alone and inserted it to other running CKE5 instance. While the button was added and the prompt was shown, nothing happened after insertion (because there was no conversion).

All that remains to do is to add `requires` getter.

Dumluregn:

Docs: explicitly require Image plugin in plugin tutorial. Closes #10183.

### DIFF
--- a/docs/framework/guides/creating-simple-plugin.md
+++ b/docs/framework/guides/creating-simple-plugin.md
@@ -184,10 +184,14 @@ class InsertImage extends Plugin {
 			return view;
 		} );
 	}
+
+	static get requires() {
+		return [ Image ];
+	}
 }
 ```
 
-If you refresh the page, you should now be able to insert new images into the content:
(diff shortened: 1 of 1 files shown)


Dumluregn:
See https://github.com/ckeditor/ckeditor5/issues/10183#issuecomment-1209450974.

PR #10190, opened 2021-07-20
Implement API for the HTML Comments plugin 

### LINKED ISSUE #9661: HTML comments: Model

Handle creating HTML comments in the model.

*   API to create and remove a comment.
*   It'd create a marker and set the root attribute.
*   Note: not a command as it will be used by upcast converters. Let's expose these methods on the plugin.

ma2ciek:

Internal (html-support): Implemented API for the HTML Comments plugin . Closes #9661.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-html-support/src/htmlcomment.js
+++ b/packages/ckeditor5-html-support/src/htmlcomment.js
@@ -8,7 +8,7 @@
  */
 
 import { Plugin } from 'ckeditor5/src/core';
-import { uid } from 'ckeditor5/src/utils';
+import { CKEditorError, uid } from 'ckeditor5/src/utils';
 
 /**
  * The HTML comment feature.
@@ -37,8 +37,8 @@ export default class HtmlComment extends Plugin {
 			view: '$comment',
 			model: ( viewElement, { writer } ) => {
 				const root = this.editor.model.document.getRoot();
-				const markerName = `$comment:${ uid() }`;
 				const commentContent = viewElement.getCustomProperty( '$rawContent' );
+				const markerName = `$comment:${ uid() }`;
 
 				writer.setAttribute( markerName, commentContent, root );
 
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-html-support/tests/htmlcomment.js
```diff
@@ -222,6 +223,263 @@ describe( 'HtmlComment', () => {
...
+	describe( 'createHtmlComment()', () => {
+		it( 'should create an HTML comment between elements', () => {
+			editor.setData( '<p>Foo</p><p>Bar</p><p>Baz</p>' );
+
+			editor.model.change( writer => {
+				const root = editor.model.document.getRoot();
+				const position = writer.createPositionAt( root, 1 );
+
+				htmlCommentPlugin.createHtmlComment( position, 'first' );
+			} );
```
psmyrek, line 235: I think that changing the model using a callback to add a comment is not needed anymore, since we have an API.

```suggestion
			htmlCommentPlugin.createHtmlComment( model.createPositionAt( root, 1 ), 'first' );
```

Similar change could be made in other tests.

packages/ckeditor5-html-support/tests/htmlcomment.js
```diff
@@ -222,6 +223,263 @@ describe( 'HtmlComment', () => {
...
+			);
+
+			const id4 = htmlCommentPlugin.createHtmlComment(
+				createPositionFromPath( [ 2, 3 ] ),
+				'foo'
+			);
+
+			const range = editor.model.change( writer => {
+				return writer.createRangeIn( root.getChild( 2 ) );
+			} );
```
psmyrek, line 395:
```suggestion
			const range = model.createRangeIn( root.getChild( 2 ) );
```

packages/ckeditor5-html-support/tests/htmlcomment.js
```diff
@@ -222,6 +223,263 @@ describe( 'HtmlComment', () => {
...
+				'foo'
+			);
+
+			htmlCommentPlugin.createHtmlComment(
+				createPositionFromPath( [ 2 ] ),
+				'bar'
+			);
+
+			const posStart = createPositionFromPath( [ 2, 1 ] );
+			const posEnd = createPositionFromPath( [ 2, 3 ] );
```
psmyrek, line 414: The `createPositionFromPath()` is not needed since this functionality is available directly in the model.

```suggestion
			const posStart = model.createPositionFromPath( [ 2, 1 ] );
			const posEnd = model.createPositionFromPath( [ 2, 3 ] );
```

packages/ckeditor5-html-support/tests/htmlcomment.js
```diff
@@ -222,6 +223,263 @@ describe( 'HtmlComment', () => {
...
+			const id1 = htmlCommentPlugin.createHtmlComment( position, 'baz' );
+			const id2 = htmlCommentPlugin.createHtmlComment( position, 'biz' );
+
+			const range = new Range( position, position );
+
+			expect( htmlCommentPlugin.getHtmlCommentsInRange( range ) ).to.deep.equal( [ id1, id2 ] );
+		} );
+	} );
+
+	describe( 'getHtmlCommentData()', () => {
```
psmyrek, line 448: There is a missing test for `getHtmlCommentData()` method if comment does not exist.

ma2ciek:
It looks like I don't know the model API at all :see\_no\_evil: 

I'll fix it in a moment.

packages/ckeditor5-html-support/src/htmlcomment.js
```diff
@@ -89,4 +89,120 @@ export default class HtmlComment extends Plugin {
...
+		const marker = editor.model.markers.get( commentID );
+
+		if ( !marker ) {
+			/**
+			 * An HTML comment with the given ID does not exist.
+			 *
+			 * @error html-comment-does-not-exist
+			 */
+			throw new CKEditorError( 'html-comment-does-not-exist', null );
+		}
```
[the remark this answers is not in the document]
ma2ciek: > I wonder why trying to remove a non-existent comment ends with an exception, but trying to read the content of a non-existent comment only returns `null`. Maybe it should be unified?
> 
> Personally, I would do as it is in `MarkerCollection`, i.e. the method for removing a comment should return `true`, when the comment has been found and removed, or `false` otherwise. I think that throwing an exception is not needed here, because when you try to remove a non-existent comment, the editor state is the same as the one expected by the user, i.e. the comment to be removed does not exist anyway, so the user's expectation is fulfilled.

The problem here is the fact that something can be removed either via the API or via the editor. Maybe the current state isn't the best one though. Probably with a `doesHtmlCommentExist( commentID )` both of the above methods could throw errors if the comment does not exist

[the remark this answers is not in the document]
ma2ciek: After a f2f talk, I changed it so these methods do not throw errors and the `removeHtmlComment` returns `true/false` according to the presence of the removed comment.

packages/ckeditor5-html-support/src/htmlcomment.js
```diff
@@ -89,4 +89,120 @@ export default class HtmlComment extends Plugin {
...
+	/**
+	 * Removes an HTML comment with the given comment ID.
+	 *
+	 * It does nothing and returns `false` if the comment with the given ID does not exist.
+	 * Otherwise it removes the comment and returns `true`.
+	 *
+	 * Note that a comment can be removed also by removing the content around the comment.
+	 *
+	 * @param {String} commentID The ID of the comment to be removed.
+	 * @return {Boolean} `true` when the comment with the given ID was removed.
```
psmyrek, line 133:
```suggestion
	 * @returns {Boolean} `true` when the comment with the given ID was removed, `false` otherwise.
```

psmyrek (approved):
LGTM 🎉

PR #10218, opened 2021-07-26
Autosave improvements

### LINKED ISSUE #10214: Autosave is triggered on data init

Autosave callback is fired when editor is initialized which is, at the very best, redundant.

### LINKED ISSUE #10215: API to immediately call autosave callback

At the moment there is no API to in autosave plugin to force immediate call of autosave callback and cancel any already triggered and delayed callbacks. This kind of API is needed if any feature would like to integrate with autosave feature.

scofalik:

Fix (autosave): Autosave callback should not be called while the editor is initialized. Closes #10214.

Feature (autosave): Introduced \`Autosave#save()\`. Closes #10215.

### DIFF
--- a/packages/ckeditor5-autosave/src/autosave.js
+++ b/packages/ckeditor5-autosave/src/autosave.js
@@ -102,7 +102,7 @@ export default class Autosave extends Plugin {
 		this._debouncedSave = debounce( this._save.bind( this ), waitingTime );
 
 		/**
-		 * The last document version.
+		 * The last saved document version.
 		 *
 		 * @private
 		 * @type {Number}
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-autosave/src/autosave.js
```diff
@@ -213,22 +224,8 @@ export default class Autosave extends Plugin {
 	 * @private
 	 */
 	_save() {
-		const version = this.editor.model.document.version;
-
-		// Change may not produce an operation, so the document's version
-		// can be the same after that change.
-		if (
-			version < this._lastDocumentVersion ||
```
scofalik, line 221: It seems to me that this condition can never be met. At most `version` can be the same as `_lastDocumentVersion` but it can never be lower, as `_lastDocumentVersion` is set to it, and document version never goes down.

packages/ckeditor5-autosave/src/autosave.js
```diff
@@ -213,22 +224,8 @@ export default class Autosave extends Plugin {
...
-			this.editor.state === 'initializing'
```
scofalik, line 222: This is wrong assumption. Data change might be triggered during editor initialization (when editor state is `initializing`) but `_save()` might be actually called later (with delay) when the editor state is already `ready`.

packages/ckeditor5-autosave/tests/autosave.js
```diff
@@ -70,24 +70,26 @@ describe( 'Autosave', () => {
...
+						save: spy
+					},
+					initialData: '<p>Foo</p>'
 				} )
 				.then( _editor => {
 					editor = _editor;
 
 					autosave = editor.plugins.get( Autosave );
-
-					const data = '<p>paragraph1</p><p>paragraph2</p>';
```
scofalik, line 89: Moved data initialization where it should be. Didn't have any impact on tests.

packages/ckeditor5-autosave/tests/manual/autosave.js
```diff
@@ -25,14 +33,6 @@ ClassicEditor
 		destroyButton.addEventListener( 'click', () => editor.destroy() );
 
 		const autosave = editor.plugins.get( Autosave );
-		autosave.adapter = {
```
scofalik, line 28: Used config callback instead of adapter because there is no word about adapter anywhere in our docs or guides, it is just mentioned somewhere deeper in API docs. I believe this is forgotten/not recommended way to integrate autosave.

Funny, with this integration, the problem with calling autosave during initialization does not happen (because the adapter is added after the editor is initialized).

packages/ckeditor5-autosave/src/autosave.js
scofalik: Yup, as I thought it was at some point `<=` (which also should simply be `==`) and it was changed to `<` because of markers that affect data but does not bump version: 5577aa2#diff-f3cd7394728f0a6168a108ff56dc6ee631d0a71eb34f1a762264f9339de28311L161

But we don't need to check versions anyway -- `_save()` is called (debounced) only in `change:date` events, so we can be sure that there was actual change.

scofalik: BTW. TBH I don't know what was the use case but if a marker affects data it should also use operations :thinking: .

While use-operation, not-data may make sense, the other combination does not. If marker affects data it has to be send over to other clients.

scofalik: However, we need to check version lower, in the promise resolution:

`if ( this.editor.model.document.version > this._lastDocumentVersion ) {`

And there might be a problem there with the markers that don't bump version but:

1.  This change (removing `if`) is not related to the code below.
2.  As I explained, I am not sure what kind of marker affects data but does not use operations :thinking: .

psmyrek (approved):
LGTM 🎉

PR #10223, opened 2021-07-27
Introduce removing HTML comments at the document boundaries when the content is removed

### LINKED ISSUE #10117: HTML comments: setData() doesn't clear some of existing comments

Steps to reproduce:

1.  Open `HtmlComment` manual test: http://localhost:8125/ckeditor5-html-support/tests/manual/htmlcomment.html.
2.  Call `editor.setData('<!-- Comment --><p>Paragraph</p>')` couple of times.
3.  Call `editor.getData()`. The `<!-- Comment -->` comment is inserted at the end of `$root` boundary multiple times.

It is also easily reproducible using `SourceEditing`:

[https://user-images.githubusercontent.com/56868128/125046490-5c203000-e09e-11eb-9620-f8d03b397243.mp4](https://user-images.githubusercontent.com/56868128/125046490-5c203000-e09e-11eb-9620-f8d03b397243.mp4)

This only happens to comments at the root boundary because other comments are removed from the document once the content in which they are placed is removed:

```
<$root>
    [c1]
    <paragraph>[c2]content[c3]</paragraph>
    [c4]
</$root>
```

Setting data on such a root will remove c2 and c3 because of the exiting mechanism. However, the parent of c1 and c4 doesn't change so they are not cleared.

This is a similar issue to what's described in the first point in [https://github.com/ckeditor/ckeditor5/issues/10119#issuecomment-878907397:](https://github.com/ckeditor/ckeditor5/issues/10119#issuecomment-878907397:)

> *   We use markers to represent comments in the model. Markers use live ranges/positions and these behave just like we see comments to behave here – they are not moved to the graveyard together with the content but rather find their place in the closest position in the remained content. We'd need a mechanism to allow easily remove markers that were affected by content removal. It'd be potentially useful in other cases related to markers too.

ma2ciek:

Internal: Introduced removing HTML comments at the document boundaries when the document content is removed with `editor.data.set()` or `model.removeContent()` . Closes #10117.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-html-support/src/htmlcomment.js
+++ b/packages/ckeditor5-html-support/src/htmlcomment.js
@@ -104,15 +104,15 @@ export default class HtmlComment extends Plugin {
 				const firstPosition = editor.model.createPositionAt( limitElement, 0 );
 				const lastPosition = editor.model.createPositionAt( limitElement, 'end' );
 
-				let affectedCommentMarkerIDs;
+				let affectedCommentIDs;
 
 				if ( firstPosition.isTouching( range.start ) && lastPosition.isTouching( range.end ) ) {
-					affectedCommentMarkerIDs = this.getHtmlCommentsInRange( editor.model.createRange( firstPosition, lastPosition ) );
+					affectedCommentIDs = this.getHtmlCommentsInRange( editor.model.createRange( firstPosition, lastPosition ) );
 				} else {
-					affectedCommentMarkerIDs = this.getHtmlCommentsInRange( range, { skipBoundary: true } );
+					affectedCommentIDs = this.getHtmlCommentsInRange( range, { skipBoundary: true } );
 				}
 
-				for ( const commentMarkerID of affectedCommentMarkerIDs ) {
+				for ( const commentMarkerID of affectedCommentIDs ) {
 					this.removeHtmlComment( commentMarkerID );
 				}
 			}
--- a/packages/ckeditor5-html-support/src/htmlcomment.js
+++ b/packages/ckeditor5-html-support/src/htmlcomment.js
@@ -208,8 +208,8 @@ export default class HtmlComment extends Plugin {
 	 * By default it includes comments at the range boundaries.
 	 *
 	 * @param {module:engine/model/range~Range} range
-	 * @param {Object} options
-	 * @param {Boolean} options.skipBoundaries When set to `true` the range boundaries will be skipped.
+	 * @param {Object} [options]
+	 * @param {Boolean} [options.skipBoundaries=false] When set to `true` the range boundaries will be skipped.
 	 * @returns {Array.<String>} HTML comment IDs
 	 */
 	getHtmlCommentsInRange( range, { skipBoundaries = false } = {} ) {
--- a/packages/ckeditor5-html-support/src/htmlcomment.js
+++ b/packages/ckeditor5-html-support/src/htmlcomment.js
@@ -88,6 +88,35 @@ export default class HtmlComment extends Plugin {
 
 			return true;
 		} );
+
+		// Delete all comment markers from the document before setting new data.
+		editor.data.on( 'set', () => {
+			for ( const commentMarker of editor.model.markers.getMarkersGroup( '$comment' ) ) {
+				this.removeHtmlComment( commentMarker.name );
+			}
+		}, { priority: 'high' } );
+
+		// Delete all comment markers that are within a removed range.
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-html-support/src/htmlcomment.js
```diff
@@ -88,6 +88,35 @@ export default class HtmlComment extends Plugin {
...
+				const firstPosition = editor.model.createPositionAt( limitElement, 0 );
+				const lastPosition = editor.model.createPositionAt( limitElement, 'end' );
+
+				let affectedCommentIDs;
+
+				if ( firstPosition.isTouching( range.start ) && lastPosition.isTouching( range.end ) ) {
+					affectedCommentIDs = this.getHtmlCommentsInRange( editor.model.createRange( firstPosition, lastPosition ) );
+				} else {
+					affectedCommentIDs = this.getHtmlCommentsInRange( range, { skipBoundaries: true } );
+				}
```
psmyrek, line 113: Maybe these few lines could be simplified a bit?

```suggestion
				// Delete comment markers at range boundaries if limit element's boundaries touch the start and end of the removed range.
				const skipBoundaries = !firstPosition.isTouching( range.start ) || !lastPosition.isTouching( range.end );
				const affectedCommentIDs = this.getHtmlCommentsInRange( range, { skipBoundaries } );
```

packages/ckeditor5-html-support/tests/htmlcomment-integration.js
```diff
@@ -969,9 +969,7 @@ describe( 'HtmlComment integration', () => {
 
 			editor.execute( 'delete' );
 
-			// Currently, comments inside a deleted content are moved to adjacent paragraph.
-			// See https://github.com/ckeditor/ckeditor5/issues/10119.
-			//
+			// The following output will be also correct:
```
psmyrek, line 972: I don't get it: when the following output will be also correct? 🤔

packages/ckeditor5-html-support/tests/htmlcomment.js
```diff
@@ -176,51 +176,160 @@ describe( 'HtmlComment', () => {
...
+			const rootAttributes = [ ...root.getAttributeKeys() ].filter( attr => attr.startsWith( '$comment' ) );
+
+			expect( rootAttributes ).to.have.length( 0 );
+
+			const commentMarkers = [ ...editor.model.markers ].filter( marker => marker.name.startsWith( '$comment' ) );
+
+			expect( commentMarkers ).to.have.length( 0 );
+		} );
+
+		it( 'should remove replace all comments with new comments when the whole content is replaced with editor.setData()', () => {
```
psmyrek, line 221:
```suggestion
		it( 'should replace all comments with new comments when the whole content is replaced with editor.setData()', () => {
```

psmyrek (approved):
It works fine. I've found just a few minor changes to consider.

packages/ckeditor5-html-support/src/htmlcomment.js
ma2ciek: Note that the range differs in both cases.

packages/ckeditor5-html-support/tests/htmlcomment-integration.js
ma2ciek: TBH for me, both behaviors would be correct here and I'm not sure what should be the expected behavior.

But perhaps I can remove it.

ma2ciek: Currently, the `comment 4` is at the boundary of the removed range. So as the whole paragraph is removed, you could expect that the comment will be removed too.

packages/ckeditor5-html-support/src/htmlcomment.js
ma2ciek: If both positions touch the boundary of the limit element then these two positions are used to create a range instead of the original ones.

psmyrek: You are right. I assumed that `position.isTouching()` behavies like `position.isEqual()`, but there is a slight difference between them.

packages/ckeditor5-html-support/tests/htmlcomment-integration.js
psmyrek: I got your point, but IMHO it is confusing now. I think that this sentence should be either changed to include i.e. why this output is not valid currently (and `comment 4` is not removed), or just remove it.

Something like: _The following output could be considered as the correct and expected one, but currently the `comment 4` is not removed, because it is not located at the limit element's boundary:_

PR #10233, opened 2021-07-27
jsdoc: Engine>Model>Selection: optional arg in constructor

fedemp:
Given that `selectable` existence is checked at the end of the constructor, the argument is optional


pomek (approved):
LGTM.

pomek:
Thanks, @fedemp.

PR #10120, opened 2021-07-09
[WIP] Fixed baloon position when viewportTopOffset is set

dawidurbanski:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

This PR aims to apply fix to all the contextual balloons for either very tall elements (like super long table, or very tall image etc), that are taller than the editor scrollable area, and also to prevent some balloons to hide under other absolutely / fixed positioned UI elements when `viewportTopOffset` is specified in the editor config.

Type: Message. Closes #9892.

---

### Additional information

I'm going to put more insights to this big issue in the ticket comments (#9892)

Manual test case with fixed balloons for mentioned cases can be checked under (tickets/9892/1.html) test by running:

```
yarn manual --files=ui
```

All the editors and sample manual test cases that are using contextual balloons can be run this way:

```
yarn manual --files=build-balloon-block --files=build-balloon --files=editor-balloon --files=ui --files=letters
```

This PR is still work in progress. For example there are no tests for new cases.
It's done in a way to make all previous tests to pass.

Issue #9672, opened 2021-05-11
Allow modifying toolbar's viewportTopOffset in runtime


Reinmar:
## 📝 Provide a description of the improvement

Currently, `config.toolbar.viewportTopOffset` can only be set when creating the editor.

The value then ends up in `editor.ui.view.stickyPanel.viewportTopOffset` that's readonly.

Could we make this property writable? Or, maybe even better, add a method/getter in the `editor.ui.view` or somewhere deeper? 

The problem I see is that the (in the runtime) `viewportTopOffset` is not a configuration of `editor.ui.view.toolbar` but `editor.ui.view.stickyPanel` which is rather poorly named (seems to have nothing to do with the toolbar). Even if we'll make it writable, discoverability will be relatively poor.

That might be an argument for adding a method to control this property somewhere up the chain (e.g. in `editor.ui.view`) but I dont' want to overcomplicate this so I have mixed feelings.

cc @oleq 

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

oleq:
I'm OK with making `editor.ui.view.stickyPanel.viewportTopOffset` **for `ClassicEditor`**:

*   configurable first via `config.toolbar.viewportTopOffset`,
*   then writable in runtime.

---

> but `editor.ui.view.stickyPanel` which is rather poorly named (seems to have nothing to do with the toolbar). Even if we'll make it writable, discoverability will be relatively poor.

I agree that the name is not perfect (tells you what but not why). It should be renamed to `(sticky)toolbarContainer` or something similar. We can keep  `stickyPanel` for backward compatibility (and warn about deprecation).

---

Still, keep in mind that `InlineEditor` also uses `config.toolbar.viewportTopOffset` but does not have the sticky panel instance. So `editor.ui.view.toolbarContainer.viewportTopOffset` will not work as `InlineEditorUIView` uses `BalloonPanelView` available under `editor.ui.view.panel`. In this case, I'd also rename it to `toolbarContainer` to keep things similar across editor types and, similarly, have `editor.ui.view.toolbarContainer.viewportTopOffset` writable.

## Proposal

*   `editor.ui.view.stickyPanel` -> `editor.ui.view.toolbarContainer` (Classic)
    *   \+ warn when using deprecated property
*   `editor.ui.view.panel` -> `editor.ui.view.toolbarContainer` (Inline)
    *   \+ warn when using deprecated property
[...]

PR #10239, opened 2021-07-28
ViewportTopOffset and BalloonPanelView integration

### LINKED ISSUE #9892: Table toolbar does not respect viewportTopOffset configuration

Use a table longer than the viewport to reproduce the issue. In the screenshot below, I reduced the opacity of the documentation top bar to show the actual position of the toolbar. It should attach to the bottom of the page top bar just like the main editor toolbar does.

## 📃 Other details

*   Browser: …
*   OS: …

---

oleq:

Type: WIP. Closes #9892.

---

### Additional information


oleq:
@ckeditor/qa-team Could you verify if this solution for #9892 is safe, please? Essentially, it touches all kinds of balloons/popups/toolbars across the editor so there's a chance something went wrong.

Also, for more information, you can ping @dawidurbanski.

LukaszGudel:
I'm testing this branch and it works great in our manual tests (`all-features` or `tickets/9892/1.html`). I've also built docs on this branch and there it looks worse 😕 

On page `http://127.0.0.1:8080/ckeditor5/29.0.0/features/table.html`:


https://user-images.githubusercontent.com/58658048/127315047-eb4acd13-9b4f-4efa-853a-9a3c308f20f4.mp4


https://user-images.githubusercontent.com/58658048/127315066-979cdb3e-1d21-4fac-803e-496837c6fe6d.mp4

Table toolbars are hidden behind the header and never jumps to the bottom of the table.

dawidurbanski:
I'm on it. Docs have to be reconfigured to use the new `viewportTopOffset` feature.

dawidurbanski:
@LukaszGudel I updated `table` docs on this branch, but this part will probably get another update in the near future (when #9672 will be done).

You can re-check it in the current shape though.

dawidurbanski:
Also @oleq @Reinmar this PR aims to do couple things at once. One of them is to change `toolbar.viewportTopOffset` config to `ui.viewportOffset.top` for all balloons. https://github.com/ckeditor/ckeditor5/pull/10239/files#diff-a35ba47e37e29680ac1908b00da274cdde808b8e8c80fd18e60c86faf7c27781R25

This makes sense because most of them has nothing to do with the toolbar.

So now the question becomes whether we keep both of these options. One for the toolbar and another one for the rest of balloons. Or do we make all balloons and toolbars to use the new `ui.viewportOffset.top`. Or maybe even a third option, to allow setting a separate value for toolbar with `toolbar.viewportTopOffset` and if not provided then default to `ui.viewportOffset.top`

Case can be made for all of them. The fact is that probably in most of cases (if not all of them), the toolbar offset will be matched with all the other UI elements positions offsets.

In case of making it all under `ui.viewportOffset.top` we probably have to introduce some way of letting people know they now should switch from `toolbar.viewportTopOffset` to `ui.viewportOffset.top`.

I was thinking of some nice option for plugins to be able to specify deprecated config properties.

Something like this (pseudo code):

```js
editor.config.addDeprecationNotice('toolbar.vieportTopOffset', 'ui.viewportOffset.top');
```

And now if someone uses the editor like so:

```js
ClassicEditor
		plugins: [ ... ],
		toolbar: {
			items: [ ... ],
			viewportTopOffset: 100
		}
	} )
```

Will get a console warning similar to this one:

> ⚠️ The `toolbar.vieportTopOffset` configuration option is **deprecated**. It will be removed from future CKEditor versions. Use `ui.viewportOffset.top` instead.

The deprecation check could be injected in the `get` method in the main `Config` class, and more precisely in the [_getFromSource](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/config.js#L195) method.

Reinmar:
> So now the question becomes whether we keep both of these options. One for the toolbar and another one for the rest of balloons. Or do we make all balloons and toolbars to use the new `ui.viewportOffset.top`. Or maybe even a third option, to allow setting a separate value for toolbar with `toolbar.viewportTopOffset` and if not provided then default to `ui.viewportOffset.top`

Less is usually better. So one option would be my choice.

However, the question to answer is whether it's a realistic and frequent enough use case to have to configure the toolbar independently from the rest of the UI.

My intuition says - no. Hence, still single config. But maybe @oleq has some use cases in his mind.

---

As for what to do when someone defined two properties **if** we'll agree that only one makes sense: To answer that we need to think about the migration path for integrators. That's the main scenario, to me, in which people will have both defined by mistake. And then, we should indeed log a warning.

Does warning people when they use the old property (but not the new one yet) make sense? I think so too. It depends on whether the old property will work the same way the new one will. If so, warning 90% of the integrators that they need to change their config has a limited sense (limited > zero). If, however, the new property has a broader effect, then warning makes more sense for sure.

> Will get a console warning similar to this one:
> 
> ⚠️ The `toolbar.vieportTopOffset` configuration option is **deprecated**. It will be removed from future CKEditor versions. Use `ui.viewportOffset.top` instead.
> 
> The deprecation check could be injected in the `get` method in the main `Config` class, and more precisely in the [\_getFromSource](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/config.js#L195) method.

That's not going to work because no one will use `get()` for that property anymore.

dawidurbanski:
> That's not going to work because no one will use `get()` for that property anymore.

Oh, yeah. Indeed. Of course we need to hook this up when setting the property in [_setToTarget](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/config.js#L139).

LukaszGudel:
@dawidurbanski I've checked the docs after your fixes and it works all right in the `features/table` section. I see two issues tho:

1. We have dozens of other editor demos throughout the docs. Most of them do have the `table` or `image` plugins and if I understand correctly, they would also need similar fixes? Right now they are bugged just like on the videos from the https://github.com/ckeditor/ckeditor5/pull/10239#issuecomment-888237192. Also, I've noticed that not all of the demos have the sticky toolbar and on some of them (e.g. `features/export-pdf.html` or CF demos) the toolbar scrolls out of the viewport. How should they be handled?
2. Are we sure about the placement of the table/image toolbar when scrolling on large table/image? It covers the bottom half of the editor toolbar. Imagine that there were 2 more buttons on the left side of the toolbar on this screen:

image


The table/image toolbar would cover all of the alignment buttons and it would be hard to see which one aligns to the left/right/center. What's more, this image/table toolbar in this case covers the button's labels and it's impossible to read them on hover. I feel like UX in this case could be improved.

dawidurbanski:
> if I understand correctly, they would also need similar fixes?

Well yes. But maybe not. It depends on how are we going to treat the old `toolbar.viewportTopOffset` option. If we decide we switch completely to `ui.viewportOffset.top` then I'll go through every single snippet in the docs and I'll update the config. If we decide to still support `toolbar.viewportTopOffset` for some time, we can hold on to this until we eventually remove it in which case we have to update all docs snippets.

For now it's undecided yet, so I just updated all snippets in a single `table` feature page just for showcase.

> Are we sure about the placement of the table/image toolbar when scrolling on large table/image?

I think we just settled on some lesser evil here with @oleq. The idea here is that it's better than it is currently (so it sticks at least), and we added some offset so even if sticky all buttons in the toolbar are at least clickable. We know it's not ideal, but that's what we came up with at this moment.

dawidurbanski:
## Summary of what has been done and what still needs to be done in this ticket

### Done:

- First try: naive implementation of the sticky position tackled in #10120 
- Second try: PoC for adding another position `viewportStickyNorth` by @oleq in #10239

### In progress:

- Recreating #10239 step by step (with adding docs and test as we go) in 4f699f2

### Todo:

- Finishing the **in progress** task, more precisely:
    - in depth review of #10239 (@dawidurbanski is already familiar with what is going on there)
    - adding the rest of the logic from #10239 to the new branch
    - write/modify tests for all the added and modified methods and functions
    - update docs for all changes

dawidurbanski:
Closed in favour of #10373

Issue #10127, opened 2021-07-12
HTML Comments: Support for the clipboard pipeline


ma2ciek:
## 📝 Provide a description of the new feature

Currently, the HTML Comments are based on markers so the clipboard pipeline doesn't handle them during cut/paste or D&D 

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

Reinmar:
I think that there are two cases to consider:

* Copying (& pasting) comments within the editor. This should work but won't work as we don't have the mechanism to downcast markers to clipboard data. 
* Filtering comments out on paste from external sources. #10155 will resolve this situation on PFO, but the remaining part is filtering them out on other content sources. Let's split this to a separate ticket.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity over the last year. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #10275, opened 2021-07-30
Filtered out all pasted HTML comments

### LINKED ISSUE #10213: [HTML comments] Filter out pasted comments

1. Copy content from a website or an external app. Even if browsers don't put comments in the content, other apps might (just like Word/GDocs in #10155).
2. Paste that into the editor.

### ✔️ Expected result

Comments should be filtered out. Comments are most often an internal thing and it makes little sense if they leak to editor data.

### ❌ Actual result

Comments are pasted.

### More info

See also #10127.

---

psmyrek:

Fix (html-support): Filtered out all pasted HTML comments. Closes #10213.

---

### Additional information

I chose to use simple regular expression filtering instead of the recently added `skipComments` option in `DomConverter#domToView()`, because:
* to use `skipComments` option in the `ckeditor5-clipboard` package, I would have to make a few more changes: options that are now passed to `DomConverter#domToView()` would have to be additionally passed to `HtmlDataProcessor#toView()` and `XmlDataProcessor#toView()`. Alternatively, the two `toView()` methods could now only accept one `skipComments` option, but a single option would look silly.
* the existing `normalizeClipboardHtml()` function fits perfectly to the initial cleanup of the input data.

### DIFF
--- a/packages/ckeditor5-clipboard/src/utils/normalizeclipboarddata.js
+++ b/packages/ckeditor5-clipboard/src/utils/normalizeclipboarddata.js
@@ -9,6 +9,7 @@
 
 /**
  * Removes some popular browser quirks out of the clipboard data (HTML).
+ * Removes all HTML comments. Comments are considered as an internal thing and it makes little sense if they leak to editor data.
  *
  * @param {String} data The HTML data to normalize.
  * @returns {String} Normalized HTML.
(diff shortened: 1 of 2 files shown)


ma2ciek:
As I understand this will filter out all comments from the pasted HTML content. I'm not sure if we can distinguish here pasting from the editor and pasting from outside of the editor as we need still to work at copying (& pasting) comments within the editor (#10127).

ma2ciek:
Besides that, it looks fine for me.

psmyrek:
This PR targets only #10213 to filter out all pasted comments, no matter where they come from.

pomek (approved):
LGTM.

PR #10315, opened 2021-08-05
jsdoc: Table>upcastStyleToAttribute param

fedemp:


fedemp:
My bad. I misread the code.

Issue #6682, opened 2020-04-27
The second enter in the code block feature should not escape it


ma2ciek:
## 📝 Provide a description of the improvement

It's pretty common to separate code using empty lines and every editor supporting code blocks should allow such a feature.

However, currently, after pressing enter two times the caret escapes the code block area :cry: This lowers the editing experience quite heavily while working with a large number of code snippets. The only workaround which is known to me is writing code first and pressing enters as the second step of creating a snippet.

IMO only the third enter should create a paragraph and remove the last (empty) line of the code block which was created by the second enter.

## 📃 Other details

*   CKEditor version: 18

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

oleq:
> The only workaround which is known to me is writing code first and pressing enters as the second step of creating a snippet.

<kbd>Shift</kbd>+<kbd>Enter</kbd>

It does not mean I disagree with you. This is a potential UX issue.

mlewand:
@ma2ciek you can put multiple subsequent line breaks using shift + enter

Reinmar:
> IMO only the third enter should create a paragraph and remove the last (empty) line of the code block which was created by the second paragraph.

:+1: The same as in Slack, I believe. Makes sense to me. I was consider reporting this too.

mlewand:
I also use shift+enter in such cases.

If we think about three enters it is a must that it removes tailing enter, because for me subsequent enter press is the only way I know how to leave the code block in a first place - which is something that I use much more often than multiple empty lines.

oleq:
If anyone's interested in researching the feel of a triple enter, [this is where things should be changed](https://github.com/ckeditor/ckeditor5-code-block/blob/5300e22a9ceeb3ba321b1aacd74f9e5435405d1a/src/codeblockediting.js#L332-L410). 

I'm also putting an intro label here because the implementation part of this issue should be quite simple.

ma2ciek:
\><kbd>Shift</kbd>+<kbd>Enter</kbd>

Oh, thanks, IDK why I didn't discover that combo. 

However it seems to be a discoverability issue as you guys mentioned it already.

ma2ciek:
I missed that tip in the https://ckeditor.com/docs/ckeditor5/latest/features/code-blocks.html#tips-and-tweaks section - this should be fixed too.

Reinmar:
[...]

PR #10316, opened 2021-08-05
Make three Enter clicks necessary at the end of a code block to escape it

marcellofuschi:
Improvement of the `CodeBlock` feature, described in https://github.com/ckeditor/ckeditor5/issues/6682.

Closes https://github.com/ckeditor/ckeditor5/issues/6682.

### DIFF
--- a/packages/ckeditor5-code-block/docs/features/code-blocks.md
+++ b/packages/ckeditor5-code-block/docs/features/code-blocks.md
@@ -96,7 +96,7 @@ The code language {@link module:code-block/codeblock~CodeBlockConfig#languages c
 
 There could be situations when there is no obvious way to set the caret before or after a block of code and type. This can happen when the code block is preceded or followed by a widget (e.g. a table) or when the code block is the first or the last child of the document (or both).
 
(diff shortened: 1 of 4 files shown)


marcellofuschi:
Just noticed that this change also needs the GIFs in the docs to be changed.

Reinmar:
Wow, thanks for such a complete PR :) I'll pass it to the team for a review.

arkflpc:
The code looks great and seems to work as intended. I don't know if it should behave the same way at the beginning of the block, though. It doesn't solve the main issue (code snippets rarely start with a blank line) and makes the feature less intuitive (despite being well documented).

@oleq, @Reinmar - what are your opinions?

Reinmar:
I agree. The change of behavior at the beginning of a code block will make it less intuitive. Most likely – too unintuitive.

As for the change at the end of a code block – I think it's a good direction. It should be more intuitive than Shift+Enter that we require currently.

I'm curious about @oleq's opinion still.

oleq:
> I agree. The change of behavior at the beginning of a code block will make it less intuitive. Most likely – too unintuitive.

> I don't know if it should behave the same way at the beginning of the block, though. It doesn't solve the main issue (code snippets rarely start with a blank line) and makes the feature less intuitive (despite being well documented).

I agree with you, this would degrade the UX and should be excluded. 

> As for the change at the end of a code block – I think it's a good direction. It should be more intuitive than Shift+Enter that we require currently.

It makes sense at the end of the code block 👍

PR #10321, opened 2021-08-06
Added a global warning banner about local/nightly documentation

oleq:

Docs (ckeditor5): Added a global warning banner about local/nightly documentation.

---

### Additional information

Requires changes in the documentation builder (in a separate PR).


pomek:
In the proposed solution, I wouldn't say I like displaying the warning banner on localhost.

pomek:
Our documentation builder allows previewing the documentation before deploying it live. In the current solution, the banner will say that we're on a local build of the documentation on the staging server, which is wrong.

I'd change the solution to display the banner only on nightly docs.

oleq:
Alright, let's keep things simple.

pomek (approved):
LGTM.

PR #10323, opened 2021-08-07
Execute indentBlock on Tab key press (configurable)

### LINKED ISSUE #7541: Indent on Tab Key Press

_What is the expected behavior of the proposed feature?_
Apply indentation on `Tab` Key press
---
``

marcellofuschi:
Closes #7541.

Updates to the docs and manual tests are still missing.

### DIFF
--- a/packages/ckeditor5-indent/package.json
+++ b/packages/ckeditor5-indent/package.json
@@ -23,6 +23,7 @@
     "@ckeditor/ckeditor5-paragraph": "^29.1.0",
     "@ckeditor/ckeditor5-theme-lark": "^29.1.0",
     "@ckeditor/ckeditor5-ui": "^29.1.0",
+    "@ckeditor/ckeditor5-utils": "^29.1.0",
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.11"
   },
--- a/packages/ckeditor5-indent/src/indentblock.js
+++ b/packages/ckeditor5-indent/src/indentblock.js
@@ -35,7 +35,8 @@ export default class IndentBlock extends Plugin {
 
 		editor.config.define( 'indentBlock', {
 			offset: 40,
-			unit: 'px'
+			unit: 'px',
+			useTabKey: false
 		} );
 	}
 
@@ -68,6 +69,10 @@ export default class IndentBlock extends Plugin {
 			this._setupConversionUsingClasses( configuration.classes );
 			editor.commands.add( 'indentBlock', new IndentBlockCommand( editor, new IndentUsingClasses( indentConfig ) ) );
 			editor.commands.add( 'outdentBlock', new IndentBlockCommand( editor, new IndentUsingClasses( outdentConfig ) ) );
+		}
+
+		if ( configuration.useTabKey ) {
+			this._setupTabKeystrokesHandlers();
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


oleq:
Hi @marcellofuschi! 

Thank you for the PR. I looked into the code and I tested it. It looks like it introduces some collisions with other editor features that use the Tab key. Here's the summary of my research:

### TableKeyboard plugin

*   `IndentBlock` command is **enabled** when the selection is **anchored in a table cell**.
    *   ❌ Tab indents the cell content instead of navigating across cells.
        *   ✅ When the selection is anchored somewhere where there is no way to indent (e.g. a list item) only then it falls back to the cross-cell navigation.
    *   ❌ Shift+Tab outdents the cell content.
        *   ✅ When there's nothing left to outdent only then it falls back to the cross-cell navigation.
*   `IndentBlock` command is **disabled** when the selection is **over a table**.
    *   ✅ Tab moves the selection to the first cell.
    *   ✅ Shift+Tab blurs the editor (imperfect but beyond the scope of this issue).

🔴 Strong collision with the table keyboard plugin. [All cross-cell navigation listeners](https://github.com/ckeditor/ckeditor5/blob/ca33e9346e0888f51a3b8338d61179f8edf337cd/packages/ckeditor5-table/src/tablekeyboard.js#L46-L49) are attached with low priority. `IndentBlock` uses normal and overrides `TableKeyboard`.

*   I suppose `TableKeyboard` is using low priority to let features like indent list use tab inside table cells.
*   The thing is that indenting lists has its limit because the cross-cell navigation will kick in sooner or later. Indenting blocks will go indefinitely and the user will never move to another cell this way.
*   Since we accepted the collision with indenting lists maybe we can live with indent block too? (It's opt-in anyway)

### Title plugin

*   `IndentBlock` command is **disabled** when the selection is **anchored in a title**.
    *   ✅ Tab moves to content.
    *   ✅ Shift+Tab blurs the editor.
*   `IndentBlock` command is **enabled** when the selection is **anchored in the content.**
    *   ✅  Tab increases indentation
    *   ❓Shift+Tab
        *   ❓ For the **first block in the content** moves the selection back to the title.
            *   No way to outdent the first block in the content via keystroke.
        *   ✅ For the **blocks that follow**, it outdents the blocks.

🟡 There's some collision but I'd say it's negligible (only the first block of the content and `Title` always wins). It's opt-in anyway.

### List plugin

*   `IndentBlock` command is **disabled** when the selection is **anchored in a list item**.
    *   ✅ Tab is indenting a list.
    *   ✅ Shift+Tab is outdenting a list.

🟢 No collision with the list plugin because `NAND(indentBlock#isEnabled, indentList#isEnabled)`.

### Code block plugin

*   `IndentBlock` command is **disabled** when the selection is **anchored in a code block item**.
    *   ✅ Tab is indenting the code (tabbing).
    *   ✅ Shift+Tab is outdenting the code (removing tabs).

🟢 No collision with the code block plugin because `NAND(indentBlock#isEnabled, indentCodeBlock#isEnabled)`.

### LinkUI plugin

_FYI this concerns focusing the visible link balloon after clicking on a link._

🟢 No collision with the Link UI plugin because it sets a listener in `editor.keystrokes` with the `'high'` priority to be faster than other (indent\*) listeners.

### RestrictedEditing plugin

🟢 No collision with the restricted editing plugin because `NAND(indentBlock#isEnabled, goToNextRestrictedEditingException#isEnabled)`.

---

Also, I think `config.indentBlock.useKeyboard` would be more accurate than `useTabKey`. That's only a suggestion, though.

oleq (commented):
As in https://github.com/ckeditor/ckeditor5/pull/10323#issuecomment-975298424.

Issue #504, opened 2017-07-17
Typings for TypeScript


ssougnez:
Hi,

I'm currently using ckeditor5 in TypeScript and I think it would be pretty usefull to ship typings definition. Is there something planed about this? I think it should be possible to auto generated them based on the JS files.

Reinmar:
Quoting myself from https://github.com/ckeditor/ckeditor5/issues/347#issuecomment-309859172:

> We maintain a very thorough API documentation (JSDoc-based notation) so it should not be hard to generate type definitions and maintain them in https://github.com/DefinitelyTyped/DefinitelyTyped.

I agree that it'd be great if typings were available. But we won't be able to work on that now. I wonder, though, how much work it requires, in general, to produce those typings out of JSDoc comments. The problem in our case may lay in the fact that JSDoc didn't work well for ES6 so we came out with https://www.npmjs.com/package/@ckeditor/jsdoc-plugins – a set of plugins which fix stuff after JSDoc and extend it to better work in our case.

nadavsinai:
Typescript definitions would greatly help our cause of developing with ckeditor5 platform

ma2ciek:
I've found only one good looking tool - https://github.com/englercj/tsd-jsdoc, but it doesn't support the `module:` syntax and we use it in a custom way...

And there're few problems to solve anyway.

1. Events 

JSDoc provides the synthetic `@event` and `@fires`, so all events and their documentation will be lost.

From the e.g.:

```js

/**
 * Fired after {@link #initPlugins plugins are initialized}.
 *
 * @event pluginsReady
 */
```

We'd have to create such declarations:

```ts
declare class Emitter<EventMap> {
	listenTo<T extends keyof EventMap>( emitter: Emitter<EventMap>, event: T, callback: ( event: EventInfo<EventMap[ T ]> ) => void, options: any ): void;
	...
}

declare class Editor extends Emitter<EditorEvents> {

}

interface EditorEvents {
	pluginsReady: PluginsReadyEvent;
}

/**
 * Fired after {@link #initPlugins plugins are initialized}.
 */
interface PluginsReadyEvent {}

interface EventInfo<T> {}
```

to not lose the informations about the event. (Then, with every new editor event only the `EditorEvents` interface needs to be changed). I don't think that any JSDoc->TS plugin can support such syntax soon.

2. Builds

[...]

PR #10332, opened 2021-08-09
jsdoc: core>Command replace readonly with protected

fedemp:
After reading the comment at https://github.com/ckeditor/ckeditor5/issues/504#issuecomment-893361105 about properties that are marked as `readonly` when they should be writable, I realized that maybe you people meant `protected` when you used `readonly` here. _Protected_ properties are writable but should only be modified by its own parent, not from the outside e.g. `isEnabled` that should be modified by `refresh()`.

### DIFF
--- a/packages/ckeditor5-core/src/command.js
+++ b/packages/ckeditor5-core/src/command.js
@@ -50,7 +50,7 @@ export default class Command {
 		 * A concrete command class should control this value by overriding the {@link #refresh `refresh()`} method.
 		 *
 		 * @observable
-		 * @readonly
+		 * @protected
 		 * @member #value
 		 */
 		this.set( 'value', undefined );
@@ -91,7 +91,7 @@ export default class Command {
 		 *		enableBold();
 		 *
 		 * @observable
-		 * @readonly
+		 * @protected
 		 * @member {Boolean} #isEnabled
 		 */
 		this.set( 'isEnabled', false );


Reinmar:
According to [https://stackoverflow.com/questions/36843357/typescript-difference-between-private-and-protected-variables,](https://stackoverflow.com/questions/36843357/typescript-difference-between-private-and-protected-variables,) these 2 props are not protected as they can be accessed from other classes. 

We (try to) follow these rules: [https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/code-style.html#accessibility](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/code-style.html#accessibility)

In addition to that, we understand "readonly" as "it can be changed by the class itself, but not by the outside world". So `@public/protected/private` define visibility, while `@readonly` defines who can change it. 

I'm afraid that the semantic that we apply to these terms is different than what TS defined. In this case, when generating typings from our JSDoc comments, I'd suggest stripping information that's incompatible. Changing the meaning of these symbols across CKE5's entire code base is something that would need to be done thoroughly and would thus be a bigger task.

fedemp:
No need to change ckeditor source then. But I would like to make it clear how TS should understand this.

Properties marked as `readonly` in ckeditor are expected to be modified only by the class and its subclasses, right?

At the same time, these properties can be read from the outside, right too?

If those two assumptions are fine, then the best typing would be like

```ts
    get foo():string;
    protected set foo(newvalue: string);
``` 

Screenshot from 2021-08-17 11-10-11

Reinmar:
> Properties marked as `readonly` in ckeditor are expected to be modified only by the class and its subclasses, right?

Yes, that's correct.

> At the same time, these properties can be read from the outside, right too?

Yes.

> If those two assumptions are fine, then the best typing would be like

As far as I understand, yes :) @arkflpc, could you confirm?

arkflpc:
Yes, that's correct. The properties shouldn't be changed by Command clients but by derived classes. However, JSDoc accessibility attributes relate to the whole symbol, not setters and getters. That's why it's read-only.

fedemp:
Excellent. I'll close this PR and i'll start doing a quick check on our typings for those `readonly` that should actually have a `protected` setter.

PR #10336, opened 2021-08-11
Fixing ckeditor5-react/issues/249

pachidj87:
Fix (ckeditor5-ui): Fixing object member error calling this.content.trim().
### Additional information

This fix the issue encountered on https://github.com/ckeditor/ckeditor5-react/issues/249


Reinmar:
Hi! Could you clarify how your PR resolves that issue? From what I can see there, the problem was in the webpack's loaders configuration, not the editor itself.

PR #10342, opened 2021-08-11
jsdoc: Mention>domwrapperview locale

fedemp:
Locale can't be optional if element is required.


Reinmar:
Sadly, it is optional. It's passed  straight to `View()` where it's optional too. Is that conflicting with TS's rules for what can be marked as optional?

fedemp:
> Sadly, it is optional. It's passed  straight to `View()` where it's optional too. Is that conflicting with TS's rules for what can be marked as optional?

Not a problem. I can create an overload so if the first param is an element, the second parameter must be undefined.

PR #10352, opened 2021-08-12
Change viewportTopOffset config. Make viewportTopOffset writable in runtime

### LINKED ISSUE #9672: Allow modifying toolbar's viewportTopOffset in runtime

dawidurbanski:
WIP (check additional information)

Type: Change viewportTopOffset config. Make viewportTopOffset writable in runtime. Closes #9672.

---

### Additional information

I added couple `TODO` comments to point out some stuff that could be improved in this PR IMO.

@oleq if you would find some time to check this one it would be awesome.

These two manual tests are best to see this in action:

```
yarn manual --files=editor-classic/manual/tickets/9672/1.js,editor-inline/manual/tickets/9672/1.js
```

Follow ups on this one:

- [x] ~~Update all our tests to use the new config (`ui.viewportOffset.top` instead of `toolbar.viewportTopOffset`)~~ done in d114b6b
- [x] ~~Update all docs snippets to use the new config~~ done in e4429e5
- [x] ~~Update docs to reflect this change~~ done in be2339d
- [x] ~~Update TODOs when discussed~~ done in 39597fd and 83b130c

### DIFF
--- a/docs/features/toolbar.md
+++ b/docs/features/toolbar.md
@@ -72,12 +72,6 @@ toolbar: {
 
  * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar configuration without the need of defining the entire list (you can specify a couple of buttons that you want to remove instead of specifying all the buttons you want to keep). If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
 
- * **`viewportTopOffset` (deprecated)** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar. Useful when a page with which the editor is being integrated has some other sticky or fixed elements (e.g. the top menu). Thanks to setting the toolbar offset, the toolbar will not be positioned underneath or above the page's UI.
-
-	<info-box warning>
-		This property has been deprecated and will be removed in the future versions of CKEditor. Please use {@link module:core/editor/editorconfig~EditorConfig#ui `EditorConfig#ui.viewportOffset`} instead.
-	</info-box>
-
  * **`shouldNotGroupWhenFull`** &ndash; When set to `true`, the toolbar will stop grouping items and let them wrap to the next line when there is not enough space to display them in a single row. This setting is `false` by default, which enables items grouping.
 
 The demo below presents the "regular" toolbar look with `shouldNotGroupWhenFull` set to `false`. If there are excess toolbar items for the display width, the toolbar gets grouped and some of the items are accessible via the clickable "Show more items" (⋮) button.
--- a/docs/features/toolbar.md
+++ b/docs/features/toolbar.md
@@ -64,7 +64,6 @@ You can use the extended {@link module:core/editor/editorconfig~EditorConfig#too
 ```js
 toolbar: {
     items: [ 'bold', 'italic', '|', 'undo', 'redo', '-', 'numberedList', 'bulletedList' ],
-    viewportTopOffset: 30,
     shouldNotGroupWhenFull: true
 }
 ```
@@ -73,7 +72,11 @@ toolbar: {
 
  * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar configuration without the need of defining the entire list (you can specify a couple of buttons that you want to remove instead of specifying all the buttons you want to keep). If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
 
- * **`viewportTopOffset`** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar. Useful when a page with which the editor is being integrated has some other sticky or fixed elements (e.g. the top menu). Thanks to setting the toolbar offset, the toolbar will not be positioned underneath or above the page's UI.
+ * **`viewportTopOffset` (deprecated)** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar. Useful when a page with which the editor is being integrated has some other sticky or fixed elements (e.g. the top menu). Thanks to setting the toolbar offset, the toolbar will not be positioned underneath or above the page's UI.
+
+	<info-box warning>
+		This property has been deprecated and will be removed in the future versions of CKEditor. Please use {@link module:core/editor/editorconfig~EditorConfig#ui `EditorConfig#ui.viewportOffset`} instead.
+	</info-box>
 
  * **`shouldNotGroupWhenFull`** &ndash; When set to `true`, the toolbar will stop grouping items and let them wrap to the next line when there is not enough space to display them in a single row. This setting is `false` by default, which enables items grouping.
 
--- a/packages/ckeditor5-core/src/editor/editorconfig.jsdoc
+++ b/packages/ckeditor5-core/src/editor/editorconfig.jsdoc
@@ -144,8 +144,6 @@
  *			toolbar: {
  *				items: [ 'bold', 'italic', '|', 'undo', 'redo', '-', 'numberedList', 'bulletedList' ],
  *
- *				viewportTopOffset: 30,
- *
  *				shouldNotGroupWhenFull: true
  *			}
  *		};
@@ -167,9 +165,12 @@
  *
  *	Line break will work only in the extended format when `shouldNotGroupWhenFull` option is set to `true`.
  *
- * * **`toolbar.viewportTopOffset`** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar.
+ * * **`toolbar.viewportTopOffset` (deprecated)** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar.
  * Useful when a page with which the editor is being integrated has some other sticky or fixed elements
  * (e.g. the top menu). Thanks to setting the toolbar offset the toolbar will not be positioned underneath or above the page's UI.
+ *
+ * 	**This property has been deprecated and will be removed in the future versions of CKEditor. Please use `{@link module:core/editor/editorconfig~EditorConfig#ui EditorConfig#ui.viewportOffset}` instead.**
+ *
  * * **`toolbar.shouldNotGroupWhenFull`** &ndash; When set to `true`, the toolbar will stop grouping items
  * and let them wrap to the next line if there is not enough space to display them in a single row.
  *
@@ -296,4 +297,29 @@
  * See the {@glink features/editor-placeholder "Editor placeholder" guide} for more information and live examples.
  *
  * @member {String} module:core/editor/editorconfig~EditorConfig#placeholder
+ */
+
+/**
+ * The editor UI configuration.
+ *
+ *		ClassicEditor
+ *			.create( document.querySelector( '#editor' ), {
+ *				ui: { ... }
+ *			} )
+ *			.then( ... )
+ *			.catch( ... );
+ *
+ * Options which can be set using the UI config:
+ *
+ * * **`ui.viewportOffset`** &ndash; The offset (in pixels) of the viewport from every direction used when positioning a sticky toolbar or other absolutely positioned UI elements.
+ * Useful when a page with which the editor is being integrated has some other sticky or fixed elements
+ * (e.g. the top menu). Thanks to setting the UI viewport offset the toolbar and other contextual balloons will not be positioned underneath or above the page's UI.
+ *
+ *		ui: {
+ *			viewportOffset: { top: 10, right: 10, bottom: 10, left: 10 }
+ *		}
+ *
+ * 	**Note:** At the moment, only `top` property is supported. Support for other directions will be added in the future.
+ *
+ * @member {Object} module:core/editor/editorconfig~EditorConfig#ui
  */
--- a/packages/ckeditor5-core/src/editor/editorconfig.jsdoc
+++ b/packages/ckeditor5-core/src/editor/editorconfig.jsdoc
@@ -321,7 +321,7 @@
  *
  * 	**Note:** At the moment, only `top` property is supported. Support for other directions will be added in the future.
  *
- * 	**Note:** The `top` property is being used by both {@link module:editor-inline/inlineeditoruiview~InlineEditorUIView#viewportTopOffset `InlineEditorUIView#viewportTopOffset`} and {@link module:ui/panel/sticky/stickypanelview~StickyPanelView#viewportTopOffset `StickyPanelView#viewportTopOffset`}. This makes both inline editor toolbar and sticky contextual balloons to respect viewport top offset.
+ * 	**Note:** If you want to modify the viewport offset in runtime (after editor was created), you can do that by overriding {@link module:core/editor/editorui~EditorUI#viewportOffset `editor.ui.viewportOffset`}.
  *
  * @member {Object} module:core/editor/editorconfig~EditorConfig#ui
  */
--- a/packages/ckeditor5-core/src/editor/editorconfig.jsdoc
+++ b/packages/ckeditor5-core/src/editor/editorconfig.jsdoc
@@ -321,5 +321,7 @@
  *
  * 	**Note:** At the moment, only `top` property is supported. Support for other directions will be added in the future.
  *
+ * 	**Note:** The `top` property is being used by both {@link module:editor-inline/inlineeditoruiview~InlineEditorUIView#viewportTopOffset `InlineEditorUIView#viewportTopOffset`} and {@link module:ui/panel/sticky/stickypanelview~StickyPanelView#viewportTopOffset `StickyPanelView#viewportTopOffset`}. This makes both inline editor toolbar and sticky contextual balloons to respect viewport top offset.
+ *
  * @member {Object} module:core/editor/editorconfig~EditorConfig#ui
  */
--- a/packages/ckeditor5-core/src/editor/editorui.js
+++ b/packages/ckeditor5-core/src/editor/editorui.js
@@ -56,8 +56,12 @@ export default class EditorUI {
 		/**
 		 * Stores viewport offsets from every direction.
 		 *
+		 * Viewport offset can be used to constrain balloons or other UI elements into an element smaller than the viewport.
+		 * This can be useful if there are any other absolutely positioned elements that may interfere with editor UI.
+		 *
 		 * @readonly
 		 * @observable
+		 * @member {Object} #viewportOffset
 		 */
 		this.set( 'viewportOffset', this._viewportOffset );
 
--- a/packages/ckeditor5-core/src/editor/editorui.js
+++ b/packages/ckeditor5-core/src/editor/editorui.js
@@ -59,11 +59,22 @@ export default class EditorUI {
 		 * Viewport offset can be used to constrain balloons or other UI elements into an element smaller than the viewport.
 		 * This can be useful if there are any other absolutely positioned elements that may interfere with editor UI.
 		 *
+		 * Example:
+		 *
+		 * ```js
+		 * {
+		 * 	top: 50,
+		 * 	right: 50,
+		 * 	bottom: 50,
+		 * 	left: 50
+		 * }
+		 * ```
+		 *
 		 * @readonly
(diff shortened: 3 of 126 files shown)


oleq (commented):
I think this is the right direction 👍

I read your comments @dawidurbanski but instead of replying under individual lines of code, I decided to fork your solution (`ck/9672-modify-viewporTopOffset-in-runtime-v2`) with a [single commit](https://github.com/ckeditor/ckeditor5/commit/6e76b5a1d09f2912c901d225ccc2fbf40633cb79) so it is easier to understand what direction I think would be the best. And besides, unlike GH comments, you can run it straight away 😀

In short, 

1. to avoid code duplication, I'm for moving the [offset normalization to `EditorUI`](https://github.com/ckeditor/ckeditor5/commit/6e76b5a1d09f2912c901d225ccc2fbf40633cb79#diff-8f1cc4aa3b82dd2712c650152a49f78304aa1c20bf325c62f6c2ec0858708654R60). All editors with the UI inherit from this class so it's a safe spot IMO.
   * this makes the runtime config available for developers under `editor.ui.viewportOffset`
   * alternatively: this could be moved from `EditorUI` to `EditorUIView`. The only drawback is that the view needs to obtain the editor config via constructure which is, um..., an antipattern (our architecture assumes UI views are completely decoupled from the editor), but I guess this would also work. The property would be `editor.ui.view.viewportOffset`. 
   * now that I think about it, `EditorUIView` could be a better place because `EditorUI` is a controller (glue) and this whole situation is just pure geometry (presentation). All editors with the UI use `EditorUIView` at some point. WDYT?
2. speaking of `viewportOffset` (instead of `viewportTopOffset`), I think it would be best if we made it future-proof and, as you correctly assumed in your code, used an object with `top`, `bottom`, `left`, `right` properties. For now, only the `top` makes sense but this will change one day.
   * the implementation in the twin issue (the one with the "sticky position") allows restricting the viewport Rect from all sides, anyway, so these changes would work together out-of-the-box.
3. moving on, I bound [`editor.ui.view.stickyPanel.viewportTopOffset` (Classic)](https://github.com/ckeditor/ckeditor5/commit/6e76b5a1d09f2912c901d225ccc2fbf40633cb79#diff-f1cff316ea0a20ab6f81df4c5fc96be3bab302b49e04c170bf7c5d961c839c41R145) and [`InlineEditorUIView.viewportTopOffset` (Inline)](https://github.com/ckeditor/ckeditor5/commit/6e76b5a1d09f2912c901d225ccc2fbf40633cb79#diff-99b6f8366a55057302812d639ce384164db9b0ce7cb52f56ea80959a87975cc8R128) to `EditorUI#viewportOffset`.
   * this way, the whole viewport offset situation is transparent to developers. They don't need to worry about which editor type is used and which property should be changed in the runtime. There's a single source of truth for all editors with UI.

dawidurbanski:
Thanks @oleq !

> now that I think about it, EditorUIView could be a better place because EditorUI is a controller (glue) and this whole situation is just pure geometry (presentation). All editors with the UI use EditorUIView at some point. WDYT?

Let's stick to the the `EditorUI` with this. It felt awkward in the `UIView` for me.

I also added some missing pieces, small refactoring, tests, some initial docs, and proper comments in 83b130c commit.

The only thing left now is to update rest of the docs including snippets and it should be job done. I'm going to complete it soon.

docs/features/toolbar.md
```diff
@@ -73,7 +72,11 @@ toolbar: {
 
  * **`removeItems`** &ndash; An array of toolbar item names. With this setting you can modify the default toolbar configuration without the need of defining the entire list (you can specify a couple of buttons that you want to remove instead of specifying all the buttons you want to keep). If, after removing an item, toolbar will have two or more consecutive separators (`'|'`), the duplicates will be removed automatically.
 
- * **`viewportTopOffset`** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar. Useful when a page with which the editor is being integrated has some other sticky or fixed elements (e.g. the top menu). Thanks to setting the toolbar offset, the toolbar will not be positioned underneath or above the page's UI.
+ * **`viewportTopOffset` (deprecated)** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar. Useful when a page with which the editor is being integrated has some other sticky or fixed elements (e.g. the top menu). Thanks to setting the toolbar offset, the toolbar will not be positioned underneath or above the page's UI.
```
Reinmar, line 75: I'm afraid this bit needs to be rethought. Most of the people who will read this guide ever will not know that this option ever existed. So, it makes sense to only briefly mentioned it that it wasthere or not even mention at all.

Instead, we need to think on how to add a section about `config.ui.viewportSth`. It might be a new section (but it's unclear to me how it could be titled... "Configuring the viewport dimensions"? "Limiting the viewport area"?) or we can just mention this property here like it was part of toolbar configuration. No one said that toolbar config cannot be outside `config.toobar` too. The third option will be to use `<info-box>` to detach this information a bit from its surrounding.

tl;dr: When writing docs (not migration guides) focus on what's available currently.

packages/ckeditor5-core/src/editor/editorconfig.jsdoc
```diff
@@ -167,9 +165,12 @@
  *
  *	Line break will work only in the extended format when `shouldNotGroupWhenFull` option is set to `true`.
  *
- * * **`toolbar.viewportTopOffset`** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar.
+ * * **`toolbar.viewportTopOffset` (deprecated)** &ndash; The offset (in pixels) from the top of the viewport used when positioning a sticky toolbar.
```
Reinmar, line 168: This is fine that it's mentioned here. API docs are the place to cover deprecated things.

packages/ckeditor5-core/src/editor/editorconfig.jsdoc
```diff
@@ -297,3 +298,28 @@
  *
  * @member {String} module:core/editor/editorconfig~EditorConfig#placeholder
  */
+
+/**
+ * The editor UI configuration.
```
Reinmar, line 303: I miss a link to toolbar configuration guide where this option will be explained in a context of what it affects.

packages/ckeditor5-core/src/editor/editorui.js
```diff
@@ -12,8 +12,7 @@
 import ComponentFactory from '@ckeditor/ckeditor5-ui/src/componentfactory';
 import FocusTracker from '@ckeditor/ckeditor5-utils/src/focustracker';
 
-import EmitterMixin from '@ckeditor/ckeditor5-utils/src/emittermixin';
-import mix from '@ckeditor/ckeditor5-utils/src/mix';
+import { ObservableMixin, mix } from '@ckeditor/ckeditor5-utils';
```
Reinmar, line 15: Revert please. It's not consistent with how the rest of the imports look here.

I started wondering though, what's the expected way to write these imports. I need to fwd this to the devops team because this deserves to be linted.

Reinmar: Reported: #10375

packages/ckeditor5-core/src/editor/editorui.js
```diff
@@ -53,6 +52,14 @@ export default class EditorUI {
 		 */
 		this.focusTracker = new FocusTracker();
 
+		/**
+		 * Stores viewport offsets from every direction.
```
Reinmar, line 56: I'm afraid that this may be severely unclear :D What is a viewport offset? How it's used? Giving an example could be enough.

packages/ckeditor5-core/src/editor/editorui.js
```diff
@@ -53,6 +52,14 @@ export default class EditorUI {
...
+		 *
+		 * @readonly
```
Reinmar, line 58: This will not appear in the docs AFAIK. You need to use `@member` if that's set by `this.set()`. JSDoc is able to recognize `this.x = 1`, but not this.

packages/ckeditor5-core/src/editor/editorui.js
```diff
@@ -172,6 +179,46 @@ export default class EditorUI {
 		return this._editableElementsMap;
 	}
 
+	/**
+	 * Returns viewport offsets object
```
Reinmar, line 183: Use backticks to wrap code fragments.

packages/ckeditor5-core/src/editor/editorui.js
```diff
@@ -172,6 +179,46 @@ export default class EditorUI {
...
+	 * { top: Number, right: Number, bottom: Number, left: Number }
+	 * Only top property is currently supported
```
Reinmar, line 185: Missing:

*   A period at the end.
*   Blank link to separate paragraphs of text.

packages/ckeditor5-core/src/editor/editorui.js
dawidurbanski: @oleq  made this change, not sure thee reason behind. I'll make it the old way :)

packages/ckeditor5-core/src/editor/editorui.js
Reinmar: Missing `:` at the end. Also, the code snippet deserves to be a block snippet, so a triple backtick.

tldr something like:

````
Returns viewport offsets object:

```js
{ ... }
```

Only something.
````

File: packages/ckeditor5-core/src/editor/editorui.js [hunk 9]
```diff
@@ -172,6 +179,46 @@ export default class EditorUI {
...
+		const editor = this.editor;
+		const viewportOffsetConfig = editor.config.get( 'ui.viewportOffset' );
+
+		if ( viewportOffsetConfig ) {
+			return viewportOffsetConfig;
+		}
+
+		const legacyOffsetConfig = editor.config.get( 'toolbar.viewportTopOffset' );
+
+		// Fall back to deprecated toolbar config
```
Reinmar, line 200: End sentences with periods.

packages/ckeditor5-core/src/editor/editorui.js
```diff
@@ -172,6 +179,46 @@ export default class EditorUI {
...
+		if ( legacyOffsetConfig ) {
```
Reinmar, line 201: OK, I have very mixed feelings seeing a warning in a `getFoo()` place (a getter for now, but can be a function too). It's just not part of the "reading things logic" to me. However, this is a deprecation notice and after giving it more thought I'm ok after having it here.

packages/ckeditor5-core/tests/editor/editorui.js
```diff
@@ -202,4 +202,32 @@ describe( 'EditorUI', () => {
 			sinon.assert.calledWithMatch( stub, 'editor-ui-deprecated-editable-elements' );
 		} );
 	} );
+
+	describe( '_viewportOffset()', () => {
```
Reinmar, line 206: If it's private it must not be accessed by other classes. See https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/code-style.html#visibility-levels.

packages/ckeditor5-core/src/editor/editorui.js
```diff
@@ -172,6 +179,46 @@ export default class EditorUI {
...
 
+	/**
+	 * Returns viewport offsets object
+	 * { top: Number, right: Number, bottom: Number, left: Number }
+	 * Only top property is currently supported
+	 *
+	 * @private
+	 * @return {Object}
+	 */
+	get _viewportOffset() {
```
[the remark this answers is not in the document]
Reinmar: OTOH, if you need to expose it for test purposes, which we sometimes do to make them simpler, it can be a method. Just avoid such complex getters. `_readViewportOffsetFromConfig()` would work better for me.

packages/ckeditor5-core/tests/editor/editorui.js
Reinmar: Also, I didn't yet see tests, but do we have an integration test for the warning? I can see the second test here, but what if `_viewportOffset` is never accessed for some reason? Who ensures that it is?

packages/ckeditor5-core/src/editor/editorui.js
Reinmar: Also, you need to define the type of this property and please provide some example on how to override it.
As for the type, you can go the easy way and define it as `Object` and explain what keys it has in the text.

Reinmar (changes requested):
OK, I'm done with my review. The change works - the top offset is writable in runtime.

There are two things that I'd want to add:

* The migration guide is necessary and it doesn't need to be part of this ticket, but we need to do it. Please open a followup and ping me about it.
* It's a pity that the toolbar doesn't change its position once you update `editor.ui.viewportOffset`. I also checked if calling `editor.ui.update()` works but it doesn't. Could you make a quick check whether there's an option to update (recalculate) the position of it on demand. If so, please report a followup ticket too.
*

packages/ckeditor5-core/src/editor/editorconfig.jsdoc
dawidurbanski: The idea here is to not connect it with the toolbar at all. It may affect anything, not just the toolbar.

Also, the `ui` namespace may be used to add any other `ui` related config stuff in the future that may not have anything to do with the toolbar at all.

Since we are completely removing the `vieportTopOffset` from the toolbar config docs, it may be confusing IMHO.

docs/features/toolbar.md
dawidurbanski: Removed

packages/ckeditor5-core/src/editor/editorconfig.jsdoc
dawidurbanski: I linked `ui.viewportOffset` config (in a note) to both places that use the `top` value and vice-versa. I liked both of these places to the `ui` config.

If you are about to build the docs locally, you'll find these under links:

Config: http://127.0.0.1:8080/ckeditor5/29.1.0/api/module_core_editor_editorconfig-EditorConfig.html#member-ui

Inline Editor: http://127.0.0.1:8080/ckeditor5/29.1.0/api/module_editor-inline_inlineeditoruiview-InlineEditorUIView.html#member-viewportTopOffset

Sticky Panel View: http://127.0.0.1:8080/ckeditor5/29.1.0/api/module_ui_panel_sticky_stickypanelview-StickyPanelView.html#member-viewportTopOffset

packages/ckeditor5-core/src/editor/editorui.js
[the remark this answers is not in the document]
dawidurbanski: After some thought I realised what actually this method do is to normalize viewportOffset config. Shall I maybe extract it to a separate (even public) function similar to [normalizeToolbarConfig()](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-ui/src/toolbar/normalizetoolbarconfig.js)?

What do you think @Reinmar ?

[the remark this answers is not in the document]
Reinmar: It makes sense to extract it if it's going to be reused somewhere else. If not, keep it local cause KISS.

packages/ckeditor5-core/tests/editor/editorui.js
dawidurbanski: I changed it to private method, and updated tests to test against `ui.viewportOffset` instead of this private method.

> Also, I didn't yet see tests, but do we have an integration test for the warning?

Yes: [here](https://github.com/ckeditor/ckeditor5/blob/b609fcfa4502eeb3bfd28aec115f5f6c0121d0a6/packages/ckeditor5-editor-classic/tests/classiceditorui.js#L89) and [here](https://github.com/ckeditor/ckeditor5/blob/b609fcfa4502eeb3bfd28aec115f5f6c0121d0a6/packages/ckeditor5-editor-inline/tests/inlineeditorui.js#L88).

I updated both of these tests to check for

```
sinon.assert.calledWithMatch( spy, 'editor-ui-deprecated-viewport-offset-config' );
```

instead of just checking for `calledOnce` for easier searching.

The only thing is that I could potentially extract it to a separate case, but it just felt natural to test this warning in one go.

> I can see the second test here, but what if _viewportOffset is never accessed for some reason? Who ensures that it is?

I have troubles finding any possible scenario when it's not accessed. Maybe after my last update to the code this became obsolete.

We do set `viewportOffset` inside `EditorUI` constructor. So every editor that has an UI will call it. If editor has no UI we don't care about setting offset data, nor warning about deprecated property usage.

packages/ckeditor5-core/src/editor/editorui.js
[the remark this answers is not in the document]
dawidurbanski: Left it as a private method called `_readViewportOffsetFromConfig`

dawidurbanski:
> * The migration guide is necessary and it doesn't need to be part of this ticket, but we need to do it. Please open a followup and ping me about it.

Done: #10401

> * It's a pity that the toolbar doesn't change its position once you update `editor.ui.viewportOffset`. I also checked if calling `editor.ui.update()` works but it doesn't. Could you make a quick check whether there's an option to update (recalculate) the position of it on demand. If so, please report a followup ticket too.

In these two tests:

```
yarn manual --files=editor-classic/manual/tickets/9672/1.js,editor-inline/manual/tickets/9672/1.js
```

We actually just set new `viewportOffset` by clicking on a button outside of the editor. This means we need to focus back in to it after it's done. We do make this step automated by adding `editor.editing.view.focus();` on button click after we are done with updating offset to new, random one.

Apparently focusing out and back in is sufficient to reposition the toolbar.

Example:

```js
const newOffset = 200;
document.documentElement.style.setProperty( '--top-offset', `${ newOffset }px` );
editor.ui.viewportOffset = { top: newOffset };
editor.editing.view.focus();
```

I added an additional, naive case to both of these tests. An another button to start updating top offset randomly in two seconds interval. This way we do keep focused in the editor while offset is being updated.

Indeed in such case editor toolbar does not get updated.

I found we just need to fire `layoutChanged` event after updating the offset.

Example:

```js
const newOffset = 200;
document.documentElement.style.setProperty( '--top-offset', `${ newOffset }px` );
editor.ui.viewportOffset = { top: newOffset };
editor.editing.view.document.fire( 'layoutChanged' );
```

In my opinion It's just a matter of proper docs to describe these cases.

Reinmar:
> Indeed in such case editor toolbar does not get updated.

That's the case to worry about. It's clear that re-focusing resolves this issue. And it's good there's even a better workaround with `layoutChanged`. However, both of these are hacks. The straightforward solution would be if calling `editor.ui.update()` worked because that method is supposed to be called whenever the editor's UI should be updated due to some external changes. I'll write that in the issue you opened too.

Reinmar:
FYI: The merge commit message that I went with:

> Feature: Introduced \`editor.ui.viewportOffset\` that allows modifying in runtime the viewport's offset. This value is used by the editor to e.g. position its sticky toolbar and contextual balloons. Additionally, the \`config.toolbar.viewportTopOffset\` property was moved to \`config.ui.viewportOffset\` and it now accetps an object. Closes #9672.
> 
> BREAKING CHANGE: The \`config.toolbar.viewportTopOffset\` property was moved to \`config.ui.viewportOffset\` and it now accetps an object.

wimleers:
Looks like this also unblocked https://www.drupal.org/project/drupal/issues/3259380 per @lauriii! 🥳

PR #10370, opened 2021-08-17
Replace relative paths to absolute in the live demos

pomek:

Internal: Added a script for replacing all relative paths to absolute in the documentation executed on each page where live demos are available.

---

Closes cksource/ckeditor5-internal#852.

### DIFF
--- a/docs/_snippets/assets.js
+++ b/docs/_snippets/assets.js
@@ -6,6 +6,7 @@
 /* global console, window, document */
 
 import tippy from 'tippy.js';
+import isRelativeUrl from 'is-relative-url';
 
 import 'tippy.js/dist/tippy.css';
 import 'tippy.js/themes/light-border.css';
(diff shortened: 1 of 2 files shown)


psmyrek (approved):
LGTM 🎉 

One thing to consider: instead of using `window.location.href` to produce the base URL for an image, maybe it would be better to get the base URL from the `baseURI` property from the image? In some rare cases these two things can be different.

> The `baseURI` read-only property returns the absolute base URL of a `Node`.
> 
> The base URL is used to resolve relative URLs when the browser needs to obtain an absolute URL, for example when processing the HTML `` element's `src` attribute or XML `xlink:href` attribute.
> 
> In most cases the base URL is the location of the document, but it can be affected by many factors, including the `<base>` element in HTML and the `xml:base` attribute in XML.

psmyrek:
Another thing to consider: what about other elements in the editable that may also have relative paths in their attributes, i.e. the `<picture>` with multiple `<source>` elements, or the `<video>` that we could host locally? After a brief look at the documentation, I did not find such elements so far, but the question is whether this solution should be ready for it?

pomek:
@psmyrek, I pushed a commit that fixes mentioned issues: 

- Usage of `img.baseURI` instead of `window.location`
- Update children of the `<picture>` element

psmyrek (approved):
Thanks for adding these changes.

LGTM 🎉

Issue #625, opened 2017-10-20
Add code of examples


Reinmar:
We knew how important this is for the developers but we didn't manage to add code examples to live examples.

I've just checked how many visits we had from various sources to this repo and the first referer is docs.ckeditor.com. From which pages precisely?

![image](https://user-images.githubusercontent.com/156149/31837070-b32ce60e-b5d7-11e7-9229-fbfbc68fd1c0.png)

Yep, from examples. I wondered why and it gets clear when you see the website:

![image](https://user-images.githubusercontent.com/156149/31837093-c3489dd0-b5d7-11e7-9fd7-fca770c54b2f.png)

Basically, people hoped to see the code of these examples.

Let's fix this. We can do this in two steps:

1. Let's add a short paragraph of text above/below each example with a link to the Quick Start guide saying that the user can find the source there.
2. Let's work on wrapping examples with tabbed widget presenting "output", "HTML", "JS" and "CSS" if needed. A link to JSFiddle would actually be doable now too (in most of the examples), but to reduce the scope this can be done later.

@AnnaTomanek, could you handle the first part? Please also ping @m-turek so we publish the changes once they are done.

Reinmar:
https://github.com/ckeditor/ckeditor5/issues/1608#issuecomment-474338643

> If we want to add the inspector to all demos of the editor in our docs, we should/could try to do that together with displaying the source code of these samples (reported in #625). Therefore, let's for now focus on mentioning the inspector in a couple of places, rather than on enabling it.

Reinmar:
FYI: again request for source code 😄 
<img width="855" alt="Screenshot 2019-04-05 at 09 24 29" src="https://user-images.githubusercontent.com/10219857/55611098-d1e11700-5784-11e9-9815-748e815e9b54.png">

godai78:
My suggestion is to follow suit and add the code of examples as we did for full-featured and bottom-toolbar examples.  
  
Using the real code listing from the examples would make the editor (almost) immediately deployable hence of hight usability for integrators.  
  
The said examples:

*   full-featured: [https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/full-featured-editor.html](https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/full-featured-editor.html)
*   bottom toolbar: https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/bottom-toolbar-editor.html

PR #10385, opened 2021-08-20
Docs: links to QS from examples; source code.

godai78:

Docs: Adding Quick Start guide links to Examples. 

Docs: Adding HTML source to Examples.

---

### Additional information

This addresses in part (but not necessarily closes) the following issue: #625

### DIFF
--- a/docs/examples/builds/balloon-block-editor.md
+++ b/docs/examples/builds/balloon-block-editor.md
@@ -12,3 +12,68 @@ classes: main__content--no-toc
 * a {@link features/blocktoolbar block toolbar} accessible using a button attached to the editable content area and following the selection in the document (bringing additional block formatting tools).
 
 {@snippet examples/balloon-block-editor}
+
+## Source code
+
+Check out the {@link builds/guides/quick-start#balloon-block-editor Quick start} guide to learn more about implementing this kind of editor. You will find implementation steps there.
+
+You can also view the HTML code of the above example here.
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="utf-8">
+	<title>CKEditor 5 – Classic editor</title>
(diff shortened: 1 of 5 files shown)


Reinmar:
As I mentioned f2f, I don't feel like the right person to review this PR. I'd rather check it within the CCX team.

My 2 cents though on what could be checked/improved:

*   Ideally, the code of examples should be well separated from the demo itself, to avoid visual clutter. Not everyone is interested in the code itself too (all less technical people checking the docs). Alternatively, the entire source code section could be in something like `<details>`.
*   All examples should look identical in the :point\_up: aspect.
*   All examples should have the Source code section. Therefore the PR requires additional thought and work to cover more complex examples. E.g. the Source code in the inline editor example is incorrect.
    *   In fact, for the above reason, I have some doubts regarding copy-pasting the examples source code to the guide. The source code will not be tested and if we change something in an example itself, we need to update the source code too. And test.
*   We have some new examples today that require to be covered now.
*   Please remember about correct indentation in code examples. [https://github.com/ckeditor/ckeditor5/pull/10385/files#diff-6c17c8417baae8a66cd75ad31cd3a4f16b03d55abe8f040b558b76a8bb6d3c69](https://github.com/ckeditor/ckeditor5/pull/10385/files#diff-6c17c8417baae8a66cd75ad31cd3a4f16b03d55abe8f040b558b76a8bb6d3c69) has the \<div> and \<script> indented differently.
*   If we'll keep the source code in the MD files like it was proposed... someone has to test these examples.

Reinmar (changes requested):
As commented.

Reinmar:
I'd seriously consider giving this problem more thought and involving the Devops team to think how to create these examples in a safer way to resolve the issues that I mentioned.

godai78:
Deprecated and closed in favor of: #11860

PR #10386, opened 2021-08-20
Added guide section explaining our custom linter rules, added link to said section in linter errors in console

### LINKED ISSUE #10051: Make a page in docs with explanation for our custom linter rules

There are some liter rules that could use further clarification. For example, we just added rule saying:

> Imports from the \`ckeditor5\` package must use the \`src/\` directory.

That says what not to do but it does not say **why** you should not do this way.

tl;dr this would blow up usages outside of our monorepo.

I initially wanted to place explanation in the error message but we had a quick chat with @pomek who pointed that it would make the message very obscure as you can't explain it in a single sentence.

Another example I can tell are DLL-related linter errors.

It would be nice to have a dedicated page, like our error codes page where we'd explain it. **This page should be linked as a "banner text" when linter finds some errors**.

---

przemyslaw-zan:
Docs: Added guide section explaining our custom linter rules. Closes #10051.  
  
Internal: Linter errors will now display a link to custom linter rules page.

### DIFF
--- a/docs/framework/guides/contributing/code-style.md
+++ b/docs/framework/guides/contributing/code-style.md
@@ -777,3 +777,58 @@ Widely used standard files do not obey the above rules:
 * `README.md`, `LICENSE.md`, `CONTRIBUTING.md`, `CHANGES.md`
 * `.gitignore` and all standard "dot-files"
 * `node_modules`
+
+## CKEditor 5 custom ESLint rules
+### Importing from the same package
+While importing modules from the same package it is allowed to use relative paths, like so:
+```js
+import Foo from '../foo';
+import Foo from '../../foo';
+```
+### Importing from other packages
+While importing modules from other packages it is disallowed to use relative paths, and absolute paths must be used instead, like so:
+```js
+import Position from '@ckeditor5/ckeditor5-engine/src/model/position';
+```
+👎 Examples of incorrect code for this rule:
+```js
+import Foo from '../ckeditor5-media-embed/src/foo';
+import Position from '../ckeditor5-engine/src/model/position';
+import Position from '../../ckeditor5-engine/src/model/position';
+import Position from '../../../../../../../../../../ckeditor5-engine/src/model/position';
+```
+👍 Examples of correct code for this rule:
+```js
+import Foo from '@ckeditor/ckeditor5-media-embed/src/foo';
+import Position from '@ckeditor/ckeditor5-engine/src/model/position';
+```
+### Docs for new errors
+Each time new type of error is created, it needs a comment that contains description, and `@error error-name` expression, like so:
+```js
+/**
+ * Description of why the error was thrown
+ *
+ * @error method-id-is-kebab
+ */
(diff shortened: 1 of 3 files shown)


docs/framework/guides/contributing/code-style.md
```diff
@@ -777,3 +777,126 @@ Widely used standard files do not obey the above rules:
...
+```js
+// Assume we edit a file located in the path: `packages/ckeditor5-engine/src/model/model.js`
+
+import Position from './position';
+import insertContent from './utils/insertcontent';
+```
+
+While importing modules from other packages, it is not allowed to use relative paths, and the import must be done using the package name, like this:
+
+👎&nbsp; Examples of incorrect code for this rule:
```
psmyrek, line 798: One thing to consider. There is a description of the rule followed by the "like this", which is followed by the incorrect sample code. This is confusing. If the first code sample is for the incorrect case, the rule description should be followed by "not like this". But IMHO it would be better to not use the "like this" or "not like this" sentences at all.

godai78: Having thought it over, I would, too, concentrate more on showing examples of good code, while removing the wrong ones. These will google and index, too.

godai78 (approved):

docs/framework/guides/contributing/code-style.md
pomek: When we were writing these sections, we based on https://eslint.org/docs/rules/no-undef.

pomek: WDYT about removing ", like this" from the sentence?

> While importing modules from other packages, it is not allowed to use relative paths, and the import must be done using the package name.
>
> Example of incorrect code...
> Example of correct code...

godai78: And this is cool, be we do not really adopt the idea of a good example / bad example in our own docs.

godai78: I'm all for removing "like this" (even though I live this phrase and have a personal story about it).

pomek: @Reinmar, WDYT?

Reinmar: > And this is cool, be we do not really adopt the idea of a good example / bad example in our own docs.

Why? What's wrong with it? As @pomek mentioned, it's absolutely common in ESLint's documentation which, at least I, consider very well thought and helpful.

godai78: We just never did it this way in the docs; I will not oppose using this, tho.

Reinmar: I can see cases where using bad examples may be confusing (i.e. where they are not clearly marked). But if we'll make sure here that bad examples are clearly marked (in fact, @psmyrek caught an issue with that), doing this, even when for the first time, is IMO fine. We do many things for the first time in this project :D

Issue #9563, opened 2021-04-26
Inline images: Pasting content with relative src image is crashing editor


LukaszGudel:
### Steps:
1. Select all content from https://ckeditor.com/ckeditor-5/demo/ and copy it.
2. Paste it to all-features manual test.

### Result:
Editor crashes with error:
```
TypeError: null is not an object (evaluating 'modelImageElement.name')
  fire - emittermixin.js:232
  domListener - emitermixin.js:247
```

Notes:
- Only in Safari
- this is related to relative image src (`<img src="/assets/images/bg/volcano-8967c4575e.jpg" >`) in demo content
- other browsers do not crash with those steps and are showing missing image icon
- same scenario in Safari with editor on `master` branch or in our demos/docs is not crashing

pomek:
The minimum code that reproduces the issue on any browser:

```js
editor.setData( '<img alt="Three Monks walking on ancient temple.">' );
```

Note: There is no `[src]` attribute on `<img>`.

Ok, you can ask, why do I create the image element without specifying the source? That's why Safari does when pasting the image with the non-existing resource. 

Safari, `console.log` at the beginning of Clipboard pipeline (`viewDocument#clipboardInput`) when reproducing the steps specified above. It's enough to copy the image only.
![image](https://user-images.githubusercontent.com/2270764/116665774-74871680-a99a-11eb-8cc4-23eb52c57df7.png)

I'd say we have two solutions:

- do not paste anything in such a case, just remove the element,
- try to extract the content of the `figcaption` element if specified and insert it as a new paragraph.

niegowski:
Closed in https://github.com/ckeditor/ckeditor5/pull/9600.

PR #10449, opened 2021-08-30
Upcast images without or with empty `src` attribute

### LINKED ISSUE #9238: Rethink image upload flow and image model & data while it is uploading

We have a problem with how uploading image is represented and handled in model and how it is handled in conversion. The problem is that an image which is uploading does not have `src` attribute. If such image is then downcasted, there is an image with no `src` attribute in the HTML. Then, when that's upcasted, the image is gone. It is not converted, because there is no `src` attribute.

This causes problems in some features that rely heavily on the editor state. One example is data compression, which may fail, because the model before and the model after compression may differ if the image was uploading when the compression got triggered.

I have two ideas:

1. Simpler. If the image is uploading, it should be downcasted in a way that upcasting it is possible, it would still be an image but displaying some kind of "the image is broken because upload was cancelled" message.
2. Complex. The image should be downcasted in such a way, that the upload would be continued after the image is upcasted. There are two problems here:
    * the image would have to be saved ad base64 in the data content (unless we save it in local storage but this has obvious limitations when it comes to re-uploading),
    * do not send images coded in base64 to other clients in RTC as the operation would be huge.

Not sure if the second idea is viable.

After we have some idea how we want to handle uploading and upload-resuming (if we do) we should consider this also for other asynchronously loaded features (not sure if auto-embed is any problematic in this regard, maybe some future features?). It would be good to have one system for all such cases.

Also related: https://github.com/ckeditor/ckeditor5/issues/5159

dawidurbanski:

Other (image): Upcast images without or with empty `src` attribute. Closes #9238.

Other (link): Upcast linked images without or with empty `src` attribute. See #9238.

Other (media): The `figure` element should not be consumed if the media embed is unknown.

---

### Additional information

Should also close: cksource/collaboration-features#3859 and #2829.

Worth to re-check this one as well: #9563 as it essentially overrides all changes made by the corresponding PR.


dawidurbanski:
@niegowski FYI, when you will be reviewing this, after merging `master` branch into this one, it turned out there is a small problem with recently added changes `media-embed` package.

It was causing these two manual test to raise console errors:

```
 CONSOLE ERROR   ➜  1 error
  
  ❌ TypeError: Cannot read property 'getAncestors' of null
    
    …found on the following pages:
    ➥  http://localhost:8125/ckeditor5-media-embed/tests/manual/mediaembed.html
    ➥  http://localhost:8125/ckeditor5-media-embed/tests/manual/semanticmediaembed.html
```

After some debugging and with help of @maxbarnas I found that for some cases (exactly for when you try to embed item with URL that is not matched with any URL from the providers list), [in this line](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-media-embed/src/mediaembedediting.js#L266) the `modelRange` had both start and end of `0` (it was collapsed).

This was later causing some unexpected behaviour down the line.

It has been patched with [this change](https://github.com/ckeditor/ckeditor5/pull/10449/commits/3e919ea9cfaf5e9a6b3c586703cc153c6f2c2d35), and tested in one [additional test](https://github.com/ckeditor/ckeditor5/pull/10449/commits/9fd1c6354ae87f7417a75f12b1c53cbaabbf7d23).

niegowski (approved):

niegowski:
@ckeditor/qa-team could you check if this feature works correctly and it solved the mentioned issues in the additional information?

Mgsy:
## Steps to reproduce

1.  Go to the all features manual test.
2.  Start uploading some big image. I'm using a 15Mb image (https://effigis.com/wp-content/uploads/2015/02/DigitalGlobe_WorldView1_50cm_8bit_BW_DRA_Bangkok_Thailand_2009JAN06_8bits_sub_r_1.jpg).
3.  During the upload call `editor.setData( editor.getData() )`.

## Current result

After a while, the editor crashes and the image disappears.

## Error

```
filerepository.js:522 Uncaught (in promise) TypeError: Cannot read property 'isFulfilled' of undefined
    at FileLoader.abort (filerepository.js:522)
    at Document.<anonymous> (imageuploadediting.js:203)
    at Document.fire (emittermixin.js:199)
    at Document._handleChangeBlock (document.js:309)
    at Model._runPendingChanges (model.js:842)
    at Model.enqueueChange (model.js:245)
    at imageuploadediting.js:381
```

Mgsy:
## Steps to reproduce

1.  Go to the all features manual test.
2.  Remove the whole content.
3.  Start uploading some big image. I'm using a 15Mb image ([https://effigis.com/wp-content/uploads/2015/02/DigitalGlobe_WorldView1_50cm_8bit_BW_DRA_Bangkok_Thailand_2009JAN06_8bits_sub_r_1.jpg](https://effigis.com/wp-content/uploads/2015/02/DigitalGlobe_WorldView1_50cm_8bit_BW_DRA_Bangkok_Thailand_2009JAN06_8bits_sub_r_1.jpg)).
4.  During the upload call `editor.setData( editor.getData() )`.

## Current result

The empty element is inserted, however, it's not possible to type anything or create a new paragraph. The image toolbar is attached to this empty element.

## Screenshot

Mgsy:
Not sure if the above issues are strictly related to changes introduced by this PR :point\_up: .

niegowski:
> Not sure if the above issues are strictly related to changes introduced by this PR ☝️ .

Does it happen on master?

Mgsy:
As I can see, [https://github.com/ckeditor/ckeditor5/pull/10449#issuecomment-914185700](https://github.com/ckeditor/ckeditor5/pull/10449#issuecomment-914185700) could be reproduced on master, but [https://github.com/ckeditor/ckeditor5/pull/10449#issuecomment-914186461](https://github.com/ckeditor/ckeditor5/pull/10449#issuecomment-914186461) occurs only on this branch (before these changes, the image was simply removed, now it persists as an empty element).

niegowski:
@dawidurbanski could you create follow-ups for the above issues?

dawidurbanski:
Done: #10486 #10487

PR #10451, opened 2021-08-30
jsdoc: remove support for string in values for _t

fedemp:
`_t` calls `_translate` with one item from `values`, which is an array of either number or string.
But `translate` expects the `quantity` param to be a number.

### DIFF
--- a/packages/ckeditor5-utils/src/locale.js
+++ b/packages/ckeditor5-utils/src/locale.js
@@ -149,7 +149,7 @@ export default class Locale {
 	 *
 	 * @private
 	 * @param {String|module:utils/translation-service~Message} message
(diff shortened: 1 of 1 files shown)


ma2ciek:
Hi @fedemp,

That's partially true, but only the first value needs to be a number and only if the plural form is used (it will be used to determine the proper form of the sentence). Otherwise, these values can be strings that will be used to just fill the placeholders in the translated string. See the examples for the `t()` method:

https://github.com/ckeditor/ckeditor5/blob/af4bb5110342f23f7530e4ad9d592bc9c71b3dce/packages/ckeditor5-utils/src/locale.js#L78-L120

PR #10453, opened 2021-08-31
Docs: adding jsfiddle snippet to DLL guide.

godai78:

Docs: adding a live jsfiddle snippet to DLL guide.

---

### Additional information

_Closes cksource/ckeditor5-internal#921_


docs/builds/guides/development/dll-builds.md
```diff
@@ -149,6 +149,12 @@ For example:
 		} );
 ```
 
+## Live implementation sample
+
+Presented below is a working sample editor using the DLL mechanism. Observe the source and then click **"Result"** to switch to the live view of the working CKEditor 5 instance.
+
+<iframe width="100%" height="600" src="//jsfiddle.net/ckeditor/mv05bweu/8/embedded/html,result/dark/" allowfullscreen="allowfullscreen" allowpaymentrequest frameborder="0"></iframe>
```
pomek, line 156: When we set the final content of the sample, we should set a current version as the base.

image

Then, the URL will not contain the `/8/` part.

pomek (commented):
I like the idea, but I'm not sure about the sample's content.

Except for showing the editor's features, we could consider producing the content that suggests going back from the sample to the documentation.

FilipTokarski:
My two cents here - when we're ready with the content of the demo, it would be a nice finishing touch to adjust the iframe's height, so that there's no vertical scroll with the default content.

docs/builds/guides/development/dll-builds.md
godai78: Done.

PR #10459, opened 2021-08-31
Utils>Config>_getFromSource return early if not object

fedemp:
Trying to make TS understand the code base better, I suggest this change. If the `source` is not an object, just return early with `undefined`. Currently you are setting it to `null`, and breaking out of the loop. At the end, you ask if `source` is `null` and return `undefined` if so. I propose that you just return as soon as you tried to set `source` to `null`.


Reinmar:
This code indeed doesn't seem to make sense. How does it help TS, though?

Also, the CC dropped so there's a question of whether a test is missing or there's dead code after your change.

PR #10481, opened 2021-09-06
MarkerCollection has() method accept instance of marker

### LINKED ISSUE #9985: MarkerCollection#has should accept Marker instances too

I don't understand why it doesn't take marker instances as a parameters too - it was very natural way for me to call it.

Currently it accepts only marker name, and works like that:

```
model.markers.has( myMarker.name ); // true

model.markers.has( myMarker ); // false 🤔
```

---

EwelinaD:

Other (engine): The `MarkerCollection#has()` method now also accepts an instance of a marker. Closes #9985.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/model/markercollection.js
+++ b/packages/ckeditor5-engine/src/model/markercollection.js
@@ -52,7 +52,7 @@ export default class MarkerCollection {
 	}
 
 	/**
-	 * Checks if given {@link ~Marker marker} or marker with given name is in the collection.
+	 * Checks if given {@link ~Marker marker} or marker name is in the collection.
 	 *
 	 * @param {String|module:engine/model/markercollection~Marker} markerOrName Name of marker or marker instance to check.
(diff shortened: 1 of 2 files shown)


mlewand (approved):
All good, just some minor remarks:
* In the changelog message, we're using a hash character to denote methods, like so: `Comment#setAttribute()`, `RevisionTracker#updateRevision()`, `DomConverter#domToView()` etc.
* It's nice that are looking at surrounding docs and e.g. change the method description properly - as it no longer works on name alone but it accepts also instances 👍
* I pushed a tiny commit that changed the names in [`should return false if given instance of marker is not in the collection`](https://github.com/ckeditor/ckeditor5/blob/f0eafe60da05393125bd933a80b70a3d03a4d5d2/packages/ckeditor5-engine/tests/model/markercollection.js#L144-L148) to make it clear that marker collection used there is different from the asserted one.

PR #10492, opened 2021-09-08
Docs: features output guide formatting.

### LINKED ISSUE #9700: Improve the styling in "Features' HTML output" left column

There are two links for each plugin, but it wasn't clear even for me after a while which links where. Just to make it clear now – the first link leads to the docs, the second to the plugin API docs.

Let's rethink what we need there and how it could be presented.

## 📃 Other details

*   Browser: …
*   OS: …
*   CKEditor version: …

---

godai78:

Docs: formatting changes for the features' output listing. Closes #9700.

---

### Additional information

To do:

*   [x] add feature name at the top
*   [x] change current plugin and feature names to API docs/feature guides links respectively
*   [x] add `alt` tag sporting the name of feature/plugin for SEO
*   [x] add icons (use docs' guide icon for guide + `cog` icon for API)
*   [x] test

Proposed output (+ icons):  
 

output


przemyslaw-zan:
*   Title is `H4` element - please verify that it's correct element to use
*   I've used 📖 instead of `book.svg`, is that ok?

godai78:
The icon looks fine, both b/w are cool. The title does not need to be `Hx`, a bold will do IMHO.

przemyslaw-zan:
Now with SVG. I've kept H4:



Bold would look worse IMO:

godai78:
We need to have the icons aligned. I would opt for black-and-white ones instead of these. Also, I believe the `book` icon is unaligned, so we could use this instead:

przemyslaw-zan:
Inserting new `<h4>` element on top of the table breaks the validator, causing it to throw errors:



Validator likely checks for the first and second element to see if it's the correct linkt - the validator will need adjustment.

pomek:
> Inserting new `<h4>` element on top of the table breaks the validator, causing it to throw errors

Umberto iterates over heading elements when building a Table of Contents. The `[id]` property for heading elements is obligatory.

pomek:
> Validator likely checks for the first and second element to see if it's the correct link - the validator will need adjustment.

I checked the changes proposed on the branch and I found the following HTML in the built docs:

```html
<a href="#undefined">CKFinder Upload Adapter</a>
```

There are many of them. The validator works as expected, the produced content seems to be invalid.

docs/assets/img/exportword.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20"><path d="M13 0l5 4.5V20H2V0h11zm-1.5 1.5h-8v17h13v-12h-5v-5zM15 15v1H5v-1h10zm0-3v1H5v-1h10zm0-3v1H5V9h10zm-5-3v1H5V6h5zm3-4v3h3.3L13 2zm-3 1v1H5V3h5z"/></svg>
```
psmyrek, line 1: The name of this file could reflect its content, i.e. that it represents a book icon.

psmyrek (changes requested):
It does not look as it has been proposed:

Output

- icons are not aligned with the link text,
- both links (feature guide and API docs) should be positioned on the bottom of the cell,
- the `alt` attribute for the API link contains HTML element, i.e. `alt="&lt;code&gt;HorizontalLine&lt;/code&gt;"`,
- the table header should have a light gray background.

Also:
- if the package does not have an API docs link, the script creates an empty `<p>`,
- multiple JSDoc comments should reflect the introduced changes (i.e. the docs for `createFeatureLink()` currently informs that if the feature documentation is missing, just the plugin name is returned, but in fact an empty string is returned).

godai78:
I'm not really sure about aligning the links to the bottom for some long cells.

przemyslaw-zan:
Previously, Feature guide and API docs links were, respectively, plugin name, and plugin class name. After current changes, plugin class name (eg. second link on screenshot below) is not shown anywhere. Do we want to show that somewhere?

przemyslaw-zan:
I believe script changes are complete. Style changes remaining:

*   icons are not aligned with the link text
*   both links (feature guide and API docs) should be positioned on the bottom of the cell
*   the table header should have a light gray background

pomek:
The current look:

image

PR #10500, opened 2021-09-09
Docs: Small adjustment in the plugincollection api docs

### LINKED ISSUE #10478: ckeditor5-core/plugincollection: get() misses `undefined` as return type

The documentation of `PluginCollection.get()` states, that it will always return `PluginInterface`. From implementation, this is not true, as it will return `undefined` if a plugin cannot be found.

An alternative would be throwing an exception, as (referring to the documentation) it is recommended using `has` before `get`, thus any check for `undefined` is irrelevant after using `has`. This is especially relevant to CKEditor-Typings as can be found at DefinitilyTyped.

mateuszzagorski:

Docs: Updated `PluginCollection.get()` as it should return `undefined` if a plugin cannot be found. Closes #10478.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-core/src/plugincollection.js
+++ b/packages/ckeditor5-core/src/plugincollection.js
@@ -108,7 +108,7 @@ export default class PluginCollection {
 	 * to check if a plugin is available.
 	 *
 	 * @param {Function|String} key The plugin constructor or {@link module:core/plugin~PluginInterface.pluginName name}.
-	 * @returns {module:core/plugin~PluginInterface}
+	 * @returns {module:core/plugin~PluginInterface|undefined}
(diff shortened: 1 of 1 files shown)


pomek:
Is this possible that the method returns `undefined`?

https://github.com/ckeditor/ckeditor5/blob/66a28aba0536628ddfb182865d68a7f5110eda35/packages/ckeditor5-core/src/plugincollection.js#L116-L138

IMO, if a plugin couldn't be found, an exception is thrown.

mateuszzagorski:
[The issue was invalid](https://github.com/ckeditor/ckeditor5/issues/10478#issuecomment-916691691).

Issue #10413, opened 2021-08-25
Text transformation: Allow reverting by pressing backspace


Reinmar:
## 📝 Provide a description of the improvement

Currently, we allow reverting an automatic transformation result by pressing undo. It's a separate undo step.

However, it's common among other editors to also allow reverting text transformation by pressing backspace immediately after transformation happened.

It works this way in GDocs and Word:

1.  Type "1/2"
2.  Type space
3.  It gets converted to "½ "
4.  Press backspace
5.  You get: "1/2 "

Note: It's super smart in Word where this 1/2 never gets converted back to a fraction again, even if you remove that space and type it again. However, GDocs doesn't implement this mechanism and you can again trigger transformation again by typing another space. We can do KISS as both options make some sense, plus going in a Word-like direction would require more work.

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

Reinmar:
Note: In GDocs and Word backspace behaves like undo – this is – you can redo that change.

Note: What about autolinking, auto media embed, etc.

arkflpc:
Are we sure about Autoformat yet?

Reinmar:
I'd go without autoformat for now because this will be safer. But, it'd be good to test how this works with autoformat to see whether the UX is good or confusing.

Hence, if you could create a PR for the features that we want to cover and then a second draft PR (based on that former PR) for autoformat?

arkflpc:
PR to test autoformat: #10508

PR #10508, opened 2021-09-10
Undo by pressing backspace: AutoFormat

arkflpc:

Feature (autoformat): Allowed undoing automatic [Markdown-like formatting](https://ckeditor.com/docs/ckeditor5/latest/features/autoformat.html) by pressing <kbd>Backspace</kbd>. See #10413. 

---

### Additional information


oleq (commented):
Some remarks:

* The feature works fine on the technical level. 
* I'm not sure applying this behavior to headings, blockquotes, code blocks, or horizontal lines is very useful, though. It's very unlikely anyone begins their block with `###`, `>`, ` ``` ` or `---` and they really want this sequence to be preserved because this is how they start the sentence. It's more likely they'll create headings/blockquotes/code blocks/etc. and want to press backspace and get rid of the whole block because they changed their mind. 
    * OTOH, it makes sense for numbered lists because I can imagine starting a sentence with `1.` and **not** wanting to create a list. Same with ordered lists, for instance, if they write footnotes they might want their sentence to start with `*` so they can superscript it later.
    * Maybe we could disable it for certain formattings then?
* Come to think of it, I'm not sure how this feature could be useful for basic style formattings either (those that do not start with `^` in their regexps). I mean, it works fine but how is `**abc**` a useful sequence anyone would like to have in their content? Backspacing to remove "c" seems more logical than undoing the autoformatting.  
   * Unlike the case with lists, I see no strong reason against enabling it for basic styles, though. It's a nice gimmick, I guess.
* I couldn't find any other UX issues with this new behavior. I thought composition will break but I failed to find any scenario that would prove it. 
* No other text editor act like this. I'm curious why? 

---

**TL;DR**

I'd enable this feature for ordered, numbered lists, and maybe for basic styles (because it looks cool). Then I'd wait for feedback from the community.

This behavior could be an argument of `blockAutoformatEditing()` and `inlineAutoformatEditing()` so 3rd-party integrators can decide what they want.

Reinmar:
> I'd enable this feature for ordered, numbered lists,

Sounds good. The reasoning (that you often need to revert the autoformat in these cases) works for me.

> and maybe for basic styles (because it looks cool)

You wrote yourself that "I'm not sure how this feature could be useful for basic style formatting". I agree that it looks cool there, though.

My big worry here, if we start to choose arbitrarily in which cases backspace works would be, though, that it will be a constant surprise to the user.

So, I think that viable options are:

*   Make it work in all cases
*   Make it work in block cases only (because of lists)
*   Do nothing.

Reinmar:
> *   Make it work in all cases

Let's go this way. We'll be consistent and have feedback (if anyone is annoyed by that).

Issue #7, opened 2015-01-16
package.json should be indented with tabs


Reinmar:

fredck:
The current formatting is on purpose.

The point is that `npm install whatever --save-dev` reformats package.json. It will be a pain in the *** to have to always reformat the code to our standards.

So, IMHO, for package.json, we should keep the `npm install` standards.

Reinmar:
://////////////////////

Heh... ok.

Reinmar:
Actually, I'd prefer avoiding using `npm install --save-dev`, but I guess that no one will support me ;).

fredck:
> Actually, I'd prefer avoiding using npm install --save-dev, but I guess that no one will support me ;).

-1 straight :) especially because of version numbers but also because we must be ok with such so common practices.

PR #10512, opened 2021-09-10
Simplified ESLint configuration.

przemyslaw-zan:
Internal: Removed list of DLL core packages from ESLint configuration file.


pomek:
The branch name `i/7` suggests that the PR fixes #7, which is not correct. I guess `ccp/7` would match better here.

pomek (approved):
LGTM.

PR #10515, opened 2021-09-11
Added customization of the max number of visible items in the mention list.

### LINKED ISSUE #10479: Mention Plugin -- Being able to customize the max number of items in list after typing trigger character

Currently the Mention plugin limits the number of items in its list to 10.
This parameter is hardcoded. We would like to have a scrollable list with more than 10 items, possibly an infinite number of items and we are unable to edit this.
Thus we would like the number of items in the list to be a parameter that can be customized when using the plugin, in order to display 5, 10, 15 items... Ideally, the parameter would be able to take a null value and display an infinite number of items. The parameter can default to 10 for backwards compatibility.

## 📃 Other details

* Browser: Chrome
* OS: Windows 10 Pro
* CKEditor version: 5
* Installed CKEditor plugins: Mention

---

mateuszzagorski:

Feature (mention): Added configuration option that allows customizing the max number of items in the list after typing trigger character. The parameter defaults to 10 for backwards compatibility. Closes #10479.

---

### Additional information


packages/ckeditor5-mention/src/mention.js
```diff
@@ -131,6 +131,30 @@ export default class Mention extends Plugin {
...
+ *				mention: {
+ *	 				dropdownLimit: Infinity,
+ *					feeds: [
+ *						{ ... }
+ *						...
+ * 					]
+ *				}
+ *			} )
+ *			.then( ... )
+ *			.catch( ... );
```
mlewand, line 149: Give it one line of breathing room between the code listing and docs text. 🙂

packages/ckeditor5-mention/src/mentionui.js
```diff
@@ -161,7 +163,7 @@ export default class MentionUI extends Plugin {
 			}
 
 			const minimumCharacters = mentionDescription.minimumCharacters || 0;
-			const feedCallback = typeof feed == 'function' ? feed.bind( this.editor ) : createFeedCallback( feed );
+			const feedCallback = typeof feed == 'function' ? feed.bind( this.editor ) : createFeedCallback( feed, dropdownLimit );
```
mlewand, line 166: The way you applied the fix will work only for one out of two use cases:

*   ✅ Passing a simple array of mentions like you did in unit tests.
*   ❌ Passing a custom callback which in turn return an array of mentions.

See [https://ckeditor.com/docs/ckeditor5/latest/api/module_mention_mention-MentionFeed.html](https://ckeditor.com/docs/ckeditor5/latest/api/module_mention_mention-MentionFeed.html) to see some examples.

Now I see that you applied fix to the data of the feed - this is incorrect. You want to limit the entries in _view only_. The view handler is common for any methods mentions would be fetch.

I pushed a dirty unit test that will show the issue in the `ck/10479-mention-dropdownlimit-failing-test` branch.

mlewand (changes requested):
The proposed fix doesn't handle all the cases. See the comments for more details.

Good job with conceptual part of unit tests, these are good cases 👍

LukaszGudel:
LGTM :+1:

packages/ckeditor5-mention/src/mention.js
```diff
@@ -131,6 +131,45 @@ export default class Mention extends Plugin {
...
+ *
+ * Customizing the number of visible mentions allows you to specify how many available elements will users be able to see
+ * in the dropdown list. You can specify any number you'll see fit. For example, in the snippets above you will find the
+ * dropdownLimit set to `20` and `Infinity` (this will result in showing all available mentions).
```
Dumluregn, line 167: Please move this description before the code samples - without context it isn't clear what to look at in them.

packages/ckeditor5-mention/src/mentionui.js
```diff
@@ -220,6 +220,13 @@ export default class MentionUI extends Plugin {
 		mentionsView.items.bindTo( this._items ).using( data => {
 			const { item, marker } = data;
 
+			// See: #10479
```
Dumluregn, line 223: Such note without any context doesn't bring much more than checking git blame. The idea usually is to give a short, one or two sentences long explanation and then, if there is more to read on GH, leave a ticket reference. 
Here I think the most important part is that the default `10` is kept for backwards compatibility. Sure, we can still leave the ticket reference after that, but it wouldn't be a tragedy if we didn't as there is not much information left.

packages/ckeditor5-mention/tests/manual/mention-custom-dropdownlimit.md
```diff
@@ -0,0 +1,13 @@
...
+The mention configuration with a custom `config.mention.dropdownLimit` configuration and a static list of autocomplete feed.
+
+### Configuration
+
+Type "@" to display the list of 20 mentions out of 21 available or type "#" to see shorter list.
+
+### Interaction
+
+- Move arrows up/down to select an item.
+- Use <kbd>enter</kbd> or <kbd>tab</kbd> to insert a mention into the documentation.
```
Dumluregn, line 12: I know it says "documentation" in other manual tests but I think it's not worth to propagate a typo, let's change it to the "document".

packages/ckeditor5-mention/tests/mentionui.js
```diff
@@ -2168,6 +2168,82 @@ describe( 'MentionUI', () => {
...
+				model.change( writer => {
+					writer.insertText( '@', doc.selection.getFirstPosition() );
+				} );
+
+				return waitForDebounce()
+					.then( () => {
+						expect( panelView.isVisible ).to.be.true;
+						expect( mentionsView.items ).to.have.length( mentionsLimit );
+					} );
+			} );
```
Dumluregn, line 2209: As @mlewand mentioned [in the comment](https://github.com/ckeditor/ckeditor5/pull/10515#discussion_r707611498), this is a dirty test which can be easily rewritten to comply with all the other tests in the file, so please instead of `await` use promise chain here (like in the next test case).

Dumluregn (changes requested):
Besides a few minor issues I've reported, there are two more things:
1) We should update the docs feature guide:
-  I think the [Customizing the autocomplete list](https://ckeditor.com/docs/ckeditor5/latest/features/mentions.html#customizing-the-autocomplete-list) section could be subdivided into `styling` part (the existing one) and `length` (which we'll add). 
- There is also a [fully customized sample](https://ckeditor.com/docs/ckeditor5/latest/features/mentions.html#fully-customized-mention-feed) where we could also implement custom feed length (since there are only 5 items, I think the length of 4 makes sense).

2) Please merge the latest `master` branch.

packages/ckeditor5-mention/tests/manual/mention-custom-dropdownlimit.md
mateuszzagorski: I've removed the typo in all manual tests: f2826ae.

packages/ckeditor5-mention/src/mention.js
```diff
@@ -131,6 +131,45 @@ export default class Mention extends Plugin {
  * @default [ 13, 9 ] // [ Enter, Tab ]
  */
 
+/**
+ * The configuration of the custom number of visible mentions.
+ *
+ * Customizing the number of visible mentions allows you to specify how many available elements will users be able to see
+ * in the dropdown list. You can specify any number you'll see fit. For example, in the snippets above you will find the
```
Dumluregn, line 138:
```suggestion
 * in the dropdown list. You can specify any number you'll see fit. For example, in the snippets below you will find the
```

packages/ckeditor5-mention/tests/manual/mention-custom-dropdownlimit.html
```diff
@@ -0,0 +1,9 @@
+<div id="editor">
+	<p>Hello <span class="mention" data-mention="@Ted">@Ted</span>.</p>
+	<p>Hello <span class="mention" data-mention="@Ted">#Ted</span></p>
```
Dumluregn, line 3: I think we should use the actual mentions available in this test (so `@01`, `@02` etc. for `@` and `#a01`, `#a02` etc. for `#`).

Dumluregn (changes requested):
Only minor corrections left and we're good.

Dumluregn (approved):

packages/ckeditor5-mention/docs/features/mentions.md
```diff
@@ -181,6 +183,30 @@ function customItemRenderer( item ) {
 
 A full, working demo with all possible customizations and its source code is available {@link features/mentions#fully-customized-mention-feed at the end of this section}.
 
+#### Length
```
godai78, line 186:
```suggestion
#### List length
```

packages/ckeditor5-mention/src/mention.js
```diff
@@ -131,6 +131,45 @@ export default class Mention extends Plugin {
  * @default [ 13, 9 ] // [ Enter, Tab ]
  */
 
+/**
+ * The configuration of the custom number of visible mentions.
+ *
+ * Customizing the number of visible mentions allows you to specify how many available elements will users be able to see
```
godai78, line 137:
```suggestion
 * Customizing the number of visible mentions allows you to specify how many available elements will the users be able to see
```

packages/ckeditor5-mention/src/mention.js
```diff
@@ -131,6 +131,45 @@ export default class Mention extends Plugin {
...
+ * in the dropdown list. You can specify any number you'll see fit. For example, in the snippets below you will find the
```
godai78, line 138:
```suggestion
 * in the dropdown list. You can specify any number you see fit. For example, in the snippets below you will find the
```

godai78 (approved):

PR #10536, opened 2021-09-15
Table rows and columns reordering when `headingRows` or `headingColums` is set

### LINKED ISSUE #10463: Reordering header rows or columns with non-headers breaks view

There are two cases with different behaviour, but in both the model and data downcast work fine and editing downcast is bugged.

## Scenario 1 - moving the header row

1. Create the table with one header row.
2. Move the first row down using `writer`.
3. Examine the model and both downcasts.

### ✔️ Expected result

There should still be one header row and rows content should be swapped.

### ❌ Actual result

There is no more header rows in the editing downcast. Also the content of rows didn't swap in the view.

## Scenario 2 - moving the non-header row

1. Create the table with one header row.
2. Move the second row up using `writer`.
3. Examine the model and both downcasts.

### ✔️ Expected result

There should still be one header row and rows content should be swapped.

### ❌ Actual result

There are two header rows in the view. Still only one in model and data downcast.

### ❓ Possible solution

The only broken pipeline is the editing downcast, so I'd look for the solution there.

## 📃 Other details

Same issues occur during header column reordering. On the [`header-row-column-test` branch](https://github.com/ckeditor/ckeditor5/tree/header-row-column-test) I've pushed some unit tests reproducing the problem.

* Browser: any
* OS: any
* Installed CKEditor plugins: table

---

dawidurbanski:

Fix (table): Make reordering table rows and columns possible without breaking view in tables with `headingRows` or `headingColumns` is set. Closes #10463.


packages/ckeditor5-table/src/converters/table-heading-rows-refresh-post-fixer.js
```diff
@@ -29,14 +29,31 @@ function tableHeadingRowsRefreshPostFixer( model ) {
...
-			continue;
-		}
+		if ( change.type === 'attribute' ) {
+			const element = change.range.start.nodeAfter;
+
+			if ( element && element.is( 'element', 'table' ) && change.attributeKey === 'headingRows' ) {
+				tablesToRefresh.add( element );
+			}
+		} else {
+			/* istanbul ignore else */
```
dawidurbanski, line 39: Alternatively instead of nested if with ignored else we can completely drop below check:

```
if ( change.type === 'insert' || change.type === 'remove' ) {
```

niegowski (approved):

Issue #4594, opened 2019-10-09
Selection converter should mark the top-most widget only


Reinmar:
![image](https://user-images.githubusercontent.com/156149/48080052-1758e280-e1ed-11e8-9f3e-de5e7d4927f9.png)

Right now the selection converter (https://github.com/ckeditor/ckeditor5-widget/blob/49afa6e3216334b18573ae0be42d6abb2af324cb/src/widget.js#L67-L81) marks all widget elements in the selection, not just the top-most one.

Issue #57, opened 2016-01-07
Don't watch on changes in packages which weren't installed in the dev env


Reinmar:
If `node_modules/ckeditor5-foo/` is a symlink, then watch.

If not, then we may assume that it does not contain a development code, but a package installed directly from NPM so we don't have to watch.

However, in some cases developers may want to have a single `ckeditor5/` repo with some packages included directly inside this repo (just like Babel does). So there should be an option to enable watching also on not-symlinked packages

Reinmar:
I think that I reported this issue too early. We don't know about any performance issues which the current watcher could cause (we had some issues in the past in Bender.js) and performance is the only thing that comes to my mind as a reason.

PR #10556, opened 2021-09-19
Widget - hide selection handle for nested widget

### LINKED ISSUE #9491: The nested widget selection handle is visible while the outer table cells are selected

1.  https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/table.html
2.  make nested table
3.  select multiple cells in the outer table

### ✔️ Expected result

The selection handle is not visible

## 📃 Other details

*   Browser: …
*   OS: …

---

EwelinaD:

Fix (widget): Selection handle no longer visible on selected nested widgets. Closes #9491.

---

### Additional information

Fix is based on assumption that there are multiple ranges during data selection in table.

### DIFF
--- a/packages/ckeditor5-widget/src/widget.js
+++ b/packages/ckeditor5-widget/src/widget.js
@@ -124,18 +124,34 @@ export default class Widget extends Plugin {
 			const viewWriter = conversionApi.writer;
 			const viewSelection = viewWriter.document.selection;
 
+			let lastMarked = null;
+
+			let rangeCounter = 0;
+			const widgetsToSelect = [];
+
 			for ( const range of viewSelection.getRanges() ) {
+				// if there is more than 1 range it means that we are inside table and we don't want to select nested widgets.
+				// See #9491.
+				if ( rangeCounter > 0 ) {
+					return;
+				}
+				rangeCounter++;
 				// Note: There could be multiple selected widgets in a range but no fake selection.
 				// All of them must be marked as selected, for instance [<widget></widget><widget></widget>]
 				for ( const value of range ) {
 					const node = value.item;
+					// Do not mark nested widgets in selected one. See: #4594
+					if ( isWidget( node ) && !isChild( node, lastMarked ) ) {
+						widgetsToSelect.push( node );
 
-					// Do not mark nested widgets in selected one. See: #9491.
-					if ( isWidget( node ) && !hasWidgetParent( node ) ) {
-						viewWriter.addClass( WIDGET_SELECTED_CLASS_NAME, node );
 						this._previouslySelected.add( node );
+						lastMarked = node;
 					}
 				}
+			}
+
+			for ( const node of widgetsToSelect ) {
+				viewWriter.addClass( WIDGET_SELECTED_CLASS_NAME, node );
 			}
 		}, { priority: 'low' } );
 
@@ -463,18 +479,15 @@ function isInsideNestedEditable( element ) {
 	return false;
 }
 
-/**
- * Checks if {@link module:engine/model/element~Element element} has any widget as parent
- *
- * @param {module:engine/view/element~Element} element - element to check.
- * @returns {Boolean}
- */
-function hasWidgetParent( element ) {
-	for ( const ancestor of element.getAncestors() ) {
-		if ( isWidget( ancestor ) ) {
-			return true;
-		}
+// Checks whether the specified `element` is a child of the `parent` element.
+//
+// @param {module:engine/view/element~Element} element An element to check.
+// @param {module:engine/view/element~Element|null} parent A parent for the element.
+// @returns {Boolean}
+function isChild( element, parent ) {
+	if ( !parent ) {
+		return false;
 	}
 
-	return false;
+	return Array.from( element.getAncestors() ).includes( parent );
 }
--- a/packages/ckeditor5-widget/src/widget.js
+++ b/packages/ckeditor5-widget/src/widget.js
@@ -124,20 +124,16 @@ export default class Widget extends Plugin {
 			const viewWriter = conversionApi.writer;
 			const viewSelection = viewWriter.document.selection;
 
-			let lastMarked = null;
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-widget/src/widget.js
```diff
@@ -126,21 +126,33 @@ export default class Widget extends Plugin {
 
 			let lastMarked = null;
 
+			let rangeCounter = 0;
+			const widgetsToSelect = [];
+
 			for ( const range of viewSelection.getRanges() ) {
+				// if there is more than 1 range it means that we are inside table and we don't want to select nested widgets.
```
dawidurbanski, line 133: If really the only correct behaviour is when we do have only one range, and we ignore other ones, then we do not need to loop through all of them at all. We can just pick the first range from `viewSelection.getRanges()`.

In such case there is no need for `rangeCounter`, widgets accumulation and later for the additional loop.

dawidurbanski (changes requested):
I made one comment/suggestion to consider.

Also I'm pretty sure we need to add a test to cover scenario with multiple ranges.

packages/ckeditor5-widget/src/widget.js
dawidurbanski: Also, I would think about it twice. Is there really no other case where there are multiple ranges but we are not inside table? In other words is there a situation where multiple ranges are expected and we need to handle all of them?

oleq (changes requested):
Let's circle back a little bit. There's a decision concerning UX we must make before we jump to writing code. I'm moving the discussion to the original issue: https://github.com/ckeditor/ckeditor5/issues/9491.

packages/ckeditor5-widget/src/widget.js
EwelinaD: This change is reverted.

EwelinaD:
I've just pushed suggested changes to tableselection.css.

oleq:
@ckeditor/qa-team Can you give us a hand, please? :pray: You can find inspiration in comments #9491 🙃

packages/ckeditor5-theme-lark/theme/ckeditor5-table/tableselection.css
```diff
@@ -32,8 +32,16 @@
 			background-color: transparent;
 		}
 
-		& .ck-widget_selected {
+		/*
+		 * To reduce the amount of noise, all widgets in the table selection have no outline and no selection handle.
+		 *See https://github.com/ckeditor/ckeditor5/issues/9491.
```
oleq, line 37:
```suggestion
		 * See https://github.com/ckeditor/ckeditor5/issues/9491.
```

FilipTokarski:
When blurring the multi cell selection, nested widget gets a flashing black border (also the table handle is visible for a split second):


https://user-images.githubusercontent.com/34380544/134646207-f046a804-ca0f-40ca-b34a-c34bf0839a81.mp4

FilipTokarski:
Not sure if important.
I found this behaviour on master - https://github.com/ckeditor/ckeditor5/issues/10600. I wouldn't mind about this, but in the context of this PR - apart from flashing border, the widget handler shows up for a split second as well, but I guess it shouldn't be here:


https://user-images.githubusercontent.com/34380544/134654874-7cc9938c-f1fc-4637-9eca-ee068649fa84.mp4

oleq:
> Not sure if important.  
> I found this behaviour on master - #10600. I wouldn't mind about this, but in the context of this PR - apart from flashing border, the widget handler shows up for a split second as well, but I guess it shouldn't be here:

Let's address it in a separate issue.

oleq:
@dawidurbanski I don't feel like reviewing a solution I suggested myself. Can you do the review?

BTW: What's going on with the CI?

EwelinaD:
Hmm, build details:
`The following packages did not provide required code coverage: engine`
But there were no changes in the engine package

FilipTokarski:
Apart from the comment above, I didn't find anything else.

oleq:
> `The following packages did not provide required code coverage: engine`

There's something wrong with the CI then. Don't worry.

pomek:
> WARN: 'Detected `console.warn()`:', 'Selection change observer detected an infinite rendering loop.'

It decreases the CC. Restarting a build usually helps.

packages/ckeditor5-widget/src/widget.js
```diff
@@ -131,11 +131,9 @@ export default class Widget extends Plugin {
 				// All of them must be marked as selected, for instance [<widget></widget><widget></widget>]
 				for ( const value of range ) {
 					const node = value.item;
-
-					// Do not mark nested widgets in selected one. See: #57.
+					// Do not mark nested widgets in selected one. See: #4594
```
dawidurbanski, line 134: Please revert the original comment. It's confusing right now that it points to another ticket which is unrelated imho for this part of the code.

I believe this file should remain unchanged.

dawidurbanski (changes requested):

dawidurbanski:
> @dawidurbanski I don't feel like reviewing a solution I suggested myself. Can you do the review please?

@oleq done

@EwelinaD I requested one last change. Thanks!

packages/ckeditor5-widget/src/widget.js
LukaszGudel: @dawidurbanski it seems like the original comment was referring to the issue from some package and it was before merging our repos and issues into monorepo.

If you open #57 you will get some `resolution:invalid` issue, not related to this part of the code, while #4594 is indeed related to showing nested widgets. Also if you go to https://github.com/ckeditor/ckeditor5-widget/issues/57 you will be redirected to #4594.

dawidurbanski: Yeah, just had a chat with Ewelina. Seems like it's all good.

dawidurbanski (approved):
LGTM 👍🏻

PR #10561, opened 2021-09-20
Dos: change master to stable for source files.

### LINKED ISSUE #9701: The "source file" in "Features' HTML output" should probably link to the stable branch

We don't have these files on the stable branch yet, but I think that's where we should point people from documentation most of the time.

So, once we ship the next release, let's update these links.

---

godai78:

Dos: change `master` to `stable` for source files. Closes #9701.

---

### Additional information

May require additional logic to exclude linking to the source for `internal` and 3rd party packages. One reviewer will do for a decision.


godai78:
Moved to: #10900

PR #10570, opened 2021-09-21
Disable edit and save buttons in html embed if the editor is in read-only mode or updateHtmlEmbed command is disabled

### LINKED ISSUE #10182: HtmlEmbed UI doesn't reflect isEnabled/isReadOnly state

1. insert a HtmlEmbed block in the editor, type something and save
2. switch the editor to isReadOnly mode
3. It is now impossible to add new HtmlEmbed blocks to the content via the toolbar
4. BUT the "pencil" button to edit existing blocks in the content still works and allows you to modify the html code inside the block.
5. This is a UI bug rather that a functionnality bug since the "save" button on the htmlembed block is disabled, so it is indeed impossible to actually modify the content of the editor. 

### ✔️ Expected result

The "pencil" button of the existing htmlembed blocks in the content should be in disabled state. 

### ❌ Actual result

The "pencil" button of the existing htmlembed blocks is enabled and allows one to change the html code in the block (but not to save it)

dawidurbanski:

Fix (html-embed): Disable edit and save buttons in html embed element if the editor is in read-only mode or `updateHtmlEmbed` command is disabled. Closes #10182.

---

### Additional information

I played around with it and I found some interesting facts.

According to this [comment](https://github.com/ckeditor/ckeditor5/issues/10182#issuecomment-915018495) it was supposed to be enough to maybe add similar binding to `isEnabled` to the `edit` as in the `save` button case.

https://github.com/ckeditor/ckeditor5/blob/1fc24358c2bc70f0dbc674a49213bfbf955f67b9/packages/ckeditor5-html-embed/src/htmlembedediting.js#L376-L389

Well, indeed it was necessary, but not for the reason of blocking the button.

In case of buttons in html embed widget line

```js
buttonView.bind( 'isEnabled' ).to( command, 'isEnabled' ); 
```

will only make some button attributes react to the `updateHtmlEmbed` command `isEnabled`.

It won't stop any buttons from being clickable. You can try this easily by doing this:

1. Go to https://ckeditor.com/docs/ckeditor5/latest/features/html-embed.html
2. Try clicking "Edit" and "Save" buttons in demo editor. They are enabled and clickable.
3. Execute `editor.isReadOnly = true` in console.
4. Try clicking "Edit" and "Save" buttons in demo editor. They are still clickable! (but save button received `ck-disabled` class and `aria-disabled` attribute but not `disabled` attribute).

Not getting `disabled` attribute is done on purpose to keep buttons focusable and it's described in `buttonview.js`.

https://github.com/ckeditor/ckeditor5/blob/1fc24358c2bc70f0dbc674a49213bfbf955f67b9/packages/ckeditor5-ui/src/button/buttonview.js#L158-L168

The thing is, buttons in HTML embed feature are overriding default button on click handler!

That's why save button is still clickable even if we bound its `isEnabled` state to the `updateHtmlEmbed` command `isEnabled`.

https://github.com/ckeditor/ckeditor5/blob/1fc24358c2bc70f0dbc674a49213bfbf955f67b9/packages/ckeditor5-html-embed/src/htmlembedediting.js#L287-L295

https://github.com/ckeditor/ckeditor5/blob/1fc24358c2bc70f0dbc674a49213bfbf955f67b9/packages/ckeditor5-html-embed/src/htmlembedediting.js#L302-L305

So the solution here was to:

1. Bind edit button `isEnabled` similar to save button in order to apply some attributes.
2. Prevent `rawHtmlApi.makeEditable()` (edit button) and `rawHtmlApi.save( newValue );` (save button) if command or editor is disabled.

### DIFF
--- a/packages/ckeditor5-html-embed/src/htmlembedediting.js
+++ b/packages/ckeditor5-html-embed/src/htmlembedediting.js
@@ -126,6 +126,8 @@ export default class HtmlEmbedEditing extends Plugin {
 			view: ( modelElement, { writer } ) => {
 				let domContentWrapper, state, props;
 
+				const command = editor.commands.get( 'updateHtmlEmbed' );
+
 				const viewContainer = writer.createContainerElement( 'div', {
 					class: 'raw-html-embed',
 					'data-html-embed-label': t( 'HTML snippet' ),
@@ -207,10 +209,14 @@ export default class HtmlEmbedEditing extends Plugin {
 					textareaPlaceholder: t( 'Paste raw HTML here...' ),
 
 					onEditClick() {
-						rawHtmlApi.makeEditable();
+						if ( command.isEnabled ) {
+							rawHtmlApi.makeEditable();
+						}
 					},
 					onSaveClick( newValue ) {
-						rawHtmlApi.save( newValue );
+						if ( command.isEnabled ) {
+							rawHtmlApi.save( newValue );
+						}
 					},
 					onCancelClick() {
 						rawHtmlApi.cancel();
@@ -286,12 +292,12 @@ export default class HtmlEmbedEditing extends Plugin {
 
 				clonedDomSaveButton.addEventListener( 'click', evt => {
 					evt.preventDefault();
-					props.onSaveClick( );
+					props.onSaveClick();
 				} );
 
 				clonedDomCancelButton.addEventListener( 'click', evt => {
 					evt.preventDefault();
-					props.onCancelClick( );
+					props.onCancelClick();
 				} );
 
 				domButtonsWrapper.appendChild( clonedDomSaveButton );
@@ -379,6 +385,7 @@ function createDomButton( editor, type ) {
 			label: t( 'Edit source' ),
 			class: 'raw-html-embed__edit-button'
 		} );
+		buttonView.bind( 'isEnabled' ).to( command, 'isEnabled' );
 	} else if ( type === 'save' ) {
 		buttonView.set( {
 			icon: icons.check,
--- a/packages/ckeditor5-html-embed/tests/htmlembedediting.js
+++ b/packages/ckeditor5-html-embed/tests/htmlembedediting.js
@@ -825,6 +825,38 @@ describe( 'HtmlEmbedEditing', () => {
 				}
 			} );
 		} );
+
+		describe( 'with updateHtmlEmbed command disabled', () => {
+			it( 'does not allow editing the source after clicking the "edit" button', () => {
+				setModelData( model, '<rawHtml value="foo"></rawHtml>' );
+				const widget = viewDocument.getRoot().getChild( 0 );
+				const contentWrapper = widget.getChild( 1 );
+				const domContentWrapper = editor.editing.view.domConverter.mapViewToDom( contentWrapper );
+
+				sinon.stub( editor.commands.get( 'updateHtmlEmbed' ), 'isEnabled' ).value( false );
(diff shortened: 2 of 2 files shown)


Reinmar:
We agreed on a f2f call that this PR resolves this issue only partially because, since it does not use the actual view instances but rather a cloned DOM structure of those, the buttons do not get updated. You can see this issue manifested by some buttons not getting the proper classes and hence not getting greyed out:



If they all were live instances of button views, then they'd be fully updated.

Yet, blocking the switch to the edit mode is still gain, so let's merge this and forget about this issue.

Reinmar:
I take the above back.

There's actually an ugly regression introduced in this PR.

1.  Open the manual test. Check the color of edit buttons. All grey (on this PR).
2.  Edit one of the embeds.
3.  Check the colors again. That one is black, all other still grey.

This happens only on this branch. When this button is initially rendered (cloned) it apparently has the isEnabled set to false and renders with that class.

Reinmar:
What I'd propose: If there's no trivial fix (some change of order of things) to this regression, I'd skip this PR for now and if we'll have the time in the sprint, get back to this issue with a proper fix (using live instances of these button views).

dawidurbanski:
Well, to make this truly work as we want we need to put some more work to this.

The current regression happens because previously we didn't bind button `isEnabled` to command `isEnabled` for the **edit** button. Thanks to this, it was enabled by default at the start.

```js
buttonView.bind( 'isEnabled' ).to( command, 'isEnabled' ); 
```

And just in case you wonder why it's disabled by default it's because of using current selection to determine if this command should be enabled:

https://github.com/ckeditor/ckeditor5/blob/e0ff106ede45618f6d1e3ad4922290e967eca3fc/packages/ckeditor5-html-embed/src/updatehtmlembedcommand.js#L29-L32

### TL; DR;

We could just remove binding `isEnabled` for **edit** button and it should be good enough.

It's still better than currently. Just for comparison I made these screencasts:

### Pure master branch:

Buttons always enabled, even in read-only mode. You cannot save content in read-only mode.

https://user-images.githubusercontent.com/12466568/135588600-58481dbd-4b45-4fe8-a110-435e9644084f.mov

### Pure master branch with **edit** button bound to command `isEnabled`:

Visually similar to the current PR. Buttons are disabled* by default (* but only visually). You can still click on them and navigate between edit/preview.

https://user-images.githubusercontent.com/12466568/135589354-bd3d1a6e-dfbe-4dd0-89ef-e26b9831f9bc.mov

### This PR without binding `isEnabled` to the **edit** button:

Buttons are properly enabled/disabled. Edit buttons are enabled by default. Edit button can be disabled by adding `isReadOnly` while in edit mode, and switching back to preview by clicking "Cancel" button, which is not disabled because you should always be able to cancel.

https://user-images.githubusercontent.com/12466568/135589877-124ffb9d-7aea-4841-941e-62a2123cb6e4.mov

IMHO it's good enough. @Reinmar let me know if I should push the fix and let's make this happen.

oleq:
[I created another PoC](https://github.com/ckeditor/ckeditor5/compare/ck/10182b?expand=1) which uses true `ButtonView` instances to render UI buttons to solve all issues 👆. The solution is short and I think it works fine:



There's one downside to it, though: I had to get rid of the `isEnabled` state of the `UpdateHtmlEmbedCommand` so the save and edit buttons can bind to it and work predictably when the command was force-disabled or the editor went into read-only. 

**Previously** the command was disabled when the selection was not on the html embed widget, which made that kind of binding act like this:



The command feels internal so I'm not sure if this will be an issue. In any case, let's talk it over.

dawidurbanski:
@oleq Looks great to me!

PR #10586, opened 2021-09-23
Changed downcast dispatcher event flow

### LINKED ISSUE #10376: Change the order/flow of events fired by the DowncastDispatcher

Currently `DowncastDispatcher` is firing events for the inserted range by walking on the range with `TreeWalker` so for the provided structure:

```html
<x foo="bar">
	<y bar="123" baz="abc">
		foobar
	</y>
</x>
```

It fires events like a flat sequence:

```
insert:x
attribute:foo:x
insert:y
attribute:bar:y
attribute:baz:y
insert:$text
```

but if any of the elements' converter uses slots (and with this triggers conversion of nested element) the flow becomes mixed:

```
insert:x
	insert:y
	attribute:bar:y
	attribute:baz:y
	insert:$text
attribute:foo:x
```

The `x` element is triggering its children's conversion by using slots. This makes the order of events different, the flow becomes partly recursive (for the `insert:x`). This could be a complication hard to understand for developers. Also with a flat sequence of events, there is no way to listen on the whole element with children got converted (that could be useful for side effects that could listen on a low priority to notice when that conversion is completed).

In the `UpcastDispatcher` the flow is different, it's always recursive - there is the lowest priority listener that triggers nested nodes conversion (and also many converters can trigger it if appropriate).

I propose to align the `DowncastDispatcher` to the same recursive flow as it's used in the `UpcastDispatcher` so the sequence of events would look like this:

```
insert:x
	insert:y
		insert:$text
	attribute:bar:y
	attribute:baz:y
attribute:foo:x
```

or maybe even to:

```
insert:x
	attribute:foo:x
	insert:y
		attribute:bar:y
		attribute:baz:y
		insert:$text
```

by introducing nested trigger for attributes conversion (sth similar to `conversionApi.convertInsert()`). This way the order of events will be exactly the same but conversion of attributes and child nodes would be nested (on the call stack) inside the parent `insert` event handler.

niegowski:

Other (engine): The attribute and child nodes conversion events are fired by the lowest priority handler for the `insert` event instead of by `DowncastDispatcher` itself. Closes #10376.

Other (engine): Events are fired by the `DowncastDispatcher` even if they were previously consumed. It's the conversion handler's responsibility to check if it can be consumed or if it was already consumed by other converters.

Feature (engine): Introducing `convertItem`, `convertChildren`, `convertAttributes` in the downcast conversion API interface.

Tests (html-support): Fixed test for checking if an attribute was consumed while converting.

Other (list): Downcast conversion should consume downcasted attributes.

---

### Additional information

In this PR `DowncastDispatcher` is using `conversionApi.consumable._getSymbolForTextProxy( data.item )` to reuse it's `TextProxy` -> `Symbol` mapping to keep track of which conversion events were already fired and which were not. I'm not sure if it's acceptable to do it this way (the method itself was changed to the protected visibility scope).


packages/ckeditor5-engine/src/conversion/downcasthelpers.js
```diff
@@ -734,7 +734,7 @@ export default class DowncastHelpers extends ConversionHelpers {
  */
 export function insertText() {
 	return ( evt, data, conversionApi ) => {
-		if ( !conversionApi.consumable.consume( data.item, 'insert' ) ) {
+		if ( !conversionApi.consumable.consume( data.item, evt.name ) ) {
```
Reinmar, line 737: Kuba, the messiah :D

PR #10623, opened 2021-09-30
Mention panel disappear in read-only mode

### LINKED ISSUE #4645: Mention panel should disappear when editor becomes read–only
ATM it stays:

Kapture 2019-04-04 at 15 44 15

mateuszzagorski:

Fix (mention): Mention panel will now hide when the editor becomes read-only. Closes #4645.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-mention/src/mentionui.js
+++ b/packages/ckeditor5-mention/src/mentionui.js
@@ -170,6 +170,7 @@ export default class MentionUI extends Plugin {
 			this._mentionsConfigurations.set( marker, definition );
 		}
 
+		editor.on( 'change:isReadOnly', () => this._hideUIAndRemoveMarker() );
 		this.on( 'requestFeed:response', ( evt, data ) => this._handleFeedResponse( data ) );
 		this.on( 'requestFeed:error', () => this._hideUIAndRemoveMarker() );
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-mention/src/mentionui.js
```diff
@@ -170,6 +170,7 @@ export default class MentionUI extends Plugin {
 			this._mentionsConfigurations.set( marker, definition );
 		}
 
+		editor.on( 'change:isReadOnly', () => this._hideUIAndRemoveMarker() );
```
Dumluregn, line 173: It may look unimportant, but in other places in code where we listen for the `change:isReadOnly` event we never create listener on the `editor`, but on the particular plugin, like [here](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-find-and-replace/src/findcommand.js#L40):

```js
this.listenTo( editor, 'change:isReadOnly', () => {
...
```

This implementation ensures that listener is bound to the plugin instead of the editor itself, so if e.g. plugin becomes turned off (maybe at some point we'll have dynamic plugin management) the listener will be dismissed. Such an approach also allows to have a bit cleaner listener lists (for both editor and plugin) and understand what depends on what.

Dumluregn (changes requested):
The fix works fine, I've only asked for a small implementation change. Also, please merge the latest `master` branch.

Dumluregn (approved):

PR #10688, opened 2021-10-13
[Docs] Implemented a quick installation quiz in the "Installation" landing page

oleq:

Docs: Implemented a quick installation quiz in the "Installation" landing page.

---

### Additional information

TODOs:

*   [x] The rest of the quiz flow (currently 6 questions and 6 resolutions - predefined, OB, from source, DLLs, Frameworks, migrate from CKE4),
*   [ ] Proofreading


godai78:
We also need to include Online Builder in the docs. And I believe we at least need a guide that mentions the OB (I even have had written an OB step-by-step quick start last year, but it does not need to be that long). Right now it's basically mentioned in "download options" which is not exactly the best way to showcase it.

oleq:
> Right now it's basically mentioned in "download options" which is not exactly the best way to showcase it.

The decision panel could link directly to https://ckeditor.com/ckeditor-5/online-builder/.

godai78:
I allowed myself to align this with the current work.

godai78:
Webpack drama unleashed.

PR #10736, opened 2021-10-21
Avoid UI clash with native text selection handles in Safari/iOS

### LINKED ISSUE #7707: It's almost impossible to click certain buttons in the balloon toolbar (Safari, iOS)

1.  Open the balloon/[balloon block](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/balloon-block-editor.html) example.
2.  Select some text.
3.  Try clicking buttons in the toolbar that horizontally align with the visual selection focus handler, like Bold in the example above (it could be a different button depending on the geometry):

### ✔️ Expected result

All toolbar buttons should be clickable.

### ❌ Actual result

For some unknown reason buttons next to the selection handler cannot be clicked as if something covered them (native iOS behavior/optimization next to selection handles?).

## 📃 Other details

*   Browser: Safari
*   OS: iOS
*   CKEditor version: master

---

oleq:

Fix (ui): Changed the look and position of the `BalloonToolbar` in Safari on iOS to avoid a clash with native text selection handles. Closes #7707.

Feature (utils): Introduced `env.isiOS` for detection of user agents running in iOS (see #7707).

Internal (utils,minimap): Moved the `RectDrawer` dev helper from `ckeditor5-minimap` to `ckeditor5-utils`.


oleq:
@ckeditor/qa-team Can you take a look at this PR, please? (requires native iOS: a simulator or a physical device) :pray:

Mgsy:
I've tested it on iPhone 8 and still had problems with accessing the button. I've marked the area that basically does nothing after clicking on it:



In the below case, I was not able to open the grouped toolbar:

oleq:
@Mgsy Are you sure you tested the right branch? The toolbar does not look like it was shifted vertically. This is how it looks on iPhone 11

Mgsy:
Not sure why I can't see the new behaviour in the docs built with the branch from the PR. I've modified `all-features` manual test, added the balloon toolbar and I can see the difference.

However, there are two issues:

*   On iPhone 8 I need to click multiple times on a button under the selection handler dot to activate it (the button is highlighted after each click, but it's not executed).
*   On iPad, the contextual toolbar doesn't display at all (it works on `master` with the same manual test setup)

Reinmar:
> *   On iPad Air@iOS 12.5.5, the contextual toolbar doesn't display at all (it works on `master` with the same manual test setup)

| Version | Build | Release date |
| --- | --- | --- |
| 12.5 | 16H20 | December 14, 2020; 10 months ago[\[155\]](https://en.wikipedia.org/wiki/IOS_version_history#cite_note-14.3-155) |

After: [https://en.wikipedia.org/wiki/IOS_version_history#iOS_8](https://en.wikipedia.org/wiki/IOS_version_history#iOS_8)

So, this is unfortunately something to take into consideration. It's old, but still "too" alive to ignore.

oleq:
> So, this is unfortunately something to take into consideration. It's old, but still "too" alive to ignore.

It's about 12% of all iPad devices [https://developer.apple.com/support/app-store/](https://developer.apple.com/support/app-store/)

Reinmar:
OK, I looked into our stats and I think we can ditch iOS 12. 12 and earlier versions make for 3.6% of all iOS users which in turn take a small percentage of users globally.

Reinmar:
BTW, the stats that you posted @oleq are from June 2021, so they are already a bit old.

dawidurbanski (approved):
LGTM 👍🏻

PR #10790, opened 2021-11-03
Added support for Shift+Delete on Windows to cut the selected content

### LINKED ISSUE #9326: Shift Delete/Insert support for copy/paste

1. Go to: https://ckeditor.com/ckeditor-5/demo/
2. Copy some text to you clipboard
3. In the demo editor place your cursor at the start of the first sentence
4. Press Shift + End, Delete (this will cut the text)
5. Press Shift + Insert to insert the cut text.

Actual: The text in the clipboard from step 2 is pasted
Expected: The text cut in step 4 using Shift+Delete is pasted (same behaviour as input field)

psmyrek:

Other (typing): Added support for Shift+Delete on Windows to cut the selected content. Closes #9326.

Internal (utils): Added support for detecting if the application is running on Windows.

### DIFF
--- a/packages/ckeditor5-typing/src/deleteobserver.js
+++ b/packages/ckeditor5-typing/src/deleteobserver.js
@@ -41,7 +41,7 @@ export default class DeleteObserver extends Observer {
 			// The Shift + Delete key combination should work in the same way as the `cut` event on a non-collapsed selection on Windows.
 			// In fact, the native `cut` event is actually emitted in this case, but with lower priority. Therefore, in order to handle the
 			// Shift + Delete key combination correctly, it is enough not to emit the `delete` event.
-			if ( env.isWindows && isShiftDeleteOnNonCollapsedSelection( data ) ) {
+			if ( env.isWindows && isShiftDeleteOnNonCollapsedSelection( data, document ) ) {
 				return;
 			}
 
--- a/packages/ckeditor5-typing/src/deleteobserver.js
+++ b/packages/ckeditor5-typing/src/deleteobserver.js
@@ -12,6 +12,7 @@ import DomEventData from '@ckeditor/ckeditor5-engine/src/view/observer/domeventd
 import BubblingEventInfo from '@ckeditor/ckeditor5-engine/src/view/observer/bubblingeventinfo';
 import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
 import env from '@ckeditor/ckeditor5-utils/src/env';
+import { isShiftDeleteOnNonCollapsedSelection } from './utils/utils';
 
 /**
  * Delete observer introduces the {@link module:engine/view/document~Document#event:delete} event.
@@ -35,17 +36,12 @@ export default class DeleteObserver extends Observer {
 		} );
 
 		document.on( 'keydown', ( evt, data ) => {
-			// Do not fire the `delete` event on Windows, if Shift + Delete key combination on a non-collapsed selection was pressed.
+			// Do not fire the `delete` event, if Shift + Delete key combination was pressed on a non-collapsed selection on Windows.
 			//
 			// The Shift + Delete key combination should work in the same way as the `cut` event on a non-collapsed selection on Windows.
 			// In fact, the native `cut` event is actually emitted in this case, but with lower priority. Therefore, in order to handle the
-			// Shift + Delete key combination correctly, it is enough not to emit the `delete` event and stop the original `keydown` event,
-			// so that {@link module:typing/utils/injectunsafekeystrokeshandling `injectUnsafeKeystrokesHandling()`} will not interfere.
-			//
-			// See https://github.com/ckeditor/ckeditor5/issues/9326.
-			if ( env.isWindows && this._isShiftDelete( data ) && this._isNonCollapsedSelection( data ) ) {
-				evt.stop();
-
+			// Shift + Delete key combination correctly, it is enough not to emit the `delete` event.
+			if ( env.isWindows && isShiftDeleteOnNonCollapsedSelection( data ) ) {
 				return;
 			}
 
@@ -114,27 +110,6 @@ export default class DeleteObserver extends Observer {
 	 * @inheritDoc
 	 */
 	observe() {}
-
-	/**
-	 * Checks if <kbd>Shift</kbd> + <kbd>Delete</kbd> keystroke was pressed.
-	 *
-	 * @private
-	 * @param {module:engine/view/observer/domeventdata~DomEventData} domEventData Event data.
-	 * @returns {Boolean}
-	 */
-	_isShiftDelete( domEventData ) {
-		return domEventData.shiftKey && domEventData.keyCode === keyCodes.delete;
-	}
-
-	/**
-	 * Indicates whether the view selection is not collapsed.
-	 *
-	 * @private
-	 * @returns {Boolean}
-	 */
-	_isNonCollapsedSelection() {
-		return !this.view.document.selection.isCollapsed;
-	}
 }
 
 /**
--- a/packages/ckeditor5-typing/src/deleteobserver.js
+++ b/packages/ckeditor5-typing/src/deleteobserver.js
@@ -35,20 +35,6 @@ export default class DeleteObserver extends Observer {
 		} );
 
 		document.on( 'keydown', ( evt, data ) => {
-			// Do not fire the `delete` event on Windows, if Shift + Delete key combination on a non-collapsed selection was pressed.
-			//
-			// The Shift + Delete key combination should work in the same way as the `cut` event on a non-collapsed selection on Windows.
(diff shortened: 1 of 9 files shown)


packages/ckeditor5-typing/src/deleteobserver.js
```diff
@@ -35,6 +35,20 @@ export default class DeleteObserver extends Observer {
...
+			// In fact, the native `cut` event is actually emitted in this case, but with lower priority. Therefore, in order to handle the
+			// Shift + Delete key combination correctly, it is enough not to emit the `delete` event and stop the original `keydown` event,
+			// so that {@link module:typing/utils/injectunsafekeystrokeshandling `injectUnsafeKeystrokesHandling()`} will not interfere.
+			//
+			// See https://github.com/ckeditor/ckeditor5/issues/9326.
+			if ( env.isWindows && this._isShiftDelete( data ) && this._isNonCollapsedSelection( data ) ) {
+				evt.stop();
+
+				return;
+			}
```
niegowski, line 50: Please extract this code from the `DeleteObserver` to some listener for `delete` event outside this file. Remember that events fired by view Document are the bubbling ones so you probably will need to register that handler at the `context: $capture` to make sure that you stop that event before it starts bubbling.

niegowski (changes requested):

packages/ckeditor5-typing/src/deleteobserver.js
psmyrek: I extracted this code to the `Delete` plugin, because similar (OS-related) code already exists there. And it looks like I do not need to listen in the `$capture` phase.

pomek: We wanted to avoid firing the `#delete` event when some conditions are met.

@niegowski, could you explain to us why the first proposed solution is invalid?

Reinmar: I agree with @pomek. The original approach was to not fire `delete` when... no deletion happens. The current code works backward – it fixes what should not happen in the first place. So, I'm for reverting the change.

niegowski: Sorry for not explaining. My biggest concern is that that code should not be inside the observer, those are expected to translate events from the DOM to the view space. Also the `evt.stop()` is not consistent because of that change. For the case whether it should be on the delete event or the keydown event I was considering the `sequence` handling of the delete event but maybe it would be better to listen to keydown and stop it before even the delete observer could look at it. Or maybe it should all be inside the observer but the `evt.stop()` should be consistent? Now I'm not sure

Reinmar: We have two types of observers – ones translating DOM->view events. And one giving more semantics to view events (keydown -> delete). 

I think that, in this case, the `keydown` is still a valid event (as keys were indeed pressed) and should not be stopped. I didn't check previously that @psmyrek actually stopped it so I take my previous comment back – the original approach wasn't good.

The second point is that `DeleteObserver` must not affect `keydown` because it's not its responsibility. `DeleteObserver`'s responsibility is to translate `keydown` to `delete` when `keydown` represents a deletion. And this is the only thing that should happen here.

@psmyrek is **not firing** `delete` enough to resolve this issue? If not, then the rest of the fix should, AFAICS, land in other places. E.g. I remember that we talked about "unsafe keystroke handler" and probably that the second part should be there.

psmyrek: Thanks for clarification. I extracted the logic to a separate util `isShiftDeleteOnNonCollapsedSelection()` because now it has to be used in 2 places: in `DeleteObserver` and in `injectUnsafeKeystrokesHandling()` function.

psmyrek:
The just added `isShiftDeleteOnNonCollapsedSelection()` function needs information about what button was pressed and whether the selection in the view is collapsed. In theory, both of this information are available in the second argument that is received in the `keydown` event listener. Unfortunately, the [`DomEventData#document`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/view/observer/domeventdata.js#L32-L38) property can't be used to read the selection, because hundreds of tests stopped working: they improperly creates `DomEventData()` instances by passing not the view instances, but the view's document instances as a first parameter to the `DomEventData()` constructor, which makes the `DomEventData#document` property `undefined`.

Even fixing these 150+ test cases where `DomEventData()` is instantiated improperly, it still wouldn't be enough. It turns out that in (another) hundreds of tests the `keydown` event is fired in a non-standardized way, i.e. its payload is not created using the `DomEventData()`, but in many cases only single properties needed for a given test are mocked, which causes others to be `undefined`.

For these two reasons, I couldn't use the existing `DomEventData#document` property to get the current document instance (in order to read the selection), but I had to explicitly pass that document instance to the `isShiftDeleteOnNonCollapsedSelection()` function as a second parameter.

niegowski:
> In theory, both of this information are available in the second argument that is received in the `keydown` event listener. Unfortunately, the [`DomEventData#document`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/view/observer/domeventdata.js#L32-L38) property can't be used to read the selection, because hundreds of tests stopped working: they improperly creates `DomEventData()` instances by passing not the view instances, but the view's document instances as a first parameter to the `DomEventData()` constructor, which makes the `DomEventData#document` property `undefined`.

I noticed that the problem with passing the document instance instead of the view instance is bigger than just tests. The [EnterObserver](https://github.com/ckeditor/ckeditor5/blob/03d23626cef93e03cca53e21425b1d8c5de038cb/packages/ckeditor5-enter/src/enterobserver.js#L33) and [DeleteObserver](https://github.com/ckeditor/ckeditor5/blob/f01c8278bc63f96530a3a19ae972e32f4bcba0bb/packages/ckeditor5-typing/src/deleteobserver.js#L89) are both passing the view document instance instead of the view. Could you please create a follow-up issue to fix this problem?

packages/ckeditor5-typing/tests/deleteobserver.js
```diff
@@ -261,6 +261,107 @@ describe( 'DeleteObserver', () => {
 	}
 } );
 
+describe( 'DeleteObserver - handling Shift + Delete on Windows', () => {
```
niegowski, line 264: Maybe it would be good to test backspace whether it's not affected. Also, the tests for non Windows would be good to confirm that this modified behavior is not affecting other platforms.

packages/ckeditor5-typing/tests/utils/injectunsafekeystrokeshandling.js
```diff
@@ -129,5 +130,30 @@ describe( 'unsafe keystroke handling utils', () => {
 				return editor.model.change( writer => writer.batch );
 			}
 		} );
+
+		describe( 'handling Shift + Delete on Windows', () => {
```
niegowski, line 134: Maybe tests on other platforms should be here also. Seems like all the new cases tested in `DeleteObserver` would be needed here.

psmyrek:
The follow-up for improperly initialized `DomEventData()` instance is created: https://github.com/ckeditor/ckeditor5/issues/10820.

packages/ckeditor5-typing/tests/deleteobserver.js
psmyrek: Added one new unit test for non-Windows platform and 2 tests for <kbd>Backspace</kbd> handling for `DeleteObserver`.

packages/ckeditor5-typing/tests/utils/injectunsafekeystrokeshandling.js
psmyrek: Added one test for non-Windows platform and also 2 other to check if new condition does not interfere when it should not.

It is not possible to add all new test cases that are already added for `DeleteObserver`, but only the ones that operate on a non-collapsed selection. The tests in this file uses only the `Typing` plugin (as expected) and the `injectUnsafeKeystrokesHandling()` function deletes only the selected content.

packages/ckeditor5-typing/tests/utils/injectunsafekeystrokeshandling.js
```diff
@@ -129,5 +130,94 @@ describe( 'unsafe keystroke handling utils', () => {
...
+			let view, viewDocument, oldEnvIsWindows;
+
+			beforeEach( () => {
+				view = editor.editing.view;
+				viewDocument = view.document;
+			} );
+
+			describe( 'on Windows', () => {
+				before( () => {
```
niegowski, line 143: Is `before` enough? Shouldn't this be `beforeEach`?

niegowski (approved):

packages/ckeditor5-typing/tests/utils/injectunsafekeystrokeshandling.js
psmyrek: In my opinion `before` is enough - there is no need to set `env` every time, but just once before the whole test group.

PR #10797, opened 2021-11-04
Docs: adding exports to read-only.

### LINKED ISSUE #10751: Docs: Update demos in read-only mode documentation

Currently, our [read-only mode demo](https://ckeditor.com/docs/ckeditor5/latest/features/read-only.html) shows the editor with a completely disabled toolbar, while we have some features that could be used in the read-only mode, i.e. Find and replace, Export to Word and Export to PDF. It would be nice to add such a plugin to the demo's configuration. 

Additionally, there's a common question about hiding the toolbar while entering the read-only mode. It would be beneficial to explain how to do it and introduce a new demo showing this behaviour.

**Scope**

1.  Add Find and replace, Export to Word and Export to PDF plugins to the main read-only mode demo.
2.  Create a new section explaining how to hide/show the toolbar after toggling the read-only mode and create a new demo.

godai78:

Docs: Adding exports and f&r to read-only mode demo. Closes #10751

---

### Additional information

### DIFF
--- a/docs/_snippets/features/read-only.js
+++ b/docs/_snippets/features/read-only.js
@@ -41,16 +41,18 @@ ClassicEditor
 		},
 		exportPdf: {
 			stylesheets: [
-				'EDITOR_STYLES'
+				'../assets/fonts.css',
+				'EDITOR_STYLES',
+				'../assets/read-only-export-pdf.css'
 			],
 			fileName: 'export-pdf-demo.pdf',
 			appID: 'cke5-docs',
 			converterOptions: {
 				format: 'A4',
-				margin_top: '20mm',
-				margin_bottom: '20mm',
-				margin_right: '12mm',
-				margin_left: '12mm',
+				margin_top: '15mm',
+				margin_bottom: '15mm',
+				margin_right: '15mm',
+				margin_left: '15mm',
 				page_orientation: 'portrait'
 			},
 			tokenUrl: false
@@ -72,3 +74,10 @@ ClassicEditor
 	.catch( err => {
 		console.error( err.stack );
 	} );
+
+// Demo styles.
+const link = document.createElement( 'link' );
+link.rel = 'stylesheet';
+link.href = '../assets/read-only-export-pdf.css';
+
+document.head.appendChild( link );
--- a/docs/_snippets/features/read-only.js
+++ b/docs/_snippets/features/read-only.js
@@ -5,11 +5,41 @@
 
 /* globals ClassicEditor, console, window, document */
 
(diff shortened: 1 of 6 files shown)


docs/features/read-only.md
```diff
@@ -18,14 +18,48 @@ The feature may also be used to view content that should not be edited, like fin
...
+			if ( isReadOnly ) {
+				editor.ui.view.top.remove( toolbarContainer );
+			} else {
+				editor.ui.view.top.add( toolbarContainer );
+			}
+		} );
+	} )
+	.catch( error => {
+		console.log( error );
+	} );
```
Mgsy, line 51: I'm not sure about this approach. Basically, it refers to [StickyPanelView](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_ui_panel_sticky_stickypanelview-StickyPanelView.html), not [ToolbarView](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_ui_toolbar_toolbarview-ToolbarView.html) itself. As I can see, the sticky panel is used only in the classic editor, this code can't be used with other editor types that use the toolbar, e.g. the decoupled editor.

How about using CSS for hiding and showing the toolbar? Something like this:

```javascript
const toolbarElement = editor.ui.view.toolbar.element;

editor.on( 'change:isReadOnly', ( evt, propertyName, isReadOnly ) => {
    if ( isReadOnly ) {
        toolbarElement.style.display = 'none';
    } else {
        toolbarElement.style.display = 'flex';
    }
} );
```

The above code snippet will work both with the classic editor and the decoupled editor.

docs/_snippets/features/read-only.js
```diff
@@ -10,15 +10,57 @@ import { CS_CONFIG } from '@ckeditor/ckeditor5-cloud-services/tests/_utils/cloud
...
+			fileName: 'export-pdf-demo.pdf',
+			appID: 'cke5-docs',
+			converterOptions: {
+				format: 'A4',
+				margin_top: '20mm',
+				margin_bottom: '20mm',
+				margin_right: '12mm',
+				margin_left: '12mm',
+				page_orientation: 'portrait'
+			},
```
Mgsy, line 55: This configuration doesn't reflect the way the demo is styled, so the exported document is not 1:1. Additionally, it misses content styles. While adding Export features to the editor, I think we should present it the way it should look like, so the content of the demo in the exported document should look like the one in the editor.

Mgsy: Fixed.

PR #10807, opened 2021-11-04
Improve rendering mode filtering rules

maxbarnas:

Other (engine): Improve rendering mode filtering rules.

---

### Additional information


niegowski (commented):
It would be good to add a test for the image with base64 encoded data that would be matched by `on...=` pattern.

niegowski (approved):

PR #10813, opened 2021-11-05
It should be possible to unlink some text or inline image while GHS is enabled to allow all

### LINKED ISSUE #10800: Unlinking linked inline images while GHS is enabled: wrapping <a> is impossible to remove

Discovered while working on https://www.drupal.org/project/ckeditor5/issues/3246168.

1. Load the following CKE5 plugins: `htmlSupport.GeneralHtmlSupport`, `image.Image`, `image.ImageToolbar`, `image.ImageUpload`, `link.Link`, `link.LinkImage`, `sourceEditing.SourceEditing`, with the following `htmlSupport` config:
```
{"allow":[{"name":{"regexp":{"pattern":"/.*/"}},"attributes":true,"classes":true,"styles":true},{"name":"script"}]}
```
```
<p><a href="http://www.drupal.org/association"><img src="llama.jpg"></a></p>
```
3. Click the `imageInline` widget
4. Click the `Unlink` button in the balloon.
5. Click the `Source` button in the toolbar.

### 🎥 Screencast
linked inline image

### ✔️ Expected result

```
<p><img src="llama.jpg"></p>
```

### ❌ Actual result

```
<p><a><img src="llama.jpg"></a></p>
```

Note: _no crash_, just not correct behavior.

### ❓ Possible solution

This is being caused by the presence of the empty `htmlA` model attribute, which is set by the `image` support of GHS.

It is AFAICT impossible to override using a `dataDowncast` event subscriber because it simply does not get invoked; I could not figure out why that is. I suspect this is related to GHS.

## 📃 Other details

## Issues that touched this problem space
- #7626 → wrong linked image `dataDowncast`
- #8401 → unlinking caused crash, but was pre-GHS
- #9916 → introduced the GHS + `image` integration; specifically see https://github.com/ckeditor/ckeditor5/issues/9916#issuecomment-900124846
- #9482

---

niegowski:

Fix (html-support): It should be possible to unlink some text or inline images while GHS is enabled to allow all. Closes #10800.

Other (engine): The `elementToAttribute` upcast helper should consume the element itself if its purpose was consumed. For example `<a href="...">` should consume `<a>` element after the `href` attribute was consumed. See #10800.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -872,15 +872,19 @@ function prepareToAttributeConverter( config, shallow ) {
 			return;
 		}
 
+		const matchToTestConsumable = {
+			...match.match
+		};
+
 		if ( onlyViewNameIsDefined( config.view, data.viewItem ) ) {
-			match.match.name = true;
+			matchToTestConsumable.name = true;
 		} else {
 			// Do not test or consume `name` consumable.
-			delete match.match.name;
+			delete matchToTestConsumable.name;
 		}
 
 		// Try to consume appropriate values from consumable values list.
-		if ( !conversionApi.consumable.test( data.viewItem, match.match ) ) {
+		if ( !conversionApi.consumable.test( data.viewItem, matchToTestConsumable ) ) {
 			return;
 		}
 
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -944,6 +944,13 @@ function prepareToAttributeConsumingConverter( config ) {
 			return;
 		}
 
+		// We need to convert children if some higher priority converter did not convert them.
+		// After the element got consumed the default handler for converting children will not trigger.
+		if ( !data.modelRange ) {
+			// Convert children and set conversion result as a current data.
+			Object.assign( data, conversionApi.convertChildren( data.viewItem, data.modelCursor ) );
+		}
+
 		// Consume the element itself for element-to-attribute conversion so it won't get converted by any other converter
 		// since this element's whole purpose was to provide some attributes (for example <a href=""> or <span style="font-size: ...">).
 		conversionApi.consumable.consume( data.viewItem, { name: true } );
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.js
@@ -555,12 +555,23 @@ function upcastElementToAttribute( config ) {
 	normalizeModelAttributeConfig( config );
 
 	const converter = prepareToAttributeConverter( config, false );
+	const consumingConverter = prepareToAttributeConsumingConverter( config );
+
+	const converterPriority = config.converterPriority || 'low';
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-html-support/src/converters.js
```diff
@@ -106,7 +112,11 @@ export function viewToAttributeInlineConverter( { view: viewName, model: attribu
 					conversionApi.writer.setAttribute( attributeKey, attributesToAdd, node );
 				}
 			}
-		}, { priority: 'low' } );
+		}, {
+			// With the low - 10 priority to execute after the element-to-attribute converters and their fallback converters
+			// for consuming the element itself so after consuming <a href="..."> both an attribute and the element should be consumed.
```
wimleers, line 117: I'm having trouble parsing this. It's a _very_ long sentence 😅 

It says `both an attribute and the element should be consumed` → so only a single attribute, not all?

packages/ckeditor5-engine/tests/conversion/upcasthelpers.js
```diff
@@ -355,6 +356,44 @@ describe( 'UpcastHelpers', () => {
...
+
+			upcastDispatcher.on( 'element:a', ( evt, data, conversionApi ) => {
+				conversionApi.consumable.consume( data.viewItem, { attributes: [ 'href' ] } );
+			}, { priority: 'high' } );
+
+			expectResult(
+				new ViewAttributeElement( viewDocument, 'a', { href: 'foo' }, new ViewText( viewDocument, 'foo' ) ),
+				'foo'
+			);
+		} );
```
wimleers, line 395: Unfortunately I'm not able to understand what this is testing exactly, because it's testing at the unit level — and pretty much almost at the _function_ level. I _would_ be able to understand this if it were testing `input markup → function calls → output markup` 🤓  But I understand that's not the right level to test or perhaps even too slow given the size of your test suite.

So instead of a thorough review, I'm asking a simple question: **does this test both the scenario in https://github.com/ckeditor/ckeditor5/issues/10800 and in https://github.com/ckeditor/ckeditor5/issues/10800#issuecomment-960654976**?

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -911,6 +923,40 @@ function prepareToAttributeConverter( config, shallow ) {
 	};
 }
 
+// TODO
```
wimleers, line 926: What is there left to do here?

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -911,6 +923,40 @@ function prepareToAttributeConverter( config, shallow ) {
...
+		// We need to convert children if some higher priority converter did not convert them.
+		// After the element got consumed the default handler for converting children will not trigger.
+		if ( !data.modelRange ) {
+			// Convert children and set conversion result as a current data.
+			Object.assign( data, conversionApi.convertChildren( data.viewItem, data.modelCursor ) );
+		}
+
+		// Consume the element itself for element-to-attribute conversion so it won't get converted by any other converter
+		// since this element's whole purpose was to provide some attributes (for example <a href=""> or <span style="font-size: ...">).
+		conversionApi.consumable.consume( data.viewItem, { name: true } );
```
wimleers, line 956: I believe that _this_ is what prevents `<a>` from being generated by a downcast, because only `<a href="">` makes sense. Right?

PR #10814, opened 2021-11-05
The `<figure>` element should not get replicated while GHS is enabled

### LINKED ISSUE #10279: [GHS] Figure tag gets duplicated when GHS config allows all

1. Modify `all-features` manual test, so that it has GHS plugin and configuration that allows all:
```js
		htmlSupport: {
			allow: [
				{
					name: /.*/,
					attributes: true,
					classes: true,
					styles: true
				}
			]
		}
```
2. Open the test in the browser and switch to source editing mode

Alternatively you can load the editor, or `setData()` with the following content:
```html
<figure class="media">
           <oembed url="https://www.youtube.com/watch?v=ZVv7UMQPEWk"></oembed>
</figure>
```

### ✔️ Expected result

Content is source editing mode is the same as in the source html file.

### ❌ Actual result

First case: there is an empty `<figure>` tag at the end of the content.
Alternative case: figure tag with `class="media"` is wrapped in the same duplicated figure tag.

ghs_2

---

niegowski:

Fix (image, media-embed, table): The `<figure>` element should not get replicated while GHS is enabled. Closes #10279.

Tests (html-support): Added tests for double converted `<figure>` elements. See #10279.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-html-support/tests/integrations/image.js
+++ b/packages/ckeditor5-html-support/tests/integrations/image.js
@@ -266,6 +266,24 @@ describe( 'ImageElementSupport', () => {
 			expect( editor.getData() ).to.equal( expectedHtml );
 		} );
 
+		it( 'should not double convert figure element', () => {
+			dataFilter.loadAllowedConfig( [ {
+				name: /^.*$/,
+				styles: true,
+				attributes: true,
+				classes: true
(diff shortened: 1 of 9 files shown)


maxbarnas (approved):
👯

PR #10903, opened 2021-11-25
Removed workspaces from `external/*/external` after extracting collaboration samples and cloud services outside CF

psmyrek:

Internal: Removed workspaces from `external/*/external` after extracting collaboration samples and cloud services outside CF.

---

### Additional information

See https://github.com/cksource/collaboration-features/issues/4316.

### DIFF
--- a/package.json
+++ b/package.json
@@ -194,10 +194,6 @@
       "packages/*",
       "external/*",
       "external/*/packages/*",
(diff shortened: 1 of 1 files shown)


pomek (approved):
LGTM. But this one needs to be merged at the end.

PR #10962, opened 2021-12-06
Improved the EditorWatchdog's editor destruction phase by removing a `change:data` listener before destroying the editor

### LINKED ISSUE #10643: Watchdog's _save method can be called when the editor is being destroyed
[https://github.com/ckeditor/ckeditor5-watchdog/blob/64495826718bf7e9ee414a299680775f7df4a6ea/src/editorwatchdog.js#L248](https://github.com/ckeditor/ckeditor5-watchdog/blob/64495826718bf7e9ee414a299680775f7df4a6ea/src/editorwatchdog.js#L248)

It seems that this issue is connected with the fact that selecting or deselecting a comment currently fires a `change:data` event (#9901).

I suspect that stopping the `change:data` listener just before destroying the editor should be enough to solve this issue:

https://github.com/ckeditor/ckeditor5-watchdog/blob/64495826718bf7e9ee414a299680775f7df4a6ea/src/editorwatchdog.js#L226-L240

ma2ciek:

Fix (watchdog): Prevented `EditorWatchdog` from crashing during the editor destruction process when one of the plugins tries to change data at this moment. Closes #10643.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-watchdog/src/editorwatchdog.js
+++ b/packages/ckeditor5-watchdog/src/editorwatchdog.js
@@ -235,6 +235,11 @@ export default class EditorWatchdog extends Watchdog {
 
 				this._editor = null;
 
+				// Remove the `change:data` listener before destroying the editor.
+				// Incorrectly written plugins may trigger firing `change:data` events during the editor destruction phase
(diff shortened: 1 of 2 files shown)


ma2ciek:
cc @pomek Do I need to rebuild the package?

PR #11012, opened 2021-12-15
Replaced bump-year logic with execution of an external function

### LINKED ISSUE #10909: Create a script for bumping a year in license headers
### Why

To avoid copying the same script over and over repositories.

### What

The goal is to create a script that will update the year in license headers in our files.

See: https://github.com/ckeditor/ckeditor5/blob/5e8257ce9ff51265b28cedffdcd3a4e638f3c23b/scripts/bump-year.js

It should accept a few parameters:

- Starting year, some of our projects began in 2003, some later. It should be configurable.
- Glob patterns for sources to update.
- Perhaps a callback for customizing the license.

przemyslaw-zan:
Internal: Replaced the existing script for bumping a year in license headers with the common one from the `ckeditor5-dev`. Closes #10909.

### DIFF
--- a/scripts/bump-year.js
+++ b/scripts/bump-year.js
@@ -23,7 +23,6 @@ git commit -am "Internal: Bumped the year." && git push
 require( '@ckeditor/ckeditor5-dev-env' )
 	.bumpYear( {
 		cwd: process.cwd(),
-		initialYear: '2003',
 		globPatterns: [
 			{ // LICENSE.md, .eslintrc.js, etc.
 				pattern: '*',
--- a/scripts/bump-year.js
+++ b/scripts/bump-year.js
@@ -31,11 +31,10 @@ require( '@ckeditor/ckeditor5-dev-env' )
 				}
 			},
 			{
-				pattern: '!(build|coverage|node_modules|external)/**',
+				pattern: '!(build|coverage|external)/**',
 				options: {
 					ignore: [
-						'**/ckeditor5-build-*/**',
-						'**/src/lib/**'
+						'**/ckeditor5-build-*/build/**'
 					]
 				}
 			}
--- a/scripts/bump-year.js
+++ b/scripts/bump-year.js
@@ -22,11 +22,23 @@ git commit -am "Internal: Bumped the year." && git push
 
 require( '@ckeditor/ckeditor5-dev-env' )
 	.bumpYear( {
+		cwd: process.cwd(),
 		initialYear: '2003',
 		globPatterns: [
-			{ pattern: '*', options: { dot: true } }, // LICENSE.md, .eslintrc.js, etc.
-			{ pattern: '!(build|coverage|node_modules|external)/**' }
-		],
-		cwd: process.cwd(),
-		callback: string => string.replace( /CKSource - Frederico Knabben./, 'CKSource Holding sp. z o.o.' )
+			{ // LICENSE.md, .eslintrc.js, etc.
+				pattern: '*',
+				options: {
+					dot: true
+				}
+			},
+			{
+				pattern: '!(build|coverage|node_modules|external)/**',
+				options: {
+					ignore: [
+						'**/ckeditor5-build-*/**',
+						'**/src/lib/**'
+					]
+				}
+			}
+		]
 	} );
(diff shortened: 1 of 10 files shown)


scripts/bump-year.js
```diff
@@ -20,65 +20,13 @@ git commit -am "Internal: Bumped the year." && git push
...
-		} );
+require( '@ckeditor/ckeditor5-dev-env' )
+	.bumpYear( {
+		initialYear: '2003',
+		globPatterns: [
+			{ pattern: '*', options: { dot: true } }, // LICENSE.md, .eslintrc.js, etc.
+			{ pattern: '!(build|coverage|node_modules|external)/**' }
+		],
+		cwd: process.cwd(),
+		callback: string => string.replace( /CKSource - Frederico Knabben./, 'CKSource Holding sp. z o.o.' )
```
psmyrek, line 31: The callback should be used only if the default license is not applicable for any reasons. Otherwise, you will have to provide the same callback in all the places.

scripts/bump-year.js
```diff
@@ -20,65 +20,13 @@ git commit -am "Internal: Bumped the year." && git push
...
-				fs.writeFile( fileName, updatedData, err => {
-					if ( err ) {
-						throw err;
-					}
-				} );
-			}
-		} );
+require( '@ckeditor/ckeditor5-dev-env' )
+	.bumpYear( {
+		initialYear: '2003',
```
psmyrek, line 25: The year that is used in most of the places should be defaulted. Provide `initialYear` only if it is different than 2003.

psmyrek (changes requested):

scripts/bump-year.js
```diff
@@ -20,65 +20,24 @@ git commit -am "Internal: Bumped the year." && git push
...
+				pattern: '*',
+				options: {
+					dot: true
+				}
+			},
+			{
+				pattern: '!(build|coverage|node_modules|external)/**',
+				options: {
+					ignore: [
+						'**/ckeditor5-build-*/**',
```
psmyrek, line 37: The rule `'**/ckeditor5-build-*/**'` is too generic. It should be `'**/ckeditor5-build-*/build/**'`.

scripts/bump-year.js
```diff
@@ -20,65 +20,24 @@ git commit -am "Internal: Bumped the year." && git push
...
+						'**/src/lib/**'
```
psmyrek, line 38: It looks like this rule `'**/src/lib/**'` is not needed for the CKEditor 5 repository. I know this one exists in the old script but maybe it is a leftover from the some very old directory structure?

scripts/bump-year.js
```diff
@@ -20,65 +20,24 @@ git commit -am "Internal: Bumped the year." && git push
...
+		cwd: process.cwd(),
+		globPatterns: [
+			{ // LICENSE.md, .eslintrc.js, etc.
+				pattern: '*',
+				options: {
+					dot: true
+				}
+			},
+			{
+				pattern: '!(build|coverage|node_modules|external)/**',
```
psmyrek, line 34: The `node_modules` directory will be ignored in the target script, so there is no need to specify it here.

psmyrek (changes requested):

scripts/bump-year.js
pomek: In the past, in the `ckeditor5-utils`, we had the `src/lib` directory where we kept 3rd party code. See https://github.com/ckeditor/ckeditor5-utils/tree/v10.0.0/src/lib/.

scripts/bump-year.js
```diff
@@ -20,65 +20,23 @@ git commit -am "Internal: Bumped the year." && git push
...
+				pattern: '*',
+				options: {
+					dot: true
+				}
+			},
+			{
+				pattern: '!(build|coverage|external)/**',
+				options: {
+					ignore: [
+						'**/ckeditor5-build-*/build/**'
```
psmyrek, line 37: I've got tons of false positives that translation files do not have license header in the `build` directory, i.e. `packages/ckeditor5-word-count/build/translations/hi.js`. It looks like it works the same in the original version of this script, but maybe it's time to improve it?

Skipping all `build` directories in the packages:

```suggestion
						'**/ckeditor5-*/build/**'
```

psmyrek (commented):
Just minor improvement to get rid of false positives.

psmyrek (approved):
LGTM 🎉 

It should be merged after the `ckeditor5-dev-env` is released and contains changes from https://github.com/ckeditor/ckeditor5-dev/pull/738.

PR #11097, opened 2022-01-12
Expose utilities as plugins to make them available in 3rd-party DLLs. Updated DLL docs.

ma2ciek:

Other (font,image,table): Moved utils functions as plugins to make them available in DLLs.  
Docs: Updated documentation for using and creating DLL builds. Added documentation for DLL builds with collaboration features.

MINOR BREAKING CHANGE (table): Moved table selection utilities to the `TableUtils` plugin.  
MINOR BREAKING CHANGE (image): Converted image caption utilities to the `ImageCaptionUtils` plugin. 

---

### Additional information


ma2ciek:
I'll rebase this PR soon.

ma2ciek:
I'll rebase this PR soon.

ma2ciek:
> I have one main doubt about those solutions. Now some commands use `xxxUtils` plugins. Of course, those plugins are assumed to be loaded (because a command cannot require a plugin in any way). So, for example, we have an `ImageCaptionCommand` that uses `ImageCaptionUtils` and it assumes it is loaded, because if `ImageCaptionCommand` is loaded, it means that `ImageCaption` is loaded, which means that `ImageCaptionUtils` is loaded.

> This is an indirect requirement, which is a little stinky. OTOH, it will make no problems as those parts of code will be use that way always, but I am not sure about it. [**@Reinmar**](https://github.com/Reinmar), [**@pomek**](https://github.com/pomek) do we have any examples like that already in existing codebase? Is there any sensible alternative?

:point\_up: AFAIR the `TableUtils` plugin is now imported by a few table commands, also this plugin is passed to a few table utils selection functions as a parameter :smile:

Yeah, but I need to agree that it is not perfect as it introduces some kind of plugin that should be "pure" and it should be loaded when the command is available. I have another idea to leave the functions as they were and introduce a plugin that would group all these functions and would expose them outside of the package. This way all commands would call the functions directly, but as a plugin, they would be available outside of the package. OTOH there would be more code, harder to document, and so on.

pomek:
I am sorry, but I don't remember whether we had a situation where a command assumes that some plugin must be loaded.

I'd say we cannot assume it. We never know which plugins will be loaded into an editor. You can create a command instance without using the entire set of plugins. 

I'd move the required instances of plugins to the class constructor. 

```js
class FooCommand {
	constructor( editor, imageCaptionUtils ) { /* ... */ }
}

This way, they will be continuously available. From our point of view, it does not change too much as these plugins exist anyway. From an integrator's point of view, it prevents invalid objects and unexpected results.

**Comment by ma2ciek** [comment]
We have a few places like that already:

https://github.com/ckeditor/ckeditor5/blob/638ef662f630da3478bf4a5d2b94fe87c2e497b8/packages/ckeditor5-table/src/commands/insertcolumncommand.js#L71

https://github.com/ckeditor/ckeditor5/blob/638ef662f630da3478bf4a5d2b94fe87c2e497b8/packages/ckeditor5-table/src/commands/insertrowcommand.js#L70

https://github.com/ckeditor/ckeditor5/blob/638ef662f630da3478bf4a5d2b94fe87c2e497b8/packages/ckeditor5-table/src/commands/inserttablecommand.js#L55

https://github.com/ckeditor/ckeditor5/blob/638ef662f630da3478bf4a5d2b94fe87c2e497b8/packages/ckeditor5-table/src/commands/mergecellcommand.js#L108

and so on. 

So it would not be a new thing. But I agree that passing these "special" utility plugins to the constructor would be a bit cleaner.

**Comment by scofalik** [comment]
What about moving `ImageCaptionUtils` to `ImageUtils`?

**Comment by ma2ciek** [comment]
I see that the CC is below 100%. I'll write the missing tests soon.

**Comment by ma2ciek** [comment]
@godai78 @pomek, could you help with reviewing the documentation changes that this PR affects?

**Comment by pomek** [comment]
> @godai78 @pomek, could you help with reviewing the documentation changes that this PR affects?

I read `*.md` files, and everything seems okay.

**Review by godai78 — APPROVED** [review]

File: docs/builds/guides/development/dll-builds-collaboration-features.md [hunk 1]
```diff
@@ -0,0 +1,176 @@
...
+    This guide discusses using a DLL build together with CKEditor 5 Collaboration Features and is supplemental to the {@link builds/guides/development/dll-builds CKEditor 5 DLL builds guide}.
+
+    Make sure to read the {@link builds/guides/development/dll-builds base guide} first before proceeding.
+</info-box>
+
+## Anatomy of a DLL build with CKEditor 5 collaboration features
+
+A DLL build of the editor consists of the following parts:
+
+* **Base DLL build**. It is a single JavaScript file that combines the contents of several core CKEditor 5 packages: `utils`, `core`, `engine`, `ui`, `clipboard`, `enter`, `paragraph`, `select-all`, `typing`, `undo`, `upload`, and `widget`. These packages are either the framework core, or are features used by nearly all editor installations (`ckeditor5` on npm).
```
scofalik, line 20:
```suggestion
* **Base DLL build**. It is a single JavaScript file that combines the contents of several core CKEditor 5 packages: `utils`, `core`, `engine`, `ui`, `clipboard`, `enter`, `paragraph`, `select-all`, `typing`, `undo`, `upload`, and `widget`. These packages are either the framework core, or are features used by nearly all editor installations. The build is available on NPM in `ckeditor5` package.
```

scofalik: How about wording it this way. The way it is worded now is not clear, the note "(`ckeditor5` on npm)" is sudden and is not connected with the previous sentence. If you agree with me (I might not understood something), please rewrite it in other places too.

scofalik: Or you can even leave it as it is in other places, at it will be now more clear what "(\`ckeditor5-collaboration\` on NPM)" means.

docs/builds/guides/development/dll-builds-collaboration-features.md
```diff
@@ -0,0 +1,176 @@
...
+* **Base DLL build for CKEditor 5 Collaboration Features**. It is a single JavaScript file that includes all necessary files for the collaboration features packages and extends the base DLL for CKEditor 5 (`ckeditor5-collaboration` on NPM).
+* **DLL-compatible package builds**. Every package that is not a part of the base DLL builds, is built into a DLL-compatible JavaScript file (`@ckeditor/ckeditor5-*` on NPM). The CKEditor 5 Collaboration Features DLL builds are available in this format as well.
+
+In order to create an editor, you need to use the two base DLL builds plus several DLL-compatible package builds.
```
scofalik, line 24:
```suggestion
In order to create an editor with collaboration features, you need to use the two base DLL builds plus a DLL-compatible package build for each plugin you would like to include.
```

docs/builds/guides/development/dll-builds.md
```diff
@@ -20,29 +20,27 @@ This is where the DLL builds come to the rescue.
...
+</info-box>
 
 ## Anatomy of a DLL build
 
 A DLL build of the editor consists of two parts:
 
-* **Base DLL build**. It is a single JavaScript file that combines the contents of several core CKEditor 5 packages: `utils`, `core`, `engine`, `ui`, `clipboard`, `enter`, `paragraph`, `select-all`, `typing`, `undo`, `upload`, and `widget`. These packages are either the framework core, or are features used by nearly all editor installations.
-* **DLL-compatible package builds**. Every package that is not part of the base DLL build is built into a DLL-compatible JavaScript file.
+* **Base DLL build**. It is a single JavaScript file that combines the contents of several core CKEditor 5 packages: `utils`, `core`, `engine`, `ui`, `clipboard`, `enter`, `paragraph`, `select-all`, `typing`, `undo`, `upload`, and `widget`. These packages are either the framework core, or are features used by nearly all editor installations (`ckeditor5`).
```
scofalik, line 33: Same note as earlier.

docs/builds/guides/migration/migration-to-33.md
```diff
@@ -0,0 +1,107 @@
...
+	If you want to keep the same user experience after updating the editor, configure the comments editor as shown in the example above.
+</info-box>
+
+If the configuration is not provided, a warning will be logged in the console and the the comments editor will be initialized with the most basic features, that is typing, paragraph and undo. 
+
+To hide the warning (and use the basic configuration), provide an empty configuration for the comments editor:
+
+```js
+ClassicEditor.create( document.querySelector( '#editor' ), {
+	// Other configurations...
```
scofalik, line 101:
```suggestion
	// ...
```

docs/builds/guides/development/dll-builds-collaboration-features.md
godai78: "a DLL-compatible package build for each plugin you would like to include" would be my suggestion rather.

scofalik: Hehe, I had it like this initially :).

scofalik (changes requested):
As commented.

ma2ciek:
I fixed these docs issues. Now it should be more clear.

scofalik:
It failed for the third time. Maybe the changes lead to some error that results in DOM structure not being cleared or something :/.

PR #11098, opened 2022-01-12
Integrated the enter feature with document lists

### LINKED ISSUE #10879: [Document lists] Editing part - enter, shift+enter

### LINKED ISSUE #10976: Implement SplitDocumentListItemCommand

This command sets a new random id of the list item.

*   Feels (needs confirmation) like this command will work only on the first selection position regardless of whether the selection is collapsed or not.
*   Scope
    *   \+ Write the new algorithm.
        *   The new list id applies to all blocks that used to have the original id the selection was anchored to (onwards).
    *   \+ Write new tests.

oleq:

Internal (list): Integrated the enter feature with document lists. Closes #10879. Closes #10976.


packages/ckeditor5-list/src/documentlist/documentlistediting.js
```diff
@@ -84,14 +92,68 @@ export default class DocumentListEditing extends Plugin {
...
+				else if ( isLastBlockOfListItem( positionParent ) ) {
+					editor.execute( 'splitListItem' );
+
+					data.preventDefault();
+					evt.stop();
+				}
+			}
+		}, { context: 'li' } );
+
+		// In some cases, the integration with the enter key is done after the default handler in EnterCommand.
```
niegowski, line 123:
```suggestion
		// In some cases, after the default block splitting, we want to modify the new block to become a new list item instead of an additional block in the same list item.
```

packages/ckeditor5-list/tests/documentlist/documentlistediting-integrations.js
```diff
@@ -0,0 +1,1648 @@
...
+			} );
+
+			indentCommand.on( 'afterExecute', ( evt, data ) => {
+				changedBlocks.push( ...data );
+			} );
+		} );
+
+		describe( 'collapsed selection', () => {
+			describe( 'with just one block per list item', () => {
+				it( 'should outdent if the slection in the only empty list item (convert into paragraph and turn of the list)', () => {
```
niegowski, line 538:
```suggestion
				it( 'should outdent if the slection in the only empty list item (convert into paragraph and turn off the list)', () => {
```

niegowski (approved):

PR #11117, opened 2022-01-16
HTMLDataProcessor should preserve leading non-layout elements

### LINKED ISSUE #11110: <script> or <style> that's used at the beginning of an HTML data string is lost
*   It’s moved by HTML parser to `<head>`
    *   Same issue as with HTML comments in the past (see #9861 and [https://github.com/ckeditor/ckeditor5/blob/82ce2e9671b8ec98fb20f3ee647cca94ec421949/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js#L122-L133](https://github.com/ckeditor/ckeditor5/blob/82ce2e9671b8ec98fb20f3ee647cca94ec421949/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js#L122-L133))
*   The first thing to check is why we cannot wrap the entire content with `<body>` before passing it here: [https://github.com/ckeditor/ckeditor5/blob/82ce2e9671b8ec98fb20f3ee647cca94ec421949/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js#L119](https://github.com/ckeditor/ckeditor5/blob/82ce2e9671b8ec98fb20f3ee647cca94ec421949/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js#L119)
    *   If it works, get also rid of the hack for comments.
    *   You can check with @psmyrek  why we didn’t go this way.
    *   One hypothesis: due to parsing full HTML content, e.g. coming from the clipboard. It may already contain doctype, \<html> and \<body>.
    *   If the above is true, we can still try to check with a simple regexp for `<body *>` in the original HTML. If the `<body>` already exists, don’t the HTML.
*   If not, let’s use the same hack as we used for comments.
    *   However, if we’ll apply the same hack, we may start passing through some \<script> tags that were originally in the \<head> (e.g. in a \<head> of a pasted content).
    *   👆We are not able to verify what all external applications do and whether they put any `<script>` tags in the content, so we just need to wait for feedback.

niegowski:

Fix (engine): `HTMLDataProcessor.toView()` should preserve leading non-layout elements while loading partial HTML. Closes #11110.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
+++ b/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.js
@@ -12,8 +12,6 @@
 import BasicHtmlWriter from './basichtmlwriter';
 import DomConverter from '../view/domconverter';
 
-import isComment from '@ckeditor/ckeditor5-utils/src/dom/iscomment';
-
 /**
  * The HTML data processor class.
  * This data processor implementation uses HTML as input and output data.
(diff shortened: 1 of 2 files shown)


Reinmar:
Glad to see this approach worked :)

PR #11123, opened 2022-01-17
List properties should not crash when removing a last list item from the list next to non-list element

### LINKED ISSUE #8642: Null reference in ListStyles causes error when removing last item in a list
https://github.com/ckeditor/ckeditor5/blob/0ad1d6a0587cb8223f33e0d5b6d186ebcfb4b4a0/packages/ckeditor5-list/src/liststyleediting.js#L194

stack reads from bottom up here, this was after selecting a list item to paste onto

Cannot read property 'is' of null 
    CkEditor/node_modules/@ckeditor/ckeditor5-engine/src/model/position.js:931:14 Function._createAt
    CkEditor/node_modules/@ckeditor/ckeditor5-engine/src/model/model.js:635:23 nl.createPositionAt
    CkEditor/node_modules/@ckeditor/ckeditor5-engine/src/model/writer.js:607:20 _c.createPositionAt
    CkEditor/node_modules/@ckeditor/ckeditor5-list/src/liststyleediting.js:194:32 

Seems (without looking at it too much) the the referenced line needs to be updated to handle a null

pomek:

Fix (list): The list properties feature will no longer crash when removing content from the last list item which is next to a non-list element. Closes #8642.

### DIFF
--- a/packages/ckeditor5-list/src/listpropertiesediting.js
+++ b/packages/ckeditor5-list/src/listpropertiesediting.js
@@ -214,6 +214,13 @@ export default class ListPropertiesEditing extends Plugin {
 					direction: 'forward'
 				} );
 
+				// If the selection ends in a non-list element, there are no <listItem>s that would require adjustments.
+				// See: #8642.
+				if ( !secondListMostOuterItem ) {
+					firstMostOuterItem = null;
+					return;
+				}
+
 				const items = [
 					secondListMostOuterItem,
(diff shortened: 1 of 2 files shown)


pomek:
IDK why Travis didn't start a new job. Perhaps an empty commit is required. Please, verify the fix manually, run tests, then push an empty commit.

pomek:
Ok, Travis started.

maxbarnas:
Dear @ckeditor/qa-team, could you take a look at it and verify it is working as expected?

niegowski (approved):

PR #11133, opened 2022-01-19
New conversion documentation

### LINKED ISSUE #10294: Redesign and introduce new reconversion system
### What do we have now

There is a downcast helper for `elementToElement` with a `triggerBy` option that triggers reconversion of a specific model element by removing its view representation (but keeping mappings), triggering the new view structure conversion callbacks (for that model element only), and reusing the child elements' old view representations. 

### What’s wrong with what we have now

To downcast a single model element to a slightly more complex structure we need to use the low-level API (so we need to take care of consuming what we converted and to bind model elements with view elements). That's because we are converting a subtree of a model structure. The reusing of view elements works sort of in a "magical" way. 

There is also an issue about the current implementation conflicting different `triggerBy` converters: #9641.

### What could we improve in the current solution

It could be more understandable if a single `elementToElement` converter would handle a single model element and do not touch its children elements. The conversion callback could provide a view structure with the special placeholders (slots) that model child elements should be downcasted into.

### Why do we think it can be done better?

The current `triggerBy` usage is very confusing. It requires to convert (in a single converter) not only an element itself but also its model child elements (and only their content is reused). This would be more declarative if we could provide conversion for a single model element to a bigger view structure with placeholders/slots for model child elements so that it could take care of inserting view elements in the correct positions in the view structure (not only as a direct child of the main element).

That slots could also have conditions for example to handle specific table rows (`thead` vs `tbody` that depends on `headingRows` attribute value).

### What do we have prototyped so far? Where’s that code?

The early PoC is in the [list reconversion branch](https://github.com/ckeditor/ckeditor5/compare/i/9407-list-reconversion) (that includes a lot of other experiments for `rangeToStructure` conversion). That branch is currently broken (because of changes in progress). The most interesting function is [`insertSlotted`](https://github.com/ckeditor/ckeditor5/blob/4a001997e3f522994efc0ef584cd53902d69c312/packages/ckeditor5-engine/src/conversion/downcasthelpers.js#L1084) that is supposed to get triggered after the old view was removed (but the view mapping was temporarily kept until the end of the conversion process to be reused while converting slots' content).

### A high-level overview of a new mechanism

The solution could be split into layers:

*   `DowncastDispatcher` could fire a new `reduceChanges` event before the conversion process starts to allow features (or conversion helpers) to alter the list of changes provided by the `Differ` so it could detect cases where the reconversion should be used. The re-conversion would be split into 2 events - one to remove the old view representation of a model element and one to insert a new representation. The old view elements would get removed from the view and unbound in the `Mapper` but it could keep temporal mapping to allow reusing view elements that got removed in the same conversion process. Those would get flushed at the end of the whole conversion.
*   `DowncastHelpers` `elementToElement` and `elementToStructure` would listen for the `reduceChanges` event to register callbacks for `triggerBy` changes to trigger re-conversion is a specified situation.
*   `elementToStructure` downcast helper could provide a factory for placeholders/slots so the simple view generator could put those into the structure and the helper itself would take care of converting those slot's content or reusing the view that was previously there.

### Example API proposal

The high-level API:

```javascript
editor.conversion.for( 'downcast' ).elementToStructure( {
	model: 'table', 
	view: ( modelElement, conversionApi ) => {
		const viewElement = conversionApi.writer.createContainerElement( 'div' );
		
		// ...
		
		writer.insert( writer.createPositionAt( viewElement, 0 ), conversionApi.slotFor( modelElement, 'children' ) );
		
		// ...
		
		return viewElement;
	}, 
	triggerBy: { 
		attributes: [ 'headingRows' ]
	}
} );
```

### Success criteria

*   Simple conversion for a single model element to a bigger view structure without taking care of mapping or consumables.
*   Can be used in tables to simplify the current downcast of the whole table and enable its reconversion with the preservation of table rows and cells.
*   Should be adjustable to use with a `rangeToStructure` future helper that will be used for the document list conversion.

### Risks, other things to change.

*   Temporal mappings kept for the conversion process are not yet in PoC (it was a simple one converter solution).
*   It seems that markers downcasted to view should properly reconvert themself (as it works for the current reconversion).

### Why now

The PoC was prepared while working on #9784 and #9783. This could be extracted from that PoC as it's a separate feature that would be later used by `rangeToStructure` downcast helper.

### What’s next

The next step would be to implement `rangeToStructure` downcast helper and use it for document list conversion.

---

dawidurbanski:
Docs (engine): Overhauled the conversion documentation and introduced the documentation for conversion helpers. Closes #10294.

Docs (ckeditor5): Links in the migration guide point to the new conversion documentation.


package.json
```diff
@@ -187,7 +187,8 @@
     "packages/*/build/**",
     "packages/*/src/lib/**",
     "coverage/**",
-    "external/**"
+    "external/**",
+    "packages/ckeditor5-engine/docs/framework/guides/mini-inspector"
```
dawidurbanski, line 191: Note to myself: remove this once mini inspector imported from the main inspector package.

packages/ckeditor5-engine/docs/_snippets/framework/mini-inspector-bold.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
```
dawidurbanski, line 4: Update the license header

packages/ckeditor5-engine/docs/_snippets/framework/mini-inspector-heading.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
```
dawidurbanski, line 4: Update the license header

packages/ckeditor5-engine/docs/_snippets/framework/mini-inspector-paragraph.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
```
dawidurbanski, line 4: Update the license header

packages/ckeditor5-engine/docs/_snippets/framework/mini-inspector-structure.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
```
dawidurbanski, line 4: Update the license header

packages/ckeditor5-engine/docs/_snippets/framework/mini-inspector-upcast-attribute.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
```
dawidurbanski, line 4: Update the license header

packages/ckeditor5-engine/docs/_snippets/framework/mini-inspector-upcast-element.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
```
dawidurbanski, line 4: Update the license header

packages/ckeditor5-engine/docs/_snippets/framework/mini-inspector.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
```
dawidurbanski, line 4: Update the license header

packages/ckeditor5-engine/docs/framework/guides/deep-dive/conversion/helpers/downcast.md
```diff
@@ -0,0 +1,376 @@
...
+		model: 'paragraphSeparator',
+		view: 'hr'
+} );
+```
+
+### Using view element definition
+
+You might want to output a view element that has more attributes, e.g. a class name. To achieve that you can provide [element definition](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_elementdefinition-ElementDefinition.html) in the `view` property:
+
+```jsx
```
dawidurbanski, line 33: I'm wondering why we do use `jsx` highlighting here instead of `js`

dawidurbanski: Same for all 33 occurrences.

packages/ckeditor5-engine/docs/framework/guides/deep-dive/conversion/helpers/downcast.md
```diff
@@ -0,0 +1,376 @@
...
+	} );
+```
+
+Check out the [ElementDefinition documentation](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_elementdefinition-ElementDefinition.html) for more details.
+
+### Creating a view element using a callback
+
+Another way of writing a converter from the previous section using a callback would look like this:
+
+```jsx
```
dawidurbanski, line 51: There are some issues with tabs in the following snippet

packages/ckeditor5-engine/docs/framework/guides/deep-dive/conversion/helpers/downcast.md
```diff
@@ -0,0 +1,376 @@
...
+
+```html
+<div class="horizontal-line">
+	<hr />
+</div>
+```
+
+you can use a converter similar to this:
+
+```jsx
```
dawidurbanski, line 133: There are some issues with tabs in the following snippet

packages/ckeditor5-engine/docs/_snippets/framework/mini-inspector-bold.js
maxbarnas: Fixed.

maxbarnas (approved):

PR #11165, opened 2022-01-26
Docs: Added info box to the Angular integration guide

### LINKED ISSUE #11156: Docs:Angular integration Quick Start leading to strict mode error
### Angular integration Quick Start guide in current form is leading directly to strict mode error and provide the answer far below moment in which error occurs

During integration CKEditor 5 with Angular project, someone who is following the guide and has project configured with strict mode enabled (e.g. created with Angular CLI) will certainly see `Could not find a declaration file for module` in his console.

The guide suggests that everything should work fine, but gives no clue why it might not work in that case. The answer is far below the integration, also in the Quick Start section (https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/angular.html#strict-mode-project-tips) but it might be not seen at first and users might Google it (e.g. https://stackoverflow.com/questions/62690188/error-could-not-find-a-declaration-file-for-module-ckeditor-ckeditor5-build-c) instead of having a direct answer in the documentation. 

I was thinking about some simple solution to this problem and my idea is to put one sentence of the "Strict mode project tips" at the begging of the Quick Start section inside of a small frame with a link to the section describing the problem and solution in more details (we already have it). The sentence is: "If you have the strict mode set in your project, you need to specify types for CKEditor 5 packages. Otherwise, you will get the Could not find a declaration file for module error."

The idea which stands behind it is to give users the clue at the very begging of integration that strict mode needs additional action to be done. In the current version of docs users who are not aware of that by themselves will certainly see the `Could not find a declaration file for module` error. Giving a clue at the begging of Quick Start will benefit in two ways:

- Experienced users will follow the required steps at the beginning and never see the error
- Beginners will probably see the error but will have in mind that something about it was in the docs

dufipl:
Docs: Added info box to the Angular integration guide. Closes #11156

### DIFF
--- a/docs/builds/guides/frameworks/angular.md
+++ b/docs/builds/guides/frameworks/angular.md
@@ -33,6 +33,10 @@ All available versions are [listed on npm](https://www.npmjs.com/package/@ckedit
 
 ## Quick start
 
+<info-box>
+	CKEditor 5 packages do not include TypeScript typings yet. With strict mode enabled, you must take care of the typings. **Please bear in mind that projects generated using Angular CLI have strict mode enabled by default.** Otherwise, you might see the `Could not find a declaration file for module` error. For more information, please see [Strict mode project tips](#strict-mode-project-tips).
+</info-box>
+
 In your existing Angular project, install the [CKEditor 5 WYSIWYG editor component for Angular](https://www.npmjs.com/package/@ckeditor/ckeditor5-angular):
 
 ```bash
--- a/docs/builds/guides/frameworks/angular.md
+++ b/docs/builds/guides/frameworks/angular.md
@@ -34,7 +34,9 @@ All available versions are [listed on npm](https://www.npmjs.com/package/@ckedit
 ## Quick start
 
 <info-box>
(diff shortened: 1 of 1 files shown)


docs/builds/guides/frameworks/angular.md
```diff
@@ -33,6 +33,10 @@ All available versions are [listed on npm](https://www.npmjs.com/package/@ckedit
 
 ## Quick start
 
+<info-box>
+	CKEditor 5 packages do not include TypeScript typings yet. With strict mode enabled, you must take care of the typings. **Please bear in mind that projects generated using Angular CLI have strict mode enabled by default.** Otherwise, you might see the `Could not find a declaration file for module` error. For more information, please see [Strict mode project tips](#strict-mode-project-tips).
```
godai78, line 37:
```suggestion
	CKEditor 5 packages do not include TypeScript typings yet. With strict mode enabled, you must take care of the typings yourself. Otherwise, you might see the `Could not find a declaration file for module` error. **Please bear in mind that projects generated using Angular CLI have strict mode enabled by default.**
	
	For more information, please see [Strict mode project tips](#strict-mode-project-tips).
```

Mgsy (approved):

PR #11171, opened 2022-01-27
The list style UI should call only the listStyle command

### LINKED ISSUE #11164: [Document lists] The list style UI should call only the listStyle command
Currently it calls two commands:

[https://github.com/ckeditor/ckeditor5/blob/aca6ab6f548e0a4c7334f59588f165beea6829a7/packages/ckeditor5-list/src/listpropertiesui.js#L234-L235](https://github.com/ckeditor/ckeditor5/blob/aca6ab6f548e0a4c7334f59588f165beea6829a7/packages/ckeditor5-list/src/listpropertiesui.js#L234-L235)

This makes little sense. E.g. in case of integration with TC, there will be two suggestions that could be accepted independently.

Try moving the first line into the `ListStyleCommand#execute()`.

The idea is that moving that line there should not affect the UX (what the user will get). The goal is to clean up this situation before we implement list properties for #2973.

arkflpc:

Internal (list): Makes `ListPropertiesUI` not to call `ListCommand` and `ListStyleCommand` in one go. Closes #11164.

---

### Additional information


packages/ckeditor5-list/src/listpropertiesui.js
```diff
@@ -231,7 +231,6 @@ function getStyleButtonCreator( { editor, listStyleCommand, parentCommandName }
 			// If the content the selection is anchored to is not a list, let's create a list of a desired style.
```
maxbarnas, line 231:
```suggestion
			// Otherwise, leave the creation of the styled list to the `ListStyleCommand`.
```

packages/ckeditor5-list/src/liststylecommand.js
```diff
@@ -97,4 +102,29 @@ export default class ListStyleCommand extends Command {
 
 		return numberedList.isEnabled || bulletedList.isEnabled;
 	}
+
+	/**
+	 * @param {Object} options
```
maxbarnas, line 107:
```suggestion
	 * Check if the provided list style is valid. Also change the selection to a list if it's not set yet.
	 *
	 * @param {Object} options
```

maxbarnas (approved):
I've suggested changing/adding two comments to make the code slightly more clear for the next guy. Feel free to apply/modify them. Otherwise everything is great :)

PR #11211, opened 2022-02-04
New reconversion system

### LINKED ISSUE #10358: Introduce `for().elementToStructure()`

### LINKED ISSUE #10359: Enable support for reconversion in `elementToElement()`

**Scope**

*   Introduce: Support for an object format of the model property to handle attributes + name + children
*   Conversion helpers: `insertElement()`
*   `downcastElementToElement()`

### LINKED ISSUE #10376: Change the order/flow of events fired by the DowncastDispatcher

### LINKED ISSUE #10377: Make sure we warn in cases when conversion events were stopped
This change is necessary after https://github.com/ckeditor/ckeditor5/issues/10376 when we changed the flow of events. Now, stopping a conversion event may lead to children/attributes of that node not being converted.

### Old notes from @niegowski 

First, we thought about blocking stopping events, but then, after numerous discussions around this we realized it's not the right direction. Check the first comment under this ticket.

> In order to be able to properly fall back to child nodes conversion, we must ensure that the event is not stopped before the handler at `lowest` priority is called. In the conversion process, we are using the consumables concept that allows more granular handling of what is still there to be processed. In the API docs, we are strictly saying, that [those events should be stopped](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_conversion_downcastdispatcher-DowncastDispatcher.html), but there was a [change in the code](https://github.com/ckeditor/ckeditor5-engine/pull/616/files#diff-33cbdd47655213d8604c26e3905b4aa91af0459dbff09792e4425b1f99080008) that removed that to allow converters that work as side effect triggers.
>
> Since we are mentioning this outdated behavior in the API docs I would suggest making events fired by `DowncastDispatcher` and `UpcastDispatcher` not stoppable to avoid confusion.
>
> This would be a good improvement for #10376 to make sure that the conversion will still work after the changes.

### LINKED ISSUE #10502: Migrate table feature downcast conversion to elementToStructure downcast helper

TODO

PoC is available here: https://github.com/ckeditor/ckeditor5/compare/ck/10358-elementToStructure...ck/10381-imperative-reconversion

---

### LINKED ISSUE #10610: elementToElement() should throw an error when the creator callback creates more than one element
*   Why? Because e2S() should be used. It has better capabilities for that (slots) and people should be using it.
*   Create an error.
*   Features that need to be rewritten, so they will be using `elementToStructure`:
    *   ImageBlock
        *   Both pipelines: `<figure>[SLOT for children]</figure>`
    *   ImageInline
        *   Editing pipeline only: `<span></span>` (no slots)
    *   Pagebreak
        *   Editing pipeline only, a similar case as ImageInline
    *   Horizontal line
        *   Same as above
    *   Media embed
        *   Same as ImageBlock
*   Migration guide entry
    *   We should describe if someone uses `elementToElement` which returns more than one element, `elementToStructure` should be used.

### LINKED ISSUE #10659: New API for #refreshItem() and #refreshMarker()

When researching [#9901](https://github.com/ckeditor/ckeditor5/issues/9901) we concluded that it would be good to have a dedicated method in the API to refresh, or rather reconvert, a marker.

At the moment, the way to do it is to call `writer.updateMarker( markerOrMarkerName )`. We don't like that for three reasons:

1.  The method name suggest that something about the marker should change. It is hard to discover that this method should be used if we want to reconvert a marker.
2.  Reconverting a marker is not really about changing something in the model, while writer's responsibility is to change the model. (Re)conversion is a different process, that is managed by a different entity.
3.  We already have similar case with `Differ#refreshItem()`. It also reconverts, but in that case, a model item.

When we discussed this, we concluded that:

1.  We want to have a dedicated method to reconvert a marker.
2.  Both methods should be most probably called `reconvertMarker()` and `reconvertItem()`.
3.  Neither `Differ` nor `model.Writer` are good place for them.
4.  We also thought about adding this API on `EditingController` or `Conversion` classes.

### LINKED ISSUE #10714: `DowncastWriter#createContainerElement()` should accept a list of child nodes

Right now you need to use `insert()` with `createPosition()`. There's no other method. That's super irritating if you want to create a structure of a couple elements.

This is extracted from https://github.com/ckeditor/ckeditor5/issues/7316.

---

### LINKED ISSUE #10811: Restructuring ckeditor5-list package (splitting into subdirectories)
Currently it's a flat list of files (mixed generic list, list styles, to-do list), this could be split into subdirectories as in the image package.

### LINKED ISSUE #11163: Secure slotFor() use for cases where only elements are used
Extracted from #11149.

For now, we need to go for:

> *   D: Make slots handle only elements, not text nodes.
>     *   Reasonable for now.
>     *   Throw when a text node was encountered by `slotFor()`.

Source: [https://github.com/ckeditor/ckeditor5/issues/11149#issuecomment-1021291990](https://github.com/ckeditor/ckeditor5/issues/11149#issuecomment-1021291990)

Scope: The slot mechanism should throw an error when encountering a text node.

### LINKED ISSUE #11179: Move conversionApi.slotFor() to writer.createSlot()

There is a `conversionApi.slotFor()` API introduced in the `elementToStructure` downcast helper. We need to find a better API for it. Maybe it could be a `writer.createSlot()` or sth like this.

---

dawidurbanski:

Feature (engine): The `DowncastWriter#createContainerElement()` should accept a list of children so bigger view structures can be created in one call. Closes #10714.

Feature (engine): The `elementToElement` downcast helper will log a console warning if multiple elements have been created. Closes #10610.

Feature (engine): The `DowncastDispatcher` will throw an error if any of the model items were not consumed while converting. Closes #10377.

Feature (engine): Introducing `convertItem()`, `convertChildren()` and `convertAttributes()` in the downcast conversion API interface.

Feature (engine): Added support for reconversion in `DowncastHelpers#elementToElement()` downcast helper. Closes #10359.

Feature (engine): Added `DowncastHelpers#elementToStructure()` downcast helper with reconversion support. Closes #10358.

Feature (engine): It's now possible to trigger a nested conversion while downcasting an element.

Docs (engine): Overhauled the conversion documentation and introduced the documentation for conversion helpers. Closes https://github.com/ckeditor/ckeditor5/issues/10294.

Internal (engine): `elementToStructure()` should throw when invoked for an element that allows `$text`. Closes #11163.

Internal (engine): Replaced `conversionApi.slotFor()` in  `elementToStructure` with `writer.createSlot()`. Closes #11179.

Internal (engine): Reconversion helpers can coexist with `Differ#refreshItem()` children.

Other (engine): Implemented the EditingController#reconvertMarker() method to be used instead of Writer#updateMarker() for marker reconversion purposes. Implemented the EditingController#reconvertItem() method to replace Differ#refreshItem(). Closes #10659.

Other (engine): The attribute and child nodes conversion events are fired by the lowest priority handler for the insert event instead of by `DowncastDispatcher` itself. Closes #10376.

Other (engine): Events are fired by the `DowncastDispatcher` even if they were previously consumed. It's the conversion handler's responsibility to check if it can be consumed or if it was already consumed by other converters.

Other (engine): The `DowncastDispatcher#convert()` method introduced as a replacement to previously used `convertInsert()`. The new method handles not only nodes conversion but also markers.

Internal (horizontal-line): The `horizontalLine` element downcast conversion has been changed from `elementToElement` to `elementToStructure`.

Internal (html-support): HTML elements downcast conversion have been changed from `elementToElement` to `elementToStructure`.

Internal (image): The `imageBlock` and `imageInline` elements downcast conversion have been changed from `elementToElement` to `elementToStructure`.

Internal (link): Link `inlineWidget` element downcast conversion has been changed from `elementToElement` to `elementToStructure`.

Internal (media-embed): The `media` element downcast conversion has been changed from `elementToElement` to `elementToStructure`.

Internal (page-break): The `pageBreak` element downcast conversion has been changed from `elementToElement` to `elementToStructure`.

Internal (widget): The `blockWidget` and `inlineWidget` elements downcast conversion have been changed from `elementToElement` to `elementToStructure`.

Internal (heading): Added missing consuming converter.

Internal (heading): Code adjusted to changes in `DowncastDispatcher` API.

Internal (html-embed): Code adjusted to the removal of `triggerBy` conversion option.

Other (list): The `ckeditor5-list` package was restructured into subdirectories. Closes #10811.

Other (list): Downcast conversion should consume downcasted attributes.

Other (table): Table downcast conversion migrated to `elementToStructure()` downcast helper. Closes #10502.

Tests (clipboard): Added missing downcast conversion.

Tests (image): Tests updated to properly handle non-consumed errors fired by the `DowncastDispatcher`.

Tests (alignment, autoformat, block-quote, engine, html-support, image, indent, table, word-count): Tests updated after list package restructuring.

Tests (html-support): Fixed test for checking if an attribute was consumed while converting.

Docs (ckeditor5): Links in the migration guide point to the new conversion documentation.

MAJOR BREAKING CHANGE (list): The `ListEditing`, `ListUI`, `ListStyleEditing`, `ListStyleUI`, `TodoListEditing`, `TodoListUI` plugins were moved to the dedicated subdirectories (`list`, `liststyle`, `todolist`).

MAJOR BREAKING CHANGE (engine): Removed the Differ#refreshItem() method from the public API. Replaced by EditingController#reconvertItem() (see #10659).

MAJOR BREAKING CHANGE (engine): The `DowncastDispatcher` will throw an error if any of the model items were not consumed while converting. Read the [`conversion-model-consumable-not-consumed`](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#conversion-model-consumable-not-consumed) error documentation for more information.

MAJOR BREAKING CHANGE (engine): The `DowncastDispatcher#conversionApi` property is no longer available. The instances of `DowncastConversionApi` being created at the start of conversion.

MAJOR BREAKING CHANGE (engine): The support for the `triggerBy` option for downcast helpers is removed and replaced with the new `elementToStructure()` options.

MINOR BREAKING CHANGE (media-embed): The `createMediaFigureElement` helper function first argument has been changed from `writer` object to `conversionApi` object.

MINOR BREAKING CHANGE (table): The downcast converters of the table feature has been rewritten with the use of `elementToStructure()` and the re-conversion mechanism. See #10502.


Reinmar:
Omg Yes Sweet GIF - Omg Yes Sweet Shocked GIFs

Issue #11166, opened 2022-01-26
[Document Lists] Implement list properties commands


arkflpc:
Commands to implement:
- `listStyle`
- `listStart`
- `listReversed`

These commands should be modeled after existing commands in `list` package. Verification if they work with existing `ListPropertiesUI` is in scope.

Behavior of these commands for collapsed vs non-collapsed selection differ between MS Word and Google Docs. We should go for easiest approach.

niegowski:
Closed in #11249.

PR #11232, opened 2022-02-07
Added DocumentListStyle command

arkflpc:

Internal (list): Adds `DocumentListStyleCommand`. See #11166.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/src/documentlist/utils/listwalker.js
+++ b/packages/ckeditor5-list/src/documentlist/utils/listwalker.js
@@ -7,7 +7,7 @@
  * @module list/documentlist/utils/listwalker
  */
 
-import { first } from 'ckeditor5/src/utils';
+import { first, toArray } from 'ckeditor5/src/utils';
 
 /**
  * Document list blocks iterator.
@@ -20,10 +20,7 @@ export default class ListWalker {
 	 * @param {Object} options
 	 * @param {'forward'|'backward'} [options.direction='backward'] The iterating direction.
 	 * @param {Boolean} [options.includeSelf=false] Whether start block should be included in the result (if it's matching other criteria).
-	 * @param {Boolean} [options.sameItemId=false] Whether should return only blocks with the same `listItemId` attribute
-	 * as the start element.
-	 * @param {Boolean} [options.sameItemType=false] Whether should return only blocks wit the same `listType` attribute.
-	 * @param {Array.<String>} [options.sameAttributes=[]] Additional attributes that must be the same for each block.
+	 * @param {Array.<String>} [options.sameListAttributes=[]] Additional attributes that must be the same for each block.
 	 * @param {Boolean} [options.sameIndent=false] Whether blocks with the same indent level as the start block should be included
 	 * in the result.
 	 * @param {Boolean} [options.lowerIndent=false] Whether blocks with a lower indent level than the start block should be included
@@ -49,34 +46,6 @@ export default class ListWalker {
 		this._referenceIndent = startElement.getAttribute( 'listIndent' );
 
 		/**
-		 * The `listItemId` of the start block.
-		 *
-		 * @private
-		 * @type {String}
-		 */
-		this._startItemId = startElement.getAttribute( 'listItemId' );
-
-		/**
-		 * The `listType` of the start block.
-		 *
-		 * @private
-		 * @type {String}
-		 */
-		this._startItemType = startElement.getAttribute( 'listType' );
-
-		/**
-		 * Additional attributes of the start block.
-		 *
-		 * @private
-		 * @type {Map}
-		 */
-		this._startAttributes = new Map();
-
-		for ( const attr of ( options.sameAttributes || [] ) ) {
-			this._startAttributes.set( attr, startElement.getAttribute( attr ) );
-		}
-
-		/**
 		 * The iterating direction.
 		 *
 		 * @private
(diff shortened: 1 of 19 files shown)


packages/ckeditor5-list/src/documentlist/utils/listwalker.js
```diff
@@ -95,6 +121,13 @@ export default class ListWalker {
 		 */
 		this._sameItemType = !!options.sameItemType;
 
+		/**
+		 * Whether should return only blocks with the same `listStyle`, `listReversed` and `listStart` attributes.
+		 * @private
+		 * @type {Boolean}
+		 */
+		this._sameProperties = !!options.sameProperties;
+
```
niegowski, line 130: Those properties are optional feature-related and extend the set of base list attributes (indent, type, item id). I think that there is no point in extending ListWalker with additional attributes, especially since you could stop iterating it as soon as you encounter the abort element from outside of the walker itself.

packages/ckeditor5-list/src/documentlistproperties/documentliststylecommand.js
```diff
@@ -0,0 +1,138 @@
...
+	/**
+	 * Executes the command.
+	 *
+	 * @param {Object} options
+	 * @param {String|null} [options.type] The type of the list style, e.g. `'disc'` or `'square'`. If `null` is specified, the default
+	 * style will be applied.
+	 * @protected
+	 */
+	execute( options = {} ) {
+		this._tryToConvertItemsToList( options );
```
niegowski, line 57: You should wrap both model changes (changing the list type and changing the list style) in a single change block, otherwise, it would generate 2 separate undo steps.

niegowski (changes requested):
Partial review (tests not verified).

packages/ckeditor5-list/tests/documentlist/documentlistcommand.js
```diff
@@ -30,7 +30,11 @@ describe( 'DocumentListCommand', () => {
 		model.schema.register( 'blockQuote', { inheritAllFrom: '$container' } );
 		model.schema.extend( '$container', { allowAttributes: [ 'listType', 'listIndent', 'listItemId' ] } );
 
-		stubUid();
+		stub = stubUid();
```
niegowski, line 33: Why is this change needed? Isn't `testUtils.createSinonSandbox();` handling the restore correctly?

packages/ckeditor5-list/tests/documentlistproperties/documentliststylecommand.js
```diff
@@ -0,0 +1,469 @@
...
+import { modelList } from '../documentlist/_utils/utils';
+import stubUid from '../documentlist/_utils/uid';
+
+describe( 'DocumentListStyleCommand', () => {
+	let editor, model, bulletedListCommand, numberedListCommand, listStyleCommand, stub;
+
+	beforeEach( async () => {
+		const newEditor = await VirtualTestEditor.create( {
+			plugins: [ Paragraph, DocumentListPropertiesEditing ]
+		} );
```
niegowski, line 19: You should not use the full editor instance to test commands. We are usually instantiating only required classes to test commands to make sure that the command itself is generating a correct model state. Otherwise, post-fixers would fix the model and we would not know if the command itself is generating the correct output.

niegowski (changes requested):

niegowski:
I pushed some proposals of changes: [6a84586](https://github.com/ckeditor/ckeditor5/pull/11232/commits/6a8458623f0cc2f2257e175ca91974fb158b1f82).

packages/ckeditor5-list/tests/documentlist/documentlistcommand.js
arkflpc: Well, it didn't in documentliststylecommand, because I forgot to use this sandbox there... fixing

packages/ckeditor5-list/src/documentlistproperties/documentliststylecommand.js
```diff
@@ -0,0 +1,141 @@
...
+	refresh() {
+		this.value = this._getValue();
+		this.isEnabled = this._checkEnabled();
+	}
+
+	/**
+	 * Executes the command.
+	 *
+	 * @param {Object} options
+	 * @param {String|null} [options.type] The type of the list style, e.g. `'disc'` or `'square'`. If `null` is specified, the default
```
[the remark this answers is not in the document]
arkflpc: It would break the UI. And if UI is updated, it wouldn't work with `ListStyleCommand`. And if `ListStyleCommand` is updated, that would be breaking change.

packages/ckeditor5-list/src/documentlistproperties/documentliststylecommand.js
```diff
@@ -0,0 +1,141 @@
...
+
+		return numberedList.isEnabled || bulletedList.isEnabled;
+	}
+
+	/**
+	 * Check if the provided list style is valid. Also change the selection to a list if it's not set yet.
+	 *
+	 * @param {Object} options
+	 * @param {String|null} [options.type] The type of the list style. If `null` is specified, the function does nothing.
+	 * @private
```
niegowski, line 120:
```suggestion
	 * @private
	 * @param {Object} options
	 * @param {String|null} [options.type] The type of the list style. If `null` is specified, the function does nothing.
```

PR #11235, opened 2022-02-08
Handle `<style>` tag

### LINKED ISSUE #11104: [GHS] Handle <style>
**Sibling issue of #10891.**

`<style>` is now ignored by GHS and cannot be enabled with it in any way.

Discovered downstream, by Drupal: https://www.drupal.org/project/drupal/issues/3256566

1. Load the `htmlSupport.GeneralHtmlSupport` and `sourceEditing.SourceEditing` CKE5 plugins, with the following `htmlSupport` config:
```
{"allow":[{"name":{"regexp":{"pattern":"/.*/"}},"attributes":true,"classes":true,"styles":true},{"name":"style"}]}
```
2. Access a CKEditor 5 instance, use the `Source` button to paste this:
```
<style>p { color: red}</style>
<p>
    Hi
</p>
```
3. Click the `Source` button again to access the rendered HTML representation
4. Click the `Source` button again to access the raw HTML representation

### ✔️ Expected result
Exactly what we pasted:
```
<style>p { color: red}</style>
<p>
    Hi
</p>

```

### ❌ Actual result
```
<p>
    Hi
</p>
```

## 📃 Other details

maxbarnas:

Other (engine): The `<style>` tag will not interfere with the editing experience. Closes #11104.

Feature (html-support): Added `<style>` tag support. Closes #11104.

---

### Additional information


packages/ckeditor5-engine/src/view/domconverter.js
```diff
@@ -1546,7 +1547,9 @@ export default class DomConverter {
 	 * @returns {Boolean}
 	 */
 	_shouldRenameElement( elementName ) {
-		return this.renderingMode == 'editing' && elementName.toLowerCase() == 'script';
+		const name = elementName.toLocaleLowerCase();
```
Reinmar, line 1550: Why `toLocaleLowerCase()`? Is this backed by tests?

packages/ckeditor5-engine/src/view/domconverter.js
```diff
@@ -1644,8 +1661,15 @@ function hasBlockParent( domNode, blockElements ) {
  * {@glink framework/guides/architecture/editing-engine#editing-pipeline editing pipeline} of the editor. To avoid this,
  * the `<script>` element was renamed to `<span data-ck-unsafe-element="script"></span>`.
  *
- * @error domconverter-unsafe-element-detected
- * @param {module:engine/model/element~Element|HTMLElement} unsafeElement The editing view or DOM element
```
Reinmar, line 1648: Why did you remove the element itself from here?

packages/ckeditor5-engine/src/view/domconverter.js
```diff
@@ -1644,8 +1661,15 @@ function hasBlockParent( domNode, blockElements ) {
...
+ * @error domconverter-unsafe-script-element-detected
+ * that was renamed.
+ */
+
+/**
+ * The {@link module:engine/view/domconverter~DomConverter} detected a `<style>` element that may affect the editing experience.
```
Reinmar, line 1669: "While rendering the editor content, the ..."

"To avoid this, the `<style>` element was replaced with ..."

packages/ckeditor5-html-support/tests/integrations/style.js
```diff
@@ -0,0 +1,179 @@
...
+import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
+import GeneralHtmlSupport from '../../src/generalhtmlsupport';
+import { getModelDataWithAttributes } from '../_utils/utils';
+import { getData as getModelData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
+
+/* global console, document */
+
+describe( 'StyleElementSupport', () => {
+	const STYLE = 'div { color: red; }';
+	const CODE_CPP = 'cout << "Hello World" << endl;';
```
Reinmar, line 16: Isn't it a leftover from the `<script>`'s tests?

packages/ckeditor5-engine/src/view/domconverter.js
maxbarnas: Crap. Stupid autocorrect! Of course it should be `toLowerCase()` 🤦

packages/ckeditor5-engine/src/view/domconverter.js
maxbarnas: I noticed that it was not consistent - in one place we were providing the DOM element as a context, but in the other we were giving the View element and DOM element we could provide would be already a replaced element. I decided to get rid of it altogether. 

The error description is clear enough and finding offending `<style>` or `<script>` tag in the raw document shouldn't be so hard that we have to point it out to the user.

packages/ckeditor5-html-support/tests/integrations/style.js
maxbarnas: Fixed.

packages/ckeditor5-engine/src/view/domconverter.js
maxbarnas: Fixed.

Reinmar:
I improved the manual tests a bit because there was no place where I could easily test the change.

### Issue 1.

I stumbled upon an odd issue (probably not a blocker for this PR).

The below content works fine:

```html
<p>
    x
</p>
<script>a</script><style>a</style>
<p>
    x
</p>
```

However, this gets miss-formatted after being passed via setData->getData:

```html
<p>
    x
</p>
<script>a</script>
<style>a</style>
<p>
    x
</p>
```

Upon `setData()` this becomes:

```html
<p>x</p><script>a</script><p>&nbsp;<style>a</style></p><p>x</p>
```

For some reason, the new line character is not correctly ignored by `DomConverter` and probably then we have a chain reaction of incorrect autoparagraphing.

### Issue 2.

This issue is not a blocker as well, but should be resolved soon.

Try this:

```
editor.setData( '<script>a</script>');
editor.getData();
```

An empty string is returned. Same for a lone `<style>`. This is probably an excessive "empty content trimming".

maxbarnas:
Second issue is definitely a separate topic that exists for a long time. The first one though... that's new. Should I create a ticket for that?

Reinmar:
Both deserve tickets. Could you search whether they exist already and if not open new ones? 

I'd also like to prioritize the second one because it may lead to data loss.

Reinmar:
I commented under the second error, but it makes sense to update both. I'll do that.

maxbarnas:
> 
> 
> I commented under the second error, but it makes sense to update both. I'll do that.

Ah! I thought it was specifically directed at the `<style>` error. Especially since the old one seemed fine before. Thanks for the fix.

Reinmar:
Both had the same issue – they missed the context of what's happening. Something was detected... but when? Integrators have no idea what DomConverter is. They might even start worrying that their script tags were completely lost. My changes make the former clearer and slightly (although, not completely) help with the latter. So the error could be made even better, explaining that this is happening only inside the editor.

maxbarnas:
Here are those two tickets for the problems found on review: https://github.com/ckeditor/ckeditor5/issues/11247 and https://github.com/ckeditor/ckeditor5/issues/11248.

Issue #11187, opened 2022-01-31
Merge the reconversion branch to master


Reinmar:
Goal: Merge `ck/10294/reconversion` to `master` and ensure that CI works

TODO:

*  [x] Merge latest `master` to that branch.
*  [x] Merge the documentation branch to `ck/10294/reconversion`.
*  [x] Update license headers to 2022.
*  [x] Check that all CIs pass (including CF and cke5-internal).
*  [x] If slotFor() is already moved to writer, update all its occurrences in the docs.
*  [x] Create a changelog entry summing up all the changes.
*  [x] Prepare merge commit message
*  [x] Remove `conversion-element-to-element-created-multiple-elements` warning and bring back `conversion-element-to-structure-disallowed-text` with better docs (explain to use `elementToElement()` instead) https://github.com/ckeditor/ckeditor5/issues/11163.
*  [x] Review and merge https://github.com/ckeditor/ckeditor5/pull/11236 to `ck/10294/reconversion`.
*  [x] Remember to merge corresponding CF pull request https://github.com/cksource/collaboration-features/pull/4261

Idea: Use the tool that @oleq wrote when working on inline images to extract merge commits from specific tickets to the `ck/10294/reconversion` branch.

dawidurbanski:
> > * [x]  Create a changelog entry summing up all the changes.
> 
> Where is it done?

After consultation with @pomek I've been told that the merge message I prepared is enough for the changelog entry.

> > * [x]  Prepare merge commit message
> 
> Where is it done?

Here: https://github.com/ckeditor/ckeditor5/pull/11211

Reinmar:
> After consultation with @pomek I've been told that the merge message I prepared is enough for the changelog entry.

Right :) Makes sense. Probably my mistake to have two actions for the same thing :)

PR #11236, opened 2022-02-08
Refactor createBlockImageViewElement() helper function.

dawidurbanski:

Internal (image): Refactor `createBlockImageViewElement()` helper function.

________________

Why: Recently we changed the `slotFor()` API by moving it into the writer and renaming the function. This PR aims to do a small refactoring by removing second parameter `slotElement` which is created using `writer` since we already pass the whole `writer` into the `createBlockImageViewElement()`.

It's documented in another PR review and acts as a follow up to it: https://github.com/ckeditor/ckeditor5/pull/11211#pullrequestreview-875583814


Reinmar:
The PR should explain why is this change needed and where it originated from.

I can see a link in #11187 but I don't understand why is it done.

dawidurbanski:
@Reinmar updated the PR description.

Issue #10892, opened 2021-11-23
Provide a solution for how to downcast a table to table>caption


Reinmar:
By default we use `<figure class=table>` with `<figcaption>` which was a design choice made at the beginning mostly because:

*   it's consistent with other features contained in `<figure>`
*   and that, in turn, is a result of `<figure>` being more generic

However, in some cases, e.g. in Drupal, `<table>` + `<caption>` is the expected markup.

I don't think we should have it as a configuration option, because the table feature exists for a long time and this was requested for the second time to my knowledge. But we can provide (even here) a code sample on how to change the output of the editor.

Other remarks:

*   The goal is to mimic CKE4's output.
*   Can we base this on the reconversion work or should we start from master?
    *   If we'll go with the reconversion approach, then we need to start from that branch obviously.
    *   Using reconversion as a base means that we must release reconversion in February's release at the latest.
*   Post-fixer or overriding table's converter?
    *   Post-fixer would be ugly. But it'd be safest. Also: we don't support data view post-fixers yet.
    *   Overriding table's converter:
        *   Would be consistent with how they override images.
        *   Would need to be based on the reconversion branch.
*   The new markup needs to be compatible with:
    *   GHS (table's and caption's attributes)
        *   Needs custom integration
        *   We do not support upcasting custom attributes from `<caption>` -> separate ticket (https://github.com/ckeditor/ckeditor5/issues/10896)
    *   Table properties
        *   On master table's width is set on `<figure>`. After this customization, `<figure>` disappears, so the style needs to be moved to `<table>`.
*   Due to the amount of work, complexity, and integration options, let's deliver this as a plugin, with tests. But let's not cover this in the guides, at least for now.

dawidurbanski:
This ticket might also require to look at this one: https://github.com/ckeditor/ckeditor5/issues/9360

Although strictly speaking it's not a blocker to complete the current one, but it makes it obvious that there is something wrong with the simplest of payloads. For example:

```html
<table>
	<caption>Monthly savings</caption>
	<tr>
		<th>Month</th>
		<th>Savings</th>
	</tr>
	<tr>
		<td>January</td>
		<td>$100</td>
	</tr>
</table>
```

Will be upcasted to:

```jsx
<table headingColumns="2">...</table>
```

Which later will be downcasted in the data pipeline to:

```html
<table?
	<caption>Monthly savings</caption>
	<tbody>
		<tr>
			<th>Month</th>
			<th>Savings</th>
[...]

PR #11240, opened 2022-02-08
Support for plain table + caption in the data pipeline

dawidurbanski:

Feature (table): Introduced the `PlainTableOutput` plugin to override default `figure>caption` markup in the data pipeline. Closes: #10892.

Tests (table): Added tests for the `TableCaptionUI` plugin.

_________

Please note that as per the original ticket we don't document this feature in any guides.

> Let's deliver this as a plugin, with tests. But let's not cover this in the guides, at least for now.


packages/ckeditor5-table/src/plaintableoutput.js
```diff
@@ -0,0 +1,122 @@
...
+	 * @inheritDoc
+	 */
+	init() {
+		const editor = this.editor;
+
+		// Override default table data downcast converter.
+		editor.conversion.for( 'dataDowncast' ).elementToStructure( {
+			model: {
+				name: 'table',
+				attributes: [ 'headingRows' ]
```
maxbarnas, line 45: I am wondering why there's `headingRows` attribute here. Could you explain?

maxbarnas (approved):
Nice, clean work! 🎉

packages/ckeditor5-table/src/plaintableoutput.js
```diff
@@ -0,0 +1,122 @@
...
+	 */
+	static get pluginName() {
+		return 'PlainTableOutput';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	static get requires() {
+		return [ Table, TablePropertiesEditing, TableCaption ];
```
niegowski, line 32: Does it really requires those except table? This will enforce those plugins loaded. Is it expected?

packages/ckeditor5-table/src/plaintableoutput.js
```diff
@@ -0,0 +1,122 @@
...
+
+// The plain table downcast converter callback.
+//
+// @private
+// @param {module:engine/model/element~Element} Table model element.
+// @param {module:engine/conversion/downcastdispatcher~DowncastConversionApi} conversionApi The conversion API object.
+// @returns {module:engine/view/containerelement~ContainerElement} Created element.
+function downcastTableElement( modelElement, { writer } ) {
+	// Table body rows slot.
+	const rowsSlot = writer.createSlot( element => element.is( 'element', 'tableRow' ) );
```
niegowski, line 71: Is it expected that we don't use thead?

packages/ckeditor5-table/src/plaintableoutput.js
dawidurbanski: Well, it was added in order to trigger conversion after `headingRows` attribute change, but it actually turned out it's unnecessary. It's being handled by another converter in the main `Table` plugin.

packages/ckeditor5-table/src/plaintableoutput.js
dawidurbanski: It turned out they are not required. I just had to add some small amends to handle situations where these are not loaded. I also added some tests to cover this.

packages/ckeditor5-table/src/plaintableoutput.js
dawidurbanski: Yes. It is expected.

The markup for "plain" table looks like this:

```html
<table>
  <caption>Monthly savings</caption>
  <tr>
    <th>Month</th>
    <th>Savings</th>
  </tr>
  <tr>
    <td>January</td>
    <td>$100</td>
  </tr>
</table>
````
https://www.w3schools.com/tags/tag_caption.asp

There are no `<thead>` not `<tbody>` tags in it nor they should be. The `<tbody>` is actually added automatically by the browser so it won't hurt to add it on our end.

Header rows are just a normal rows with `<th>` cells inside `<tbody>`.

File: packages/ckeditor5-table/src/plaintableoutput.js [hunk 4]
```diff
@@ -0,0 +1,119 @@
...
+			view: downcastTableElement,
+			converterPriority: 'high'
+		} );
+
+		// Make sure <caption> is always downcasted to <caption> in the data pipeline.
+		editor.conversion.for( 'dataDowncast' ).elementToElement( {
+			model: 'caption',
+			view: 'caption',
+			converterPriority: 'high'
+		} );
```
niegowski, line 51: Does it work ok with images with captions? AFAIK image caption should not be affected by this change, but this converter looks like it would take over it.

niegowski, line 51: Does it work ok with images with captions? AFAIK image caption should not be affected by this change, but this converter looks like it would take over it.

niegowski, line 51: Does it work ok with images with captions? AFAIK image caption should not be affected by this change, but this converter looks like it would take over it.

packages/ckeditor5-table/src/plaintableoutput.js
dawidurbanski: Well, I understand it in a different way, as per [ticket](https://github.com/ckeditor/ckeditor5/issues/10892) description:

> The goal is to mimic CKE4's output.

1. Go here: https://ckeditor.com/docs/ckeditor4/latest/examples/table.html
2. Go to the source editing
3. Paste the table code from my previous comment.
4. Go back to editor
5. Check the source editing again

You get:

```html
<table>
	<caption>Monthly savings</caption>
	<tbody>
		<tr>
			<th>Month</th>
			<th>Savings</th>
		</tr>
		<tr>
			<td>January</td>
			<td>$100</td>
		</tr>
	</tbody>
</table>
```

That's all I care about to meet requirements.

packages/ckeditor5-table/src/plaintableoutput.js
dawidurbanski: Indeed, good catch. I'm working on it.

packages/ckeditor5-table/src/plaintableoutput.js
Reinmar: CKEditor 4's table plugin will output `<thead>`:

1.  Insert table (via the toolbar button)
2.  It indeed has only `<tbody>`
3.  Open table properties
4.  Set "Table headers" to "First row"
5.  Check out the source again:

```html
<table border="1" cellpadding="1" cellspacing="1" style="width:500px">
	<thead>
		<tr>
			<th scope="col">&nbsp;</th>
			<th scope="col">&nbsp;</th>
		</tr>
	</thead>
	<tbody>
		<tr>
			<td>&nbsp;</td>
			<td>&nbsp;</td>
		</tr>
		<tr>
			<td>&nbsp;</td>
			<td>&nbsp;</td>
		</tr>
	</tbody>
</table>
```

packages/ckeditor5-table/src/plaintableoutput.js
dawidurbanski: Fixed.

packages/ckeditor5-table/src/plaintableoutput.js
Reinmar: The goal is to mimic CKE4 output, as you mentioned, and CKE4 will use `<thead>` if necessary.

dawidurbanski: > if necessary

How do I determine when it's necessary and when not?

dawidurbanski: I can't distinguish between the example I posted, and the example you posted @Reinmar.

So I just changed in this way: if there are some heading rows in the model, it will create `<thead>`. If not, it won't.

This means it will work as per your example.

In case someone paste in source editing (or sets the data with `setData()`):

```html
<table>
	<caption>Monthly savings</caption>
	<tbody>
		<tr>
			<th>Month</th>
			<th>Savings</th>
		</tr>
		<tr>
			<td>January</td>
			<td>$100</td>
		</tr>
	</tbody>
</table>
```

The output will be:

```html
<table>
	<caption>Monthly savings</caption>
	<thead>
		<tr>
			<th>Month</th>
			<th>Savings</th>
		</tr>
	</thead>
	<tbody>
		<tr>
			<td>January</td>
			<td>$100</td>
		</tr>
	</tbody>
</table>
```

So it will be affected.

Reinmar:
Ooops. I mistakenly rebased the entire branch when fixing one of my commits:facepalm: 

Sorry for the future readers of this PR.

packages/ckeditor5-widget/src/widgettoolbarrepository.js
```diff
@@ -124,10 +124,18 @@ export default class WidgetToolbarRepository extends Plugin {
...
 			/**
-			 * When {@link #register} a new toolbar, you need to provide a non-empty array with
+			 * When {@link #register registering} a new widget toolbar, you need to provide a non-empty array with
 			 * the items that will be inserted into the toolbar.
 			 *
+			 * If you see this error when integrating the editor, you likely forgot to configure one of the widget toolbars.
+			 *
+			 * See for instance:
+			 *
+			 * * {@link module:table/table~TableConfig#contentToolbar `editor.config.table.contentToolbar`}
```
Reinmar, line 134: It should be `config.table.contentToolbar`. I will fix this on master.

PR #11273, opened 2022-02-14
Docs: Installation docs revamp

Mgsy:

Docs: Installation docs revamp.

---

### Additional information

Requires: cksource/collaboration-features#4462 and cksource/ckeditor5-internal#1437.  
  
Post-merge follow-up: cksource/docs#420.


docs/installation/getting-started/quick-start.md
```diff
@@ -0,0 +1,315 @@
...
+	Some of these plugins are **premium features** which require additional license to run.
+</info-box>
+
+Once you choose all the desired plugins, press the **Next step** button on the top right.
+
+### Toolbar composition
+
+Next step allows you to compose the toolbar. A simple drag-and-drop workspace allows for adding buttons (representing the plugins chosen in the previous step) to the toolbar. The user may change the order of the buttons and dropdowns and group them accordingly.
+
+{@img assets/img/online-builder-04-toolbar-configurator.png 753 The toolbar drag-and-drop configurator.}
```
AnnaTomanek, line 74: A GIF would make better here, I think. cc @godai78

docs/installation/getting-started/quick-start.md
```diff
@@ -0,0 +1,315 @@
...
+</info-box>
+
+If everything worked correctly, you should see:
+
+```
+p@m /workspace/quick-start> ./node_modules/.bin/webpack --mode development
+Hash: c96beab038124d61568f
+Version: webpack 4.15.1
+Time: 3023ms
+Built at: 2018-07-05 17:37:38
```
AnnaTomanek, line 270: Is this what you will see in 2022?

pomek: > Version: webpack 4.15.1

Make sure to review webpack and other dependencies because of the PR downgraded versions. See https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/quick-start.html#lets-start.

pomek (commented):
Make sure to review webpack and other dependencies because of the PR downgraded versions. See https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/quick-start.html#lets-start.

CHANGELOG.md
```diff
@@ -976,15 +976,15 @@ Read more in the blog post: https://ckeditor.com/blog/CKEditor-5-v29.0.0-with-bo
...
 
-* **[build-decoupled-document](https://www.npmjs.com/package/@ckeditor/ckeditor5-build-decoupled-document)**: The official preconfigured [decoupled document build](https://www.npmjs.com/package/@ckeditor/ckeditor5-build-decoupled-document) now ships with the [`ImageResize`](https://ckeditor.com/docs/ckeditor5/latest/api/module_image_imageresize-ImageResize.html) plugin enabled by default. Learn more about it in the [Migration to v.29.x guide](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/migration/migration-to-29.html).
+* **[build-decoupled-document](https://www.npmjs.com/package/@ckeditor/ckeditor5-build-decoupled-document)**: The official preconfigured [decoupled document build](https://www.npmjs.com/package/@ckeditor/ckeditor5-build-decoupled-document) now ships with the [`ImageResize`](https://ckeditor.com/docs/ckeditor5/latest/api/module_image_imageresize-ImageResize.html) plugin enabled by default. Learn more about it in the [Migration to v.29.x guide](https://ckeditor.com/docs/ckeditor5/latest/updating/migration-to-29.html).
```
pomek, line 979: Should we update changelog entries if we don't update GitHub Releases? These links should work anyway for backward compatibility.

docs/framework/guides/contributing/git-commit-message-convention.md
```diff
@@ -172,7 +172,7 @@ Changelog
 
 ## [1.0.0](https://github.com/ckeditor/ckeditor5/compare/v1.0.0...v0.0.1) (2017-01-04)
 
-### MAJOR BREAKING CHANGES [ℹ️](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/versioning-policy.html#major-and-minor-breaking-changes)
+### MAJOR BREAKING CHANGES [ℹ️](https://ckeditor.com/docs/ckeditor5/latest/support/versioning-policy.html#major-and-minor-breaking-changes)
```
pomek, line 175: It would be good to update source scripts too: https://github.com/ckeditor/ckeditor5-dev/blob/8ebb83bd777a4ad05c9c421d7246fde17abba1c7/packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogformonorepository.js#L27-L28.

docs/installation/advanced/predefined-builds.md
```diff
@@ -0,0 +1,651 @@
...
+
+<info-box hint>
+	If you are looking for an easy way to create a custom build of CKEditor 5, check also the [online builder](https://ckeditor.com/ckeditor-5/online-builder/), which allows you to create easily a custom build through a simple and intuitive UI.
+</info-box>
+
+### Requirements
+
+In order to start developing CKEditor 5 you will require:
+
+* [Node.js](https://nodejs.org/en/) 12.0.0+
```
pomek: CKEditor 5 requires Node@14.

https://github.com/ckeditor/ckeditor5/blob/0a9e5a9c34a83324fec2242f2d5c86669c08ed02/package.json#L135

docs/installation/advanced/using-two-editors.md
```diff
@@ -0,0 +1,166 @@
...
+diff --git a/webpack.config.js b/webpack.config.js
+index c57e371..04fc9fe 100644
+--- a/webpack.config.js
++++ b/webpack.config.js
+@@ -21,7 +21,7 @@ module.exports = {
+
+     output: {
+         // The name under which the editor will be exported.
+-        library: 'ClassicEditor',
++        library: 'CKEDITOR',
```
pomek, line 106: * `CKEDITOR` can be misleading with CKEditor 4, which uses the name for exporting the library.
* DLL builds use `CKEditor5`
* Perhaps `Editors` will be enough.

docs/installation/getting-started/quick-start.md
```diff
@@ -0,0 +1,313 @@
...
+
+Then, install packages needed to build CKEditor 5:
+
+```bash
+npm install --save \
+	postcss-loader@3 \
+	raw-loader@3 \
+	style-loader@1 \
+	webpack@4 \
+	webpack-cli@3
```
pomek, line 114: CKEditor 5 uses webpack 5:

https://github.com/ckeditor/ckeditor5/blob/0a9e5a9c34a83324fec2242f2d5c86669c08ed02/package.json#L118-L131

docs/installation/getting-started/quick-start.md
```diff
@@ -0,0 +1,313 @@
...
+	// https://webpack.js.org/configuration/entry-context/
+	entry: './app.js',
+
+	// https://webpack.js.org/configuration/output/
+	output: {
+		path: path.resolve( __dirname, 'dist' ),
+		filename: 'bundle.js'
+	},
+
+	module: {
```
pomek, line 137: Webpack 5 requires changes in the `module` object. See:
https://github.com/ckeditor/ckeditor5/blob/0a9e5a9c34a83324fec2242f2d5c86669c08ed02/packages/ckeditor5-build-classic/webpack.config.js#L60-L93

docs/installation/index.md
```diff
@@ -0,0 +1,344 @@
...
+Find out more about the {@link installation/advanced/plugins plugin development}, how to {@link installation/advanced/saving-data handle and save the data} and what the {@link installation/advanced/features-html-output-overview features' HTML output} is for each plugin. Learn about alternative setups such as {@link installation/advanced/dll-builds DLL builds} or {@link installation/advanced/integrating-from-source integrating CKEditor 5 from source}.
+
+## Related links
+
+ * {@link updating/index Updating CKEditor 5} &ndash; Find out how to keep you installation up-to-date at all times.
+ * {@link features/index Features} &ndash; Learn more about the CKEditor 5 features.
+ * {@link examples/index Examples} &ndash; Try live demos of available predefined builds and custom solutions.
+ * * {@link framework/guides/overview CKEditor 5 Framework} &ndash; Learn how to work with CKEditor 5 Framework, customize it, create your own plugins or custom editors, how to change the UI or even bring your own UI to the editor.
+
+<script type="text/javascript">
```
pomek, line 48: Couldn't we extract the `<script>` source to a new snippet? Then, in a snippet HTML file, execute it.

```html
<script>startQuiz();</script>
```

packages/ckeditor5-core/src/editor/editorconfig.jsdoc
```diff
@@ -24,7 +24,7 @@
  *			.then( ... )
  *			.catch( ... );
  *
- * Check the {@glink builds/guides/integration/configuration Configuration guide} for more information
+ * Check the {@glink /installation/advanced/predefined-builds Configuration guide} for more information
```
pomek, line 27: Absolute URLs may lead to non-existing resources.

image

image

packages/ckeditor5-editor-balloon/src/ballooneditor.js
```diff
@@ -17,7 +17,7 @@ import BalloonEditorUI from './ballooneditorui';
 import BalloonEditorUIView from './ballooneditoruiview';
 
 /**
- * The {@glink builds/guides/predefined-builds/overview#balloon-editor balloon editor} implementation (Medium-like editor).
+ * The {@glink /installation/advanced/predefined-builds#balloon-editor balloon editor} implementation (Medium-like editor).
```
pomek, line 20: image

Please, review all absolute URLs.

pomek (changes requested):
Wow! Amazing work. However, I found a few issues that we should resolve before merging.

* Absolute URLs will not work when building the entire CKEditor documentation.
* I'm not sure about updating the main CHANGELOG.md file.
* Changes in `packages/ckeditor5-build-*/build/ckeditor.js.map` files should be reverted.
* Quiz source could be a snippet.

AnnaTomanek:
I've been reading the docs the entire week and I'm totally impressed about the depth of the changes that this PR covers :heart\_eyes: 

I committed a lot of changes directly to the branch, but here are some more thoughts:

1.  I have a feeling that not all the guides were converted to the new "two path" mindset. As a result, anywhere we talk about customizing your build (like: the installing plugins guide), we refer to and think about the "building from source" way. Online builder path still just appears as a tip, like a curiosity but not really a valid way of doing things. I would get back to it, because I feel all our recommendations should actually be revisited, asking whether we really treat the online builder path as a complete one.
2.  "What’s next" and "Quick recap" notes should have anchors to link to. Otherwise, UX suffers.
3.  Is “Getting and saving data” really an advanced guide? :thinking: I'm a bit on the fence with this one as yeah, maybe it's not a part of the installation itself, but an editor is pretty useless without the ability to manipulate some content with it (and then do something with this content).
4.  Screenshots in `predefined-builds.html` should be updated to show current build configurations.
    1.  Also, the existing inline editor screenshot with such a limited toolbar looks 😱, like this was a build without features.
5.  I really stumbled when I got to the `predefined-builds.html#forking-an-existing-build` section. It is a leftover from the previous builds-first approach, it looked out of place and I had a deja vu feeling when reading it. It feels a bit strange - don’t we have the same thing covered in another article/other articles? Should it be a part of the Alternative setups > Predefined builds article at all? Do we actually still recommend this sort of customization? Quick start with its two paths are two approaches that we chose to promote and I'm honestly wondering why this way would make more sense to anyone. Or maybe if there is something really valuable in it, like the whole "publish your build to npm" thing, this could actually be added as a step in the two recommended paths?

CHANGELOG.md
godai78: There are redirects in place, so we may safely drop changes in the changelog indeed.

docs/installation/getting-started/quick-start.md
godai78: Addressed with PR, together with new predefined builds screenshots.

godai78:
Not removing the .png for tollbar configurator, as I will have to redo the gif - something was changed and now GiFox forces a watermark in the free version.

CHANGELOG.md
godai78: I reverted these.

packages/ckeditor5-editor-balloon/src/ballooneditor.js
godai78: Addressed in [#11422](https://github.com/ckeditor/ckeditor5/pull/11422)

packages/ckeditor5-core/src/editor/editorconfig.jsdoc
godai78: Addressed in [#11422](https://github.com/ckeditor/ckeditor5/pull/11422)

docs/installation/index.md
Mgsy: It will be covered in a follow-up: https://github.com/cksource/ckeditor5-internal/issues/1577.

Mgsy:
Merged the latest stable and resolved conflicts.

PR #11278, opened 2022-02-14
Fixed missing value in the React guide

### LINKED ISSUE #11261: React component docs use an outdated onError signature

The snippet in [Using the document editor build](https://ckeditor.com/docs/ckeditor5/32.0.0/builds/guides/integration/frameworks/react.html#using-the-document-editor-build) shows following `onError` usage:

```
                    onError={ ( { willEditorRestart } ) => {
                        // If the editor is restarted, the toolbar element will be created once again.
                        // The `onReady` callback will be called again and the new toolbar will be added.
                        // This is why you need to remove the older toolbar.
                        if ( willEditorRestart ) {
                            this.editor.ui.view.toolbar.element.remove();
                        }
                    } }
```

The problem is that `willEditorRestart` is passed to a second parameter instead of the first, just like it is mentioned in [Context feature properties](https://ckeditor.com/docs/ckeditor5/32.0.0/builds/guides/integration/frameworks/react.html#context-feature-properties):

> *   `onError` – A function called when the context has crashed during the initialization or during the runtime. It receives two arguments: the error instance and the error details.  
>     Error details is an object that contains two properties:
>     *   `{String} phase`: `'initialization'|'runtime'` – Informs when the error has occurred (during the editor or context initialization, or after the initialization).
>     *   `{Boolean} willContextRestart` – When `true`, it means that the context component will restart itself.

przemyslaw-zan:

Docs: Fixed missing value in the React guide. Closes #11261.

---

### Additional information

### DIFF
--- a/docs/builds/guides/frameworks/react.md
+++ b/docs/builds/guides/frameworks/react.md
@@ -213,7 +213,7 @@ class App extends Component {
 
 						this.editor = editor;
 					} }
-					onError={ ( { phase, willEditorRestart } ) => {
+					onError={ ( error, { willEditorRestart } ) => {
 						// If the editor is restarted, the toolbar element will be created once again.
 						// The `onReady` callback will be called again and the new toolbar will be added.
 						// This is why you need to remove the older toolbar.


pomek (changes requested):
As noted in the documentation, an ~editor~ error instance is the first argument passed in the `onError()` callback.

You can also verify it with the code of the component – https://github.com/ckeditor/ckeditor5-react/blob/0056e33292d1958a1b7ed9b9259584ddbc9a5396/src/ckeditor.jsx#L118.

pomek:
Quoting https://github.com/ckeditor/ckeditor5/issues/11261:

> A function called when the context has crashed during the initialization or during the runtime. **It receives two arguments: the error instance and the error details.**

pomek (approved):
LGTM.

PR #11329, opened 2022-02-22
Ck/11167 document list properties post fixer

### LINKED ISSUE #11167: [Document Lists] Implement a generic post-fixer ensuring correct model of ListProperties
Goal: Make sure that all items in a single list (items at the same level) have the same start, reversed, style.

Most likely, the above will (in the future) be ensured by a combination of:

* (a) DocListProperties and DocList commands cooperating together.
* (b) Listeners for `deleteContent` and `insertContent`.
* (c) A post-fixer that we need anyway to handle e.g. RTC-related issues.

For now, let's focus only on the last part – the post-fixer. We need to implement it anyway.

Any UX, performance or logical issues that can't be solved generically by postfixer or are not easy enough, are out of scope. For instance:
- merging two lists
- fixing list after indentation changes
- pasting lists
- collaboration.

We should implement now at least some basic tests for these scenarios. If the post-fixer won't be able to handle them in a reasonable (from UX perspective) way, we can skip them for now and use them as an input when working on the **(a)** and **(b)**.

arkflpc:

Internal (list): Adds postfixer that makes sure that all items in a single list have the same start, reversed and style properties. Closes #11167.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/src/documentlist/documentlistediting.js
+++ b/packages/ckeditor5-list/src/documentlist/documentlistediting.js
@@ -448,12 +448,7 @@ function modelChangePostFixer( model, writer, emitter ) {
 			findAndAddListHeadToMap( entry.position, itemToListHead );
 		}
 		// Changed list item indent or type.
-		else if (
-			entry.type == 'attribute' && (
-				entry.attributeKey == 'listIndent' ||
-				emitter._sameListDefiningAttributes.includes( entry.attributeKey )
-			)
-		) {
+		else if ( entry.type == 'attribute' && entry.attributeKey.startsWith( 'list' ) ) {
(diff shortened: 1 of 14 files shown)


Reinmar:
I found two scenarios that may need our attention:

*   Indenting a list to a sub list: The properties of the items being indented are not reset.
    *   For list style this is not a big deal, although, it works differently than in previous lists. 
    *   However, for reversed option and, most importantly, the start index, this is very visible and a significant UX issue.
    *   This deserves a followup ticket as this will not be handled by a post-fixer.
*   When deleting a block that separates two differently styled lists, these lists will not be merged.
    *   This worked in the previous list feature as well.
    *   Let's try to cover it in this ticket.

PR #11338, opened 2022-02-25
The default configuration options for `ImageStyle` plugin are not properly cloned

### LINKED ISSUE #11328: The default configuration options for `ImageStyle` plugin are not properly cloned

It's difficult for me right now to prepare a few short steps to reproduce this bug, so let me just describe the problem.

The problem occurred during updating CKE5-related dependencies in the `ckeditor5-angular` repository, because one test started to fail in this line: https://github.com/ckeditor/ckeditor5-angular/blob/master/src/ckeditor/ckeditor.component.spec.ts#L324.

This test fails because unexpectedly the second component in this test also started to emit an error, but it shouldn't. The reason for this is that since CKE5 v29 we have an `ImageInline` plugin and the default options for `ImageStyle` have changed. It looks like they are not properly cloned when they are created, but they share the same arrays as `modelElements` in each editor, i.e. https://github.com/ckeditor/ckeditor5/blob/f52226ff5f6015c8736255debf1e39c2e0bb1ed3/packages/ckeditor5-image/src/imagestyle/utils.js#L47

### ✔️ Expected result

Editors do not share common properties.

### ❌ Actual result

Editors share common properties from `ImageStyle` plugin.

### ❓ Possible solution

Make sure that default options are properly cloned.

## 📃 Other details

* First affected CKEditor version: v29

---

psmyrek:

Fix (image): Always create new instances of the default options for the `ImageStyle` plugin. Closes #11328.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-image/package.json
+++ b/packages/ckeditor5-image/package.json
@@ -44,6 +44,7 @@
     "@ckeditor/ckeditor5-undo": "^32.0.0",
     "@ckeditor/ckeditor5-upload": "^32.0.0",
     "@ckeditor/ckeditor5-utils": "^32.0.0",
+    "@ckeditor/ckeditor5-watchdog": "^32.0.0",
     "@ckeditor/ckeditor5-widget": "^32.0.0",
     "webpack": "^5.58.1",
     "webpack-cli": "^4.9.0"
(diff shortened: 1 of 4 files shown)


pomek (approved):
LGTM.

PR #11388, opened 2022-03-02
Introduced Marker#getData(). Improved marker-related change detection in Differ.

ma2ciek:
Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Other: Introduced `Marker#getData()`. Improved marker-related change detection in Differ. Closes https://github.com/ckeditor/ckeditor5/issues/11389

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/model/differ.js
+++ b/packages/ckeditor5-engine/src/model/differ.js
@@ -337,7 +337,8 @@ export default class Differ {
 	 *
 	 * * model structure changes,
 	 * * attribute changes,
-	 * * changes of markers which were defined as `affectsData`.
+	 * * changes of markers which were defined as `affectsData`,
+	 * * changes of markers `affectsData` property.
 	 *
 	 * @returns {Boolean}
 	 */
--- a/packages/ckeditor5-engine/src/model/differ.js
+++ b/packages/ckeditor5-engine/src/model/differ.js
@@ -60,7 +60,7 @@ export default class Differ {
 		 * The keys of the map are marker names.
 		 * The values of the map are objects with the following properties:
 		 * - `oldMarkerData`,
-		 * - `newMarkerData`,
+		 * - `newMarkerData`.
 		 *
 		 * @private
 		 * @type {Map.<String, Object>}
@@ -331,7 +331,7 @@ export default class Differ {
 	 * * model structure changes,
 	 * * attribute changes,
 	 * * changes of markers which were defined as `affectsData`,
-	 * * changes of markers `affectsData` property.
+	 * * changes of markers' `affectsData` property.
 	 *
 	 * @returns {Boolean}
 	 */
--- a/packages/ckeditor5-engine/src/model/differ.js
+++ b/packages/ckeditor5-engine/src/model/differ.js
@@ -252,10 +252,8 @@ export default class Differ {
 	 * Buffers a marker change.
 	 *
 	 * @param {String} markerName The name of the marker that changed.
-	 * @param {module:engine/model/markercollection~MarkerData} oldMarkerData Marker range before the change
-	 * or `null` if the marker has just been created.
-	 * @param {module:engine/model/markercollection~MarkerData} newMarkerData Marker range after the change
-	 * or `null` if the marker was removed.
+	 * @param {module:engine/model/markercollection~MarkerData} oldMarkerData Marker data before the change.
+	 * @param {module:engine/model/markercollection~MarkerData} newMarkerData Marker data after the change.
 	 */
 	bufferMarkerChange( markerName, oldMarkerData, newMarkerData ) {
 		const buffered = this._changedMarkers.get( markerName );
--- a/packages/ckeditor5-engine/src/model/differ.js
+++ b/packages/ckeditor5-engine/src/model/differ.js
@@ -58,8 +58,11 @@ export default class Differ {
 		 * A map that stores all changed markers.
 		 *
 		 * The keys of the map are marker names.
-		 * The values of the map are objects with the `oldRange` and `newRange` properties. They store the marker range
-		 * state before and after the change.
+		 * The values of the map are objects with the following properties:
+		 * - `oldRange`,
+		 * - `newRange`,
+		 * - `affectsData`,
+		 * - `changesDataAffecting`.
 		 *
 		 * @private
 		 * @type {Map}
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-engine/tests/model/markercollection.js
```diff
@@ -53,7 +53,14 @@ describe( 'MarkerCollection', () => {
 			expect( marker.managedUsingOperations ).to.be.false;
 			expect( marker.affectsData ).to.be.false;
 			expect( marker.getRange().isEqual( range ) ).to.be.true;
-			sinon.assert.calledWithExactly( markers.fire, 'update:name', result, null, range );
+
+			expect( markers.fire.firstCall.args ).to.deep.equal( [
```
ma2ciek, line 57: I tried to write a test with `sinon.assert` +  `sinon.match()` but `sinon.match()` has a bug when an object has circular refs.

packages/ckeditor5-engine/src/model/markercollection.js
```diff
@@ -418,6 +429,19 @@ class Marker {
...
+	/**
+	 * Gets the current marker's data.
+	 *
+	 * @returns {module:engine/model/markercollection~MarkerData}
+	 */
+	getData() {
+		return {
+			range: this.getRange(),
+			affectsData: this.affectsData,
+			managedUsingOperations: this.managedUsingOperations
```
ma2ciek, line 441: Not 100% sure about this property being here, but I can remove it if it does not make sense.

packages/ckeditor5-engine/src/model/markercollection.js
```diff
@@ -418,6 +429,19 @@ class Marker {
 		return this._affectsData;
 	}
 
+	/**
+	 * Gets the current marker's data.
+	 *
+	 * @returns {module:engine/model/markercollection~MarkerData}
+	 */
+	getData() {
```
ma2ciek, line 437: This method may be enhanced in the future and expose things like marker attributes for the differ.

packages/ckeditor5-engine/src/model/differ.js
```diff
@@ -249,23 +252,29 @@ export default class Differ {
...
 
 		if ( !buffered ) {
 			this._changedMarkers.set( markerName, {
-				oldRange,
-				newRange,
-				affectsData
+				oldRange: oldMarkerData.range,
+				newRange: newMarkerData.range,
+				affectsData: oldMarkerData.affectsData,
+				changesDataAffecting: oldMarkerData.affectsData !== newMarkerData.affectsData
```
ma2ciek, line 268: I didn't want to change this API now to limit the number of changes, but perhaps it would make sense to have the following one:

```typescript
{
	oldMarkerData: MarkerData;
	newMarkerData: MarkerData;
	changesDataAffecting: Boolean;
}
```

Or maybe even:

```typescript
{
	oldMarkerData: MarkerData;
	newMarkerData: MarkerData;
}
```

I was thinking that changing the `affectsData` twice should also produce the event, but it seems to be a super edge case (IDK if we have anywhere even a case of changing marker from affecting data to not affecting). So perhaps we can just check the first marker data from the first change with the last marker data from the last change for the same marker to determine changes.

packages/ckeditor5-engine/src/model/markercollection.js
```diff
@@ -418,6 +429,19 @@ class Marker {
 		return this._affectsData;
 	}
 
+	/**
+	 * Gets the current marker's data.
```
scofalik, line 433:
```suggestion
	 * Returns the marker data (properties defining the marker).
```

scofalik (changes requested):

packages/ckeditor5-engine/src/model/document.js
```diff
@@ -157,14 +157,20 @@ export default class Document {
...
 		// are modified using `model.markers` collection, not through `MarkerOperation`).
-		this.listenTo( model.markers, 'update', ( evt, marker, oldRange, newRange ) => {
+		this.listenTo( model.markers, 'update', ( evt, marker, oldRange, newRange, oldMarkerData ) => {
 			// Whenever marker is updated, buffer that change.
-			this.differ.bufferMarkerChange( marker.name, oldRange, newRange, marker.affectsData );
+			this.differ.bufferMarkerChange( marker.name, oldMarkerData, { ...marker.getData(), range: newRange } );
 
 			if ( oldRange === null ) {
 				// If this is a new marker, add a listener that will buffer change whenever marker changes.
 				marker.on( 'change', ( evt, oldRange ) => {
```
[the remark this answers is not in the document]
ma2ciek: I was thinking also about that `change` event. We'll need to change the way how it is delegated as the default delegation won't make sense from now as we want to pass the whole data object there.

packages/ckeditor5-engine/src/model/differ.js
ma2ciek: Fixed.

ma2ciek:
Some tests are still failing and I'm trying to figure it out.

packages/ckeditor5-engine/src/model/document.js
```diff
@@ -157,14 +157,20 @@ export default class Document {
 		// Buffer marker changes.
 		// This is not covered in buffering operations because markers may change outside of them (when they
 		// are modified using `model.markers` collection, not through `MarkerOperation`).
-		this.listenTo( model.markers, 'update', ( evt, marker, oldRange, newRange ) => {
+		this.listenTo( model.markers, 'update', ( evt, marker, oldRange, newRange, oldMarkerData ) => {
 			// Whenever marker is updated, buffer that change.
-			this.differ.bufferMarkerChange( marker.name, oldRange, newRange, marker.affectsData );
+			this.differ.bufferMarkerChange( marker.name, oldMarkerData, { ...marker.getData(), range: newRange } );
```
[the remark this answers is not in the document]
ma2ciek: The `update` event fires before the change when the marker is being deleted. And I tried to change it quickly but it turned out to be not so trivial. I'd need to mock `oldMarker._liveRange` during the marker removal.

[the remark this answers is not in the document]
ma2ciek: I can report it as well in the follow-up

packages/ckeditor5-engine/src/model/document.js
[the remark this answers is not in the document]
ma2ciek: Reported in #11415.

ma2ciek:
A follow-up for the `Differ#getChangedMarkers`: #11416

scofalik (approved):
Two minor things that I already fixed.

PR #11425, opened 2022-03-07
Add insertObject() function.

### LINKED ISSUE #11198: [Document lists] Make sure that lists isn’t split when making certain changes inside it
Scenarios (some examples, there's definitely more of those):

*   Widget type around, `insertParagraph`, `insertTable`, pasting images into empty list items, etc
*   Inline image ↔ block image, moving list attributes to e.g. the paragraph is created.
*   A scenario in handling Backspace:

```
* li bar
* li [block image]
* li foo

+ backspace

expected:

* li bar
* li []
* li foo

actual:

* li bar
p []
* li foo
```

Idea: Make paragraph inherit attributes allowed by the schema of the block/element that was removed.

*   This could be implemented directly in `deleteContent()`.
*   And apply only in cases where we create some new element (e.g. merging one block into the preceding one doesn’t count).
*   Do we need a flag? Maybe it should always be true?
*   But then, what about conflicting attributes (allowed on many elements)?
    *   GHS: `htmlAttributes`, and specific ones: `htmlImgAttributes`
*   Introduce `copyOnReplace` in `schema.setAttributeProperties()`
    *   And mark `list*` attributes

CatStrategist:

Feature (engine): Added a new function `insertObject()` to the `Model` for inserting elements defined as objects by schema into a model.

Feature (engine): Added a new function `setAllowedAttributes()` to the `Schema` that validates attributes if they are allowed on given element before setting them.

Feature (engine): Added a new function `getAttributesWithProperty()` to the `Schema` that retrieves attributes from a node which have given property.

Feature (media-embed): Added optional `findOptimalPosition` parameter to `insertMedia()` function that allows for inserting `media` element without breaking content.

Feature (paragraph): Added optional `options.attributes` parameter to `InsertParagraph` command that allows setting attributes on created paragraph.

Fix (list): Fixed fixing indent of inserted content into document list in `documentlistediting` fixer.

Internal (engine): `findOptimalPosition()` is now also included in engine for internal use. 

Internal (table, page-break, horizontal-line, media-embed, html-embed, image): `table`, `pageBreak`, `horizontalLine`, `media`, `imageBlock`,`imageInline` elements are now inserted with `insertObject()` function instead of `insertContent()`. Closes #11198.

---

### Additional information

Changes to deletecontent, widgettypearound that handles copyOnReplace attributes' property are not described above. Changes to moving findOptimalPosition to engine not described yet.


packages/ckeditor5-image/src/imageutils.js
```diff
@@ -84,6 +84,14 @@ export default class ImageUtils extends Plugin {
...
 		const selection = model.document.selection;
 
+		let insertionContext;
+
+		if ( imageType === 'imageInline' ) {
+			insertionContext = model.document.selection.getSelectedElement();
+		} else {
+			insertionContext = model.document.selection.anchor.parent;
+		}
+
```
Reinmar, line 94: imageType is calculated below, so the order should be different

packages/ckeditor5-image/src/imageutils.js
```diff
@@ -103,13 +103,17 @@ export default class ImageUtils extends Plugin {
...
+				{
+					setSelection: 'on',
+					// If we want to insert a block image (for whatever reason) then we don't want to split text blocks.
+					// This applies only when we don't have the selectable specified (i.e., we insert multiple block images at once).
+					findOptimalPosition: !selectable && imageType != 'imageInline'
+				}
+			);
 
 			// Inserting an image might've failed due to schema regulations.
 			if ( imageElement.parent ) {
```
CatStrategist, line 119: Should this condition be also included in setSelection inside insertObject?

packages/ckeditor5-widget/src/insertobject.js
```diff
@@ -0,0 +1,100 @@
...
+		if ( optionsWithDefaults.findOptimalPosition && model.schema.isBlock( object ) ) {
+			const range = findOptimalInsertionRange( selection, model );
+			insertionSelection = writer.createSelection( range, offset );
+		}
+
+		// Get and set attributes
+		const firstSelectedBlock = selection.getSelectedBlocks().next().value;
+		const attributesToCopy = model.schema.getAttributesWithProperty( firstSelectedBlock, 'copyOnReplace', true );
+
+		writer.setAttributes( attributesToCopy, object );
```
niegowski, line 52: This should be set on the block element (so for an inline image that requires auto paragraphing those attributes should be set on the paragraph that wrapped an inline image.

Also attributes should be verified if are applicable to a specified object (according to the schema).

packages/ckeditor5-widget/src/insertobject.js
```diff
@@ -0,0 +1,100 @@
...
+
+		// Autoparagraph
+		if ( !model.schema.checkChild( insertionSelection.anchor.parent, object ) && model.schema.checkChild( 'paragraph', object ) ) {
+			const paragraph = writer.createElement( 'paragraph', attributesToCopy );
+
+			if ( !insertionSelection.isCollapsed ) {
+				model.deleteContent( insertionSelection, { doNotAutoparagraph: true } );
+			}
+
+			writer.insert( paragraph, insertionSelection.anchor );
```
niegowski, line 62: I would wrap the object with this paragraph and use the paragraph with the inline object as an object to be inserted. On the new object (the paragraph) we need to set those attributes that should persist.

packages/ckeditor5-widget/src/insertobject.js
```diff
@@ -0,0 +1,100 @@
...
+
+			insertionSelection = writer.createPositionAt( paragraph, 0 );
+		}
+
+		const affectedRange = model.insertContent( object, insertionSelection, offset, options );
```
niegowski, line 67:
```suggestion
		const affectedRange = model.insertContent( object, insertionSelection, offset );
```

packages/ckeditor5-engine/src/model/utils/insertobject.js
```diff
@@ -0,0 +1,227 @@
...
+ * * When `'after'`, the document selection will move to the closest text node after the inserted object. If there is no
+ * such text node, a paragraph will be created and the document selection will be moved inside it.
+ * @returns {module:engine/model/range~Range} A range which contains all the performed changes. This is a range that, if removed,
+ * would return the model to the state before the insertion. If no changes were preformed by `insertObject()`, returns a range collapsed
+ * at the insertion position.
+ */
+export default function insertObject( model, object, selectable, placeOrOffset, options = {} ) {
+	if ( !model.schema.isObject( object ) ) {
+		/**
+		 * Tried to insert an element by {@link module:engine/model/utils/insertobject insertObject()} function
```
oleq, line 50: Worth mentioning that `insertContent()` is the right tool for that.

packages/ckeditor5-engine/tests/model/utils/insertobject.js
```diff
@@ -0,0 +1,775 @@
...
+		model.schema.extend( '$text', { allowIn: 'span' } );
+	} );
+
+	describe( 'handled element types', () => {
+		it( 'should not throw an incorrect type of element to insert error if element is an object', () => {
+			const widget = new Element( 'blockWidget', [], [] );
+
+			expect( () => {
+				insertObject( model, widget );
+			} ).to.not.throw( CKEditorError, /insertobject-todo/ );
```
oleq, line 56: TODO

packages/ckeditor5-engine/tests/model/utils/insertobject.js
```diff
@@ -0,0 +1,775 @@
...
+
+		it( 'should throw an error if element is not an object', () => {
+			const paragraph = new Element( 'paragraph', [], [ new Text( 'bar' ) ] );
+
+			expectToThrowCKEditorError( () => insertObject( model, paragraph ), 'insertobject-element-not-an-object' );
+		} );
+	} );
+
+	describe( 'insertion selection', () => {
+		describe( 'with findOptimalSelection', () => {
```
oleq, line 67: findOptimalPosition?

packages/ckeditor5-engine/tests/model/utils/insertobject.js
```diff
@@ -0,0 +1,775 @@
...
+				insertObject( model, widget, undefined, undefined, { findOptimalPosition: 'auto' } );
+
+				const selectableArg = insertContentSpy.getCall( 0 ).args[ 1 ];
+
+				expect( selectableArg ).to.not.equal( model.document.selection );
+				expect( selectableArg.anchor.path ).to.deep.equal( [ 0 ] );
+			} );
+		} );
+
+		describe( 'without findOptimalSelection', () => {
```
oleq, line 108: findOptimalPosition?

packages/ckeditor5-paragraph/src/insertparagraphcommand.js
```diff
@@ -37,11 +37,17 @@ export default class InsertParagraphCommand extends Command {
 	 */
 	execute( options ) {
 		const model = this.editor.model;
+		const attributes = options.attributes;
```
oleq, line 40: Missing docs for this option.

oleq:
I guess pasting a widget on a fake typearound selection caret should also respect list attributes. The paste should be in the same list item.

oleq (approved):
LGTM 👍. Waiting for feedback from the QA team.

PR #11429, opened 2022-03-08
Certain markup breaks the Source Editing mode

### LINKED ISSUE #10703: Certain markup breaks the Source Editing mode

1. Go to https://ckeditor.com/docs/ckeditor5/latest/features/general-html-support.html#demo,
2. In the **Source Editing** mode paste in the  `<dir>` into the editor,
3. **Switch back** to a regular mode.

### ✔️ Expected result

Editor parses markup and shows the results.

### ❌ Actual result

An error is thrown in the console and the Source Editing mode is not allowing to change editor's content.

```
Uncaught TypeError: Cannot read properties of null (reading 'getAttribute')
    at setAttributeOnItem (writer.js:1489)
    at Writer.setAttribute (writer.js:380)
    at preserveElementAttributes (image.js:102)
    at UpcastDispatcher.dispatcher.on.priority (image.js:90)
    at UpcastDispatcher.fire (emittermixin.js:199)
    at UpcastDispatcher._convertChildren (upcastdispatcher.js:282)
    at UpcastDispatcher.<anonymous> (upcasthelpers.js:783)
    at UpcastDispatcher.fire (emittermixin.js:199)
```

### ❓ Possible solution

Initial check points to the image integration plugin in ckeditor5-html-support.

psmyrek:

Fix (html-support): Skip inline image upcast conversion inside not supported element. Closes #10703.

### DIFF
--- a/packages/ckeditor5-html-support/src/integrations/image.js
+++ b/packages/ckeditor5-html-support/src/integrations/image.js
@@ -84,6 +84,10 @@ export default class ImageElementSupport extends Plugin {
 function viewToModelImageAttributeConverter( dataFilter ) {
 	return dispatcher => {
 		dispatcher.on( 'element:img', ( evt, data, conversionApi ) => {
+			if ( !data.modelRange ) {
+				return;
+			}
+
 			const viewImageElement = data.viewItem;
 			const viewContainerElement = viewImageElement.parent;
 
--- a/packages/ckeditor5-html-support/tests/integrations/image.js
+++ b/packages/ckeditor5-html-support/tests/integrations/image.js
@@ -1221,6 +1221,25 @@ describe( 'ImageElementSupport', () => {
 				'<p><img src="/assets/sample.png"></p>'
 			);
 		} );
+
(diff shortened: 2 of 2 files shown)


psmyrek:
It was quite difficult for me to find out what is going on here. Below there are some notes I was able to find out while debugging this problem:
- The `DataFilter#_registerElementHandlers()` registers converters based on the type of a given element. But if a given element is both an object and an inline type, then `$text` will not be extended with new attributes introduced by this element. Is this approach correct? Probably yes, as they will be stored in the `htmlAttributes` attribute, but it was a surprise to me a bit.
- Initially, I wanted to fix this issue in another way. If an element is an object and an inline type (there are several such elements, i.e. `htmlImg`), then we should additionally register attribute converters by calling the `DataFilter#_registerInlineElement()` for this element. This, however, would not be enough because the upcast converter from image integration is called before the default converter for inline elements, so the one from the image integration had to be reordered with `'low' - 1` priority. In the end, I decided that it is too complicated and the current fix in this PR seems to be enough.

PR #11441, opened 2022-03-09
Introduce lock mechanism for Editor#isReadOnly

### LINKED ISSUE #10496: Editor#isReadOnly should work like disabling commands (blockable multiple times by different features)

At the moment, `Editor#isReadOnly` is a simple property that can be set to either `true` or `false`. This is limiting and leads to incorrect code.

Consider two mechanisms (features), that may want to set editor to readonly mode. At the beginning, both of them may decide that "there's some kind of error" and switch editor to read-only. But then, one problem is solved, so one of the mechanisms switches back `isReadOnly` to `true`. This is incorrect, because the other mechanism still wants to block the editor.

This is the same situation as with commands, where multiple features may want to set a command as disabled.

My proposal is to:

*   change `Editor#isReadOnly` to ... wait for it ... readonly property.
*   add `Editor#setReadOnly( id:String|Symbol )` and `Editor#unsetReadOnly( id )` or (`unlockReadOnly( id )`) -- similarly as in commands,
*   fix all places where we set `Editor#isReadOnly`.

One note is that `Editor#isReadOnly` could also be directly linked with document editing permission. Maybe document editing permission should be the only thing that actually changes `Editor#isReadOnly` and it should be dynamically affected by other mechanisms/features? However, I don't think that this really makes sense. For example, consider losing connection to Cloud Services. In such situation, we switch the editor to read-only. I don't think that it would make sense to switch user permission to "cannot edit" in this situation.

ma2ciek:

Feature (core): Introduced locking mechanism for `Editor#isReadOnly`. Read-only mode can now be separately enabled and disabled by multiple features which allows for proper control without conflicts between features. Closes #10496.

Other (core): `Editor#isReadOnly` is now a read-only property.

MAJOR BREAKING CHANGE: The `Editor#isReadOnly` property is now read-only. From now, this property is controlled by `Editor#enableReadOnlyMode( lockId )` and `Editor#disableReadOnlyMode( lockId )`, which allow controlling the `Editor#isReadOnly` state by more than one feature without collisions.

---

### Additional information


pomek:
This major breaking change breaks our integrations that set `editor.isReadOnly=true`.

https://github.com/ckeditor/ckeditor5-react/blob/dc3093c98ef566a0a422413e326e4bd086ecd54a/src/ckeditor.jsx#L63-L65

Now, integrations have to support both forms - the previous (as mentioned above), and the new one (that the PR adds).

ma2ciek:
Yes... Perhaps we can still support the old behavior and when encountering this kind of assignments we can try to set/remove a custom lock instead. I'll research it tomorrow.

ma2ciek:
During a recent discussion with @scofalik we decided that we can do a breaking change and then release major releases in our integrations (the BC in these integrations will be the fact that they will be working only with CKEditor 5 v34+), so they will not have to support two versions at the same time.

Introducing a lock mechanism looks pretty bad, imagine this situation:

```
editor.isReadOnly = false; // clears a custom assignment lock on the `isReadOnly` property.

editor.isReadOnly // still `true` as other features may still have their locks on the `isReadOnly` property.
```

pomek:
I have nothing to do with the API, as you know better what you want to achieve. I agree that the proposed method is much better than a single property. 

However, **we cannot assume** people update their builds. They don't have to, and it's our responsibility to think about it. So, integrations must detect (somehow) a version of the CKEditor 5 API and, based on that, use old or new API.

In the worst scenario, if we cannot support both APIs, we need to release new versions of integrations with adjusted API.

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -148,15 +148,23 @@ export default class Editor {
 		 * In read-only mode the editor {@link #commands commands} are disabled so it is not possible
 		 * to modify the document by using them. Also, the editable element(s) become non-editable.
 		 *
-		 * In order to make the editor read-only, you can set this value directly:
+		 * In order to make the editor read-only, you need to call the {@link #setReadOnlyLock} method:
```
scofalik, line 151:
```suggestion
		 * In order to make the editor read-only, you need to call the {@link #setReadOnly}.
		 * As a parameter, this method takes a unique id. The id is used to clear the lock later on.
```

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -227,6 +235,97 @@ export default class Editor {
...
+	 * When at least one feature sets the read-only lock, then the {@link #isReadOnly `isReadOnly` property} will be set to `true`.
+	 * Otherwise it is set to `false`.
+	 *
+	 * 		editor.setReadOnlyLock( 'my-feature-lock-id' );
+	 *
+	 * You can remove the lock using the {@link #clearReadOnlyLock} method afterwards.
+	 *
+	 * 		editor.clearReadOnlyLock( 'my-feature-lock-id' );
+	 *
+	 * It is possible to pass the additional argument - `value` to determine if the lock should be set or removed.
```
scofalik, line 269:
```suggestion
     * It is possible to pass the additional argument - `value` to determine if the lock should be set or removed.
     * Passing `false` works the same way as calling `clearReadOnlyLock()`.
```

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -227,6 +235,97 @@ export default class Editor {
...
+	 *
+	 * 		let shouldMyFeatureSetReadOnlyMode = false;
+	 * 		editor.setReadOnlyLock( 'my-feature-lock-id', shouldMyFeatureSetReadOnlyMode );
```
scofalik, line 272:
```suggestion
	 * 		myFeature.on( 'change:isConnected', () => {
	 * 			editor.setReadOnlyLock( 'my-feature-lock-id', myFeature.isConnected );
	 * 		} );
```

scofalik (changes requested):
I don't like current method names, I propose:

* `setReadOnly( id )` and `clearReadOnly( id )`, or
* `enableReadOnly( id )` and `disableReadOnly( id )`.

I understand that `setReadOnly( 'foo' )` looks like we are setting the property value to `'foo'`, hence the second proposal.

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -227,6 +235,97 @@ export default class Editor {
...
+		 */
+		this._readOnlyStack = new Set();
+	}
+
+	/**
+	 * Returns `true` if some feature with the given lock ID ensures the editor being in the read-only state.
+	 *
+	 * @param {String} lockId The lock ID for setting the editor to the read-only state.
+	 */
+	hasReadOnlyLock( lockId ) {
```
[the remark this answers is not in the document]
ma2ciek: From what I remember there is no place where this method is used in the source code, but as for the tests there is currently no other way to write it without using the `hasReadOnlyLock()`. I was thinking about just using the `Editor.isReadOnly` in manual tests but then these buttons will look broken :sweat\_smile:

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -148,15 +148,23 @@ export default class Editor {
...
+		 * 		editor.clearReadOnlyLock( 'feature-lock-id' );
+		 *
+		 * @readonly
 		 * @observable
 		 * @member {Boolean} #isReadOnly
 		 */
 		this.set( 'isReadOnly', false );
 
+		// Stop the event to not allow changing the property by hand.
+		this.on( 'set:isReadOnly', stopReadOnlyChangeEvent, { priority: 'highest' } );
```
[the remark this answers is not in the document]
ma2ciek: I was thinking about it and I came to the same conclusions. I don't think it's a good idea to make it a getter as the integration with the \`ObservableMixin\` will be messy and allowing detecting changes of that property is one of the main use cases so I can't imagine not supporting it :)

packages/ckeditor5-core/src/editor/editor.js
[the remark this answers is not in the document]
ma2ciek: So without using `hasReadOnlyLock` we would have to either use internals, `Editor.isReadOnly`, I don't have other ideas.

scofalik:
LGTM except of the new method names. 

> I propose:
> 
> *   `setReadOnly( id, value )` and `clearReadOnly( id )`, or
> *   `enableReadOnly( id, value )` and `disableReadOnly( id )`.

ma2ciek:
@ckeditor/qa-team  Can I ask you to test this PR? I changed also the `ckeditor5-core/tests/manual/readonly` test to allow testing setting a few locks at the same time.

Unfortunately, there are a lot of places where we use or observer `editor.isReadOnly` and set this property via `editor.setReadOnly()`

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -229,6 +222,113 @@ export default class Editor {
...
+	 * 		editor.clearReadOnly( 'my-feature-lock-id' );
+	 *
+	 * It is possible to pass the additional argument - `value` to determine if the lock should be set or removed.
+	 * Passing `false` works the same way as calling {@link #clearReadOnly `clearReadOnly()`}.
+	 *
+	 * 		myFeature.on( 'change:isConnected', () => {
+	 * 			editor.setReadOnly( 'my-feature-lock-id', myFeature.isConnected );
+	 * 		} );
+	 *
+	 * @param {String|Symbol} lockId The lock ID for setting the editor to the read-only state.
```
scofalik, line 277:
```suggestion
	 * @param {String|Symbol} lockId Unique ID for setting the editor to the read-only state.
```

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -229,6 +222,113 @@ export default class Editor {
...
+			throw new CKEditorError( 'editor-read-only-lock-id-invalid', null, { lockId } );
+		}
+
+		if ( !this._readOnlyLocks.has( lockId ) ) {
+			return;
+		}
+
+		this._readOnlyLocks.delete( lockId );
+
+		if ( this._readOnlyLocks.size == 0 ) {
```
scofalik, line 326:
```suggestion
		if ( this._readOnlyLocks.size === 0 ) {
```

ma2ciek: Can it be something else? :smile:

LukaszGudel:
I've tested this PR, and also CF branch related to this changes and everything seems fine. The lock mechanism is working correctly. I've also tested CF features where read-only mode can be set and that was also working correctly :+1:

scofalik:
Migration guide missing 😬

ma2ciek:
Can we add this guide as a follow-up? The DLL changes already introduce the migration guide for the next iteration and I'd be happy to not merge these files if I do not have to :)

docs/updating/migration-to-34.md
```diff
@@ -34,3 +34,34 @@ From now on, additional plugins will be required, when the following CKEditor 5
 	import RealTimeCollaborativeRevisionHistory from '@ckeditor/ckeditor5-real-time-collaboration/src/realtimecollaborativerevisionhistory';
 	import RevisionHistory from '@ckeditor/ckeditor5-revision-history/src/revisionhistory';
 	```
+
+### Changed mechanism for setting and clearing the editor read-only mode
+
+With this update, the previously editable {@link module:core/editor/editor~Editor#isReadOnly `editor.isReadOnly`} become a getter, manual setting `editor.isReadOnly` is no longer permitted.
```
scofalik, line 40:
```suggestion
With this update, {@link module:core/editor/editor~Editor#isReadOnly `editor.isReadOnly`} becomes a read-only property. Setting it manually is no longer permitted and will result in an error.
```

docs/updating/migration-to-34.md
```diff
@@ -34,3 +34,34 @@ From now on, additional plugins will be required, when the following CKEditor 5
...
+
+Changing the editor mode it now possible only via the lock mechanism, and thanks to that, various features that set the read-only mode will not conflict with themselves and will not have to know about each other.
```
scofalik, line 42:
```suggestion
Changing the editor mode it now possible only via dedicated methods that introduce a lock mechanism. Thanks to that, various features that set the read-only mode will not collide and will not have to know about each other. Basically, the editor will become editable again only if all features (that earlier set it to read-only) will allow for that.
```

docs/updating/migration-to-34.md
```diff
@@ -34,3 +34,34 @@ From now on, additional plugins will be required, when the following CKEditor 5
...
+	```js
+	// ❌ Old usage:
+	function makeEditorReadOnly() {
+		editor.isReadOnly = true;
+	}
+
+	function makeEditorEditable() {
+		editor.isReadOnly = false;
+	}
+	// ✅ New usage:
```
scofalik, line 57:
```suggestion
	
	// ✅ New usage:
```

docs/updating/migration-to-34.md
```diff
@@ -34,3 +34,34 @@ From now on, additional plugins will be required, when the following CKEditor 5
...
+
+### Changed mechanism for setting and clearing the editor read-only mode
+
+With this update, the previously editable {@link module:core/editor/editor~Editor#isReadOnly `editor.isReadOnly`} become a getter, manual setting `editor.isReadOnly` is no longer permitted.
+
+Changing the editor mode it now possible only via the lock mechanism, and thanks to that, various features that set the read-only mode will not conflict with themselves and will not have to know about each other.
+
+The lock mechanism makes the editor read-only if there is at least one feature lock, otherwise it makes the editing allowed.
+
+ The new methods on the `Editor` are called {@link module:core/editor/editor~Editor#enableReadOnlyMode `editor.enableReadOnlyMode( featureLockId )`}  and {@link module:core/editor/editor~Editor#clearReadOnlyMode `editor.clearReadOnlyMode( featureLockId )`}, which respectively set and clear the read-only lock on the editor.
```
scofalik, line 46:
```suggestion
The new methods on the `Editor` class are {@link module:core/editor/editor~Editor#enableReadOnlyMode `editor.enableReadOnlyMode( lockId )`}  and {@link module:core/editor/editor~Editor#disableReadOnlyMode `editor.disableReadOnlyMode( lockId )`}, which respectively enable and disable the read-only mode.

The lock mechanism makes the editor read-only if there is at least one lock. If all locks are removed, the content becomes editable again. Because each feature is responsible only for setting and removing its own lock, the features do not conflict each other. Earlier, such issues could result in the editor content being editable when it shouldn't.
```

docs/updating/migration-to-34.md
```diff
@@ -34,3 +34,34 @@ From now on, additional plugins will be required, when the following CKEditor 5
...
+	}
+	// ✅ New usage:
+	const myFeatureLockId = Symbol( 'my-feature' );
+
+	function makeEditorReadOnly() {
+		editor.enableReadOnlyMode( myFeatureLockId );
+	}
+
+	function makeEditorEditable() {
+		editor.clearReadOnlyMode( myFeatureLockId );
```
scofalik, line 65:
```suggestion
		editor.disableReadOnlyMode( myFeatureLockId );
```

scofalik (approved):
I reworded the migration guide a bit 😬

scofalik:
It's not at all from scratch. In one place it looks like a lot of changes but I just changed the order.

PR #11461, opened 2022-03-14
Allowed excluding certain elements from CSS reset

### LINKED ISSUE #11451: Research possibility to use donut CSS rule to exclude pieces of DOM from our CSS reset
The case is that someone may want to use their UI inside our UI. Their UI may expect no reset rules to be used.

How can they say that starting from some container our CSS reset rules should not apply?

oleq:

Feature (theme-lark): Implemented a `.ck-reset_all-excluded` CSS class that excludes certain elements from CSS reset. Closes #11451.

---

### Additional information

Demo in the manual test in the `theme-lark` package.

### DIFF
--- a/packages/ckeditor5-theme-lark/tests/manual/reset.html
+++ b/packages/ckeditor5-theme-lark/tests/manual/reset.html
@@ -1,0 +1,43 @@
+<h2>CSS reset exclusion</h2>
+
+<div class="ck ck-reset_all" style="border: 1px dotted gray; padding: 10px;">
+	<h3>A box with the <code>ck-reset_all</code> CSS class.</h3>
+	<p>Paragraph of text.</p>
+	<button type="button">Button</button>
+	<input type="text" value="Text of input" />
+	<div class="ck-rtl">
+		<p>.ck-rtl paragraph.</p>
+	</div>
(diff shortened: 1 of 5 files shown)


oleq:
@ckeditor/qa-team Could you please check if this change does not break the UI of the editor? 🙏 The scope is vast, any piece of UI could be affected and there's no easy way to confirm that other than going through all guides/features (also in CF) 😬.

dufipl:
@oleq
We have tested this branch on docs on Chrome, Firefox and Safari and it looks good. Also, it does not look like it is breaking CF manual tests or collaboration samples so it looks good for us 👍

Reinmar:
TODO @oleq add a note about this PR being a breaking change.

oleq:
> TODO @oleq add a note about this PR being a breaking change.

False alert. I thought using `.ck.ck-reset_all *` won't be possible but the I realized `.ck` stays in DOM so we're good. Sorry for the confusion.

Reinmar:
What about the `border-collapse` style? It seems that it somehow leaks from... something.

Reinmar:
Oleq explained to me that :point\_up: is a dev tools issue.

All looks good then :)

PR #11494, opened 2022-03-21
GHS should not convert already consumed inline elements

### LINKED ISSUE #11447: GHS converts elements that have been already converted by other plugins

Drupal wraps `` and `<drupal-media>` elements with `<a>` elements. In the CKEditor model, both of these elements are likely to contain `<caption>` model element. If the `` or `<drupal-media>` element is wrapped with a `<a>` element and contains a caption, it leads to the text inside `<caption>` getting wrapped with an empty `<a>` element.

Here's a failing test for this: https://github.com/lauriii/ckeditor5/commit/a8f8912e428a50daec8dae68c9dedbffa886a84c.

### ✔️ Expected result

The `href` and all other attributes are consumed from the `<a>` element, meaning that there shouldn't be any processing expected from GHS on that element.

The expected markup would be:
```
<a href="...">
	<drupal-media>
		<caption>
			...
		</caption>
	</drupal-media>
</a>
```

### ❌ Actual result
```
<a href="...">
	<drupal-media>
		<caption>
			<a>...</a>
		</caption>
	</drupal-media>
</a>
```

---

niegowski:

Fix (html-support): GHS should not convert already consumed inline elements. Closes #11447.

---

### Additional information


lauriii:
Thank you @niegowski for the PR! I tested this manually with Drupal, and this allows us to consume the `<a>` element in the plugin that provides support for linked media, and therefore resolves https://github.com/ckeditor/ckeditor5/issues/11447. 💯

lauriii:
Realized there's also a Link plugin specific part of this. The Link plugin is consuming the `href` attribute without consuming the tag: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-link/src/linkimageediting.js#L134.

niegowski:
> Realized there's also a Link plugin specific part of this. The Link plugin is consuming the `href` attribute without consuming the tag: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-link/src/linkimageediting.js#L134.

This might be related to #10800.

wimleers:
And it looks like a very tightly scoped bug fix for a edge case that simply was not covered before — with explicit test coverage. Looks _great_ 🤩

PR #11496, opened 2022-03-21
The view list split converter should not fire if change was already consumed

### LINKED ISSUE #11490: CKEditor 33.0.0 List plugin not compatible with Drupal Media
Drupal Media has following conversion:

Model:
```
<drupalMedia><caption>...</caption></drupalMedia>
```

Data:
```
<drupal-media data-caption></drupal-media>
```

This is not compatible with the list plugin. It could be that Drupal Media was never compatible with the List plugin but for some reason the `mapping-view-position-parent-not-found` exception was not thrown prior to 33.0.0. This seems to be caused by the `<caption>` not having a view element that it could be mapped to.

It could be possible to reproduce this with the `should allow nesting downcast conversion` CKEditor test when list plugin is enabled.

---

niegowski:

Fix (list): The view list split converter should not fire if the change was already consumed. Closes #11490.

---

### Additional information


lauriii:
Confirmed manually with Drupal that this PR addresses https://github.com/ckeditor/ckeditor5/issues/11490 🎉  Great work @niegowski! 👏

Issue #11460, opened 2022-03-14
Upgrade PostCSS to the latest version


pomek:
## Provide a description of the task

@dependabot explains why it is needed in [PRs](https://github.com/ckeditor/ckeditor5-dev/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed+postcss+author%3Aapp%2Fdependabot).

## Scope

* Upgrade `postcss` from 7 to 8.
	* See: https://github.com/postcss/postcss/releases/tag/8.0.0.
	* Adjust plugins in `ckeditor5-dev` repository.
	* Adjust plugins in `ckeditor5` repository (content styles).
* Make sure the following tasks still work:
	* Verify whether `ckeditor5-build-*` packages are built properly.
	* Verify whether manual tests work.
	* Verify whether automated tests work.
	* Verify whether DLLs work.
	* Verify whether documentation works.
	* Verify whether the content-styles extractor still works.
* PostCSS is used in many places. Make sure to update all of them:
	* Mentioned `ckeditor5` and `ckeditor5-dev` repositories.
	* [Package generator](https://github.com/ckeditor/ckeditor5-package-generator).
	* Online Builder and builds procured by the tool.
* ~Create a migration guide on why did we do it.~
	_This is not needed, it's not a major bump._

## Tips

Before starting anything, prepare test scenarios that you can follow before and after upgrading the environment. It will help understand issues you spot.

### Content-styles extractor

Our script implements a plugin that iterates over-produced CSS rules. We need to adjust its API. See: https://github.com/ckeditor/ckeditor5/commit/64ffb7b28a7b2a1971953290d945f21af2a1a464.

### Upgrading PostCSS plugins in `ckeditor5-dev`

In the past, I thought we could support both versions. It does not work as we cannot predict which PostCSS plugins will be installed. Hence, let's support only PostCSS@8: https://github.com/ckeditor/ckeditor5-dev/commit/a3174e6451b541071560a4596e3941e581e3b733.

pomek:
Changes in `ckeditor5-dev` are released: https://github.com/ckeditor/ckeditor5-dev/releases/tag/v30.0.0.

pomek:
CKEditor 5 on the `#master` branch uses PostCSS 8 too.

pomek:
Starting from the next release, Online Builder will produce builds that use PostCSS 8.

pomek:
All work done.

PR #11498, opened 2022-03-22
Bumped `postcss` to v8

przemyslaw-zan:

Other: CKEditor 5 uses PostCSS@8 now. Closes See #11460.

---

### Additional information

### DIFF
--- a/docs/installation/index.md
+++ b/docs/installation/index.md
@@ -43,7 +43,7 @@ Find out more about the {@link installation/advanced/plugins plugin development}
  * {@link updating/index Updating CKEditor 5} &ndash; Find out how to keep you installation up-to-date at all times.
  * {@link features/index Features} &ndash; Learn more about the CKEditor 5 features.
  * {@link examples/index Examples} &ndash; Try live demos of available predefined builds and custom solutions.
- * * {@link framework/guides/overview CKEditor 5 Framework} &ndash; Learn how to work with CKEditor 5 Framework, customize it, create your own plugins or custom editors, how to change the UI or even bring your own UI to the editor.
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -131,6 +131,9 @@
     "webpack": "^5.58.1",
     "wrapper-webpack-plugin": "^2.2.0-beta.0"
   },
+  "peerDependencies": {
+    "postcss": "^8.4.12"
+  },
```
pomek, line 136:
```suggestion
```

package.json
```diff
@@ -173,7 +176,7 @@
   },
   "resolutions": {
     "eslint-config-ckeditor5": "^4.0.0",
-    "postcss-loader": "^4.0.0",
+    "postcss-loader": "^4.3.0",
```
pomek, line 179:
```suggestion
    "postcss-loader": "^4.0.0",
```

The previous and proposed change will install the same package.

pomek (approved):
LGTM.

pomek:
We need a new ticket for updating/adjusting docs:

(1) Add the `postcss` dependency to all `npm install` commands:

```diff
npm install --save \
    @ckeditor/ckeditor5-vue2 \
    @ckeditor/ckeditor5-dev-webpack-plugin \
    @ckeditor/ckeditor5-dev-utils \
    postcss-loader@4 \
+   postcss@8 \
    raw-loader@4
```

(2) We need to create a migration guide that explains what to do after the migration to postcss@8.

package.json
```diff
@@ -114,7 +114,7 @@
     "mkdirp": "^1.0.4",
     "nyc": "^15.0.1",
     "popper": "^1.0.1",
-    "postcss": "^7.0.39",
+    "postcss": "^8.4.12",
```
pomek, line 117: `ckeditor5-dev-utils` requires the `postcss` library so we don't need it here.

PR #11510, opened 2022-03-23
Fixes column resizing not working in Safari.

mmotyczynska:

Fix (table-resize): fixes column resizing not working in Safari.

---

### Additional information

I changed the way the width of table in px is calculated. getComputedStyle for colgroup in Safari always returns 0px, so I changed the calculation to be based on tbody instead, which works fine in all main browsers.


mlewand:
@ckeditor/qa-team I don't have safari at hand, can any of you folks test whether this PR fixes the issue?

mlewand (approved):
LGTM

If I was doing this fix I would add a unit test where:
* I stub window.getComputedStyles() only for `colgroup` inside my table that would return 0 (emulating safari bug).
* Shortly document why it is stubbing it like that and reference the issue.

But the stub needs to return normal results for other calls (because it will be called with `tbody` after your fix).

This should cover this case without a need to having run it on Safari.

If you want to implement this unit test that would be nice addition. But if you skip it it's fine too. 🙂

dufipl:
During testing of this PR one Safari specific issue occurred. Arrow of balloon toolbar is blinking when the resized column is below:
blinking column resize

mlewand (changes requested):
The overall direction is fine, but there are multiple minor things to be ironed out.

Looks to me like you didn't verify whether the added unit test actually tests what it's supposed to test.

If you undo changes in the sources you'll see that your unit test passes even with the old implementation - implementation that doesn't work.

You can verify this by doing `git checkout 8bc2f4a2071e876d333fcde6bd656000d797b810 packages/ckeditor5-table-resize/src/tablecolumnresize/utils.js` in CKE5 directory (where `8bc2f4a20` is a parent commit of first commit in this PR).

packages/ckeditor5-table-resize/tests/tablecolumnresizer/utils.js
```diff
@@ -487,6 +433,67 @@ describe( 'TableColumnResize utils', () => {
...
+							</tr>
+						</tbody>
+					</table>
+				</figure>`
+			);
+
+			const table = editor.model.document.getRoot().getChild( 0 );
+			const getComputedStyleOriginal = window.getComputedStyle;
+
+			const modifiedGetComputedStyle = function( element ) {
```
mlewand, line 477: You could achieve the same using dedicated sinon API.

Like so:

```
			const getComputedStyleStub = sinon.stub( window, 'getComputedStyle' ).callThrough();

			// Emulate safari's bug.
			getComputedStyleStub.withArgs( sinon.match.has( 'localName', 'colgroup' ) ).returns( { width: '0px' } );
```

mlewand (commented):
Minor things to be cleaned up but the test does what it's supposed to do 👍

I'll push fixes in a second.

mlewand (approved):
Added corrections in ce4d65aba3

PR #11524, opened 2022-03-28
GHS should integrate with removing a native feature attribute removing

### LINKED ISSUE #10800: Unlinking linked inline images while GHS is enabled: wrapping <a> is impossible to remove

### LINKED ISSUE #10954: PoC of more precise consumable use in GHS and other features
PoC of a solution for #10800 and related problems.

Notes from recent discussions:

---

> *   \<a href> → \<a> should be consumed by LinkEditing plugin
>     *   This means that elementToAttribute() (upcast and two-way) will need to consume the element name if it was listed in the `view: { name: ... }` part.
> *   \<span style=x> → \<span>s are a specific case – the only one where we're not interested in the element itself when some things were already consumed on it. Thus, GHS should ignore such spans on which everything was already consumed.
> *   But why we have the \<strong> problem too?
> 
> PoC is needed:
> 
> *   Link problem: \[elementToAttribute()\] consume the element name
> *   Strong problem: \[GHS\] should check whether the element wasn't consumed already
> *   Span problem: \[GHS\] Thus, GHS should ignore such spans on which everything was already consumed.

niegowski:

Fix (html-support): Inline elements handled by an editor plugin should not be handled by the GHS. Closes #10800. Closes #10954.

Fix (html-support): Unlinking should remove a link even if there were some additional attributes handled by the GHS. See #10800.

Fix (engine): The `elementToAttribute()` upcast helper should consume an element while consuming its attribute. See #10800.


Reinmar:
I scanned the changes and they look good. I didn't test them manually yet (cc @ckeditor/qa-team).

The biggest problem that I have is with... the naming :rofl: `coupledAttribute` doesn't explain what kind of attribute it is and in what way they are coupled. To make it more explicit it'd have to become something like: `coupledFeatureTextAttribute` but it's ugly :D So if anyone can come up with something better, it'd be great.

The second thing is – how can a 3rd party addon developer specify a coupled attribute for their feature? Currently we keep these things only in `SchemaDefinitions` – I'd expect it to see also somewhere in the public API in the `DataFilter`.

packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -908,6 +909,15 @@ function prepareToAttributeConverter( config, shallow ) {
 		// It may happen that a converter will try to set an attribute that is not allowed in the given context.
 		// In such a situation we cannot consume the attribute. See: https://github.com/ckeditor/ckeditor5/pull/9249#issuecomment-815658459.
 		if ( attributeWasSet ) {
+			// Verify if the element itself wasn't consumed yet. It could be consumed already while upcasting some other attribute
+			// on the same element (for example <span class="big" style="color: red">foo</span>).
+			// We need to consume it so other features (especially GHS) won't try to convert it.
+			// Note that it's not tested by the other element-to-attribute converters whether an element was consumed before
+			// (in case of converters that the element itself is just a context and not the primary information to convert).
+			if ( conversionApi.consumable.test( data.viewItem, { name: true } ) ) {
```
scofalik, line 917: Do we need this `if`? Isn't it the same without it?

Or maybe `consume()` won't consume anything if at least one part is not consumable?

niegowski: > Or maybe consume() won't consume anything if at least one part is not consumable?

Exactly

packages/ckeditor5-html-support/src/converters.js
```diff
@@ -89,7 +89,17 @@ export function createObjectView( viewName, modelElement, writer ) {
 export function viewToAttributeInlineConverter( { view: viewName, model: attributeKey }, dataFilter ) {
 	return dispatcher => {
 		dispatcher.on( `element:${ viewName }`, ( evt, data, conversionApi ) => {
-			const viewAttributes = dataFilter._consumeAllowedAttributes( data.viewItem, conversionApi );
+			let viewAttributes = dataFilter._consumeAllowedAttributes( data.viewItem, conversionApi );
+
+			// Do not apply an attribute if the element itself is already consumed and there is no view attributes to store.
+			if ( !viewAttributes && !conversionApi.consumable.test( data.viewItem, { name: true } ) ) {
```
scofalik, line 95: It seems to me that the fragment from line 95 to 107 can be simplified to:

```
if ( !viewAttributes ) {
	conversionApi.consumable.consume( data.viewItem, { name: true } );
	
	return true;
}
```

Or am I missing something?

If there are no attributes, the converter makes sure that the element is consumed and that's it.

packages/ckeditor5-html-support/src/datafilter.js
```diff
@@ -106,8 +106,18 @@ export default class DataFilter extends Plugin {
 		*/
 		this._dataInitialized = false;
 
+		/**
+		 * Cached map of coupled attributes. The map is keyed by the feature attribute name
+		 * and coupled GHS attribute names are stored in the value array.
```
scofalik, line 111:
```suggestion
		 * Cached map of coupled attributes. Keys are the feature attributes names
		 * and values are arrays with coupled GHS attributes names.
```

scofalik: I am not sure if after my changes this is correct, but I think that simply writing "keys are, values are" is simpler to understand.

packages/ckeditor5-html-support/src/converters.js
niegowski: There is a case of an element that does not have any attributes but has to be converted to an empty attribute object to be preserved. So for example `<cite>` must not only be consumed but also converted to `htmlCite={}`.

niegowski: Ugh, I just noticed that there is a merge error.

packages/ckeditor5-html-support/src/datafilter.js
```diff
@@ -288,6 +301,83 @@ export default class DataFilter extends Plugin {
...
 
+	/**
+	 * Registers a model post-fixer that is removing coupled GHS attributes of inline elements. Those attributes
+	 * are removed if a coupled feature attribute is removed.
+	 *
+	 * For example while unlinking a text (by the remove of `linkHref` attribute):
+	 *
+	 *		<$text linkHref="foo" htmlA="{ ... }">bar</$text>
+	 *
+	 * The `htmlA` attribute would be left behind in the model and would cause GHS to generate an `<a>` element
```
scofalik, line 312:
```suggestion
	 * For example, consider following HTML:
	 *
	 *		<a href="foo.html" id="myId"></a>
	 *
	 * Which would be upcasted to following text node in the model:
	 *
	 *		<$text linkHref="foo.html" htmlA="{ id: 'myId' }">bar</$text>
	 *
	 * When the user removes the link from that text (using UI), only `linkHref` attribute would be removed:
	 *
	 *		<$text htmlA="{ id: 'myId' }">bar</$text>
	 *
	 * The `htmlA` attribute would stay in the model and would cause GHS to generate an `<a>` element.
	 * This is incorrect from UX point of view, as the user wanted to remove the whole link.
```

scofalik: More verbose example is needed to understand while this mechanism is necessary.

I am not sure if `htmlA` value is correct in the proposed example.

packages/ckeditor5-html-support/src/converters.js
scofalik: Yes, there is the same `if` twice there.

scofalik: Okay, now it probably makes more sense.

packages/ckeditor5-html-support/src/datafilter.js
```diff
@@ -288,6 +301,83 @@ export default class DataFilter extends Plugin {
...
+	 * that should not be there anymore because that text was supposed to be unlinked completely (not only a `href`).
```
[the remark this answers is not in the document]
scofalik: This line should be also included to remove in the previous suggestions.

FilipTokarski:
The bug with unlinking seems to be fixed now 👍

Mgsy:
LGTM :+1:

Reinmar:
We still need a written answer to questions raised by @scofalik, @niegowski :P

Reinmar:
Actually, my questions are still unanswered too: [https://github.com/ckeditor/ckeditor5/pull/11524#issuecomment-1082746504](https://github.com/ckeditor/ckeditor5/pull/11524#issuecomment-1082746504)

niegowski:
> The biggest problem that I have is with... the naming 🤣 `coupledAttribute` doesn't explain what kind of attribute it is and in what way they are coupled. To make it more explicit it'd have to become something like: `coupledFeatureTextAttribute` but it's ugly :D So if anyone can come up with something better, it'd be great.

This model attribute is not limited to text attributes registered by features, it's a pretty loose relation so I wanted to avoid too strict relation in the name. That relation is limited only to the case when some attribute is removed, then the GHS attribute should be removed too.

> The second thing is – how can a 3rd party addon developer specify a coupled attribute for their feature? Currently we keep these things only in `SchemaDefinitions` – I'd expect it to see also somewhere in the public API in the `DataFilter`.

I just forgot to add it in docs, it's registered by `DataSchema#registerInlineElement()`.

niegowski:
> May be stupid to ask it at this point, but... Did you try making this fix without using "coupledAttribute" mechanism? I dislike hardcoded relations like this one as they may lead to errors by omission, later in the project's life.
> 
> I wonder if this could be automated. Using the example with `<a>` tag. Let's say you have model and view:
> 
> ```
> <$text linkHref="foo.html" htmlA="{ id: 'myId' }">bar</$text>
> ```
> 
> ```
> <a href="foo.html" id="myId">bar</a>
> ```
> 
> When you remove `linkHref`, in downcast, you generate `<a href="foo.html">` and use it to unwrap the view.
> 
> Theoretically, at least using this example, you would be able to figure out that only GHS attributes are left and they should be removed as well. Of course, there is `<span>` case, where both core and GHS attributes are added to `<span>` tag and removing, for example, font size, would remove GHS attributes as well. Also there are cases where the whole tag is created and handled by GHS and removing one of attributes of such tag should not lead to removing the whole tag or other attributes (`<cite id="foo" class="bar"></cite>` and then remove `class` attribute _somehow_).
> 
> Still, I have a strong feeling that this should be supported by the engine, that we are missing something.
> 
> `<span>`s can be recognized too, because converters that create `<span>`s simply do not use element name, or do not care about element name. So, when they unwrap, they will not want to unwrap the tag.
> 
> OTOH, I also feel that in some cases it is the feature that knows the best what it wants to do. In the case of link feature, the feature knows that it wants to remove **everything** connected with `<a>` the whole tag and all attributes not matter how they ended up there. But of course, it does not feel good at all to make link feature know about `htmlA` attribute... But, again, I feel this is the same case for all other features too, like bold, italic, etc. You want the whole tag gone.
> 
> But then what with `<span style="font-weight:bold;" id="foo"></span>` case? Does it become `<strong id="foo">\</strong\>`? 🤔
> 
> Anyway, the changes look fine, overall, but I have this feeling that this should simply work out of the box and coupling should not be needed.

The biggest problem with automatic resolving the connection between view elements and model attributes is that we don't have there any mapping. We can't tell whether an attribute that was removed, was the one responsible for the element itself, or it was an additional data for it. There might be a way by triggering some fake conversion to check what element is generated but still, it could be problematic. We decided to go with the configuration of mapping because it seemed easiest and most consistent with other places in GHS (it's using mapping for block elements already).

packages/ckeditor5-html-support/src/dataschema.js
```diff
@@ -251,5 +251,8 @@ function testViewName( pattern, viewName ) {
  * @property {Number} [priority] Element priority. Decides in what order elements are wrapped by
  * {@link module:engine/view/downcastwriter~DowncastWriter}.
  * Set by {@link module:html-support/dataschema~DataSchema#registerInlineElement} method.
+ * @property {String} [coupledAttribute] The name of the model attribute that generates the same view element. GHS inline attribute
+ * will be removed from the model tree as soon as the coupled attribute is removed. See
+ * {@link module:html-support/datafilter~DataFilter#_registerModelPostFixer GHS post-fixer} for more details.
```
Reinmar, line 256: Hehe, I see what you did there :) It's not ideal to link to a private method from a public API but taken that we don't have a good dev-oriented GHS's documentation yet, it makes sense to have it here.

PR #11525, opened 2022-03-29
Ck/11297 styles dropdown ghs config

### LINKED ISSUE #11297: Styles dropdown configuration: GHS
### **Why**

Implementation needed for [#5700](https://github.com/ckeditor/ckeditor5/issues/5700).

We need to provide an option to configure GHS feature using Styles Dropdown feature config.

### **What** 

Based on the UI (#11269). Implement the engine layer of the feature to support configuring GHS.

### **Scope**

- Let’s follow the format described in the [spec](https://github.com/ckeditor/ckeditor5/issues/5700#issue-516228316).
    - Affects:
        - GHS’s config
        - The Style command’s configuration

### **Other notes**

- Does configuration have to differentiate block, inline and object styles?
- What does it affect?
    - The type of the style (block vs inline) affects the command (how the style is supposed to he applied and how the command queries whether the style is available)
- GHS defines element types (e.g. that `<p>` is a block)
    - Thus, the config of SD cannot allow overriding this
    - It has to read the type from GHS
- Future: Perhaps GHS might expose some utils/API.

oleq:

Internal (style): Configuring the Styles feature now also configures GHS. Closes #11297.

---

### Additional information

In this PR I also made GHS a soft requirement of `StyleEditing`. The hard requirement that we had does not make sense in DLLs.


dawidurbanski (approved):
Nailed it! Thank you!

PR #11534, opened 2022-03-30
Fixes column resize crash on pasting subsequent table

mmotyczynska:

Fix (table-resize): fixes error on pasting subsequent table.

---

### Additional information

Actually, this error occurred when pasting a table that increased number of rows and columns _at the same time_. At first paste, the colgroup was deleted, and at the second paste the missing colgroup caused the error in the console (but also the whole column resize functionality stopped working without the colgroup).


packages/ckeditor5-table-resize/src/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -680,4 +681,24 @@ export default class TableColumnResizeEditing extends Plugin {
...
+	_registerColgroupFixer() {
+		const editor = this.editor;
+
+		this.listenTo( editor.editing.view.document, 'layoutChanged', () => {
+			const table = editor.model.document.selection.getFirstPosition().findAncestor( 'table' );
+			const tableView = editor.editing.view.document.selection.getFirstPosition().getAncestors().reverse().find(
+				element => element.name === 'table'
+			);
+
+			if ( table && table.hasAttribute( 'columnWidths' ) && tableView && tableView.getChild( 0 ).name !== 'colgroup' ) {
```
Dumluregn, line 699: I think the hard assumption that `colgroup` element is at the index `0` at some point may prove us wrong if a client has some custom element within the `figure` - in the rest of the plugin we always operate using the `find()` method (e.g. `[ ...viewTable.getChildren() ].find(...)`) and therefore I think we should stick with it. The same applies to the unit test.

Dumluregn (changes requested):
Works well, just one protection needs to be added.

Dumluregn (approved):

PR #11571, opened 2022-04-06
Allow insertImage without requiring uploadImage

scotteuser:
### Suggested merge commit message 

Allow insertImage without requiring uploadImage

---

### Warning

This is my first time contributing to CKE, apologies if I have gotten the approach, etc wrong. Happy to get feedback!

### Approach

Change the status quo to not require imageUpload, moving existing manual tests and automated tests to {existingfilename}withupload.{extension}.

- ckeditor5-image/tests/imageinsert/imageinsertui.js -> ckeditor5-image/tests/imageinsert/imageinsertwithuploadui.js
- ckeditor5-image/tests/manual/imageinsertviaurl.html -> ckeditor5-image/tests/manual/imageinsertviaurlwithupload.html
- ckeditor5-image/tests/manual/imageinsertviaurl.js -> ckeditor5-image/tests/manual/imageinsertviaurlwithupload.js
- ckeditor5-image/tests/manual/imageinsertviaurl.md -> ckeditor5-image/tests/manual/imageinsertviaurlwithupload.md

### Results

Existing behaviour moved to:
/ckeditor5-image/tests/manual/imageinsertviaurlwithupload.html
imageinsertviaurlwithupload
(screencast limited to browser tab, but at the end it opens up OS file selection dialogue).

Merge request behaviour:
/ckeditor5-image/tests/manual/imageinsertviaurl.html
imageinsertviaurl

### Additional information

This comes after discussion with @lauriii on how to solve the Drupal issue: https://www.drupal.org/project/drupal/issues/3222756


Reinmar:
Hi @scotteuser! Congrats on your first contribution :clap:  It seems complete or nearly complete.

I know you were in contact with @lauriii with whom I discussed this in the past. Unfortunately, it took a PR to make me realize that this change will bring a breaking change. It even showed in the last of your commits: 2af8fa1.

Theoretically it's not a bad moment to introduce it. However, I have second thoughts due to it.

After the change, the integrator will need to pass yet another plugin to the list. And I feel that in this particular case it's wrong because from what we see most integrators want to have the upload, so the default configuration should reflect that.

This made me think that keeping these plugins coupled and allowing configuring the range of available options differently would be better. In other words I'd see something like:

```javascript
config.image.insertUIOptions = [ 'url', 'upload', 'myCustomAssetsBrowser' ];
```

Reinmar:
We'll have to discuss this first internally to see if other folks agree and how to move this forward.

Reinmar:
We resolved this issue in #11980. Thanks for the inspiration :+1:

ThiefMaster:
How exactly do I use this if I still want to provide the ImageUpload plugin in my build? As soon as the plugin exists I get the upload logic, but if I use `removePlugins: ['ImageUpload'],` I get the `plugincollection-required {"plugin":"ImageUpload","requiredBy":"ImageInsert"}` error...

Basically, I want to be able to determine when initializing the editor whether to use insertImageViaUrl behavior or also allow uploading (since it depends on the context in my webapp whether uploading is possible or not).

PR #11612, opened 2022-04-11
Table column resize tests improvement

mmotyczynska:

Fix (table-resize): Table column resize tests improvement.

---

### Additional information

Besides fixing 3 tests that failed on my hardware I also changed util functions to make it possible to use them with any table (not only with the first table in the root, but also with nested tables for example).


Dumluregn (approved):
That was a nice refactor of both the `tableColumnResizeMouseSimulator` object and the other helpers 👏🏻 

I was only wondering for a while if we shouldn't name the helpers like `getFirstModelTable()` instead of `getModelTable()` as it only works if the table is at the beginning of the document - but that would also be inaccurate as it wouldn't fetch the first table if there is any other content before it. So I think it can stay as is.

PR #11623, opened 2022-04-12
Table column resize text formatting error fix

mateuszzagorski:

Fix (table-resize): Resolves issue when adding & removing text formatting was leading to an error.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table-resize/tests/tablecolumnresizer/utils.js
+++ b/packages/ckeditor5-table-resize/tests/tablecolumnresizer/utils.js
@@ -14,14 +14,6 @@ import Table from '@ckeditor/ckeditor5-table/src/table';
 import InsertOperation from '@ckeditor/ckeditor5-engine/src/model/operation/insertoperation';
 import MoveOperation from '@ckeditor/ckeditor5-engine/src/model/operation/moveoperation';
 import AttributeOperation from '@ckeditor/ckeditor5-engine/src/model/operation/attributeoperation';
-import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
-import Bold from '@ckeditor/ckeditor5-basic-styles/src/bold';
-import { setData, getData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
-import LinkCommand from '@ckeditor/ckeditor5-link/src/linkcommand';
-import UnlinkCommand from '@ckeditor/ckeditor5-link/src/unlinkcommand';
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-table-resize/tests/tablecolumnresizer/utils.js
```diff
@@ -330,6 +338,131 @@ describe( 'TableColumnResize utils', () => {
...
+		describe( 'should not crash if no affected table has been found', () => {
+			let model, differ, editor, editorElement, linkCommand, unlinkCommand, highlightCommand;
+
+			beforeEach( async () => {
+				editorElement = document.createElement( 'div' );
+				document.body.appendChild( editorElement );
+
+				editor = await ClassicEditor.create( editorElement, Object.assign( {}, {
+					plugins: [ Table, Paragraph, LinkEditing, HighlightEditing, Bold ]
+				} ) );
```
Dumluregn, line 351: The syntax here (I guess copied from the `createEditor()` function) is useful if we want to allow the config extensions, but here we can pass the config object directly.

packages/ckeditor5-table-resize/tests/tablecolumnresizer/utils.js
```diff
@@ -330,6 +338,131 @@ describe( 'TableColumnResize utils', () => {
...
+					'</table>'
+				);
+
+				expect( linkCommand.value ).to.be.undefined;
+
+				linkCommand.execute( 'url' );
+
+				expect( linkCommand.value ).to.be.equal( 'url' );
+
+				unlinkCommand.execute();
```
Dumluregn, line 383: Correct me if I'm wrong, but I think we are only interested in the second part of the test - removing the attribute - so we can simplify the test by applying the necessary markup directly to the model. Same for the other tests.

Dumluregn (changes requested):
The solution works well. I have a comment regarding the tests.

To keep things in order and be able to control what we test and what we don't, we have tests split into files corresponding to the source code structure. With that comes the level of abstraction in tests. As you've probably noticed, none of the tests in `utils.js` file uses the editor instance and we should keep it this way.

The tests you wrote are ok as the integration tests - they cover the particular use cases where we found issues and ensure they no longer exist - but we have an additional puzzle in between which is testing the actual `getAffectedTables()` method, like in the other tests here.

I think the place for your tests is in the `tablecolumnresizeediting.js` file in the integration part (it's an integration with attribute removal). And in the `utils.js` let's have tests using the `model.change()` block like all the other ones there (but removing the attributes).

Dumluregn (approved):

PR #11634, opened 2022-04-14
Improve History API

### LINKED ISSUE #11226: Improve `History` API

`History` is a simple class that is mostly an array with a few additional functionalities. Recently, `History` was used in real time collaboration and revision history. Some API was lacking, and it would be nice to introduce it:

*   make `_operations` public+readonly.
*   `#clear()` - removes all data from the history.
*   `#last` - returns the most recent operation from the history.
*   `#length` - returns the total amount of operations in the history.

Other than that it would be good to think about making `getOperations()` faster (now it looks through all operations). It is not a problem right now -- I will make a separate issue if it proves problematic. But it would not harm to think on this matter for some several minutes.

### LINKED ISSUE #11644: RestrictedEditing feature may produce multiple operations with the same base version
While working on the #11634, I realized that a few tests break with the new API, mainly because the `Document#version` check has been moved to another place where versions are checked when they are added to the History instead of before being applied.

In [https://github.com/ckeditor/ckeditor5/pull/11634#issuecomment-1099655852](https://github.com/ckeditor/ckeditor5/pull/11634#issuecomment-1099655852) I debugged the issue and came to these conclusions:

1.  The addMarker() is called
2.  The `applyMarkerOperation()` is called the first time and a new version for the operation is created
3.  During the operation apply, but before it's done the RestrictedEditing detected that the marker has changed and is not flat (`ensureNewMarkerIsFlat`). As this is not a normal postfixer - `model.markers.on( 'update:restrictedEditingException', ensureNewMarkerIsFlat( editor ) );` - the `ensureNewMarkerIsFlat` is called just after detecting the update of the marker.
4.  The `ensureNewMarkerIsFlat` detects a marker that is not flat and updates it.
5.  The `udpateMarker()` is called
6.  The `applyMarkerOperation()` is called the second time and a new version for the operation is created - the same as during the 2. step.
7.  The second operation is applied successfully and the document version is incremented.
8.  The first operation can't be applied - the version of the operation is the same as the second one.

ma2ciek:

Closes #11226. Closes #11644

MINOR BREAKING CHANGE: The `Document#version` became a getter and changing it is now impossible. `document.applyOperation()` increments the `Document#version`, so changing it directly should not be necessary, but for advanced use cases, it can be changed using the new `History#version` property. 

Feature (engine): Improved the  `History` API.  More information in #11634.

Fix (restricted-editing): Standard editing mode postfixers will no longer create marker operations with invalid base versions.

---

### Additional information

*   Before changes, the `Document#version` property referred to the base version of the last document operation present in the `History`, thus the property has been moved to the `History#version` and `Document#history` became a getter to provide backward compatibility and easier access.
*   The `History#addOperation()` method now automatically increments the `History#version` property. You should not need to set the `version` manually, but if it is necessary in your case, it is possible to change it and then call `History#addOperation()`. Note, that changing the `version` property may easily lead to breaking the document.
*   The `History#reset()` method has been introduced to allow resetting `History` to the initial state.
*   The `History#getOperations(from, to)` method was redesigned to be faster - now it will not have to iterate over all operations so the speed of this method for small ranges should be improved.
*  The `History#lastOperation` has been introduced to quickly access the last operation added to the history.


ma2ciek:
I see that there is a couple of tests failing. I'll fix them.

ma2ciek:
I got some weird logs from one of the RestrictedEditing tests (`should convert <span class="restricted-editing-exception"> inside table to marker` and 2 more) - maybe it's a hidden bug (?)

```
LOG: '----------------------------------------------------------------'
LOG: '{"baseVersion":0,"position":{"root":"main","path":[0],"stickiness":"toNone"},"nodes":[],"shouldReceiveAttributes":false,"__className":"InsertOperation"}'
LOG: ''
LOG: '----------------------------------------------------------------'
LOG: '{"baseVersion":1,"position":{"root":"main","path":[0],"stickiness":"toNone"},"nodes":[{"name":"paragraph"}],"shouldReceiveAttributes":false,"__className":"InsertOperation"}'
LOG: ''
LOG: '----------------------------------------------------------------'
LOG: '{"baseVersion":2,"sourcePosition":{"root":"main","path":[0],"stickiness":"toNext"},"howMany":1,"targetPosition":{"root":"$graveyard","path":[0],"stickiness":"toNone"},"__className":"MoveOperation"}'
LOG: ''
LOG: '----------------------------------------------------------------'
LOG: '{"baseVersion":3,"position":{"root":"main","path":[0],"stickiness":"toNone"},"nodes":[{"name":"table","children":[{"name":"tableRow","children":[{"name":"tableCell","children":[{"name":"paragraph","children":[{"data":"bar"}]}]}]}]}],"shouldReceiveAttributes":false,"__className":"InsertOperation"}'
LOG: ''
LOG: '----------------------------------------------------------------'
LOG: '{"baseVersion":4,"name":"restrictedEditingException:1","oldRange":{"start":{"root":"main","path":[0,0,0,0],"stickiness":"toNext"},"end":{"root":"main","path":[0,0,0,0,3],"stickiness":"toPrevious"}},"newRange":{"start":{"root":"main","path":[0,0,0,0,0],"stickiness":"toNext"},"end":{"root":"main","path":[0,0,0,0,3],"stickiness":"toPrevious"}},"affectsData":true,"__className":"MarkerOperation"}'
LOG: ''
LOG: '----------------------------------------------------------------'
LOG: '{"baseVersion":4,"name":"restrictedEditingException:1","oldRange":null,"newRange":{"start":{"root":"main","path":[0,0,0,0],"stickiness":"toNext"},"end":{"root":"main","path":[0,0,0,0,3],"stickiness":"toPrevious"}},"affectsData":true,"__className":"MarkerOperation"}'
LOG: ''
```

ma2ciek:
So after a bit of debugging it seems that:

1.  The addMarker() is called
2.  The `applyMarkerOperation()` is called the first time and a new version for the operation is created
3.  During the operation apply, but before it's done the RestrictedEditing detected that the marker has changed and is not flat (`ensureNewMarkerIsFlat`). As this is not a normal postfixer - `model.markers.on( 'update:restrictedEditingException', ensureNewMarkerIsFlat( editor ) );` - the `ensureNewMarkerIsFlat` is called just after detecting the update of the marker.
4.  The `ensureNewMarkerIsFlat` detects a marker that is not flat and updates it.
5.  The `udpateMarker()` is called
6.  The `applyMarkerOperation()` is called the second time and a new version for the operation is created - the same as during the 2. step.
7.  The second operation is applied successfully and the document version is incremented.
8.  The first operation can't be applied - the version of the operation is the same as the second one.

ma2ciek:
Blocked by #11644.

packages/ckeditor5-engine/src/model/document.js
```diff
@@ -179,6 +152,20 @@ export default class Document {
 		} );
 	}
 
+	/**
+	 * The document version. Every operation increases the version number. It is used to ensure that
+	 * operations are applied on a proper document version.
+	 *
+	 * If the {@link module:engine/model/operation/operation~Operation#baseVersion base version} does not match the document version,
+	 * a {@link module:utils/ckeditorerror~CKEditorError model-document-applyoperation-wrong-version} error is thrown.
```
scofalik, line 160:
```suggestion
	 * The document version.
	 *
	 * See {@link module:engine/model/history~History#version}.
```

scofalik: How about keeping it short and having doc in one place?

scofalik: Or we could go the other way around and keep the documentation here and make it simple in `History`. But I also see now that these docs differ a bit. This one says about the error, while `History` doc says about changing version property.

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -39,43 +41,170 @@ export default class History {
...
+		this._gaps = new Map();
 	}
 
 	/**
-	 * Adds an operation to the history.
+	 * The version of the last operation in the history.
+	 *
+	 * The history version is incremented automatically when a new operation is added to the history.
+	 * Setting the version manually should be done only in rare circumstances when there is a planned gap
+	 * between history versions. When doing so, the
```
scofalik, line 81: Unfinished doc.

packages/ckeditor5-engine/src/model/document.js
scofalik, line 160:
```suggestion
	 * The document version. Every applied operation increases the version number. It is used to
	 * ensure that operations are applied on a proper document version.
	 *
	 * This property is equal to the {@link module:engine/model/history~History#version `Document#history#version`}.
	 *
	 * If the {@link module:engine/model/operation/operation~Operation#baseVersion base version} does not match the document version,
	 * a {@link module:utils/ckeditorerror~CKEditorError model-document-applyoperation-wrong-version} error is thrown.
```

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -39,43 +41,170 @@ export default class History {
...
+	 *
+	 * @readonly
+	 * @type {module:engine/model/operation/operation~Operation|undefined}
+	 */
+	get lastOperation() {
+		return this.operations[ this.operations.length - 1 ];
+	}
+
+	/**
+	 * Adds an operation to the history and increments the history version.
```
scofalik, line 120: It is a bit internal method but I'd add that the operation's base version should be equal to the history version or an error is thrown.

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -39,43 +41,170 @@ export default class History {
...
+		}
 
-		for ( const operation of this._operations ) {
-			if ( operation.baseVersion >= from && operation.baseVersion < to ) {
-				operations.push( operation );
+		if ( fromBaseVersion === undefined ) {
+			fromBaseVersion = this.firstOperation.baseVersion;
+		}
+
+		// Change exclusive `to` to inclusive, so the `to` will refer to the actual index.
```
scofalik, line 164:
```suggestion
		// Change exclusive `toBaseVersion` to inclusive, so it will refer to the actual index.
```

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -39,43 +41,170 @@ export default class History {
...
+		// Thanks to that mapping from base versions to operation indexes are possible.
+		let inclusiveTo = toBaseVersion - 1;
+
+		// Check if the `from` or `to` points to the gap between versions.
```
scofalik, line 168:
```suggestion
		// Check if "from" or "to" point to the gap between versions.
```

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -39,43 +41,170 @@ export default class History {
...
 
-		return operations;
+		// If the whole range is outside of the operation versions, then return an empty array.
+		if ( inclusiveTo < this.firstOperation.baseVersion || fromBaseVersion > this.lastOperation.baseVersion ) {
+			return [];
+		}
+
+		let fromIndex = this._baseVersionToOperationIndex.get( fromBaseVersion );
+
+		// TODO Change
```
scofalik, line 188: What about this TODO?

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -39,43 +41,170 @@ export default class History {
...
-		this._operations.push( operation );
+		this.operations.push( operation );
+		this._version++;
+
+		this._baseVersionToOperationIndex.set( operation.baseVersion, this.operations.length - 1 );
 	}
 
 	/**
-	 * Returns operations added to the history.
+	 * Returns operations chronologically that were added to the history from the given range of operation base versions.
```
scofalik, line 145:
```suggestion
	 * Returns operations from the given range of operation base versions that were added to the history.
```

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -39,43 +41,170 @@ export default class History {
...
 	 *
-	 * @param {Number} [from=Number.NEGATIVE_INFINITY] Base version from which operations should be returned (inclusive).
-	 * Defaults to `Number.NEGATIVE_INFINITY`, which means that operations from the first one will be returned.
-	 * @param {Number} [to=Number.POSITIVE_INFINITY] Base version up to which operations should be returned (exclusive).
-	 * Defaults to `Number.POSITIVE_INFINITY` which means that operations up to the last one will be returned.
-	 * @returns {Array.<module:engine/model/operation/operation~Operation>} Operations added to the history.
+	 * Note that operation versions may contain gaps.
```
scofalik, line 147:
```suggestion
	 * Note that there may be gaps in operations base versions.
```

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -18,10 +20,10 @@ export default class History {
 		/**
 		 * Operations added to the history.
 		 *
-		 * @protected
-		 * @member {Array.<module:engine/model/operation/operation~Operation>} module:engine/model/history~History#_operations
+		 * @readonly
+		 * @type {Array.<module:engine/model/operation/operation~Operation>}
 		 */
-		this._operations = [];
+		this.operations = [];
```
scofalik, line 26: I know that I suggested this change in the original issue, but now as I think of it, I am not sure if it is needed. In places where we use `_operations` maybe we can use `getOperations()`. It will be faster than now (won't need to look through all ops).

scofalik (changes requested):
Minor suggestions. In general, LGTM.

packages/ckeditor5-engine/src/model/history.js
ma2ciek: We iterate through `operations` from the end in RTC Client. Using the `getOperations()` would slow down it significantly, but maybe it's not a problem.

packages/ckeditor5-engine/src/model/history.js
```diff
@@ -39,43 +41,170 @@ export default class History {
...
+
+	/**
+	 * The first history operation.
+	 *
+	 * @readonly
+	 * @type {module:engine/model/operation/operation~Operation|undefined}
+	 */
+	get firstOperation() {
+		return this.operations[ 0 ];
+	}
```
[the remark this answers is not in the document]
scofalik: AFAIR I had local changes that removed that part of the code :D.

PR #11660, opened 2022-04-22
Export DocumentList{,Properties}{,Editing}

### LINKED ISSUE #11610: [Document list] How to expose its DLL?
Part of #11377.

The problem right now is that list's `src/index.js` exposes only the old list feature.

How can we expose document list's DLL?

Options:

### Option 1. Add to `ckeditor5-list/src/index.js`

*   Pros:
    *   KISS
    *   Will motivate us to sunset old lists faster ;)
*   Cons:
    *   Doubled size of the DLL (and you'll never use two of these features at once)
    *   If we'll go this way, moving to opt 2 or 3 later on will be a BC

### Option 2. Move entire DocLists to a new package

*   Pros:
    *   Clean
    *   Optimal size
*   Cons:
    *   Both features depend on the same UI, so where would we keep it and how would we import it? Answer: we'd import from `ckeditor5-list/src/...`.
    *   What will we do once the old lists are sunset in the future?

### Option 3. Create a new `ckeditor5-document-list` package just for the DLL

And have only `src/index.js` there. It'd import module directly from `ckeditor5-list`. We'd ignore the cross-package import rule there.

*   Pros:
    *   Smart
    *   Optimal size
    *   No issue with sharing the same UI
*   Cons:
    *   Slightly confusing (it's a workaround rather than a solution)

arkflpc:

Other (list): Exports `DocumentList` related plugins. Closes #11610.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/src/index.js
+++ b/packages/ckeditor5-list/src/index.js
@@ -7,6 +7,10 @@
  * @module list
  */
 
+export { default as DocumentList } from './documentlist';
+export { default as DocumentListEditing } from './documentlist/documentlistediting';
+export { default as DocumentListProperties } from './documentlistproperties';
+export { default as DocumentListPropertiesEditing } from './documentlistproperties/documentlistpropertiesediting';
 export { default as List } from './list';
 export { default as ListEditing } from './list/listediting';
 export { default as ListUI } from './list/listui';


Reinmar:
It needs to be a PR to `stable` because we want to release it ASAP.

Reinmar:
@arkflpc I can see you changed the base branch here on GH. However, it means that the merge contains entire content of `master` now. You should've rebased the branch too.

Reinmar:
I tested it and both features work fine when combined in one DLL.

PR #11661, opened 2022-04-22
Fixed table column resize editor crash when adding caption

mateuszzagorski:

Fix (table-resize): Editor will no longer crash when adding caption for the table.

---

### Additional information


mlewand (approved):
LGTM

PR #11667, opened 2022-04-25
GHS - split converting figure attributes when converting table/img/media to seperate figure converter

### LINKED ISSUE #11688: [GHS] Suggestions and comments are lost on elements with enabled GHS attributes
## Steps to reproduce

    * Add GHS to http://localhost:8125/ckeditor5/external/collaboration-features/packages/ckeditor5-track-changes/tests/manual/track-changes.html sample.

    * Enable all:

```
htmlSupport: {
	allow: [
		{ name: /.*/, attributes: true, styles: true, classes: true }
	]
}
```

    * Add the insert table suggestion/add a comment on the table.

    * Re-load the data: `editor.setData( editor.getData() )`.

## Actual result

The suggestion/comment on the table is lost.
## Notes

After re-loading the data:

----

### Debugging

@scofalik Found the following:

This is because of problems with upcast converters execution order (for figure element). In general, it seems to me that converters in GHS table integration are written incorrectly. table converter should not convert figure attributes. I think it should add converters for figure.

This is what happens now:

1. Upcast figure.
2. Callbacks order are: figure converter from table plugin, then marker attributes (data-suggestion-...).
3. figure converter from table plugin fires table conversion.
4. table is converted and then GHS converter kicks in.
5. GHS converter then handles figure attributes. Since data-suggestion-... has not been converted yet, GHS handles them.
6. Then, when we finally get back to the original figure conversion (point 2.) and we get to marker attributes conversion, the attributes are already converted.

In my opinion, upcast converters should not "look up". If you convert given element, it means that its conversion was already triggered by its parent. All conversion should be handled by parent's converters, and those converters will be fired at proper time. Same is with "looking down", but there the problem is that if you handle whole figure in one element, you can't write custom stuff for <td>s etc.

CatStrategist:

Fix (html-support): Extracted upcasting attributes of `figure` view element to separate converters for `table`, `image` and `media` integrations. Closes #11688.

---

### Additional information
- I was thinking about extracting `preserveElementAttributes()` to utils, but that would require passing four parameters to it if we would remove closure. Not sure if it's worth it.
- It was kind of tricky to understand why extracted upcasting of figure elements had to have 'low' priority in order to make markers work and to not break current functionality. It required understanding how upcast converters are registered in GHS, that there is fallback converter in GHS registered with low priority and knowing that markers priority is limited to range from -999 to -1001.


packages/ckeditor5-html-support/src/integrations/image.js
```diff
@@ -110,19 +113,40 @@ function viewToModelImageAttributeConverter( dataFilter ) {
...
-					preserveElementAttributes( viewContainerElement.parent, 'htmlFigureAttributes' );
+			function preserveElementAttributes( viewElement, attributeName ) {
+				const viewAttributes = dataFilter._consumeAllowedAttributes( viewElement, conversionApi );
+
+				if ( viewAttributes ) {
+					conversionApi.writer.setAttribute( attributeName, viewAttributes, data.modelRange );
 				}
 			}
-		}, { priority: 'low' } );
+		}, { priority: priorities.get( 'low' ) } );
```
scofalik, line 149: This is the same as `priority: 'low'`... right?

packages/ckeditor5-html-support/src/integrations/image.js
```diff
@@ -110,19 +113,40 @@ function viewToModelImageAttributeConverter( dataFilter ) {
...
+function viewToModelFigureAttributeConverter( dataFilter ) {
+	return dispatcher => {
+		dispatcher.on( 'element:figure', ( evt, data, conversionApi ) => {
+			const viewFigureElement = data.viewItem;
+
+			if ( !data.modelRange || !viewFigureElement.hasClass( 'image' ) ) {
+				return;
+			}
+
+			preserveElementAttributes( viewFigureElement, 'htmlFigureAttributes' );
```
scofalik, line 140: I guess that we don't need this to be in a separate function.

packages/ckeditor5-html-support/src/integrations/mediaembed.js
```diff
@@ -89,6 +89,34 @@ function viewToModelMediaAttributesConverter( dataFilter, mediaElementName ) {
 	}
 }
 
+// View-to-model conversion helper preserving allowed attributes on {@link module:table/table~Table Table}
+// feature model element from figure view element.
+//
+// @private
+// @param {module:html-support/datafilter~DataFilter} dataFilter
+// @returns {Function} Returns a conversion callback.
+function viewToModelFigureAttributesConverter( dataFilter ) {
```
scofalik, line 98: Same comments as above.

packages/ckeditor5-html-support/src/integrations/table.js
```diff
@@ -96,7 +95,35 @@ function viewToModelTableAttributeConverter( dataFilter ) {
...
+	};
+}
+
+// View-to-model conversion helper preserving allowed attributes on {@link module:table/table~Table Table}
+// feature model element from figure view element.
+//
+// @private
+// @param {module:html-support/datafilter~DataFilter} dataFilter
+// @returns {Function} Returns a conversion callback.
+function viewToModelFigureAttributeConverter( dataFilter ) {
```
scofalik, line 108: Same comments as above.

packages/ckeditor5-html-support/src/integrations/image.js
scofalik: Also, we were talking about passing `+0.5` priority. I understand that you found another solution (adding `register:figure`?)

scofalik (changes requested):
As in comments.

packages/ckeditor5-html-support/src/integrations/image.js
CatStrategist: Yes, forgot to mention the happy news. It doesn't require +0.5 anymore, found a better solution.

packages/ckeditor5-html-support/src/integrations/image.js
CatStrategist: My reasoning was to keep it the same as in image/table/media converter, so if anyone decide it would be better to extract this peace of code to util - it would be easier to spot and extract, but no problem - I will inline this.

packages/ckeditor5-html-support/src/integrations/image.js
```diff
@@ -96,9 +101,7 @@ function viewToModelImageAttributeConverter( dataFilter ) {
 
 			preserveElementAttributes( viewImageElement, 'htmlAttributes' );
 
-			if ( viewContainerElement.is( 'element', 'figure' ) ) {
-				preserveElementAttributes( viewContainerElement, 'htmlFigureAttributes' );
-			} else if ( viewContainerElement.is( 'element', 'a' ) ) {
+			if ( viewContainerElement.is( 'element', 'a' ) ) {
 				preserveLinkAttributes( viewContainerElement );
```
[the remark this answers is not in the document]
CatStrategist: Do you mean scenario like this?
```
'<figure class="image" data-commented-end-after="foo:id" data-commented-start-before="foo:id">' +
		'<a href="www.example.com">' +
			'<img src="/assets/sample.png">' +
		'</a>' +
'</figure>'
```
If yes - then it is covered with unit test.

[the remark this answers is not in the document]
scofalik: I asked just in case as I was not sure how if `<a>` is over or inside `<figure>`. But in this case it should work fine.

Reinmar:
This is a non-trivial bug in CKEditor 5, so it'd be great to have an issue here in this repo explaining this part of the issue. We don't do that sometimes in simpler cases, but ideally we should.

niegowski:
@ckeditor/qa-team could you check this PR? We had issues before in a `<figure>` handling in GHS so I'd like to make sure that there is no regression.

LukaszGudel:
It seems like custom attributes added to `figure` will not be retained when there is a comment or a suggestion added.

### Steps
1. Allow all with htmlSupport config.
2. Set data: `editor.setData('<figure class="table" foo="bar"><table><tbody><tr><td>&nbsp;</td></tr></tbody></table></figure>')`.
3. Call `editor.getData()` -> custom attribute is still there.
4. Turn on TC and remove table as a suggestion.
5. Call `editor.setData( editor.getData() )`.

### Result

Custom attribute `foo="bar"` is lost.

LukaszGudel:
Other than the issue mentioned above, I did not find anything else. The original issue is fixed :+1:

CatStrategist:
@LukaszGudel The bug you found should be fixed, please let me know when you test it :)

LukaszGudel:
@CatStrategist it seems like custom attributes are now correctly preserved, but I don't see any colors/markers in the editor after reloading the data.

### Steps:
1. Open manual sample with TC and GHS.
2. Turn on TC.
3. Insert a new table.
4. Call `editor.setData( editor.getData() )`.


### Result:
Green highlight on table is lost.

https://user-images.githubusercontent.com/58658048/168823264-8cd9374b-8115-44e9-bdd1-d215c94c2edc.mp4

### Notes:
- this issue occurs by simply opening the sample, as there is a suggestion in initial data, but this suggestion is not visible in the editor

CatStrategist:
@LukaszGudel Can you give more details? Did you do something else maybe? I can't reproduce it, video attached

https://user-images.githubusercontent.com/34556745/168828636-be007ca5-122d-4a43-80d7-b84e90c6952f.mov

LukaszGudel:
My bad, I was on wrong branch 😅  This scenario works and you can ignore my comments

CatStrategist:
@niegowski Hi! :) PR was approved by QA, you can tak a look, thanks!

niegowski (approved):

niegowski:
Waiting for CI.

PR #11668, opened 2022-04-25
Add DocumentLists to ckeditor5-metadata.json

### LINKED ISSUE #11613: [Document list] Add DocumentLists to ckeditor5-metadata.json
Part of #11377.

Here: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-list/ckeditor5-metadata.json

arkflpc:

Internal (list): Adds `DocumentList` and `DocumentListProperties` to `ckeditor5-metadata.json`. Closes #11613.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/ckeditor5-metadata.json
+++ b/packages/ckeditor5-list/ckeditor5-metadata.json
@@ -92,6 +92,58 @@
 					"styles": "list-style-type"
 				}
 			]
+		},
+		{
+			"name": "Document list",
+			"className": "DocumentList",
+			"description": "Implements bulleted and numbered list features. Supports block content inside list elements. Incompatible with the list plugin.",
+			"path": "src/documentlist.js",
+			"uiComponents": [
+				{
+					"type": "Button",
+					"name": "numberedList",
+					"iconPath": "theme/icons/numberedlist.svg"
+				},
+				{
+					"type": "Button",
+					"name": "bulletedList",
+					"iconPath": "theme/icons/bulletedlist.svg"
+				}
+			],
+			"htmlOutput": [
+				{
+					"elements": [
+						"ol",
+						"ul"
+					]
+				},
+				{
+					"elements": "li",
+					"implements": "$block"
+				}
+			]
+		},
+		{
+			"name": "Document list properties",
+			"className": "DocumentListProperties",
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-list/ckeditor5-metadata.json
```diff
@@ -92,6 +92,58 @@
...
+			"htmlOutput": [
+				{
+					"elements": [
+						"ol",
+						"ul"
+					]
+				},
+				{
+					"elements": "li",
+					"implements": "$block"
```
Reinmar, line 122: We mentioned this in the issue that this bit (the `implements` property) should not be used in this case.

Its meaning is that the `li` will inherit e.g. text alignment classes or text indent inline styles. This is true in normal lists, but not in the document lists. Why? Because in document lists, if you apply something like text alignment, the feature will generate `<li><p text-align>...</p></li>`. So, there will be the `<p>` on which the alignment was set. Similar to how we handle this in table cells.

packages/ckeditor5-list/ckeditor5-metadata.json
```diff
@@ -92,6 +92,58 @@
 					"styles": "list-style-type"
 				}
 			]
+		},
+		{
+			"name": "Document list",
+			"className": "DocumentList",
+			"description": "Implements bulleted and numbered list features. Supports block content inside list elements. Incompatible with the list plugin.",
```
Reinmar, line 99:
```suggestion
			"description": "Implements bulleted and numbered list features. Supports block content inside list elements. Incompatible with the List plugin.",
```

Reinmar:
I left some minor comments. Other than that, it looks good. Good idea with mentioning the lack of compatibility between the plugins.

packages/ckeditor5-list/ckeditor5-metadata.json
wimleers: This is very interesting … because it means that `\Drupal\ckeditor5\HTMLRestrictions::getTextContainerElementList()` introduced in https://www.drupal.org/node/3268307 will need to change.

I think that's fine though.

@lauriii , thoughts?

lauriii: Looks like we indeed should update the list on our side to include `<li>`. I went through the related code changes to confirm. 👍

wimleers: Making the necessary metadata change on the Drupal side in the issue where we're switching from `list.List` to `list.DocumentList` 👍 

https://www.drupal.org/project/drupal/issues/3274651#comment-14502537

PR #11673, opened 2022-04-26
Column resize nested tables can be resized correctly after paste

mateuszzagorski:

Fix (table-resize): Resolved issue with nested tables not being resizable after paste.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table-resize/src/tablecolumnresize/utils.js
+++ b/packages/ckeditor5-table-resize/src/tablecolumnresize/utils.js
@@ -73,21 +73,18 @@ function getAffectedTable( change, model ) {
 		return null;
 	}
 
-	const allTables = [];
-	const document = model.document;
+	const affectedTables = [];
+	const tableNode = ( referencePosition.nodeAfter && referencePosition.nodeAfter.name === 'table' ) ?
+		referencePosition.nodeAfter : referencePosition.findAncestor( 'table' );
+	const range = model.createRangeOn( tableNode );
 
-	for ( const rootName of document.getRootNames() ) {
-		const root = document.getRoot( rootName );
-		const range = model.createRangeIn( root );
-
-		for ( const node of range.getItems() ) {
-			if ( node.is( 'element' ) && node.name === 'table' ) {
-				allTables.push( node );
-			}
+	for ( const node of range.getItems() ) {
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-table-resize/src/tablecolumnresize/utils.js
```diff
@@ -42,8 +45,9 @@ export function getAffectedTables( changes ) {
 //
 // @private
 // @param {module:engine/model/differ~DiffItem} changes
+// @param {module:engine/model/model~Model} model
 // @returns {module:engine/model/element~Element|null}
```
mlewand, line 49: You didn't change the returned type. You're returning an array now.

packages/ckeditor5-table-resize/src/tablecolumnresize/utils.js
```diff
@@ -42,8 +45,9 @@ export function getAffectedTables( changes ) {
...
-function getAffectedTable( change ) {
```
mlewand, line 46: This function no longer matches to its name. It says that it returns a **table** - while it returns multiple tables. And there's a `getAffectedTables()` right next to it.

mlewand: I think that there's no longer need to have a separation for both functions and it should be just a single function (`getAffectedTables`).

packages/ckeditor5-table-resize/tests/tablecolumnresizer/tablecolumnresizeediting.js
```diff
@@ -549,6 +549,34 @@ describe( 'TableColumnResizeEditing', () => {
 
 			expect( Math.abs( 100 - finalViewColumnWidthsPx[ 1 ] ) < PIXEL_PRECISION ).to.be.true;
 		} );
+
+		it( 'should find and allow for resizing of all tables - when nested', () => {
```
mlewand, line 553:
```suggestion
		it( 'should find and allow for resizing nested tables', () => {
```

mlewand (changes requested):
Couple of things to be fixed in the source changes.

The tests are looking good 👍

mlewand (approved):

PR #11676, opened 2022-04-26
[Document lists] The editor should strip list attributes from solo blocks to enhance UX on paste or drop

### LINKED ISSUE #11608: [Document list] Copy/cut/drag (getSelectedContent()) rooted in a single list item should not return a content wrapped in a list item
Part of #11377.

This ticket originated from #11351.

Part of the problem in #11351 is that the content that ends up in the clipboard is wrapped with `<ul><li>` so e.g. [this logic](https://github.com/ckeditor/ckeditor5/blob/7c871770d81e73131fec0919fb000eeabdbd78f6/packages/ckeditor5-image/src/image/imageinlineediting.js#L128-L204) does not discover this scenario.

However, a more general issue is this: If I copy an image from a list and paste it in the root of the editor, should be it still wrapped in a list item?

The answer is: NO.

Thus, whenever a selection starts & ends in the same list item, `getSelectedContent()` should strip list-related attributes from the returned content.

It's worth to remember that `getSelectedContent()` is a decorated method.

oleq:

Other (list): The editor should strip list attributes from solo blocks to enhance UX on paste or drop in document lists. Closes #11608.

### DIFF
--- a/packages/ckeditor5-list/src/documentlist/documentlistediting.js
+++ b/packages/ckeditor5-list/src/documentlist/documentlistediting.js
@@ -34,7 +34,8 @@ import {
 	isLastBlockOfListItem,
 	isSingleListItem,
 	getSelectedBlockObject,
-	isListItemBlock
+	isListItemBlock,
+	removeListAttributes
 } from './utils/model';
 import {
 	getViewElementIdForListType,
@@ -134,6 +135,7 @@ export default class DocumentListEditing extends Plugin {
 		this._setupDeleteIntegration();
 		this._setupEnterIntegration();
 		this._setupTabIntegration();
+		this._setupClipboardIntegration();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-list/src/documentlist/documentlistediting.js
```diff
@@ -455,6 +457,50 @@ export default class DocumentListEditing extends Plugin {
...
+		// * This prevents splitting list items when such a content is pasted/dropped by the user (block widgets in particular),
+		//   * Clipboard integration heuristics of ImageInlineEditing can kick in.
+		// * This avoids pasting single-list-item lists (orphans) anywhere else (in the root, in a paragraph).
+		//
+		//	                       ┌─────────────────────┬───────────────────┐
+		//	                       │ Selection           │ Clipboard content │
+		//	                       ├─────────────────────┼───────────────────┤
+		//	                       │ * [<blockObject />] │ <blockObject />   │
+		//	                       ├─────────────────────┼───────────────────┤
+		//	                       │ * Foo [bar] baz     │ bar               │
```
oleq, line 481: I guess I over-engineered this bit. I don't think there's any way for this kind of selection (even if in spans the entire block) to include list attributes in `getSelectedContent()`. Or at least I couldn't reproduce this. If my concerns are confirmed, I'll simplify this code because it looks like it only applies to block object only after all.

niegowski (approved):

PR #11686, opened 2022-04-27
GHS: DualContent should search for block in all nested elements

### LINKED ISSUE #11513: DualContentModelElementSupport doesn't support inline element upcasting into block element

The `html-support/integrations/dualcontent~DualContentModelElementSupport` is not handling cases where a child node that is an inline element in view changes into a block element in model. Example case where this is happening (in Drupal) is when `<a>` is wrapping a `<drupal-media>` element.

Following data:
```
<a href="example.com">
  <drupal-media></drupal-media>
</a>
```

Upcasts into (because the links plugin doesn't support linking block elements):
```
<drupalMedia linkHref="example.com"></drupalMedia>
```

So imagine the following use case:
```
<div>
  <a href="example.com">
      <drupal-media></drupal-media>
  </a>
</div>
```

### ✔️ Expected result

```
<htmlDiv>
  <drupalMedia linkHref="example.com"></drupalMedia>
</htmlDiv>
```

### ❌ Actual result

What actually happens is:
```
<htmlDivParagraph>
  <drupalMedia linkHref="example.com"></drupalMedia>
</htmlDivParagraph>
```

Which is later on reduced to since `<drupalMedia>` is a block element:
```
<drupalMedia linkHref="example.com"></drupalMedia>
```

---

arkflpc:

Fix (html-support): `DualContentModelElementSupport` should upcast inline elements as block when they contain block inside. Closes #11513.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-html-support/src/integrations/dualcontent.js
+++ b/packages/ckeditor5-html-support/src/integrations/dualcontent.js
@@ -114,6 +114,9 @@ export default class DualContentModelElementSupport extends Plugin {
 		const view = this.editor.editing.view;
 		const blockElements = view.domConverter.blockElements;
 
+		// Traversing the viewElement subtree looking for block elements.
+		// Especially for the cases like <div><a href="#"><p>foo</p></a></div>.
+		// https://github.com/ckeditor/ckeditor5/issues/11513
 		for ( const viewItem of view.createRangeIn( viewElement ).getItems() ) {
 			if ( viewItem.is( 'element' ) && blockElements.includes( viewItem.name ) ) {
 				return true;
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-html-support/src/integrations/dualcontent.js
```diff
@@ -113,8 +113,23 @@ export default class DualContentModelElementSupport extends Plugin {
...
+			if ( node.is( 'element' ) ) {
+				return hasBlockContent( node );
+			}
+
+			return false;
+		}
+
+		function hasBlockContent( viewElement ) {
+			return Array.from( viewElement.getChildren() ).some( isBlock );
+		}
```
niegowski, line 132: I would suggest using a `TreeWalker`:

```javascript
	_hasBlockContent( viewElement ) {
		const view = this.editor.editing.view;
		const blockElements = view.domConverter.blockElements;

		for ( const viewItem of view.createRangeIn( viewElement ).getItems() ) {
			if ( viewItem.is( 'element' ) && blockElements.includes( viewItem.name ) ) {
				return true;
			}
		}

		return false;
	}
```

It is more consistent with how we solve such issues and IMHO is easier to understand.

PR #11699, opened 2022-05-04
<ul type> & <ol type>

### LINKED ISSUE #11615: <ul type> and <ol type> not supported by List nor ListProperties

While bringing the `<ol start>`  and `<ol reversed>`  functionality to Drupal core, we’re also looking at `<ol type>` and `<ul type>`, because as https://ckeditor.com/blog/ckeditor-5-v32.0.0-with-new-list-properties-support-for-the-script-tag-and-enhanced-mentions/ mentions, this is now supported too and has a magnificent UX. 

https://ckeditor.com/docs/ckeditor5/latest/api/module_list_listproperties-ListPropertiesConfig.html#member-styles says:
> When set, the list style feature will be enabled. It allows changing the list-style-type HTML attribute of the lists.

But:
1. There is no `list-style-type` HTML attribute. Only a `type` attribute.
2. There seems to be no way to configure CKEditor 5 to generate a type attribute, only `list-style-type`.
3. Searching the CKE5 codebase finds plenty of matches for the strings `<ul style="list-style-type` and `<ol style="list-style-type`, but zero for `<ul type` and `<ol type`

**Conclusion:**

Because `ListProperties`' support for list style types does _not_ work for `<ol type>` or `<ul type>`, despite that being in the HTML spec:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#attr-type → not deprecated
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#attr-type → deprecated _but supported_

→ No downcast support, but also, more worryingly still, no upcast support.

Therefore the only work-around is to use GHS to add support for these. But … _that also does not work_:

1. Load the following CKE5 plugins: `htmlSupport.GeneralHtmlSupport`, `list.List`, with the following `htmlSupport` config:
```
{"allow":[{"name":"ul","attributes":{"type":true},"classes":true,"styles":true}]}
```
```
<ol type="A">
  <li>test</li>
</ol>
```

**I strongly suspect that `v32` caused this to not work with GHS anymore, because I do recall testing this a long time ago and it worked fine then.**

### ✔️ Expected result
CKEditor 5 retains the `type` attribute.

### ❌ Actual result
It does not.

## 📃 Other details

* First affected CKEditor version: `v33.0.0`

---

arkflpc:

Feature (list): Adds support for `type` attribute of `<ul>` and `<ol>` elements in addition to `list-style-type` style. Closes #11615.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/ckeditor5-metadata.json
+++ b/packages/ckeditor5-list/ckeditor5-metadata.json
@@ -87,8 +87,7 @@
 					],
 					"attributes": [
 						"start",
-						"reversed",
-						"type"
+						"reversed"
 					],
 					"styles": "list-style-type"
 				}
--- a/packages/ckeditor5-list/ckeditor5-metadata.json
+++ b/packages/ckeditor5-list/ckeditor5-metadata.json
@@ -87,7 +87,8 @@
 					],
 					"attributes": [
 						"start",
-						"reversed"
+						"reversed",
+						"type"
 					],
 					"styles": "list-style-type"
 				}
@@ -138,7 +139,8 @@
 					],
 					"attributes": [
 						"start",
-						"reversed"
+						"reversed",
+						"type"
 					],
 					"styles": "list-style-type"
 				}
--- a/packages/ckeditor5-list/src/listproperties.js
+++ b/packages/ckeditor5-list/src/listproperties.js
@@ -65,7 +65,7 @@ export default class ListProperties extends Plugin {
  */
 
 /**
- * When set, the list style feature will be enabled. It allows changing the `list-style-type` HTML attribute of the lists.
+ * When set, the list style feature will be enabled. It allows changing the `list-style-type` style or `type` HTML attribute of the lists.
  *
  * @default true
  * @member {Boolean|module:list/listproperties~ListPropertiesStyleConfig} module:list/listproperties~ListPropertiesConfig#styles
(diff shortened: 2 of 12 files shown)


packages/ckeditor5-list/src/listproperties.js
```diff
@@ -68,7 +68,7 @@ export default class ListProperties extends Plugin {
  * When set, the list style feature will be enabled. It allows changing the `list-style-type` HTML attribute of the lists.
```
oleq, line 68: It's worth mentioning here that it's controlling either `list-style-type` style or the `type` attribute depending on the value.

packages/ckeditor5-list/src/listproperties/listpropertiesui.js
```diff
@@ -277,7 +277,12 @@ function createListPropertiesView( {
 			listStyleCommand
 		} );
 
-		styleButtonViews = styleDefinitions.map( styleButtonCreator );
+		// The command can be ListStyleCommand or DocumentListStyleComman. Let's checks if it has expected function.
```
oleq, line 280: `DocumentListStyleCommand`

packages/ckeditor5-list/ckeditor5-metadata.json
```diff
@@ -87,7 +87,8 @@
 					],
 					"attributes": [
 						"start",
-						"reversed"
+						"reversed",
+						"type"
```
oleq, line 91: Are you sure this is OK? Only the `DocumentListProperties` supports the `type` attribute. `ListProperties` is incapable of returning `type` in the data pipeline.

wimleers:
YAY! This unblocks https://www.drupal.org/project/drupal/issues/3274635 — which means Drupal will be able to offer the superior list type UX that CKE5 supports, instead of the GHS-based work-around we were planning to ship with in https://www.drupal.org/project/drupal/issues/3274651 👍 🥳

PR #11735, opened 2022-05-09
Column resize: removes columnIndex attribute from table cell

mmotyczynska:

Fix (table-resize): Removes columnIndex attribute from table cell.

---

### Additional information

Requires #11673 as this fix partly depends on it (some unit tests were failing due to missing resizers in nested tables).


mlewand (approved):
LGTM but there's a conflict that needs to be resolved before the merge

PR #11744, opened 2022-05-11
[GHS] custom elements

### LINKED ISSUE #11248: Line break characters around <script> and <style> should be ignored

1. Set in Source Editing mode in `ckeditor5-html-support/tests/manual/ghs-all-features.html`:
```html
<p>
    x
</p>
<script>a</script>
<style>a</style>
<p>
    x
</p>
```
2. Retrieve the data back.

### ✔️ Expected result

Output is the same as the input.

### ❌ Actual result

Notice the extra paragraph with `&nbsp;` and `<style>` inside it:
```html
<p>x</p><script>a</script><p>&nbsp;<style>a</style></p><p>x</p>
```

### LINKED ISSUE #11432: GHS: `name: /.*/` should allow arbitrary HTML elements, including custom/non-existent ones
**Sibling issue of #10891 and #11104? 🤔 **

I cannot configure CKEditor 5 to _not_ strip away custom elements (f.e. `<drupal-media>`) or even arbitrary elements (f.e. `<yo>`).

1. Load the `htmlSupport.GeneralHtmlSupport` and `sourceEditing.SourceEditing` CKE5 plugins, with the following `htmlSupport` config:
```
{"allow":[{"name": /.*/}},"attributes":true,"classes":true,"styles":true}]}
```
2. Access a CKEditor 5 instance, use the `Source` button to paste this:
```
<p>p</p>
<p>yo</p>
```
3. Click the `Source` button again to access the rendered HTML representation
4. Click the `Source` button again to access the raw HTML representation

### ✔️ Expected result
Exactly what we pasted:
```
<p>p</p>
<p>yo</p>
```

### ❌ Actual result
```
<p>p</p>
<p>yo</p>
```

## 📃 Other details

* First affected CKEditor version: `v32.0.0`

niegowski:

Feature (html-support): Custom elements should be preserved by GHS. Closes #11432.

Fix (engine): Whitespaces between block elements should not trigger auto-paragraphing. Closes #11248.

---

### Additional information


packages/ckeditor5-engine/src/conversion/upcasthelpers.js
```diff
@@ -465,6 +465,11 @@ export function convertText() {
 				return;
 			}
 
+			// Do not auto-paragraph whitespaces.
```
Reinmar, line 468: This breaks the following test: https://github.com/ckeditor/ckeditor5/blob/30286f77b39526fce2856b03b9be0ba4cc91d1c7/packages/ckeditor5-engine/tests/view/domconverter/whitespace-handling-integration.js#L110

We need to choose to either trim that nbsp in that case or correctly handle scenarios like:

```
<custom>x</custom>

<custom someattr></custom>
```

where the space (normal space) should be trimmed.

Unfortunately, since these are custom elements we can't easily make the decision on DOM->view conversion whether these are spaces between block or inline elements.

Additionally, unfortunately we lose the information whether this was an nbsp or normal space upon DOM->view conversion. Currently, all are normalized to normal spaces at this stage. if they weren't, the below code would actually correctly differentiate these situation.

Hence, we need to:

* either lose intentional nbsps between blocks (may happen in HTML generated by external tools)
* or make sure custom elements don't break things

oleq (approved):
Yes.

Issue #11728, opened 2022-05-06
[TS MVP] Create typings for editor config


arkflpc:
To achieve the first goal of #11704 (Improving the DX for the community), it would be very helpful to have editor config statically typed.

The config can be used in different contexts:

1.  As a parameter to `Editor.create()`
2.  Within plugins: `editor.config.define()` in constructor and `editor.config.get()` later.

I would try to solve the first one and at least investigate if we can achieve the second.

If possible, we should avoid centralized type for the config and assemble it somehow from definitions near plugins that define them. We wouldn't introduce unnecessary coupling, and the user would be able to extend it as well. Maybe we could use some magic type interference based on plugin classes provided to `create()` method. If that fails, the centralized config type covering only plugins included in `ckeditor5` is better than the fully dynamic one.

 Part of #11708.

arkflpc:
At least one plugin that declares its config should be rewritten first. Then, we could test in different package if this config is visible as a member of `EditorConfig` interface.

arkflpc:
Seems to be done in other tickets

Issue #11723, opened 2022-05-06
[TS] Revise code style guide to cover TypeScript


arkflpc:
We need to revise [Code style](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/code-style.html) to cover TypeScript features.

It'd also be a good idea to write here tips & tricks that we learn while porting the code.

 Part of #11708.

Witoso:
To follow up with some team brainstorming.

arkflpc:
Draft of scope:
- Review current examples
- Add new rules from esLint
- Module import formatting
- Mixins
- Crating read-only properties

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #11755, opened 2022-05-13
Rewrite ckeditor5-utils to TypeScript

### LINKED ISSUE #11715: [TS MVP] Rewrite ckeditor5-utils package to TypeScript
We need to start moving CKEditor5 to TypeScript somewhere, and `ckeditor5-utils` is the perfect package to start with. That's because:

1.  It doesn't depend on anything else, so we won't introduce a TS package that depends on JS one.
2.  Most classes and functions are independent.
3.  It's quite stable, so the code won't diverge much from the master branch (it takes several tasks to merge it there).

Because we don't have any tooling supporting TS right now, let's keep the code on feature branch.

Exploiting the full potential of statically typed classes like `Emitter` or `Observable` can be tricky to do at the first attempt, so I wouldn't overcomplicate that. We can get back later. For instance, when rewriting `ckeditor5-engine`.

arkflpc:

Other (utils): Rewrites `ckeditor5-utils` in TypeScript. Closes #11715.

---

### Additional information


packages/ckeditor5-utils/src/collection.ts
```diff
@@ -91,7 +101,7 @@ export default class Collection {
 		 * @private
 		 * @member {String}
 		 */
-		this._idProperty = options.idProperty || 'id';
+		this._idProperty = options.idProperty || 'id' as I;
```
arkflpc, line 104: The cast wouldn't be necessary if we make options mandatory (so we can check `idProperty` against `I` type parameter). I think it would be possible to only do it for non-default `idProperty` using conditional types. That would reduce readability

packages/ckeditor5-utils/src/config.ts
```diff
@@ -13,13 +13,15 @@ import { isPlainObject, isElement, cloneDeepWith } from 'lodash-es';
...
+	private readonly _config: Record<string, any>;
+
 	/**
 	 * Creates an instance of the {@link ~Config} class.
 	 *
 	 * @param {Object} [configurations] The initial configurations to be set. Usually, provided by the user.
 	 * @param {Object} [defaultConfigurations] The default configurations. Usually, provided by the system.
 	 */
-	constructor( configurations, defaultConfigurations ) {
+	constructor( configurations: Record<string, any>, defaultConfigurations: Record<string, any> ) {
```
arkflpc, line 24: The config is completely untyped right now. Let's defer it for now till #11728.

packages/ckeditor5-utils/src/difftochanges.ts
```diff
@@ -25,23 +27,23 @@
...
+	let lastOperation: Change<T> | null = null;
 
 	diff.forEach( change => {
 		if ( change == 'equal' ) {
 			pushLast();
 
 			index++;
 		} else if ( change == 'insert' ) {
-			if ( isContinuationOf( 'insert' ) ) {
+			if ( lastOperation && lastOperation.type == 'insert' ) {
```
arkflpc, line 46: `isContinuationOf` inlined here, so TypeScript can see type guards.

packages/ckeditor5-utils/src/focustracker.ts
```diff
@@ -155,3 +161,8 @@ export default class FocusTracker {
 
 mix( FocusTracker, DomEmitterMixin );
 mix( FocusTracker, ObservableMixin );
+
+type ObservableDomEmitter = DomEmitter & Observable;
```
arkflpc, line 165: Workaround. `interface FocusTracker extends DomEmitter, Observable {}` would complain about `listenTo` not being the same.

packages/ckeditor5-utils/src/observablemixin.ts
```diff
@@ -708,6 +736,35 @@ function attachBindToListeners( observable, toBindings ) {
  * @interface Observable
  * @extends module:utils/emittermixin~Emitter
  */
+export interface Observable extends Emitter {
+	set<K extends keyof this & string>( name: K, value: this[ K ] ): void;
+	set( values: { readonly [ K in keyof this ]?: unknown } ): void;
```
arkflpc, line 741: I initially tried to `set( values: { readonly [ K in keyof this ]?:` **`this[ K ]`** `} ): void;` It compiled, but in every use site there was an error about incompatible types. I also tried using `Partial<this>` here with the same result. That's why I decided to keep `unknown` here, so the compiler would only check property names, not it's value types. Unfortunately, this style of set method is used a lot in ui-related code.

packages/ckeditor5-utils/src/observablemixin.ts
```diff
@@ -906,3 +963,122 @@ function attachBindToListeners( observable, toBindings ) {
...
  */
+
+interface SingleBindChain<TKey extends string, TVal> {
+	toMany<O extends Observable, K extends keyof O>(
+		observables: readonly O[],
+		key: K,
+		callback: ( ...values: O[ K ][] ) => TVal
+	): void;
+
+	to<O extends Observable & { [ P in TKey ]: TVal }>(
```
arkflpc, line 974: There can be many combinations here, like `x.bind( 'x' ).to( c1, c2, 'y', c3 )` which would equal to `x.bind( 'x' ).to( c1, 'x', c2, 'y', c3, 'x' )`. Unless we create a generator for these typings, let's keep only the common combinations up to 4 observables (the greatest number I've found in our code-base).

packages/ckeditor5-utils/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
arkflpc, line 1: Just an example to make VS Code working. It'll be refined along with the tooling that is using this.

packages/ckeditor5-utils/src/emittermixin.ts
```diff
@@ -335,6 +338,64 @@ export default EmitterMixin;
  *
  * @interface Emitter
  */
+export interface Emitter {
```
arkflpc, line 341: Investigation how to do statically types events is a time-consuming undertaking.

Note for myself: create a ticket to do that and link it here.

packages/ckeditor5-utils/src/keystrokehandler.ts
```diff
@@ -60,7 +63,7 @@ export default class KeystrokeHandler {
 	 *
 	 * @param {module:utils/emittermixin~Emitter} emitter
 	 */
-	listenTo( emitter ) {
+	listenTo( emitter: HTMLElement | Window ): void {
```
niegowski, line 66: This could be any emitter, especially a view document instance in the case of `EditingKeystrokeHandler` that extends this class.

packages/ckeditor5-utils/src/dom/tounit.ts
```diff
@@ -14,7 +14,7 @@
  * @param {String} unit An unit like "px" or "em".
  * @returns {module:utils/dom/tounit~helper}
```
arkflpc, line 15: Add JsDoc & TS definitions of tounit~helper (renamed to uppercase, perhaps).

packages/ckeditor5-utils/src/emittermixin.ts
```diff
@@ -335,6 +338,64 @@ export default EmitterMixin;
...
+	off( event: string, callback: Function ): void;
+
+	listenTo(
+		emitter: Emitter,
+		event: string,
+		callback: ( this: this, ev: EventInfo, ...args: any[] ) => void,
+		options?: CallbackOptions
+	): void;
+
+	stopListening( emitter?: Emitter, event?: string, callback?: Function ): void;
```
arkflpc, line 363: It won't work for instance when `stopListening(undefined, 'myevent')` Write overloads, like for `stopDelegating` to ensure all previous parameters are passed.

packages/ckeditor5-utils/src/collection.ts
```diff
@@ -62,7 +70,9 @@ export default class Collection {
 	 * @param {String} [options.idProperty='id'] The name of the property which is used to identify an item.
 	 * Items that do not have such a property will be assigned one when added to the collection.
 	 */
-	constructor( initialItemsOrOptions = {}, options = {} ) {
+	constructor( options?: { idProperty?: I } )
```
arkflpc, line 73: readonly

arkflpc, line 73: readonly

packages/ckeditor5-utils/src/collection.ts
```diff
@@ -633,9 +656,9 @@ export default class Collection {
 	 * @param {Object} item Item to be added.
 	 * @returns {String}
 	 */
-	_getItemIdBeforeAdding( item ) {
+	private _getItemIdBeforeAdding( item: { [ id in I ]?: string } ): string {
```
arkflpc, line 659: item: T

packages/ckeditor5-utils/src/keyboard.ts
```diff
@@ -204,18 +215,21 @@ function getEnvKeyCode( key ) {
...
+	contentLanguageDirection: LanguageDirection
+): boolean {
 	const localizedKeyCodeDirection = getLocalizedArrowKeyCodeDirection( keyCode, contentLanguageDirection );
 
 	return localizedKeyCodeDirection === 'down' || localizedKeyCodeDirection === 'right';
 }
 
-function generateKnownKeyCodes() {
-	const keyCodes = {
+function generateKnownKeyCodes(): { [keyCode: string]: number } {
```
arkflpc, line 231: readonly

arkflpc: Consider typing each code

packages/ckeditor5-utils/src/observablemixin.ts
```diff
@@ -7,16 +7,16 @@
  * @module utils/observablemixin
  */
 
-import EmitterMixin from './emittermixin';
+import EmitterMixin, { Emitter } from './emittermixin';
 import CKEditorError from './ckeditorerror';
 import { extend, isObject } from 'lodash-es';
```
arkflpc, line 12: remove `extend`

packages/ckeditor5-utils/src/observablemixin.ts
```diff
@@ -258,47 +255,73 @@ const ObservableMixin = {
...
+		properties: string[];
+	}[];
+}
+
 // Init symbol properties needed for the observable mechanism to work.
 //
 // @private
 // @param {module:utils/observablemixin~ObservableMixin} observable
-function initObservable( observable ) {
+function initObservable<T extends Observable>( observable: T ): asserts observable is T & {
```
arkflpc, line 320: not readable

packages/ckeditor5-utils/src/keyboard.ts
arkflpc: It's indexed with string in few places, so it must be indexable by any key.

packages/ckeditor5-utils/src/priorities.ts
```diff
@@ -23,12 +24,12 @@ const priorities = {
 	 * Converts a string with priority name to it's numeric value. If `Number` is given, it just returns it.
 	 *
 	 * @static
-	 * @param {module:utils/priorities~PriorityString|Number} priority Priority to convert.
+	 * @param {module:utils/priorities~PriorityString|Number} [priority] Priority to convert.
 	 * @returns {Number} Converted priority.
 	 */
-	get( priority ) {
+	get( priority?: PriorityString | number ): number {
```
arkflpc, line 30: Add default value

packages/ckeditor5-utils/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "compilerOptions": {
+    "lib": [
+      "DOM"
+    ],
+    "noImplicitAny": true,
+    "outDir": "./dist/",
+    "strict": true,
+    "module": "es6",
+    "target": "es2017",
```
arkflpc, line 10: esnext?

packages/ckeditor5-utils/src/focustracker.ts
```diff
@@ -29,7 +29,13 @@ import mix from './mix';
  * @mixes module:utils/dom/emittermixin~EmitterMixin
  * @mixes module:utils/observablemixin~ObservableMixin
  */
-export default class FocusTracker {
+class FocusTracker {
```
arkflpc, line 32: Cleanup docs about firing events.

packages/ckeditor5-utils/src/dom/emittermixin.ts
```diff
@@ -166,15 +180,21 @@ export default DomEmitterMixin;
...
-	 * @param {Node} node DOM Node that fires events.
+	 * @param {Node|Window} node DOM Node that fires events.
 	 * @param {Object} [options] Additional options.
 	 * @param {Boolean} [options.useCapture=false] Indicates that events of this type will be dispatched to the registered
 	 * listener before being dispatched to any EventTarget beneath it in the DOM tree.
 	 * @param {Boolean} [options.usePassive=false] Indicates that the function specified by listener will never call preventDefault()
 	 * and prevents blocking browser's main thread by this event handler.
 	 */
-	constructor( node, options ) {
+	constructor(
```
CatStrategist, line 194: Maybe we could utilize TypeScript access modifiers for constructors parameters so there is no need to redeclare and assign them in class?

packages/ckeditor5-utils/src/eventinfo.ts
```diff
@@ -14,11 +14,18 @@ import spy from './spy';
  * manipulate it.
  */
 export default class EventInfo {
+	readonly source: object;
```
CatStrategist, line 17: Should we be explicit about access modifiers and mark these properties with public keyword? It would be consistent with protected/private properties so we don't have a case where private and protected properties have access modifier keyword, but public ones don't.

CatStrategist (approved):

packages/ckeditor5-utils/src/dom/rect.ts
```diff
@@ -418,12 +431,19 @@ export default class Rect {
...
+ */
+export type RectSource = HTMLElement | Range | Window | { readonly [ Prop in RectProperty ]: number };
+
 // Acquires all the rect properties from the passed source.
 //
 // @private
 // @param {module:utils/dom/rect~Rect} rect
 // @param {ClientRect|module:utils/dom/rect~Rect|Object} source
-function copyRectProperties( rect, source ) {
+function copyRectProperties( rect: Rect, source: { readonly [ prop in RectProperty ]: number } ): void {
```
arkflpc, line 446: RectLike

packages/ckeditor5-utils/src/dom/resizeobserver.ts
```diff
@@ -209,6 +219,11 @@ ResizeObserver._elementCallbacks = null;
  * @mixes module:utils/domemittermixin~DomEmitterMixin
  */
 class ResizeObserverPolyfill {
```
arkflpc, line 221: Remove polyfill. https://caniuse.com/resizeobserver

packages/ckeditor5-utils/src/dom/tounit.ts
```diff
@@ -14,7 +14,7 @@
...
  */
-export default function toUnit( unit ) {
+export default function toUnit( unit: string ): ( value: unknown ) => string {
```
arkflpc, line 17:
```
string|number
```

packages/ckeditor5-utils/src/dom/position.ts
```diff
@@ -267,18 +261,36 @@ function getRectForAbsolutePositioning( rect ) {
...
 	 * @param {Object} [options] options object.
 	 * @param {module:utils/dom/rect~Rect} options.elementRect The positioned element rect.
 	 * @param {module:utils/dom/rect~Rect} options.targetRect The target element rect.
-	 * @param {module:utils/dom/rect~Rect} options.viewportRect The viewport rect.
+	 * @param {module:utils/dom/rect~Rect|null} options.viewportRect The viewport rect.
+	 * @param {module:utils/dom/rect~Rect} [options.limiterRect] The limiter rect.
 	 * @param {HTMLElement|null} [options.positionedElementAncestor] Nearest element ancestor element which CSS position is not "static".
 	 */
-	constructor( positioningFunction, options ) {
+	constructor(
```
arkflpc, line 284: Mark constructor as private (in JsDocs)

packages/ckeditor5-utils/src/dom/emittermixin.ts
arkflpc: It would be difficult to find the best place to put JsDoc comment. Let's discuss that in #11723.

packages/ckeditor5-utils/src/collection.ts
```diff
@@ -25,7 +25,17 @@ import mix from './mix';
...
  */
-export default class Collection {
+class Collection<T extends { [ id in I ]?: string }, I extends string = 'id'> implements Iterable<T> {
+	private readonly _items: T[];
+	private readonly _itemMap: Map<string, T>;
+	private readonly _idProperty: I;
+	private _bindToCollection?: Collection<any, any> | null;
+	private readonly _bindToExternalToInternalMap: WeakMap<any, T>;
+	private readonly _bindToInternalToExternalMap: WeakMap<T, any>;
+	private _skippedIndexesFromExternal: number[];
```
niegowski, line 35: JSDocs shouldn't be here? As discussed, it's more readable than when it's in the constructor.

packages/ckeditor5-utils/src/dom/resizeobserver.js
```diff
@@ -1,378 +0,0 @@
...
- */
-ResizeObserver._elementCallbacks = null;
-
-/**
- * A polyfill class for the native [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver).
- *
- * @private
- * @mixes module:utils/domemittermixin~DomEmitterMixin
- */
-class ResizeObserverPolyfill {
```
niegowski, line 211: Why did you remove this file completely? We were expected only to remove the polyfill AFAIK.

niegowski: Ugh, it's below... diff tricked me :D

packages/ckeditor5-utils/src/elementreplacer.ts
```diff
@@ -12,6 +12,8 @@
  * the original elements from the DOM.
  */
 export default class ElementReplacer {
+	private _replacedElements: { element: HTMLElement, newElement?: HTMLElement }[];
```
niegowski, line 15: Please move JSDoc here.

PR #11771, opened 2022-05-17
Improve editor-is-readonly-has-no-setter error message.

### LINKED ISSUE #11695: Improve the error for using `Editor.isReadonly` setter
Currently it can be found here: [https://ckeditor.com/docs/ckeditor5/latest/support/error-codes.html#error-editor-isreadonly-has-no-setter](https://ckeditor.com/docs/ckeditor5/latest/support/error-codes.html#error-editor-isreadonly-has-no-setter)

And looks like this:

> `Editor#isReadOnly` does not have a setter and should be set using `Editor#enableReadOnlyMode( lockId )` and `Editor#disableReadOnlyMode( lockId )`.

There are at least a couple of things that can be improved here:

*   Those should be links to the methods. Save people time whenever you can.
*   We should clearly state that this is a change introduced in a version XYZ. That before this you should use the setter. But starting from XYZ you need to use these methods.
*   You can add a code example of how you used it before and how it looks now. Again, save the time.

CatStrategist:

Docs (core): Improved an error message when trying to set the editor in read-only mode by using `Editor#isReadOnly` setter. Closes #11695.

---

### Additional information

How it looks now

obraz

### DIFF
--- a/packages/ckeditor5-core/src/editor/editor.js
+++ b/packages/ckeditor5-core/src/editor/editor.js
@@ -246,8 +246,19 @@ export default class Editor {
 
 	set isReadOnly( value ) {
 		/**
-		 * `Editor#isReadOnly` does not have a setter and should be set using `Editor#enableReadOnlyMode( lockId )` and
-		 * `Editor#disableReadOnlyMode( lockId )`.
+		 * {@link #isReadOnly Editor#isReadOnly} property is read-only since the version 34.0.0 and can be set only using
+		 * {@link #enableReadOnlyMode Editor#enableReadOnlyMode( lockId )} and
+		 * {@link #disableReadOnlyMode Editor#disableReadOnlyMode( lockId )}.
+		 *
+		 * Before the version 34.0.0:
+		 *
+		 *		editor.isReadOnly = true;
+		 * 		editor.isReadOnly = false;
+		 *
+		 * Since the version 34.0.0:
+		 *
+		 *		editor.enableReadOnlyMode( 'my-feature-id' );
+		 * 		editor.disableReadOnlyMode( 'my-feature-id' );
 		 *
 		 * @error editor-isreadonly-has-no-setter
 		 */
--- a/packages/ckeditor5-core/src/editor/editor.js
+++ b/packages/ckeditor5-core/src/editor/editor.js
@@ -246,16 +246,16 @@ export default class Editor {
 
 	set isReadOnly( value ) {
 		/**
-		 * {@link #isReadOnly Editor#isReadOnly} property is read-only since the version 34.0.0 and can be set only using
-		 * {@link #enableReadOnlyMode Editor#enableReadOnlyMode( lockId )} and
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -246,8 +246,19 @@ export default class Editor {
 
 	set isReadOnly( value ) {
 		/**
-		 * `Editor#isReadOnly` does not have a setter and should be set using `Editor#enableReadOnlyMode( lockId )` and
-		 * `Editor#disableReadOnlyMode( lockId )`.
+		 * {@link #isReadOnly Editor#isReadOnly} property is read-only since the version 34.0.0 and can be set only using
```
scofalik, line 249:
```suggestion
		 * {@link #isReadOnly Editor#isReadOnly} property is read-only since version `34.0.0` and can be set only using
```

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -246,8 +246,19 @@ export default class Editor {
...
+		 * {@link #enableReadOnlyMode Editor#enableReadOnlyMode( lockId )} and
+		 * {@link #disableReadOnlyMode Editor#disableReadOnlyMode( lockId )}.
+		 *
+		 * Before the version 34.0.0:
```
scofalik, line 253:
```suggestion
		 * Before version `34.0.0`:
```

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -246,8 +246,19 @@ export default class Editor {
...
+		 *
+		 *		editor.isReadOnly = true;
+		 * 		editor.isReadOnly = false;
+		 *
+		 * Since the version 34.0.0:
```
scofalik, line 258:
```suggestion
		 * Since version `34.0.0`:
```

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -246,8 +246,19 @@ export default class Editor {
 
 	set isReadOnly( value ) {
 		/**
-		 * `Editor#isReadOnly` does not have a setter and should be set using `Editor#enableReadOnlyMode( lockId )` and
-		 * `Editor#disableReadOnlyMode( lockId )`.
+		 * {@link #isReadOnly Editor#isReadOnly} property is read-only since the version 34.0.0 and can be set only using
+		 * {@link #enableReadOnlyMode Editor#enableReadOnlyMode( lockId )} and
+		 * {@link #disableReadOnlyMode Editor#disableReadOnlyMode( lockId )}.
```
scofalik, line 251:
```suggestion
		 * {@link #enableReadOnlyMode `Editor#enableReadOnlyMode( lockId )`} and
		 * {@link #disableReadOnlyMode `Editor#disableReadOnlyMode( lockId )`}.
```

scofalik (approved):

PR #11847, opened 2022-05-30
Not able to select text that overlaps with column resize handler

### LINKED ISSUE #11770: Not able to select text that overlaps with column resize handler using the mouse

1. On latest `ck/epic/3284-table-column-resize` open `tablecolumnresize` manual test
2. Merge two cells so that there is a column resize handler from other rows going through the middle of it.
3. In this merged cell, add text that fills the whole cell width
4. Try to select the text in such a way, that the mouse hovers over the column resize handler

### ✔️ Expected result

Text selection has priority over the resize handler - you are always able to select the whole text.

### ❌ Actual result

Not able to select the text. When you hover the mouse over the resize handler, it takes the priority and selects the whole cell instead.
**NOTE: if you do this very quickly, you will be able to select the text (see the recording below).**

### Screencast

https://user-images.githubusercontent.com/34380544/168823761-f05385e6-4b45-4719-a63f-dc5d6eb23135.mov

## 📃 Other details

* Browser: any

---

### LINKED ISSUE #11772: Unable to drop the content in place which overlaps with cell resize handler

1. On latest `ck/epic/3284-table-column-resize` open `tablecolumnresize` manual test
2. Merge two cells so that there is a column resize handler from other rows going through the middle of it.
3. In this merged cell, add some text that fills the whole cell width
4. Select some text and try to drag and drop it into place where the resize handler overlaps with the cell's text

### ✔️ Expected result

You are able to insert the text.

### ❌ Actual result

When drop cursor overlaps the resize handler, selection jumps to the cell below. You cannot insert the text.
Other:
- in Firefox, the selection jumps on the whole table and back to the cell

### Screencast

https://user-images.githubusercontent.com/34380544/168825896-4884fa83-334e-4697-98ce-0b466db74218.mov

## 📃 Other details

* Browser: any

---

mmotyczynska:

Fix (table): Not able to select text that overlaps with column resize handler. Closes #11770.

---

### Additional information


packages/ckeditor5-table/src/tablecolumnresize/utils.js
```diff
@@ -365,3 +365,21 @@ export function removeColumnResizerElements( viewWriter, viewCell ) {
 
 	viewWriter.remove( viewTableColumnResizerElement );
 }
+
+// Shows/hides resizers in specified view.
+//
+// @param {module:ui/view~View} view
+// @param {true|false} value
+export function setResizersVisibility( view, value ) {
```
mlewand, line 373: `setResizersVisibility()` sounds like it would show the resizer if it's set to `true` which is not entirely true. If it's set to true the resizers will be available but doesn't necessary have to be shown.

I think like `hideAllColumnResizers()` / `restoreAllColumnResizers()` would be a more fitting. But at the same time I'd prefer going with a single function like you did but I don't have good name for that 😄

packages/ckeditor5-table/src/tablecolumnresize/utils.js
```diff
@@ -365,3 +365,21 @@ export function removeColumnResizerElements( viewWriter, viewCell ) {
 
 	viewWriter.remove( viewTableColumnResizerElement );
 }
+
+// Shows/hides resizers in specified view.
+//
+// @param {module:ui/view~View} view
+// @param {true|false} value
```
mlewand, line 372:
```suggestion
// @param {Boolean} value
```

It's a boolean value - just document its type.

mlewand: Also I think that a value that tells the purpose would be better, e.g. `isVisible`.

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -450,42 +450,47 @@ export default class TableColumnResizeEditing extends Plugin {
...
-		const editingView = editor.editing.view;
-
-		editingView.addObserver( MouseEventsObserver );
-		editingView.document.on( 'mousedown', this._onMouseDownHandler.bind( this ), { priority: 'high' } );
-
 		const domEmitter = Object.create( DomEmitterMixin );
 
+		domEmitter.listenTo( global.window.document, 'mousedown', this._onMouseDownHandler.bind( this ), { priority: 'high' } );
 		domEmitter.listenTo( global.window.document, 'mouseup', this._onMouseUpHandler.bind( this ) );
+		domEmitter.listenTo( global.window.document, 'drop', this._onMouseUpHandler.bind( this ), { useCapture: true } );
```
mlewand, line 457: Why do we use drop listener? I believe that dropping something should also trigger `mouseup` event. Unless... we're cancelling it in our drag and drop handler 🙂

Either way we should shortly explain this in a comment.

mlewand (changes requested):
There's a weird behavior regression on the proposed branch:

[https://user-images.githubusercontent.com/5353898/172794187-9a12f21c-43a2-47d8-a3b0-2b6ecd2bba0b.mp4](https://user-images.githubusercontent.com/5353898/172794187-9a12f21c-43a2-47d8-a3b0-2b6ecd2bba0b.mp4)

We can't let it work like that.

But overall the direction is good 👌

packages/ckeditor5-table/src/tablecolumnresize/utils.js
mmotyczynska: True... ;)

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
mmotyczynska: While working on this issue, the thing with drag&drop got complicated and cannot be handled here anymore. So I decided to revert this change and work on it in other PR as we have a dedicated ticket for this: #11772.

packages/ckeditor5-table/src/tablecolumnresize/utils.js
mmotyczynska: I prefer a single function so I changed its name to `setResizersEnablement`, WDYT?

packages/ckeditor5-table/src/tablecolumnresize/utils.js
```diff
@@ -365,3 +365,21 @@ export function removeColumnResizerElements( viewWriter, viewCell ) {
...
+export function setResizersEnablement( view, isEnabled ) {
+	view.change( viewWriter => {
+		for ( const root of view.domRoots.values() ) {
+			const viewRoot = view.domConverter.mapDomToView( root );
+
+			if ( isEnabled ) {
+				viewWriter.removeClass( 'ck-resizers-hidden', viewRoot );
+			} else {
+				viewWriter.addClass( 'ck-resizers-hidden', viewRoot );
+			}
```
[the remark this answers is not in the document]
Dumluregn: Also the class name should be more precise to better reflect the scope it affects - right now it may suggest that e.g. image resizers will be hidden too.

packages/ckeditor5-table/tests/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -1946,6 +1946,104 @@ describe( 'TableColumnResizeEditing', () => {
...
+				setModelData( model,
+					'<table columnWidths="100%">' +
+					'<tableRow>' +
+					'<tableCell>' +
+					'<paragraph>' +
+					'[foo]' +
+					'</paragraph>' +
+					'</tableCell>' +
+					'</tableRow>' +
+					'</table>'
```
Dumluregn, line 2004: You can use a `modelTable` util here as well (just as you did for the other test).

packages/ckeditor5-table/tests/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -1946,6 +1946,104 @@ describe( 'TableColumnResizeEditing', () => {
...
+				);
+
+				const resizer = view.getDomRoot().querySelector( '.table-column-resizer' );
+
+				resizer.dispatchEvent( new MouseEvent( 'mousedown', { bubbles: true } ) );
+
+				const viewRoot = view.domConverter.mapDomToView( view.getDomRoot() );
+
+				expect( viewRoot.hasClass( 'ck-resizers-hidden' ) ).to.be.false;
```
Dumluregn, line 1991: Grouping the commands a bit differently may improve the readability of test flow:

```suggestion
                const resizer = view.getDomRoot().querySelector( '.table-column-resizer' );
                const viewRoot = view.domConverter.mapDomToView( view.getDomRoot() );

                resizer.dispatchEvent( new MouseEvent( 'mousedown', { bubbles: true } ) );

                expect( viewRoot.hasClass( 'ck-resizers-hidden' ) ).to.be.false;
```

Dumluregn: Also you can access a `viewRoot` a bit simpler, using `view.document.getRoot()`.

packages/ckeditor5-table/tests/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -1946,6 +1946,104 @@ describe( 'TableColumnResizeEditing', () => {
...
+			} );
+
+			it( 'should hide resizers when table cell is clicked and show resizers on mouseup', () => {
```
Dumluregn, line 1994: There is one missing chain in our tests. The logic looks like this:

`mousedown` / `mouseup` -> `setResizersEnablement()` -> add/remove class.

We should ensure all the links are tested. The second one (between function and class) is already covered in the `utils` file. The test that we have here skips the first link - it jumps from the start to the end directly. Instead we should check, if the mousedown and mousup events trigger the `setResizersEnablement()` function (and don't trigger them when they shouldn't). The integration test like the one here can also be useful to ensure that all the links are connected into a reasonable chain, but it is not always mandatory.

packages/ckeditor5-table/tests/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -1946,6 +1946,104 @@ describe( 'TableColumnResizeEditing', () => {
...
+
+				td.dispatchEvent( new MouseEvent( 'mousedown', { bubbles: true } ) );
+
+				const viewRoot = view.domConverter.mapDomToView( view.getDomRoot() );
+
+				expect( viewRoot.hasClass( 'ck-resizers-hidden' ) ).to.be.true;
+
+				td.dispatchEvent( new MouseEvent( 'mouseup', { bubbles: true } ) );
+
+				expect( viewRoot.hasClass( 'ck-resizers-hidden' ) ).to.be.false;
```
Dumluregn, line 2017: After we split the tests this will be irrelevant, but having a few boolean `expect`s without messages will cause that if this test fails, you won't know which expect is responsible.

Ok, here you'd still know because one of them should be `true` and the other `false`, but you get the point.

packages/ckeditor5-table/tests/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -1946,6 +1946,104 @@ describe( 'TableColumnResizeEditing', () => {
...
+
+		describe( 'on resizer mouse drag', () => {
+			it( 'does not change table selection', () => {
+				setModelData( model, modelTable( [
+					[ '00[]', { contents: '01', colspan: 2 } ],
+					[ '10', '11', '12' ]
+				], { columnWidths: '50%,25%,25%' } ) );
+
+				const paragraph = view.domConverter.mapViewToDom(
+					view.document.getRoot().getChild( 0 ).getChild( 1 ).getChild( 1 ).getChild( 1 ).getChild( 2 ).getChild( 0 )
```
Dumluregn, line 2029: This chain doesn't look healthy :smile: We could improve it in various ways:

1.  Split the chain links to separate lines with comments which element it is, like:

```
view.document.getRoot() // div
	.getChild( 0 ) // figure
	.getChild( 1 ) // table
	(...)
```

2\. Get the dom root right away and use the `querySelector()`.

The same goes for the resizer const.

packages/ckeditor5-table/tests/tablecolumnresize/utils.js
```diff
@@ -574,6 +575,43 @@ describe( 'TableColumnResize utils', () => {
...
+			const viewRoot = view.domConverter.mapDomToView( view.getDomRoot() );
+
+			setResizersEnablement( view, false );
+
+			expect( viewRoot.hasClass( 'ck-resizers-hidden' ) ).to.be.true;
+
+			setResizersEnablement( view, true );
+
+			expect( viewRoot.hasClass( 'ck-resizers-hidden' ) ).to.be.false;
+		} );
```
Dumluregn, line 613: Currently this test checks two things. Ideally each unit test should check only one for easier problem detecting, so let's split it into two parts - first one adding and second one removing the class.

The second test should not repeat the steps from the first one (so setting the class on view by running the tested helper), but get the expected conditions right away. There is a `setData()` helper in `@ckeditor5-engine/src/dev-utils/view`, so let's use it here.

Dumluregn: Also it would be nice to cover the cases where nothing happens - so trying to remove the class if it's not applied and applying a class on the root that already has this class.

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -450,42 +450,51 @@ export default class TableColumnResizeEditing extends Plugin {
...
-	 * Handles the `mousedown` event on column resizer element.
+	 * Handles the `mousedown` event on the document.
+	 *
+	 * The element clicked can be a resizer (then we handle it by enabling resizing process) or other element (then we hide resizers).
 	 *
 	 * @private
-	 * @param {module:utils/eventinfo~EventInfo} eventInfo
-	 * @param {module:engine/view/observer/domeventdata~DomEventData} domEventData
+	 * @param {EventInfo} eventInfo
+	 * @param {MouseEvent} domEvent
```
Dumluregn, line 472: Good job with keeping the docs up to date.

Dumluregn (changes requested):
Functionally this PR not only solves #11770, but also improves the integration with drag&drop a lot (see https://github.com/ckeditor/ckeditor5/issues/11772) 👏🏻 - it works almost fine, besides the fact that after d&d the resizers are not visible until the first click anywhere in the editor. I think after we merge this PR we could also close that issue and extract a second one for fixing that mishap.

Other than that - the usual refactoring/renaming stuff here and there.

Dumluregn:
Taken over by #12280.

PR #11860, opened 2022-06-01
Docs: adding demos' code to the predefined builds example guides.

godai78:

Docs: adding demos' code to the predefined builds example guides.

---

### Additional information

This addresses in part (but not necessarily closes) the following issue: #625   
  
Verify the need to use CS in these examples.


docs/examples/builds/classic-editor.md
```diff
@@ -11,3 +12,52 @@ classes: main__content--no-toc
...
+
+Check out the {@link installation/advanced/predefined-builds#installation-example Quick start} guide to learn more about implementing this kind of editor. You will find implementation steps there. You can see this example editor’s code below.
+
+<details>
+<summary>View editor configuration script</summary>
+
+```js
+
+import ClassicEditor from '@ckeditor/ckeditor5-build-classic/src/ckeditor';
+import CloudServices from '@ckeditor/ckeditor5-cloud-services/src/cloudservices';
```
niegowski, line 26: Same as for the balloon editor.

docs/examples/builds/document-editor.md
```diff
@@ -12,3 +13,234 @@ See the {@link framework/guides/document-editor tutorial} to learn how to create
...
+
+Check out the {@link installation/advanced/predefined-builds#installation-example-5 Quick start} guide to learn more about implementing this kind of editor. You will find implementation steps there. You can see this example editor’s code below.
+
+<details>
+<summary>View editor configuration script</summary>
+
+```js
+
+import DecoupledEditor from '@ckeditor/ckeditor5-build-decoupled-document/src/ckeditor';
+import CloudServices from '@ckeditor/ckeditor5-cloud-services/src/cloudservices';
```
niegowski, line 26: This is not needed (as in other snippets).

docs/examples/builds/inline-editor.md
```diff
@@ -11,3 +12,106 @@ classes: main__content--no-toc
...
+
+Check out the {@link installation/advanced/predefined-builds#installation-example-2 Quick start} guide to learn more about implementing this kind of editor. You will find implementation steps there. You can see this example editor’s code below.
+
+<details>
+<summary>View editor configuration script</summary>
+
+```js
+
+import InlineEditor from '@ckeditor/ckeditor5-build-inline/src/ckeditor';
+import CloudServices from '@ckeditor/ckeditor5-cloud-services/src/cloudservices';
```
niegowski, line 26: Not needed (as in other snippets, it's included in the build).

docs/examples/builds/inline-editor.md
```diff
@@ -11,3 +12,106 @@ classes: main__content--no-toc
...
+				top: window.getViewportTopOffsetConfig()
+			}
+		},
+		toolbar: {},
+        cloudServices: {
+                    // PROVIDE CORRECT VALUES HERE:
+                    tokenUrl: 'https://example.com/cs-token-endpoint',
+                    uploadUrl: 'https://your-organization-id.cke-cs.com/easyimage/upload/',
+                    webSocketUrl: 'your-organization-id.cke-cs.com/ws/'
+                },
```
niegowski, line 43: Indentation is mixed (you should use tab only).

FilipTokarski:
Now that you mentioned `viewportOffset` I'm thinking whether we should get rid of them from every example. I understand that it's here only because of issues on our docs (the top navigation bar interfering with editor UI). So maybe let's not confuse users with additional config options that they mostly don't need.
  
Now I also think about `cloudServices` option, which is not needed unless you want to use Easy Image. This makes an impression that in order to use standard builds you need to provide CS credentials, which is not true.

@Mgsy WDYT? Maybe we should get rid of those options and present the users only the minimal (most real-life) config?

Mgsy:
> Now that you mentioned `viewportOffset` I'm thinking whether we should get rid of them from every example. I understand that it's here only because of issues on our docs (the top navigation bar interfering with editor UI). So maybe let's not confuse users with additional config options that they mostly don't need.

I think that's a good idea, as this configuration is used to fix our docs-related issue.

> Now I also think about `cloudServices` option, which is not needed unless you want to use Easy Image. This makes an impression that in order to use standard builds you need to provide CS credentials, which is not true.

All builds include Easy Image and our demos have the `cloudServices` configuration specified. I think that removing it could make a false impression that the image upload will work out of the box. However, I will improve this section a bit:

*   Remove the `webSocketUrl` property from the configuration, as it is required by real-time collaboration, not included in the build.
*   Add the comment above the `cloudServices` configuration object explaining why it is needed. Something like this: _All predefined builds include the Easy Image feature that is responsible for image upload. Providing this configuration is required for using this service. Read more about Easy Image in the documentation -_ [_https://ckeditor.com/docs/ckeditor5/latest/features/images/image-upload/easy-image.html._](https://ckeditor.com/docs/ckeditor5/latest/features/images/image-upload/easy-image.html.) _For alternative image upload methods see the following guide - https://ckeditor.com/docs/ckeditor5/latest/features/images/image-upload/image-upload.html._ 
*   :point\_up: it's a bit long, so maybe we could rephrase it :D

godai78:
I decided to split the information for multiple reasons, and the visual presentation is one of them, too.

godai78:
@Mgsy - is this what you meant?

docs/examples/builds/balloon-block-editor.md
```diff
@@ -12,3 +13,60 @@ classes: main__content--no-toc
...
+	.create( document.querySelector( '#snippet-balloon-block-editor' ), {
+        cloudServices: {
+                    // PROVIDE CORRECT VALUES HERE:
+                    tokenUrl: 'https://example.com/cs-token-endpoint',
+                    uploadUrl: 'https://your-organization-id.cke-cs.com/easyimage/upload/',
+                    webSocketUrl: 'your-organization-id.cke-cs.com/ws/'
+                },
+		ui: {
+			viewportOffset: {
+				top: window.getViewportTopOffsetConfig()
```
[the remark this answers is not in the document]
godai78: Removed.

Mgsy (approved):

Issue #5473, opened 2019-10-09
Keystroke for BlockToolbar


oskarwrobel:
It is possible to navigate through block toolbar items using <kbd>arrow left</kbd> and <kbd>arrow right</kbd> keys. It is possible to close the block toolbar using <kbd>Esc</kbd> or moving focus back to the editable using <kbd>Tab</kbd>. The only missing feature is keystrok for opening and focusing the block toolbar.

![jun-27-2018 20-39-18](https://user-images.githubusercontent.com/1057056/41992948-3784a946-7a4a-11e8-80f0-7954079cc9ac.gif)

Issue #5146, opened 2018-04-19
Unable to focus balloon toolbars (image toolbar, table toolbar etc)


jodator:
As selected here (sorry for image with text) I'm not able to move focus to a visible baloon toolbar with keyboard.

![selection_011](https://user-images.githubusercontent.com/247363/38980532-5a7d243c-43bd-11e8-884f-edf2798c5b7d.png)

I've tried:
- ALT+F10 - obviously moves to main toolbar
- TAB - moves to image caption filed

Same is true for other toolbar like table, media embed etc.

---

If you'd like to see this feature implemented, add 👍 to this post.

Reinmar:
The idea here would be that Alt+F10 could cycle through all possible toolbars (imaga/table/widget, balloon, classic, block, and so on...).

fhalna:
Hello everyone,
Many thanks for the work done for the accessibility of Ckeditor 5, unfortunately this bug related to the possibility of inserting alternatives to images in the editor makes the editor completely inaccessible to blind users.

is it possible to change the milestone of this ticket?

sunesimonsen:
It would be nice with a general fix that allows navigating to a contextual toolbar registered with `WidgetToolbarRepository`. 

I would expect `Tab` to focus the first item and escape to move focus back to the selected editor element.

wimleers:
FYI: @bramd re-discovered this bug using NVDA at Drupal Dev Days Ghent 2022 — see https://wimleers.com/blog/ddd-ghent-2022 & https://twitter.com/gaborhojtsy/status/1512059083129593858.



> The idea here would be that Alt+F10 could cycle through all possible toolbars (imaga/table/widget, balloon, classic, block, and so on...).

This sounds relatively simple to implement I think? If you can give us guidance, maybe we could help make this happen, to address this hard blocker ASAP. 🤞 

But it is also _not_ what https://ckeditor.com/docs/ckeditor5/latest/features/keyboard-support.html#user-interface-and-navigation says:

> Move focus to the visible contextual balloon: `Tab`

So … what is the "visible contextual balloon" actually referring to? 🤔

mlewand:
First of all let's make this issue a generic issue for any dynamically added/context toolbar (image toolbar, table toolbar, media embed toolbar etc).

One idea as mentioned is to make one hotkey cycle though all toolbars (good benefit is the discoverability of the feature).

However, let's see if there are any applications out there that have this problem solved so that we can have interoperability with it.

[...]

PR #11876, opened 2022-06-06
Enabled toolbar focus in source editing mode

### LINKED ISSUE #10368: It's impossible to focus the toolbar with the keyboard in the source editing mode

1.  Use the keyboard to enter the source editing mode.
2.  Focus on the editable and type something.
3.  Try to move back to the toolbar with the keyboard.

Related - #9906.

---

mateuszzagorski:

Fix (source-editing): It is possible now to focus the toolbar with the keyboard in the source editing mode. Closes #10368.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-source-editing/src/sourceediting.js
+++ b/packages/ckeditor5-source-editing/src/sourceediting.js
@@ -278,36 +278,34 @@ export default class SourceEditing extends Plugin {
 			this._dataFromRoots.set( rootName, data );
 		}
 
-		if ( document.querySelector( '.ck-source-editing-area > textarea' ) ) {
-			const textarea = document.querySelector( '.ck-source-editing-area > textarea' );
-			const toolbar = editor.ui.view.toolbar;
+		const textarea = document.querySelector( '.ck-source-editing-area > textarea' );
+		const toolbar = editor.ui.view.toolbar;
 
-			// Start listening for the keystrokes coming from the textarea.
-			this._keystrokes.listenTo( textarea );
+		// Start listening for the keystrokes coming from the textarea.
+		this._keystrokes.listenTo( textarea );
 
-			if ( toolbar ) {
-				this._focusTracker.add( toolbar.element );
-				this._focusTracker.add( textarea );
+		if ( toolbar ) {
+			this._focusTracker.add( toolbar.element );
+			this._focusTracker.add( textarea );
 
-				// Listen for the keystrokes coming from the editor's toolbar.
-				this._keystrokes.listenTo( toolbar.element );
+			// Listen for the keystrokes coming from the editor's toolbar.
+			this._keystrokes.listenTo( toolbar.element );
 
-				this._keystrokes.set( 'Alt+F10', ( data, cancel ) => {
-					if ( this._focusTracker.isFocused && !toolbar.focusTracker.isFocused ) {
-						toolbar.focus();
+			this._keystrokes.set( 'Alt+F10', ( data, cancel ) => {
+				if ( this._focusTracker.isFocused && !toolbar.focusTracker.isFocused ) {
+					toolbar.focus();
 
-						cancel();
-					}
-				} );
+					cancel();
+				}
+			} );
 
-				this._keystrokes.set( 'Esc', ( data, cancel ) => {
-					if ( toolbar.focusTracker.isFocused ) {
-						textarea.focus();
+			this._keystrokes.set( 'Esc', ( data, cancel ) => {
+				if ( toolbar.focusTracker.isFocused ) {
+					textarea.focus();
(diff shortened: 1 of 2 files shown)


Dumluregn (changes requested):
My findings:

1) The tests coverage report shows, that the code you added in the editing part of the plugin isn't tested. You can check it locally running `yarn test -f source-editing --coverage` and inspecting the report created in the `coverage` directory. It suggests something is wrong with tests.
2) If you comment the code you added, the tests are still passing. Another indicator that tests don't actually test the code they are intended to.
3) Inspecting the tests code:
- You are creating a new toolbar separated from the editor used in the test and don't attach it anyhow. That means that after `const toolbar = editor.ui.view.toolbar;`, `toolbar` is `undefined` and the keystrokes handlers are not registered.
- So why the tests pass? Let's take a look at the `enableToolbarKeyboardFocus()` function used there. Its docs say:
```
Enables focus/blur toolbar navigation using `Alt+F10` and `Esc` keystrokes.
```
So it does the same job we try to do, and it seems it does it well since the tests pass.
- The conclusion - we are using the solution we should in tests, but not in the plugin itself. We should use the dedicated helper and then remember to use the correct toolbar in the tests.


Other than that, I've noticed that the `Esc` key works only after the first switch to the source editing mode - when you switch again, it stops working. But maybe that issue will be solved by changing the implementation.

Dumluregn (approved):
LGTM, requesting another review from the core team. A couple of notes to whomever checks this PR later:
1) We create a separate `focusTracker` here - I'm not 100% sure if that's the right approach or we should reuse the one from the editor. 
2) AFAICS the `focusTracker.destroy()` method does not clear the list of tracked elements - on contrary to its docs: https://github.com/ckeditor/ckeditor5/blob/c41a8e6c0d920668b2b724fe7c918e964386b396/packages/ckeditor5-utils/src/focustracker.js#L109-L116. In did in the first version: https://github.com/ckeditor/ckeditor5/commit/594d6c3c80ffa58f8f3d91d0482a592e84ed9c33, but later that line was removed due to some redundancy - but I don't see it: https://github.com/ckeditor/ckeditor5/commit/950bf6c05d5bba4dd92efa06c308eb22b15f15ff. I think it should be restored - asking for the confirmation.
3) I believe the commit that I've recalled in 2) should be enough to close https://github.com/ckeditor/ckeditor5/issues/4971 - also waiting for the approval.

wimleers:
Tracked downstream at https://www.drupal.org/project/drupal/issues/3283800#comment-14561193 👍

oleq (changes requested):
Aside from the concerns expressed by @Dumluregn, I must admit that I don't like this whole approach from the architectural point of view.

By hardcoding Alt+F10 and Esc logic in `SourceEditing` we're straying further from `enableToolbarKeyboardFocus()` that was an attempt to keep things in one place. The following code works because `SourceEditing's` `KeystrokeHandler` listens to the `editor.ui.view.toolbar.element`

https://github.com/ckeditor/ckeditor5/blob/ef2565f184bed48af61b7df89d79e8fa6814ca4b/packages/ckeditor5-source-editing/src/sourceediting.js#L295-L301

And it works magically probably because it won the race in the order of execution upon the `Esc` key that supposed to be encapsulated in `enableToolbarKeyboardFocus()` (two `KeystrokeHandlers`, two listeners for the same key)

https://github.com/ckeditor/ckeditor5/blob/30286f77b39526fce2856b03b9be0ba4cc91d1c7/packages/ckeditor5-ui/src/toolbar/enabletoolbarkeyboardfocus.js#L53-L63

We should centralize the logic, not scatter it across features. Since `enableToolbarKeyboardFocus()` is not enough to handle all this (it'a 5yo helper developed when there was a single editor, no source editing etc.), what we need is a piece of code that:
* will know about the `editor.ui.view.toolbar` (for instance, `BalloonEditorUIView` does not have one, but it has something else), so probably it should be provided by the implementation of the editor, not `SourceEditing`
    * will know that in some editors, focusing toolbar is not enough, some must be shown first
* will allow registering DOM elements (editing DOM root(s), source editing's textarea, what else?)
* will offer an API to focus the toolbar (since it knows about it) that individual features' `KeystrokeHandlers` could hook up to `Alt+F10` or, even better, an API that will hide `Alt+F10` from the features altogether,
* will be able to revert focus back upon `Esc` to the place where it originally came from

Also, since we want to have Alt+F10 a focal point of navigation across different toolbars (https://github.com/ckeditor/ckeditor5/issues/5146, https://github.com/ckeditor/ckeditor5/issues/5473, etc.), the design of the new system should this requirement into account.

mlewand:
Closing the PR in favor of #12064.

Issue #11867, opened 2022-06-02
Missing labels for all the widget editables


mateuszzagorski:
## :memo: **Provide detailed reproduction steps (if any)**

1.  Open e.g. https://ckeditor.com/docs/ckeditor5/latest/features/table.html#basic-table-features
2.  Launch VoiceOver or other screen reader
3.  Focus the table cell.

### :heavy\_check\_mark: **Expected result**

The editable should be announced as an appropriate element (e.g. table cell).

### :x: **Actual result**

There is no label.

### :question: **Possible solution**

Solution to all the different widget editables should be based on the results from the #11837 issue.

---

If you'd like to see this fixed sooner, add a :+1: reaction to this post.

PR #11880, opened 2022-06-06
Missing labels for nested editables

### LINKED ISSUE #11837: Missing labels for nested editables

1.  Open https://ckeditor.com/docs/ckeditor5/latest/examples/builds/classic-editor.html
2.  Launch VoiceOver or other screen reader
3.  Focus the caption of the image.

### ✔️ Expected result

The editable should be announced as an image caption.

### ❌ Actual result

There is no label.

### ❓ Possible solution

The caption can be announced as, e.g. "Caption for \<image's alt> image".

## 📃 Other details

*   Browser: all
*   OS: all
*   Installed CKEditor plugins: image, table

The same issue exists for other nested editables, e.g. table cells.

---

mmotyczynska:

Fix (image, table): Adds missing labels for nested editables. Closes #11837.

---

### Additional information

In this PR, the `label` is added for figcaption element and the `role` is added to all editables. Labels for other elements will be added in #11867.


mlewand (changes requested):
We already have an interesting problem that we need to solve 🙂

It works well for the first render. However, the image's `alt` value can change at a runtime (the author can change alternative text at any time or people collaborating might change it). We should have this information bound so that it doesn't get out of sync.

mmotyczynska:
Good point @mlewand :+1: I updated the code to cover also changing image `alt` attribute at runtime.

packages/ckeditor5-image/src/imagecaption/imagecaptionediting.js
```diff
@@ -132,7 +133,10 @@ export default class ImageCaptionEditing extends Plugin {
 					keepOnFocus: true
 				} );
 
-				return toWidgetEditable( figcaptionElement, writer );
+				const imageAlt = modelElement.parent.getAttribute( 'alt' );
+				const label = ( imageAlt ) ? t( 'Caption for %0 image', [ imageAlt ] ) : t( 'Caption for the image' );
```
oleq, line 137: Do you think replacing `%0` with `"%0"` would make sense? I ran this in a screen reader both ways and I was able to better understand the description of an image and the caption label with the quotes (the boundary). But maybe it's just me. 

cc @Comandeer

packages/ckeditor5-image/src/imagecaption/imagecaptionediting.js
```diff
@@ -132,7 +133,10 @@ export default class ImageCaptionEditing extends Plugin {
...
+
```
oleq, line 138: Missing entries in `contexts.json`.

packages/ckeditor5-image/src/imagecaption/imagecaptionediting.js
oleq: Also, obsolete parentheses around the first operand of the ternary operator.

packages/ckeditor5-image/src/imagecaption/imagecaptionediting.js
```diff
@@ -242,4 +246,36 @@ export default class ImageCaptionEditing extends Plugin {
...
+	/**
+	 * Reconverts image caption when image alt attribute changes.
+	 * The change of alt attribute is reflected in caption's aria-label attribute.
+	 *
+	 * @private
+	 */
+	_registerCaptionReconversion() {
+		const model = this.editor.model;
+
+		model.document.on( 'change:data', () => {
```
oleq, line 259: @niegowski Could you take a look at this bit please? For instance, I'm worried about about accessing `image.name` without checking whether it exists but I'm not an expert in reconversion.

oleq (changes requested):
The whole idea looks good. I checked with JAWS and VoiceOver. I left some comments.

Looks like there are tests failing in ckbox after merging master (see the CI) and decreased code coverate.

packages/ckeditor5-image/src/imagecaption/imagecaptionediting.js
niegowski: Accessing `change.range.start.nodeAfter` is safe there because the range of a Differ attribute change diff item always starts before the changed node. For an element attribute change it's: `[<el>]</el>` but for the text, the range is flat: `<$text>foo</$text>[<$text alt="x">bar</$text>]<$text>baz</$text>`. This makes `const caption = change.range.end.nodeAfter;` unsafe for cases where `alt` model attribute would be enabled (for whatever reason may be some 3rd party feature) on text nodes. So I would suggest to use a dedicated util to find the caption element: https://ckeditor.com/docs/ckeditor5/latest/api/module_image_imagecaptionutils_utils-ImageCaptionUtils.html#function-getCaptionFromImageModelElement.

mmotyczynska:
Requested changes are addressed. Unit tests fixed. Code coverage improved.

packages/ckeditor5-image/lang/contexts.json
```diff
@@ -22,5 +22,6 @@
 	"Insert": "The label of the form submit button if the image source URL input has no value.",
 	"Update": "The label of the form submit button if the image source URL input has a value.",
 	"Insert image via URL": "The input label for the Insert image via URL form.",
-	"Update image URL": "The input label for the Insert image via URL form for a pre-existing image."
+	"Update image URL": "The input label for the Insert image via URL form for a pre-existing image.",
+	"Caption for image: %0": "Text used by screen readers do describe an image, e.g. 'Caption for image: this is a description of the image.'"
```
oleq, line 26:
```suggestion
	"Caption for the image": "Text used by screen readers do describe an image when the image has no text alternative.",
	"Caption for image: %0": "Text used by screen readers do describe an image when there is a text alternative available, e.g. 'Caption for image: this is a description of the image.'"
```

packages/ckeditor5-image/src/imagecaption/imagecaptionediting.js
```diff
@@ -242,4 +246,38 @@ export default class ImageCaptionEditing extends Plugin {
...
+			const changes = model.document.differ.getChanges();
+
+			for ( const change of changes ) {
+				if ( change.attributeKey !== 'alt' ) {
+					continue;
+				}
+
+				const image = change.range.start.nodeAfter;
+
+				if ( image.name === 'imageBlock' ) {
```
oleq, line 271:
```suggestion
				if ( imageUtils.isBlockImage( image ) ) {
```

packages/ckeditor5-image/src/imagecaption/imagecaptionediting.js
```diff
@@ -242,4 +246,38 @@ export default class ImageCaptionEditing extends Plugin {
...
+	/**
+	 * Reconverts image caption when image alt attribute changes.
+	 * The change of alt attribute is reflected in caption's aria-label attribute.
+	 *
+	 * @private
+	 */
+	_registerCaptionReconversion() {
+		const editor = this.editor;
+		const model = editor.model;
+		const imageCaptionUtils = editor.plugins.get( 'ImageCaptionUtils' );
```
oleq, line 259:
```suggestion
		const imageUtils = editor.plugins.get( 'ImageUtils' );
		const imageCaptionUtils = editor.plugins.get( 'ImageCaptionUtils' );
```

oleq (changes requested):
Do you think the nested editable label would make sense for the table captions as well?

https://github.com/ckeditor/ckeditor5/blob/6040c00a53dd84a977f1c8da0f0f72bbbd1c1647/packages/ckeditor5-table/src/tablecaption/tablecaptionediting.js#L107

mmotyczynska:
> Do you think the nested editable label would make sense for the table captions as well?
> 
> https://github.com/ckeditor/ckeditor5/blob/6040c00a53dd84a977f1c8da0f0f72bbbd1c1647/packages/ckeditor5-table/src/tablecaption/tablecaptionediting.js#L107

Yes, and I guess we already have an issue for that #11867.

PR #11884, opened 2022-06-07
Update .gitignore

yoyo837:
`.gitignore` Add `.DS_Store`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ mrgit.json
 
 build/
 !packages/ckeditor5-build-*/build
+.DS_Store


Reinmar:
@pomek I know this is a minor issue and the entire core team has it configured globally (and I recommend this approach anyway). But it's not the first time there's a PR to add DS\_Store to this list so perhaps we should just add it and forget.

Also, I can see that we have this very inconsistent across our main repos (where ideally some consistency would be good). I actually prefer the CF's version with:

```
# All "dot directories".
.*/**
!.github/**

node_modules/
build/
```

WDYT about using the first two patterns in cke5, cke5-i and cf?

Issue #10845, opened 2021-11-15
[Feature Request] Trusted Types support


tosmolka:
## Trusted Types support

CKEditor 5 should support Trusted Types API so that it can be seamlessly integrated into applications that enforce Trusted Types for all DOM XSS Injection Sinks (such as innerHTML setters) via CSP directive require-trusted-types-for. Trusted Types are now fully supported in major browsers such as Chrome or Edge. 

To support Trusted Types we should identify all instances where CKeditor 5 calls such methods and propose re-factoring. We need to be careful to keep supporting browsers without Trusted Types support. This is usually done by testing whether window.trustedTypes is defined and fall-back to current behavior if it's not. 

Initial list to consider: 

- Here we could maybe avoid calling Injection Sink methods altogether
  - https://github.com/ckeditor/ckeditor5/blob/4c228fb0549507f01d1f1ba4f9862817ac589279/packages/ckeditor5-clipboard/src/dragdrop.js#L554
  - ~~https://github.com/ckeditor/ckeditor5/blob/8f0b1c306108b3e7f849b9ce5664ddb48ac2ab80/packages/ckeditor5-engine/src/view/filler.js#L59~~
  - https://github.com/ckeditor/ckeditor5/blob/99a56669b8cf4a5fb0d3e461aec8824596eed96a/packages/ckeditor5-minimap/src/utils.js#L219
  - https://github.com/ckeditor/ckeditor5/blob/bce8267e16fccb25448b4c68acc3bf54336aa087/packages/ckeditor5-paste-from-office/src/filters/space.js#L44
  - https://github.com/ckeditor/ckeditor5/blob/487d5444e72718725dbbf64b14af42dbd6b996a9/packages/ckeditor5-widget/src/widgettypearound/widgettypearound.js#L42

- Here we could leverage [TrustedTypePolicyFactory.emptyHTML](https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedtypepolicyfactory-emptyhtml)
  - https://github.com/ckeditor/ckeditor5/blob/bce8267e16fccb25448b4c68acc3bf54336aa087/packages/ckeditor5-ui/src/icon/iconview.js#L106

[...]

PR #11897, opened 2022-06-08
Fix (engine): Use innerText instead of innerHTML in view/filler

tosmolka:
### Suggested merge commit message
Fix (engine): Use innerText instead of innerHTML in view/filler

---

### Additional information

This PR fixes one of the issues mentioned in #10845 . This should be a low-risk change IMHO, happy to redo if needed.

### DIFF
--- a/packages/ckeditor5-engine/tests/view/filler.js
+++ b/packages/ckeditor5-engine/tests/view/filler.js
@@ -10,7 +10,8 @@ import {
 	INLINE_FILLER,
 	startsWithFiller,
 	isInlineFiller,
-	getDataWithoutFiller
+	getDataWithoutFiller,
+	MARKED_NBSP_FILLER
 } from '../../src/view/filler';
 
 describe( 'filler', () => {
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-engine/tests/view/filler.js
```diff
@@ -119,4 +120,12 @@ describe( 'filler', () => {
...
 		} );
 	} );
+
+	describe( 'MARKED_NBSP_FILLER', () => {
+		it( 'should return node with correct HTML', () => {
+			const node = MARKED_NBSP_FILLER( document ); // eslint-disable-line new-cap
+
+			expect( node.outerHTML ).to.equal( '<span data-cke-filler="true">&nbsp;</span>' );
+		} );
+	} );
```
niegowski, line 130: Unfortunately, this test is not verifying whether `innerText` is used. It's passing even without changing the helper implementation. It would be good to have it tested to ensure the implementation won't get rolled back.

niegowski (changes requested):

packages/ckeditor5-engine/tests/view/filler.js
tosmolka: @niegowski , I added one more test case, can you pls take another look? Thank you.

niegowski (approved):

PR #11920, opened 2022-06-13
Replacing data-placeholder attribute with aria-placeholder.

### LINKED ISSUE #11848: Placeholder is treated as regular text (at least in voiceover)

1. Open [full-featured editor](https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/full-featured-editor.html).
2. Delete all editor contents so the placeholder appears: _Type or paste your content here_.
4. Run voiceover to read the editor contents.

### ✔️ Expected result

Voiceover reads the text _Type or paste your content here_ but it points out that this is a placeholder.

### ❌ Actual result

Voiceover just reads the text _Type or paste your content here_ as it was a normal content.

## 📃 Other details

---

mateuszzagorski:

MINOR BREAKING CHANGE (engine): Replaced the `data-placeholder` attribute with `aria-placeholder`. Closes #11848.

Tests (editor-balloon, editor-classic, editor-decoupled, editor-inline, engine, heading, html-support, image, link, media-embed, table): Tests updated after the attribute replacement.

---

### Additional information

We’ve discussed adding the `aria-placeholder` alongside already existing `data-placeholder` because of the accessibility improvements, but decided it is a more coherent approach to replace it altogether.

So all the places in the project where we’ve been using the `data-placeholder` attribute should’ve been updated and now properly use `aria-placeholder`.

Additionally, in the original issue there was a comment about adding the `aria-hidden` attribute to prevent screen readers from reading the text rendered within the editable. Short research shows that:  
  
Image caption without the `aria-hidden` attribute is being read as follows:

*   **NVDA:**
    *   Enter image caption  caption  multi line  editable  blank
*   VoiceOver:
    *   Enter image caption, text editing, Enter image caption, courser at the end of the word: caption

Image caption with the `aria-hidden="true"` attribute is being read as follows:

*   NVDA:
    *   Enter image caption caption multi line editable selected Enter image caption
*   VoiceOver:
    *   Enter image caption, text editing

Different approach would be to alter the existing code and remove content being added with `::before` and replace it with a span, set with `aria-hidden="true"` attribute. Sample code:

```
<div contenteditable="true" aria-placeholder="A placeholder text">
	<span aria-hidden="true">A placeholder text</span>
</div>
```

This is being read by screen readers:

*   NVDA:
    *   A placeholder text  section  multi line  editable  A
*   VoiceOver:
    *   A placeholder text, text editing

### DIFF
--- a/packages/ckeditor5-core/src/editor/editorconfig.jsdoc
+++ b/packages/ckeditor5-core/src/editor/editorconfig.jsdoc
@@ -286,9 +286,9 @@
  *		};
  *
  * The placeholder text is displayed as a pseudo–element of an empty paragraph in the editor content.
- * The paragraph has the `.ck-placeholder` CSS class and the `data-placeholder` attribute.
+ * The paragraph has the `.ck-placeholder` CSS class and the `aria-placeholder` attribute.
  *
- *		<p data-placeholder="Type some text..." class="ck-placeholder">
+ *		<p aria-placeholder="Type some text..." class="ck-placeholder">
  *			::before
  *		</p>
  *
--- a/packages/ckeditor5-editor-balloon/tests/ballooneditorui.js
+++ b/packages/ckeditor5-editor-balloon/tests/ballooneditorui.js
@@ -125,7 +125,7 @@ describe( 'BalloonEditorUI', () => {
 					.then( newEditor => {
 						const firstChild = newEditor.editing.view.document.getRoot().getChild( 0 );
 
-						expect( firstChild.getAttribute( 'data-placeholder' ) ).to.equal( 'placeholder-text' );
+						expect( firstChild.getAttribute( 'aria-placeholder' ) ).to.equal( 'placeholder-text' );
 
 						return newEditor.destroy();
 					} );
@@ -144,7 +144,7 @@ describe( 'BalloonEditorUI', () => {
 					.then( newEditor => {
 						const firstChild = newEditor.editing.view.document.getRoot().getChild( 0 );
 
-						expect( firstChild.getAttribute( 'data-placeholder' ) ).to.equal( 'placeholder-text' );
+						expect( firstChild.getAttribute( 'aria-placeholder' ) ).to.equal( 'placeholder-text' );
 
 						return newEditor.destroy();
 					} );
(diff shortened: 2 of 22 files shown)


Comandeer (changes requested):
I've tested with http://localhost:8125/ckeditor5-editor-inline/tests/manual/placeholder.html in VoiceOver and different browsers. Results are not promising:

- Chrome reads the placeholder… even when the editor contains some text,
- Safari reads the placeholder only if there is absolutely nothing inside the editor (so even placing the element with a visual placeholder makes Safari ignore the placeholder),
- Firefox does not read the placeholder at all.

Test in NVDA:

- Chrome seems to work correctly (the placeholder is read only when the field is empty),
- Firefox still does not read the placeholder at all.

So probably we would need to come up with some other solution that works also outside Chrome on Windows :thinking:

I'm thinking about something like https://codepen.io/Comandeer/pen/BaYXBzN

- Instead of `[aria-placeholder]` we use `[aria-describedby]` and point it to the element with a visual placeholder.
- The element with a visual placeholder uses the `::before` technique we use currently for displaying the placeholder.
- The element with a visual placeholder has also the `[aria-hidden]` attribute to be ignored by screen readers and not treated as a text.
- If the user starts typing, the `[aria-describedby]` attribute alongside the element, which is pointed by it, are removed and the placeholder is no more announced .
- When all the content from the editor is removed, the element alongside the `[aria-describedby]` attribute are readded and the placeholder should be announced correctly.

It's not exactly the same behavior as with the `[aria-placeholder]` but IMO it's *close enough* and does the job (user is informed about the possible action in the currently focused field). The only downside: Firefox does not read the placeholder (it's probably connected with the use of the `::before` pseudoelement; yet not using it generates issue with a visual placeholder being treated as text so we'd be back to square one…).

A similar technique should work with image captions – after adding some element that would hold the visual placeholder, e.g.:

```html
<span id="placeholder" data-placeholder="Placeholder" aria-hidden="true"></span>
```

IMO the `[aria-hidden=true]` part is essential, otherwise the element is treated as a part of the editor's content – and that's the very issue we're trying to fix.

mlewand:
Closing as there's still work to do and we won't be doing it any time soon.

PR #11939, opened 2022-06-20
Rewrite ckeditor5-engine/src/model to TypeScript

arkflpc:

Internal (engine): Migrates `ckeditor5-engine/src/model` to TypeScript.

---

### Additional information


packages/ckeditor5-engine/src/model/differ.ts
```diff
@@ -19,12 +32,21 @@ import Range from './range';
...
 export default class Differ {
+	private readonly _markerCollection: MarkerCollection;
+	private readonly _changesInElement: Map<Element | DocumentFragment, ChangeItem[]>;
+	private readonly _elementSnapshots: Map<Element | DocumentFragment, { name: string; attributes: Map<string, unknown> }[]>;
+	private readonly _changedMarkers: Map<string, { newMarkerData: MarkerData; oldMarkerData: MarkerData }>;
+	private _changeCount: number;
+	private _cachedChanges: DiffItem[] | null;
+	private _cachedChangesWithGraveyard: DiffItem[] | null;
+	private _refreshedItems: Set<Item>;
+
```
niegowski, line 43: Missing JSDoc comments.

packages/ckeditor5-engine/src/model/differ.ts
```diff
@@ -1216,13 +1273,24 @@ function _changesInGraveyardFilter( entry ) {
  * @interface DiffItem
  */
 
+export type DiffItem = DiffItemInsert | DiffItemRemove | DiffItemAttribute;
```
niegowski, line 1276: Missing JSDoc.

packages/ckeditor5-engine/src/model/differ.ts
```diff
@@ -1260,6 +1328,15 @@ function _changesInGraveyardFilter( entry ) {
...
  */
 
+export interface DiffItemRemove {
+	type: 'remove';
+	name: string;
+	attributes: Map<string, unknown>;
+	position: Position;
+	length: number;
+	changeCount?: number;
+}
```
niegowski, line 1338: Same as above.

packages/ckeditor5-engine/src/model/differ.ts
```diff
@@ -1297,6 +1374,17 @@ function _changesInGraveyardFilter( entry ) {
...
+export interface DiffItemAttribute {
+	type: 'attribute';
+	attributeKey: string;
+	attributeOldValue: unknown;
+	attributeNewValue: unknown;
+	range: Range;
+	position?: Position;
+	changeCount?: number;
+	length?: number;
+}
```
niegowski, line 1386: Missing JSDoc. Internal properties should be marked not to confuse API users.

packages/ckeditor5-engine/src/model/document.ts
```diff
@@ -38,12 +45,21 @@ const graveyardName = '$graveyard';
...
+class Document {
+	public readonly model: Model;
+	public readonly history: History;
+	public readonly selection: DocumentSelection;
+	public readonly roots: Collection<RootElement, 'rootName'>;
+	public readonly differ: Differ;
+
+	private readonly _postFixers: Set<( writer: Writer ) => unknown>;
+	private _hasSelectionChangedFromTheLastChangeBlock: boolean;
+
```
niegowski, line 57: Missing JSDoc.

packages/ckeditor5-engine/src/model/document.ts
```diff
@@ -270,7 +286,7 @@ export default class Document {
 	 *
 	 * @param {Function} postFixer
 	 */
-	registerPostFixer( postFixer ) {
+	public registerPostFixer( postFixer: ( writer: Writer ) => unknown ): void {
```
niegowski, line 289: Post-fixer callbacks should return `true` if they modified the model. Maybe we could be more verbose here?

packages/ckeditor5-engine/src/model/documentfragment.ts
```diff
@@ -23,7 +35,13 @@ import isIterable from '@ckeditor/ckeditor5-utils/src/isiterable';
...
  * {@link module:engine/model/writer~Writer#insert} function.
  */
-export default class DocumentFragment {
+export default class DocumentFragment implements Iterable<Node> {
+	public readonly markers: Map<string, Range>;
+	declare public name?: undefined;
+	declare public rootName?: undefined;
+
+	private readonly _children: NodeList;
+
```
niegowski, line 44: Missing JSDoc.

packages/ckeditor5-engine/src/model/documentfragment.ts
```diff
@@ -23,7 +35,13 @@ import isIterable from '@ckeditor/ckeditor5-utils/src/isiterable';
  * will be set to the {@link module:engine/model/model~Model#markers model markers} by a
  * {@link module:engine/model/writer~Writer#insert} function.
  */
-export default class DocumentFragment {
+export default class DocumentFragment implements Iterable<Node> {
+	public readonly markers: Map<string, Range>;
+	declare public name?: undefined;
+	declare public rootName?: undefined;
```
niegowski, line 41: Maybe those should be marked as internal?

niegowski (commented):
Partial review.

packages/ckeditor5-engine/src/model/documentfragment.ts
arkflpc: Nope. They are for the cases when you have something like this:

```typescript
let parent: Element | DocumentFragment;
if ( parent.name == 'xyz' /* would fail if DocumentFragment didn't have `name` property */ ) {
	DoSomething( parent ); // it know that parent is Element, because on DocumentFragment it's undefined and can't equal 'xyz'
}
```

arkflpc: Nope. They are for the cases when you have something like this:

```typescript
let parent: Element | DocumentFragment;
if ( parent.name == 'xyz' /* would fail if DocumentFragment didn't have `name` property */ ) {
	DoSomething( parent ); // it know that parent is Element, because on DocumentFragment it's undefined and can't equal 'xyz'
}
```

arkflpc: And on `Node` they're public

PR #11942, opened 2022-06-21
Docs: adding CDN solution to Quick start guide.

godai78:

Docs: adding CDN solution to Quick start guide. Closes cksource/ckeditor5-internal#1803

---

### Additional information


docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,770 @@
...
+	CKEDITOR.ClassicEditor
+		.create( document.querySelector( '#editor' ) )
+		.catch( error => {
+			console.error( error );
+		} );
+</script>
 ```
 
-Based on these packages you can create a simple application.
+Remove the plugins you do not need with the `removePlugins` method. In this example, we will remove the premium collaboration features that require credentials to work. We need to do this, otherwise the editor will throw an error.
```
Mgsy, line 121:
```suggestion
Remove the plugins you do not need with the `removePlugins` configuration option. In this example, we will remove the premium collaboration features that require credentials to work. We need to do this, otherwise the editor will throw an error.
```

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,770 @@
...
+```html
+<script src="https://cdn.ckeditor.com/ckeditor5/{@var ckeditor5-version}/super-build/ckeditor.js"></script>
 ```
 
-<info-box>
-    If you cannot use the latest webpack (at the moment of writing this guide, it is 5), the provided configuration will also work with webpack 4.
-</info-box>
-
-### Creating an editor
+Call the {@link module:editor-classic/classiceditor~ClassicEditor#create `ClassicEditor.create()`} method.
```
Mgsy, line 109:
```suggestion
In the superbuild, all editor classes are stored under the `CKEDITOR` object. To create the classic editor, you need to access `CKEDITOR` object first, then call the {@link module:editor-classic/classiceditor~ClassicEditor#create `ClassicEditor.create()`} method.
```

I've added a small suggestion to make it clear how to access editor classes in the superbuild.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,770 @@
...
-
-ClassicEditor
+CKEDITOR.ClassicEditor
 	.create( document.querySelector( '#editor' ), {
-		plugins: [ Essentials, Paragraph, Bold, Italic ],
-		toolbar: [ 'bold', 'italic' ]
-	} )
-	.then( editor => {
-		console.log( 'Editor was initialized', editor );
+		removePlugins: [ 'Comments, RealTimeCollaborativeComments, TrackChanges, RealTimeCollaborativeTrackChanges, RevisionHistory, RealTimeCollaborativeRevisionHistory, RealTimeCollaborativeEditing, ExportWord, ExportPdf, Pagination, WProofreader, MathType' ],
```
Mgsy, line 130: Each item should be wrapped in `''`. Also, we should add `PresenceList` and `TrackChangesData` to this list.

```suggestion
		removePlugins: [ 'Comments', 'TrackChanges', 'TrackChangesData', 'RevisionHistory', 'RealTimeCollaborativeComments', 'RealTimeCollaborativeTrackChanges', 'RealTimeCollaborativeRevisionHistory', 'RealTimeCollaborativeEditing', 'PresenceList', 'ExportWord', 'ExportPdf', 'Pagination', 'WProofreader', 'MathType' ],
```

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,770 @@
...
+		toolbar: {
+			items: [
+				'undo', 'redo',
+				'|',
+				'exportPdf', 'exportWord',
```
Mgsy, line 135: Exports are removed in this snippet, so the toolbar shouldn't contain items related to these plugins.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,770 @@
...
+</head>
+<body>
+	<h1>Full-featured editor</h1>
+	<div id="editor">
+		<p>This is some sample content.</p>
+	</div>
+	<script>
+		CKEDITOR.ClassicEditor
+			.create( document.querySelector( '#editor' ), {
+				removePlugins: [ 'Comments, RealTimeCollaborativeComments, TrackChanges, RealTimeCollaborativeTrackChanges, RevisionHistory, RealTimeCollaborativeRevisionHistory, RealTimeCollaborativeEditing, ExportWord, ExportPdf, Pagination, WProofreader, MathType' ],
```
Mgsy, line 258: We should update this part. thesame way as above.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,770 @@
...
+				toolbar: {
+					items: [
+						'undo', 'redo',
+						'|',
+						'exportPdf', 'exportWord',
```
Mgsy, line 263: We should update this part. thesame way as above.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,770 @@
...
+						'|',
+						'wproofreader', 'findAndReplace', 'selectAll',
```
Mgsy, line 265: We should update this part same way as above.

docs/installation/getting-started/quick-start.md
godai78: Aren't exports available in this version?

docs/installation/getting-started/quick-start.md
godai78: Addressed in commit.

docs/installation/getting-started/quick-start.md
godai78: Addressed in commit.

docs/installation/getting-started/quick-start.md
godai78: Addressed in commit.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,770 @@
...
-	.then( editor => {
-		console.log( 'Editor was initialized', editor );
+		removePlugins: [ 'Comments, RealTimeCollaborativeComments, TrackChanges, RealTimeCollaborativeTrackChanges, RevisionHistory, RealTimeCollaborativeRevisionHistory, RealTimeCollaborativeEditing, ExportWord, ExportPdf, Pagination, WProofreader, MathType' ],
+		toolbar: {
+			items: [
+				'undo', 'redo',
+				'|',
+				'exportPdf', 'exportWord',
+				'|',
+				'wproofreader', 'findAndReplace', 'selectAll',
```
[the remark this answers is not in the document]
godai78: Addressed in commit.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,1720 @@
...
+		table: {
+			contentToolbar: [
+				'tableColumn', 'tableRow', 'mergeTableCells', 'tableProperties', 'tableCellProperties', 'toggleTableCaption'
+			]
+		},
+	} )
+	.catch( error => {
+		console.log( error );
+	} );
+```
```
dufipl, line 248: In general, I agree that step by step attitude and in the end presenting full implementation is good, but in this case, pasting a large amount of source code to say that plugins can be removed and just after that place another large amount of almost the same code with full implementation looks like too much. 
I would suggest removing this snippet and within the part about eliminating plugins and customising the toolbar refer to the full implementation below.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,1720 @@
...
-	webpack@5 \
-	webpack-cli@4
+```html
+<script>
+	CKEDITOR.ClassicEditor
+		.create( document.querySelector( '#editor' ) )
+		.catch( error => {
+			console.error( error );
+		} );
+</script>
```
dufipl, line 130: This section suggests that doing those steps is enough to make the editor work (which is true for the previous tutorial about using CDN, it is possible to just paste similar lines to an HTML document and it is working), for this guide it is not true and misleading. When you paste just those steps you start receiving errors about missing config of multiple plugins. e.g. RH. I see two options here: delete those steps and refer to full examples or make a disclaimer that due to the complex character of the superbuild there are more configurations to do which are explained in detail below. At this stage, it might be very confusing for readers who do not have experience with editors and want to give it a try. 

I would go with the attitude to include a full fully configured example of an editor without premium features and then an example with premium features. We should focus on quick experience first and then go deep into details.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,1720 @@
...
+
+		toolbarContainer.appendChild( editor.ui.view.toolbar.element );
 	} )
 	.catch( error => {
-		console.error( error.stack );
+		console.log( error );
 	} );
 ```
 
-You can now run webpack to build the application. To do that, call the `webpack` executable:
```
dufipl, line 237: As above, the unnecessary huge amount of code and repetition of most of it just behind, I think we might delete this and refer to the main example.

docs/installation/getting-started/quick-start.md
```diff
@@ -4,318 +4,1720 @@
...
+	} )
+	.then( editor => {
+		const toolbarContainer = document.querySelector( '.editor-toolbar' );
+
+		toolbarContainer.appendChild( editor.ui.view.toolbar.element );
+	} )
+	.catch( error => {
+		console.log( error );
+	} );
+```
```
dufipl, line 1344: As above, the unnecessary huge amount of code and repetition of most of it just behind, I think we might delete this and refer to the main example explaining the details.

dufipl (changes requested):

docs/installation/getting-started/quick-start.md
godai78: @dufipl - please compare the solution we proposed with collapsible code listings.

PR #11976, opened 2022-06-27
Updated editor's textbox label

### LINKED ISSUE #11836: Repetitive and non-informative editor's label

1.  Open https://ckeditor.com/docs/ckeditor5/latest/examples/builds/classic-editor.html
2.  Launch VoiceOver (or some other screen reader).
3.  Focus the editor.

### ✔️ Expected result

The editor is announced in a concise and non-redundant way.

### ❌ Actual result

The editor is announced as "Rich Text Editor, \<root name>, edit text. \<content and selection info>. Rich Text Editor, application.

### ❓ Possible solution

Both labels – for the application region and editable ones – should be different and easily differentiated. See [CKE4's implementation](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_a11y.html#announcing-the-editor-on-the-page) as an example.

## 📃 Other details

*   Browser: all
*   OS: all
*   Installed CKEditor plugins: N/A

---

mateuszzagorski:

Fix (editor-balloon,editor-inline,editor-classic,editor-decoupled,ui): The editing area of the editor should have an accessible informative label. Closes #11836.


oleq (changes requested):
I'd avoid passing a reference to the entire UI of the editor the inline editable view class. These things are orthogonal and the latter is just a building block that should not be aware of the bigger picture. For instance, this creates a strange logic like this one https://github.com/ckeditor/ckeditor5/pull/11976/files#diff-a8d5a9f1066f55f490c2415d80b9eac40e95e853a3535d02062b2244a17cf29cR59

We don't know what the implementation of the editor view looks like (it's up to the integrator) and we cannot presume anything based on the presence or absence of the DOM element.

Instead, I'd pass a label generator down the structure. The integration (type of the editor) decides what sort of label should be used because it knows the full picture. I PoCed this in a fork. You can see what it looks like here https://github.com/ckeditor/ckeditor5/compare/ck/11836-repetetive-editor-label...ck/11836-repetetive-editor-label-v2?expand=1.

In the PoC I also streamlined the labels to be more informative. For classic editor:
Screenshot 2022-07-08 at 14 53 09


For balloon editor:
Screenshot 2022-07-08 at 14 53 21

It would be great if you could take it over from here. Things left to do:
* Other kinds of editor (only classic and balloon covered in the PoC)
* Tests. There are no tests for individual editors, only for the UI in this PR.
* Entries in `contexts.json` in ckeditor5-ui.
* Probably some update to the multi root guide that we have in our docs.  

Good luck 🙂

oleq:
The PR is on review again.

niegowski (commented):
Looks good. Also, remember to test the classic editor.

PR #11980, opened 2022-06-28
Introducing the ImageInsertViaUrl plugin enabling insertion of images via URL without the requirement of the UploadPlugin

### LINKED ISSUE #11654: Allow using images from external source without allowing image upload

The CKEditor 5 `ImageInsert` plugin has a dependency on `ImageUpload` which prevents configuring CKEditor 5 in a way that it would be possible to allow using images from external source without allowing image upload. 

Ideally all of these combinations would be possible:

1. Image insert from external URL
2. Image upload
3. Image insert from external URL + Image upload

Right now, only options 2 and 3 are possible. What Drupal needs for providing an upgrade path from its CKEditor 4 implementation is options 1 and 2.

The dependency seems to be on the UI level, meaning that it should be straight forward to make `ImageInsert` plugin work without `ImageUpload`. A proof of concept of this was implemented in https://github.com/ckeditor/ckeditor5/pull/11571.

However, if we simply remove the dependency `ImageInsert` plugins dependency on `ImageUpload`, it does introduce a BC break where all instances configured to use `ImageInsert` without explicitly also depending on `ImageUpload`, would see a change in behavior.

---

niegowski:

Other (image): Introducing the `ImageInsertViaUrl` plugin enabling insertion of images via URL without the requirement of the `UploadPlugin`. Closes #11654.

MINOR BREAKING CHANGE (image): `ImageInsertPanelView#dropdownView` property is no longer accessible. Use `editor.plugins.get( 'ImageInsertUI' ).dropdownView` instead to access the image insert dropdown instance. See #11654.

---

### Additional information

It's a replacement for the `ImageInsert` plugin which does not require `ImageUpload`.

### DIFF
--- a/packages/ckeditor5-image/src/imageinsert.js
+++ b/packages/ckeditor5-image/src/imageinsert.js
@@ -9,6 +9,7 @@
 
 import { Plugin } from 'ckeditor5/src/core';
 import ImageUpload from './imageupload';
+import ImageInsertViaUrl from './imageinsertviaurl';
 import ImageInsertUI from './imageinsert/imageinsertui';
 
 /**
@@ -37,7 +38,7 @@ export default class ImageInsert extends Plugin {
 	 * @inheritDoc
 	 */
 	static get requires() {
-		return [ ImageUpload, ImageInsertUI ];
+		return [ ImageUpload, ImageInsertViaUrl, ImageInsertUI ];
 	}
 }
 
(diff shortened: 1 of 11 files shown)


oleq (approved):
LGTM. FYI I pushed a commit that untangles the weird ownership graph of the UI components.

PR #11998, opened 2022-06-30
Editor's application unique label

### LINKED ISSUE #11863: Make editor's application labels unique

1. Open https://ckeditor.com/docs/ckeditor5/latest/features/lists/lists.html
2. Open VoiceOver.
3. Open [rotor](https://support.apple.com/en-gb/guide/voiceover/mchlp2719/mac) and switch to landmarks list.

### ✔️ Expected result

Every editor has a unique name on the list displayed by the rotor.

### ❌ Actual result

Every editor has the same name, "Rich Text Editor".

### ❓ Possible solution

Add the editor name after the static part of the label: "Rich Text Editor, &lt;editor name&gt;".

## 📃 Other details

* Browser: all (tested on Chrome)
* OS: all

The analogous issue is present in other screen readers.

---

mateuszzagorski:

Fix (core, editor-classic, editor-inline, editor-decoupled, editor-balloon, ui): Every editor should have a unique application label. Closes #11863.

---

### Additional information

The label is based on the editor's config 'label' value. If not present, it will use a name pattern of \`'editor{n}'\`.


packages/ckeditor5-editor-classic/src/classiceditoruiview.js
```diff
@@ -24,12 +24,13 @@ export default class ClassicEditorUIView extends BoxedEditorUIView {
 	 * @param {module:utils/locale~Locale} locale The {@link module:core/editor/editor~Editor#locale} instance.
 	 * @param {module:engine/view/view~View} editingView The editing view instance this view is related to.
 	 * @param {Object} [options={}] Configuration options for the view instance.
+	 * @param {String} [options.editorName] Editor's name.
```
oleq, line 27: Where it is displayed? What it is for?

packages/ckeditor5-ui/src/editorui/boxed/boxededitoruiview.js
```diff
@@ -21,8 +21,9 @@ export default class BoxedEditorUIView extends EditorUIView {
 	 * Creates an instance of the boxed editor UI view class.
 	 *
 	 * @param {module:utils/locale~Locale} locale The locale instance..
+	 * @param {Object} [options]
```
oleq, line 24: Missing docs for `options.editorName`.

packages/ckeditor5-ui/src/editorui/boxed/boxededitoruiview.js
```diff
@@ -43,6 +44,13 @@ export default class BoxedEditorUIView extends EditorUIView {
 		 */
 		this.main = this.createCollection();
 
+		/**
+		 * Editor's title based on the passed object options.editorName.
```
oleq, line 48:
```suggestion
		 * Editor's label defined by the `options` passed to the constructor . 
		 * Used to generate the label used by assistive technologies.
```

packages/ckeditor5-ui/src/editorui/boxed/boxededitoruiview.js
```diff
@@ -107,7 +115,11 @@ export default class BoxedEditorUIView extends EditorUIView {
 		const t = this.t;
 		const voiceLabel = new LabelView();
 
-		voiceLabel.text = t( 'Rich Text Editor' );
+		if ( this.label && this.label !== '' ) {
```
oleq, line 118: If `editorName` was not in `options` the `#label` is `''`. If it was passed in `options`, let's assume it was not `''` and we should use its value as it was passed. This will simplify the code below. This will always come down to `voiceLabel.text = t( 'Rich Text Editor, ' + this.label );`

packages/ckeditor5-ui/src/editorui/boxed/boxededitoruiview.js
```diff
@@ -107,7 +115,11 @@ export default class BoxedEditorUIView extends EditorUIView {
...
+			voiceLabel.text = t( 'Rich Text Editor, ' + this.label );
+		} else {
+			voiceLabel.text = t( 'Rich Text Editor' );
+		}
```
oleq, line 122: `BoxedEditorUIView` tests missing for this logic.

packages/ckeditor5-ui/src/editorui/boxed/boxededitoruiview.js
```diff
@@ -43,6 +44,13 @@ export default class BoxedEditorUIView extends EditorUIView {
 		 */
 		this.main = this.createCollection();
 
+		/**
+		 * Editor's title based on the passed object options.editorName.
+		 *
+		 * @readonly
+		 */
+		this.label = options ? options.editorName : '';
```
oleq, line 52: *   `BoxedEditorUIView` tests missing for this logic.
*   Also, does this thing need to be public? I see no good reason to expose it yet.
*   Also, let's name it `_editorName` maybe (?). No reason to rename it here, the option is pretty clear as to what information it carries.

packages/ckeditor5-ui/src/editorui/boxed/boxededitoruiview.js
oleq: Let's try to say not only what it is but also what it is for.

packages/ckeditor5-editor-classic/src/classiceditor.js
```diff
@@ -77,8 +77,9 @@ export default class ClassicEditor extends Editor {
 		this.model.document.createRoot();
 
 		const shouldToolbarGroupWhenFull = !this.config.get( 'toolbar.shouldNotGroupWhenFull' );
+		const editorName = this.name;
 		const view = new ClassicEditorUIView( this.locale, this.editing.view, {
-			shouldToolbarGroupWhenFull
+			shouldToolbarGroupWhenFull, editorName
```
oleq, line 82:
```suggestion
			shouldToolbarGroupWhenFull, 
			editorName: this._name
```

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -220,6 +220,31 @@ export default class Editor {
...
+		 */
+		Editor._editorCount = Editor._editorCount ? ++Editor._editorCount : 1;
+
+		/**
+		 * A unique editor's name.
+		 *
+		 * @observable
+		 * @member {String} #name
+		 */
+		this.name = this.generateEditorLabel();
```
oleq, line 238: cc @Reinmar Are we OK with this kind of property on every editor instance? Could be helpful for some people but this will stay forever.

*   Also, what about the name of this property?

If not OK, let's move this whole label generation logic to `EditorUIView` maybe and hide it from the outside world.

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -433,6 +458,21 @@ export default class Editor {
...
+	 * Generates a label for the editor.
+	 *
+	 * It will be based on the editor.config.get( 'label' ), otherwise a name
+	 * pattern of `'editor{n}'` will be used.
+	 *
+	 * @readonly
+	 * @returns {String} label
+	 */
+	generateEditorLabel() {
+		const label = this.config.get( 'label' ) ? this.config.get( 'label' ) : 'editor' + Editor.editorCount;
```
oleq, line 471: There's no documentation for `label` in `EditorConfig`. This a critical API for integrators and must be documented.

packages/ckeditor5-editor-classic/src/classiceditoruiview.js
```diff
@@ -24,12 +24,13 @@ export default class ClassicEditorUIView extends BoxedEditorUIView {
...
 	 * @param {Boolean} [options.shouldToolbarGroupWhenFull] When set `true` enables automatic items grouping
 	 * in the main {@link module:editor-classic/classiceditoruiview~ClassicEditorUIView#toolbar toolbar}.
 	 * See {@link module:ui/toolbar/toolbarview~ToolbarOptions#shouldGroupWhenFull} to learn more.
 	 */
 	constructor( locale, editingView, options = {} ) {
-		super( locale );
+		super( locale, { editorName: options.editorName } );
```
oleq, line 33: Tests missing for `ClassicEditorUIView` to verify this is passed correctly.

packages/ckeditor5-editor-classic/src/classiceditor.js
oleq, line 82: Tests missing for `ClassicEditor` that would verify that `editorName` is passed to its view.

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -220,6 +220,31 @@ export default class Editor {
...
 		this.keystrokes = new EditingKeystrokeHandler( this );
 		this.keystrokes.listenTo( this.editing.view.document );
+
+		/**
+		 * Current number of editors instances created.
+		 *
+		 * @private
+		 * @type {Number}
+		 */
+		Editor._editorCount = Editor._editorCount ? ++Editor._editorCount : 1;
```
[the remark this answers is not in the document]
oleq: I'd solve this file like this

```javascript
diff --git a/packages/ckeditor5-core/src/editor/editor.js b/packages/ckeditor5-core/src/editor/editor.js
index 5f1d06d387..3988704360 100644
--- a/packages/ckeditor5-core/src/editor/editor.js
+++ b/packages/ckeditor5-core/src/editor/editor.js
@@ -221,30 +221,13 @@ export default class Editor {
         this.keystrokes = new EditingKeystrokeHandler( this );
         this.keystrokes.listenTo( this.editing.view.document );

-        /**
-         * Current number of editors instances created.
-         *
-         * @private
-         * @type {Number}
-         */
-        Editor._editorCount = Editor._editorCount ? ++Editor._editorCount : 1;
-
         /**
          * A unique editor's name.
          *
          * @observable
          * @member {String} #name
          */
-        this.name = this.generateEditorLabel();
-    }
-
-    /**
-     *  Static getter for the _editorCount property.
-     *
-     * @returns {module:core/editor/editor~Editor._editorCount}
-     */
-    static get editorCount() {
-        return Editor._editorCount;
+        this.name = this._generateEditorLabel();
     }

     /**
@@ -464,15 +447,22 @@ export default class Editor {
      * It will be based on the editor.config.get( 'label' ), otherwise a name
      * pattern of `'editor{n}'` will be used.
      *
-     * @readonly
+     * @private
      * @returns {String} label
      */
-    generateEditorLabel() {
-        const label = this.config.get( 'label' ) ? this.config.get( 'label' ) : 'editor' + Editor.editorCount;
+    _generateEditorLabel() {
+        const label = this.config.get( 'label' ) ? this.config.get( 'label' ) : 'editor' + ++Editor._editorCount;

         return label;
     }

+    /**
+     * Editor instance counter for labeling and accessibility.
+     *
+     * @private
+     * @property {number} #_editorCount
+     */
+
     /**
      * Creates and initializes a new editor instance.
      *
@@ -490,6 +480,8 @@ export default class Editor {
      */
 }

+Editor._editorCount = 0;
+
 mix( Editor, ObservableMixin );

 /**
```

packages/ckeditor5-core/src/editor/editor.js
oleq: Regardless of 👆 the docs must explain more about what it is and where it is used and why.

packages/ckeditor5-core/src/editor/editor.js
```diff
@@ -220,6 +220,31 @@ export default class Editor {
...
+
+		/**
+		 * A unique editor's name.
+		 *
+		 * @observable
```
oleq, line 235: This property is not observable. It should be readonly, though.

oleq (changes requested):
We could use a manual test with multiple editors of a different kind here.

Also, what about other editor types? Inline, Balloon, etc.?

packages/ckeditor5-core/src/editor/editor.js
Reinmar: As for the naming of this property. If we move it to `EditorUI` it could be renamed to `editor.ui.editorName`. And we should explain it in the comments that it's related to a11y and how it's set (via a config prop?).

mateuszzagorski:
I've replaced the initially added all-editors manual test, with the one proposed by Oleq in #12064.

tony1377:
Is there a known ETA on this fix?

oleq (changes requested):
Quoting myself from the last [review](https://github.com/ckeditor/ckeditor5/pull/11998#pullrequestreview-1028661925):

> Also, what about other editor types? Inline, Balloon, etc.?

PR #11999, opened 2022-06-30
Handle call to navigator API for envs without navigator API.

CatStrategist:

Internal (utils): Wrapped accessing browser's API in try/catch blocks to safely initialize editor in environments without browser's APIs.

---

### Additional information

`env.ts`, `global.ts` and `translation-service.ts` files might be imported in environments without browsers' APIs - like Node.
The best solution would be to get it out of a global scope to a function call, so it's not executed on the file import, but that would require a change in fifty files wherever it's imported.


Reinmar:
This change requires tests. How do we know that it actually doesn't crash later on when the string is empty?

Also, perhaps it'd be better if a "no browser" tried to mimick some known browser? Chrome for instance as it's most popular?

Issue #11851, opened 2022-05-30
Discovery: research keyboard handling for grid components


mlewand:
## Provide a description of the task

Research best practices on how to handle grid navigation using the keyboard.

We have multiple grid-like UI components:

*   [Colors in the font color dropdown.](https://github.com/ckeditor/ckeditor5/issues/11841)
*   [List styles (in a dropdown)](https://github.com/ckeditor/ckeditor5/issues/11041).
*   [Table dimension grid in insert table button dropdown.](https://github.com/ckeditor/ckeditor5/issues/3176)
*   [Special characters](https://github.com/ckeditor/ckeditor5/issues/9037).

To make it clear: this issue is just discovery. It's meant just to figure out best way to implement it and see why we have it implemented the way we did. We also can make some PoC to see how much effort will it require to fix it.

Once we have figured how this should be solved, we can move on to each separate issue.

mlewand:
There's a curious thing in how we handle grids as _some_ of them have not predefined columns.

For example, list styles have responsive grid, based only on CSS. However font colors do have predefined columns count.

mlewand:
As far as the implementation goes we might consider doing it as a helper rather than making this feature a UI component class.

An example of such composition-based approach is click outside handler: https://github.com/ckeditor/ckeditor5/blob/30286f77b39526fce2856b03b9be0ba4cc91d1c7/packages/ckeditor5-ui/src/bindings/clickoutsidehandler.js#L27

oleq:
For the record, it looks like the table insertion grid should handle keyboard differently because when reaching the right boundary of the grid, the right arrow key should not move focus back to the first column (which would be expected in the color grid).

mmotyczynska:
Yet another approach could be using a FocusCycler as it's already partly used in grid-based features. We'd have to enhance it for grids so e.g. the next/previous methods would 'know' that at some point a row of elements ends and we have to go to the beginning/end of the row instead of to the next/previous element. And of course we'd have to add vertical navigation as well for handling arrow up/down events.

Reinmar:
BTW: https://github.com/ckeditor/ckeditor5/issues/5772

mmotyczynska:
### Navigation (tab vs arrow keys):

There should be two different navigation handled (they shouldn’t be mixed together):

*   Between sections in the dropdown (tab)
*   Between items of the grid (arrow keys)

[...]

Issue #5859, opened 2019-11-28
The selected language in code block dropdown is not visible


jodator:
## 📝 Provide detailed reproduction steps (if any)

1. Open code block feature guide
2. Place caret in either JavaScript or HTML/XML block
3. Open code block dropdown

### ✔️ Expected result

A selected language is visible (is scrolled to if possible)

### ❌ Actual result

The active dropdown action is not visible.

## 📃 Other details

![Selection_075](https://user-images.githubusercontent.com/247363/69799859-609f1a00-11d4-11ea-9a2a-a21c270dc2b9.png)

* Browser: Safari, Firefox probably Chrome also
* CKEditor version: latest master

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

jodator:
It's worse on Safari as they don't render the scrollbar there and the list looks like finite:

![Zrzut ekranu 2019-11-28 11 55 25](https://user-images.githubusercontent.com/247363/69800925-7c0b2480-11d6-11ea-8f75-07be4c1bb392.png)

bbenjamin:
Something to note, that may mean this is easier to solve: This is a problem on pages that have loaded with existing code blocks. If a new code block is added, or the language in an existing code block is changed, then the active language is automatically scrolled to when the select is opened. This can be reproduced on https://ckeditor.com/docs/ckeditor5/latest/features/code-blocks.html

Issue #1037, opened 2018-05-23
Improve accessibility of drop–downs


oleq:
1.  The screen reader should properly announce that the button has a popup, (#12258)
    1.  ~The screen reader should explain the way to open it, (this will also be handled by #12258)~
2.  ~The screen reader focus should move properly to the drop–down and follow the navigation across its content~ (#11838)
3.  ~Exiting the drop–down should also be explained by the text reader.~ (not needed as [discussed later](https://github.com/ckeditor/ckeditor5/issues/1037#issuecomment-1147652455))

oleq:
1. Our dropdowns have no `aria-expanded` but should have (see an example from w3c https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html)
   * The only places that have that attribute are CollapsibleView (list properties: start index, reversed section) and SplitButtonView (a button for dropdowns with an “extra arrow”).

wimleers:
AFAICT this overlaps with #11568, #11041, #10368, #3184, #2000, #3215, #3185, #3176 and #11838.

What seems _different_ about this ticket is that it seems to stress the _screen reader_ aspect.

So I propose to retitle/rescope this to `Improve screenreader support of dropdowns`, and to do this _after_ it's been made keyboard-accessible. Doing that first is necessary anyway, and may solve some of the problems outlined here 😊

Comandeer:
> The screen reader should properly announce that the button has a popup,

It can be done by adding the [`[aria-haspopup]`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup) and [`[aria-expanded]`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) attributes to buttons triggering dropdowns. Depending on what the dropdown is, the `[aria-haspopup]` attribute can take different values instead of generic `true` (e.g. `menu` if the dropdown is some kind of a menu). We also need to ensure that the dropdown itself has an appropriate role (e.g. `[role=menu]` in case of menu dropdown). ARIA Practices contains a decent [menu button sample](https://www.w3.org/WAI/ARIA/apg/example-index/menu-button/menu-button-actions.html).

> The screen reader should explain the way to open it,

Using the combination of `[aria-haspopup]` and `[aria-expanded]` attributes should be enough for that.

> The screen reader focus should move properly to the drop–down and follow the navigation across its content,

[...]

PR #12003, opened 2022-07-01
Implemented focusing first item after opening a dropdown

### LINKED ISSUE #2000: What should happen with the focus after someone opened a dropdown (with the mouse)?
Currently, if you open a dropdown **with your mouse** the focus stays in the editor (of course, it behaves differently when you first focused the toolbar with alt+f10):

image

Which means that we have issues like https://github.com/ckeditor/ckeditor5-font/issues/54.

While the above mentioned problem is indeed an issue and the solution to it seems trivial – let's close the dropdown when the selection changes (CKEditor 4 does that), I'd like to know what should actually happen with focus. Because, I can imagine that it's more useful for the focus to be moved to the dropdown so you can use your keyboard to navigate there (may be useful sometimes).

cc @Comandeer

mlewand:

Fix (ui): Opening a dropdown will cause the first enabled item to be focused. If there are no enabled items inside it will focus just the first item. Closes #2000.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ui/src/dropdown/dropdownview.js
+++ b/packages/ckeditor5-ui/src/dropdown/dropdownview.js
@@ -265,6 +265,15 @@ export default class DropdownView extends View {
 			}
 		} );
 
+		// Focus the first item in the dropdown when the dropdown opened
+		this.on( 'change:isOpen', () => {
+			if ( !this.isOpen ) {
+				return;
+			}
+
+			this.panelView.focus();
+		} );
+
 		// Listen for keystrokes coming from within #element.
 		this.keystrokes.listenTo( this.element );
 
--- a/packages/ckeditor5-ui/src/dropdown/utils.js
+++ b/packages/ckeditor5-ui/src/dropdown/utils.js
@@ -127,8 +127,10 @@ export function createDropdown( locale, ButtonClass = DropdownButtonView ) {
  *
  * @param {module:ui/dropdown/dropdownview~DropdownView} dropdownView A dropdown instance to which `ToolbarView` will be added.
  * @param {Iterable.<module:ui/button/buttonview~ButtonView>} buttons
+ * @param {Object} [options]
+ * @param {Boolean} [options.skipAutoFocus=false]
  */
-export function addToolbarToDropdown( dropdownView, buttons ) {
+export function addToolbarToDropdown( dropdownView, buttons, options = {} ) {
 	const locale = dropdownView.locale;
 	const t = locale.t;
 	const toolbarView = dropdownView.toolbarView = new ToolbarView( locale );
@@ -143,19 +145,21 @@ export function addToolbarToDropdown( dropdownView, buttons ) {
 
 	buttons.map( view => toolbarView.items.add( view ) );
 
-	dropdownView.on( 'change:isOpen', () => {
-		// A listener to focus the first active dropdown option upon opening, see #11838.
-		if ( !dropdownView.isOpen ) {
-			return;
-		}
-
-		for ( const toolbarItem of toolbarView.items ) {
-			if ( toolbarItem.isOn ) {
-				toolbarItem.focus();
-				break;
+	if ( !options.skipAutoFocus ) {
+		dropdownView.on( 'change:isOpen', () => {
+			// A listener to focus the first active dropdown option upon opening, see #11838.
+			if ( !dropdownView.isOpen ) {
+				return;
 			}
-		}
-	}, { priority: 'low' } );
+
+			for ( const toolbarItem of toolbarView.items ) {
+				if ( toolbarItem.isOn ) {
+					toolbarItem.focus();
+					break;
+				}
+			}
+		}, { priority: 'low' } );
+	}
 
 	dropdownView.panelView.children.add( toolbarView );
 	toolbarView.items.delegate( 'execute' ).to( dropdownView );
--- a/packages/ckeditor5-ui/src/dropdown/utils.js
+++ b/packages/ckeditor5-ui/src/dropdown/utils.js
@@ -143,6 +143,20 @@ export function addToolbarToDropdown( dropdownView, buttons ) {
 
 	buttons.map( view => toolbarView.items.add( view ) );
 
+	dropdownView.on( 'change:isOpen', () => {
+		// A listener to focus the first active dropdown option upon opening, see #11838.
+		if ( !dropdownView.isOpen ) {
+			return;
+		}
+
+		for ( const toolbarItem of toolbarView.items ) {
+			if ( toolbarItem.isOn ) {
+				toolbarItem.focus();
+				break;
+			}
+		}
+	}, { priority: 'low' } );
+
 	dropdownView.panelView.children.add( toolbarView );
 	toolbarView.items.delegate( 'execute' ).to( dropdownView );
 }
@@ -227,11 +241,7 @@ export function addListToDropdown( dropdownView, items ) {
 
 			if ( item.model && item.model.isOn ) {
 				const listItemView = listView.items.get( i );
-
-				// Calling focus tracker to focus the view is not enough. It sets the focused element
-				// internally, but doesn't actually call focus method on a given view.
-				listView.focusTracker._focus( listItemView );
-				listView._focusCycler._focus( listItemView );
+				listItemView.focus();
 				break;
 			}
 		}
--- a/packages/ckeditor5-ui/src/dropdown/utils.js
+++ b/packages/ckeditor5-ui/src/dropdown/utils.js
@@ -216,6 +216,27 @@ export function addListToDropdown( dropdownView, items ) {
 		}
 	} );
 
+	dropdownView.on( 'change:isOpen', () => {
+		// A listener to focus the first active dropdown option upon opening, see #11838.
+		if ( !dropdownView.isOpen ) {
+			return;
+		}
+
+		for ( let i = 0; i < items.length; i++ ) {
+			const item = items.get( i );
+
+			if ( item.model && item.model.isOn ) {
+				const listItemView = listView.items.get( i );
+
+				// Calling focus tracker to focus the view is not enough. It sets the focused element
+				// internally, but doesn't actually call focus method on a given view.
+				listView.focusTracker._focus( listItemView );
+				listView._focusCycler._focus( listItemView );
+				break;
+			}
+		}
+	}, { priority: 'low' } );
+
 	dropdownView.panelView.children.add( listView );
 
 	listView.items.delegate( 'execute' ).to( dropdownView );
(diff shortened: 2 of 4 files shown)


mlewand:
@mmotyczynska based on your branch I added behavior for toolbar dropdowns and to prioritize focusing the active option if possible.

There's some failing tests that probably were written on a little bit more integration rather than unit test side. So this needs to be addressed.

Furthermore, there's insufficient code coverage so that needs to be fixed too.

Finally, there are 2 known uncovered cases which should be extracted to a followup:

*   list style dropdown is not affected by this change (because it's implemented in a much custom way) - it needs a follow up issue
*   font color (and font bg color) dropdown - also requires a followup

mlewand:
Also, @ckeditor/qa-team can you test CF UIs with this change? To make sure that it doesn't break anything (less likely) and still make sense.

dufipl:
cc @mlewand @mmotyczynska 
## Issue with selection outside of the editor after accepting/discarding suggestions in TC
### Steps

1. Open http://localhost:8125/ckeditor5/external/collaboration-features/packages/ckeditor5-revision-history/tests/manual/revision-history-autosave.html
2. Enable Track Changes
3. Create a suggestion
4. Click on the TC dropdown and choose "Accept all suggestions"

### Expected result:
Focus gets back to the editor, the user can write instantly, `ctrl+z` is possible without manually focusing on the editor:

### Actual result:
Focus does not get back to the editor

### Additional details
It is a regression compared to the `master` branch.

dufipl:
cc @mlewand @mmotyczynska simpler case:
## Issue with selection outside of the editor after enabling TC
### Steps

1. Open http://localhost:8125/ckeditor5/external/collaboration-features/packages/ckeditor5-revision-history/tests/manual/revision-history-autosave.html
2. Enable Track Changes

### Expected result:
Focus inside of the editor, the user can type after enablement. 

### Actual result:
Focus does not get back to the editor
Focus moved to the toolbar element: 
Zrzut ekranu 2022-07-4 o 15 43 05


### Additional details
It is a regression compared to the `master` branch.

oleq (changes requested):
I think we're getting close with this task 🙂

I noticed you didn't take advantage of the new `addToolbarToDropdown()` API, though. I decided to give it a try.  I wasn't sure if this was the right way thing to do so sorry for the poor communication. 

Upon closer inspection I unified the focus-on-open logic under a helper that gets used internally in `addToolbarToDropdown()` and `addListToDropdown()` as well as in some features that need a custom handling (list style and font color). You can see how it looks in a [fork of this PR](https://github.com/ckeditor/ckeditor5/compare/ck/11838-placement-of-focus-after-opening-dropdowns...ck/11838-placement-of-focus-after-opening-dropdowns-with-helper?expand=1). Feel free to merge it here if you find it acceptable (it will need tests, though).

## Things left to do:

* I noticed a glitch when clicking one dropdown of the balloon toolbar and then another. The latter won't open and the toolbar hides. This must be addressed. My gut tells me this could be due to the fact that until this PR, there was no way to focus anything in this kind of toolbar. It could be that something was never registered in the focus tracker because if you attach this listener you'll find that the entire editor gets blurred in the process
```js
editor.ui.focusTracker.on( 'change:isFocused', () =>{
    console.log( editor.ui.focusTracker.isFocused );
});
```


**Update: I suppose the problem is somewhere else.** 

Add the following listeners:

```js
window.addEventListener('focus', () => { console.log( 'activeElement', document.activeElement ) }, true)
window.addEventListener('blur', () => { console.log( 'activeElement', document.activeElement ) }, true)
```

and the following style:


```css
*:focus {
    outline: 3px dashed red !important;
    outline-offset: -1px !important;
}
```

and see what it looks like on master and then in this PR. The thing is that on master, the focus always stays in the root as you click across the UI (e.g. open one dropdown and then another). In this PR however, if you already got one dropdown open it takes the focus, and when you click another, the click outside handler logic kicks in 

https://github.com/ckeditor/ckeditor5/blob/1f40c3e11889df2f4ed413337d90829fd2393b1d/packages/ckeditor5-ui/src/dropdown/utils.js#L275-L286

To put it shortly, this logic is responsible for closing the dropdown when somebody click outside of it (on `mousedown`). When the dropdown is open, it holds the focus of the entire editor. When you click on another dropdown (hold the button down to see this better), this logic kicks in and the dropdown closes. The focal point of the entire editor gets hidden as the panel goes off the screen and the browser, seeing that the thing that used to be focused (a button) is no longer visible, moves the focus to `<body>`. `editor.ui.focusTracker` seeing the focus went to `<body>` thinks the editor was blurred and all sorts of UI (**including all balloons**) hide upon that. That's why you can't click between dropdowns in the widget toolbar (or any ballon for what is worth) but it kinda works in the classic editor toolbar (although notice that the `editor.focusTracker.isFocused` goes in and out and the editing root's outline is blinking).

I'm not sure how to address this yet, though. Maybe we should move the focus somewhere else before `dropdownView.isOpen = false;`?

```diff
diff --git a/packages/ckeditor5-ui/src/dropdown/utils.js b/packages/ckeditor5-ui/src/dropdown/utils.js
index 78b80d6442..1ace5146bc 100644
--- a/packages/ckeditor5-ui/src/dropdown/utils.js
+++ b/packages/ckeditor5-ui/src/dropdown/utils.js
@@ -278,6 +278,7 @@ function closeDropdownOnBlur( dropdownView ) {
                            emitter: dropdownView,
                            activator: () => dropdownView.isOpen,
                            callback: () => {
+                                    dropdownView.buttonView.focus();
                                     dropdownView.isOpen = false;
                            },
                            contextElements: [ dropdownView.element ]
```

seems to help but it looks weird. I think it would we better if we allowed all toolbar buttons to be focusable on click (ATM [we prevent that](https://github.com/ckeditor/ckeditor5/blob/30286f77b39526fce2856b03b9be0ba4cc91d1c7/packages/ckeditor5-ui/src/button/buttonview.js#L154-L156)). There would be no interruption in the focus then.

* (cc @Reinmar) I'm worried about integrators and their `View` instances that do not come with the `focus()` method. When in the `DropdownView` we call `this.panelView.focus();` and the panel does `this.children.first.focus();` we assume that the first child has the `focus()` method. If not, there will be a JS crash. And because this will happen upon opening of every dropdown (lots of points of failure), I see two approaches:
   1. Implement `View#focus()` in `view.js` that will `logWarning()` saying that ~"We were about to focus this view but it looks like it doesn't have the `focus()` method implementation.". No crash will follow. We'll have to disable the warning in case someone calls `super()` in their `focus()` implementation but I suppose something like ` if (this.focus === View.prototype.focus) {}` should do the trick.
   2. Check if `this.children.first.focus` in the `DropdownPanelView` and `childToFocus.focus` in `focusChildOnDropdownOpen()` exist. It's cheaper and easier but it secures the dropdown ecosystem only. If we progress with the a11y we may find out there are more systems that require smart automatic focus movement and other views will throw because they don't have the `focus()` interface.

oleq:
An update regarding 

> I noticed a glitch when clicking one dropdown of the balloon toolbar and then another. \[...\]

Together with @Reinmar we decided to ditch

[https://github.com/ckeditor/ckeditor5/blob/30286f77b39526fce2856b03b9be0ba4cc91d1c7/packages/ckeditor5-ui/src/button/buttonview.js#L154-L156](https://github.com/ckeditor/ckeditor5/blob/30286f77b39526fce2856b03b9be0ba4cc91d1c7/packages/ckeditor5-ui/src/button/buttonview.js#L154-L156)

and make all buttons focusable. Let's see how it works during QA session.

---

An update regarding

> I'm worried about integrators and their `View` instances that do not come with the `focus()` method

Together with @Reinmar we decided to have a check in `DropdownPanelView#focus()` to make sure the first child has the `focus()` method. If not, don't do anything but `logWarning()` telling the integrator that their view instance is missing some API. 

We'll come up with a more systemic solution when the migration to TypeScript is over. A `FocusableView` interface would make things a lot easier.

oleq (changes requested):
I added some docs and I did a minor code refactoring here and there. There are still some changes missing tests, though:

image

---

@ckeditor/qa-team Could you please take a look at this PR? The scope is as follows:

* Whenever a dropdown (any dropdown except maybe table insertion dropdown, there's a dedicated ticket for that) is open, the focus should move somewhere into its panel.
* When a dropdown contains a list (e.g. headings) or a grid (font color), etc., the focus should move to the active item straight away.
* Note: It does not matter whether the dropdown was open by keyboard or by clicking the face of the button, it should work all the same
* Note: To make this happen we made every single button (also: dropdown) in the editor get focused when it gets clicked. This is an a11y standard across the industry but this could also have an impact on all UI related features (also: collaboration features). Please verify that the entire UI of the editor acts predictably and the focus is never lost 🙏 We don't want the editor to get blurred.
* Note: Like I said, ignore the table insertion dropdown for now. It's one of a kind 🙃.

dufipl:
@oleq @mmotyczynska Generally it looks like working fine, I have however two comments:
1. TC enabling still (https://github.com/ckeditor/ckeditor5/pull/12003#issuecomment-1173839094) does not bring the focus back to the editor as it does before that change. From the UI perspective button looks the same as the highlight which brings the focus back to the editor 
image
Zrzut ekranu 2022-07-13 o 13 04 06
2. Bullet list styling does not allow to navigate through potential styles using the keyboard after pressing the dropdown:
Zrzut ekranu 2022-07-13 o 13 05 07
even if the focus is being kept. I'm not sure if that is in the scope of this task, but this dropdown looks similar to others which offer such a possibility.

Besides that, I have not found any significant issues 👍

Reinmar:
> 1.  C enabling still ([Implemented focusing first item after opening a dropdown #12003 (comment)](https://github.com/ckeditor/ckeditor5/pull/12003#issuecomment-1173839094)) does not bring the focus back to the editor as it does before that change. From the UI perspective button looks the same as the highlight which brings the focus back to the editor

Followup ticket for TC. It's ok to extract it from the scope of the current work in this PR.

> 1.  Bullet list styling does not allow to navigate through potential styles using the keyboard after pressing the dropdown:

Separate topic: #11851

Reinmar:
@oleq Am I right that this PR is ready to merge?

Reinmar:
Aren't these two tickets reasolved as well?

*   #5859
*   #1037

Reinmar:
Aren't these two tickets reasolved as well?

*   #5859
*   #1037

oleq:
> Aren't these two tickets reasolved as well?
> 
> * [The selected language in code block dropdown is not visible #5859](https://github.com/ckeditor/ckeditor5/issues/5859)
> * [Improve accessibility of drop–downs #1037](https://github.com/ckeditor/ckeditor5/issues/1037)

As for the second one, I think it's more than this. For instance 

> Exiting the drop–down should also be explained by the text reader.

was not covered by this PR. I think we should refine this issue first to understand what is done from that list and what is left and what is really necessary.

dufipl:
@Reinmar created a follow-up ticket for the TC issue: https://github.com/cksource/collaboration-features/issues/4668
cc @scofalik

PR #12024, opened 2022-07-06
Basic IME support

### LINKED ISSUE #11636: [beforeinput] Handle changes triggered by direct modification of DOM
Part of #11438.

Although `beforeinput` makes mutation-based typing obsolete, some external applications (spell checkers, etc.) may still want to modify the editing root via native DOM API. We may need a basic mutation observer logic to be able to respond to these changes.

However, any other changes applied directly to the DOM, without being applied to the editing view first must be reverted (that's the current behavior on `master`). In other words, this mechanism must ensure that any changes applied to the DOM, that were not also applied to the editing view, must be reverted.

A separate question is whether we should keep the current `mutations` event.

Relevant pieces of code that ensures that changes applied to DOM are then checked by the rendered and reverted if needed:

*   https://github.com/ckeditor/ckeditor5/blob/f1ab675836473ca1e29cc3e3409fb433808e0d13/packages/ckeditor5-engine/src/view/observer/mutationobserver.js#L201-L224
*   https://github.com/ckeditor/ckeditor5/blob/f1ab675836473ca1e29cc3e3409fb433808e0d13/packages/ckeditor5-engine/src/view/observer/mutationobserver.js#L253-L254

### LINKED ISSUE #11736: [beforeinput] Turn the PoC of IME handling into a complete, production-ready implementation
Next step in the #11438 drama.

During #11626 we PoCed an approach to composition handling in which we block rendering during the composition. This approach gave really nice results (was tasted by multiple native CJK users) and we'd like to turn that PoC into a production-ready code.

The unfortunate followup of this work is #12058. Unfortunately, Android requires a completely different approach due to its "unorthodox" integration with onscreen keyboards.

niegowski:

Since this PR is just a part of a bigger epic those changes here are listed here only as a draft. Closes #11636. Closes #11736.

Other (engine): The `Renderer` and the `SelectionObserver` are blocked while `view.Document#isComposing` flag is set to `true`. See #11626. 

Other (typing): In order to handle the results of a composition "session", `InsertTextObserver` will fire the `insertText` event upon `compositionend`. Additionally, the `Input` plugin calls `Model#deleteContent()` on `compositionstart` to clean up the selected fragment (including widgets). See #11626.

Other (widget): The `WidgetTypeAround` fake caret handling is integrated with the `compositionstart` event. See #11626.

Other (engine): The `MutationObserver` is reintroduced but with a different role than the previous one. The new implementation is responsible to detect any changes in the editing DOM tree and marking those to revert by the `Renderer` (including attribute changes). See #11626.

---

### Additional information

This PR is a result of the research on the IME:

[https://github.com/ckeditor/ckeditor5/issues/11626#issuecomment-1121226440](https://github.com/ckeditor/ckeditor5/issues/11626#issuecomment-1121226440)

Changes included in this PR:

*   Added typing logs, hidden behind the `-d typing` CLI flag and the `window.logCKETyping` runtime flag (both must be enabled at the same time to see typing logs). There is a new manual test that toggles that runtime flag: http://fake.ckeditor.com:8125/ckeditor5-typing/tests/manual/beforeinput.html
*   The `Renderer` is blocked while `view.document.isComposing`
    *   This is to avoid any DOM changes while the user is composing since composition is very fragile and could easily break if DOM changes.
*   The `SelectionObserver` is blocked while `view.document.isComposing`
    *   This is needed since the `Renderer` is blocked while composing so the model and the view could get desynchronized with the DOM. 
*   The `InsertTextObserver` is extended (from the before-input branch) to fire `insertText` event on `compositionend` event
*   The `Input` plugin is listening to `compositionstart` before the `Renderer` is blocked to trigger `model.deleteContent()` before the composition start
    *   This is needed for cleaning the non-collapsed selection before the composition starts because since the `Renderer` will be blocked, the browser would remove selected content from DOM and view to DOM mapping would drift apart and could cause major mapping problems that are especially visible while displaying any UI that is positioned in the relation to a DOM Rect mapped from the view to a removed DOM node
*   The `WidgetTypeAround` plugin is also listening to `compositionstart` event to redirect the composition target to a new paragraph in case of a fake caret being active
*   The `MutationObserver` was reintroduced but with a different role than the previous one
    *   The new implementation is responsible to detect any changes in the editing DOM tree and marking those to revert by the Renderer (including attribute changes).
    *   It's not trying to find out what was typed or how the DOM changed.

This PR is not covering any cases with IME on Android. Those will be handled in a follow-up: #12058


Reinmar:
TODO:

* Remove TypingSimulator. It's a dead end - it's too complex.
* Make sure CI is clean.
* Document the PR.
* Turn into a real PR (it's a draft now).

packages/ckeditor5-typing/tests/inserttextobserver.js
```diff
@@ -49,7 +49,7 @@ describe( 'InsertTextObserver', () => {
 		} ).to.not.throw();
 	} );
 
-	it( 'should not work if the observer is disabled', () => {
+	it( 'should not work if the observer is disabled (beforeinput)', () => {
```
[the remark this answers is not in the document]
Reinmar: Ok, clear now. I saw the other test :D

packages/ckeditor5-widget/src/widgettypearound/widgettypearound.js
```diff
@@ -644,6 +647,15 @@ export default class WidgetTypeAround extends Plugin {
 				data.selection = viewDocument.selection;
 			}
 		}, { priority: 'high' } );
+
+		// Note: The priority must precede the default Input plugin compositionstart handler (to call it before delete content).
+		this._listenToIfEnabled( viewDocument, 'compositionstart', () => {
+			if ( model.document.selection.isCollapsed ) {
```
Reinmar, line 653: What does this check mean?

Reinmar: Previously we've done that on `keydown` which is fired just before `compositionstart` on non-Safari browsers. On Safari browsers it's actually fired later, so on master this scenario didn't work at all.

In other words, on Safari we fixed this scenario (it's apparently ok there to modify DOM on `compositionstart`). It didn't work and it works now.

Reinmar: I think that the solution for Chrome will be to bring back some handling on `keydown`. But we'll need to be smart to only touch composition-related ones.

Reinmar: I reported #12088

Reinmar: Actually, I still don't understand that `if()` and it's worth explaining in the code.

Reinmar:
I left a couple of comments here and there, but for me this PR is ready to merge.

@niegowski, do we want to wait for one QA session (to verify that we didn't break anything on the last straight) before mering it or should we do it straight away?

packages/ckeditor5-widget/src/widgettypearound/widgettypearound.js
Reinmar: Bingo, then :D So, to avoid more people being confused about why it's there we need to do something.

packages/ckeditor5-typing/src/inserttextobserver.js
```diff
@@ -63,6 +63,46 @@ export default class InsertTextObserver extends Observer {
...
+			//     this isn't necessarily the selection on which the user started composing.
+			//     We cannot even know whether it's still collapsed (there might be some weird
+			//     editor feature that changed it in unpredictable ways for us). But it's by far
+			//     the simplest solution and should be stable (the selection is definitely correct)
+			//     and probably mostly predictable (features usually don't modify the selection
+			//     unless called explicitly by the user).
+			//   - Try to follow it from the `beforeinput` events. This would be really complex as each
+			//     `beforeinput` would come with just the range it's changing and we'd need to calculate that.
+			// We decided to go with the 2nd option for its simplicity and stability.
+			// TODO maybe we should not pass the DOM event and only translate what we could need in the view/model
```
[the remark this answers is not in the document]
Reinmar: So this TODO is more of a generic ticket needed to be reported? Or do we have one? Either way – it'd be good to get rid of this TODO.

huzedong2015:
I use online build, and then replace `engine`, `typing`, and `widget` packages from the ["ck/epic/11438 migrate to beforeinput"](https://github.com/ckeditor/ckeditor5/tree/ck/epic/11438-migrate-to-beforeinput) branch. The IME problem still exists
version: v35.2.0

Issue #11922, opened 2022-06-14
Content styles contains wrong entries regarding table column resize


pomek:
## 📝 Provide detailed reproduction steps (if any)

**Building content styles using the documentation:**

1. Use `yarn run docs` for building the documentation.
2. Visit [the content styles page](https://ckeditor.com/docs/ckeditor5/34.1.0/installation/advanced/content-styles.html#the-full-list-of-content-styles)
3. Find the `.table-column-resizer` definition.

**Building content styles using a node script:**

1. Create the script file (e.g. `content-styles.js` in the root CKEditor 5 directory):

	  ```js
	  #!/usr/bin/env node
	  
	  /**
	   * @license Copyright (c) 2003-2022, CKSource Holding sp. z o.o. All rights reserved.
	   * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
	   */
	  
	  /* eslint-env node */
	  
	  const buildContentStyles = require( './scripts/docs/build-content-styles' );
	  
	  buildContentStyles();
	  ```

2. Call `node content-styles.js`.
3. Open a file pointed out by the script (_Content styles have been extracted to ..._)
4. Find the `.table-column-resizer` definition.

### ✔️ Expected result

The content styles should not include `.table-column-resizer` definitions.

### ❌ Actual result

Content styles contain the following rules:

```css
/* ckeditor5-table/theme/tablecolumnresize.css */
.ck-content .table .table-column-resizer {
	position: absolute;
	top: -999999px;
	bottom: -999999px;
	right: var(--ck-table-column-resizer-position-offset);
	width: var(--ck-table-column-resizer-width);
	cursor: col-resize;
	user-select: none;
	z-index: var(--ck-z-default);
}
/* ckeditor5-table/theme/tablecolumnresize.css */
.ck-content .table[draggable] .table-column-resizer {
	display: none;
}
/* ckeditor5-table/theme/tablecolumnresize.css */
.ck-content .table .table-column-resizer:hover,
.ck-content .table .table-column-resizer__active {
	background-color: var(--ck-color-table-column-resizer-hover);
	opacity: 0.25;
}
/* ckeditor5-table/theme/tablecolumnresize.css */
.ck-content[dir=rtl] .table .table-column-resizer {
	left: var(--ck-table-column-resizer-position-offset);
	right: unset;
}
/* ckeditor5-table/theme/tablecolumnresize.css */
[...]

PR #12031, opened 2022-07-07
Hide resizers when resizing commands are disabled (e.g. in comments-only mode)

### LINKED ISSUE #11886: Column resize doesn't work with comments-only (or any other feature that disables commands)

Currently the comments-only feature doesn't disable the column resize feature as it disables all the commands.

Since currently column resize doesn't expose any command it doesn't work.

We're planning to add commands so that it also works well with the Track Changes feature.

## 📃 Other details

*   Browser: Any
*   OS: Any
*   First affected CKEditor version: 34.0.0
*   Installed CKEditor plugins: Table column resize

---

Dumluregn:

Fix (table): Table column resizers should be hidden when resizing commands are disabled. Closes #11886.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
+++ b/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
@@ -78,13 +78,25 @@ export default class TableColumnResizeEditing extends Plugin {
 		this._isResizingActive = false;
 
 		/**
-		 * A flag indicating if the column resizing is allowed. It is not allowed if the editor is in read-only mode or the
-		 * `TableColumnResize` plugin is disabled.
+		 * A flag indicating if the column resizing is allowed. It is not allowed if the editor is in read-only
+		 * or comments-only mode or the `TableColumnResize` plugin is disabled.
 		 *
 		 * @private
+		 * @observable
 		 * @member {Boolean}
 		 */
-		this._isResizingAllowed = true;
+		this.set( '_isResizingAllowed', true );
+		this.on( 'change:_isResizingAllowed', ( evt, name, value ) => {
+			if ( value ) {
+				editor.editing.view.change( writer => {
+					writer.removeClass( 'ck-comments-only', editor.editing.view.document.getRoot() );
+				} );
+			} else {
+				editor.editing.view.change( writer => {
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -581,12 +598,18 @@ export default class TableColumnResizeEditing extends Plugin {
 				}
 			} else {
 				// In read-only mode revert all changes in the editing view. The model is not touched so it does not need to be restored.
+				// This case can occur if the read-only mode kicks in during the resizing process.
 				editingView.change( writer => {
 					if ( isColumnWidthsAttributeChanged ) {
-						const columnWidths = columnWidthsAttributeOld.split( ',' );
+						if ( columnWidthsAttributeOld ) {
```
mlewand, line 604: Let's add a short comment what is it needed for.

mlewand (commented):
Couple of minor things.

Please, don't forget to provide the commit message for the fix. This PR has just a placeholder "Type: Message.".

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
Dumluregn: I added a comment - this `if` is necessary to avoid running the `split()` method if there is no previous value (so it would result in an error). I see that I didn't add a test for this case here, but in a CC fixing PR - https://github.com/ckeditor/ckeditor5/pull/12049/files#diff-d08d4a33389c9c57bdda1abc834750454c8f1d7f135bee1dd9fa4ff5a13eb883R888-R907. I'd rather not copy this one test to avoid issues during branch merging - do you think we can do it like that?

Dumluregn: Ok, #12049 was just merged, so we have this test here already :see\_no\_evil:

Dumluregn:
To test how the feature behaves with the `comments-only` mode best to use `editor-modes` manual test from CF repo.

Dumluregn:
I also modified the CSS selector hiding the resizers according to https://github.com/ckeditor/ckeditor5/issues/11922.

mlewand (approved):
LGTM

PR #12033, opened 2022-07-07
Better BR elements handling

### LINKED ISSUE #10217: Extra paragraph and break elements when copy/pasting from Google Docs
## 📝 Provide detailed reproduction steps

1. Go to https://docs.google.com/document/d/1abo1Ca_RHEmuXnAX6k_z0itHihQb3g9rJRM3x4vQ25w/edit
2. Copy some text
3. Go to https://ckeditor.com/docs/ckeditor5/latest/features/basic-styles.html#demo
4. Paste the text

### ✔️ Expected result

image

image

### ❌ Actual result

image

image

## 📃 Other details

* Browser: Tested on Google Chrome Version 91.0.4472.164, Firefox 91.0b7 (64-bit), Safari Version 14.1.1 (16611.2.7.1.4)
* OS: macOS Big Sur Version 11.4
* First affected CKEditor version: 26.0.0
* Installed CKEditor plugins: 

```JSON
    "@ckeditor/ckeditor5-alignment": "^29.0.0",
    "@ckeditor/ckeditor5-autoformat": "^29.0.0",
    "@ckeditor/ckeditor5-basic-styles": "^29.0.0",
    "@ckeditor/ckeditor5-block-quote": "^29.0.0",
    "@ckeditor/ckeditor5-build-classic": "^29.0.0",
    "@ckeditor/ckeditor5-core": "^29.0.0",
    "@ckeditor/ckeditor5-dev-utils": "^25.2.6",
    "@ckeditor/ckeditor5-dev-webpack-plugin": "^25.2.6",
    "@ckeditor/ckeditor5-editor-classic": "^29.0.0",
    "@ckeditor/ckeditor5-essentials": "^29.0.0",
    "@ckeditor/ckeditor5-font": "^29.0.0",
    "@ckeditor/ckeditor5-heading": "^29.0.0",
    "@ckeditor/ckeditor5-highlight": "^29.0.0",
    "@ckeditor/ckeditor5-indent": "^29.0.0",
    "@ckeditor/ckeditor5-link": "^29.0.0",
    "@ckeditor/ckeditor5-list": "^29.0.0",
    "@ckeditor/ckeditor5-media-embed": "^29.0.0",
    "@ckeditor/ckeditor5-paragraph": "^29.0.0",
    "@ckeditor/ckeditor5-paste-from-office": "^29.0.0",
    "@ckeditor/ckeditor5-react": "^3.0.2",
    "@ckeditor/ckeditor5-remove-format": "^29.0.0",
    "@ckeditor/ckeditor5-special-characters": "^29.0.0",
    "@ckeditor/ckeditor5-table": "^29.0.0",
    "@ckeditor/ckeditor5-theme-lark": "^29.0.0",
    "@ckeditor/ckeditor5-typing": "^29.0.0",
    "@ckeditor/ckeditor5-ui": "^29.0.0",
```

Built from source for React with Webpack as per official guideline.

Trying to update from 23.1.0 to the latest. Currently updated up to 25.0.0

---

niegowski:

Fix (enter): A `<br>` element between blocks should be converted to an empty paragraph. Closes #10217.

Test (clipboard): Added test for pasting content with `<br>` element between blocks. See #10217.

Internal (engine): Handling of `<br>` elements in the data pipeline moved from `DomConverter` to the `softBreak` conversion.

---

### Additional information

The original issue is only about pasting such content (trimmed for readability):

```html
<p>A</p>
<br />
<p>B</p>
```

This content expects 3 paragraphs and the middle one is empty but got converted (because of auto-paragraphing, softBreak is not allowed in root context) as:
```html
<paragraph>A</paragraph>
<paragraph><softBreak/></paragraph>
<paragraph>B</paragraph>
```

But since this is very generic HTML and the behavior of `<br>` elements is very specific in web browsers I did some research to find out how different `<br>` scenarios are handled by a web browser and CKE5. 

<details>
<summary>Here is a result (from a manual test: http://fake.ckeditor.com:8125/ckeditor5-enter/tests/manual/softbreak-paragraphing.html):</summary>

</details>

#### The broken cases

<table><tbody><tr><td><strong>Input</strong></td><td><strong>Actual</strong></td><td><strong>Expected</strong></td></tr><tr><td><pre><code class="language-html">&lt;br&gt;
&lt;p&gt;foo&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;&lt;br&gt;&amp;nbsp;&lt;/p&gt;
&lt;p&gt;foo&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;&amp;nbsp;&lt;/p&gt;
&lt;p&gt;foo&lt;/p&gt;</code></pre></td></tr><tr><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;br&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;&lt;br&gt;&amp;nbsp;&lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;&amp;nbsp;&lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td></tr><tr><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
foo&lt;br&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;foo&lt;br&gt;&amp;nbsp;&lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;foo&lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td></tr><tr><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
foo&lt;br&gt; {whitespaces}
&lt;p&gt;after&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;foo&lt;br&gt;&amp;nbsp;&lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;foo&lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td></tr><tr><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;foo&lt;br&gt; {whitespaces} &lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;foo&lt;br&gt;&amp;nbsp;&lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;before&lt;/p&gt;
&lt;p&gt;foo&lt;/p&gt;
&lt;p&gt;after&lt;/p&gt;</code></pre></td></tr><tr><td><pre><code class="language-html">&lt;p&gt;foo&lt;/p&gt;
&lt;br&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;foo&lt;/p&gt;
&lt;p&gt;&lt;br&gt;&amp;nbsp;&lt;/p&gt;</code></pre></td><td><pre><code class="language-html">&lt;p&gt;foo&lt;/p&gt;
&lt;p&gt;&amp;nbsp;&lt;/p&gt;</code></pre></td></tr></tbody></table>

Above cases were failing also while `<br>` was wrapped with some inline elements (like a link or strong).

The only case that was explicitly handled by the editor was `<p><br></p>` and it was fragile to any whitespace around BR element.

#### First solution attempt

I was trying to solve the issue by extending the `<p><br></p>` case that was handled in DomConverter:

[https://github.com/ckeditor/ckeditor5/blob/2a22cd92fba9d2e7a8d03d7838c61e379ef41aa2/packages/ckeditor5-engine/src/view/domconverter.js#L1080-L1083](https://github.com/ckeditor/ckeditor5/blob/2a22cd92fba9d2e7a8d03d7838c61e379ef41aa2/packages/ckeditor5-engine/src/view/domconverter.js#L1080-L1083)

But it had problems:

*   it required using a lot of text trimming logic (that logic is already implemented in `DomConverter` but it would require triggering it multiple times to handle look-ahead checks (for checking block fillers and to trim text itself)
*   it was a "hack" because in the data pipeline only NBSP block fillers are legit, BRs were an additional workaround because they would not land normally in the data pipeline
*   while replacing `<br/>` by a `<p></p>` in some cases, it generated paragraphs nested in other paragraphs or other block elements that could contain text so the generated view structure would be messed up (but upcast conversion to the model would handle it properly)

#### Final solution

I decided to move the main solution to `softBreak` upcast converter because:

*   the proper NBSP block fillers would be already removed
*   all text nodes would be properly trimmed (especially useful for lookahead checks)
*   scenarios of handling BR elements in a specific contexts are much clearer there

You could check how browsers and the editor are handling `<br>`s in some cases in the new [manual test](http://fake.ckeditor.com:8125/ckeditor5-enter/tests/manual/softbreak-paragraphing.html).


oleq (dismissed):
The functionality looks OK.

* There's a problem with GHS.

* I'd add a regular test for paste from office just for the reference and future developers who may come across this issue.

Reinmar:
> Internal (engine): Handling of `<br>` elements in the data pipeline moved from `DomConverter` to the `softBreak` conversion.

I don't know whether we can call this an internal change. The feature was moved from the engine to a plugin. Did you think on how the behavior of the editor without SoftBreak changed? A thing to consider is that it also touches pasting. Editors without this plugin will be quite rare so we can be rather liberal here, but let's at least have this thought through.

Reinmar:
Thanks for writing down these thoughts. It's an interesting idea to move the code to the plugin. It's definitely more elegant if `<br>` handling is done on a conversion layer – fewer things to worry about this at that stage.

OTOH, we're now moving logic that was resolving issues related to how semantic data has to be enhanced to show well in a browser (or strip from these enhancements) to the conversion layer. But on the third hand, that's where we clean up the content and fix where it was misformatted (e.g. autoparagraphs, wrong nesting, disallowed children, etc.).

So, an important thing to consider here is that those BRs are not part of the enhancement that we did. We use `<br>`s only in the BR mode and those are still ["deenhanced" correctly by DomConverter](https://github.com/ckeditor/ckeditor5/pull/12033/files#diff-3faad2836937de50b583589a4fdf4e0bc1cc2e03e019c106d1f3a5d5f7214e6dR1076): which guarantees its symmetricity. Other `<br>`s can be part of the cleanup done on the conversion layer.

So after giving it a deeper thought, I think the change makes sense.

Reinmar:
One more thing to consider: Performance.

We had poor implementation, but probably negligible from the performance pov. Now, it's more complete, but at what cost?

packages/ckeditor5-clipboard/tests/clipboardpipeline.js
```diff
@@ -359,6 +360,24 @@ describe( 'ClipboardPipeline feature', () => {
 			expect( stringifyModel( spy.args[ 0 ][ 0 ] ) ).to.equal( 'x<paragraph>y</paragraph>' );
 		} );
 
+		it( 'converts BR elements between paragraphs to an empty paragraph', () => {
```
Reinmar, line 363:
```suggestion
		// See https://github.com/ckeditor/ckeditor5/issues/10217
		it( 'converts BR elements between paragraphs to an empty paragraph', () => {
```

packages/ckeditor5-typing/tests/input.js
```diff
@@ -514,6 +514,43 @@ describe( 'Input feature', () => {
 			expect( getViewData( view ) ).to.equal( '<p><strong>Foo</strong> {}</p>' );
 		} );
 
+		it( 'should handle bogus br correctly (with custom br to softBreak converter)', () => {
```
Reinmar, line 517: How did you come up with these tests? Did you encounter some regressions after making this change? Does it anyhow affect beforeinput-related work?

Reinmar:
The first test I done was ctrl+a, ctrl+c in this page (PR) and paste it to http://localhost:8125/ckeditor5/tests/manual/all-features.html

Unfortunately, on this branch crashes with:

```
converters.js:849 Uncaught TypeError: Cannot read properties of null (reading 'start')
    at viewToModelListItemChildrenConverter (converters.js:849:1)
    at UpcastDispatcher.viewModelConverter (converters.js:388:1)
    at UpcastDispatcher.fire (emittermixin.js:200:1)
    at UpcastDispatcher._convertItem (upcastdispatcher.js:258:1)
    at UpcastDispatcher._convertChildren (upcastdispatcher.js:291:1)
    at UpcastDispatcher.<anonymous> (upcasthelpers.js:441:1)
    at UpcastDispatcher.fire (emittermixin.js:200:1)
    at UpcastDispatcher._convertItem (upcastdispatcher.js:258:1)
    at UpcastDispatcher._convertChildren (upcastdispatcher.js:291:1)
    at UpcastDispatcher.<anonymous> (upcasthelpers.js:790:1)
```

The best way to test conversion and anything related to that is to take random pieces of HTML from various websites and just paste :D

Reinmar:
I was curious whether the PR indeed fixes the original issue. So, how paste from this document works: https://docs.google.com/document/d/1abo1Ca_RHEmuXnAX6k_z0itHihQb3g9rJRM3x4vQ25w/edit

**NOTE:** This document is really oddly formatted. When you test a normal document, created with normal paragraphs in Google Docs (I tested some other docs I had), the result is better.

### Before

```html
<p>
    <span style="background-color:transparent;color:#000000;font-family:Arial;font-size:11pt;">Sometimes, there’s nothing scarier than a blank page waiting to be filled - especially when it comes to figuring out how to start a resume.&nbsp;</span>
</p>
<p>
    <br>
    &nbsp;
</p>
<p>
    <span style="background-color:transparent;color:#000000;font-family:Arial;font-size:11pt;">For every position that you apply for, you will need to stand out in a pile of applications through your professional experience, achievements, and education, to impress recruiters.&nbsp;</span>
</p>
<p>
    <br>
    &nbsp;
</p>
<p>
    <span style="background-color:transparent;color:#000000;font-family:Arial;font-size:11pt;">Just thinking about all this may make starting your resume seem like a Herculean task.&nbsp;</span>
</p>
<p>
    <br>
    &nbsp;
</p>
```

### On this PR

```html
<p>
    <span style="background-color:transparent;color:#000000;">Sometimes, there’s nothing scarier than a blank page waiting to be filled - especially when it comes to figuring out how to start a resume.&nbsp;</span>
</p>
<p>
    &nbsp;
</p>
<p>
    <span style="background-color:transparent;color:#000000;">For every position that you apply for, you will need to stand out in a pile of applications through your professional experience, achievements, and education, to impress recruiters.&nbsp;</span>
</p>
<p>
    &nbsp;
</p>
<p>
    <span style="background-color:transparent;color:#000000;">Just thinking about all this may make starting your resume seem like a Herculean task.&nbsp;</span>
</p>
<p>
    &nbsp;
</p>
```

Hmm... there are still empty paragraphs that aren't visible in the original document. If paragraphs are styled to have margins (as in our docs) it looks like this:



So, differently than in Google Docs. However, as I mentioned,  the problem is with the styling, not data.

I decided to check CKE4 and TinyMCE's powerpaste out of curiosity too as they worked on their filtering systems for a couple more years. Yet, the results are identical.

### CKEditor 4



### TinyMCE



### To sum up

The PR resolves the reported issue but it's probably even impossible to ensure that the content look very much alike between Google Docs and other systems.

Reinmar:
I did some quick performance tests and for a 12kb of content that's relatively heavy with `<br>`s, the difference was negligible: 190ms -> 210ms.

packages/ckeditor5-typing/tests/input.js
niegowski: The case is that those bogus BRs are stripped while converting to the model (input compares the current model and the model generated from the modified DOM). This case won't affect the beforeinput because it's not used there anymore.

niegowski:
> Unfortunately, on this branch crashes with:
> 
> ```
> converters.js:849 Uncaught TypeError: Cannot read properties of null (reading 'start')
>     at viewToModelListItemChildrenConverter (converters.js:849:1)
>     at UpcastDispatcher.viewModelConverter (converters.js:388:1)
>     at UpcastDispatcher.fire (emittermixin.js:200:1)
>     at UpcastDispatcher._convertItem (upcastdispatcher.js:258:1)
>     at UpcastDispatcher._convertChildren (upcastdispatcher.js:291:1)
>     at UpcastDispatcher.<anonymous> (upcasthelpers.js:441:1)
>     at UpcastDispatcher.fire (emittermixin.js:200:1)
>     at UpcastDispatcher._convertItem (upcastdispatcher.js:258:1)
>     at UpcastDispatcher._convertChildren (upcastdispatcher.js:291:1)
>     at UpcastDispatcher.<anonymous> (upcasthelpers.js:790:1)
> ```

I narrowed the crashing content to:

```html
<ul>
	<li>foo<br><custom-element></custom-element></li>
</ul>
```

niegowski:
> I narrowed the crashing content to:
> 
> ```html
> <ul>
> 	<li>foo<br><custom-element></custom-element></li>
> </ul>
> ```

So the problem is that `<custom-element>` and any inline object (like ``) were ignored while looking for the BR's sibling so the BR converter consumed it but without providing any model element. This crashed on this line:

[https://github.com/ckeditor/ckeditor5/blob/833f3958d64a8ce28dd6392ef55d29b216e561df/packages/ckeditor5-list/src/list/converters.js#L849](https://github.com/ckeditor/ckeditor5/blob/833f3958d64a8ce28dd6392ef55d29b216e561df/packages/ckeditor5-list/src/list/converters.js#L849)

because there is no `result.modelRange` for not converted element.

The case of known inline object elements (like ``) is now solved but the case where there is an unknown inline object is not solved. These cases are fixed:

```html
foo<br><img/>
foo<br><custom-element><img/></custom-element>
<p>foo<br><img/></p>
<p>foo<br><custom-element><img/></custom-element></p>
```

But if the custom element does not contain any content then it crashes:

```html
foo<br><custom-element></custom-element>
```

But there might be more cases where this would fail because of the line mentioned above.

Reinmar:
I'm worried that in ab3e3ad you added just a single test. I'd generate a couple more for weird variations: elements before, multiple elements one by one, other inline elements, etc. I found that the tree-walking logic is often incomplete.

Reinmar:
I did some more smoke tests and also checked whether we didn't start losing more `<br>`s. The results are good:

(left is on the PR, right is master)



If there are differences, they look like this one.

Reinmar:
> So if any feature is using `conversionApi.convertItem()` and is not verifying whether the given view node got converted then it could crash (in the feature code). Of course, this applies to generic cases that do not use it to convert a specific view element (we use it in the table, and image features but we pass a very specific element that we know will be converted - `img` or `table`). The `conversionApi.convertChildren()` is safe because it verifies whether the `modelRange` was returned from the single child conversion and also always returns the range (sometimes collapsed at the `modelCursor` position).

In other words, the frequency will be really low. There has to be a combination of a custom converter, using `convertItem()`, not checking the range, and having `<br>` in the content inside that element.

niegowski:
> In other words, the frequency will be really low. There has to be a combination of a custom converter, using `convertItem()`, not checking the range, and having `<br>` in the content inside that element.

Yes.

Reinmar:
CI was clean apart from a usual 404 in manual tests.

Reinmar:
This change will be reverted in #12183.

Issue #12038, opened 2022-07-11
Release tools: add a hook for modifying a `package.json` file before publishing it on npm


pomek:
### Story

* **Given**: I am a developer who releases CKEditor 5 on npm.
* **When**: I prepare a new release.
* **Then**: The tools should detect a package written in TypeScript.
* **And**: The tools should align a path specified under the `#main` key in a `package.json` file.

### Background

* Under the `#main` key, we keep a relative path to an entry point for a package. 
* For TypeScript packages, it points to a `src/index.ts` file.
* We publish `*.js` files on npm. Hence, we need to modify the path (align an extension).

### Hints

* We don't want to commit this change. It should be visible only for packages published on npm.
* We can add an optional callback executed for each package. It receives a JSON object representing the `package.json` structure.
* We also need to revert this change. Hence, we need two callbacks: an update and a reversion.

pomek:
### Input:

**packages/ckeditor5-alignment/src/alignment.js**

```js
export default class Alignment extends Plugin {
	/**
	 * @inheritDoc
	 */
	static get requires() {
		return [ AlignmentEditing, AlignmentUI ];
	}

	/**
	 * @inheritDoc
	 */
	static get pluginName() {
		return 'Alignment';
	}

	init() {
		const t = editor.locale.t;

		t( 'Whatever. I am a JS file.' );
	}
}
```

**packages/ckeditor5-alignment/src/alignmentui.ts**

```ts
// Somewhere in the class.

dropdownView.buttonView.set( {
	label: t( 'Text alignment An Example' ),
	tooltip: true
} );
```

### Results:

```
Context for the message id is missing ('Whatever. I am a JS file.' from /Users/pomek/Projects/ckeditor/ckeditor5/packages/ckeditor5-alignment/src/alignment.js).
Context for the message id is missing ('Text alignment An Example' from /Users/pomek/Projects/ckeditor/ckeditor5/packages/ckeditor5-alignment/src/alignmentui.ts).
```

pomek:
After adding missing contexts as the tool suggests.

```json
{
	"Whatever. I am a JS file.": "Description for 'Whatever. I am a JS file.'",
	"Text alignment An Example": "Description for 'Text alignment An Example'",
	"Align left": "Toolbar button tooltip for aligning the text to the left.",
	"Align right": "Toolbar button tooltip for aligning the text to the right.",
	"Align center": "Toolbar button tooltip for aligning the text to center.",
[...]

PR #12044, opened 2022-07-13
Added steps required for publishing the package written in TypeScript.

### LINKED ISSUE #11720: [TS MVP] Update NPM release process
We should be able to publish NPM packages from sources written in both JavaScript and TypeScript.

The open question is: how much backward compatibility we want to achieve? If the users import modules directly reaching `src` directories, should we continue to support that or ask them to import from `dist` (or any other name) directory instead?

pomek:

Internal: Prepared the repository to release a package written in TypeScript. Closes #11720.

---

### Additional information

* We still need https://github.com/ckeditor/ckeditor5/issues/12038 to be resolved but it does not block the PR.


.gitignore
```diff
@@ -19,3 +19,7 @@ yalc.lock
 
 build/
 !packages/ckeditor5-build-*/build
+
+# Ignore compiled TypeScript files.
+packages/*/src/**/*.js
```
arkflpc, line 24: I'd keep packages/ckeditor5-utils/src/**/*.js only. This could interfere with work on other packages

pomek, line 24:
```suggestion
packages/ckeditor5-utils/src/**/*.js
```

arkflpc (approved):

PR #12079, opened 2022-07-18
Translation utils - support for skipping packages

pomek:

Internal: Translation utils allow specifying packages that should not be processed.


psmyrek (approved):
LGTM 🎉

PR #12134, opened 2022-07-24
Fixes disappearing toolbar on click in Safari.

### LINKED ISSUE #12115: Toolbars disappearing on Safari

1. Open the editor, and scroll up, so that the toolbar sticks on top.
2. Use any feature from the toolbar on a portion of the content (eg. insert a link, apply formatting)

### ✔️ Expected result

The feature works as expected and the toolbar stays put.

### ❌ Actual result

The toolbar disappears (probably "jumps" to the top, out of viewport).

### ❓ Possible solution

@FilipTokarski suggested this might be due to changes in the focus area.

## 📃 Other details

* Browser: Safari
* OS: MacOS
* First affected CKEditor version: 35.0.0

---

mmotyczynska:

Internal: Fixes disappearing toolbar on click in Safari. Closes #12115.

---

### Additional information

Used internal commit category, since the source bug was not released yet.

### DIFF
--- a/packages/ckeditor5-ui/src/button/buttonview.js
+++ b/packages/ckeditor5-ui/src/button/buttonview.js
@@ -13,6 +13,7 @@ import TooltipView from '../tooltip/tooltipview';
 
 import uid from '@ckeditor/ckeditor5-utils/src/uid';
 import { getEnvKeystrokeText } from '@ckeditor/ckeditor5-utils/src/keyboard';
+import env from '@ckeditor/ckeditor5-utils/src/env';
 
 import '../../theme/components/button/button.css';
 
@@ -127,7 +128,7 @@ export default class ButtonView extends View {
 			this._getTooltipString.bind( this )
 		);
 
-		this.setTemplate( {
+		const template = {
 			tag: 'button',
 
 			attributes: {
@@ -163,7 +164,17 @@ export default class ButtonView extends View {
 					}
 				} )
 			}
-		} );
+		};
+
+		// For safari we have to force the focus as it is implemented differently than in other browsers.
+		if ( env.isSafari ) {
+			template.on.mousedown = bind.to( evt => {
+				this.focus();
+				evt.preventDefault();
+			} );
+		}
+
+		this.setTemplate( template );
 	}
 
 	/**
--- a/packages/ckeditor5-ui/src/button/buttonview.js
+++ b/packages/ckeditor5-ui/src/button/buttonview.js
@@ -166,7 +166,8 @@ export default class ButtonView extends View {
 			}
 		};
 
-		// For safari we have to force the focus as it is implemented differently than in other browsers. See #12115.
+		// On Safari we have to force the focus on a button on click as it's the only browser
+		// that doesn't do that automatically. See #12115.
 		if ( env.isSafari ) {
 			template.on.mousedown = bind.to( evt => {
 				this.focus();
--- a/packages/ckeditor5-ui/src/button/buttonview.js
+++ b/packages/ckeditor5-ui/src/button/buttonview.js
@@ -166,7 +166,7 @@ export default class ButtonView extends View {
 			}
 		};
 
-		// For safari we have to force the focus as it is implemented differently than in other browsers.
+		// For safari we have to force the focus as it is implemented differently than in other browsers. See #12115.
 		if ( env.isSafari ) {
 			template.on.mousedown = bind.to( evt => {
 				this.focus();
--- a/packages/ckeditor5-ui/src/dropdown/dropdownview.js
+++ b/packages/ckeditor5-ui/src/dropdown/dropdownview.js
@@ -9,6 +9,7 @@
 
 import View from '../view';
 import KeystrokeHandler from '@ckeditor/ckeditor5-utils/src/keystrokehandler';
+import env from '@ckeditor/ckeditor5-utils/src/env';
 
 import '../../theme/components/dropdown/dropdown.css';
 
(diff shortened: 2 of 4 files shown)


mmotyczynska:
@ckeditor/qa-team could you please verify if the current fix solves the issue? If possible, please check it in older/slower MacBooks as we have to find rather small value for the timeout which still works even for older machines. Currently it's set to 150ms and it works on MacBook Air (M1, 2020).

Mgsy:
## Steps to reproduce

1.  In Safari, go to http://localhost:8080/ckeditor5/34.2.0/examples/builds-custom/full-featured-editor.html.
2.  Focus the editor and scroll down, so the sticky toolbar appears.
3.  Click on the split button arrow, e.g. in the media embed icon.
4.  Repeat step 3.

## Current result

The toolbar disappears.

Mgsy:
## Steps to reproduce

1.  In Safari, go to http://localhost:8080/ckeditor5/34.2.0/examples/builds/balloon-editor.html.
2.  Select some text.
3.  Click on the split button arrow, e.g. in the media embed icon.

## Current result

The toolbar disappears.

Reinmar:
Please check https://github.com/ckeditor/ckeditor4/blob/master/core/focusmanager.js#L64 (history of that line may give more context).

Reinmar:
BTW, CC is missing in utils 😕 ?

Reinmar:
> BTW, CC is missing in utils 😕 ?

Ah, right – that's due to the conditional statement. Some quick test that stubs the `env.isSafari` should be enough

Reinmar:
I found one blocker: table dropdown doesn't manage the focus correctly. Upon opening the dropdown the focus disappears so the toolbar disappears too.

The solution here will be to do something reasonable with the focus on opening the dropdown – e.g. moving it to the first button in the grid. Right now, we don't have keyboard navigation in grids, so this change can be temporary, but we need to do something now.

Once this issue is fixed, the next priority will be to manage the focus correctly in the dropdown+splitbutton case. E.g. a second click on the arrow in the "highlight" button will indeed lose focus. That's because on the second click on the arrow we do nothing with the focus. The correct solution will be to move it correctly onto the arrow button.

Mgsy:
Note: It seems there is a difference between clicking on an item in the toolbar by "tapping" and hard-clicking the touchpad. Also, clicking and holding the touchpad makes the toolbar disappear almost every time.

Reinmar:
> Note: It seems there is a difference between clicking on an item in the toolbar by "tapping" and hard-clicking the touchpad. Also, clicking and holding the touchpad makes the toolbar disappear almost every time.

One thing that we need to understand first what events are fired in both cases. Apparently, Safari behaves differently in them.

mmotyczynska:
One more issue: when you press and hold any button in toolbar for at least 200ms (or whatever we set the timeout to) then the toolbar disappears.

mmotyczynska:
> could you please verify if the current fix solves the issue

@Mgsy did you manage to check the best value of timeout for simple cases like bold button (or other not-dropdown buttons)? We have several issues here and we have to solve them one by one, so let's start with this simple case.

Mgsy:
> @Mgsy did you manage to check the best value of timeout for simple cases like bold button (or other not-dropdown buttons)? We have several issues here and we have to solve them one by one, so let's start with this simple case.

It seems that the default value introduced by this PR works fine in the most basic cases.

mmotyczynska:
@ckeditor/qa-team I changed the fix completely. There is no timeout anymore so it can be tested on any Mac. Please kindly verify if all the issues mentioned earlier are solved now in Safari by this fix.

godai78:
Recreated the problem in this branch:

LukaszGudel:
I'm testing changes on this branch and it looks really good so far. I didn't found any places where the toolbar would unexpectedly disappear or focus would behave differently. Still, I will test these changes for some time to make sure that everything look good.

---

> In Safari, go to http://fake.ckeditor.com:8888/cke5-workspace/ckeditor5/build/docs/ckeditor5/34.2.0/examples/builds-custom/full-featured-editor.html
> Focus the content.
> Try using any toolbar item (with different results).

@godai78 I'm not able to recreate this issue on this branch. Are you on the latest commit `a59e996`?

godai78:
> Are you on the latest commit `a59e996`?

I was convinced about it, but now I am not. Now I am testing this once again.

LukaszGudel (approved):

Mgsy:
Unfortunately, I can still reproduce the disappearing toolbar issue with the media embed dropdown 😞 

https://user-images.githubusercontent.com/20988892/181271136-15789d0d-9c83-407f-afdf-33cc3a2d1e22.mp4

mlewand:
> Unfortunately, I can still reproduce the disappearing toolbar issue with the media embed dropdown (only in Safari) :disappointed:

This should be fixed by Darek's fix - shouldn't it? (#12162)

packages/ckeditor5-ui/src/button/buttonview.js
```diff
@@ -163,7 +164,17 @@ export default class ButtonView extends View {
 					}
 				} )
 			}
-		} );
+		};
+
+		// For safari we have to force the focus as it is implemented differently than in other browsers. See #12115.
```
Reinmar, line 169:
```suggestion
		// On Safari we have to force the focus on a button on click as it's the only browser
		// that doesn't do that automatically. See #12115.
```

Reinmar:
Code-wise the change looks good to me too.

Just some inspiration for manual tests:

*   iOS
*   different editor types
*   clicking in different panels and buttons and inputs
*   clicking on the toolbar itself (between buttons)

At least the last bit wasn't working (deserves a followup, not a blocker).

Mgsy:
> > Unfortunately, I can still reproduce the disappearing toolbar issue with the media embed dropdown (only in Safari) :disappointed:
> 
> This should be fixed by Darek's fix - shouldn't it? ([#12162](https://github.com/ckeditor/ckeditor5/pull/12162))

Unfortunately, it doesn't fix it. The media embed button is the last one that fails in Safari :/

mmotyczynska:
> The media embed button is the last one that fails in Safari :/

A followup issue for this one #12171.

mmotyczynska:
> At least the toolbar clicking wasn't working (deserves a followup, not a blocker, broken on master too).

A followup issue for this one #12172.

mlewand:
Let's do a final QA testing, that includes:

*   CF features
*   iOS
*   Strong focus on split buttons.

The CKE5 itself has been already well covered by @Mgsy so he don't need to do this particular part that much (contrary to @FilipTokarski 😈).

Let's test `focus-release-hot-fixes` branch that contains both fixes (#12134, #12162).

Mgsy (approved):

FilipTokarski (approved):

PR #12140, opened 2022-07-25
Ck/12077 mixins

### LINKED ISSUE #12077: [TS Followup] Change mixins style
There are two styles of mixins suggested by TypeScript [Handbook](https://www.typescriptlang.org/docs/handbook/mixins.html): suggested and alternative one.

Currently, the usage of mixins in ckeditor5 looks like:

```javascript
export default class MyClass { ... }
mix(MyClass, EmitterMixin);
```

After migration to TypeScript, we could use [the alternative pattern](https://www.typescriptlang.org/docs/handbook/mixins.html#alternative-pattern) and translate the above into:

```typescript
class MyClass { ... }
mix(MyClass, EmitterMixin);
interface MyClass extends Emitter {}
export default MyClass;
```

However, when we want to mix both `DomEmitterMixin` and `ObservableMixin`:

```typescript
class MyClass { ... }
mix(MyClass, DomEmitterMixin);
mix(MyClass, ObservableMixin);

// won't work, because listenTo must be identical in all super-interfaces:
// interface MyClass extends DomEmitter, Observable {}

// But type intersection works differently, so let's use that:
type ObservableDomEmitter = DomEmitter & Observable;
interface MyClass extends ObservableDomEmitter {}
export default MyClass;
```

This is doable, but seems like a hack. I propose to use [the pattern from handbook](https://www.typescriptlang.org/docs/handbook/mixins.html#how-does-a-mixin-work):

```typescript
export default class MyClass extends ObservableMixin( DomEmitterMixin( Object ) ) { ... }
```

This could be made backward-compatible (`XyzMixin` functions would have all the methods also exposed on the function object itself, at least for the existing mixins).

arkflpc:

Other (utils): Introduces new way to apply mixins. Instead of using `mix()` util, classes should extend the base created by `EmitterMixin()`, `ObservableMixin()` and `DomEmitterMixin()` or extend `Emitter` or `Observable` directly. Closes #12077.

---

### Additional information


niegowski (approved):

packages/ckeditor5-utils/src/dom/emittermixin.ts
```diff
@@ -45,116 +44,115 @@ import mix from '../mix';
...
-	 * * To stop listening to all events fired by all object.
-	 *
-	 * @param {module:utils/emittermixin~Emitter|Node|Window} [emitter] The object to stop listening to.
-	 * If omitted, stops it for all objects.
-	 * @param {String} [event] (Requires the `emitter`) The name of the event to stop listening to. If omitted, stops it
-	 * for all events from `emitter`.
-	 * @param {Function} [callback] (Requires the `event`) The function to be removed from the call list for the given
-	 * `event`.
-	 */
-	stopListening(
```
niegowski, line 103: This should probably not get lost and be moved to the interface method same as the above method.

PR #12145, opened 2022-07-25
DomConverter should use a separate DOM document instance in the data pipeline

niegowski:

Other (engine): The `DomConverter` should use a separate DOM document instance in the data pipeline.

---


Reinmar:
Why was the \`document\` parameter there? I can imagine that it might've been useful, but my guess that it wasn't added there to solve an issue with working with multiple documents but for some less important reason (like tests).

niegowski:
> Why was the `document` parameter there? I can imagine that it might've been useful, but my guess that it wasn't added there to solve an issue with working with multiple documents but for some less important reason (like tests).

It was never used with other document instances, there even were no tests verifying if the provided document was used to create DOM nodes. My guess is that it was supposed to be used for the data downcast but since there was no direct distinction between data and editing pipelines in the DomConverter (the `renderingMode` constructor option was added recently) the path of passing the document as a parameter was chosen.

PR #12160, opened 2022-07-27
Improve selecting enabled styles in styles dropdown refresh.

### LINKED ISSUE #12158: Styles dropdown does not apply block style for block element nested in non-limit element

1. Example model:
```
<div> // schema of div: inherit all from $block
   [<paragraph>Well hello there!</paragraph>]
</div>
```
2. Selection is on paragraph, not inside.
3. Execute block style for paragraph

### ✔️ Expected result

Block style is applied on paragraph

### ❌ Actual result

No style is applied, following warning is logged to a console: `style-command-executed-with-incorrect-style-name`

---

CatStrategist:

Fix (style): Fixed selecting enabled block styles when a selection's range is on block element. Closes #12158.

---

### Additional information

https://github.com/ckeditor/ckeditor5/issues/12158#issuecomment-1196559680

### DIFF
--- a/packages/ckeditor5-style/src/stylecommand.js
+++ b/packages/ckeditor5-style/src/stylecommand.js
@@ -91,17 +91,7 @@ export default class StyleCommand extends Command {
 		}
 
 		// Block styles.
-		const selectedElement = selection.getSelectedElement();
-		const selectedBlocks = Array.from( selection.getSelectedBlocks() );
-
-		// If there is a selected element it means that there is just a single tree of elements selected
-		// so we should start from the deepest block which is the last element from `getSelectedBlocks()`.
-		// Otherwise there could be neighbouring blocks selected for a non-collapsed selection - as we want
-		// to display enabled styles for a position where selection started, we should use first element from
-		// `getSelectedBlocks()`.
-		const index = selectedElement ? selectedBlocks.length - 1 : 0;
-
-		const startingBlock = selectedBlocks[ index ];
+		const startingBlock = getStartingBlockElement( selection );
 
 		if ( startingBlock ) {
 			const ancestorBlocks = startingBlock.getAncestors( { includeSelf: true, parentFirst: true } );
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-style/src/stylecommand.js
```diff
@@ -91,10 +91,20 @@ export default class StyleCommand extends Command {
...
+		const selectedBlocks = Array.from( selection.getSelectedBlocks() );
 
-		if ( firstBlock ) {
-			const ancestorBlocks = firstBlock.getAncestors( { includeSelf: true, parentFirst: true } );
+		// If there is a selected element it means that there is just a single tree of elements selected
+		// so we should start from the deepest block which is the last element from `getSelectedBlocks()`.
+		// Otherwise there could be neighbouring blocks selected for a non-collapsed selection - as we want
+		// to display enabled styles for a position where selection started, we should use first element from
+		// `getSelectedBlocks()`.
+		const index = selectedElement ? selectedBlocks.length - 1 : 0;
```
niegowski, line 102: Would it be true for content like this?
```html
[<div><p>foo</p><h2>bar</h2></div>]
```

scofalik:
This PR fixes a thing and it probably could be merged, but...

I dislike the current strategy for deciding whether given style is enabled or not (can be chosen in dropdown). It takes first block from the selection, which doesn't make sense when there's a lot of content in the selection. It's bad representation of what's possible and how the feature could behave. We should either enable all styles that can be set on any of the selected elements (sum), or only those styles that can be selected on all elements (intersection).

I'd go with enabling the sum of allowed styles.

Also, the algorithm for applying the style is bugged in some cases involving nested blocks. For example:

```html
<htmlSection><paragraph>[Foo</paragraph></htmlSection>
<htmlSection><paragraph>Bar]</paragraph></htmlSection>
<htmlSection><paragraph>Baz</paragraph></htmlSection>
```

:point\_up: In this case, paragraph style will be applied to paragraphs Foo and Bar. However in this case:

```html
<htmlSection><paragraph>[Foo</paragraph></htmlSection>
<htmlSection><paragraph>Bar</paragraph></htmlSection>
<htmlSection><paragraph>Baz]</paragraph></htmlSection>
```

It will be applied only to Foo and Baz.

This is because command execution is based on `selection.getSelectedBlocks()` which isn't the best solution for this case as this returns only top-level block elements. Instead, we would like to get all the elements in the selection, not just blocks and not just top-level ones.

I propose to:

1.  Introduce changes described in this comment.
2.  Get back to this PR and see if these changes are still relevant.
3.  At the very least maybe we could reuse some of the tests introduced in this PR.

Reinmar:
I wouldn't like to make such decisions in a PR. AFAIU, we're talking here about two topics related to how all/most block styles will work. This is potentially a too big change to just make it without proper assessment. 

I'd like to:

*   understand the reasoning behind changing this, before/after comparison, impact on other features, etc. 
*   have the change in the shared code (assuming that you want to modify it) done separately from the changes in styles dropdown

Thus, I'd propose to open a new ticket and extract this part there.

scofalik:
Added: #12416 and #12417.

scofalik:
@oleq commented on #12416. Looks like we want to make even bigger refactoring of this feature in the future. So, for now, I won't push with #12416. I still think that we should solve #12417 though and come back here after it is done.

scofalik (changes requested):
#12417 should be solved first.

Reinmar:
Is this PR still necessary? Or can we close it?

scofalik:
My bad, I closed the issue but left the PR.

PR #12183, opened 2022-07-29
Revert "Better BR elements handling"

Reinmar:
Reverts #12033

Context: We realized that we haven't verified the symmetry of the data pipeline. This PR completely rebuilds how BRs are handled and there's a high chance of a regression that we may not be able to spot easily now.


Reinmar:
CI failed but it's an unrelated issue (we actually don't know why it fails, because locally it's green).

PR #12205, opened 2022-08-02
Fix removing tables

### LINKED ISSUE #12201: Deleting a table throws an error

1. Checkout branch `ck/epic/11857-table-resize-improvements`
2. Open manual test with table column resize feature: http://localhost:8125/ckeditor5-table/tests/manual/tablecolumnresize.html.
3. Delete a table.

### ✔️ Expected result

Table was deleted.

### ❌ Actual result

There's an error in the console:

image

### ❓ Possible solution

In one of the PRs refactoring the feature we removed the line checking if `tableNode` is nullish: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-table/src/tablecolumnresize/utils.js#L60. Seems it was necessary for the case of removing the node from model. Restoring it (with comment and tests this time) should be enough.

## 📃 Other details

* Browser: all
* OS: all
* First affected CKEditor version: `ck/epic/11857-table-resize-improvements` branch
* Installed CKEditor plugins: `TableColumnResize`

---

Dumluregn:

Internal (table): Fixed removing tables from the editor. Closes #12201.

---

### Additional information

As described in the issue, previously this was handled by the `if` statement located in https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-table/src/tablecolumnresize/utils.js#L60. Writing the unit tests revealed though that this solution works because of a coincidence.

When a table is removed, the table plugin immediately inserts a paragraph instead. That means that (fortunately) there is always a paragraph as a `position.nodeAfter`, so the removed table is not classified as "affected" by change. However, if that behaviour is altered (i.e. a paragraph won't be inserted) and there will be another table right after a removed one, the code will classify that second table as affected - which is not true.

For that reason there are two unit tests - one performing a "pure" table removal, and one that mimics  the table plugin behaviour, so inserting a paragraph (effectively replacing the table). As the previous solution didn't work for the first test, I proposed a more general fix.

### DIFF
--- a/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
+++ b/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
@@ -27,7 +27,7 @@ import {
 	clamp,
 	fillArray,
 	sumArray,
-	getAffectedTables,
+	getChangedTables,
 	getColumnIndex,
 	getColumnMinWidthAsPercentage,
 	getElementWidthInPixels,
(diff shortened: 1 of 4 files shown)


mlewand (approved):
I added some docs clarifications. Also renamed the helper function as it no longer returns all the affected tables.

PR #12214, opened 2022-08-03
Enable table column resize feature in GHS

### LINKED ISSUE #12191: Table column resize doesn't work well with GHS

1.  Add `TableColumnResize` to http://localhost:8125/ckeditor5-html-support/tests/manual/ghs-all-features.html.
2.  Open the manual test.
3.  Insert a table and resize the columns.
4.  Run `editor.setData( editor.getData() )`.

### ✔️ Expected result

The column remains resized.

### ❌ Actual result

The column is not resized.

### Notes

It works fine after removing GHS.

---

Dumluregn:

Fix (table,html-support): Fixed table column resize integration with General HTML Support. Closes #12191.

---

### Additional information

It's a two-parts fix:

1.  On the GHS side, it adds the `colgroup` and `col` elements schema. It resolves the issue where these elements were filtered out as `unsafe`. Please note that it only happened after using `editor.setData()` - if you just open a demo and insert a table, it can be resized normally.
2.  On the table column resize feature the `ck-table-resized` CSS class handling had to be improved as it was not always present when it should be. It was causing incorrect table rendering during the upcast in some cases (e.g. for a resized table with a lot of text in cell). Also it didn't work well for a table with `<colgroup>` element but without `width` attribute in the editor that was turned into read-only mode during resizing - the class was removed while it should stay intact (there is a unit test for that case - `if tableWidth was set for the first time`).

### Testing

PR should be tested against the epic branch it is targeted into - [ck/**epic**/11857-table-resize-improvements](https://github.com/ckeditor/ckeditor5/compare/ck/epic/11857-table-resize-improvements...ck/12191-table-resize-ghs?expand=1). **Please note that the behaviour on that branch can be different than on the `master` one, but it is still broken.** 

I created a separate branch to facilitate testing: `ck/12191-table-resize-ghs-testing`. Changes:

*   Unit tests - all table column resize tests are focused, so you can just run `yarn test -f table -wc`.
*   Manual tests:
    *   Primary focus should be put on http://localhost:8125/ckeditor5-html-support/tests/manual/ghs-all-features.html test. There is a table column resize feature included. Also there is already a resized table in the content. 
    *   Secondary focus - in http://localhost:8125/ckeditor5-table/tests/manual/tablecolumnresize.html I've adjusted it to test the case with enabling read-only that I described above - it:
        *   has a table with `<colgroup>` element but without `width` attribute,
        *   the editor turns into read-only mode after 2 seconds.

### DIFF
--- a/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
+++ b/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
@@ -599,8 +599,8 @@ export default class TableColumnResizeEditing extends Plugin {
 				// In read-only mode revert all changes in the editing view. The model is not touched so it does not need to be restored.
 				// This case can occur if the read-only mode kicks in during the resizing process.
 				editingView.change( writer => {
-					// If table was already resized before, restore the previous column widths.
-					// Otherwise clean up the view from the temporary resizing markup.
+					// If table had resized columns before, restore the previous column widths.
+					// Otherwise clean up the view from the temporary column resizing markup.
 					if ( columnWidthsAttributeOld ) {
 						const columnWidths = columnWidthsAttributeOld.split( ',' );
 
@@ -612,8 +612,8 @@ export default class TableColumnResizeEditing extends Plugin {
 					}
 
 					if ( isTableWidthAttributeChanged ) {
-						// If table was already resized before, restore the previous table width.
-						// Otherwise clean up the view from the temporary resizing markup.
+						// If the whole table was already resized before, restore the previous table width.
+						// Otherwise clean up the view from the temporary table resizing markup.
 						if ( tableWidthAttributeOld ) {
 							writer.setStyle( 'width', tableWidthAttributeOld, viewFigure );
 						} else {
@@ -621,6 +621,8 @@ export default class TableColumnResizeEditing extends Plugin {
 						}
 					}
 
+					// If a table and its columns weren't resized before,
+					// prune the remaining common resizing markup.
 					if ( !columnWidthsAttributeOld && !tableWidthAttributeOld ) {
 						writer.removeClass(
 							'ck-table-resized',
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -610,12 +609,15 @@ export default class TableColumnResizeEditing extends Plugin {
...
 						} else {
 							writer.removeStyle( 'width', viewFigure );
-							writer.removeClass(
-								'ck-table-resized',
-								[ ...viewFigure.getChildren() ].find( element => element.name === 'table' )
-							);
 						}
 					}
+
+					if ( !columnWidthsAttributeOld && !tableWidthAttributeOld ) {
```
mlewand, line 615: We need explanation / reference for reason of this check.

mlewand (changes requested):

mlewand:
Somehow it didn't make into the review notes, sorry:

---

Let's add an integration test that will case from #12191.

So there should be call to `editor.setData( editor.getData() )` with GHS. I think that it makes sense to add this test in a separate test suite, because configuration will

mlewand (approved):
The solution looks good, however there are some DOM element leaks (see https://app.travis-ci.com/github/ckeditor/ckeditor5/jobs/579209722). Fixed in daac6ad.

It make sense to add column resize to the GHS manual test. Added in d11c8fc.

Dumluregn:
>  however there are some DOM element leaks

Sorry, I was about to take care of it and forgot 👎🏻

mlewand:
> > however there are some DOM element leaks
> 
> Sorry, I was about to take care of it and forgot 👎🏻 Thanks for taking care of it.

img

mlewand:
Issue with

```
🔥 The following errors have been found:
 RESPONSE FAILURE   ➜  1 error
  
  ❌ Failed to load resource from upload.wikimedia.org (HTTP response status code: 404)
    
    …found on the following page:
    ➥  http://localhost:8125/ckeditor5/tests/manual/performance/editorinit.html
error Command failed with exit code 1.
Closing the manual test server...
```

Has been happening in epic (and probably master) branch already. Merging.

Issue #12229, opened 2022-08-05
Inserting image does not move focus out of the toolbar 


mateuszzagorski:
## 📝 Provide detailed reproduction steps (if any)

After using the insert image, the focus stays on the toolbar button:  

https://user-images.githubusercontent.com/11699569/183067079-b926fe8f-c25b-4595-b32f-dface3f32956.mov

### ✔️ Expected result

Focus should be moved onto the image.

### ❌ Actual result

Focus stays on the toolbar button.

## 📃 Other details

This is also happening on a non-collapsed toolbar.
All of the ui components in the `ckeditor5-image` should work correctly.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Dumluregn:
While working on it we should also ensure that https://github.com/ckeditor/ckeditor5/issues/12178#issuecomment-1205363543 is working as expected (it should be fixed with the same code but let's be sure).

mlewand:
> While working on it we should also ensure that [#12178 (comment)](https://github.com/ckeditor/ckeditor5/issues/12178#issuecomment-1205363543) is working as expected (it should be fixed with the same code but let's be sure).

I wonder how is it possible that we never noticed it 😅 IMO it has a major impact on the UX of image feature with balloon toolbars.

PR #12162, opened 2022-07-27
Ck/12125 dropdown focus

Dumluregn:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Internal: Dropdown will move focus to the button after an option is picked from the list. Closes #12125. Closes #12121.

---

### Additional information

See the details in https://github.com/ckeditor/ckeditor5/issues/12125#issuecomment-1197865322.

PR #12239, opened 2022-08-08
Fix the focus management for buttons inside the grouped toolbar

### LINKED ISSUE #12178: Toolbar 'Show more items' button is focused after clicking a button from it

1.  Open full features demo http://localhost:8125/ckeditor5/tests/manual/all-features.html
2.  Reduce the width of the editor so that some buttons are grouped.
3.  Click a button that should move the focus to the editable (e.g. choose language) from the collapsed part of toolbar.

### ✔️ Expected result

Focus is moved to the editable.

### ❌ Actual result

Focus is moved to the "Show more items" toolbar button.

### ❓ Possible solution

It is a regression introduced in #12162. Due to moving the focus when any dropdown is closed, it also affects the toolbar dropdown which is closed after the button's command is executed, so it focuses the toolbar back.

The problem occurs on every type of button - generic one, dropdown and splitbutton.

## 📃 Other details

*   Browser: any
*   OS: any
*   First affected CKEditor version: #12162 - [ck/12125-dropdown-focus](https://github.com/ckeditor/ckeditor5/tree/ck/12125-dropdown-focus) branch
*   Installed CKEditor plugins: toolbar

---

Dumluregn:

Fix (ui): Grouped toolbar no longer steals the focus from the editor after executing a button inside it. Closes #12178.

---

### Additional information

1.  The solution was tested and accepted by QA team (see comments in the issue - https://github.com/ckeditor/ckeditor5/issues/12178#issuecomment-1206434202).
2.  Two tests started failing after this change (for `esc` and `arrowleft` keystrokes). They were assuming that the dropdown button will be focused when those keystrokes are executed, which isn't true in the given scenario - as the dropdown is never focused, it no longer gains focus out of nowhere. I've simply reversed that one assertion, but I'm not sure if it makes any sense now. Perhaps it would be better to remove this checks altogether.
3.  As agreed, the fix was implemented in the `render()` method of the `DropdownView` class as it's a follow-up after #12162. However I'm not sure if that's the right way. The dropdown focus behavior is mostly managed through the helpers in `utils.js` file, like `closeDropdownOnBlur()`. Further splitting this logic between two files will make the whole flow harder to follow. I'd rather add a helper called `focusButtonOnDropdownClose()` which would add a separate listener:

```javascript
function focusButtonOnDropdownClose( dropdownView ) {
	dropdownView.on( 'change:isOpen', () => {
		if ( !dropdownView.isOpen ) {
			if ( dropdownView.element.contains( document.activeElement ) ) {
				dropdownView.focus();
			}

			return;
		}
	}
}
```

and execute it in the `addDefaultBehavior()` function. Please reconsider this approach during the review. Functionally such a change should be transparent and not require the extensive re-testing.


Dumluregn:
Hi @ckeditor/qa-team, since we've fixed #12229 please take a quick peek (no longer than 15 min I think) if the image insert focus issues mentioned in comments under #12178 are solved.

mlewand:
@ckeditor/qa-team when tested please give R+ using GH's review system so that it's clearly visible for us 🙂 I believe that R+ from one of you is enough in this case.

Acrophost:
Issue mentioned in [https://github.com/ckeditor/ckeditor5/issues/12178#issuecomment-1205363543](https://github.com/ckeditor/ckeditor5/issues/12178#issuecomment-1205363543) is still not fully fixed.

*   On balloon editor while adding image via url focus is still going missing (except desktop safari, where focus goes in a wrong place)
*   In case of Safari@iOS (any editor) while inserting image via button the keyboard does not appear after the image is added

packages/ckeditor5-ui/tests/dropdown/dropdownview.js
```diff
@@ -469,4 +470,42 @@ describe( 'DropdownView', () => {
...
+			// Create a button inside the dropdown panel to enable focus.
+			view.panelView.children.add( new ButtonView( locale ) );
+
+			view.isOpen = true;
+
+			expect( view.panelView.element.contains( document.activeElement ) ).to.be.true;
+
+			view.isOpen = false;
+
+			expect( view.element.contains( document.activeElement ) ).to.be.true;
```
oleq, line 485: Let's be precise here. We know that the focus should be on `view#buttonView`.

packages/ckeditor5-ui/tests/dropdown/dropdownview.js
```diff
@@ -469,4 +470,42 @@ describe( 'DropdownView', () => {
...
 	} );
+
+	describe( 'on dropdown close', () => {
+		it( 'should focus a dropdown button if focus is inside the dropdown', () => {
+			// Create a button inside the dropdown panel to enable focus.
+			view.panelView.children.add( new ButtonView( locale ) );
+
+			view.isOpen = true;
+
+			expect( view.panelView.element.contains( document.activeElement ) ).to.be.true;
```
oleq, line 481: Let's be precise here. We know that `document.activeElement` is the `new ButtonView` from L477. There could be some mistakes in the code that would make this test pass by accident (what if `panelView` becomes more complex in the future and there's always something else focusable there?)

packages/ckeditor5-ui/tests/dropdown/dropdownview.js
```diff
@@ -469,4 +470,42 @@ describe( 'DropdownView', () => {
...
+			button.classList.add( 'test' );
+
+			// Create a button inside the dropdown panel.
+			view.panelView.children.add( new ButtonView( locale ) );
+
+			view.isOpen = true;
+
+			expect( view.panelView.element.contains( document.activeElement ) ).to.be.true;
+
+			document.querySelector( '.test' ).focus();
```
oleq, line 501: What if there's some test runner element with the same class? You could've saved the reference to the DOM element in L490 and re-used it here. Also, no reason to give a specific class then.

packages/ckeditor5-ui/tests/dropdown/dropdownview.js
```diff
@@ -3,12 +3,13 @@
...
  */
 
+/* global document */
+
 import DropdownView from '../../src/dropdown/dropdownview';
 import KeystrokeHandler from '@ckeditor/ckeditor5-utils/src/keystrokehandler';
 import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
 import ButtonView from '../../src/button/buttonview';
 import DropdownPanelView from '../../src/dropdown/dropdownpanelview';
-import global from '@ckeditor/ckeditor5-utils/src/dom/global';
```
oleq, line 11: This change has nothing to do with the PR.

oleq (changes requested):
LGTM except for some comments.

Dumluregn:
Requested changes were applied. As for my own refactor proposal:

> As agreed, the fix was implemented in the `render()` method of the `DropdownView` class as it's a follow-up after [**Ck/12125 dropdown focus #12162**](https://github.com/ckeditor/ckeditor5/pull/12162). However I'm not sure if that's the right way. The dropdown focus behavior is mostly managed through the helpers in `utils.js` file, like `closeDropdownOnBlur()`. Further splitting this logic between two files will make the whole flow harder to follow. I'd rather add a helper called `focusButtonOnDropdownClose()` which would add a separate listener:

I won't be able to cover it, perhaps someone from the team will handle it instead.

mmotyczynska:
> Requested changes were applied. As for my own refactor proposal:
> 
> > As agreed, the fix was implemented in the `render()` method of the `DropdownView` class as it's a follow-up after [**Ck/12125 dropdown focus #12162**](https://github.com/ckeditor/ckeditor5/pull/12162). However I'm not sure if that's the right way. The dropdown focus behavior is mostly managed through the helpers in `utils.js` file, like `closeDropdownOnBlur()`. Further splitting this logic between two files will make the whole flow harder to follow. I'd rather add a helper called `focusButtonOnDropdownClose()` which would add a separate listener:
> 
> I won't be able to cover it, perhaps someone from the team will handle it instead.

Actually, it's not that simple, moving the code to `utils.js` causes some tests to fail. This is probably because this code is being executed on `render` and the code in `addDefaultBehavior()` is executing before `render`, so it may need some additional adjustments.  
Besides, Olek is thinking about fixing this problem in a different way, so first let's wait and see how the code changes, and then we will decide if it's still worth moving it to `utils.js`.

oleq:
I noticed that this PR does not address the issue of the "Show more items" toolbar staying visible. I forked this PR and addressed the issue there #12319 + some refactoring.

Dumluregn:
Closing this PR in favor of https://github.com/ckeditor/ckeditor5/pull/12319.

PR #12256, opened 2022-08-10
Keyboard support for insert table dropdown grid

### LINKED ISSUE #3176: Add keyboard support for insert table dropdown
While implementing insert table feature:

selection_053

I've realized that it has no keyboard support defined. But I'd leave it as a followup to focus first on other missing features from MVP.

I also had some minor issues with getting this to work with `FocusCycler` and using `<div tabindex="-1">`. Also maybe I'm missing something but with current implementation of above dropdown I'd like to use methods like `focus( 10 )` and `focus( -10 )` to skip a whole row on `arrowup` and `arrowdown`.

For clarity this issue is only about properly **handling the keyboard** in table dropdown.

mmotyczynska:

Fix (table): Add keyboard support for insert table dropdown. Closes [#3176](https://github.com/ckeditor/ckeditor5/issues/3176).

---

### Additional information


packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -98,18 +110,19 @@ export default class InsertTableView extends View {
 
 				click: bind.to( () => {
 					this.fire( 'execute' );
+				} ),
+
+				keydown: bind.to( evt => {
+					if ( evt.key === 'Enter' ) {
+						this.fire( 'execute' );
```
mmotyczynska, line 117: No sure which piece of logic is responsible for this but clicking a toolbar button followed by enter creates a 2x2 table (should be 1x1, I guess)
image

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -119,14 +132,38 @@ export default class InsertTableView extends View {
...
+		this.keystrokes.listenTo( this.element );
 	}
 
 	/**
 	 * @inheritDoc
 	 */
 	focus() {
-		// The dropdown panel expects DropdownPanelFocusable interface on views passed to dropdown panel. See #30.
-		// The method should be implemented while working on keyboard support for this view. See #22.
+		this.items.get( 0 ).focus();
```
mmotyczynska, line 166: A CSS-related comment but I'm leaving it around here: I noticed the focused grid element has a weird (bold) outline. I think we should get rid of it, the mouse and keyboard selections should look the same IMO.

image

packages/ckeditor5-table/src/ui/inserttableview.js
mmotyczynska: This was happening on the second (and next) opening dropdown. It was caused by resetting selected rows/cols on dropdown opening in table.ui. I removed this code as now on opening dropdown we always focus the first tile so there is no need of resetting it somewhere else.

oleq (approved):

PR #12257, opened 2022-08-10
Keyboard support for special characters dropdown grid

### LINKED ISSUE #9037: Special Characters plugin is not keyboard accessible

1. Load a ClassicEditor with SpecialCharacters plugin
2. Move focus to the toolbar by keyboard
3. Move focus to the Special Characters toolbar button by keyboard
4. Activate button -> focus doesn't move inside the dropdown, so it's unusable by keyboard

### ✔️ Expected result

I can choose and insert a special character from the toolbar by keyboard only

### ❌ Actual result

Focus returns to the editable content and the dropdown is stuck open

## 📃 Other details

* Browser: Chrome 87
* OS: MacOS 10.15.7
* First affected CKEditor version: 22.0.0
* Installed CKEditor plugins: BlockQuote, Bold, Essentials, Font, Heading, Indent, IndentBlock, Italic, ListStyle, **SpecialCharacters**, **SpecialCharactersEssentials**, Strikethrough, Subscript, Superscript, Table, TableCellProperties, TableProperties, TableToolbar, Underline, WordCount,

---

mmotyczynska:

Fix (special-characters): Add keyboard support for Special Characters. Closes [#9037](https://github.com/ckeditor/ckeditor5/issues/9037).

---

### Additional information

https://user-images.githubusercontent.com/95343030/183960326-f5d10b2c-8130-4696-a1b1-0aecdbeaca83.mp4


mmotyczynska:
After setting the row to always contain 10 items (CSS: `grid-template-columns: repeat(10, 1fr)`) I also changed `overflow-x` to `auto` because in small screens (e.g. mobile) the items of the grid at the end of the row were not visible.

mmotyczynska:
There is still one issue to be fixed: updating info section when navigating with keyboard.

For now, the info section is updated on mouseover. We should also update it when navigating with arrow keys. So there are two cases to be discussed (the 1st one depends on the 2nd):

1. If on `mouseover` the info section is being changed, what should happen on `mouseout`? Should we update the info section based on the currently focused tile (for now focus is changed only when navigating with keyboard)?
2. Should the `mouseover` also `focus` the tile? If so, then we wouldn't have to think about 1. because the mouse would take over the current selection.

mmotyczynska:
> setting the row to always contain 10 items (CSS: `grid-template-columns: repeat(10, 1fr)`)

This also causes the special characters grid to be not fully visible in some mobile devices. So for smaller screens I changed the number of items from 10 to 7 in a single row which solves the issue for the 2 android devices a tested. Changing the number to be less than 7 is still possible but would require more adjustments in CSS. This is because at some point the header section does not have enough space for the title and category dropdown.



Unfortunately navigation with keyboard is not dynamic, i.e. it does not check the number of items in a row in runtime but always operates on the hardcoded value for the given feature (for special characters this is equal 10). As a result, on smaller devices when the number of items will be other than 10 (7 in our case), the navigation with arrow up/down will be broken. However it is rather not a common case to use keyboard with small screens so we can improve this in the future.

packages/ckeditor5-special-characters/tests/ui/charactergridview.js
```diff
@@ -75,4 +75,39 @@ describe( 'CharacterGridView', () => {
...
+			const spy = sinon.spy( tile, 'focus' );
+
+			view.tiles.add( tile );
+			view.focus();
+
+			sinon.assert.calledOnce( spy );
+		} );
+	} );
+
+	describe( 'FocusTracker', () => {
```
oleq, line 91: Let's move it to `render()` section where it happens in the code.

oleq (changes requested):
The feature looks OK. There are some tests missing, though and some docs too. 

BTW, I added the support for responsive grids in the keyboard navigation helper because it turned out very easy and works fine with the special chars grid.

PR #12262, opened 2022-08-11
Implemented a Tooltip plugin to handle the JS-driven tooltip system across the UI

### LINKED ISSUE #1074: [FF/Edge] Text in tooltips jumps

🐞 Bug report

## 💻 Version of CKEditor

Latest master

## 📋 Steps to reproduce

1. Hover some icon in the toolbar.

## ✅ Expected result
Text in the tooltip doesn't jump.

## ❎ Actual result
Text in the tooltip jumps.

**GIF**

bug_cke5

**Notes**

I'm able to reproduce it only on Windows.

### LINKED ISSUE #5478: Use BalloonPanelView for tooltips
There are more and more arguments for tooltips refactoring (switching to`BalloonPanelView`). At this moment with hidden overflow of element it's impossible to properly display tooltip.

With this change we should also consider to introduce tooltips in the Comments Plugin ( for edit, delete, save & cancel actions) and use tooltips in the user labels in the Collaboration Plugin.

cc @oleq 

screen shot 2018-07-11 at 11 59 13

### LINKED ISSUE #5636: "Show more items" tooltip placement improvement

Thanks to the toolbar grouping feature the editor looks much better on the smaller devices.

But, there is one thing we could improve - tooltip position. In the situation where an editor has small padding or margin there is a possibility of trimmed tooltip content. We know that by default "show more" button will be always the last one, so we could think about this "improvement".

It all boils down to [introduce a new class of tooltip component](https://github.com/ckeditor/ckeditor5-theme-lark/blob/master/theme/ckeditor5-ui/components/tooltip/tooltip.css) - `_sw`. 

cc @oleq 

### ✔️ Expected result
Screen Shot 2019-10-21 at 10 17 56

*What is the expected result of the above steps?*

### ❌ Actual result

Screen Shot 2019-10-21 at 10 19 21

---

### LINKED ISSUE #5821: The tooltips are slightly disconnected from pointing triangles in certain environments
The tooltips are slightly disconnected from pointing triangles in certain environments

I just noticed that the tooltip body is slightly disjointed from the pointer.

### ✔️ Expected result

The tooltip pointer should be aligned without spacing.

### ❌ Actual result

This is how it looks on Chrome78 @ win10:

Chrome

I went to Firefox to check if it's browser dependent, and seems like it is. Here's the screenshot from Firefox:

Firefox

## 📃 Other details

I believe the display settings are critical here.

I'm using `window.devicePixelRatio = 2.5` (3840x2160 resolution).

**It looks correct in the same browsers in lowDPI environments.**

* Browser: Any
* OS: Win10
* CKEditor version: 15.0.0

---

### LINKED ISSUE #7327: Tooltips / floating windows cause horizontal scrollbar on desktop at small sizes

1. Go to the [CKEditor 5 demo page](https://ckeditor.com/ckeditor-5/demo/)
2. Resize your window to less than ~700px wide (_without_ using the devtools "responsive"/"mobile" viewer; just a regular resize)
3. Observe horizontal scrollbar

### ✔️ Expected result

There should be no horizontal scrollbar when the page is resized to less than 700px on desktop, the same as it is on a "true" mobile environments (incl. i.e. Chrome devtools)

### ❌ Actual result

A regular page resize causes a horizontal scrollbar, whereas a resize using mobile devtools does not.

## 📃 Other details

* Browser: Tested on both Chrome and Firefox
* OS: MacOS
* CKEditor version: 5
* Installed CKEditor plugins: None

From what I can see in my inspector, it's "floating" elements that seem to cause this – toolbars/tooltips/menus that are normally hidden but are showing up in the layout anyway.

**Edit:** This also happens when putting CKeditor in thin containers that butt up against page sides, which is (I think) a perfectly valid use that covers more than just "developers who resize their windows constantly".

---

Screen Shot 2020-05-28 at 10 38 50 AM

### LINKED ISSUE #10435: Toolbar tooltips are covered by the dropdowns

1. Open some dropdown on the toolbar
2. Hover over other toolbar icons

### ✔️ Expected result

Tooltips that show up are visible.

### ❌ Actual result

Tooltips are covered by the dropdowns.

### Screencast

https://user-images.githubusercontent.com/34380544/131097848-bc71ca0d-21a3-4fe5-82bd-de98f7f31681.mp4

---

### LINKED ISSUE #10939: Toolbar and tooltips get cropped when the editor is wrapped in a narrow container

When you place CKEditor in a narrow container with overflow auto or hidden, its toolbar and tooltips get cropped.
I have created a jsfiddle that demonstrates the issue: https://jsfiddle.net/gfox1984/9nz0aqvL/

### ✔️ Expected result

The tooltip and toolbar should either display above the container (so they are not cropped) or they should wrap in the available space.

### ❌ Actual result

Tooltips are cropped:
image
Toolbar is cropped:
image

### ❓ Possible solution

Handle the toolbar and tooltips the same as the image toolbar for instance:
image

Or wrap the toolbar on several lines, as in https://github.com/ckeditor/ckeditor5/pull/8553 (but within the overflow container, rather than within the window)
image

## 📃 Other details

Tested with:
* Browser: Chrome 96
* OS: Windows 10
* First affected CKEditor version: probably all (tested with 31.0.0)
* Installed CKEditor plugins: classic preset

---

### LINKED ISSUE #12067: Toolbar button tooltips should also be visible for keyboard users – let's try again
We first introduced tooltips to focus buttons in https://github.com/ckeditor/ckeditor5/issues/5581. It seemed very simple.

However, in https://github.com/ckeditor/ckeditor5/issues/12066 we figured out that it introduces some annoying UX-glitches.

We decided to revert the original fix for #5581. And try again.

Steps:

1. Let's try to figure out a better, more sophisticated but feasible solution.
2. Let's implement it.
3. If the outcome of step 1. will be "it's damn tricky, we don't have enough time now", let's go with an original solution to #5581 but with some small improvements:
   * turning the alignment menu to horizontal
   * removing tooltips for buttons with visual text labels

oleq:

Feature (ui): Implemented a `TooltipManager` class to manage UI tooltips across components and features. Closes #12067.

MINOR BREAKING CHANGE (ui): The `TooltipView` UI component has been removed, please use the new [tooltip API](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focusmanager.html) instead. Please note that this change **does not** affect integrations that configure tooltips using [`ButtonView#tooltip`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_button_buttonview-ButtonView.html#member-tooltip) property.

MINOR BREAKING CHANGE (ui): The static properties of `BalloonPanelView` have been renamed. `BalloonPanelView.arrowVerticalOffset` is `heightOffset` and `BalloonPanelView.arrowHorizontalOffset` is `sideOffset`.

**Other issues that most likely got fixed** (to be checked individually): 

* [ ] Closes #10939. 
* [x] Closes cksource/collaboration-features#1989. 
* [x] Closes #7327. 
* [x] Closes #1074. 
* [x] Closes #5636. 
* [x] Closes #5821. 
* [x] Closes cksource/collaboration-features#3692. 
* [x] Closes #5478. 
* [x] Closes #10435.

**⚠️ Requires a [PR in CF](https://github.com/cksource/collaboration-features/pull/4711) that aligns the code base to the new plugin and removal of `TooltipView`.**

---

### Status of things.

* ~~This is a WIP without tests (although quite prod-ready)~~
* The prototype works on top of the balloon system (`BalloonPanelView`) and gets controlled by JS.
* The prototype depends on certain `data-` set on DOM elements. If an element has these attributes (e.g. a button), the system will display a tooltip when it gets hovered/focused.
* It supports mouse and keyboard interactions (see screencasts).
* The implementation works so that moving over/leaving an element with the tooltip displayed will get rid of that tooltip no matter if shown by mouse or keyboard in the first place (see screencasts).
* There's a single tooltip at a time in the UI now.
* I introduced some delay before showing up the tooltip for better experience (avoid flashing when moving cursor across the toolbar).
* I removed a show (opacity) transition because with a delay things got messy and the UX was bad (too many things happening).

Also: 
* There's a new manual test exclusively for tooltips.
* `BalloonPanelView` got two new positions (west and east)

### Concerns

* ~~The implementation is a plugin and the plugin needs to be loaded.~~
   * Quite easy to forget, I'd say. Even when I added this plugin to `Essentials`, we have tons of tests that do not load `Essentials`. Same with integrators, they could miss it.
   * OTOH forgetting this plugin is not critical for the users. These are just tooltips.
   * There is no easy work around. This is JS and must be run whenever the editor has UI. There are two options:
      1. Run it as a plugin (this PR).
      2. Have it somewhere around the base UI `View` class (so that when you use any `View`, you get tooltips support out of the box). This isn't going to be pretty, to say the least (the implementation uses `BalloonPanelView` that requires `View` on its own...) ☹️
      * **Decision**: Let's use a singleton loaded by EditorUI
- ~~The implementation is listening to DOM events on the document: What about iframes?~~
    * **Decision**: The editor does not support iframes and shadow roots. The problem concerns all balloons anyway.
- ~~Previous implementation offered a chance change the tooltip text dynamically when it is displayed. The new one will only update text when the tooltip gets hidden and shows up again. This is because the new implementation is based on `data-` attributes of HTML elements.~~
   - I'd don't recall any critical feature relying on this, though. It could've been a side effect of our UI framework.
   - **Decision**: This is not important ATM.
- ~~The implementation is affected by the technical debt of the balloon panel system, notably when the editor is located in a scrollable container~~
   - **Decision**: Hiding tooltips on scroll is a decent workaround. 
 

## TODOs

* [x] Tests.
* [x] Docs.
* [x] (TBD) Make the alignment feature toolbar horizontal.
   * Decision: Let's postpone the decision.
* [x] (TBD) Remove tooltips for buttons with visual text labels.
   * Decision: A dedicated issue (e.g. "remove color").
* [ ] (BC) Rename `BalloonPanelView.arrowVerticalOffset` → `BalloonPanelView.arrowHeightOffset`


Reinmar:
* The tooltip is positioned a bit differently now:
    * After: <br>image
    * Before: <br>image
    * Interestingly, there's a difference in placement of these tooltips between the Classic editor demo and the Chat with mentions example where they look like this. Some rounding issue? <br>image

* I wondered whether these tooltips@focus + the new focus handling will make the tooltips intrusive. My first feelings are: Definitely not. The thing that save us are:
    * The delay. But I'd even consider making it longer.
    * The fact they disappear when I do something so when I really use the editor with mouse they are either still invisible or disappear before I notice them.

Reinmar:
Short answer: This works great. Let's go for it :)

mlewand:
Some issues that I noticed:

## Content changes affecting toolbar position

1. Open manual test with [real time collaboration](http://localhost:8125/ckeditor5/external/collaboration-features/packages/ckeditor5-real-time-collaboration/tests/manual/real-time-editing.html) on two browser.
2. Place focus in the table.
3. Open column dropdown from balloon toolbar.
4. Press esc to move the focus back to the balloon toolbar.
5. **On the second browser connected to the same session:** modify the content so that the table moves below.

Actual: tooltip gets mispositioned.

https://user-images.githubusercontent.com/5353898/184868420-0d183e5e-7e71-48f1-852d-ee3fdc61c458.mp4

## Find and replace

1.  Place focus in the find & replace form
2. Move focus with `tab` key to the "replace with" field.
3. Press tab to move focus to "show options"

Actual: Tooltip is offset to bottom.

The problem comes from the fact that "replace with" adds this tip that changes the height of the dropdown. If you move focus in the reverse order this does not happen.

https://user-images.githubusercontent.com/5353898/184868942-4f54a04f-a411-436d-8adc-fc827756b62a.mp4

## Tooltips not hiding after clicking the button

I believe this is a small nuisance but somehow it sticks to my eye a lot during clicking 😅 Previously tooltip got hidden upon clicking a button. Now it stays there until you move your cursor.

Actual:

https://user-images.githubusercontent.com/5353898/184870519-e8dda3b9-7c8e-4575-a4d5-9c2ba6800365.mp4

Before:

https://user-images.githubusercontent.com/5353898/184870571-f57c55ad-9439-4238-8efd-9be0323a5867.mp4

Having it the way it is proposed now makes me feel like after clicking I'm partially in previous stage of my interaction and partially in the new stage (interacting with the dropdown content).

Reinmar:
@oleq Could you better document the last change (in both places where it was needed): 6ea5bf4.

Also, please open a followup and cc @scofalik there. Maybe we missed something.

oleq:
> @oleq Could you better document the last change (in both places where it was needed): [6ea5bf4](https://github.com/ckeditor/ckeditor5/commit/6ea5bf420d1acb0649ce5c0e27b1716eaeb69961).

#12292

oleq:
@ckeditor/qa-team Could you please take a look at this PR? 

Scope:

*   Well... basically all tooltips, everywhere.
*   Tooltips on mouse hover.
*   Tooltips on keyboard focus.
*   Interaction when the tooltip was displayed using the keyboard then the mouse kicks in (and vice-versa)
*   Bugs [discovered by Marek](https://github.com/ckeditor/ckeditor5/pull/12262#issuecomment-1216518788) that should be addressed.
*   Tooltips in the presence list (a.k.a. avatars) in collaboration features (this PR [requires another PR](https://github.com/cksource/collaboration-features/pull/4711) in CF to work).
    *   There should be no degradation of the UX there.
*   There's a "tooltip" manual test in the ckeditor5-ui package that could help you test tooltips when the container with the editor starts to scroll.

Mgsy:
@dufipl @Acrophost could you please take a look at this PR?

Acrophost:
Issue described here: #10939 still persist on [one of our documentation examples](http://127.0.0.1:8080/ckeditor5/35.0.1/examples/builds-custom/bottom-toolbar-editor.html) after build
Screenshot 2022-08-22 at 13 48 12
Screenshot 2022-08-22 at 13 48 21

PR #12283, opened 2022-08-17
Added aria-expanded for dropdownbuttonview

### LINKED ISSUE #12258: Buttons with dropdown should feature proper semantic markup

Extracted from ckeditor/ckeditor5#1037

The screen reader should properly announce that the button has a popup,

We already cover `aria-haspopup` well enough.

However, we must handle `aria-expanded` for dropdown button views. Interestingly [split buttons already have it implemented](https://github.com/ckeditor/ckeditor5/blob/5e8cdcb9e493ddfbff6f56a12cd8599d3aa664b7/packages/ckeditor5-ui/src/dropdown/button/splitbuttonview.js#L227), so this could serve as a reference.

---

mateuszzagorski:

Fix (ui): Added aria-expanded attribute for the dropdownbuttonview. Closes #12258.

---

### Additional information

Tested briefly with VoiceOver and it seems that the screen reader will now properly announce if the button has a popup.  
You can check this behavior with all-features manual tests and insertImageViaUrl.

### DIFF
--- a/packages/ckeditor5-ui/src/dropdown/button/dropdownbuttonview.js
+++ b/packages/ckeditor5-ui/src/dropdown/button/dropdownbuttonview.js
@@ -49,7 +49,8 @@ export default class DropdownButtonView extends ButtonView {
 
 		this.extendTemplate( {
 			attributes: {
-				'aria-haspopup': true
+				'aria-haspopup': true,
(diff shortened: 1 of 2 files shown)


mlewand (approved):
LGTM, I also confirmed that it works well with NVDA + Chrome

PR #12319, opened 2022-08-23
Addressed issues related to closing dropdowns and "Show more items" menu in the toolbar

### LINKED ISSUE #12005: Toolbar dropdowns should hide when they lose focus

This is a followup task that should only be done after https://github.com/ckeditor/ckeditor5/issues/11838 is implemented.

Currently we don't hide dropdown panels when we focus out from their panels. This leads to weird situations like that:

image

There are numerous ways how you can reproduce this. With the latest release:

1.  Open https://ckeditor.com/docs/ckeditor5/34.2.0/examples/builds/document-editor.html
2.  Put the focus inside of the editor ediable.
3.  Use alt + f10 to focus the toolbar.
4. Press space to open heading dropdown.
5. Press down arrow to make sure that the focus goes inside of a dropdown.
6. Press `tab` to move the focus back to editable.

### ✔️ Expected result

Dropdown is hiden.

### ❌ Actual result

It remains open.

image

## 📃 Other details

There are many other ways to move focus out of dropdown after we open it.  For example in browser you can use builtin search function to move focus to any place on the page.

---

### LINKED ISSUE #12178: Toolbar 'Show more items' button is focused after clicking a button from it

### LINKED ISSUE #12186: Adding media embed does not move focus out of the toolbar

1. Open the examples, classic-editor predefined build in docs
2. Either focus editor's toolbar with keyboard, or select media-embed with mouse
3. Insert media embed.

### ✔️ Expected result

Media embed is correctly selected, and you can continue to type freely.

### ❌ Actual result

https://user-images.githubusercontent.com/11699569/181756554-c817575b-2a19-453d-bc93-d07705275f24.mov

It looks as if the focus is on both the toolbar and on the added media. You can't type, and when pressing enter it will open the media-embed dropdown again.

## 📃 Other details

* Browser: Chrome, Firefox, Safari

---

oleq:

Fix (ui): `DropdownView` should close when it gets blurred. Also, `DropdownView` should focus its `#buttonView` when it gets closed and the focus was inside its `panelView`. Closes #12178. Closes #12186. Closes #12005.

~~❓ **To be verified**: Does it resolve https://github.com/ckeditor/ckeditor5/issues/12005?~~
* Confirmed

---

### Additional information

This is a successor to #12239. It fixes both [#12178](https://github.com/ckeditor/ckeditor5/issues/12178) and [#12186](https://github.com/ckeditor/ckeditor5/issues/12186).

In this PR I did some considerable refactoring. I moved some logic from `DropdownView` to individual behaviors in dropdown utils. The assumption is that the bare `DropdownView` should be just a dumb button+panel that offers a lot of flexibility for developers who use it in different ways (e.g. see the [formatting dropdown example in docs](https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/bottom-toolbar-editor.html)).


Reinmar:
If I understand correctly: if the `addDefaultBehavior()` was used, by default we're still focusing the button. And we still expect this to be overridden in `dropdown#execute` event listener if needed?

While cleaning up the "where to put code that overrides the default behavior" problem in the previous iteration I thought that perhaps this could be approached differently (although, that would be a precedent) – e.g. via some `#execute`'s event property like `data.moveFocusTo`. Do you think it could be a next step to clean up the way how dropdowns are used?

oleq:
> If I understand correctly: if the `addDefaultBehavior()` was used, by default we're still focusing the button. And we still expect this to be overridden in `dropdown#execute` event listener if needed?
> 
> While cleaning up the "where to put code that overrides the default behavior" problem in the previous iteration I thought that perhaps this could be approached differently (although, that would be a precedent) – e.g. via some `#execute`'s event property like `data.moveFocusTo`. Do you think it could be a next step to clean up the way how dropdowns are used?

Let's have a meeting about this maybe?

oleq:
Besides, I just noticed (CI) that there are some adjustments necessary in other packages.

Dumluregn:
I checked the PR functionally and everything works fine in all editor types. I confirm it also fixes #12005 as mentioned in the PR comment. As we discussed, I'm not diving into the code review until the QA team approves it too.

oleq:
Can we have some feedback on this PR please @ckeditor/qa-team?

Scope:

*   In all editor types...
*   when drop-downs are either 
    *   in the visible part of the toolbar,
    *   under "Show more items" button
*   we need to test opening various dropdowns (media, insert table, insert image)...
*   and then closing them by either
    *   executing the command (inserting media, inserting table, etc.)
    *   clicking somewhere outside
    *   using escape key
*   to make sure both the feature (media, table, etc.) dropdown gets closed and the "Show more items" dropdown/toolbar gets closed.

There's also #12005 on top of that.

Thanks :pray:

packages/ckeditor5-ui/tests/dropdown/utils.js
```diff
@@ -184,6 +184,62 @@ describe( 'utils', () => {
...
+					document.body.appendChild( dropdownView.element );
+					document.body.appendChild( externalFocusableElement );
+				} );
+
+				afterEach( () => {
+					dropdownView.element.remove();
+					externalFocusableElement.remove();
+				} );
+
+				it( 'should close the dropdown when the focus was in the #panelView but it went somewhere else', done => {
```
Dumluregn, line 206: This and the next test are failing for me if browser running them is not focused. If they can't be rewritten to make them work in such a scenario, let's drop the comment explaining it.

packages/ckeditor5-ui/tests/dropdown/utils.js
```diff
@@ -3,10 +3,10 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
-/* globals document, Event, console */
+/* globals document, Event, console, setTimeout */
 
 import { assertBinding } from '@ckeditor/ckeditor5-utils/tests/_utils/utils';
-import { keyCodes } from '@ckeditor/ckeditor5-utils/src/keyboard';
+import { global, keyCodes } from '@ckeditor/ckeditor5-utils';
```
Dumluregn, line 9: `document` is already ignored by the linter in line 6 and I don't see another usage of `global` here, so I wonder if it's needed? Also I'd be happy to clarify if there is a preferred approach for us in this matter as we are pretty inconsistent about it in the codebase. 

From the DX point of view I prefer using the `globals` clause as it saves some code, but API docs of `global` package suggest that in some environments (node?) such a solution won't work. Do you have some knowledge to drop?

packages/ckeditor5-ui/src/dropdown/utils.js
```diff
@@ -352,6 +367,39 @@ function focusDropdownContentsOnArrows( dropdownView ) {
...
+// @param {module:ui/dropdown/dropdownview~DropdownView} dropdownView
+function focusDropdownPanelOnOpen( dropdownView ) {
+	dropdownView.on( 'change:isOpen', ( evt, name, isOpen ) => {
+		if ( !isOpen ) {
+			return;
+		}
+
+		// Focus the first item in the dropdown when the dropdown opened.
+		dropdownView.panelView.focus();
+	}, { priority: 'low' } ); // Let the panel show up first.
```
Dumluregn, line 400: This listener works on the same event and priority as in `focusChildOnDropdownOpen()` function. I checked and the order of calls is correct (at first the panel is focused and then the focus is moved to the child), but at the first glance it looks like a race condition or coincidence. I'd rather set the priority to some arbitrary number above the `low` priority like `-100` or at least add a comment that "it will always be executed before `focusChildOnDropdownOpen()` so don't worry child".

Dumluregn (approved):
LGTM, I left some discussion comments but they are not mandatory from my perspective. Still we're waiting for the QA check before merging the PR.

dufipl:
>   * using escape key
> * to make sure both the feature (media, table, etc.) dropdown gets closed and the "Show more items" dropdown/toolbar gets closed.

@oleq, both using the feature and clicking somewhere else close both the feature (media, table, etc.) dropdown and the "Show more items" dropdown/toolbar. However, the `escape` key closes only the feature dropdown/toolbar returning focus back to the "Show more items" dropdown/toolbar. In the context of the above quote, should I consider it incorrect behaviour?

dufipl:
With @Acrophost we have finished testing and we've not found any more issues with this PR beyond the ones above.

packages/ckeditor5-ui/tests/dropdown/utils.js
oleq: The point of the `global` module is having an elegant failsafe proxy to the `global/window` objects regardless of the environment ([see a recent PR](https://github.com/ckeditor/ckeditor5/pull/11999)). 

Technically speaking, there's nothing wrong with using the `/* globals */` declaration but given that every now and then we want to [stub the entire window object](https://github.com/ckeditor/ckeditor5/blob/bacb8cb5547d040ab3a4c175a78c8d6a98176eb3/packages/ckeditor5-ui/tests/panel/sticky/stickypanelview.js#L43) (because you cannot stub some of its properties, DOM is weird) I'd love to see the entire code-base use the `global` module just for consistency and ease of testing. 

If we introduce the editor to some new environment in the future, `global` would be the only point of failure (and the place to fix) instead of dozens of editing/UI features accessing `document` and `window` on their own.

Personally, I think we should forbid using `document` and `window` in `/* globals */` per ESLint configuration but this could be my OCD kicking in.

packages/ckeditor5-ui/src/dropdown/utils.js
oleq: Nice catch 🙂 Still, the order of execution will always be safe because `focusDropdownPanelOnOpen()` is always called first (when the dropdown is created) and `focusChildOnDropdownOpen()` is called next by the integrator and if the priorities are the same, callbacks are executed in the same order they were registered in the `EmitterMixin`.

I suppose `focusDropdownPanelOnOpen, focusChildOnDropdownOpen` this is the right order of things, even if they both get called. 

If it was the other way around (`focusChildOnDropdownOpen, focusDropdownPanelOnOpen`), the focus (in the headings dropdown) would first move to the selected heading, then back to the entire panel, which would focus the first item in the list, which is not necessarily the active heading.

oleq: I'll change the priorities just to keep things easier to understand without the knowledge about internals of `EmitterMixin`.

packages/ckeditor5-ui/tests/dropdown/utils.js
oleq: Addressed in bacb8cb.

packages/ckeditor5-ui/src/dropdown/utils.js
oleq: Addressed in d096b11.

oleq:
> However, the `escape` key closes only the feature dropdown/toolbar returning focus back to the "Show more items" dropdown/toolbar. In the context of the above quote, should I consider it incorrect behaviour?

This is correct, sorry for not being specific. Esc should close UI elements "one by one" until you reach the editing root. Clicking the editing root should close everything at once.

oleq:
> Special characters dropdown closes when clicked between the character buttons. In case of balloon/inline editor, the toolbar closes as well.

Addressed in d3a4bb7.

packages/ckeditor5-media-embed/tests/mediaembedui.js
```diff
@@ -142,7 +149,7 @@ describe( 'MediaEmbedUI', () => {
 				sinon.assert.calledOnce( spy );
 			} );
 
-			it( 'executes the command and closes the UI (if the form is valid)', () => {
+			it( 'executes the command and closes the UI (if the form is valid)', done => {
```
Dumluregn, line 152: Tests for media embed have the same problem when running without focus as the ones fixed in bacb8cb.

Dumluregn (changes requested):

packages/ckeditor5-media-embed/tests/mediaembedui.js
oleq: Addressed in aa85b9f.

Dumluregn (approved):
LGTM, I think the code of `DropdownView` class looks much cleaner now with all the logic moved to the utils.

PR #12320, opened 2022-08-23
Keyboard navigation for table color picker

### LINKED ISSUE #12193: Add keyboard navigation for color picker in table properties

Adding keyboard navigation to font color dropdown does not automatically cover color pickers in table properties - they have to be handled separately. This was checked during the PoC #11851.

*   It has to be possible to enter into color picker with the keyboard.
    *   So it should be possible to focus it with a tab when e.g. table cell properties dialog is open.
*   The colors need to act like a grid.
*   Focus cycler **within** color picker should act as expected.

## 📃 Other details

*   Browser: …
*   OS: …
*   CKEditor version: …

mmotyczynska:

Fix (table): Adds keyboard navigation for table color picker. Closes #12193.

---

### Additional information


packages/ckeditor5-table/src/tableproperties/ui/tablepropertiesview.js
```diff
@@ -360,8 +360,10 @@ export default class TablePropertiesView extends View {
 		[
 			this.borderStyleDropdown,
 			this.borderColorInput,
+			this.borderColorInput.fieldView._dropdownView.buttonView,
```
oleq, line 363: I think this is the right moment do make `_dropdownView` public in the component. `ColorInputView` is private anyway (for now) so this will not hurt anyone.

oleq (changes requested):
## Issues with borders

I noticed two issues related to the border being covered by other elements:

### "Removed color"



We have a twin UI for font color, and everything looks OK. 



Can you figure it out?

---

### Color picker button



I suppose some `z-index` for focused button might help (?).

---

## Issues with aria labels

Also, there are some issues with aria labeling for the color `<input>` element. Take a look at the a11y tree for the width `<input>` element:



Apart from the entire form being unnamed (we have an issue for that), it's pretty clear that there's a "Border" > "Width" relation existing there and screen readers can use it.

Now let's take a look at the a11y tree of the border color `<input>`



There's no label for the `<input>` and this will probably be a problem for screen readers (you can check it out). 

It could be fixed easily in the `ColorInputView#_createInputTextView` method. I suppose it's all about setting `#ariaDescribedById` to the `id` of the whole input.

mmotyczynska:
> I noticed two issues related to the border being covered by other elements:

AFAICS the issues were related to the bottom and left (right for rtl) borders of the `remove color` and `color picker` buttons accordingly because until recently there was no focus in them so the issue was simply not visible.

For the `remove color` button, I just restricted the gray color for the left border so it's not applied for the focused button.

For the `color picker` button, it was actually hidden (the left or right border) therefore it was not visible. This was because the element next to it also has the border so it would be doubled. So my solution was to add a transparent left (right for rtl) border and then the focused element gets the blue border by default. The reason for setting the (left or right) border to transparent instead of hiding it is because when I simply remove it, the content inside slightly changes its position.

mmotyczynska:
> there are some issues with aria labeling for the color `<input>` element

Yes, there were labels missing for all 4 color input fields in table properties. This was because we were adding a label to the input color wrapper (containing a dropdown and the input field) and not to the input field itself.

After discussion with @oleq I also simplified the template for the ColorInputView.

oleq (changes requested):
Everything's fine except that we totally forgot about `TableCellPropertiesView` 😛

mmotyczynska:
> Everything's fine except that we totally forgot about `TableCellPropertiesView` 😛

This is so true :see\_no\_evil: Already fixed it. This change was simple because a lot of logic is shared between both views :+1:

PR #12308, opened 2022-08-22
code-blocks.md: Fix typos

cclauss:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Type: Message. Closes #000.

---

### Additional information

_For example – encountered issues, assumptions you had to make, other affected tickets, etc._

PR #12322, opened 2022-08-24
Docs (markdown files): Fix typos discovered by codespell

cclauss:
Follow-up to #12308

[`codespell -w **/*.md`](https://pypi.org/project/codespell/)


tests/manual/abbreviation-level-2.md
```diff
@@ -311,7 +311,7 @@ Let's write a basic `_createFormView()` function, just to create an instance of
 
 We also need to create a function, which will give us the target position for our balloon from user's selection. We need to convert selected view range into DOM range. We can use {@link module:engine/view/domconverter~DomConverter#viewRangeToDom `viewRangeToDom()` method} to do so.
 
-Finally, let's add our balloon amd fpr, view to the `init()` method.
+Finally, let's add our balloon and fpr, view to the `init()` method.
```
cclauss, line 314: Should `fpr,` be `form`?  On qwerty, that would be one-key-to the-right on both o->p and m->,

mlewand: I'm pretty sure it should be form. We can look at the inline comment in the code snippet itself:

> // Create the balloon and the form view.

So

> Finally, let's add our balloon amd fpr, view to the \`init()\` method.

Should become:

> Finally, let's add our balloon and form view to the \`init()\` method.

godai78:
Great job, thanks!

godai78 (approved):
LGTM

PR #12327, opened 2022-08-24
Keyboard navigation for styles grids

### LINKED ISSUE #12250: Keyboard navigation for styles grids

There should be a proper navigation with keyboard for styles dropdown.
There are two grids - switching between them should be possible with TAB key.
In every grid there should be o proper navigation among items with arrow keys (right/left/up/down).

## 📃 Other details

We should use a new helper for grid navigation introduced in https://github.com/ckeditor/ckeditor5/pull/12202.

The dropdown we should work on:
image

* Browser: …
* OS: …
* CKEditor version: …

mmotyczynska:

Fix (style): Improves keyboard navigation for style feature. Closes #12250.

---

### Additional information

Now `(shift) tab` keystroke navigates between the two sections in style dropdown whilst `arrow keys` navigate between grid items within a single grid.

### DIFF
--- a/packages/ckeditor5-style/tests/ui/stylegridview.js
+++ b/packages/ckeditor5-style/tests/ui/stylegridview.js
@@ -152,6 +152,16 @@ describe( 'StyleGridView', () => {
 		} );
 	} );
 
+	describe( 'focus()', () => {
+		it( 'should focus the first style', () => {
+			const spy = sinon.spy( grid.children.first, 'focus' );
+
+			grid.focus();
+
+			sinon.assert.calledOnce( spy );
+		} );
+	} );
+
 	describe( 'destroy()', () => {
 		it( 'should destroy the FocusTracker instance', () => {
 			const destroySpy = sinon.spy( grid.focusTracker, 'destroy' );
--- a/packages/ckeditor5-style/tests/ui/stylegridview.js
+++ b/packages/ckeditor5-style/tests/ui/stylegridview.js
@@ -118,10 +118,6 @@ describe( 'StyleGridView', () => {
 	} );
 
 	describe( 'render()', () => {
-		beforeEach( () => {
-			grid.render();
-		} );
-
 		it( 'should register styleGridView children elements in #focusTracker', () => {
 			const grid = new StyleGridView( new Locale(), [
 				{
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-style/tests/ui/stylegridview.js
```diff
@@ -108,4 +116,140 @@ describe( 'StyleGridView', () => {
...
 		} );
 	} );
+
+	describe( 'render()', () => {
+		beforeEach( () => {
+			grid.render();
+		} );
+
+		it( 'should register styleGridView children elements in #focusTracker', () => {
+			expect( grid.focusTracker._elements ).to.include( grid.children.first.element );
```
oleq, line 126: We can't just access private properties of other classes like this. They are private for a reason: they may change often and that will break unrelated tests like this one. So either we do this properly by spying on `add()` or we don't do this at all because this is tested indirectly by keyboard navigation tests below.

packages/ckeditor5-style/tests/ui/stylepanelview.js
```diff
@@ -308,4 +271,27 @@ describe( 'StylePanelView', () => {
...
+
+		it( 'should register styleGroupView grids in #_focusables', () => {
+			expect( panel._focusables.map( f => f ) ).to.have.members( [
+				panel.blockStylesGroupView.gridView,
+				panel.inlineStylesGroupView.gridView
+			] );
+		} );
+
+		it( 'should register styleGroupView grid elements in #focusTracker', () => {
+			expect( panel.focusTracker._elements ).to.include( panel.blockStylesGroupView.gridView.element );
```
oleq, line 293: We can't just access private properties of other classes like this. They are private for a reason: they may change often and that will break unrelated tests like this one. So either we do this properly by spying on add() or we don't do this at all because this is tested indirectly by keyboard navigation tests in this test file.

packages/ckeditor5-style/tests/ui/stylegridview.js
```diff
@@ -108,4 +116,140 @@ describe( 'StyleGridView', () => {
 			expect( grid.element.lastChild ).to.equal( grid.children.last.element );
 		} );
 	} );
+
+	describe( 'render()', () => {
+		beforeEach( () => {
```
oleq, line 121: This is used by a single test only, the rest does their own prep. Might be worth it to move this prep to the `it()` that needs it.

oleq (changes requested):
The UX feels good. As for the rest: see the comments.

I added a minor improvement to the screen reader support.

PR #12350, opened 2022-08-25
Fix error when stylesView is missing in ListPropertiesUI.

### LINKED ISSUE #12337: Opening a dropdown panel for list properties throws an error

This error occurs only in **lists without styles**, but configured with extra properties: start index, or reversed order, or both.

1. Open "List start index" demo: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/lists/lists.html#demo-3
2. Open the list dropdown panel to change the start index.

### ✔️ Expected result

No error in the console.

### ❌ Actual result

```
Uncaught TypeError: Cannot read properties of null (reading 'children')
    at eval (listpropertiesui.js:207:129)
    at dropdownView.on.priority (utils.js:289:24)
    at DropdownView.fire (emittermixin.ts:168:47)
    at DropdownView.set [as isOpen] (observablemixin.ts:87:30)
    at SplitButtonView.eval (dropdownview.js:254:16)
    at SplitButtonView.fire (emittermixin.ts:168:47)
    at fireDelegatedEvents (emittermixin.ts:443:17)
    at ButtonView.fire (emittermixin.ts:187:25)
    at TemplateToBinding.eval (buttonview.js:162:12)
    at ProxyEmitter.callback (template.js:973:11)
```

https://user-images.githubusercontent.com/56868128/186606635-6442aad8-84df-4ae9-9d21-b58ac8f012c3.mp4

### 💡 Remarks

- The first version where the error occurs is 35.0.1. This error does not exist in version 34.2.0 and earlier.
- The editor is still fully functional after the error is thrown. The "list start index" and "reversed list" features work fine.
- The same error occurs in the "Reversed list" demo: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/lists/lists.html#demo-4.
- The same error occurs in the manual test for the list properties: http://localhost:8125/ckeditor5-list/tests/manual/list-properties.html (scroll down to the "Without styles, just extra properties" section).

## 📃 Other details

* Browser: Firefox, Chrome
* OS: Win 11
* First affected CKEditor version: 35.0.1

---

mmotyczynska:

Type: Message. Closes #12337.

---

### Additional information


oleq (changes requested):
I'm not so sure about it. The logic you removed was responsible for focusing the active style when the dropdown gets open (and the styles are configured to be displayed).

Removing this one will fall back to the [default focus behavior](https://github.com/ckeditor/ckeditor5/blob/7dfb504a3339fe287627bf422cae8c2a295303bf/packages/ckeditor5-ui/src/dropdown/dropdownview.js#L276) that focuses the first one but not the first active one.

So my gut tells me we need this code. All we need to do is enable it only when the list styles are enabled. In other cases, the default logic that I linked will focus either the start index field or the reversed switch button.

oleq:
For the record, I proposed some [refactoring in another PR](https://github.com/ckeditor/ckeditor5/pull/12319/files#diff-5fde4b4d58cd2f29a4bc8a9292cb60d3d44fd2e57c93ae62aeb3a2b0fefcce27R304) but this should not change my [previous comment](https://github.com/ckeditor/ckeditor5/pull/12350#pullrequestreview-1085566818).

mmotyczynska:
> The logic you removed was responsible for focusing the active style

Yes, you're right. I'll change it.

packages/ckeditor5-list/src/listproperties/listpropertiesui.js
```diff
@@ -185,7 +185,11 @@ function getDropdownViewCreator( { editor, parentCommandName, buttonLabel, butto
 		dropdownView.panelView.children.add( listPropertiesView );
 
 		// Accessibility: focus the first active style when opening the dropdown.
-		focusChildOnDropdownOpen( dropdownView, () => listPropertiesView.stylesView.children.find( child => child.isOn ) );
+		focusChildOnDropdownOpen( dropdownView, () => {
```
oleq, line 188: Do you think we could move this to `createListPropertiesView`? Feels like the right place. Also, I'd wrap it in `if ( listPropertiesView.stylesView ) {` instead of enabling it always and checking this in the callback.

PR #12376, opened 2022-08-31
Rewrite core & ui into TypeScript

### LINKED ISSUE #11726: [TS MVP] Rewrite ckeditor5-ui

### LINKED ISSUE #11727: [TS MVP] Rewrite ckeditor5-core
Rewrite `ckeditor5-core` to TypeScript.

Make sure that automated and manual tests run after the rewrite. Tests themselves are out of scope.

This would be the first serious application of `EmitterMixing` and typing of `Emitter` done in #11715.

### LINKED ISSUE #12218: [TS] Merge the "number" type into PriorityString
Followup of: https://github.com/ckeditor/ckeditor5/issues/11708
Part of: https://github.com/ckeditor/ckeditor5/issues/12027

Right now we define a type of e.g. `on()`'s `options.priority` param as `PriorityString | number`.

For me, the number is part of `PriorityString`. Everywhere where we expect the priority, we also allow a number version of it.

So, the proposal is to:

* Move `number` to `PriorityString` to simplify its use.
* Keep the current name of this union to avoid increasing visibility of the fact that a number can be used (which should be done only in special cases).

arkflpc:

Other (ui): Rewrite `ckeditor5-ui` package to TypeScript. Closes #11726.
Other (core): Rewrite `ckeditor5-core` package to TypeScript. Closes #11727.
Other (utils): Add types to `Config` and merge `number` into `PriorityString`. Closes #12218.

---

### Additional information


packages/ckeditor5-core/src/context.ts
```diff
@@ -41,23 +44,34 @@ import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
  * See {@link module:core/context~Context.create `Context.create()`} for usage examples.
  */
 export default class Context {
+	public readonly config: Config<ContextConfig>;
+	public readonly plugins: PluginCollection<Context | Editor>;
+	public readonly locale: Locale;
+	public readonly t: Locale[ 't' ];
```
niegowski, line 50: Maybe this one deserves a dedicated type?

packages/ckeditor5-ui/src/model.ts
```diff
@@ -7,23 +7,26 @@
...
 import { extend } from 'lodash-es';
 
 /**
  * The base MVC model class.
  *
  * @mixes module:utils/observablemixin~ObservableMixin
  */
-export default class Model {
+export default class Model extends Observable {
+	[ x: string ]: any;
```
niegowski, line 19: Maybe we could make it more typed? 😉

packages/ckeditor5-ui/src/tooltipmanager.ts
```diff
@@ -362,7 +399,7 @@ export default class TooltipManager {
 	 * @param {String} position Name of the position (`s`, `se`, `sw`, `n`, `e`, or `w`).
 	 * @returns {Array.<module:utils/dom/position~PositioningFunction>} Positioning functions to be used by the {@link #balloonPanelView}.
 	 */
-	static getPositioningFunctions( position ) {
+	public static getPositioningFunctions( position: TooltipPosition ): PositioningFunction[] {
```
niegowski, line 402: Let's move it just after other public methods.

packages/ckeditor5-ui/src/template.ts
```diff
@@ -829,38 +849,55 @@ export default class Template {
...
+type AttributeValues = ( TemplateSimpleValue | TemplateBinding )[] |
+	[ NamespacedValue<( TemplateSimpleValue | TemplateBinding )[]> ];
 
 /**
  * Describes a binding created by the {@link module:ui/template~Template.bind} interface.
  *
  * @protected
+ * @internal
  */
 export class TemplateBinding {
```
niegowski, line 880: Maybe this could be abstract?

packages/ckeditor5-ui/src/template.ts
```diff
@@ -918,7 +955,11 @@ export class TemplateBinding {
...
 	 */
-	activateAttributeListener( schema, updater, data ) {
+	public activateAttributeListener(
+		schema: ( TemplateSimpleValue | TemplateBinding )[],
+		updater: Updater,
+		data: RenderData
+	): () => void {
 		const callback = () => syncValueSchemaValue( schema, updater, data );
 
 		this.emitter.listenTo( this.observable, 'change:' + this.attribute, callback );
```
niegowski, line 965: Event type?

packages/ckeditor5-core/tsconfig.json
```diff
@@ -1,15 +1,3 @@
 {
-  "compilerOptions": {
```
niegowski: A relative path outside the package would require a strict directory structure while developing plugins. Maybe we could use the package name prefixed path same as we do for imports between packages?

niegowski:
```json
{
  "extends": "ckeditor5/tsconfig.json"
}
```

packages/ckeditor5-ui/src/toolbar/toolbarview.ts
```diff
@@ -410,7 +427,7 @@ export default class ToolbarView extends View implements DropdownPanelFocusable
 					 * @error toolbarview-item-unavailable
 					 * @param {String} name The name of the component.
 					 */
-					logWarning( 'toolbarview-item-unavailable', { name } );
+					logWarning( 'toolbarview-item-unavailable', { name: item } );
```
niegowski, line 430: Is this still a name only? Maybe we should pass the item there and update the param above?

arkflpc: It's not only a name anymore. Do you want to make it `item`?

niegowski: > It's not only a name anymore. Do you want to make it `item`?

👍  and please remember to update the param in jsdoc above

PR #12401, opened 2022-09-02
Aligned table can now be resized with the proper border

### LINKED ISSUE #11785: When table is right-aligned, cannot resize with left border

1. On latest `ck/epic/3284-table-column-resize` open `tablecolumnresize` manual test
2. Add a table 
3. Align it to the right

### ✔️ Expected result

User can resize the table using left table border.

### ❌ Actual result

User can only resize the table using right table border. Left border doesn't even have the resize handler and nothing happens when you try resizing it. You need to use the right border, which is a counter intuitive UX.

https://user-images.githubusercontent.com/34380544/169052781-51ef27c6-0f7c-4ac9-a79c-187465f3673d.mov

## 📃 Other details

* Browser: any

---

mateuszzagorski:

Fix (table): Updated resize handler position for side aligned tables. Closes #11785.

---

### Additional information

It is now possible to resize the table using the:

*   left border handler on a right aligned table.
*   right border handler on a left aligned table when using the `rtl editor`.


packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -643,28 +651,43 @@ export default class TableColumnResizeEditing extends Plugin {
...
 		const viewFigureWidth = getElementWidthInPixels( editor.editing.view.domConverter.mapViewToDom( viewFigure ) );
 		const tableWidth = getTableWidthInPixels( modelTable, editor );
 		const leftColumnWidth = columnWidths[ leftColumnIndex ];
-		const rightColumnWidth = isRightEdge ? undefined : columnWidths[ leftColumnIndex + 1 ];
+		let rightColumnWidth = isDraggableEdge ? undefined : columnWidths[ leftColumnIndex + 1 ];
+
+		// The data needs to be updated on right/left table alignment. See #11785.
+		if ( ( modelTable.getAttribute( 'tableAlignment' ) === 'right' && isLtrContent ) ||
+			( modelTable.getAttribute( 'tableAlignment' ) === 'left' && !isLtrContent ) ) {
+			viewRightColumn = isDraggableEdge ? undefined : viewColgroup.getChild( leftColumnIndex - 1 );
```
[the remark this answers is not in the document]
mlewand: As a side note the initial assignment:

```
let viewRightColumn = isTableResizeEdge ? undefined : viewColgroup.getChild( leftColumnIndex + 1 );
```

and reassignment in the `if`:

```
viewRightColumn = isTableResizeEdge ? undefined : viewColgroup.getChild( leftColumnIndex - 1 );
```

Differs only by using `+1` and `-1`.

So I suppose a better idea would be to just assign the offset to a constant and use it for both assignments.

```
cosnt nextColumnOffset = <condition> ? 1 : -1;
```

packages/ckeditor5-table/tests/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -1291,46 +1292,137 @@ describe( 'TableColumnResizeEditing', () => {
 					}
 				} );
 
-				it( 'shrinks the table twice as much when resizing centered table as compared to aligned table', () => {
-					const columnToResizeIndex = 1;
-					const mouseMovementVector = { x: -10, y: 0 };
+				describe( 'where there are 2 tables: centered and aligned', () => {
```
mlewand, line 1295:
```suggestion
				describe( 'with 2 tables: centered and aligned', () => {
```

mlewand: I know that it's from the code that was before but this wording is a little unfortunate.

mlewand (changes requested):
It works fine, the overall algorithm is good. However there are quite a few things to simplify and clarify before we can merge it.

mlewand:
I also merged the latest `master` branch to this PR.

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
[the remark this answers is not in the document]
mlewand: @mateuszzagorski I have no answer to the question

> Why do we need to pick different right column in case of a table that is side aligned away?

I don't see rationale why are we setting `viewRightColumn` to `col` element describing column marked on the screen below:



I don't understand that.

Just thinking of it the thing that we should change here is [the `isTableResizeEdge` constant](https://github.com/ckeditor/ckeditor5/blob/aa0d99bd17c066d73518927157d7c4c7cfeda4be/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js#L668) - and I believe it should work already.

I have a feeling that we'll simply have to assume that "left" column might not always be availably, similar like https://github.com/ckeditor/ckeditor5/blob/aa0d99bd17c066d73518927157d7c4c7cfeda4be/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js#L680

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.js
```diff
@@ -730,3 +748,16 @@ export default class TableColumnResizeEditing extends Plugin {
 		}, { priority: 'lowest' } );
 	}
 }
+
+// Helper function for checking if the table is:
+// 	- aligned right with left-to-right content language direction
+// 	- aligned left with right-to-left content language direction
+//
+// The function got extracted to avoid code duplication as we are using it multiple times.
```
mlewand, line 756: Good docs except this very line.

This should be obvious for everyone, no need to mention detail like that.

Also add API docs for params, returned value.

As an example see:

https://github.com/ckeditor/ckeditor5/blob/aa0d99bd17c066d73518927157d7c4c7cfeda4be/packages/ckeditor5-list/src/listproperties/listpropertiesediting.js#L817-L818

mlewand (changes requested):
We need to be get this `_getResizingData` function to be refactored correctly.

It's already not nearly as nice as we'd like it to be - so let's sure it doesn't degraded any further with our change.

This is the only remaining problem.

PR #12409, opened 2022-09-05
Show to the end user that the alternative text is applied or not.

### LINKED ISSUE #12268: Image alternative text presence should be conveyed to the user

Currently the end user has no idea whether an alternative text is applied to a certain image or not.

1.  Open [full featured example for the latest release](https://ckeditor.com/docs/ckeditor5/35.0.1/examples/builds-custom/full-featured-editor.html).
2.  Focus the first image.
3.  Change alternative text to non-empty value.

### ✔️ Expected result

The fact that the image has an alternative text should be conveyed to the user with UI.

**In addition** to that let's change the button name to "Image text alternative".

### ❌ Actual result

No information is present.

### ❓ Possible solution

IMO just marking the UI button as `isOn` would be sufficient and consistent with how ImageCaption, TableCaption and the alikes are doing this:

For describing link state we even go extra mile. Not only we're setting `isOn` to `true` on link button, but there's this little "status" icon in the top-right corner of the image:

## 📃 Other details

*   Browser: Any
*   OS: Any

---

pszczesniak:

Fix (image): Show to the end user that the alternative text is applied or not. Closes #12268.

---

### Additional information


packages/ckeditor5-image/tests/imagetextalternative/imagetextalternativeui.js
```diff
@@ -110,6 +110,16 @@ describe( 'ImageTextAlternativeUI', () => {
 
 			sinon.assert.callOrder( disableCssTransitionsSpy, addSpy, selectSpy, enableCssTransitionsSpy );
 		} );
+
+		it( 'has isOn bound to command\'s value', () => {
+			const button = editor.ui.componentFactory.create( 'imageTextAlternative' );
```
mmotyczynska, line 115: It is not necessary to create here this button as we already have a `button` instance that we can re-use, created in the beforeEach block.

mmotyczynska (commented):
The fix looks fine, great work :+1: 

The only thing I would change is this unnecessary button instance creation, please check the comment.

packages/ckeditor5-image/tests/imagetextalternative/imagetextalternativeui.js
pszczesniak: Fixed, ty for help.

mmotyczynska (approved):

mlewand (approved):
LGTM

PR #12415, opened 2022-09-05
Fixes possibility to insert image url with keyboard

### LINKED ISSUE #12215: Not possible to insert image URL with keyboard

1.  Open editor with Insert Image feature, e.g. https://ckeditor.com/docs/ckeditor5/30.0.0/features/images/image-upload/images-inserting.html
2.  Navigate to toolbar with keyboard (alt/opt + F10)
3.  Navigate to Insert Image button with arrow keys
4.  Press arrow down to open the dropdown

### ✔️ Expected result

The dropdown opens and is fully visible.

https://user-images.githubusercontent.com/95343030/182612417-10c88621-7391-40b6-a7ce-8b655a2e70cb.mov

### ❌ Actual result

The dropdown opens but is empty.

https://user-images.githubusercontent.com/95343030/182612049-7a7a46a8-5b8c-43a0-a7fc-4ce58964ff6f.mov

### ❓ Possible solution

According to @oleq it's _likely_ that rendering is done in `execute` event.

However, pressing down arrow doesn't cause `execute` event (unlike space and enter key).

If that's the case we should simply fire `execute` event as well (for down arrow too) to make sure that other/future implementations doesn't fall into the same issue.

## 📃 Other details

Be sure to check whether nothing in CF broke after this.

*   Browser: …
*   OS: …

---

mmotyczynska:

Fix (ui): Fixes possibility to insert image url with keyboard. Closes #12215.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
+++ b/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
@@ -122,7 +122,7 @@ export default class ImageInsertUI extends Plugin {
 
 		// Defer the children injection to improve initial performance.
 		// See https://github.com/ckeditor/ckeditor5/pull/8019#discussion_r484069652.
-		dropdownView.buttonView.once( 'open', () => {
+		dropdownView.once( 'change:isOpen', () => {
 			panelView.children.add( imageInsertView );
 		} );
 
--- a/packages/ckeditor5-image/tests/imageinsert/imageinsertui.js
+++ b/packages/ckeditor5-image/tests/imageinsert/imageinsertui.js
@@ -117,6 +117,13 @@ describe( 'ImageInsertUI', () => {
 
 			expect( dropdown.panelView.children.length ).to.equal( 1 );
 			expect( dropdown.panelView.children.first ).to.be.instanceOf( ImageInsertPanelView );
+
+			dropdown.isOpen = false;
+			dropdown.isOpen = true;
+
+			// Make sure it happens only once.
+			expect( dropdown.panelView.children.length ).to.equal( 1 );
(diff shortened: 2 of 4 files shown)


oleq (commented):
What would you say if we did a simpler fix

```diff
diff --git a/packages/ckeditor5-image/src/imageinsert/imageinsertui.js b/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
index 21c58f1b9a..a3729f0fba 100644
--- a/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
+++ b/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
@@ -122,7 +122,7 @@ export default class ImageInsertUI extends Plugin {
 
                   // Defer the children injection to improve initial performance.
                   // See https://github.com/ckeditor/ckeditor5/pull/8019#discussion_r484069652.
-                  dropdownView.buttonView.once( 'open', () => {
+                  dropdownView.once( 'change:isOpen', () => {
                            panelView.children.add( imageInsertView );
                   } );
```

The difference is that `change:isOpen` is fired always whenever the dropdown opens. It does not matter if this is a result of pressing a button, pressing a keystroke, or doing whatever really. 

For instance, that would keep the feature safe (no empty dropdown) if anyone decided to open it by `dropdownView.isOpen = true`. This `keydown` thing will not secure this case.

mmotyczynska:
> What would you say if we did a simpler fix?

This would be a really simple fix so I already implemented it as you suggested.

What I wanted to achieve earlier was to ensure that future features do not fall into the same issue if they have lazy loaded dropdown content. That's why I proposed a change in the SplitButtonView instead of simply fixing the ImageInsert feature itself. But you pointed out that case with `dropdownView.isOpen = true` so I decided to go with your simpler fix, thanks :+1:

PR #12419, opened 2022-09-06
Fix list ordering when block quote was inserted

### LINKED ISSUE #11082: Ordering problem when block quote is being inserted into the middle of the list

1. Prepare six lines of text
2. Create numbered list (one number for each text line)
3. Select lines 3 and 4 and add a "Block quote"
4. Change the order of the list in the block quote.

### ✔️ Expected result

Order changes only in the block quoted list.

### ❌ Actual result

Order changes are also in the list below.
block quote order issue
Changing the order of the list above or below the block quote impacts only desired list. 

## 📃 Other details

Browser: Chrome, Firefox, Safari
OS: Mac Os Monterey
First affected CKEditor version: 31.1.0

---

DawidKossowski:

Fix (list): Ordering problem when block quote was being inserted into the middle of the list. Closes #11082.

---

### Additional information

This issue was caused by by incorrectly ignoring closing tags when searching for list items.

### DIFF
--- a/packages/ckeditor5-list/src/list/utils.js
+++ b/packages/ckeditor5-list/src/list/utils.js
@@ -289,7 +289,7 @@ export function findNestedList( viewElement ) {
  *
  * It means that values of `listIndent`, `listType`, `listStyle`, `listReversed` and `listStart` for all items are equal.
  *
- * Additionally, if the `position` is inside a list item that list item will be returned as well.
+ * Additionally, if the `position` is inside a list item, that list item will be returned as well.
  *
  * @param {module:engine/model/position~Position} position Starting position.
  * @param {'forward'|'backward'} direction Walking direction.
--- a/packages/ckeditor5-list/src/list/utils.js
+++ b/packages/ckeditor5-list/src/list/utils.js
@@ -297,7 +297,7 @@ export function getSiblingNodes( position, direction ) {
 	const items = [];
 	const listItem = position.parent;
 	const walkerOptions = {
-		ignoreElementEnd: true,
+		ignoreElementEnd: false,
 		startPosition: position,
 		shallow: true,
 		direction
@@ -395,7 +395,6 @@ export function getSelectedListItems( model ) {
 		.filter( element => element.is( 'element', 'listItem' ) )
 		.map( element => {
 			const position = model.change( writer => writer.createPositionAt( element, 0 ) );
-
 			return [
 				...getSiblingNodes( position, 'backward' ),
 				...getSiblingNodes( position, 'forward' )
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-list/tests/list/utils.js
```diff
@@ -307,8 +308,8 @@ describe( 'utils', () => {
 				'<listItem listType="bulleted" listIndent="0">3.</listItem>' +
 				'<listItem listType="bulleted" listIndent="0">4.</listItem>'
 			);
-
```
scofalik, line 310: Looks like you accidentally removed a line break here.

packages/ckeditor5-list/tests/list/utils.js
```diff
@@ -444,10 +449,31 @@ describe( 'utils', () => {
 			);
 
 			expect( getSiblingNodes( document.selection.getFirstPosition(), 'forward' ) ).to.deep.equal( [
+				document.getRoot().getChild( 2 ),
 				document.getRoot().getChild( 3 ),
 				document.getRoot().getChild( 4 )
 			] );
 		} );
+
+		it( 'should return only list items from block quoted list when selection is inside (direction="forward")', () => {
```
scofalik, line 458: Is there a similar test but for direction = "backward"? I know we didn't fix that case but it would be good to have symmetry in test, in case if someone changes something in the future.

packages/ckeditor5-list/src/list/utils.js
```diff
@@ -395,7 +395,6 @@ export function getSelectedListItems( model ) {
 		.filter( element => element.is( 'element', 'listItem' ) )
 		.map( element => {
 			const position = model.change( writer => writer.createPositionAt( element, 0 ) );
-
```
scofalik, line 398: Accidentally removed line break. Maybe something is wrong with your IDE settings?

packages/ckeditor5-list/tests/list/utils.js
scofalik, line 458: I'd like you change the name for this test. This test is not about block quote only, it is about returning items that are contained in a different element. My proposal: "should return only list items that are inside the same parent element". Or "...the same container".

scofalik (changes requested):
Good job, although some changes are needed.

Apart from what I commented, please change `getSiblingsNodes` description so that it mentions that if the `position` is inside a list item, that list item is returned as well.

DawidKossowski:
@scofalik I added all the changes you raised.

scofalik (approved):
LGTM. Waiting for CLA.

PR #12424, opened 2022-09-07
Corrected autolinking urls without protocol

### LINKED ISSUE #11040: Auto-link correction

1. …type www.google.com in content (without http)
2. …inspect the link, you'll find it with href='www.google.com' which is invalid

### ✔️ Expected result

href='https://www.google.com'

### ❌ Actual result

href='www.google.com'

### ❓ Possible solution

Either auto-link would add https:// to the link, or doesn't link www.google.com

kbinieda:

Fix (link): Corrected autolinking URLs without protocol. Closes #11040.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-link/src/autolink.js
+++ b/packages/ckeditor5-link/src/autolink.js
@@ -10,7 +10,7 @@
 import { Plugin } from 'ckeditor5/src/core';
 import { Delete, TextWatcher, getLastTextLine } from 'ckeditor5/src/typing';
 
-import { addLinkProtocolIfApplicable } from './utils';
+import { addLinkProtocolIfApplicable, linkHasProtocol } from './utils';
 
 const MIN_LINK_LENGTH_WITH_SPACE_AT_END = 4; // Ie: "t.co " (length 5).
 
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-link/src/utils.js
```diff
@@ -165,15 +165,26 @@ export function isEmail( value ) {
...
 
 	return link && isProtocolNeeded ? protocol + link : link;
 }
 
+/**
+ * Checks if protocol is already included in the link
+ *
+ *
+ * @param {String} link
+ * @returns {boolean}
```
scofalik, line 182:
```suggestion
 * @returns {Boolean}
```

packages/ckeditor5-link/src/autolink.js
```diff
@@ -233,16 +233,16 @@ export default class AutoLink extends Plugin {
...
+		if ( !this.isEnabled || !isLinkAllowedOnRange( range, model ) || !linkHasProtocol( parsedUrl ) ) {
 			return;
 		}
 
 		// Enqueue change to make undo step.
 		model.enqueueChange( writer => {
-			const defaultProtocol = this.editor.config.get( 'link.defaultProtocol' );
-			const parsedUrl = addLinkProtocolIfApplicable( link, defaultProtocol );
 			writer.setAttribute( 'linkHref', parsedUrl, range );
-
```
scofalik, line 245: Please don't remove existing empty lines :) (unless this was automated by IDE, btw. @DawidKossowski had the same thing).

Personally, I am a big fan of empty lines and separating code blocks. I think that in our codebase most of use like adding extra lines for readability reasons.

packages/ckeditor5-link/tests/autolink.js
```diff
@@ -237,6 +237,15 @@ describe( 'AutoLink', () => {
 			);
 		} );
 
+		it( 'not autolinks address without protocol if defaultProtocol is not valid', () => {
```
scofalik, line 240:
```suggestion
		it( 'does not autolink address without protocol (defaultProtocol is not set or not valid)', () => {
```

scofalik (changes requested):
Very minor things, as commented.

packages/ckeditor5-link/src/autolink.js
kbinieda: Right, I am also a fan of separating code blocks, but in this case I've removed most of the code from the block so it made sense to me to merge both code blocks together. On the other hand it makes sense to visually separate actual command changes from broadcasting another, undo command.

packages/ckeditor5-link/src/utils.js
```diff
@@ -165,15 +165,26 @@ export function isEmail( value ) {
...
-	const isProtocolNeeded = !!protocol && !PROTOCOL_REG_EXP.test( link );
+	const isProtocolNeeded = !!protocol && !linkHasProtocol( link );
 
 	return link && isProtocolNeeded ? protocol + link : link;
 }
 
+/**
+ * Checks if protocol is already included in the link
+ *
+ *
```
scofalik, line 180:
```suggestion
 * Checks if protocol is already included in the link.
 *
```

PR #12429, opened 2022-09-07
Created custom splice method to deal with very big number of elements

### LINKED ISSUE #12428: Array.splice causes editor to crash
Array.splice causes editor to crash when big number of nodes are inserted into an element.

DawidKossowski:

Fix (engine): Fixed editor throwing `Maximum call stack size exceeded` error when huge number of nodes were inserted at once into an element. Closes #12428.

---

### Additional information

Generally, the editor crashes with error _"Maximum call stack size exceeded"_ when there are at least 200 000 elements in one paragraph and someone opens the revision history feature. This issue has been solved, however, there is still situation when browser crashes because of large content (like 1-2 mln of chars in one paragraph) and this issue cannot be fix on our side.

### DIFF
--- a/packages/ckeditor5-engine/src/model/nodelist.ts
+++ b/packages/ckeditor5-engine/src/model/nodelist.ts
@@ -18,7 +18,7 @@ import spliceArray from '@ckeditor/ckeditor5-utils/src/splicearray';
  * or {@link module:engine/model/documentfragment~DocumentFragment DocumentFragment}.
  */
 export default class NodeList implements Iterable<Node> {
-	private readonly _nodes: Node[];
+	private _nodes: Node[];
 
 	/**
 	 * Creates an empty node list.
@@ -195,7 +195,7 @@ export default class NodeList implements Iterable<Node> {
 			}
 		}
 
-		spliceArray<Node>( this._nodes, Array.from( nodes ), index, 0 );
+		this._nodes = spliceArray<Node>( this._nodes, Array.from( nodes ), index, 0 );
 	}
 
 	/**
--- a/packages/ckeditor5-engine/src/model/nodelist.ts
+++ b/packages/ckeditor5-engine/src/model/nodelist.ts
@@ -10,6 +10,7 @@
 import Node from './node';
 
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+import spliceArray from '@ckeditor/ckeditor5-utils/src/splicearray';
 
 /**
  * Provides an interface to operate on a list of {@link module:engine/model/node~Node nodes}. `NodeList` is used internally
@@ -194,7 +195,7 @@ export default class NodeList implements Iterable<Node> {
 			}
 		}
 
-		this._nodes.splice( index, 0, ...nodes );
+		spliceArray<Node>( this._nodes, Array.from( nodes ), index, 0 );
 	}
 
 	/**
--- a/packages/ckeditor5-utils/src/splicearray.ts
+++ b/packages/ckeditor5-utils/src/splicearray.ts
@@ -13,6 +13,8 @@ const BIG_CHUNK_SIZE = 10000;
  * Splices one array into another. To be used instead of `Array.prototype.splice` as the latter may
  * throw "Maximum call stack size exceeded" when passed huge number of items to insert.
  *
+ * Note: in contrary to Array.splice, this function does not modify the original `target`.
+ *
  * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 0, 0 );		// [ 3, 4, 1, 2 ]
  * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 1, 1 );		// [ 1, 3, 4 ]
  * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 1, 0 );		// [ 1, 3, 4, 2 ]
--- a/packages/ckeditor5-utils/src/splicearray.ts
+++ b/packages/ckeditor5-utils/src/splicearray.ts
@@ -7,16 +7,11 @@
  * @module utils/splicearray
  */
 
+const BIG_CHUNK_SIZE = 10000;
+
 /**
- * Function that splices Arrays and could be used instead of Array.prototype.splice as it results
- * into "Maximum call stack size exceeded" when dealing with very big number.
- *
- * @private
- * @param {Array} target Array to be spliced.
- * @param {Array} source Array of elements to be inserted to target.
- * @param {Number} start Index at which nodes should be inserted/removed.
- * @param {Number} count Number of items.
- * @param {Number} [chunkSize=10000] Chunk size of source array.
+ * Splices one array into another. To be used instead of `Array.prototype.splice` as the latter may
+ * throw "Maximum call stack size exceeded" when passed huge number of items to insert.
  *
  * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 0, 0 );		// [ 3, 4, 1, 2]
  * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 1, 1 );		// [ 1, 3, 4 ]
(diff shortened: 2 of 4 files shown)


packages/ckeditor5-engine/src/model/nodelist.ts
```diff
@@ -194,7 +195,7 @@ export default class NodeList implements Iterable<Node> {
 			}
 		}
 
-		this._nodes.splice( index, 0, ...nodes );
+		spliceArray<Node>( this._nodes, Array.from( nodes ), index, 0 );
```
scofalik, line 198: Although it is not needed by current implementation (as the function "works in place"), I'd go with `this._nodes = spliceArray...` to be future-proof.

scofalik: Same thing in other places in this PR.

packages/ckeditor5-utils/src/splicearray.ts
```diff
@@ -0,0 +1,39 @@
...
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module utils/splicearray
+ */
+
+/**
+ * Function that splices Arrays and could be used instead of Array.prototype.splice as it results
+ * into "Maximum call stack size exceeded" when dealing with very big number.
```
scofalik, line 12:
```suggestion
 * Splices one array into another. To be used instead of `Array.prototype.splice` as the latter may
 * throw "Maximum call stack size exceeded" when passed huge number of items to insert.
```

packages/ckeditor5-utils/src/splicearray.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param {Number} [chunkSize=10000] Chunk size of source array.
+ *
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 0, 0 );		// [ 3, 4, 1, 2]
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 1, 1 );		// [ 1, 3, 4 ]
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 1, 0 );		// [ 1, 3, 4, 2 ]
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 2, 0 );		// [ 1, 2, 3, 4 ]
+ * 		spliceArray( [ 1, 2 ], [], 0, 1 );				// [ 2 ]
+ *
+ */
+export default function spliceArray<T>( target: Array<T>, source: Array<T>, start: number, count: number, chunkSize = 10000 ): Array<T> {
```
scofalik, line 28: I have doubts about `chunkSize`. Is it really needed? Do you see a scenario where you might want to use different chunk sizes? It could be simply hardcoded in a property in this file.

Also, if we would change the implementation, for example to:

```js
return target.slice( ... ).concat( source ).concat( target.slice( ... ) );
```

The parameter would be useless and would pollute code. To me, the parameter looks like implementation detail and not something you would want to change in various calls.

BTW. I know you did some research on this problem. I wonder if you found any solution using slice+concat or maybe you tried your own perf tests? I am asking because slice+concat solution looks much more elegant to me when I look at it now (I know I suggested the implemented solution in the beginning :)).

packages/ckeditor5-utils/src/splicearray.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param {Number} count Number of items.
+ * @param {Number} [chunkSize=10000] Chunk size of source array.
+ *
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 0, 0 );		// [ 3, 4, 1, 2]
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 1, 1 );		// [ 1, 3, 4 ]
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 1, 0 );		// [ 1, 3, 4, 2 ]
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 2, 0 );		// [ 1, 2, 3, 4 ]
+ * 		spliceArray( [ 1, 2 ], [], 0, 1 );				// [ 2 ]
+ *
+ */
```
scofalik, line 27: Missing `@returns` doc.

packages/ckeditor5-utils/src/splicearray.ts
```diff
@@ -0,0 +1,39 @@
...
+ * into "Maximum call stack size exceeded" when dealing with very big number.
+ *
+ * @private
+ * @param {Array} target Array to be spliced.
+ * @param {Array} source Array of elements to be inserted to target.
+ * @param {Number} start Index at which nodes should be inserted/removed.
+ * @param {Number} count Number of items.
+ * @param {Number} [chunkSize=10000] Chunk size of source array.
+ *
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 0, 0 );		// [ 3, 4, 1, 2]
```
scofalik, line 21: The example should be before docs annotations.

scofalik (changes requested):

packages/ckeditor5-engine/src/model/nodelist.ts
scofalik: :thinking: I am not sure why it is readonly, it isn't like that in original .js file. @arkflpc

packages/ckeditor5-utils/src/splicearray.ts
DawidKossowski: I'm back with some info. I implemented the new method as you proposed (let's call it `proposal`) and then compared it with my method (let's call it `actual`) and the conclusions are interesting. I used jsbench to compare the performance providing different datasets:

1.  `target.length = 2`    `source.length = 2`  
    Result: `proposal` is **27,48% slower** 
2.  `target.length = 2`    `source.length = 10000`  
    Result: `proposal` is **66,80% faster**
3.  `target.length = 10`    `source.length = 2`  
    Result: `proposal` is **22,74% slower** 
4.  `target.length = 10 000`  `source.length = 2`  
    Result: `proposal` is **42,43% faster**
5.  `target.length = 10 000`  `source.length = 10000`  
    Result: `proposal` is **58,33% faster**

As you can see... the more data we process the slower `actual` solution works.  
  
**Proposal:** we can implement some kind of hybrid method which will perform normal `Array.prototype.splice` for small set of data and the `propsal` solution for large arrays (because of `splice` method).  
**Drawback:** `Array.prototype.splice` works with reference and `proposal` solution returns new Array.

packages/ckeditor5-engine/src/model/nodelist.ts
arkflpc: Private fields are not in the contract of the class. You can drop `readonly` modifier anytime.

DawidKossowski:
@scofalik all done. Can we merge it?

packages/ckeditor5-utils/src/splicearray.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @private
+ * @param {Array} target Array to be spliced.
+ * @param {Array} source Array of elements to be inserted to target.
+ * @param {Number} start Index at which nodes should be inserted/removed.
+ * @param {Number} count Number of items.
+ *
+ * @returns {Array} New spliced array.
+ *
+ */
```
scofalik, line 30:
```suggestion
 * @returns {Array} New spliced array.
 */
```

packages/ckeditor5-utils/src/splicearray.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @module utils/splicearray
+ */
+
+const BIG_CHUNK_SIZE = 10000;
+
+/**
+ * Splices one array into another. To be used instead of `Array.prototype.splice` as the latter may
+ * throw "Maximum call stack size exceeded" when passed huge number of items to insert.
+ *
+ * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 0, 0 );		// [ 3, 4, 1, 2 ]
```
scofalik, line 16:
```suggestion
 * throw "Maximum call stack size exceeded" when passed huge number of items to insert.
 *
 * Note: in contrary to Array.splice, this function does not modify the original `target`.
 *
 * 		spliceArray( [ 1, 2 ], [ 3, 4 ], 0, 0 );		// [ 3, 4, 1, 2 ]
```

PR #12439, opened 2022-09-08
Make all grid items at InsertTableGrid as an instances of ButtonView

### LINKED ISSUE #12344: Pressing the `space` key doesn't execute the `insertTable` command

1.  Go to a demo with table, e.g. https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/examples/builds-custom/full-featured-editor.html.
2.  Open an insert table toolbar.
3.  Press `space`.

### ✔️ Expected result

Table should be inserted.

### ❌ Actual result

Command is not executed.

### ❓ Possible solution

`Space` should probably just get the same listener as `enter` key.

## 📃 Other details

*   Browser: all
*   OS: all
*   First affected CKEditor version: the one introducing grid keyboard handling (probably 35.1.0)
*   Installed CKEditor plugins: table

---

pszczesniak:

Fix (table): Added keyboard (spacebar) support for table insertion dropdown. Closes #12344 .

---

### Additional information


packages/ckeditor5-theme-lark/theme/ckeditor5-table/inserttable.css
```diff
@@ -23,6 +23,8 @@
 .ck .ck-insert-table-dropdown-grid-box {
 	width: var(--ck-insert-table-dropdown-box-width);
 	height: var(--ck-insert-table-dropdown-box-height);
```
mmotyczynska, line 25: As you added `min-width` and `min-height` these two can be now removed. Regarding the styling I'll add some more information in the review comment.

mmotyczynska (changes requested):
The fix works so nice work :+1:

However there are some changes you have to apply before we merge it. Besides comments in the code please additionally:

1.  Add docs to `this.keystrokes` (my fault, I forgot about this last time :sweat\_smile: )
2.  Update CSS styles so the grid looks just like before. The difference is in the additional outline for the focused item and some transitions which also can be seen when opening the dropdown for the second time - for a short moment we can see the previous selection which does not look good.

Change this:



To this:



And here are transitions that should be removed:

https://user-images.githubusercontent.com/95343030/189202068-62f998a5-225c-4ce1-8050-7ba87d4be1e5.mov

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -124,22 +114,14 @@ export default class InsertTableView extends View {
...
 				} )
 			}
 		} );
 
 		// #rows and #columns are set via changes to #focusTracker on mouse over.
 		this.on( 'boxover', ( evt, domEvt ) => {
 			const { row, column } = domEvt.target.dataset;
-
-			this.items.get( ( row - 1 ) * 10 + ( column - 1 ) ).focus();
+			this.items.get( ( parseInt( row ) - 1 ) * 10 + ( parseInt( column ) - 1 ) ).focus();
```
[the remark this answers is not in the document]
pszczesniak: I've updated it a little bit more.

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -227,7 +239,7 @@ export default class InsertTableView extends View {
 			const row = Math.floor( index / 10 );
 			const column = index % 10;
 
-			boxes.push( new TableSizeGridBoxView( this.locale, row + 1, column + 1, this._geometryLabelId ) );
+			boxes.push( this.createGridButton( this.locale, row + 1, column + 1 ) );
```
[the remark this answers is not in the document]
pszczesniak: Agree.

packages/ckeditor5-table/tests/ui/inserttableview.js
```diff
@@ -52,6 +54,15 @@ describe( 'InsertTableView', () => {
 			expect( view.element.children[ 1 ].classList.contains( 'ck-insert-table-dropdown__label' ) ).to.be.true;
 		} );
 
+		describe( 'createGridButton()', () => {
```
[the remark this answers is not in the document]
pszczesniak: Yes, it makes sense.

packages/ckeditor5-theme-lark/theme/ckeditor5-table/inserttable.css
pszczesniak: Yes, width an height are not needed.

mmotyczynska (approved):

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -215,6 +203,36 @@ export default class InsertTableView extends View {
...
+	 * @private
+	 * @returns {module:ui/button/buttonview~ButtonView}
+	 */
+	_createGridButton( locale, row, column, label ) {
+		const button = new ButtonView( locale );
+		const bind = this.bindTemplate;
+
+		button.set( {
+			label,
+			withText: false,
```
oleq, line 218: This property is false by default, there's no need to set it here https://github.com/ckeditor/ckeditor5/blob/3b33b8fd8291b9bda184f1d3bdd94255e8aa12e3/packages/ckeditor5-ui/src/button/buttonview.js#L62

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -215,6 +203,36 @@ export default class InsertTableView extends View {
...
+			'isOn': false
```
oleq, line 219: `isOn` property is false by default too https://github.com/ckeditor/ckeditor5/blob/3b33b8fd8291b9bda184f1d3bdd94255e8aa12e3/packages/ckeditor5-ui/src/button/buttonview.js#L53

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -215,6 +203,36 @@ export default class InsertTableView extends View {
...
+		} );
+
+		button.extendTemplate( {
+			attributes: {
+				class: [
+					'ck-insert-table-dropdown-grid-box',
```
oleq, line 225: Use `ButtonView#class` instead https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_button_buttonview-ButtonView.html#member-class

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -215,6 +203,36 @@ export default class InsertTableView extends View {
...
+					bind.if( 'isOn', 'ck-on' )
```
oleq, line 226: This is handled internally by `ButtonView` https://github.com/ckeditor/ckeditor5/blob/3b33b8fd8291b9bda184f1d3bdd94255e8aa12e3/packages/ckeditor5-ui/src/button/buttonview.js#L132. You do not have to create another binding here.

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -215,6 +203,36 @@ export default class InsertTableView extends View {
 		} );
 	}
 
+	/**
+	 * Creates a new Button for the grid.
+	 *
+	 * @private
+	 * @returns {module:ui/button/buttonview~ButtonView}
```
oleq, line 210: Missing docs `@param` for `locale`, `row`, `column`, and `label`.

oleq (changes requested):
Works fine, I didn't notice any performance issues.

I left some comments regarding implementation, though.

BTW: I noticed that the whole thing uses `flex` https://github.com/ckeditor/ckeditor5/blob/3b33b8fd8291b9bda184f1d3bdd94255e8aa12e3/packages/ckeditor5-table/theme/inserttable.css#L7 instead of CSS grid. I suppose 4 years ago CSS grid was not widely supported (or there were bugs). Maybe since we're at it it's worth cleaning this up? (probably a new issue)

packages/ckeditor5-table/src/ui/inserttableview.js
pszczesniak: filled

packages/ckeditor5-table/src/ui/inserttableview.js
pszczesniak: Thank you, I need to work more closely with using documentation and code.

pszczesniak:
Yes, when i looked at support for grid: [grid support](https://caniuse.com/css-grid), i think that we can change flex to grid without any issues (but of course would be good to check it on some major devices/browsers (would be good to have it in separate ticket).

oleq (changes requested):

packages/ckeditor5-table/tests/ui/inserttableview.js
```diff
@@ -58,6 +56,12 @@ describe( 'InsertTableView', () => {
 				expect( view.items ).to.have.length( 100 );
 			} );
 
+			it( 'should create a single new grid button', () => {
+				const btn = view._createGridButton( locale, 1, 1, '1 × 1' );
```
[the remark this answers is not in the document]
pszczesniak: I should ask at the beginning about testing private methods.

oleq (approved):

oleq:
> Yes, when i looked at support for grid: [grid support](https://caniuse.com/css-grid), i think that we can change flex to grid without any issues (but of course would be good to check it on some major devices/browsers (would be good to have it in separate ticket).

Extracted to #12471

PR #12445, opened 2022-09-08
Set tabindex to -1 to prevent focusing the checkbox element with Tab key

### LINKED ISSUE #6535: To-do list is a focus trap

1.  Open https://ckeditor.com/docs/ckeditor5/latest/features/todo-lists.html
2.  Focus the editor by placing the selection anywhere in its content
3.  Press the `tab` key

### ✔️ Expected result

The focus should go outside of the editor

### ❌ Actual result

Focus for a brief moment goes to the first input checkbox (you can see it because _focus ring_ blinks for a moment) and goes back to your selection position.

## 📃 Other details

This is a rather major accessibility issue.

*   Browser: Any
*   OS: Any

---

pszczesniak:

Fix (list): To-do list items should not interrupt the Tab key navigation across the editor content. Closes #6535 .

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/tests/todolist/todolistediting.js
+++ b/packages/ckeditor5-list/tests/todolist/todolistediting.js
@@ -1265,7 +1265,6 @@ describe( 'TodoListEditing - checkbox rendering', () => {
 
 		expect( checkboxElement.tagName ).to.equal( 'INPUT' );
 		expect( checkboxElement.checked ).to.equal( false );
-		expect( checkboxElement.hasAttribute( 'tabindex' ) ).to.be.true;
 		expect( checkboxElement.getAttribute( 'tabindex' ) ).to.equal( '-1' );
 	} );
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-list/tests/todolist/todolistediting.js
```diff
@@ -1265,6 +1265,8 @@ describe( 'TodoListEditing - checkbox rendering', () => {
 
 		expect( checkboxElement.tagName ).to.equal( 'INPUT' );
 		expect( checkboxElement.checked ).to.equal( false );
+		expect( checkboxElement.hasAttribute( 'tabindex' ) ).to.be.true;
```
oleq, line 1268: This assertion is obsolete. `Element#getAttribute()` always returns `null` for a non-existing name. The next assertion will handle this.

oleq (changes requested):
Please keep in mind that the commit merge message you propose will go directly to `CHANGELOG.md` file and https://ckeditor.com/docs/ckeditor5/latest/updating/changelog.html read by integrators. Try focusing on whys (user story) instead of whats (technical details) because this will allow developers understand the scope of the release. They may not understand the internals that we change but they can certainly relate to the problems of the user/business requirements.

For instance, something along the lines of "To-do list items should not interrupt the Tab key navigation across the editor content." should do the trick.

Bonus tip: Almost every "Fix (...)" change can be described using "should/should not".

PR #12446, opened 2022-09-08
IME support on Android

### LINKED ISSUE #12058: [beforeInput] Bring back basic typing support in Android
In a course of action in https://github.com/ckeditor/ckeditor5/issues/11438 we wrote a composition support for non-Android browsers (see https://github.com/ckeditor/ckeditor5/pull/12024).

Unfortunately, while it works really well on every other system and browser and every language (including CJK), it does not work on Android at all.

We need to implement a completely custom handling for Android browsers.

Keyboards that we'll try to support from day one:

* The default one
* GBoard
* Swift

## ToC

* PoC ([done](https://github.com/ckeditor/ckeditor5/issues/12058#issuecomment-1220904956) ✅ )
* And then stabilize:
  * #12391
  * #12369
  * #12368

niegowski:

Internal (engine, typing, enter, widget): Support for IME on Android. Closes #12058.

---

### Additional information


packages/ckeditor5-engine/tests/view/renderer.js
```diff
@@ -4743,7 +4909,7 @@ describe( 'Renderer', () => {
 					renderer.isSelecting = false;
 				} );
 
-				it( 'should remove the inline filler despite the user making selection', () => {
+				it.skip( 'should remove the inline filler despite the user making selection', () => {
```
arkflpc, line 4912: What do we want to do with those tests?

packages/ckeditor5-widget/src/widgettypearound/widgettypearound.js
```diff
@@ -647,10 +648,21 @@ export default class WidgetTypeAround extends Plugin {
...
 
-		// Note: The priority must precede the default Input plugin compositionstart handler (to call it before delete content).
-		this._listenToIfEnabled( viewDocument, 'compositionstart', () => {
-			this._insertParagraphAccordingToFakeCaretPosition();
-		}, { priority: 'high' } );
+		if ( env.isAndroid ) {
+			// On Android with English keyboard, the composition starts just by putting caret
+			// at the word end or by selecting a table column. This is not a real composition started.
+			// Trigger delete content on first composition key pressed.
+			this.listenTo( viewDocument, 'keydown', ( evt, data ) => {
```
arkflpc, line 655: \_listenToIfEnabled?

arkflpc (approved):

packages/ckeditor5-engine/src/view/observer/inputobserver.ts
```diff
@@ -68,32 +70,50 @@ export default class InputObserver extends DomEventObserver<'beforeinput'> {
 		// in the fake selection container.
```
Reinmar, line 70: Please update this comment with the note what we're doing for Android and WHY.

packages/ckeditor5-engine/src/view/observer/inputobserver.ts
```diff
@@ -68,32 +70,50 @@ export default class InputObserver extends DomEventObserver<'beforeinput'> {
...
 		}
 
-		this.fire( domEvent.type, domEvent, {
-			data,
-			dataTransfer,
-			isComposing: domEvent.isComposing,
-			targetRanges,
-			inputType: domEvent.inputType
-		} );
+		if ( env.isAndroid && domEvent.inputType == 'insertCompositionText' && data && data.endsWith( '\n' ) ) {
```
Reinmar, line 103: Missing a comment explaining what is this case. Especially that we even had a ticket for that so it's easy to explain why it's needed.

packages/ckeditor5-engine/src/view/observer/inputobserver.ts
```diff
@@ -68,32 +70,50 @@ export default class InputObserver extends DomEventObserver<'beforeinput'> {
...
+			this.fire( domEvent.type, domEvent, {
+				inputType: 'insertParagraph',
+				targetRanges: [ view.createRange( targetRanges[ 0 ].end ) ]
+			} );
+		} else {
```
Reinmar, line 108: I don't like that the default (expected) behavior of this listener is in an `else{}` block because it made it harder for me to understand that the `if()` part is an exception and then there goes the default.

More naturally for me it would be like this:

```js
if ( specialCondition ) {
   doSpecialThings();
   return;
}

doDefaultThings();
```

However, then, there's the `console.groupEnd()` thing, so we can't actually make it an early return (because we'd miss the console thing).

So... maybe just document the `else{}` block:

```js
// Special case for Android: ...
if( ) {
   ...
}
// Default action: Fire (what do we actually fire here?! Document what it could bew)
else {
   ...
}
```

packages/ckeditor5-engine/src/view/observer/selectionobserver.ts
```diff
@@ -182,7 +183,7 @@ export default class SelectionObserver extends Observer {
 			// @if CK_DEBUG_TYPING // 	);
 			// @if CK_DEBUG_TYPING // }
 
-			if ( this.document.isComposing ) {
+			if ( this.document.isComposing && !env.isAndroid ) {
```
Reinmar, line 186: Why not on Android? Missing a comment.

packages/ckeditor5-engine/src/view/renderer.ts
```diff
@@ -240,7 +239,7 @@ export default class Renderer extends Observable {
 	 * removed as long as the selection is in the text node which needed it at first.
 	 */
 	public render(): void {
-		if ( this.isComposing ) {
+		if ( this.isComposing && !env.isAndroid ) {
```
Reinmar, line 242: Explanation needed.

packages/ckeditor5-engine/src/view/renderer.ts
```diff
@@ -562,6 +565,12 @@ export default class Renderer extends Observable {
 			return false;
 		}
 
+		// Do not use inline filler while typing outside inline elements on Android.
+		// The deleteContentBackward would remove part of the inline filler instead of removing last letter in a link.
+		if ( env.isAndroid && ( nodeBefore || nodeAfter ) ) {
```
Reinmar, line 570: I like the comment. Could we also add a link to a ticket? I think we had this reported.

packages/ckeditor5-engine/src/view/renderer.ts
```diff
@@ -1163,3 +1227,35 @@ function createFakeSelectionContainer( domDocument: DomDocument ): DomElement {
 
 	return container;
 }
+
+function updateTextNode( domText: DomText, expectedText: string ) {
```
Reinmar, line 1231: Docs needed: what's the general algorithm it follows.

Reinmar: **And why it does what it does.**

Reinmar: I can see it's a copy paste from another place. So perhaps there's not much "why". But I'd still prefer if we had at least one sentence of what it does.

packages/ckeditor5-engine/src/view/renderer.ts
```diff
@@ -667,6 +680,12 @@ export default class Renderer extends Observable {
 
 		const diff = this._diffNodeLists( actualDomChildren, expectedDomChildren );
 
+		// We need to check whether there are some text nodes to update data to make sure that replacing such nodes
```
Reinmar, line 683: I couldn't understand this explanation.

I'd start from what's different between the flow for Android and other browsers. Explain first the context.

Then try to explain why it's here and how it changes the behavior later on. It's hard for me e.g. to understand what's the difference between `diff` initially and after passing it via `actions`. And then I'd need to understand why that difference was needed.

We may also need to clarify why we don't do `findReplaceActions()` on other browsers. Especially if the reason was "we don't want to change the behavior on other browsers for safety, but maybe one day cause it seems to make sense".

packages/ckeditor5-engine/src/view/renderer.ts
```diff
@@ -676,21 +695,45 @@ export default class Renderer extends Observable {
...
 				insertAt( domElement as DomElement, i, expectedDomChildren[ i ] );
 				i++;
+			} else if ( action === 'replace' ) {
+				// @if CK_DEBUG_TYPING // if ( window.logCKETyping ) {
+				// @if CK_DEBUG_TYPING // 	console.group( '%c[Renderer]%c Update text node',
+				// @if CK_DEBUG_TYPING // 		'color: green;font-weight: bold', ''
+				// @if CK_DEBUG_TYPING // 	);
+				// @if CK_DEBUG_TYPING // }
+
+				updateTextNode( actualDomChildren[ i ] as DomText, ( expectedDomChildren[ i ] as DomText ).data );
```
Reinmar, line 731: AFAICS this method so far was only touching elements but now it also updates text nodes. This only happens on Android, right? If so:

* Add info about what this method does for Android to its documentation.
* Document this `else if()` that it's Android only.

packages/ckeditor5-engine/src/view/renderer.ts
```diff
@@ -826,9 +882,11 @@ export default class Renderer extends Observable {
 		// Render selection.
 		if ( this.selection.isFake ) {
```
Reinmar, line 883: The chain of these conditions is hard for me to understand.

I also can't understand the difference between what we had before and what's there now.

So far it seems that we were calling `updateDomSelection()` when `isComposing` was `true`. Right now we don't do that (or more precisely: do this only when there was fakeSelContainer). This sounds like a big change compared to what we had before. Could you explain why you made it (in this PR)?

Reinmar: PS. And most of that needs to end up in the comments :D

packages/ckeditor5-engine/src/view/view.ts
```diff
@@ -75,7 +75,6 @@ export default class View extends Observable {
 	public readonly domRoots: Map<string, HTMLElement>;
 
 	declare public isFocused: boolean;
-	declare public isSelecting: boolean;
```
Reinmar, line 78: Breaking change?

packages/ckeditor5-enter/src/enter.js
```diff
@@ -38,7 +38,11 @@ export default class Enter extends Plugin {
 		editor.commands.add( 'enter', new EnterCommand( editor ) );
 
 		this.listenTo( viewDocument, 'enter', ( evt, data ) => {
-			data.preventDefault();
+			// Rendering is disabled while composing so prevent events that will be rendered by the engine
```
Reinmar, line 41: I'd reword it because I had troubles understanding this comment. Maybe something like:

> When not in composition, we handle the action, so prevent the default one.
> When in composition, it's the browser who modify the DOM (renderer is disabled).

packages/ckeditor5-typing/src/deleteobserver.js
```diff
@@ -155,25 +155,20 @@ export default class DeleteObserver extends Observer {
...
 
-			// Android IMEs have a quirk. Sometimes it may change the DOM selection on `beforeinput` event so that
-			// the selection contains all the text that the IME wants to remove. But sometimes it is only expanding
-			// it by a single character (in case of the collapsed selection).
-			//
-			// The code below checks if the former scenario occurred (the latter is fine, needs no correction) and it
-			// uses this information to correct the "delete" event so it knows the proper part of the content to be removed.
-			//
-			// **Note**: See the Delete plugin for the second part of this quirk.
+			// Split deleting a single character from deleting the selected (wider) range.
```
Reinmar, line 158: Why? You removed the comment that actually explained why this code is needed here. Although, I can see the code changed too, so perhaps the old comment was outdate. Anyway, new one is needed.

packages/ckeditor5-typing/src/input.js
```diff
@@ -10,6 +10,7 @@
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import InsertTextCommand from './inserttextcommand';
 import InsertTextObserver from './inserttextobserver';
+import env from '@ckeditor/ckeditor5-utils/src/env';
```
Reinmar, line 13: I like the comments in this file :D I spent 2 minutes scanning it thanks to them, instead of 5x more as on some other files. Just to give the perspective on why comments are important even for the review.

packages/ckeditor5-typing/src/inserttextobserver.js
```diff
@@ -37,6 +38,10 @@ export default class InsertTextObserver extends Observer {
 	constructor( view ) {
 		super( view );
 
+		if ( env.isAndroid ) {
```
Reinmar, line 41: Why?

packages/ckeditor5-typing/src/inserttextobserver.js
```diff
@@ -66,7 +71,7 @@ export default class InsertTextObserver extends Observer {
 
 		// Note: The priority must be lower than the CompositionObserver handler to call it after the renderer is unblocked.
 		viewDocument.on( 'compositionend', ( evt, { data, domEvent } ) => {
-			if ( !this.isEnabled ) {
+			if ( !this.isEnabled || env.isAndroid ) {
```
Reinmar, line 74: Why?

Reinmar:
I left some comments in the code. Pretty much all about missing comments. Implementation-wise, I noticed only small iffy places where non-Android browsers were affected and I'd like to know why (whether it was on purpose or not). But there are just a few such places – I can see you cleaned up this branch significantly since the last time I saw it.

Other than that, it's all about "does it work" so we've got it covered already. But I want to ensure now that we can figure out why some pieces of this code are needed. Especially that some were even related to bugs that we found and we have specific tickets regarding those.

Anyways, once the "whys" are rediscovered and commented, it's ready to be merged to the base beforeInput branch. And let's then see this on `master` and... :crossed\_fingers: 

Once again: great job! Let's now polish it and have it merged.

packages/ckeditor5-engine/src/view/renderer.ts
niegowski: Can't find the ticket. I'm not sure if it was written on any ticket.

packages/ckeditor5-engine/src/view/renderer.ts
niegowski: https://github.com/ckeditor/ckeditor5/blob/2b9e30f34e5b87c0dd27613a3e6477129798e111/packages/ckeditor5-engine/src/view/renderer.ts#L897-L913

packages/ckeditor5-engine/src/view/view.ts
niegowski: Nope, it's just a `declare` that wasn't used anywhere.

PR #12460, opened 2022-09-13
Fix (link): Don't undo autolinking when any other change has been made.

### LINKED ISSUE #11040: Auto-link correction

### LINKED ISSUE #12447: Auto-link undo correction

1. type "https://ckeditor.com abc" in content (hyperlink should be created for ckeditor.com, while "abc" still remains as text, which is fine)
2. slowly (one character at a time) hit backspace to remove " abc" text (including space).
3. Instead of removing space, editor undo character removal to initial "https://ckeditor.com abc" situation.

### ✔️ Expected result

just "https://ckeditor.com" in editor

### ❌ Actual result

"https://ckeditor.com abc" in editor

kbinieda:

Fix (link): Don't undo autolinking when any other change has been made. Closes #12447 .

### DIFF
--- a/packages/ckeditor5-link/src/autolink.js
+++ b/packages/ckeditor5-link/src/autolink.js
@@ -243,7 +243,7 @@ export default class AutoLink extends Plugin {
 	}
 
 	/**
-	 * Enqueue autolink changes in model.
+	 * Enqueues autolink changes in the model.
 	 *
 	 * @param {String} url The URL to link.
 	 * @param {module:engine/model/range~Range} range The text range to apply the link attribute to.
--- a/packages/ckeditor5-link/src/autolink.js
+++ b/packages/ckeditor5-link/src/autolink.js
@@ -223,26 +223,39 @@ export default class AutoLink extends Plugin {
 	}
 
 	/**
-	 * Applies a link on a given range.
+	 * Applies a link on a given range if the link should be applied.
 	 *
 	 * @param {String} url The URL to link.
 	 * @param {module:engine/model/range~Range} range The text range to apply the link attribute to.
 	 * @private
 	 */
-	_applyAutoLink( link, range ) {
+	_applyAutoLink( url, range ) {
 		const model = this.editor.model;
-		const deletePlugin = this.editor.plugins.get( 'Delete' );
 
 		const defaultProtocol = this.editor.config.get( 'link.defaultProtocol' );
-		const parsedUrl = addLinkProtocolIfApplicable( link, defaultProtocol );
+		const fullUrl = addLinkProtocolIfApplicable( url, defaultProtocol );
 
-		if ( !this.isEnabled || !isLinkAllowedOnRange( range, model ) || !linkHasProtocol( parsedUrl ) ) {
+		if ( !this.isEnabled || !isLinkAllowedOnRange( range, model ) || !linkHasProtocol( fullUrl ) || linkIsAlreadySet( range ) ) {
 			return;
 		}
 
+		this._persistAutoLink( fullUrl, range );
+	}
+
+	/**
+	 * Enqueue autolink changes in model.
+	 *
+	 * @param {String} url The URL to link.
+	 * @param {module:engine/model/range~Range} range The text range to apply the link attribute to.
+	 * @protected
+	 */
+	_persistAutoLink( url, range ) {
+		const model = this.editor.model;
+		const deletePlugin = this.editor.plugins.get( 'Delete' );
+
 		// Enqueue change to make undo step.
 		model.enqueueChange( writer => {
-			writer.setAttribute( 'linkHref', parsedUrl, range );
+			writer.setAttribute( 'linkHref', url, range );
 
 			model.enqueueChange( () => {
 				deletePlugin.requestUndoOnBackspace();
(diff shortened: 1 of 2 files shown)


scofalik:
As far as I can see, the fix here is to prevent additional auto-linking of already linked text. AFAICS, This can happen in two scenarios:

* when you move selection before the link and type space again,
* when you start removing text, e.g. `http://foo.pl abc[]` -> `http://foo.pl []`.

I am quite sure that we don't want to double auto-link in such cases :).

However, I have two thoughts regarding this.

One, I wonder if `TextWatcher` should fire an event on text deletion. IDK if there are cases where this could be useful/used. In general, if you are deleting stuff, you must have added it earlier, which means that the `TextWatcher` already had a chance to act. Unless you pasted some text (e.g. you pasted `lorem http://foo.pl ipsum`). But in this case, AFAIK, we don't want `TextWatcher` to autolink after removing `ipsum` right? So, maybe we should also change how `TextWatcher` works? Nevertheless, the proposed if will help in the first case (moving selection before the link and typing space again).

Second, I think that I'd add one more unit test. Proposed unit test is fine, as it targets exactly the scenario described in the ticket. However, I'd add one more unit test, that would make sure that we don't autolink already linked text (you can use the scenario with moving selection and typing space again). This test will target the actual change.

cc @Reinmar (regarding `TextWatcher`).

scofalik (changes requested):
Please add one more test (as described in my previous comment).

kbinieda:
Aside from adding additional test, I've broke down _applyAutoLink function so that the moment of applying linkHref can be easily tested.

packages/ckeditor5-link/tests/autolink.js
```diff
@@ -414,6 +425,27 @@ describe( 'AutoLink', () => {
 				'<paragraph>https://www.cksource.com []</paragraph>'
 			);
 		} );
+
+		it( 'should not undo auto-linking by pressing backspace after any other change has been made', () => {
```
scofalik, line 429: Could you add a link to the issue in the comment above this line? This test will seem very weird to someone looking at it in a year :).

scofalik (changes requested):
Small fixes required.

packages/ckeditor5-link/tests/autolink.js
```diff
@@ -246,6 +246,17 @@ describe( 'AutoLink', () => {
 			);
 		} );
 
+		it( 'does not autolink if link is already created', () => {
```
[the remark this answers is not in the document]
kbinieda: That's a good point. The "different text" approach gives us different result which we can then easily compare, I haven't thought of that. On the other hand If we want to check if same text/link hasn't been applied twice I would go for a method spy as it is more readable than operations counting.

Reinmar:
> One, I wonder if `TextWatcher` should fire an event on text deletion. IDK if there are cases where this could be useful/is already used. In general, if you are deleting stuff, you must have added it earlier, which means that the `TextWatcher` already had a chance to act. Unless you pasted some text (e.g. you pasted `lorem http://foo.pl ipsum`). But in this case, AFAIK, we don't want `TextWatcher` to autolink after removing `ipsum` right? So, maybe we should also change how `TextWatcher` works?

TextWatcher is also used by mentions. I tested quickly how mentions work e.g. in Slack and when you start with this:

```
Hello, @reinmarX[]
```

And press delete, the mention panel will be triggered once "X" is removed.

Our current implementation, from what I quickly tested, works the same way as Slack's. So I consider this correct.

However, there's a chance that TextWatcher should indeed be improved. For instance, if you a content like this in GH Writer:

```
Text[]
#123
#234
Text
```

And start to navigate through this content with the up/down arrows, you'll be constantly triggering mentions, blocking the navigation. GH's native editor does not have this shortcoming. But... Slack's has if you have content with unfinished mentions:

```
Hello, @re[]
Hello, @re
```

Long story short, I'd be really cautious when changing something, but there's a chance that we should.

There's also a chance that autolink should work differently than mentions.

packages/ckeditor5-link/src/autolink.js
```diff
@@ -223,26 +223,39 @@ export default class AutoLink extends Plugin {
...
-		if ( !this.isEnabled || !isLinkAllowedOnRange( range, model ) || !linkHasProtocol( parsedUrl ) ) {
+		if ( !this.isEnabled || !isLinkAllowedOnRange( range, model ) || !linkHasProtocol( fullUrl ) || linkIsAlreadySet( range ) ) {
 			return;
 		}
 
+		this._persistAutoLink( fullUrl, range );
+	}
+
+	/**
+	 * Enqueue autolink changes in model.
```
Reinmar, line 246:
```suggestion
	 * Enqueue autolink changes in the model.
```

scofalik:
I forgot that `TextWatcher` has uses outside of autoformatting text.

> There's also a chance that autolink should work differently than mentions.

Yes. But having mentions in mind, it seems that it is better to add modifications in auto-formatting features (autoformat, autolink, etc.) if needed. We could also manage this behavior through an option parameter when creating `TextWatcher` instance (to what it should react: typing, deleting, selection change).

But this is outside of the scope of this PR. I will create a new issue for this discussion.

Issue #12432, opened 2022-09-07
Improve schema-related vocabulary and Schema API


scofalik:
## 📝 Provide a description of the improvement

Related to: #12430.

I feel there is confusion regarding naming when it comes to schema. It's about "block" word.

We have two categories of elements, that are "blocks":

* non-object block elements, which hold inline content (mostly text), such as paragraph or heading,
* object block elements, such as image or table.

These two are totally different, they behave differently, etc. It is a problem when providing descriptions or explaining, because you always have to mention the other category or specify whether you are talking about object-block or non-object-block. Finally, some blocks break schema rules. In Schema guide it is said that blocks cannot contain blocks. But table can be inside a table. I think this rule make sense only for non-object-blocks.

It is also a problem when writing features. When it comes to `Schema` API, we have `.isBlock()` but it is not enough. We always have to use `&& schema.isObject( ... )` (or negation).

I don't have a good proposition for names because images or tables feel like something that can be called "block". And also, we are used to calling paragraphs "blocks". My main problem is that calling `.isBlock()` doesn't give me enough information. And saying that something is "block" is not enough. 

Additionally, we should start thinking in behavior categories. I can identify these categories:

* container (block quote, section),
* limit-container (some custom feature? maybe templates?),
* block object (non-inline image, table), ("object"?)
* block element (paragraph, heading), ("block"?)
* inline (text, inline image).

All elements from given category should behave the same (maybe inline elements are a bit different than text when it comes to selection handling). Apart from inline objects, I think that other categories are exclusive (i.e. you can't have something that is container and block at the same time, etc.).

A lot of time we wonder how an element should behave regarding things like: enter key, selection handling, auto-paragraphing, automatic schema-checks, `insertContent()`, track changes feature, revision history feature. This gets complicated with features like GHS or with custom elements introduced in 3rd party features. This is why I'd like to focus on these five categories and assume that any model element will always fit into one of these (unless a new category emerges).

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
[...]

PR #12461, opened 2022-09-13
Fix (html-support): Redefine most GHS-elements as not blocks in schema.

### LINKED ISSUE #12430: Multiple GHS-elements are incorrectly defined as blocks in schema

GHS feature introduces multiple model elements that represent related HTML tags. Most of those elements/tags are for creating document structure, like `<article>` (`htmlArticle`) or `<section>` (`htmlSection`). We call such elements "containers". Considering CKE5 features, there is just one container defined and it is `blockQuote` element.

Those GHS-related elements are incorrectly marked as blocks in model schema. I didn't make full list but I guess that most these elements are incorrect. However, `htmlDivParagraph` should stay a block.

Due to this incorrect setting, [we have problems with styles feature](https://github.com/ckeditor/ckeditor5/issues/12417) and also with other features (e.g. heading feature will change section into a heading...).

Additional context:

Block elements are:

*   paragraph, heading, list item, etc. (non-objects),
*   table, image, media, etc. (objects).

Block elements are "low level" when it comes to document structure, they are either content-holders (like paragraph) or they "are content" themselves (like image). Non-object block elements are expected to allow inline elements in them. It is also possible to split them (using enter key) and merge them together (delete). Finally, non-object block elements cannot contain other block elements.

_(Side note, I dislike, that images are blocks and paragraphs are blocks as well, as they are totally different: #12432_.)

On the other hand, containers keep other containers and blocks to provide more complex document structure. [Containers are those elements that are not any other type of element, like block quote](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/schema.html#defining-additional-semantics). Block quote works totally different than paragraphs. We intuitively feel that paragraph and block quote are in different groups of elements.

Finally, we use `selection.getSelectedBlocks()` in many places in our features and we expect that it returns paragraph-like elements.

kbinieda:

Fix (html-support): Redefine most GHS-elements as not blocks in schema. Closes #12430.

### DIFF
--- a/packages/ckeditor5-html-support/src/schemadefinitions.js
+++ b/packages/ckeditor5-html-support/src/schemadefinitions.js
@@ -135,7 +135,7 @@ export default {
 			view: 'tbody',
 			modelSchema: {
 				allowIn: 'htmlTable',
-				isBlock: true
+				isBlock: false
 			}
 		},
 		{
@@ -143,7 +143,7 @@ export default {
 			view: 'thead',
 			modelSchema: {
 				allowIn: 'htmlTable',
-				isBlock: true
+				isBlock: false
 			}
 		},
 		{
@@ -151,7 +151,7 @@ export default {
 			view: 'tfoot',
 			modelSchema: {
 				allowIn: 'htmlTable',
-				isBlock: true
+				isBlock: false
 			}
 		},
 		{
@@ -160,7 +160,8 @@ export default {
 			modelSchema: {
 				allowIn: 'htmlTable',
 				allowChildren: '$text',
-				isBlock: true
+				isBlock: false,
+				isLimit: true
 			}
 		},
 		{
@@ -169,7 +170,7 @@ export default {
 			modelSchema: {
 				allowIn: 'htmlTable',
 				allowChildren: 'col',
-				isBlock: true
+				isBlock: false
 			}
 		},
 		{
@@ -177,7 +178,7 @@ export default {
 			view: 'col',
 			modelSchema: {
 				allowIn: 'htmlColgroup',
-				isBlock: true
+				isBlock: false
 			}
 		},
 		{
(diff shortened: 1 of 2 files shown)


scofalik:
I think that `htmlCaption` and `htmlFigCaption` should not be blocks. They are not splittable and merge'able (using enter and backspace). In general, they aren't like paragraphs. They can include text, though. I'd say that's an exception, as I earlier thought that `$text` can be only in `$block`s.

As we discussed with @kbinieda, we don't have a strict definitions for schema items and this is causing confusion.

Other than that, I think that, `htmlFigure` should be block (and object). This is another confusing part, where block+object is something completely different than block+non-object. I wrote about it here: #12432. However, I'd test how this behaves.

scofalik:
Actually, I'd check how `htmlCaption` and `htmlFigCaption` behaves when it is `isBlock: false` and `isLimit: true`. That is how our own caption is defined is schema but we have special handling for our own caption (as it is our own feature). I am not sure if the editor will have good UX for a "generic" block-non-object-limit.

packages/ckeditor5-html-support/src/schemadefinitions.js
```diff
@@ -211,7 +212,7 @@ export default {
 			view: 'figure',
 			modelSchema: {
 				inheritAllFrom: '$container',
-				isBlock: true
+				isBlock: false
```
scofalik, line 215:
```suggestion
				isBlock: true
```

scofalik:
I proposed some additional changes for some elements however I am not 100% sure about them.

I changed `tr`, `td` and `th` to better reflect schema that we have for them in the actual feature. Previously, you were able to do such silly things like removing `<td>` if you clicked backspace too much and removed whole content.

I also reverted changes for `figcaption` and `caption`. When changed to `isLimit` it was difficult to remove them and, as a result, it was difficult to remove `figure` that contained them. It was possible, but you had to make a selection that contained the whole figure. I believe that enter should not split `figcaption` and unfortunately, after revert, it does split it. But I think this is a smaller evil. I'd rather be able to easily remove the `figcaption`.

PR #12472, opened 2022-09-14
Ck/12426 column resize conversion for width style on cell

### LINKED ISSUE #12426: Column resize should provide conversion for width style on individual cell

Currently if you copy (to your clipboard) a table like:

```html
<table style="width:399px">
	<tbody>
		<tr>
			<td style="width:94px">&nbsp;</td>
			<td style="width:291px">&nbsp;</td>
		</tr>
		<tr>
			<td style="width:94px">&nbsp;</td>
			<td style="width:291px">&nbsp;</td>
		</tr>
	</tbody>
</table>
```

Table in CKEditor 4

And paste it into the CKEditor 5 [column resize sample](https://ckeditor.com/docs/ckeditor5/35.1.0/features/table.html#table-column-resize) it will ignore the cell widths and look in a following way:

The reason for this is that we didn't include _cell properties plugin_ in this sample. This plugin brings [conversion and command for cell width](https://zoom.us/j/89758913109?pwd=eGlWZFZ6R1BWdEd0WlBJN3pGSnNKQT09).

The problem is that it includes also plenty of other features (like cell background, border) that are not desired in case someone enables the column resize plugin. So we need to extract cell width to an atomic plugin that will be required by both `CellPropertiesEditing` and `TableColumnResizeEditing`.

To sum up:

1.  Let's extract cell width to `CellWidthEditing` plugin.
2.  This plugin needs to go to `requires` list of cell properties editing and column resize editing plugin.

## 📃 Other details

*   Browser: Any
*   OS: Any
*   CKEditor version: …

pszczesniak:

Other (table): Pasting table with setted width values on cells should be done correctly now. Closes #12426.

---

### Additional information


packages/ckeditor5-table/src/tablecellproperties.js
```diff
@@ -35,7 +36,7 @@ export default class TableCellProperties extends Plugin {
 	 * @inheritDoc
 	 */
 	static get requires() {
-		return [ TableCellPropertiesEditing, TableCellPropertiesUI ];
+		return [ TableCellPropertiesEditing, TableCellPropertiesUI, TableCellWidthEditing ];
```
mlewand, line 39: The `TableCellWidthEditing` plugin should be required by `TableCellPropertiesEditing` rather than the glue plugin.

Reason for that is that previously it was the `TableCellPropertiesEditing` plugin that implemented conversion. However, now when you moved it to the "glue plugin", the `TableCellPropertiesEditing` doesn't do that anymore - which is a breaking change 🙂

packages/ckeditor5-table/src/tablecellwidth/tablecellwidthediting.js
```diff
@@ -0,0 +1,84 @@
...
+		const defaultTableCellProperties = getNormalizedDefaultProperties(
+			editor.config.get( 'table.tableCellProperties.defaultProperties' )
+		);
+
+		enableProperty( schema, conversion, {
+			modelAttribute: 'tableCellWidth',
+			styleName: 'width',
+			defaultValue: defaultTableCellProperties.width
+		} );
+		editor.commands.add( 'tableCellWidth', new TableCellWidthCommand( editor, defaultTableCellProperties.width ) );
```
Dumluregn, line 62: Also, I just noticed that we have both a command and a model attribute called `tableCellWidth` :scream: it seems like a dangerous idea as it can be confusing, but AFAICS all (or most) of the table and table cell commands follow this pattern, so it looks like an intentional design :thinking: Still, we can't make such a change in isolation and it would be a big breaking change for all the integrators, so we'll have to live with it.

Dumluregn (changes requested):
Besides Marek's comment, I miss a unit test for the table column resize feature - it should be enough to just set the data to sth like

```html
<table style="width:399px">
	<tbody>
		<tr>
			<td style="width:94px">&nbsp;</td>
			<td style="width:291px">&nbsp;</td>
		</tr>
		<tr>
			<td style="width:94px">&nbsp;</td>
			<td style="width:291px">&nbsp;</td>
		</tr>
	</tbody>
</table>
```

and check if the widths were trimmed or stayed.

Other than that mostly refactors and cleanups.

packages/ckeditor5-table/src/tablecellwidth/tablecellwidthediting.js
pszczesniak: Yes, i've seen that pattern and i followed with it.

packages/ckeditor5-table/src/tablecellwidth/tablecellwidthediting.js
```diff
@@ -0,0 +1,84 @@
...
+	 */
+	static get pluginName() {
+		return 'TableCellWidthEditing';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	static get requires() {
+		return [ TableEditing ];
```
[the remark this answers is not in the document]
Dumluregn: Actually it is required due to the `schema.extend()` call in the plugin - without `TableEditing` the editor throws an error. Sorry for that.

Dumluregn (approved):
👍🏻 

I did some final polishing.

Issue #11719, opened 2022-05-06
[TS MVP] Update ESLint configuration


arkflpc:
We need ESLint helping us with TypeScript code as well as with JavaScript.

Enable ESLint for `*.ts` files and verify that:

*    `yarn run lint` checks them,
*   `git commit` checks them,
*   there is any support in IDE (like VS Code).

We should keep all existing rules for JavaScript that have been working for us, but we need to decide about TypeScript-specific ones.

Reinmar:
Potentially, we'll need to release a new packages.

We have https://www.npmjs.com/package/eslint-config-ckeditor5 and https://www.npmjs.com/package/eslint-plugin-ckeditor5-rules already. E.g. the latter package contains our custom rules – we may need to port or extend these rules so they can be applied to TS code.

It's also unclear where we should define the TS-specific rules. Can it be done in `eslint-config-ckeditor5`? Or should we introduce a new package?

Reinmar:
So, there seems to be three steps:

* Agree (team-wide) on TS-specific rules to be enabled.
* Write them down in the configs and ensure that our setup (`yarn lint`, precommit hooks, etc) execute them.
* Update the existing TS code (let's assume this is only ckeditor5-utils at this stage) to adhere to these new rules.

Reinmar:
Potential risk: We defined our ESLint rules ~5 years ago. It may happen that we use some deprecated rules that will not work with TS. Thus, we may need to review more of our rules, than just the ones related to TS.

arkflpc:
I'm starting the discussion here, and this is my proposal.

Let's start with all the [recommended rules](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin#supported-rules), but:

| Rule | My recommendation | Comment |
| --- | --- | --- |
| [ban-types](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/ban-types.md) | Don't ban `Function` | I used the `Function` type in a few places in #11755. Although, I don't insist on keeping it there. |
| [no-explicit-any](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-explicit-any.md) | Off | It's not going to work in our code-base (good for projects written from scratch in TypeScript). |
[...]

PR #12491, opened 2022-09-20
Ck/12467 insert content does not insert markers

### LINKED ISSUE #12467: model.insertContent() does not insert markers (added in a document fragment)

When `model.DocumentFragment` is inserted to the model using `model.insertContent()`, its children are inserted but markers saved in `model.DocumentFragment#markers` are skipped. They should be inserted as well.

Most probably here: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/model/utils/insertcontent.ts#L90

The difficulty in this task is that we need to re-calculate the markers ranges. Additional difficulty is introduced by the fact that `model.insertContent()` may heavily affect the model, so re-calculation would be different depending on how some elements are split, merged, etc. I don't think that just "using math" will be enough.

One idea to simplify this calculations that I have is to add temporary, fake elements to the document fragment at the positions were marker boundaries are. Then, those elements would be inserted to the model together with regular children. After this is done, fake elements positions would be saved, the elements would be removed and markers would be added instead. I am not sure this method would be 100% accurate in all cases (e.g. when markers are at the beginning/end of the document fragment, or when the insertion causes massive splitting and merging in the document model).

The other idea is to handle markers in `insertion.handleNodes()` (maybe change the method name). Markers could be passed as second parameter and then processed in `Insertion` class as the nodes are inserted to the model. This way we could very precisely calculate new positions for markers boundaries.

I think the second idea would be more precise but it is probably tougher to implement. We can do it in two steps, first use fake nodes and then implement better solution if needed.

Another thing is that I wonder whether we should put it behind some kind of a flag for backward compatibility but I am leaning towards "no". At the moment, the markers are "silently dropped", so, yeah, we will introduce backward incompatibility. But will it be problematic?

1. In most cases you don't have any markers in the document fragment (or you don't even insert a document fragment). We noticed this problem only now, after long time and it was probably never raised by our users.
2. In cut/copy+paste, markers are removed earlier (probably during copy) and they are not available inside `insertContent()`. So this won't be a problem.
3. Internally, we use `insertContent()` in a couple of places. We can check if there's a possibility that in those places document fragment with marker might be inserted.
4. If the user has problems, the fix is very easy (just remove markers from document fragment before inserting it).
5. The fact that markers are skipped can be treated as a bug, actually. We can treat this as fixing incorrect behavior of `insertContent()`. If your document fragment has markers and you don't want them, remove them.

So, I propose no flag and marking this as minor breaking change + adding simple migration guide.

DawidKossowski:
### **Suggested merge commit message (**[**convention**](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html)**)**

Other (engine): `Model#insertContent()` will now insert markers added to `DocumentFragment#markers`. Closes #12467.

---

### **Additional information**

This improvement has been implemented to allow Import from Word plugin to work correctly.

### DIFF
--- a/packages/ckeditor5-engine/src/model/utils/insertcontent.ts
+++ b/packages/ckeditor5-engine/src/model/utils/insertcontent.ts
@@ -22,6 +22,7 @@ import type Schema from '../schema';
 import type Writer from '../writer';
 
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
+import { LiveRange } from '../../index';
 
 /**
  * Inserts content into the editor (specified selection) as one would expect the paste functionality to work.
@@ -77,10 +78,61 @@ export default function insertContent(
 		}
 
 		const insertion = new Insertion( model, writer, selection.anchor! );
+		const fakeMarkerElements = [];
 
 		let nodesToInsert: any;
 
 		if ( content.is( 'documentFragment' ) ) {
+			// If document fragment has any markers, these markers should be inserted into the model as well.
+			if ( content.markers.size ) {
+				const markersPosition = [];
+
+				for ( const [ name, range ] of content.markers ) {
+					const { start, end } = range;
+					const isCollapsed = start.isEqual( end );
+
+					markersPosition.push(
+						{ position: start, name, isCollapsed },
+						{ position: end, name, isCollapsed }
+					);
+				}
+
+				// Markers position is sorted backwards to ensure that the insertion of fake markers will not change
+				// the position of the next markers.
+				markersPosition.sort( ( { position: posA }, { position: posB } ) => posA.isBefore( posB ) ? 1 : -1 );
+
+				for ( const { position, name, isCollapsed } of markersPosition ) {
+					let fakeElement = null;
+					let collapsed = null;
+					const isAtBeginning = position.parent === content && position.isAtStart;
+					const isAtEnd = position.parent === content && position.isAtEnd;
+
+					// We have two ways of handling markers. In general, we want to add temporary <$marker> model elements to
+					// represent marker boundaries. These elements will be inserted into content together with the rest
+					// of the document fragment. After insertion is done, positions for these elements will be read
+					// and proper, actual markers will be created in the model and fake elements will be removed.
+					//
+					// However, if the <$marker> element is at the beginning or at the end of the document fragment,
+					// it may affect how the inserted content is merged with current model, impacting the insertion
+					// result. To avoid that, we don't add <$marker> elements at these positions. Instead, we will use
+					// `Insertion#getAffectedRange()` to figure out new positions for these marker boundaries.
+					if ( !isAtBeginning && !isAtEnd ) {
+						fakeElement = writer.createElement( '$marker' );
+						writer.insert( fakeElement, position );
+					} else if ( isCollapsed ) {
+						// Save whether the collapsed marker was at the beginning or at the end of document fragment
+						// to know where to create it after the insertion is done.
+						collapsed = isAtBeginning ? 'start' as const : 'end' as const;
+					}
+
+					fakeMarkerElements.push( {
+						name,
+						element: fakeElement,
+						collapsed
+					} );
+				}
+			}
+
 			nodesToInsert = content.getChildren();
 		} else {
 			nodesToInsert = [ content ];
@@ -88,7 +140,77 @@ export default function insertContent(
 
 		insertion.handleNodes( nodesToInsert );
 
-		const newRange = insertion.getSelectionRange();
+		let newRange = insertion.getSelectionRange();
+
+		if ( content.is( 'documentFragment' ) && fakeMarkerElements.length ) {
+			// After insertion was done, the selection was set but the model contains fake <$marker> elements.
+			// These <$marker> elements will be now removed. Because of that, we will need to fix the selection.
+			// We will create a live range that will automatically be update as <$marker> elements are removed.
+			const selectionLiveRange = newRange ? LiveRange.fromRange( newRange ) : null;
+
+			// Marker name -> [ start position, end position ].
+			const markersData: Record<string, Position[]> = {};
+
+			// Note: `fakeMarkerElements` are sorted backwards. However, now, we want to handle the markers
+			// from the beginning, so that existing <$marker> elements do not affect markers positions.
+			// This is why we iterate from the end to the start.
+			for ( let i = fakeMarkerElements.length - 1; i >= 0; i-- ) {
+				const { name, element, collapsed } = fakeMarkerElements[ i ];
+				const isStartBoundary = !markersData[ name ];
+
+				if ( isStartBoundary ) {
+					markersData[ name ] = [];
+				}
+
+				if ( element ) {
+					// Read fake marker element position to learn where the marker should be created.
+					const elementPosition = writer.createPositionAt( element, 'before' );
+
+					markersData[ name ].push( elementPosition );
+
+					writer.remove( element );
+				} else {
+					// If the fake marker element does not exist, it means that the marker boundary was at the beginning or at the end.
+					const rangeOnInsertion = insertion.getAffectedRange();
+
+					if ( !rangeOnInsertion ) {
+						// If affected range is `null` it means that nothing was in the document fragment or all content was filtered out.
+						// Some markers that were in the filtered content may be removed (partially or totally).
+						// Let's handle only those markers that were at the beginning or at the end of the document fragment.
+						if ( collapsed ) {
+							markersData[ name ].push( insertion.position );
+						}
+
+						continue;
+					}
+
+					if ( collapsed ) {
+						// If the marker was collapsed at the beginning or at the end of the document fragment,
+						// put both boundaries at the beginning or at the end of inserted range (to keep the marker collapsed).
+						markersData[ name ].push( rangeOnInsertion[ collapsed ] );
+					} else {
+						markersData[ name ].push( isStartBoundary ? rangeOnInsertion.start : rangeOnInsertion.end );
+					}
+				}
+			}
+
+			for ( const [ name, [ start, end ] ] of Object.entries( markersData ) ) {
+				// For now, we ignore markers if they are included in the filtered-out content.
+				// In the future implementation we will improve that case to create markers that are not filtered out completely.
+				if ( start && end && start.root === end.root ) {
+					writer.addMarker( name, {
+						usingOperation: true,
+						affectsData: true,
+						range: new Range( start, end )
+					} );
+				}
+			}
+
+			if ( selectionLiveRange ) {
+				newRange = selectionLiveRange.toRange();
+				selectionLiveRange.detach();
+			}
+		}
 
 		/* istanbul ignore else */
 		if ( newRange ) {
@@ -128,7 +250,7 @@ class Insertion {
 	private _documentFragmentPosition: Position;
 	private _firstNode: Node | null;
 	private _lastNode: Node | null;
-	private _lastAutoParagraph: Node | null;
+	private _lastAutoParagraph: Element | null;
 	private _filterAttributesOf: Node[];
 	private _affectedStart: LivePosition | null;
 	private _affectedEnd: LivePosition | null;
--- a/packages/ckeditor5-engine/tests/model/utils/insertcontent.js
+++ b/packages/ckeditor5-engine/tests/model/utils/insertcontent.js
@@ -12,6 +12,9 @@ import Position from '../../../src/model/position';
 
 import { setData, getData, parse, stringify } from '../../../src/dev-utils/model';
 import Range from '../../../src/model/range';
+import testUtils from '@ckeditor/ckeditor5-core/tests/_utils/utils';
+
+/* global console */
 
 describe( 'DataController utils', () => {
 	let model, doc, root;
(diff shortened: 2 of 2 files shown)


packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -306,7 +358,7 @@ class Insertion {
 			return Range._createOn( this._nodeToSelect );
 		}
 
-		return this.model.schema.getNearestSelectionRange( this.position );
+		return this.model.schema.getNearestSelectionRange( this._affectedEnd ? this._affectedEnd.toPosition() : this.position );
```
scofalik, line 361: I wouldn't make the change here.

First, although it seems logical in most cases, I am not 100% sure that `this._affectedEnd` is always "same enough" as `this.position`. It seems it should be but it is not apparent from the code. Actually, the logic behind setting `this._affectedEnd` seems complicated. When `this._affectedEnd` is set, it is always set to `this.position` but then, `this.position` may change differently than `this._affectedEnd`.

Second, `getSelectionRange()` makes sense only straight after handling nodes insertion. Most probably it would even make more sense to return the selection range in `handleNodes()` rather than have a separate method. What happens in your code is this: you insert nodes, then you tinker with the inserted content outside of `Insertion` class and then you expect `Insertion` class to return correct results. It kind-of worked for you, because `this._affectedEnd` is a `LiveRange`, so it got updated after you removed the nodes.

But it is not a good design, as you make change in `Insertion` class that is triggered by a new functionality added in a different place. Unless it is a trivial change, an unavoidable change, or a bug we should solve it in a different place. It seems that results returned by `getSelectionRange()` make sense and there were no bugs related strictly to how `Insertion` class works.

Instead, you should reorganize the code, so that the new functionality (handling markers) solves the problem that it introduces. After all, the selection is wrong because we inserted and removed the fake elements:

1.  `handleNodes()`
2.  `let newRange = getSelectionRange()`
3.  if there are fake markers elements to handle:
    1.  create a new live range, same as the range returned by `getSelectionRange()`
    2.  remove fake elements, create markers
    3.  overwrite `newRange` with the live range (that will be updated in step 3.ii)

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,358 @@ describe( 'DataController utils', () => {
...
+
+				expect( getData( model ) )
+					.to.equal( '<paragraph>Bar[]foo</paragraph>' );
+
+				expect( expectedMarker ).to.exist;
+				expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 0, 0 ] );
+				expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 0, 2 ] );
+			} );
+
+			it( 'should create collapsed marker after inserting paragraph at the beginning', () => {
```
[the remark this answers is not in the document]
scofalik: I am also missing a test where the collapsed marker is before/after paragraph, not only inside it.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,358 @@ describe( 'DataController utils', () => {
...
+				const expectedMarker = model.markers.get( 'comment-thread-1' );
+
+				expect( getData( model ) )
+					.to.equal( '<paragraph>foo</paragraph><paragraph>test<imageInline></imageInline>[]</paragraph>' );
+
+				expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 1, 4 ] );
+				expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 1, 5 ] );
+			} );
+
+			it( 'should create marker on imageInline after inserting when content is empty', () => {
```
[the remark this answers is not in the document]
scofalik: And block image into empty content.

scofalik (changes requested):
Requested changes as in comments.

Regarding `getSelectionRange()` - similar problem may happen with `affectedRange` which is returned by `insertContent()`. In the test, you do not test this value (it is returned by `insertHelper()`). I am almost positive that `affectedRange` will be correct though, since it is made of live positions. But it would be nice to test it and be sure.

Regarding the unit tests. I am almost happy with the amount of test you prepared _for this category._ You prepared a lot of unit tests. I'd like to see a few more, as commented, because such core functionality must be heavily tested. But, in general, you considered a lot of different cases.

But, those tests are only from one category. All these tests are inserting inline content (text, inline image) into block-non-object element (paragraph). I don't feel like there are a lot of tests where marker is on whole `<paragraph>`, there are only few. I am also missing a simple test where we insert `[<paragraph>Bar</paragraph>]`. Other than that we need full suite of tests that will have markers on/around block-object element (block image). Then we need tests that insert paragraphs/text/inline images/block images:

*   insert next to paragraph
*   insert next to block image
*   insert next to/into paragraph which is in a block quote
*   insert into a limit element
*   insert at a place where insertion is forbidden (see "filtering out" suite)

We probably don't need to duplicate all the different scenarios that you already came up with for these configurations. But for each configuration we need at least a few tests. Finally, we need a test that inserts empty document fragment with a (collapsed) marker.

So, I see a potential for at least 30-40 more unit tests. I know it is a chore and writing tests requires a lot of creativity. But this is a core functionality. Please, try to look at the tests that are already existing for different configurations of pasted content / receiving content.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,358 @@ describe( 'DataController utils', () => {
...
+				const expectedNeastedMarker = model.markers.get( 'comment-thread-2' );
+
+				expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 0, 3 ] );
+				expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 0, 5 ] );
+
+				expect( expectedNeastedMarker.getRange().start.path ).to.deep.equal( [ 0, 4 ] );
+				expect( expectedNeastedMarker.getRange().end.path ).to.deep.equal( [ 0, 6 ] );
+			} );
+
+			it( 'should create inline marker after inserting paragraph', () => {
```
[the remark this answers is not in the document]
scofalik: The same probably will happen if you test pasting `[<blockImage></blockImage>]` next to, or into `<paragraph>`.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,358 @@ describe( 'DataController utils', () => {
...
+
+				expect( getData( model ) )
+					.to.equal( '<paragraph>Foo<imageInline></imageInline></paragraph><paragraph>>bar[]</paragraph>' );
+
+				expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 0, 3 ] );
+				expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 1, 2 ] );
+			} );
+
+			it( 'should create 500 markers', () => {
+				setData( model, '<paragraph>foo[]</paragraph>' );
```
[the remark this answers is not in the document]
DawidKossowski: I wanted to check if nothing crashed when the number of markers is large

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,358 @@ describe( 'DataController utils', () => {
...
+			} );
+
+			it( 'should create marker which starts in paragraph and ends inside the beginning of next paragraph', () => {
+				setData( model, '<paragraph>Foo[]</paragraph>' );
+
+				insertHelper( '<paragraph>>bar</paragraph><paragraph></paragraph>', {
+					'comment-thread-1': { start: [ 0, 1 ], end: [ 1, 0 ] }
+				} );
+
+				const expectedMarker = model.markers.get( 'comment-thread-1' );
```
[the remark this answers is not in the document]
DawidKossowski: My intention wasn't to check getData output and only check markers positions, because I created similar tests where I checked getData. Anyways, I will update it.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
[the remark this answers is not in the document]
DawidKossowski: That's true. I also added test case with `blockImage`

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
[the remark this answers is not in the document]
DawidKossowski: I created multiple combinations with collapsed markers

DawidKossowski:
@scofalik all done (I hope so...). You can review it again.

DawidKossowski:
I forgot to mention that the issue with merging `$marker` element is still existing. We need to figure out how to fix it.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,1251 @@ describe( 'DataController utils', () => {
...
+				insertHelper( '<imageBlock></imageBlock>', {
+					'marker-a': { start: [ 0 ], end: [ 1 ] }
+				} );
+
+				const expectedMarker = model.markers.get( 'marker-a' );
+
+				// <paragraph><imageInline></imageInline></paragraph>[<imageBlock></imageBlock>]
+				expect( getData( model ) ).to.equal( '<paragraph><imageInline></imageInline></paragraph>[<imageBlock></imageBlock>]' );
+
+				// <paragraph><imageInline></imageInline>{</paragraph><imageBlock></imageBlock>}
```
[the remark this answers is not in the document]
scofalik: Test cases are good but most/all expects will need to be changed anyway after the solution is fixed.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,1251 @@ describe( 'DataController utils', () => {
...
+
+					// <paragraph>{}<imageInline></imageInline>()</paragraph><imageBlock></imageBlock>
+					expect( expectedMarkerA.getRange().start.path ).to.deep.equal( [ 0, 0 ] );
+					expect( expectedMarkerA.getRange().end.path ).to.deep.equal( [ 0, 0 ] );
+					expect( expectedMarkerB.getRange().start.path ).to.deep.equal( [ 0, 1 ] );
+					expect( expectedMarkerB.getRange().end.path ).to.deep.equal( [ 0, 1 ] );
+				} );
+			} );
+
+			it( 'should does not create marker when insertion is forbidden', () => {
```
scofalik, line 2424:
```suggestion
			it( 'should not create marker when insertion is forbidden', () => {
```

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,1251 @@ describe( 'DataController utils', () => {
...
+				setData( model, '<wrapper><limit><paragraph>[]</paragraph></limit></wrapper>', {
```
scofalik, line 2425: Umm wait wait wait. Markers declaring should be in `insertHelper` not here :).

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,1251 @@ describe( 'DataController utils', () => {
...
+				}
+
+				insertHelper( '<paragraph>Bar</paragraph>', mockMarkers );
+
+				for ( let i = 0; i < 500; i++ ) {
+					expect( model.markers.get( `comment-thread-${ i }` ) ).to.exist;
+				}
+			} );
+
+			describe( 'affected range', () => {
```
scofalik, line 2455: I thought you forgot about testing `affectedRange` :). I thought you will add these checks in current tests.

scofalik (changes requested):
The testing suite is much better now.

Four things:

1.  Make sure that in tests like `setData( '[]<paragraph>Foo</paragraph>' )` the content is actually inserted at `[ 0 ]`, not `[ 0, 0 ]`.
2.  Fix expects.
3.  We need to fix problems when `$marker` is at the beginning or end of the document fragment. We already discussed that on Slack.
4.  I am missing a couple tests more:
    1.  Test for inserting `<image></image>[]<image></image>` - I expect that `$marker` elements in such inserted content may behave interestingly. Similar case for `<paragraph>`s.
    2.  I am not sure, maybe I missed it, do we have a test where we insert empty document fragment, just with marker(s)?
    3.  A few tests where marker is wholly or partially on filtered-out element, e.g. `{<limit><p>Foo</p></limit><p>B}ar</p>` added into another limit (I think this should be filtered).

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,6 +1299,1251 @@ describe( 'DataController utils', () => {
...
+			it( 'should create marker after inserting paragraph', () => {
+				setData( model, '<paragraph>foo[]</paragraph>' );
+
+				insertHelper( '<paragraph>Bar</paragraph>', {
+					'marker-a': { start: [ 0, 0 ], end: [ 0, 3 ] }
+				} );
+
+				const expectedMarker = model.markers.get( 'marker-a' );
+
+				// <paragraph>fooBar[]</paragraph>
```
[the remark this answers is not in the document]
scofalik: Or maybe for better readability:

```
// <paragraph>foo[]</paragraph>
// <paragraph>{Bar}</paragraph>
//
// <paragraph>foo{Bar}</paragraph>
```

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -88,7 +116,35 @@ export default function insertContent(
...
+
+			for ( const [ name, [ start, end ] ] of Object.entries( markersData ) ) {
+				writer.addMarker( name, {
+					usingOperation: true,
+					affectsData: true,
+					range: new Range( start, end )
+				} );
+			}
+
+			newRange = markerRange && markerRange.toRange();
```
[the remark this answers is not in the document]
scofalik: Also `markerRange` doesn't seem like a correct name. `selectionLiveRange` is better. It also has `liveRange` in it, so when you see it you instantly know that you need to detach live range :D.

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -88,7 +116,35 @@ export default function insertContent(
 
 		insertion.handleNodes( nodesToInsert );
 
-		const newRange = insertion.getSelectionRange();
+		let newRange = insertion.getSelectionRange();
+
+		if ( fakeMarkerElements.length ) {
+			const markerRange = newRange && LiveRange.fromRange( newRange );
```
scofalik, line 122: For readability reasons, I'd prefer `= newRange ? LiveRange.fromRange( newRange ) : null`.

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -85,21 +85,28 @@ export default function insertContent(
...
 				} );
 			}
 
 			markersPosition.sort( ( { position: posA }, { position: posB } ) => posA.isBefore( posB ) ? 1 : -1 );
 
 			for ( const { position, name } of markersPosition ) {
-				const fakeElement = writer.createElement( '$marker', { 'data-name': name } );
+				let fakeElement = null;
+				const isBeginning = position.path.length === 1 && position.path[ 0 ] === 0;
+				const isEnding = position.path.length === 1 && position.path[ 0 ] === content.childCount;
```
scofalik, line 104: `content.maxOffset`. But you should also be able to use `Position#isAtEnd()` 🤔.

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -449,6 +490,8 @@ class Insertion {
 
 		const livePosition = LivePosition.fromPosition( this.position, 'toNext' );
 
+		this._positionBeforeFirstInsert = livePosition.toPosition();
```
scofalik, line 493: This has to be set only if it hasn't been set before.

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
scofalik: I think I'd go with `position.parent === content && position.isAtStart` and `position.isAtEnd`. Will be more clear what is going on here.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
DawidKossowski: What do you mean? Markers are sent in the second parameter

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
DawidKossowski: Can we stay with these tests?

DawidKossowski:
@scofalik it's ready to review. However, there are still issues that we discussed and will be fixed in the following tasks:
1. We use `affectedRange` to calculate starting and ending markers' positions. It causes that sometimes markers are created in the wrong place (like inserting imageBlock inside paragraph):
```
// <paragraph>Fo[]o</paragraph>
// {<imageBlock></imageBlock>}
//
// <paragraph>Fo{</paragraph><imageBlock></imageBlock><paragraph>}o</paragraph>
```

2. Sometimes, when we set the custom selection the resulting insertion is wrong (it's not connected to markers):
```
setData( model, '<imageBlock></imageBlock>[]<imageBlock></imageBlock>' );

insertHelper( '<imageBlock></imageBlock>', null, [ 1 ] );

expect( getData( model ) ).to.equal( '[<imageBlock></imageBlock>]<imageBlock></imageBlock><imageBlock></imageBlock>' );
```

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+
+				expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 1 ] );
+				expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 2 ] );
+			} );
+
+			it( 'should create marker on imageBlock and imageInline after inserting', () => {
+				// <paragraph>[]</paragraph>
+				// {<imageBlock></imageBlock><imageInline></imageInline>}
+				//
+				// {<imageBlock></imageBlock><paragraph><imageInline></imageInline></paragraph>}
```
scofalik, line 2133: This is an interesting, although rather not probable case. Interesting as it is one big marker and the inline image will need to be paragraphed, so the marker has to be on `<paragraph>` but it does not have to be on `</paragraph>`.

I think this will get also fixed when we will change from basing on `affectedRange` to some more precise solution.

scofalik: For the moment, I don't care much about this scenario as I think it will change. You can leave it like this (and it should start failing after we improve our solution).

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+
+				expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 0 ] );
+				expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 2 ] );
+			} );
+
+			it( 'should create marker on imageBlock after inserting in block quoted paragraph', () => {
+				// <blockQuote><paragraph>f[]oo</paragraph></blockQuote>
+				// {<imageBlock></imageBlock>}
+				//
+				// <blockQuote><paragraph>f{</paragraph><imageBlock></imageBlock>}<paragraph>oo</paragraph></blockQuote>
```
scofalik, line 2152: Change expect + off.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+
+					expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 0, 3 ] );
+					expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 0, 3 ] );
+				} );
+
+				it( 'should create collapsed marker after inserting imageInline next to paragraph', () => {
+					// <paragraph>foo</paragraph>[]
+					// {}<imageInline></imageInline>
+					//
+					// <paragraph>foo[]</paragraph>{<paragraph><imageInline></imageInline></paragraph>}
```
scofalik, line 2426: It should either be before `<paragraph>` or before `<imageInline>` but not like this. AFAICS, the expect is correct but the comment is wrong.

scofalik: The question here is what behavior we see as correct regarding the positioning of the collapsed marker, i.e., should it be at the beginning of inserted content, or should it stay next to element that it was in the original content.

In other words, what is correct: `{}<paragraph><imageInline>` or `<paragraph>{}<imageInline>`. I believe the second one and we can see what will happen to this test after we improve the solution.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+
+					expect( getData( model ) ).to.equal( '<paragraph>f</paragraph>[<imageBlock></imageBlock>]<paragraph>oo</paragraph>' );
+
+					expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 0, 1 ] );
+					expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 0, 1 ] );
+				} );
+
+				it( 'should create collapsed marker after inserting imageBlock at the end of paragraph', () => {
+					// <paragraph>foo[]</paragraph>
+					// {}<imageBlock></imageBlock>
```
scofalik, line 2500: Interesting that in this case it was moved out of the paragraph, but in previous case it stayed in the paragraph.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+
+					const expectedMarker = model.markers.get( 'marker-a' );
+
+					expect( getData( model ) ).to.equal( '<paragraph>foo[]</paragraph><imageBlock></imageBlock>' );
+
+					expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 1 ] );
+					expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 1 ] );
+				} );
+
+				it( 'should create collapsed marker after inserting imageBlock before paragraph', () => {
```
scofalik, line 2536: In general, I miss some tests where `{}` is at the end of pasted content. Like this test, but `{}` after `</imageBlock>`.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+					expect( getData( model ) ).to.equal( '<paragraph><imageInline></imageInline></paragraph>[<imageBlock></imageBlock>]' );
+
+					expect( expectedMarkerA.getRange().start.path ).to.deep.equal( [ 0, 0 ] );
+					expect( expectedMarkerA.getRange().end.path ).to.deep.equal( [ 0, 0 ] );
+					expect( expectedMarkerB.getRange().start.path ).to.deep.equal( [ 0, 1 ] );
+					expect( expectedMarkerB.getRange().end.path ).to.deep.equal( [ 0, 1 ] );
+				} );
+			} );
+
+			it( 'should not create marker after inserting empty document fragment', () => {
```
scofalik, line 2579: Why not? IMO it should.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+				const expectedMarker = model.markers.get( 'marker-a' );
+
+				expect( getData( model ) ).to.equal( '<paragraph>fo[]o</paragraph>' );
+
+				expect( expectedMarker ).to.be.null;
+			} );
+
+			it( 'should not create marker when insertion is forbidden', () => {
+				// <wrapper><limit><paragraph>[]</paragraph></limit></wrapper>
+				// {<wrapper><limit><paragraph>foo</paragraph></limit></wrapper>}
```
scofalik, line 2599: Here is a more interesting example but, in the end, similar to the one before. In this case it is more discussable whether the marker should be added or not.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+
+				const expectedMarker = model.markers.get( 'marker-a' );
+
+				expect( getData( model ) ).to.equal( '<wrapper><limit><paragraph>Bar[]</paragraph></limit></wrapper>' );
+
+				expect( expectedMarker.getRange().start.path ).to.deep.equal( [ 0, 0, 0 ] );
+				expect( expectedMarker.getRange().end.path ).to.deep.equal( [ 0, 0, 0, 2 ] );
+			} );
+
+			it( 'should not create marker that is next to forbidden content', () => {
```
scofalik, line 2635: Test name looks incorrect given that the marker is added.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
scofalik: My problem is that these test cover much less scenarios than the rest of the test you added. And it will be easier to add `affectedRange` checks to the tests you wrote above, than to recreated those scenarios here.

Our current solution looks like it should not change `affectedRange`. I would be fine with adding tests for `affectedRange` in a followup issue. In this case, we can leave this tests suite, add an issue and when solving that new issue, we will remove this suite and make `affectedRange` checks in other markers-related tests.

BTW. It may be easier for you to check `affectedRange` if you use `stringify()` (from the same file as `getData()`). You can pass there a range that should be printed. It would be also easier for checking markers. I should have pointed this out earlier :(.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+					} );
+
+					expect( affectedRange.start.path ).to.deep.equal( [ 0 ] );
+					expect( affectedRange.end.path ).to.deep.equal( [ 1 ] );
+				} );
+			} );
+		} );
+	} );
+
+	describe( 'integration with limit elements', () => {
```
scofalik, line 2788: Just to confirm: you didn't change anything here but the GH diff got it incorrectly?

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
```diff
@@ -1299,128 +1302,1629 @@ describe( 'DataController utils', () => {
...
+			doc.selection;
+
+		const markersMap = new Map();
+
+		if ( typeof content == 'string' ) {
+			content = parse( content, model.schema, {
+				context: [ '$clipboardHolder' ]
+			} );
+
+			if ( markers && !content.markers ) {
```
scofalik, line 2913:
```suggestion
			if ( markers && !content.is( 'documentFragment' ) ) {
```

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -77,9 +78,50 @@ export default function insertContent(
 		}
 
 		const insertion = new Insertion( model, writer, selection.anchor! );
+		const markersPosition = [];
+		const fakeMarkerElements = [];
 
 		let nodesToInsert: any;
 
+		if ( content.is( 'documentFragment' ) && content.markers ) {
```
scofalik, line 86: We could merge this `if` with the one below:

```javascript
if ( content.is( 'documentFragment' ) ) {
	nodesToInsert = content.getChildren();
	
	if ( content.markers ) {
		...
	}
}
```

scofalik (changes requested):

scofalik:
Regarding your previous post:

1.  Please create an issue to improve current solution. Title: "`insertContent()` should not depend on affected range when inserting markers".
2.  I think this happens because in `setData()` (or rather: just after) the selection is autofixed to `[<imageBlock></imageBlock>]<imageBlock></imageBlock>`. After you insert between the paragraphs, it does not affect document selection in any way. Since you passed the selection by hand to `insertContent()`, `insertContent()` does not change the document selection either. So, everything looks good here, after all.

scofalik:
What about a case where marker starts in filtered content but ends in non-filtered content (and vice versa)?

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -88,7 +130,52 @@ export default function insertContent(
...
+					if ( markersData[ name ] ) {
+						markersData[ name ].push( elementPosition );
+					} else {
+						markersData[ name ] = [ elementPosition ];
+					}
+
+					writer.remove( element );
+				} else {
+					if ( markersData[ name ] ) {
+						markersData[ name ].push( collapsed ? rangeBetweenInsertion[ collapsed ] : rangeBetweenInsertion.end );
```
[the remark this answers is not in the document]
DawidKossowski: `collapsed` variable can have 3 possible values:  `false`, `'start'`, `'end'`. Based on those values I determine the position of collapsed markers

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
DawidKossowski: That's true. I didn't change those tests. Screenshot:

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
DawidKossowski: It's since in the previous example the insertion of `imageBlock` split the original paragraph into two and `affectedRange` starts from the beginning of the first split paragraph.  
  
In that case the `imageBlock` element is inserted just after the paragraph even if the selection is inside.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
DawidKossowski: I updated this test and the answer is the first one: `{}<paragraph><imageInline>`

DawidKossowski:
@scofalik you can take a look at it again.

packages/ckeditor5-engine/tests/model/utils/insertcontent.js
scofalik: :+1:, I expected this to end up as `{}<paragraph>`, what I meant by "correct" is what we want to achieve / what is the user's expected behavior.

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -77,18 +78,111 @@ export default function insertContent(
...
+				markersPosition.sort( ( { position: posA }, { position: posB } ) => posA.isBefore( posB ) ? 1 : -1 );
+
+				for ( const { position, name, isCollapsed } of markersPosition ) {
+					let fakeElement = null;
+					let collapsed = null;
+					const isBeginning = position.parent === content && position.isAtStart;
+					const isEnding = position.parent === content && position.isAtEnd;
+
+					if ( !isBeginning && !isEnding ) {
+						fakeElement = writer.createElement( '$marker', { 'data-name': name } );
```
scofalik, line 112: Since this is not an HTML element, you can simply use "name" as the attribute ... name. Nothing bad should happen.

scofalik: BTW. it seems that you don't even read the name stored in this attribute :D.

scofalik (changes requested):
I added one more comment and I have one more request.

This new code in `insertContent()` handles some complicated logic. There are some tricky cases that we cover. I would expect that this new code has several comments with explanations why we are doing this or that, what will happen next, what is the whole idea of handling the markers.

Other than that, LGTM.

We have two issues that we need to create tickets for: using `affectedRange` and markers on filtered content.

scofalik:
@arkflpc Could you look at it from TS perspective?

packages/ckeditor5-engine/src/model/utils/insertcontent.ts
```diff
@@ -77,18 +78,111 @@ export default function insertContent(
...
 		}
 
 		insertion.handleNodes( nodesToInsert );
 
-		const newRange = insertion.getSelectionRange();
+		let newRange = insertion.getSelectionRange();
+
+		if ( content.is( 'documentFragment' ) && fakeMarkerElements.length ) {
+			const selectionLiveRange = newRange ? LiveRange.fromRange( newRange ) : null;
+			const markersData: Record<string, Array<Position>> = {};
```
arkflpc, line 137: I'd rather use `Position[]` instead of `Array<Position>`. It would be consistent with the rest of `engine`. We decided not to force any convention (see [comment](https://github.com/ckeditor/ckeditor5/issues/11719#issuecomment-1148731592)), so it is up to you.

Issue #12496, opened 2022-09-21
Icon registry


oleq:
## 📝 Provide a description of the new feature

It would be nice if there was a registry of all available icons in the editor instance (build? constructor?) in the map `<name,svg string>` format. 

Use-cases:

*   In #12490 developers could use just any icon for their declarative drop-down (sub-toolbar), not just a few icons that will be shipped with the feature. 
*   Integrators who develop their plugins as functions injected in editor config (without using webpack) would be able to re-use existing icons in their UI.

This is a twin feature to #11749 (UI component registry).

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

PR #12506, opened 2022-09-23
Declarative nested toolbars

### LINKED ISSUE #12490: Declarative nested toolbars

Let's add a new format of configuration to `config.toolbar` to make it possible to group toolbar items into drop-downs so that we can squeeze a massive toolbar

into something compact

Screenshot 2022-09-14 at 11 50 41
Screenshot 2022-09-14 at 11 50 46
Screenshot 2022-09-14 at 11 50 51
image

## Proposed scope of changes

We're going to need a functionality in the `ToolbarView` component to allows declarative drop-downs with nested toolbars. 

Example toolbar configuration

```javascript
toolbar: [
	{
		label: 'Import/Export',
		icon: 'importExport',
		items: [
			'exportPdf',
			'exportWord'
		]
	},
	'sourceEditing',
	'|',
	'heading',
	'|',
	'style',
	'|',
	'bold', 'italic', 'link',
	'|',
	{
		label: 'Insert',
		icon: 'plus',
		items: [
			'insertImage', 'ckbox', 'insertTable', 'mediaEmbed', 'htmlEmbed',
			'pageBreak', 'horizontalLine', 'specialCharacters', '|',
			'blockQuote', 'codeBlock'
		]
	},
	
	// ...
```

**Notes**:

*  `label` can be just any string (we could provide translations for several popular ones like we do in the heading feature),
*  `icon` maps to an object with pre-defined icons (or a SVG string like in `ImageStyle` feature)
*  `items` corresponds to the standard items configuration syntax. It support recursive drop-downs with toolbars.

### Changes to the source editing feature

To save space, I say we remove the label from the button. Not sure why we decided to have it there in the first place. I suppose we wanted to replicate the look of CKE4.

*   Alternative 1.: Source editing feature could register two components in the factory. One with the label and one without.
*   Alternative 2.: Use **advanced toolbar item config** (see below) to avoid multiple components in the factory.

image

### Changes to the text language feature

To save space, let's remove the label from the button and have a new icon there. This will degrade the UX but OTOH this feature is not very popular.

*   Alternative 1.: Text part language feature could register two components in the factory. One with the label and one without.
*   Alternative 2.: Use **advanced toolbar item config** (see below) to avoid multiple components in the factory.

image

### Other works

* Let's have a new icon for all export features.
* Let's have more icons developers could use for their dropdowns. I think they already exist so we only have to move them to `ckeditor5-core`.

#### Drawback: Deep stacking

The drawback of this solution is that in some cases, there's a deep stack of drop-downs/toolbars to reach an option

Screenshot 2022-09-14 at 11 51 18

#### Drawback: Drop-downs are passive

It could be confusing for some users that these drop-downs (their icons and states) are static:

image

It may look like the icon should change depending on the current selection but it won't. This is just a static icon to give users an idea of what's in there.

### Proposal: More config for the dropdown

The syntax I propose is as follows:

```
{
	label: 'Insert',
	icon: 'plus',
	items: [ ... ]
},
```

but there could be extra fields like `withText` so the dropdown' looks

image

instead of 

image

### Proposal: Advanced atomic toolbar item config

To avoid having multiple components registered by a feature just to have one with a label and the other without it, I think we could have advanced configuration format for toolbar items:

```javascript
toolbar: [
	// ...
	
	{ name: 'sourceEditing', withText: false, otherOption: ... }
	
	// ...
]
```

This should be pretty straightforward. When `ToolbarView` uses component factory to create a component named `sourceEditing`

https://github.com/ckeditor/ckeditor5/blob/e86c4029b1b4ea184f2451f23b24ce603afead55/packages/ckeditor5-ui/src/toolbar/toolbarview.js#L359-L369

it will then call

```js
const component = factory.create( name );

if ( component instanceof DropdownView ) {
	component.set( component.buttonView, options );
} else if ( component instanceof ButtonView ) {
	component.set( component, options );
}
```

to set options passed in the config. Since things like `withText` or `tooltip` or `label` are observable properties of `ButtonView`, this will simply use the public API to customize the component.

**Note**: It will fail if the component has these properties bound to some external state/model but for many cases this should work.

---

oleq:

Feature (ui): Allowed grouping toolbar items into drop-downs by using a declarative config. Closes #12490.

BREAKING CHANGE (basic-styles): The bold icon has been moved to the `ckeditor5-core` package.  
BREAKING CHANGE (paragraph): The paragraph icon has been moved to the `ckeditor5-core` package.

---

### Additional information

**TODOs**:

*   [x] Are we OK that the bold and paragraph icons moved to the `ckeditor5-core` package (e.g. while italic and heading1 didn't)?
*   [x] Docs proofreading.
*   [x] Do we want a default icon? (three vertical dots is a solid candidate)


mlewand:
@oleq regarding moving the icons, why `core` package? Wouldn't UI package be more fitting?

Regarding the default icon, where do you see it kicking in? Would that it's added as an opt-out (so that integrator has to pass `icon: false` to remove the icon)? Or would it be present if integrator didn't provide neither the `icon` nor `withText` property?

oleq:
> @oleq regarding moving the icons, why `core` package? Wouldn't UI package be more fitting?

Because we're already collecting icons there. Alignment icons are there, object alignment icons too. And blockquote. All icons that could be (or already are) re-used by other features.

I don't say this is the best place, though. Honestly I don't know why we have icons scattered across the project. They don't make sense without `ckeditor5-ui` and the `IconView` class anyway. Maybe because sb. else could have their custom UI but use our icons? I don't know (remember), really. I'd rather keep all icons in the project in the `ckeditor5-ui` package, but this would be a huge change. It probably overlaps with #12496 too.

> Regarding the default icon, where do you see it kicking in? Would that it's added as an opt-out (so that integrator has to pass `icon: false` to remove the icon)? Or would it be present if integrator didn't provide neither the `icon` nor `withText` property?

The latter. But this is not critical I guess. There are [warnings logged](https://github.com/ckeditor/ckeditor5/pull/12506/files#diff-352f905722b4f5937b9a4a3c027c0b8da7fe04c7d0365d73f1ce6f470d60cad8R466-R486) if you fail to provide either an icon or `withText: true`, anyway.

packages/ckeditor5-ui/src/toolbar/toolbarview.js
```diff
@@ -405,6 +422,85 @@ export default class ToolbarView extends View {
 			} );
 	}
 
+	/**
+	 * Creates a user-defined dropdown containing a toolbar with items.
+	 *
+	 * @private
+	 * @param {Object} definition A definition of the nested toolbar drop-down.
+	 * @param {String} definition.label A label of the dropdown.
+	 * @param {String} definition.icon An icon of the drop-down. One of 'bold', 'plus', 'text', 'importExport', 'alignLeft', 'paragraph'
```
arkflpc, line 431:
```suggestion
	 * @param {String} [definition.icon] An icon of the drop-down. One of 'bold', 'plus', 'text', 'importExport', 'alignLeft', 'paragraph'
```

arkflpc (approved):

packages/ckeditor5-ui/src/toolbar/toolbarview.js
```diff
@@ -405,6 +422,85 @@ export default class ToolbarView extends View {
...
+			 */
+			logWarning( 'toolbarview-nested-toolbar-drop-down-missing-presentational-config', definition );
+		}
+
+		dropdownView.class = 'ck-toolbar__nested-toolbar-dropdown';
+		dropdownView.buttonView.set( {
+			label,
+			tooltip,
+			withText: !!withText,
+			icon: NESTED_TOOLBAR_ICONS[ icon ] || icon
```
arkflpc, line 494:
```suggestion
			icon: NESTED_TOOLBAR_ICONS[ icon ] || icon || NESTED_TOOLBAR_ICONS.threeVerticalDots
```

arkflpc: Do we need automated tests for that? I think, we do, to have the coverage.

arkflpc: And `if ( !icon && !withText )` is to be removed, I think.

godai78:
I made some changes, as we use "drop-down" (hyphenated) when it's an adjective, and here we mostly had nouns.

PR #12512, opened 2022-09-26
Performance improvements in CKE5

scofalik:

Other (utils): Dropped using lodash `isElement()` to achieve better performance in `Rect` class.
Other (engine): Reimplemented `Position#isTouching()` to achieve better performance.

---

@oleq could you check changes in `Rect`?

@arkflpc could you check this PR from TS point of view?

### DIFF
--- a/packages/ckeditor5-engine/src/model/position.ts
+++ b/packages/ckeditor5-engine/src/model/position.ts
@@ -525,57 +525,47 @@ export default class Position extends TypeCheckable {
 	 * @returns {Boolean} True if positions touch.
 	 */
 	public isTouching( otherPosition: Position ): boolean {
-		let left = null;
-		let right = null;
-		const compare = this.compareWith( otherPosition );
+		if ( this.root !== otherPosition.root ) {
+			return false;
+		}
 
-		switch ( compare ) {
-			case 'same':
-				return true;
+		const commonLevel = Math.min( this.path.length, otherPosition.path.length );
 
-			case 'before':
-				left = Position._createAt( this );
-				right = Position._createAt( otherPosition );
-				break;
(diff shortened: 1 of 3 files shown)


scofalik:
For the `Rect` changes - another option was to change constructor so that all types other than DOM are directly checked and DOM Element path is in `else`. However, DOM Element is the most common path, so we would have a few unnecessary checks before. OTOH, new `isDomElement` makes a few checks itself anyway, so, IDK. I can change this implementation if you have doubts.

arkflpc (approved):

oleq (commented):
I'm OK with changes to the `Rect` module. I didn't run tests but I assume you did :P. 

What I don't like though is that we still use lodash `isElement` in many places. What you did here should probably land in `ckeditor5-utils/dom` as `isElement` and replace lodash imports across the project. Less code, less confusion.

scofalik:
I wasn't sure why we are using complicated lodash method instead of something simpler. This is why I wanted to limit the scope of changes. If you are sure that this will not ruin something, I could make a change in `utils/dom`. But we can also do this in a separate ticket.

oleq:
> I wasn't sure why we are using complicated lodash method instead of something simpler. This is why I wanted to limit the scope of changes. If you are sure that this will not ruin something, I could make a change in `utils/dom`. But we can also do this in a separate ticket.

Let's have a new issue for that then.

BTW, why `nodeType === 1` but not `nodeType === Node.ELEMENT_NODE`?

scofalik:
Honestly, I must have taken it from [loadash `isElement()`](https://github.com/lodash/lodash/blob/master/isElement.js#L20).

oleq (approved):

PR #12572, opened 2022-10-05
Tooltip should hide when an element the tooltip is attached to also hides

### LINKED ISSUE #12492: Tooltips in the link actions view won't go away when hidden using Esc key press

1.  Create a link
2.  Click it
3.  Hover an UI in the link actions view
4.  Press ESC

### ✔️ Expected result

The tooltip is gone.

### ❌ Actual result

The tooltip stays

### ❓ Possible solution

It does not happen in the `LinkFormView`, though. This is probably because one of the handlers (`blur` or `mouseleave`) gets a chance to hide the tooltip in that case. The key to the right fix is figuring out the difference in behavior between `LinkActionsView` and `LinkFormView`.

[https://github.com/ckeditor/ckeditor5/blob/e86c4029b1b4ea184f2451f23b24ce603afead55/packages/ckeditor5-ui/src/tooltipmanager.js#L147-L153](https://github.com/ckeditor/ckeditor5/blob/e86c4029b1b4ea184f2451f23b24ce603afead55/packages/ckeditor5-ui/src/tooltipmanager.js#L147-L153)

Also note that the link UI is probably only one of many examples of this problem. We need a generic solution in `TooltipManager`, not a hack in `ckeditor5-link`.

---

oleq:

Fix (ui): Tooltip should hide when an element the tooltip is attached to also hides. Closes #12492.


oleq:
@ckeditor/qa-team Could you take a look at this one please? Reproduction steps in #12492.

dufipl:
With @Acrophost we've tested that and it looks like working correctly, https://github.com/ckeditor/ckeditor5/issues/12492 is fixed. 👍

mmotyczynska (approved):
LGTM :+1:

PR #12585, opened 2022-10-06
enhance special-characters.md

rbalet:
Type (special-characters.md) : Bolden the text to make it faster to read


godai78 (approved):

godai78:
Thanks for this @rbalet!

PR #12603, opened 2022-10-07
Internal (table, list): Use insertContent in table caption instead of using writer directly

kbinieda:

Internal (table): Use insertContent in table caption instead of using writer directly.

---

### Additional information

This is needed / going to help with track changes integration

### DIFF
--- a/packages/ckeditor5-table/src/tablecaption/toggletablecaptioncommand.js
+++ b/packages/ckeditor5-table/src/tablecaption/toggletablecaptioncommand.js
@@ -89,7 +89,7 @@ export default class ToggleTableCaptionCommand extends Command {
 		// Try restoring the caption from the TableCaptionEditing plugin storage.
 		const newCaptionElement = savedCaptionElement || writer.createElement( 'caption' );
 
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-table/src/tablecaption/toggletablecaptioncommand.js
```diff
@@ -114,7 +114,8 @@ export default class ToggleTableCaptionCommand extends Command {
 		// Store the caption content so it can be restored quickly if the user changes their mind.
 		tableCaptionEditing._saveCaption( tableElement, captionElement );
 
+		writer.setSelection( captionElement, 'on' );
+		model.deleteContent( model.document.selection );
```
scofalik, line 118: `model.deleteContent()` should be able to take `Range` instance as the parameter. Did you try `model.deleteContent( model.createRangeOn( captionElement ) )`? Was it failing for some reason?

scofalik: In this case, have you tried `model.deleteContent( model.createSelection( captionElement, 'on' ) )`?

PR #12631, opened 2022-10-11
Improvements to the IconView to allow rich colorful icons

### LINKED ISSUE #12597: IconView should preserve (some) attributes of the <svg> tag

If you create an icon with 

```
<svg width="45" height="45" viewBox="0 0 45 45" fill="none" xmlns="http://www.w3.org/2000/svg">
```

the `fill="none"` will be ignored because `IconView` creates the `<svg>` element from scratch. Instead, we should preserve these attributes because they affect rendering. Without this attribute the icon can look broken.

### ✔️ Expected result

`fill="none"` stays

### ❌ Actual result

`fill="none"` is gone

### ❓ Possible solution

*   To make it clear, we need to pass only certain attributes from predefined white list.
*   We're concerned only about attributes that are added to the root `svg` element.

The fix should land around [https://github.com/ckeditor/ckeditor5/blob/8254ae2dc7e0cfb23d15dc040a2c94f32ac0d3c5/packages/ckeditor5-ui/src/icon/iconview.js#L98-L112.](https://github.com/ckeditor/ckeditor5/blob/8254ae2dc7e0cfb23d15dc040a2c94f32ac0d3c5/packages/ckeditor5-ui/src/icon/iconview.js#L98-L112.)

we should copy all attributes except `width`, `height` (handled by our viewbox), `viewbox` (already copied), `xmlns` (handled by `createElementNS`).

---

### LINKED ISSUE #12599: IconView should have a property that disables all customizations brought by CKE5

ATM icons in CKE5 inherit colors from buttons to allow dynamic theming. To do that, it is assumed there are no paths with strokes in the `<svg>`. There's also `ck-reset_all` applying to the internals of the icon. 

This prevents users from using their custom rich icons in the `IconView` component in different places. They look totally broken if not adapted to the requirements of CKE5.

Things to do:

*   Add `ck-reset_all-excluded` to the `IconView` so the reset does not apply to the internals of the icon.
*   Add an `isMonochrome = true (default)` to the icon that would translate to a CSS class that adds the current rules (without `cursor` and `will-change`)

https://github.com/ckeditor/ckeditor5/blob/8254ae2dc7e0cfb23d15dc040a2c94f32ac0d3c5/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/icon/icon.css#L17-L36

---

oleq:

```
Fix (ui): Improvements to the `IconView` component to allow rich colorful icons. Closes #12597. Closes #12599.

Preserved presentational attributes from the icon source on the `<svg>` element (#12597). 
Made it possible to opt out from icon color inheritance (#12599). 
Excluded icon internals from CSS reset (#12599). 
```

---

### Additional information

Closes #12597. Closes #12599.


packages/ckeditor5-ui/src/icon/iconview.ts
```diff
@@ -23,6 +23,8 @@ export default class IconView extends View {
 	declare public content: string | undefined;
 	declare public viewBox: string;
 	declare public fillColor: string;
+	declare public isColorInherited: boolean;
+	declare public static presentationalAttributeNames: string[];
```
arkflpc, line 27:
```suggestion
	declare public static presentationalAttributeNames: Array<string>;
```

arkflpc: It will be enforced this way very soon.

packages/ckeditor5-ui/tests/manual/icon/icon.md
```diff
@@ -14,3 +14,8 @@
 
 ## ViewBox
 1. First icon should be large and green, same as the previous one.
+
+## Colorful icons
+
+1. There should seven colorful icon rendered without any distortions.
```
arkflpc, line 20:
```suggestion
1. There should be seven colorful icon rendered without any distortions.
```

arkflpc (changes requested):

PR #12652, opened 2022-10-14
Changed event type naming convention

### LINKED ISSUE #12649: [TS MVP] Change naming convention of event describing types
In TypeScript we describe events using auxiliary types, that can be passed to `Emitter` methods like `listenTo`, `fire`, etc. However, many have the same names. I.e., `'change'` event is fired by: `ObservableMixin`, `Selection`, `Document`, `Range`, `MarkerCollection`, `Collection` and `Position`. All of them described as `ChangeEvent`. If a code consumes two or more `ChangeEvent`s, they can be renamed in `import`. However, DX is poor, finding the proper one is error-prone, and copy-paste code may not respect all local names of particular event. Some events (in observers) even clash with DOM-provided ones.

Let's change the names of all event-describing types to match the name convention: `${ className }${ eventName }Event`.

arkflpc:

Internal: Changes event type names to make them consistent. Closes #12649.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/model/documentselection.ts
+++ b/packages/ckeditor5-engine/src/model/documentselection.ts
@@ -12,15 +12,15 @@
 import TypeCheckable from './typecheckable';
 import LiveRange from './liverange';
 import Selection, {
-	type ChangeAttributeEvent as SelectionChangeAttributeEvent,
-	type ChangeRangeEvent as SelectionChangeRangeEvent
+	type SelectionChangeAttributeEvent,
+	type SelectionChangeRangeEvent
 } from './selection';
 import Text from './text';
 import TextProxy from './textproxy';
 
-import type { default as Document, ChangeEvent as DocumentChangeEvent } from './document';
-import type { default as Model, ApplyOperationEvent } from './model';
-import type { Marker, UpdateEvent as MarkerUpdateEvent } from './markercollection';
+import type { default as Document, DocumentChangeEvent } from './document';
+import type { default as Model, ModelApplyOperationEvent } from './model';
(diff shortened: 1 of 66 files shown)


packages/ckeditor5-ui/src/input/inputview.ts
```diff
@@ -230,7 +230,7 @@ function isInputElementEmpty( domElement: HTMLInputElement ) {
 	return !domElement.value;
 }
 
-export type InputEvent = {
```
niegowski, line 233: Maybe this should be `InputViewInputEvent`?

packages/ckeditor5-engine/src/model/documentselection.ts
```diff
@@ -567,7 +567,7 @@ DocumentSelection.prototype.is = function( type: string ): boolean {
  * which mean that they are not updated once the document changes.
  */
 
-export type ChangeRangeEvent = SelectionChangeRangeEvent;
+export type DocumentSelectionChangeRangeEvent = SelectionChangeRangeEvent;
```
niegowski, line 570: Maybe we should be consistent with the view selection?

packages/ckeditor5-engine/src/model/selection.ts
```diff
@@ -856,22 +856,22 @@ Selection.prototype.is = function( type: string ): boolean {
 	return type === 'selection' || type === 'model:selection';
 };
 
-export type ChangeEvent = {
+export type SelectionChangeEvent = {
```
niegowski, line 859: Maybe we should prefix it as in the view?

PR #12688, opened 2022-10-20
GHS should mark formatting elements

### LINKED ISSUE #12626: GHS should mark formatting elements (so that remove format works with it)

Currently GHS doesn't work well with remove format feature.

That's because this feature [determines whether something should be removed by looking up at `isFormatting` attribute property](https://github.com/ckeditor/ckeditor5/blob/4a52f6c079c7d4a518aee4d87f15bdb8ae2cc575/packages/ckeditor5-remove-format/src/removeformatcommand.js#L109).

Our [hand-created features add this attribute](https://github.com/ckeditor/ckeditor5/blob/4a52f6c079c7d4a518aee4d87f15bdb8ae2cc575/packages/ckeditor5-basic-styles/src/underline/underlineediting.js#L40) but runtime attributes created by GHS does not:

If GHS would include a map of known formatting elements this would enable it to work well with remove formatting.

Now, I know there are dozens of formatting elements and attributes in HTML. To keep things simple we don't have to necessarily implement all possible. Just having parity with our hand-crafted features (basic styles, font size/color/bg/face) would be a good start.

## 📃 Other details

*   Browser: Any
*   OS: Any

---

mmotyczynska:

Fix (html-support): Basic styles should now work well with remove formatting feature. Closes #12626.

---

### Additional information

I added `RemoveFormat` to `ghs-all-features.html` manual test to make it easier to verify if this works.


mlewand (approved):
LGTM

PR #12726, opened 2022-10-25
Added toggleImageCaption that was missing from the metadata of ckeditor5-image

przemyslaw-zan:

Internal (image): Added the `toggleImageCaption` button to the image toolbar.

---

### Additional information


psmyrek:
Good catch @przemyslaw-zan, but the `ImageBlock` or `ImageInline` plugins should know nothing about the `toggleImageCaption` button, which is registered by the `ImageCaption` plugin. The `ImageCaption` plugin is a separate plugin, which should be responsible for registering own UI components in a toolbar.

I've pushed changes to your branch.

PR #12760, opened 2022-10-28
MutationObserver is no longer watching attribute changes.

### LINKED ISSUE #12759: [GHS] details/summary cannot be expanded by click
On `master`, clicking the arrow `►` next to Summary does not have any effect.

On the previous release, it works fine.

It changed because we extended the "revert changes that don't come from our View" mechanism of the mutation observer when working on support for IME.

We extended that mechanism in the initial phase of working on IME. It was necessary back then, but right now (we changed the implementation a couple of times) doesn't seem that critical.

We agree that a mechanism that reverts direct DOM attribute changes that were potentially made by mistake by e.g. an integrator should, in general, have positive effect on DX because it will immediately give feedback that such change should not be made this way.

However, it backfired at us here (and in one more case related to WSC). We're afraid that it may affect more integration scenarios. 

Hence: let's revert mutation observer to what it did previously: Only reverting structural changes.

niegowski:

Internal (engine): MutationObserver is no longer watching attribute changes. Closes #12759.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/view/observer/mutationobserver.ts
+++ b/packages/ckeditor5-engine/src/view/observer/mutationobserver.ts
@@ -53,7 +53,6 @@ export default class MutationObserver extends Observer {
 		this._config = {
 			childList: true,
 			characterData: true,
-			attributes: true,
 			subtree: true
 		};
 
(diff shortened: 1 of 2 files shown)


arkflpc (approved):

Reinmar:
@niegowski Could you write down some manual testing ideas for the QA team? I wonder if there are some areas affected by this PR that are worth doublechecking now.

niegowski:
> @niegowski Could you write down some manual testing ideas for the QA team? I wonder if there are some areas affected by this PR that are worth doublechecking now.

@ckeditor/qa-team Maybe you could check the composition on a block and an inline widget, and on the widget fake caret (before and after a widget). I'm not sure if there could be any other scenarios.

Mgsy:
We've tested it in all browsers and looks good :+1: .

PR #12765, opened 2022-10-31
Create the working example of the external resource widget

pszczesniak:
Draft of the external resource widget (working example).


pszczesniak:
Suggested approach is not a good in resolving this issue.

PR #12773, opened 2022-11-02
Manage raw html data outside the editor

pszczesniak:
Compare two approaches of external source widget


pszczesniak:
Was created to compare two branches - not needed at all.

PR #12795, opened 2022-11-04
Normalizing beforeInput:insertText that contains new-line characters

### LINKED ISSUE #2045: [iOS] Dictation is not working ("new line" breaks the editor)

🐞 Bug report

## 📋 Steps to reproduce

1. Open the classic editor on https://ckeditor.com/ckeditor-5/demo/ in Safari on an iPhone/iPad.
2. Activate dictation
3. Say "test text new line test text"

## ✅ Expected result
The following text in the editor:
Test text
Test text
## ❎ Actual result
Test text
It seems like after I say "new line" the dictation is not working anymore.

### LINKED ISSUE #11443: Mac multiline text replacement shortcuts not working

On macs, you can create multiline text replacement shortcuts. (See accepted answer here: https://superuser.com/questions/1541755/multi-line-text-replacement-on-macos-or-ios). When using these shortcuts in CKEditor the shortcuts aren't expanded.

For example, if my shortcut is: `mline` which is replaced with:
```
make

multiline

magic
```

(Only the first line of the shortcut is displayed)
image

then when using that shortcut in CKEditor:
Screen Shot 2022-03-09 at 12 06 54 PM

the result is:
Screen Shot 2022-03-09 at 12 06 58 PM

### ✔️ Expected result

The shortcut is expanded and all lines are entered into the editor.
image

### ❌ Actual result

Only the first line is inserted:
Screen Shot 2022-03-09 at 12 06 58 PM

### ❓ Possible solution

I'm pretty sure Apple doesn't really support multiline shortcuts, however, you can sort've trick the text replacement input field into using a multiline shortcut by seperating lines with soft-breaks. My guess is that something with the soft-break is messing up some logic in CKEditor.

## 📃 Other details

* Browser: Chrome Version 98.0.4758.109 (Official Build) (x86_64)
* OS: macOS Big Sur 11.6.2
* Installed CKEditor plugins: was able to reproduce on the classic editor build

---

niegowski:

Other (engine): Normalizing `beforeInput:insertText` that contains new-line characters in data. Closes #2045. Closes #11443.

---

### Additional information

This PR is addressing VoiceControl (not the dictation available on the keyboard).


arkflpc (approved):

niegowski:
@ckeditor/qa-team could you test this PR?

Acrophost:
Since there is no option to enable VoiceControl on simulator, @dufipl could you test the https://github.com/ckeditor/ckeditor5/issues/2045 part? Thank you!

I have 2 remarks for #11443:

- in captions, we have two behaviours: the enter does not work, so when the option to expand is clicked, we get the supposedly multi-line text in one line or if we click anywhere besides the option, the expanded text appears on top of editor content

https://user-images.githubusercontent.com/66953380/200556933-33f78fbb-d64d-4a7b-958a-65c3ff9d1cba.mov

- in restricted mode the text doesn't expand with enter - it does with mouse click, backspace or spacebar. The multi-line text appears in one line. This one acts a little wonky in general, since when added with backspace, the whitespace after the expanded text gets deleted and when added with spacebar, selection appears after expanded text instead after added whitespace.

https://user-images.githubusercontent.com/66953380/200555639-07a03488-da9a-4541-b6b7-e256200e7d24.mov

dufipl:
Regarding #2045 iOS Voice Control works well on `ck/2045` (an example scenario from this issue was working on `master` also. 

However, this PR does not resolve #2045 IMO, because dictation still works very poorly.

When trying to dictate "Test text new line test text" different weird things happen, having in common that the editor moves to the following line and nothing happens but dictation is still enabled. When I disable dictation more text shows up and it is duplicated in different ways e.g.:

IMG_4A4A20B37C5F-1

IMG_560FFB6EB140-1

As I see here in the @Reinmar comment https://github.com/ckeditor/ckeditor5/issues/2045#issuecomment-1302191245, we might want to consider this as an upstream issue, fill a ticket to WebKit and resolve #2045 as upstream anyway.

Reinmar:
> As I see here in the @Reinmar comment [#2045 (comment)](https://github.com/ckeditor/ckeditor5/issues/2045#issuecomment-1302191245), we might want to consider this as an upstream issue, fill a ticket to WebKit and resolve #2045 as upstream anyway.

Yep, this is an upstream issue. In that mode, iOS does crazy things on all editors. Nothing we can do about this now.

> I have 2 remarks for #11443:

I think in both cases the behavior wasn't that bad. It's hard to say what should happen to a multi-line text in places where new lines are not allowed. It didn't crash, it just stripped them so that's more or less fine.

PR #12829, opened 2022-11-10
Special characters form header should use a heading markup

### LINKED ISSUE #12464: Special characters form header should use a heading markup

The special characters form heading is represented as a `<span>`. To improve the accessibility we should consider changing it to a proper `<h*>` tag.

---

illia-stv:

Fix (Special Characters): Special characters form header should use a heading markup. Closes #12464.

---

### Additional information
In this pull request I also provided little changes of neighbouring unit test, I mean "should set the template". First reason is better to getting to property through template not through DOM. Second reason is to increase the readability of test in case it failed. 

_For example – It will no display "**expect false to be true**", however it will display "**expect [ 'ck', 'ck-form__header'] should include  'wrong data'** "_


Dumluregn (approved):
:+1:

PR #12848, opened 2022-11-15
Added backward compatibility check for DLLs

### LINKED ISSUE #12756: Check if the upcoming CKEditor 5 DLL build is backward compatible

We want to verify whether a new CKEditor 5 DLL build (built from the code that will be released) is compatible with a DLL build produced using the latest published CKEditor 5 DLL build.

The goal is to avoid a problem like #12752 in the future.

We need to figure out how a sample or a new manual test that loads the latest CKEditor 5 DLL build and an older (external) plugin.

martnpaneq:
Internal (ci): Added backward compatibility check for DLLs. Closes #12756

### DIFF
--- a/scripts/check-core-backward-compability.sh
+++ b/scripts/check-core-backward-compability.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# @license Copyright (c) 2003-2022, CKSource Holding sp. z o.o. All rights reserved.
+# For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+
+# Checkout release branch and update with origin
+git checkout release
+git pull origin release
+
+# Build dlls for core
+yarn dll:build
+
+# Checkout previous commit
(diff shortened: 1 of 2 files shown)


martnpaneq:
In order to test this PR, change `HEAD~1` to `6104c4bdec`. It's the commit that broke the backward compatibility last time.
image

pomek (approved):
We implemented a few changes:

* Use HEAD instead of switching and pulling branches.
* Use the latest tag (release) instead of a commit before HEAD.
* Compile a single manual test. It's enough.

Then, we checked how it works.

1. In CKEditor 5 DLL manifest file, we changed JS paths to TypeScript files. We expected to detect an error.
	image
2. In the same manifest file, use JS files. It should pass tests.
	image

So, LGTM. It does the job.

pomek:
So far, we didn't focus on including the check on CI, as the script needs to know the latest tag. CI clones a particular number of commits, so that might be unavailable.

PR #12858, opened 2022-11-16
The special character label in dropdown doesn't get updated during the keyboard navigation

### LINKED ISSUE #12393: The special character label in dropdown doesn't get updated during the keyboard navigation

1.  Go to a demo with Special characters feature, e.g. https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/full-featured-editor.html.
2.  Open Special characters dropdown.
3.  Press tab to move the focus to the characters grid.
4.  Observe the label view below the grid.

### ✔️ Expected result

The label of focused element is displayed.

### ❌ Actual result

The label doesn't react on changing the focus.

The issue is also visible if you hover over some symbol with mouse, then move the mouse cursor away of the dropdown (the label is still displayed) and then start navigating with keyboard - the label doesn't update.

## 📃 Other details

*   Browser: any
*   OS: any
*   First affected CKEditor version: 35.0.1 - introduction of keyboard grid navigation
*   Installed CKEditor plugins: `special-characters`

---

illia-stv:

Fix (special-characters): The special character label in dropdown should be updated when navigating with keyboard. Closes #12393.

---

### Additional information


mmotyczynska (changes requested):
The solution you proposed works, but I would suggest some small changes to keep the code more consistent. As described in comments below please use `focus` instead of `keyup`. Besides that please fire `tileFocus` on focus instead of `tileHover`. After these changes update the unit test and remember to add the new event to API docs.

packages/ckeditor5-special-characters/tests/ui/charactergridview.js
```diff
@@ -179,6 +179,17 @@ describe( 'CharacterGridView', () => {
 			sinon.assert.calledOnce( spy );
 			sinon.assert.calledWithExactly( spy, sinon.match.any, { name: 'foo bar baz', character: 'ε' } );
 		} );
+
+		it( 'delegates #tileFocus from the tile to the grid on keyup event', () => {
```
mmotyczynska, line 183:
```suggestion
		it( 'delegates #tileFocus from the tile to the grid on focus the tile', () => {
```

packages/ckeditor5-special-characters/src/ui/charactergridview.js
```diff
@@ -128,14 +128,19 @@ export default class CharacterGridView extends View {
...
+				focus: tile.bindTemplate.to( 'focus' )
 			}
 		} );
 
 		tile.on( 'mouseover', () => {
 			this.fire( 'tileHover', { name, character } );
 		} );
 
+		tile.on( 'focus', () => {
+			this.fire( 'tileFocus', { name, character } );
```
mmotyczynska, line 141: The `tileFocus` you introduced here has to be documented in our API docs. You can check how the `tileHover` was documented and create a similar description based on that: https://github.com/ckeditor/ckeditor5/blob/38dd963e029cdff97e15fc41d8421804221f5644/packages/ckeditor5-special-characters/src/ui/charactergridview.js#L97-L105

mmotyczynska (changes requested):
Almost ready :smile: Please correct test description and add API docs for `tileFocus` as described in the comments.

mmotyczynska (approved):
LGTM! :+1:  Thanks for updating the PR description too. Well done :smiley:

PR #12860, opened 2022-11-16
Ck/12720 cleaner imports

### LINKED ISSUE #12720: [TS] Research how to make import and re-export cleaner
Currently, we have the following ways of importing from other packages:
1. In all non-essential packages, we import from `ckeditor5/src/*`. In practice, that mean importing from respective `index.js`.
2. In essential packages, we import either from package's `index.js` or directly from a specific file. The way seems to depend on what IDE suggested at the time of adding this import.

The stuff being exported from `index.js`es (and, from `ckeditor5/src`) are helper functions and classes meant to be instantiated directly. However, when moving to TypeScript, a code have to import way more classes (and interfaces) from other packages. That is because every local helper function needs to annotate its parameters. And there are local variables where type inference is not enough.
There are also 

Let's research how to make all of it look clean and consistent. Consider the following:
- An Essential package importing stuff from other essential package.
- Not essential package importing stuff from essential one.
- Not essential package importing types from another non-essential package (new case added by TypeScript).

Scope:
1. Create a branch with example how it could look like.
2. Propose tasks to create.

arkflpc:

Internal: Changes import to target packages instead of individual files. Closes #12720.

---

### Additional information


packages/ckeditor5-ui/src/labeledfield/labeledfieldview.ts
```diff
@@ -8,13 +8,14 @@
  */
 
 import View from '../view';
-import uid from '@ckeditor/ckeditor5-utils/src/uid';
 import LabelView from '../label/labelview';
+
+import { uid, type Locale } from '@ckeditor/ckeditor5-utils';
+
 import '../../theme/components/labeledfield/labeledfieldview.css';
```
niegowski, line 15: Let's move it to the bottom of imports block.

packages/ckeditor5-ui/src/labeledinput/labeledinputview.ts
```diff
@@ -8,12 +8,13 @@
  */
 
 import View from '../view';
-import uid from '@ckeditor/ckeditor5-utils/src/uid';
 import LabelView from '../label/labelview';
+
+import { uid, type Locale } from '@ckeditor/ckeditor5-utils';
+
 import '../../theme/components/labeledinput/labeledinput.css';
```
niegowski, line 15: Let's move it to the bottom of imports block.

PR #12866, opened 2022-11-17
Docs (clipboard): Update clipboardInput event signature

### LINKED ISSUE #12839: `Clipboard:clipboardInput` event docs list 4 parameters when only 2 are passed.
The documentation for the [`clipboardInput` event](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_document-Document.html#event-event:clipboardInput) incorrectly shows that 4 arguments are passed to the event handler. In reality, only the `event` and `data` arguments are passed. `method` and `target` are both part of the `data` object, not separate arguments.
This can be confirmed in the [code](https://github.com/ckeditor/ckeditor5/blob/8b87a8497fe20a8f2913108ff47ff2b1dfcb1cf2/packages/ckeditor5-clipboard/_src/clipboardobserver.js#L57):

```js
viewDocument.fire(eventInfo, {
  dataTransfer: data.dataTransfer,
  method: evt.name,
  targetRanges,
  target: data.target,
});
```
The [docs](https://github.com/ckeditor/ckeditor5/blob/8b87a8497fe20a8f2913108ff47ff2b1dfcb1cf2/packages/ckeditor5-clipboard/_src/clipboardobserver.js#L111-L135) should be updated to reflect that.

tomasr8:
Closes #12839

Rebuilt docs show the correct function signature:

image

### DIFF
--- a/packages/ckeditor5-clipboard/_src/clipboardobserver.js
+++ b/packages/ckeditor5-clipboard/_src/clipboardobserver.js
@@ -126,8 +126,8 @@ function getDropViewRange( view, domEvent ) {
  * @event module:engine/view/document~Document#event:clipboardInput
  * @param {Object} data The event data.
  * @param {module:engine/view/datatransfer~DataTransfer} data.dataTransfer Data transfer instance.
- * @param {'paste'|'drop'} method Whether the event was triggered by a paste or drop operation.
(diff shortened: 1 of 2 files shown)


Reinmar:
@arkflpc I guess we have a conflict due to the migration to TS but perhaps we fixed the same issue there when migrating. Could you verify?

PR #12868, opened 2022-11-17
Add support for "is" attribute in Template and createElement

### LINKED ISSUE #12867: Support the "is" attribute in createElement and Template

It is currently not possible to use custom HTML elements that [extend existing elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#customized_built-in_elements) (such as a custom input) in plugins as neither [utils/dom/createelement](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/dom/createelement.ts) nor [ui/template](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-ui/src/template.ts) support the "is" attribute.

## 📃 Other details

Methods that call `createElementNS` and `createElement` should probably check for an `is` attribute and use it to populate the "[options](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS#parameters)" parameter.

---

semiaddict:
Closes: #12867

Other (ui): Add support for the "is" attribute in Template. Closes #12867.

Other (utils): Add support for the "is" attribute in createElement. Closes #12867.

---

### Additional information

This MR allows using custom elements that extend native HTML elements in templates and via the createElement utility method. It basically checks if the "is" attribute is set, and, if so, passes it to the createElement or createElementNS native methods.

### DIFF
--- a/packages/ckeditor5-ui/src/template.ts
+++ b/packages/ckeditor5-ui/src/template.ts
@@ -460,7 +460,7 @@ export default class Template extends EmitterMixin() {
 		let node = data.node;
 
 		if ( !node ) {
-			const options = this.attributes && this.attributes.is ? { is: this.attributes.is } : undefined;
+			const options = this.attributes && this.attributes.is ? { is: this.attributes.is as any } : undefined;
 			node = data.node = document.createElementNS( this.ns || xhtmlNs, this.tag!, options ) as any;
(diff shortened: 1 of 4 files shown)


Reinmar:
cc @oleq

PR #12895, opened 2022-11-21
Redesigning the Update section TOC

godai78:

Type: Redesigning the Update section TOC and structure. Closes cksource/ckeditor5-internal#2238


AnnaTomanek:
Just a minor note from me: I'd be against using shortcuts such as "cke4" and "cke5" in URLs - while meaningful for us internally, they are unclear for the external audience so it's not just a SEO problem, it's simply confusing. Better to have a URL that is clear from the first look what it is about, and the migration/updating pages are extremely important.

godai78:
> Just a minor note from me: I'd be against using shortcuts such as "cke4" and "cke5" in URLs - while meaningful for us internally, they are unclear for the external audience so it's not just an SEO problem, it's simply confusing. Better to have a URL that is clear from the first look what it is about, and the migration/updating pages are extremely important.

This was due to Umberto returning an error of "duplicate files". We can revisit this and check if there is a solution since this was moved deeper into the structure.

PR #12899, opened 2022-11-21
Keyboard navigation in insert table dropdown should not be reversed in rtl editor

### LINKED ISSUE #12871: Keyboard navigation in grid dropdowns should not be reversed in RTL editor

1.  Open [demo with RTL editor](https://ckeditor.com/docs/ckeditor5/latest/features/ui-language.html#demo-2).
2.  Open the insert table dropdown.
3.  Press arrow left.

### ❓ Possible solution

The right/left arrow navigation is reversed. It should work naturally instead.

The same thing happens for all grids in the editor UI:
- color picker
- styles
- special characters
- list styles

## 📃 Other details

*   Browser: …
*   OS: …

---

illia-stv:

Fix (Table): Keyboard navigation in insert table dropdown should not be reversed in RTL editor. Closes #12871.

---

### Additional information


packages/ckeditor5-ui/tests/bindings/addkeyboardhandlingforgrid.js
```diff
@@ -275,6 +275,72 @@ describe( 'addKeyboardHandlingForGrid()', () => {
 		} );
 	} );
 
+	describe( 'arrows moves in rtl', () => {
+		let view, keystrokes, focusTracker, gridElementsCollection, numberOfColumnsStub, uiLanguageDirection;
```
Dumluregn, line 279: You don't really need any of those variables:

*   `view, keystrokes, focusTracker, gridElementsCollection` are all already in the scope so you can reassign them.
*   `numberOfColumnsStub, uiLanguageDirection` have always the same values in your tests, so these values can be used directly.

Dumluregn (changes requested):
Besides the inline comments - there are still two more grids awaiting the handling: list styles:



and styles dropdown:

packages/ckeditor5-special-characters/src/ui/charactergridview.js
```diff
@@ -82,7 +82,8 @@ export default class CharacterGridView extends View {
 				.getComputedStyle( this.element.firstChild ) // Responsive .ck-character-grid__tiles
 				.getPropertyValue( 'grid-template-columns' )
 				.split( ' ' )
-				.length
+				.length,
+			uiLanguageDirection: this.locale && this.locale.uiLanguageDirection
```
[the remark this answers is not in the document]
illia-stv: Screenshot 2022-11-24 at 12 34 01

packages/ckeditor5-ui/src/bindings/addkeyboardhandlingforgrid.ts
```diff
@@ -84,4 +82,52 @@ export default function addKeyboardHandlingForGrid(
 			evt.preventDefault();
 		};
 	}
+
+	// Function returning a right index relatively current index.
```
Dumluregn, line 86: The function actually doesn't know, how many elements there are in the grid row. It sees the collection linearly - only knows where it starts and ends. So the more correct description is that it returns a next index, and if the last element is being checked, returns the first index (0).

packages/ckeditor5-ui/src/bindings/addkeyboardhandlingforgrid.ts
```diff
@@ -84,4 +82,52 @@ export default function addKeyboardHandlingForGrid(
...
+	// If current index is at the left conner, function return last index of previous row.
+	//
+	// before: [ ][ ][ ]	after: [ ][ ][x]
+	//         [x][ ][ ]	       [ ][ ][ ]
+	//         [ ]      	       [ ]
+	// index = 2            index = 0
+	//
+	// @param {number} [elementIndex] Number of current index.
+	// @param {module:ui/viewcollection~ViewCollection} [gridItems] A collection of grid items.
+	function getLeftElementIndex( elementIndex: number, gridItems: ViewCollection ) {
```
Dumluregn, line 126: Same comments as for the previous function, only with a change in the docs - if the first index is passed, it returns the last one.

packages/ckeditor5-ui/src/bindings/addkeyboardhandlingforgrid.ts
```diff
@@ -84,4 +82,52 @@ export default function addKeyboardHandlingForGrid(
...
+	//
+	// @param {number} [elementIndex] Number of current index.
+	// @param {module:ui/viewcollection~ViewCollection} [gridItems] A collection of grid items.
+	function getRightElementIndex( elementIndex: number, gridItems: ViewCollection ) {
+		if ( elementIndex === gridItems.length - 1 ) {
+			return 0;
+		} else {
+			return elementIndex + 1;
+		}
+	}
```
Dumluregn, line 108: It's usually a good practise to only pass what is necessary to the function - it makes it easier to understand, refactor etc. Here, if you look carefully, we're passing `gridItems` object, but the function actually only uses its length - so it makes sense to pass just that - the length of the collection, and name it e.g. `collectionLength`. It will show that we don't do any operations on the collection inside the function.

packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.js
```diff
@@ -203,7 +203,8 @@ export default class ListPropertiesView extends View {
 					.getComputedStyle( this.stylesView.element )
 					.getPropertyValue( 'grid-template-columns' )
 					.split( ' ' )
-					.length
+					.length,
+				uiLanguageDirection: this.locale.uiLanguageDirection
```
Dumluregn, line 207: I did some digging and I see that `locale` is optional in all the packages that are already rewritten to TS, so we can safely assume it will happen also for this one - let's add the check you've provided elsewhere: `this.locale && this.locale.uiLanguageDirection`.

packages/ckeditor5-style/src/ui/stylegridview.js
```diff
@@ -130,7 +130,8 @@ export default class StyleGridView extends View {
 			keystrokeHandler: this.keystrokes,
 			focusTracker: this.focusTracker,
 			gridItems: this.children,
-			numberOfColumns: 3
+			numberOfColumns: 3,
+			uiLanguageDirection: this.locale.uiLanguageDirection
```
Dumluregn, line 134: I did some digging and I see that `locale` is optional in all the packages that are already rewritten to TS, so we can safely assume it will happen also for this one - let's add the check you've provided elsewhere: `this.locale && this.locale.uiLanguageDirection`.

packages/ckeditor5-table/src/ui/inserttableview.js
```diff
@@ -158,7 +158,8 @@ export default class InsertTableView extends View {
 			keystrokeHandler: this.keystrokes,
 			focusTracker: this.focusTracker,
 			gridItems: this.items,
-			numberOfColumns: 10
+			numberOfColumns: 10,
+			uiLanguageDirection: this.locale.uiLanguageDirection
```
Dumluregn, line 162: I did some digging and I see that `locale` is optional in all the packages that are already rewritten to TS, so we can safely assume it will happen also for this one - let's add the check you've provided elsewhere: `this.locale && this.locale.uiLanguageDirection`.

packages/ckeditor5-ui/src/bindings/addkeyboardhandlingforgrid.ts
```diff
@@ -21,31 +21,29 @@ import type ViewCollection from '../viewcollection';
  * @param {module:ui/viewcollection~ViewCollection} options.gridItems A collection of grid items.
  * @param {Number|Function} options.numberOfColumns Number of columns in the grid. Can be specified as a function that returns
  * the number (e.g. for responsive grids).
+ * @param {String|Undefined} options.uiLanguageDirection String of ui language direction.
```
Dumluregn, line 24: We use lowercase `undefined` for types (same with `null`).

Dumluregn (changes requested):

Dumluregn (approved):
LGTM, well done :clap:

PR #12944, opened 2022-11-29
Markers should not be lost while upcasting a plain table (without a <figure> element).

niegowski:

Fix (html-support): Markers should not be lost while upcasting a plain table (without a `<figure>` element).

---

### Additional information

### DIFF
--- a/packages/ckeditor5-html-support/tests/integrations/table.js
+++ b/packages/ckeditor5-html-support/tests/integrations/table.js
@@ -942,6 +942,223 @@ describe( 'TableElementSupport', () => {
 		await editor.destroy();
 	} );
 
+	it( 'should upcast GHS attributes at the low priority (feature attribute converter at low + 1 priority)', () => {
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-html-support/tests/integrations/table.js
```diff
@@ -942,6 +942,223 @@ describe( 'TableElementSupport', () => {
...
+			styles: true,
+			classes: true
+		} ] );
+
+		editor.model.schema.extend( 'table', { allowAttributes: [ 'barAttr' ] } );
+
+		editor.conversion.for( 'upcast' ).attributeToAttribute( {
+			view: 'foo-attr',
+			model: 'barAttr',
+			converterPriority: priorities.get( 'low' ) + 1
```
arkflpc, line 958:
```suggestion
			converterPriority: priorities.low + 1
```

arkflpc (approved):

PR #12946, opened 2022-11-29
Fixed infinite loop after transpilation to ES5

psmyrek:

Internal: Fixed infinite loop after transpilation to ES5. Closes ckeditor/ckeditor5-react#345.

---

### Additional information

The description how to reproduce this issue is in ckeditor/ckeditor5-react#345.

In short, `create-react-app` uses by default very loose `browserslist` rules for production environment, that cause the generated (minified) code of the application targets the ES5. This in turn causes that the Babel (probably) does not handle properly the inline destructuring inside a `for of` loop with a generator.

### DIFF
--- a/packages/ckeditor5-list/src/documentlist/utils/postfixers.js
+++ b/packages/ckeditor5-list/src/documentlist/utils/postfixers.js
@@ -30,7 +30,9 @@ export function findAndAddListHeadToMap( position, itemToListHead ) {
 	} else {
 		let listHead = previousNode;
 
-		for ( { node: listHead } of iterateSiblingListBlocks( listHead, 'backward' ) ) {
+		for ( const { node } of iterateSiblingListBlocks( listHead, 'backward' ) ) {
+			listHead = node;
+
 			if ( itemToListHead.has( listHead ) ) {
 				return;
 			}


psmyrek:
The difference in generated minified code for `findAndAddListHeadToMap()` before (red) and after (green) this change:

```diff
function Vg( e, t ) {
     var n = e.nodeBefore;

     if ( pg( n ) ) {
         var r, i = n, o = f( dg ( i, "backward" ) );

         try {
             for ( o.s(); !( r = o.n() ).done; )
-                if ( e = r.value.node, t.has( i ) )
+                if ( i = r.value.node, t.has( i ) )
                     return
         } catch( s ) {
             o.e( s )
         } finally {
             o.f()
         }

         t.set( n, i )
     } else {
         var a = e.nodeAfter;

         pg( a ) && t.set( a, a )
     }
}
```

From what I see, the `o = f( dg ( i, "backward" ) )` creates an iterator, starting from `i`, which is `position.nodeBefore` stored as `listHead` in the source code. But then the `i` was never updated, so the loop never ended.

For a reference, below is the original function:

```js
function findAndAddListHeadToMap( position, itemToListHead ) {
    const previousNode = position.nodeBefore;

    if ( !isListItemBlock( previousNode ) ) {
        const item = position.nodeAfter;

        if ( isListItemBlock( item ) ) {
            itemToListHead.set( item, item );
        }
    } else {
        let listHead = previousNode;

        for ( { node: listHead } of iterateSiblingListBlocks( listHead, 'backward' ) ) {
            if ( itemToListHead.has( listHead ) ) {
                return;
            }
        }

        itemToListHead.set( previousNode, listHead );
    }
}
```

PR #12991, opened 2022-12-08
Ck/12853 move EditorUI to ui

### LINKED ISSUE #12853: Move EditorUI to ckeditor5-ui
There is circular dependency between `ckeditor5-core` and `ckeditor5-ui` packages. The problematic class is `EditorUI` that has dependency to `ckeditor5-ui`. The idea is to move `EditorUI` to `ckeditor5-ui`.

It would allow full implementation of #12720 (so that imports from other packages are only through `index.ts`, now it is not fully possible because of this circular dependency).

Scope:

*   Move `EditorUI` to `ckeditor5-ui` along with its tests.
*   Move `ckeditor5-ui` from `dependencies` to `devDependencies` in `packages/ckeditor5-core/package.json`.
*   Update all imports.
*   Update the documentation: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/ui/custom-editor-creator.html + https://ckeditor.com/docs/ckeditor5/latest/examples/framework/multi-root-editor.html  and test implementations of the MultiRootEditor.
*   Create a migration guide (we know there are integrators that created their own editor implementations, so let's help them a bit).

Out of scope:

*   `ClassicTestEditor` etc. can stay where it is.
*   `Editor` class may import `EditorUI` as `import type` (circular type dependencies are ok).

This will be minor breaking change.

arkflpc:

Other (core, ui): Moves `EditorUI` from `ckeditor5-core` to `ckeditor5-ui`. Closes #12853.

MAJOR BREAKING CHANGE: The `EditorUI` class was moved from `@ckeditor/ckeditor5-core` to `@ckeditor/ckeditor5-ui`. Thus, it's enough to update imports (depending on the format used):

* from `'ckeditor5/src/core'` to `'ckeditor5/src/ui'`
* from `'@ckeditor/ckeditor5-core/src/editor/editorui'` to `'@ckeditor/ckeditor5-ui/src/editorui/editorui'`

---

### Additional information


docs/updating/update-to-35.md
```diff
@@ -138,7 +138,7 @@ export default class MyEditorUI extends EditorUI {
 **After**:
 
 ```js
-import { EditorUI } from 'ckeditor5/src/core';
+import { EditorUI } from 'ckeditor5/src/ui';
```
Reinmar, line 141: Requires additional comment here. The previous path will work when upgrading to 35.0.0. But when upgrading to 36.0.0 it needs to be the proposed one.

Reinmar: It makes sense to explain in this guide that this file's path changed in 36.0.0.

Reinmar:
> MAJOR BREAKING CHANGE: The `EditorUI` class can no longer be imported from `'ckeditor5/src/core'`. Imports should be changed to `'ckeditor5/src/ui'`.

vs

image

I'm unsure what makes a bigger sense in the changelog, but I suppose the path that we used in the docs.

niegowski (approved):

docs/updating/update-to-35.md
arkflpc: I wouldn't. This is an example for dealing with `enableToolbarKeyboardFocus` which is unrelated.

docs/updating/update-to-36.md
```diff
@@ -0,0 +1,43 @@
...
+
+The `EditorUI` class was moved from @ckeditor/ckeditor5-core to @ckeditor/ckeditor5-ui. Thus, it's enough to update imports (depending on the format used):
+
+* from `'ckeditor5/src/core'` to `'ckeditor5/src/ui'`
+* from `'@ckeditor/ckeditor5-core/src/editor/editorui'` to `'@ckeditor/ckeditor5-ui/src/editorui/editorui'`
+
+**Before**:
+
+```js
+import { EditorUI } from 'ckeditor5/src/core';
```
arkflpc, line 32:
```suggestion
import { EditorUI } from '@ckeditor/ckeditor5-core/src/editor/editorui';
```

docs/updating/update-to-36.md
```diff
@@ -0,0 +1,43 @@
...
+
+export default class MyEditorUI extends EditorUI {}
+```
+
+**After**:
+
+```js
+import { EditorUI } from 'ckeditor5/src/ui';
```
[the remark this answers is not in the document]
arkflpc:
```suggestion
import { EditorUI } from '@ckeditor/ckeditor5-ui/src/editorui/editorui';
```

PR #12992, opened 2022-12-08
Delete unnecessary branding from links

gorzelinski:

Docs: Delete unnecessary branding from links.

Closes [ckeditor5-internal#2312](https://github.com/cksource/ckeditor5-internal/issues/2312)

---

### Additional information

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -8,13 +8,13 @@ meta-description: Learn how to install, integrate, update, configure and develop
 
 # CKEditor 5 documentation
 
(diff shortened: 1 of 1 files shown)


dufipl:
@apadol-cksource maybe we could consider leaving "Updating and maintaining" instead of just "Updating"? 🤔

apadol-cksource:
> @apadol-cksource maybe we could consider leaving "Updating and maintaining" instead of just "Updating"? 🤔

Let's proceed with "Updating and maintaining". Just "Updates" or "Releases" would be even better but we would need to change it in several places.

dufipl (approved):

Issue #11533, opened 2022-03-30
Downcasting table cells is incorrect if there is a marker on the paragraph


scofalik:
## 📝 Provide detailed reproduction steps (if any)

1. Create a 1x1 table, with an empty cell or a cell with only one paragraph.
2. Turn on track changes.
3. Change paragraph into list item.
4. Call `editor.getData()`.

### ✔️ Expected result

`<p>` is in the data, and there is a marker on it.

### ❌ Actual result

Invalid data, `<p>` is not in the result, and the marker is "moved" onto the table cell. Loading such data is incorrect (suggestion is on table cell).

Additionally, this is problematic for revision history, because `<p>` is not available but there is a marker tag in `<td>`. This means that `<p>` will not be inserted during upcast, so it "magically" disappears, messing the compare algorithm.

### ❓ Possible solution

I have three ideas:

1. First is obvious - check for markers when downcasting `paragraph` in table cell. Similarly as here: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-table/src/converters/downcast.js#L122. Problem: I am not sure if it will be easy to get list of markers at this point in the code.
1. Remove `<p>` in ~the view postfixer~ a lowest-priority converter. I am not sure why it is not done this way and I also have a feeling that in the past, we have been doing this in view... but maybe it was done differently... IDK.
1. Improve upcasting. I am not sure but it seems, that for a table cell with text (`<td>Foo</td>`) the paragraph is added in the post-fixer, not during upcasting. We have this: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-table/src/converters/upcasttable.js#L147 but it only works for empty `<td></td>`. So it covers neither `<td>Foo</td>` nor `<td><suggestion-end ... /></td>`. So post-fixer is *needed* to correctly upcast data. This is wrong (see below - "One more thing").

Bonus: Actually, I think this converter is unnecessary and harmful as it works a bit "randomly". At the moment we have model post-fixer anyway. Without the converter, now, at least upcast effects will be more stable. Removing this converter fixes the issue in revision history (I never get `paragraph` elements if they weren't in the data, while now I sometimes get them and sometimes not).

Bonus 2: This can be fixed in revision history in a different way, but it is a hack, it would be nice to remove it at some point.

One more thing.

[...]

PR #13054, opened 2022-12-12
Fix tableCell content inconsistency after executing editor.data.set

### LINKED ISSUE #13053: Markers change positions in tableCell after calling editor.data.set
Markers change positions in tableCells after multiple `editor.data.set( editor.data.get() )` commands.

More information: https://github.com/cksource/collaboration-features/issues/4061

kbinieda:

Fix (table): Fix tableCell content inconsistency after executing editor.data.set. Closes #13053.

### Additional information

Connected to: https://github.com/cksource/collaboration-features/issues/4061

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
@@ -523,7 +523,16 @@ export function convertText() {
 				return;
 			}
 
+			// Wrap $text in paragraph and include any marker that is directly before $text. #13053
+			const nodeBefore = position.nodeBefore;
+
 			position = wrapInParagraph( position, writer );
+
+			if ( nodeBefore && nodeBefore.is( 'element', '$marker' ) ) {
+				// Move `$marker` to the paragraph.
+				writer.move( writer.createRangeOn( nodeBefore ), position );
+				position = writer.createPositionAfter( nodeBefore );
+			}
 		}
 
 		consumable.consume( data.viewItem );
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
@@ -523,7 +523,7 @@ export function convertText() {
 				return;
 			}
 
-			// Wrap $text in paragraph and include any marker that is directly before $text. #13053
+			// Wrap `$text` in paragraph and include any marker that is directly before `$text`. See #13053.
 			const nodeBefore = position.nodeBefore;
 
 			position = wrapInParagraph( position, writer );
--- a/packages/ckeditor5-engine/tests/conversion/upcasthelpers.js
+++ b/packages/ckeditor5-engine/tests/conversion/upcasthelpers.js
@@ -1200,6 +1200,31 @@ describe( 'upcast-converters', () => {
 			expect( conversionResult.getChild( 0 ).data ).to.equal( 'foobar' );
 		} );
 
+		it( 'should also include $marker when auto-paragraphing $text.', () => {
+			// Make $text invalid to trigger auto-paragraphing.
+			schema.addChildCheck( ( ctx, childDef ) => {
+				if ( ( childDef.name == '$text' ) && ctx.endsWith( '$root' ) ) {
+					return false;
+				}
+			} );
+
+			const viewText = new ViewText( viewDocument, 'foobar' );
+			dispatcher.on( 'text', ( evt, data, conversionApi ) => {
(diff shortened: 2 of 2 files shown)


scofalik (changes requested):
This is a fine base solution but there are some things I dislike, so let's try to come up with something better.

What I don't like is that:

1.  This solution is not efficient.
2.  We leave some mess in the generated HTML.

---

In this solution, for every cell, we check every marker. This is highly inefficient in some CKE5 use cases, where you edit mostly huge tables and you use track changes (users have different roles and permissions).

We've already seen CKE5 being used with tables with hundreds of rows and, at the same time, with hundreds of suggestions. 500 rows x 5 columns x 200 suggestions gives you 500000 checks, and these are not exaggerated numbers.

To fix the efficiency but stick with this solution, we should rather look through all markers and then see whether particular marker is on a table cell. This will be faster, as it will only depend on number of markers. The question is - where and when to do it?

I have two candidates: evaluate the list of "affected cells" on every marker update (listen to `editor.markers`), or add a callback to `DataController#get` in which we will perform the check for every marker (affected cells can be passed in `options`). cc @niegowski.

---

The second problem: it would be best not to leave any mess in HTML. But, if we add extra markup, It looks like a simple `<span>` does it job, so we could skip the CSS class (not in the case of `asWidget`, of course).

However, there was already a similar problem in revision history and it was solved by forcing `<p>` rather than `<span>`. Whatever we do here, we will need to check whether a new approach works well with revision history.

---

As a next step, I'd go with the fixes I proposed so that we have a nice base solution. But it doesn't feel right that we need to hack it this way. This should "just work" ™.

I found an old ticket where I discuss the same problem (the new ticket is actually a DUP of that one: #11533). There is another ticket linked. After we have a nice base solution, we can try to solve it even better.

scofalik:
The new solution is interesting as it does not change current solution heavily but it does not work well for images, for which it is correct to have a marker starting directly in a table cell (the editor throws an error because it tries to fix the marker but the `oldRange` is `null`).

We should not disallow such markers, such markers are okay. And we (CKS) are not only ones using markers. It should be allowed to create markers wherever people want.

So, I'd go back to the original solution and solve this in following way:

1.  Define a `Map` that will hold model table cell -> integer binding, where the integer is the number of markers in the cell. Or a similar structure.
2.  On markers `update` event, update the structure (remove table cell if there are no markers in that cell).
3.  Pass this structure to the converter.
4.  In the converter check if given table cell has markers, if so, return.

scofalik (changes requested):
As described.

packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
```diff
@@ -523,7 +523,18 @@ export function convertText() {
...
+				const markerNode = position.nodeBefore;
+				position = wrapInParagraph( position, writer );
+
+				// Move $marker to paragraph
+				writer.move( writer.createRangeOn( markerNode ), position );
+
+				position = writer.createPositionAfter( markerNode );
+			} else {
+				position = wrapInParagraph( position, writer );
+			}
```
scofalik, line 537:
```suggestion
			// Wrap `$text` in a paragraph and include any marker that is directly before `$text`. See #13053.
			const nodeBefore = position.nodeBefore;
			
			position = wrapInParagraph( position, writer );
			
			if ( nodeBefore && nodeBefore.is( 'element', '$marker' ) ) {
				// Move `$marker` to the paragraph.
				writer.move( writer.createRangeOn( nodeBefore ), position );

				position = writer.createPositionAfter( nodeBefore );
			}
```

Would this work?

scofalik (changes requested):
I proposed code simplification. Other than that, a nice idea.

I am afraid that there are other places in the engine where we autoparagh text. Could you try to to find them and research whether they need fixing too?

I think we autoparagraph empty roots, although I don't know whether a root with just one paragraph and marker inside is downcasted to `<p><marker /></p>` or `<marker />` probably the first one, so this doesn't count.

We also autoparagraph in `insertContent()` but this has it's own markers handling, so I'd leave it for now.

I don't know about other places.

packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
```diff
@@ -523,7 +523,16 @@ export function convertText() {
 				return;
 			}
 
+			// Wrap $text in paragraph and include any marker that is directly before $text. #13053
```
scofalik, line 526:
```suggestion
			// Wrap `$text` in paragraph and include any marker that is directly before `$text`. See #13053.
```

PR #13149, opened 2022-12-23
Rewrite ckeditor5-page-break to TypeScript.

### LINKED ISSUE #13026: [TS] Rewrite ckeditor5-page-break into TypeScript
Part of #11704.

martnpaneq:

Other (alignment): Rewrite ckeditor5-page-break to TypeScript. Closes #13026.


martnpaneq:
I'll add the `_src` directory when the review is finished.

packages/ckeditor5-page-break/src/pagebreakediting.ts
```diff
@@ -90,19 +91,19 @@ export default class PageBreakEditing extends Plugin {
...
 
 						// The child must be the "span" element that is not displayed.
-						if ( !viewSpan.is( 'element', 'span' ) || viewSpan.getStyle( 'display' ) != 'none' ) {
-							return;
+						if ( !viewSpan!.is( 'element', 'span' ) || viewSpan.getStyle( 'display' ) != 'none' ) {
+							return null;
 						}
 					} else if ( element.childCount > 1 ) {
-						return;
+						return null;
```
martnpaneq, line 106: I'm not sure if this is the correct way to go, to replace `undefined` with `null`. This might change the logic. But the type for method `view` expects type

```( element: Element ) => Match | null )```

so `null` is expected and not `undefined`.

arkflpc: It was intentional to omit `undefined` from the callback signature. So, callbacks must be explicit that they return no match. Otherwise, `undefined` could be returned silently if not all branches have returns.

PR #13162, opened 2022-12-28
Fixes pasting images from MS Word 2016.

### LINKED ISSUE #11993: Text with images pasted from Word 2016 doesn't preserve images.

1. Got to https://ckeditor.com/docs/ckeditor5/latest/features/pasting/paste-from-word.html
2. Extract attached file press Ctrl+A, Ctrl+C and paste its content into editor. 
3. You can also select some text with an image. It doesn't have to be the whole document
[copy_paste_image_+_text.docx.zip](https://github.com/ckeditor/ckeditor5/files/9018075/copy_paste_image_%2B_text.docx.zip)

### ✔️ Expected result

Image with text gets pasted.

### ❌ Actual result

Only text gets pasted.

### ❓ Possible solution

## 📃 Other details

From what we have checked, this MS Word file has been created with MS Word 2016 - https://superuser.com/a/1144401

* Browser: Any
* OS: Any
* Installed CKEditor plugins: paste-from-office

---

arkflpc:

Fix (paste-from-office): Fixes pasting images from MS Word 2016. Closes #11993.

---

### Additional information


niegowski (changes requested):
The code looks ok but it seems not to resolve the original issue. I checked with the file attached to the issue and the images are not visible.

niegowski:
> The code looks ok but it seems not to resolve the original issue. I checked with the file attached to the issue and the images are not visible.

Ok, it seems that it's a mixed issue. It works great while pasting on Windows, but on Mac, it stops working when any heading is selected and copied (no RTF in that case on Mac).

PR #13214, opened 2023-01-04
Set correct value (false) for CKFinderCommand#affectsData

### LINKED ISSUE #13213: Make CKFinder work with track changes.
Since CKFinder uses commands that are already integrated with track changes, the only change that should be done is to mark ckfinder `affectsData` as false.

More information here: https://github.com/cksource/collaboration-features/issues/4267

kbinieda:

Feature (track-changes): Add track changes integration for CKFinder feature.

Other (ckfinder): Set correct value (false) for `CKFinderCommand#affectsData`. Now the command's state depends only on related commands (`insertImage` and `link`). Closes #13213.

### DIFF
--- a/packages/ckeditor5-ckfinder/src/ckfindercommand.js
+++ b/packages/ckeditor5-ckfinder/src/ckfindercommand.js
@@ -33,6 +33,9 @@ export default class CKFinderCommand extends Command {
 	constructor( editor ) {
 		super( editor );
 
+		// The CKFinder command does not affect data by itself.
+		this._affectsData = false;
+
 		// Remove default document listener to lower its priority.
 		this.stopListening( this.editor.model.document, 'change' );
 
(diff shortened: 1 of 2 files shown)


scofalik:
I am missing any test checking this value. I wanted to propose to check if the command is enabled in read only mode but it would be disabled anyway as link and image are disabled in this mode. So, I guess, we need to check directly `affectsData`.

scofalik (changes requested):
I am missing any test checking this value. I wanted to propose to check if the command is enabled in read only mode but it would be disabled anyway as link and image are disabled in this mode. So, I guess, we need to check directly `affectsData`.

kbinieda:
> I am missing any test checking this value. I wanted to propose to check if the command is enabled in read only mode but it would be disabled anyway as link and image are disabled in this mode. So, I guess, we need to check directly `affectsData`.

I've added the test although, since the change does not introduce any behaviour (it's just a configuration preset) IMHO I don't think it needs testing.

scofalik:
I know this test is weak. We could come up with some artificial tests that checks it though -- after all, we are adding this property with some purpose in mind. This change fixes situation in track changes after all, so there's a scenario for it. But I am not a fan of such test as well. We should add a test in track changes checking whether CKF is properly enabled.

PR #13218, opened 2023-01-05
Other (image): Encapsulate image replacement into command

### LINKED ISSUE #13217: Make image replacement a command
Currently, image replacement is not done by command. In order to make a process more consistent we should encapsulate the replacement into the command.

kbinieda:

Other (image): Encapsulate image replacement into command. Closes #13217 .

### DIFF
--- a/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
+++ b/packages/ckeditor5-image/src/imageinsert/imageinsertui.js
@@ -117,6 +117,7 @@ export default class ImageInsertUI extends Plugin {
 		const dropdownView = this.dropdownView;
 		const panelView = dropdownView.panelView;
 		const imageUtils = this.editor.plugins.get( 'ImageUtils' );
+		const replaceImageSourceCommand = editor.commands.get( 'replaceImageSource' );
 
 		dropdownView.bind( 'isEnabled' ).to( command );
 
@@ -131,7 +132,7 @@ export default class ImageInsertUI extends Plugin {
 
 			if ( dropdownView.isOpen ) {
 				if ( imageUtils.isImage( selectedElement ) ) {
-					imageInsertView.imageURLInputValue = selectedElement.getAttribute( 'src' );
+					imageInsertView.imageURLInputValue = replaceImageSourceCommand.value;
 					insertButtonView.label = t( 'Update' );
 					insertImageViaUrlForm.label = t( 'Update image URL' );
 				} else {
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-image/src/imageinsert/imageinsertui.js
```diff
@@ -161,11 +161,7 @@ export default class ImageInsertUI extends Plugin {
 			const selectedElement = editor.model.document.selection.getSelectedElement();
 
 			if ( imageUtils.isImage( selectedElement ) ) {
-				editor.model.change( writer => {
-					writer.setAttribute( 'src', imageInsertView.imageURLInputValue, selectedElement );
-					writer.removeAttribute( 'srcset', selectedElement );
-					writer.removeAttribute( 'sizes', selectedElement );
-				} );
+				editor.execute( 'replaceImage', { image: selectedElement, source: imageInsertView.imageURLInputValue } );
```
scofalik, line 164: Do we actually need to pass the image here? Shouldn't the command take it from the selection?

packages/ckeditor5-image/src/image/replaceimagecommand.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Replace image command.
+ *
+ * Changes image source to the one provided. Can be executed as follows:
+ *
+ *		editor.execute( 'replaceImage', imageElement, { source: 'http://url.to.the/image' } );
+ *
+ * @extends module:core/command~Command
+ */
+export default class ReplaceImageCommand extends Command {
```
scofalik, line 21: I'd call it `ReplaceImageSource` or `ChangeImageSource`.

We are not replacing the whole image, we are replacing the source. Such name does a better job explaining what the command really does. If the command would replace image (the whole image) I'd also expect removing caption or image alt in the "new" image. But we just want to change the source.

packages/ckeditor5-image/src/image/replaceimagecommand.js
```diff
@@ -0,0 +1,38 @@
...
+	/**
+	 * Executes the command.
+	 *
+	 * @fires execute
+	 * @param {Object} options Options for the executed command.
+	 * @param {String} [options.source] The image source to replace.
+	 * @param {module:engine/model/element~Element} [options.image] The image that is affected by the replacement.
+	 */
+	execute( options ) {
+		const image = options.image || this.editor.model.document.selection.getSelectedElement();
```
scofalik, line 31: I believe the command should work like other image commands - it should take the image from the selection.

Additionally, the command should be enabled only when an image is selected, see similar commands, e.g. image resize:

```javascript
refresh() {
	const editor = this.editor;
	const imageUtils = editor.plugins.get( 'ImageUtils' );
	const element = imageUtils.getClosestSelectedImageElement( editor.model.document.selection );

	this.isEnabled = !!element;

	if ( !element || !element.hasAttribute( 'width' ) ) {
		this.value = null;
	} else {
		this.value = {
			width: element.getAttribute( 'width' ),
			height: null
		};
	}
}
```

It should also set `value` property (to the image src), like above.

We could maybe then use this value to fill the UI?

```javascript
imageInsertView.imageURLInputValue = replaceImageSourceCommand.value;
```

scofalik (changes requested):

scofalik (approved):
Please squash the changes together and it will be ready to merge.

PR #13226, opened 2023-01-08
Internal (find-and-replace): Bulk find and replace utils into plugin

### LINKED ISSUE #13225: Make plugin out of find and replace utils functions
In order to use util functions in different packages (track-changes etc.) it would be great to have an util plugin (similar as it is in document list properties).

kbinieda:

Internal (find-and-replace): Bulk find and replace utils into plugin. Closes #13225.


packages/ckeditor5-find-and-replace/src/findandreplaceutils.js
```diff
@@ -0,0 +1,45 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	static get pluginName() {
+		return 'FindAndReplaceUtils';
+	}
+
+	/**
+	 * Executes findCallback and updates search results list.
+	 */
```
scofalik, line 26: The docs were moved here only partially, missing annotations (JSDoc tags).

scofalik (changes requested):
We did analogical changes in documents lists utils, but now I am having some doubts.

First, the tests... Added only for code coverage purposes, OTOH, may do more harm than good in future (the tests don't even check if the "aliased" function is called, because it is not possible to spy/stub them).

Second, docs duplication. In documents lists, the original utils functions are marked as protected so at the very least they are not displayed in docs by default. Another idea would be to use `@see` tag.

However, after some time, I think it was a better idea to move the whole code to the plugin. It is only a bit more effort but the result, I believe will be better. Let's do this for F+R and document lists (in a separate issue). For document lists it will not even be a breaking change because the utils were protected. For F+R we can consider marking this a minor breaking change but given how niche these utils are, I'd even skip that.

scofalik (approved):
Please squash commits and rename test file to `findandreplaceutils.js` so it matches the source file and it will be ready for merge.

PR #13237, opened 2023-01-10
Fix editable focus issue

DawidKossowski:

Fix (engine): Fixed focus issue handling on Chrome which happened when nested editable was clicked.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/_src/view/observer/focusobserver.js
+++ b/packages/ckeditor5-engine/_src/view/observer/focusobserver.js
@@ -61,7 +61,22 @@ export default class FocusObserver extends DomEventObserver {
 		 * @private
 		 * @member {Number} #_renderTimeoutId
 		 */
+
+		/**
+		 * Set to `true` if the document is in the process of setting the focus.
+		 *
+		 * The flag is used to indicate that setting the focus is in progress.
+		 *
+		 * @protected
+		 * @type {Boolean} module:engine/view/observer/focusobserver#_isFocusChanging
+		 */
 	}
+
+	/**
+	 * Finishes setting the document focus state.
+	 *
+	 * @function flush
+	 */
 
 	onDomEvent( domEvent ) {
 		this.fire( domEvent.type, domEvent );
--- a/packages/ckeditor5-engine/src/view/observer/focusobserver.ts
+++ b/packages/ckeditor5-engine/src/view/observer/focusobserver.ts
@@ -25,6 +25,7 @@ import type View from '../view';
  */
 export default class FocusObserver extends DomEventObserver<'focus' | 'blur'> {
 	private _renderTimeoutId!: ReturnType<typeof setTimeout>;
+	private _isFocusChanging: boolean = false;
 
 	constructor( view: View ) {
 		super( view );
@@ -34,8 +35,15 @@ export default class FocusObserver extends DomEventObserver<'focus' | 'blur'> {
 		const document = this.document;
 
 		document.on<ViewDocumentFocusEvent>( 'focus', () => {
-			document.isFocused = true;
-			document._isFocusChanging = true;
+			/**
+			 * Set to `true` if the document is in the process of setting the focus.
+			 *
+			 * The flag is used to indicate that setting the focus is in progress.
+			 *
+			 * @internal
+			 * @type {Boolean} module:engine/view/observer/focusobserver#_isFocusChanging
+			 */
+			this._isFocusChanging = true;
 
 			// Unfortunately native `selectionchange` event is fired asynchronously.
 			// We need to wait until `SelectionObserver` handle the event and then render. Otherwise rendering will
@@ -46,7 +54,7 @@ export default class FocusObserver extends DomEventObserver<'focus' | 'blur'> {
 			// Using `view.change()` instead of `view.forceRender()` to prevent double rendering
 			// in a situation where `selectionchange` already caused selection change.
 			this._renderTimeoutId = setTimeout( () => {
-				document._isFocusChanging = false;
+				this.flush();
 				view.change( () => {} );
 			}, 50 );
 		} );
@@ -56,6 +64,7 @@ export default class FocusObserver extends DomEventObserver<'focus' | 'blur'> {
 
 			if ( selectedEditable === null || selectedEditable === data.target ) {
 				document.isFocused = false;
+				this._isFocusChanging = false;
 
 				// Re-render the document to update view elements
 				// (changing document.isFocused already marked view as changed since last rendering).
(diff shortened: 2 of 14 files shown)


packages/ckeditor5-engine/src/view/observer/focusobserver.ts
```diff
@@ -71,6 +75,16 @@ export default class FocusObserver extends DomEventObserver<'focus' | 'blur'> {
...
+	/**
+	 * Makes the document focus state as complete.
+	 */
+	public flush(): void {
+		if ( this.document._isFocusChanging ) {
+			this.document._isFocusChanging = false;
+			this.document.isFocused = true;
+		}
+	}
+
```
scofalik, line 87: I am not sure if I like the name of this function. I'd expect that calling `flush()` would perform any associated action. In this case, I might expect that the selection is moved after calling `flush()`, so basically, that the render timeout callback is called:

```javascript
this._renderTimeoutId = setTimeout( () => {
	if ( document._isFocusChanging ) {
		document._isFocusChanging = false;
		document.isFocused = true;

		view.change( () => {} );
	}
}, 50 );
```

But it might be incorrect to do it in a place where we actually want to flush (I am not sure about that though, maybe we are in a `view.change` block there anyway).

cc @niegowski

scofalik: Also, you forgot to copy the docs to `_src` directory.

scofalik (changes requested):

packages/ckeditor5-engine/src/view/observer/focusobserver.ts
niegowski: > I am not sure if I like the name of this function.

The intention of `flush()` is to mark any pending state change (focus changing -> focused) as complete. The name of this method is a mirror from `MutationObserver#flush()`. 

> I'd expect that calling `flush()` would perform any associated action. In this case, I might expect that the selection is moved after calling `flush()`, so basically, that the render timeout callback is called:

It's performing a state change. Listeners that are watching `isFocused` are triggering appropriate actions.

scofalik: > The intention of `flush()` is to mark any pending state change (focus changing -> focused) as complete.

Not only that but also to perform any action that was delayed. This is what happens in `MutationObserver` as well and how flushing delayed events work.

I'd expect that after calling `flush()` the selection is in correct place. If that's guaranteed, then I am fine. But it looks to me it is not (I might be missing something).

niegowski: > I'd expect that after calling `flush()` the selection is in correct place. If that's guaranteed, then I am fine. But it looks to me it is not (I might be missing something).

But this is a focus observer, not a selection observer. The selection observer is calling both `flush()` methods, on focus and mutation observers to make sure that state of the model is up to date to update model selection to the one provided from DOM.

scofalik: Okay, my bad, it was incorrect to say that the flush should move selection. However, still, for some reason, we call `view.change( () => {} );` in the timeout.

My point is same, after `flush()`, I'd expect the state to be exactly like after calling this timeout:

```javascript
this._renderTimeoutId = setTimeout( () => {
	if ( document._isFocusChanging ) {
		document._isFocusChanging = false;
		document.isFocused = true;

		view.change( () => {} );
	}
}, 50 );
```

niegowski: > Okay, my bad, it was incorrect to say that the flush should move selection. However, still, for some reason, we call `view.change( () => {} );` in the timeout.

Because the timeout is just a fallback if there won't be a selectionchange event following the focus change. In case of selection change, the rendering is triggered after the selectionchange event is processed and applied to the model.
 
We need to have a `document.isFocused` flag updated before the selection is updated because editable focused state is a combination of that flag and an editable element referenced by the document selection. I know that we don't have a single point of truth for the focus state (`document.isFocused` vs `selection.editableElement`) but this would probably require renderer refactoring.

scofalik: Will the editable be focused if I call:

```js
viewDocument.fire( 'focus' );
focusObserver.flush();
// Is editable focused?
```

niegowski: this depends on the current view document selection (as in the snippet that I pasted)

niegowski: and yes, the editable that is returned by `selection.editableElement` will get focused after a timeout or after a `selectionchange` event if it arrives before the timeout.

niegowski: hmm or maybe not... I'm not sure right now

scofalik: The point is -- what is the point of `FocusObserver`. I am fine with "setting `.isFocused` flag" (as opposed to making the editable focused). But in this case, why do we fire `view.change()` at all (in the timeout)?

And to be honest: I can accept the current solution as is. I just wonder why we call `view.change()` if focus observer is for setting `.isFocused` flag. And if there is a need to call `view.change()` then it should happen on `flush()` as well.

I am ignoring other parts of CKE5 code and technical challanges there. I will be fine with agreeing to: "it's messy and `view.change()` is needed there because of _something_."

niegowski: This whole discussion got me to the point where I realized that this could be simplified: [df4fe27](https://github.com/ckeditor/ckeditor5/pull/13237/commits/df4fe27c58dc7e0d3717f9d3ca8f09e012c02152)

packages/ckeditor5-engine/_src/view/observer/focusobserver.js
```diff
@@ -63,6 +63,12 @@ export default class FocusObserver extends DomEventObserver {
 		 */
 	}
 
+	/**
+	 * Makes the document focus state as complete.
```
scofalik, line 67:
```suggestion
	 * Finishes setting the document focus state.
	 *
	 * If the focus observer is waiting to set the document focus state, it will
	 * be set to `true` immediately. Otherwise, the document focus state will not change.
```

scofalik: Also change in the other place.

scofalik (changes requested):
LGTM but Kuba made some changes that, AFAIK, needs also changes in tests.

scofalik (changes requested):
There are a lot of unnecessary changes in selectionobserver.ts in new lines (added and removed).

DawidKossowski:
@scofalik IDK why, but fixed

PR #13238, opened 2023-01-10
feat: add configurable order of special characters groups

### LINKED ISSUE #13220: The list of emoji categories should be configurable

**Use-case**:

Some categories are more important than others. For instance, in a document editor, you're more likely to use mathematical symbols than emojis. In a forum post reply editor, this will be the other way around.

Currently, the list of categories is filled according to the plugin loading order. If you use the `SpecialCharactersEssentials` glue plugin, it enforces a specific order 

https://github.com/ckeditor/ckeditor5/blob/e890b2ccbc180ba2c8a252b347cae2e20133df1f/packages/ckeditor5-special-characters/src/specialcharactersessentials.js#L35-L41.

If you want to add your own category to the list, all you can do is either add it before (`init()`) or after (`afterInit()`) essentials. You cannot reorder existing categories either.

---

filipsobol:

Feature (special-characters): add configurable order of special characters groups. Closes #13220.

### DIFF
--- a/packages/ckeditor5-special-characters/src/specialcharacters.js
+++ b/packages/ckeditor5-special-characters/src/specialcharacters.js
@@ -64,6 +64,10 @@ export default class SpecialCharacters extends Plugin {
 		 * @member {Map.<String, Set.<String>>} #_groups
 		 */
 		this._groups = new Map();
+
+		editor.config.define( 'specialCharacters', {
+			order: Array.from( this.getGroups() )
+		} );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-special-characters/src/specialcharacters.js
```diff
@@ -149,12 +149,31 @@ export default class SpecialCharacters extends Plugin {
...
+
+		if ( invalidGroup ) {
+			/**
+			 * One of the special character groups in the "specialCharacters.order" configuration doesn't exist.
+			 *
+			 * @error special-character-invalid-order-group-name
+			 */
+			throw new CKEditorError(
+				`special-character-invalid-order-group-name: Group with the name "${ invalidGroup }" doesn't exist.`
+			);
```
niegowski, line 170:
```suggestion
			throw new CKEditorError( 'special-character-invalid-order-group-name', null, { invalidGroup } );
```

packages/ckeditor5-special-characters/tests/specialcharacters.js
```diff
@@ -208,36 +203,44 @@ describe( 'SpecialCharacters', () => {
...
+			plugin.addItems( 'Custom mathematical', [
 				{
 					title: ',',
 					character: 'comma'
 				}
 			] );
 
-			const groups = [ ...plugin.getGroups() ];
-			expect( groups ).to.deep.equal( [ 'Mathematical' ] );
+			const groups = Array.from( plugin.getGroups() );
```
niegowski, line 240:
```suggestion
			const groups = Array.from( plugin.getGroups() );

```

niegowski (approved):

niegowski (approved):

PR #13263, opened 2023-01-13
Other (table): Do not execute property command on table property window open

### LINKED ISSUE #13262: Don't execute table property command on opening table property window

1.  Create a table with 2+ tableCells
2.  Change background (or any property) of one tableCell
3.  Open table cell properties window of second tableCell

The command for changed property in step 2 is called (shouldn't be called before changing any value). This leads to creating unnecessary suggestions in track changes mode.

Note: Same things happen with table and table cells properties.

kbinieda:

Type: Other (table): Do not execute property command on table property window open. Closes #13262.

### DIFF
--- a/packages/ckeditor5-table/tests/tableproperties/tablepropertiesui.js
+++ b/packages/ckeditor5-table/tests/tableproperties/tablepropertiesui.js
@@ -569,6 +569,20 @@ describe( 'table properties', () => {
 			} );
 
 			describe( 'initial data', () => {
+				it( 'should not execute commands before changing the data', () => {
+					const tableBackgroundCommand = editor.commands.get( 'tableBackgroundColor' );
+					const spy = sinon.spy( tableBackgroundCommand, 'execute' );
+
+					tablePropertiesUI._showView();
+					tablePropertiesView = tablePropertiesUI.view;
+
+					expect( spy.called ).to.be.false;
+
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-table/tests/tableproperties/tablepropertiesui.js
```diff
@@ -707,6 +721,15 @@ describe( 'table properties', () => {
...
 			} );
 
+			it( 'should not reposition the baloon if view is not visible', () => {
+				const spy = sinon.spy( contextualBalloon, 'updatePosition' );
+
+				tablePropertiesUI.view = false;
+				editor.ui.fire( 'update' );
+
+				expect( spy.called ).to.be.false;
+			} );
```
scofalik, line 731: What's up with this test? :)

kbinieda: That's a coverage test. It seems that after removing code that was not connected to that case, the coverage've dropped. I've checked manually whether it should be covered and while I'm not sure because I couldn't manually get to that path, I still added some coverage test to not create a reggrestion.

scofalik: I see that there are already similar tests there (checking repositioning), so OK.

PR #13287, opened 2023-01-17
Markers should be properly downcasted and upcasted on bogus paragraphs

### LINKED ISSUE #13285: Markers are not correctly downcasted when start before document list element

1.  Create a document list that includes only "simple" items (only text).
2.  Indent one of the items in track changes mode.
3.  Call `editor.getData()`.
4.  There is no attribute/HTML element for suggestion start.
5.  When such data will be saved and loaded, the suggestion is gone (as it is incorrect markup).

https://user-images.githubusercontent.com/1502228/212714545-d7bdbd6b-8d04-4913-a413-40ddfb661934.mov

niegowski:

Fix (engine, list, table): Markers should not be lost on list items and in table cells in the data pipeline. Closes #13285.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/view/domconverter.ts
+++ b/packages/ckeditor5-engine/src/view/domconverter.ts
@@ -29,7 +29,8 @@ import {
 	indexOf,
 	getAncestors,
 	isText,
-	isComment
+	isComment,
+	first
 } from '@ckeditor/ckeditor5-utils';
 
 import type ViewNode from './node';
@@ -558,7 +559,8 @@ export default class DomConverter {
 			}
 
 			const transparentRendering = childView.is( 'element' ) &&
-				( childView.getCustomProperty( 'dataPipeline:transparentRendering' ) as boolean | undefined );
+				!!childView.getCustomProperty( 'dataPipeline:transparentRendering' ) &&
+				!first( childView.getAttributes() );
 
 			if ( transparentRendering && this.renderingMode == 'data' ) {
 				yield* this.viewChildrenToDom( childView, options );
--- a/packages/ckeditor5-clipboard/tests/dragdrop.js
+++ b/packages/ckeditor5-clipboard/tests/dragdrop.js
@@ -863,7 +863,7 @@ describe( 'Drag and Drop', () => {
 				expect( spyClipboardOutput.firstCall.firstArg.method ).to.equal( 'dragstart' );
 				expect( spyClipboardOutput.firstCall.firstArg.dataTransfer ).to.equal( dataTransferMock );
 				expect( stringifyView( spyClipboardOutput.firstCall.firstArg.content ) ).to.equal(
-					'<figure class="table"><table><tbody><tr><td>abc</td></tr></tbody></table></figure>'
+					'<figure class="table"><table><tbody><tr><td><p>abc</p></td></tr></tbody></table></figure>'
 				);
 
 				dataTransferMock.dropEffect = 'move';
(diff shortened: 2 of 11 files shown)


packages/ckeditor5-engine/src/view/domconverter.ts
```diff
@@ -558,7 +559,8 @@ export default class DomConverter {
 			}
 
 			const transparentRendering = childView.is( 'element' ) &&
-				( childView.getCustomProperty( 'dataPipeline:transparentRendering' ) as boolean | undefined );
+				( childView.getCustomProperty( 'dataPipeline:transparentRendering' ) as boolean | undefined ) &&
```
arkflpc, line 562:
```suggestion
				!!childView.getCustomProperty( 'dataPipeline:transparentRendering' ) &&
```

arkflpc: Or leave it `unknown`. It's used in boolean context only - it doesn't matter what's inside.

Mgsy:
Not a big deal, but I can't reproduce it on `release`:

### Steps to reproduce

1.  Go to http://localhost:8125/ckeditor5/external/collaboration-features/packages/ckeditor5-collaboration/tests/manual/all-features-rtc-document-list.html.
2.  Insert an empty table and put the selection inside an empty cell.
3.  Turn on Track Changes.
4.  Select _Heading 1_ from the headings dropdown to add a formatting suggestion.
5.  Call `editor.data.set( editor.getData(), { suppressErrorInCollaboration: true } )`.
6.  Reject the suggestion.
7.  Try to create an ordered list using autoformatting in the cell that had a formatting suggestion.

### Actual result

It's impossible to insert a list using autoformatting.

scofalik:
AFAICS, this happens because there is a bogus `nbsp;` added to the table cell.

`editor.getData()` returns this:

```
<figure class="table"><table><tbody><tr><td><p data-suggestion-start-before="formatBlock:698dn3bun9mf:e321952b3e500000ea901c56954d7ddcd:e168fe479815a10c23f0c08a44996fe42"><suggestion-end name="formatBlock:698dn3bun9mf:e321952b3e500000ea901c56954d7ddcd:e168fe479815a10c23f0c08a44996fe42"></suggestion-end>&nbsp;</p></td></tr></tbody></table></figure>
```

@niegowski, can we do anything relatively simple to avoid these spaces?

niegowski:
> AFAICS, this happens because there is a bogus `nbsp;` added to the table cell.
> 
> `editor.getData()` returns this:
> 
> ```
> <figure class="table"><table><tbody><tr><td><p data-suggestion-start-before="formatBlock:698dn3bun9mf:e321952b3e500000ea901c56954d7ddcd:e168fe479815a10c23f0c08a44996fe42"><suggestion-end name="formatBlock:698dn3bun9mf:e321952b3e500000ea901c56954d7ddcd:e168fe479815a10c23f0c08a44996fe42"></suggestion-end>&nbsp;</p></td></tr></tbody></table></figure>
> ```
> 
> @niegowski, can we do anything relatively simple to avoid these spaces?

This is a block filler that was not stripped because of another element in the same block (there is a marker-end element) and `DomConverter` is not able to strip this block filler because it does not know whether the marker element is a marker or any other element. We might teach `DomConverter` what element names are describing markers (by registering element names while registering a marker converter) to ignore such elements while checking block filler but I'm not sure if we could do it in such a short time window. This looks like sth out of the scope of this ticket and I would suggest extracting this to a separate issue. WDYT @scofalik

scofalik:
I thought so.

I briefly look whether this won't mess up with revision history, but it seems it is fine.

PR #13328, opened 2023-01-24
Introduce `<tableColumnGroup>` and `<tableColumn>` elements to store columns width information

### LINKED ISSUE #11479: GHS: Support for <colgroup> and <col> attributes

Currently, `<colgroup>` and `<col>` elements are introduced by the Table resize feature. However, it seems that attributes are lost, even if we allow all in GHS.

This feature request is about preserving attributes with GHS on `<colgroup>` and `<col>` elements.

---

filipsobol:

Fix (html-support): Attributes on the `<colgroup>`, and `<col>` elements should be preserved in the data. Closes #11479.

Other (table): The table columns' widths are stored as model elements instead of model element attributes. See #11479.

Fix (engine): The `elementToStructure()` downcast helper should not skip attribute changes that happened with the modification of the children list.

MINOR BREAKING CHANGE (table): The table columns' widths are stored as model elements instead of model element attribute.


niegowski (changes requested):

niegowski:
Also we should have tests for the original issue (GHS attributes on colgroup and col).

niegowski (approved):

PR #13435, opened 2023-02-08
Ck/13433 fix imports

### LINKED ISSUE #13433: [TS] Fix importing module augmentations
We use [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) for adding properties to interfaces from `ckeditor5-core` with information provided by other packages:
- `EditorConfig` - plugin-specific config is added there,
- `PluginsMap` - it enables `editor.plugins.get()` to provide specific return type based on plugin name,
- `CommandsMap` - it enables `editor.commands.get()` to provide specific return type based on command name and also checks the parameters of `editor.execute()`.

Return types are correctly provided at each call-site (`{config,plugins,commands}.get()`) only when the file with the proper `declare module '@ckeditor/ckeditor5-core'` is imported.
However, it's not always imported directly, and the results are not-consistent. Visual Studio Code usually sees those augmentations and reports no error. But, when running automatic tests, the compilation fails randomly with `TypeScript emitted no output` error.

Let's make sure that every augmentation is directly and explicitly imported:

1. Whenever there is `plugin.get()`, the type of the plugin is imported. Make sure that `PluginsMap` is augmented in the same file as plugin class.
2. Whenever there is `commands.get()`, the type of the command is imported. Make sure that  `CommandsMap` is augmented in the same file as command class.
3. Whenever there is `config.get()`, the plugin config should be imported. Ideally, the config should be in a single file like `ckeditor5-mypackage/src/mypackageconfig.ts`.

arkflpc:

Internal: Fixes imports of `declare module`. Closes #13433.

---

### Additional information


packages/ckeditor5-language/src/textpartlanguageui.ts
```diff
@@ -11,6 +11,9 @@ import { Plugin } from 'ckeditor5/src/core';
 import { Model, createDropdown, addListToDropdown, type ListDropdownItemDefinition } from 'ckeditor5/src/ui';
 import { Collection } from 'ckeditor5/src/utils';
 import { stringifyLanguageAttribute } from './utils';
+import type TextPartLanguageCommand from './textpartlanguagecommand';
+
+import './textpartlanguageconfig';
```
filipsobol, line 16: This is already imported in `packages/ckeditor5-language/src/textpartlanguageediting.ts`

arkflpc: Yes, it is.
But `textpartlanguageediting` is not imported by this file. The goal of this PR is to make sure that every file which depends on module augmentation (which happens on line 39 here) imports the config.

niegowski (approved):

niegowski (approved):
LGTM. I have one concern. Maybe we should not put legacy command aliases to `CommandsMap`? WDYT @arkflpc @Reinmar ?

PR #13440, opened 2023-02-09
Re-export some view types. Add `DOM.Iterable` to TS config.

psmyrek:

Internal (engine): Added `DowncastAddMarkerEvent` and `ViewRenderEvent` types to the package exports.

Internal: Added `DOM.Iterable` library to known type definitions in TypeScript configuration.

---

### Additional information

The `DOM.Iterable` adds `[Symbol.iterator]()` for allowed DOM types. Therefore, the following code is not reported as incorrect by TypeScript:

```ts
const editor = this.editor;
const view = editor.editing.view;
const domRoot = view.getDomRoot()!;

const classes = [
    'ck-foo',
    'ck-bar',

    // Clone all editing root classes.
    ...domRoot.classList
];
```

### DIFF
--- a/packages/ckeditor5-engine/src/index.ts
+++ b/packages/ckeditor5-engine/src/index.ts
@@ -17,6 +17,7 @@ export { default as DataController, type DataControllerSetEvent } from './contro
 export { default as Conversion } from './conversion/conversion';
 export type {
 	default as DowncastDispatcher,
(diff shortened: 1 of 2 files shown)


pomek (approved):
LGTM.

niegowski (approved):

PR #13443, opened 2023-02-09
Mentions are written on the right side of slash in RTL when inside of a non-empty paragraph

### LINKED ISSUE #2675: Overwriting dataTransfer data on 'clipboardInput' event seems to have no effect on pasting
According to the [clipboard docs](https://ckeditor.com/docs/ckeditor5/latest/api/module_clipboard_clipboard-Clipboard.html)  which shows the way to preprocess the clipboard data:

```js
this.listenTo( editor.editing.view, 'clipboardInput', ( evt, data ) => {
    const content = customTransform( data.dataTransfer.get( 'text/html' ) );
    const transformedContent = transform( content );
    data.dataTransfer.set( 'text/html', transformedContent );
} );
```

and also looking at the clipboard code: https://github.com/ckeditor/ckeditor5-clipboard/blob/c077719bd6cb4acda6301a4966d6c090940d53c2/src/clipboard.js#L145-L178

it seems the intention here is that **one can overwrite what's inside data transfer object so it will be pasted inside the editor**. However, I am unable to make this work, it seems that native data transfer does not allow changing its data in those scenarios. Check this codepen - https://codepen.io/f1ames/pen/MBNKEo?editors=1010

On Chrome data is not simply changed and on FF it throws `NoModificationAllowedError: Modifications are not allowed for this document`.

### LINKED ISSUE #13459: Mentions are written on the right side of slash in RTL when inside of a non-empty paragraph

1.  Open: http://localhost:8125/ckeditor5-mention/tests/manual/mention.html
2.  Set language to `ar`.
3.  Try to type `@`.

### ✔️ Expected result

The mention popup is on the left side of `@`:  
 

### ❌ Actual result

Here this popup is on the right side of `@`:  
 

## 📃 Other details

*   Browser: Any

illia-stv:

Fix: Mentions are written on the right side of _**@**_ in RTL when inside of a non-empty paragraph. 

Closes #13459.

---

### Additional information


packages/ckeditor5-mention/tests/mentionui.js
```diff
@@ -131,6 +131,41 @@ describe( 'MentionUI', () => {
 		} );
 	} );
 
+	describe( 'contextual balloon in rtl', () => {
```
oleq, line 134: Let's move it to an existing "position" category below and have two dedicated tests:

```
describe( 'position', () => {
    ... // existing position category tests
    describe( 'UI language direction', () => {
        it( 'should set proper position list in RTL', () => {} );
        it( 'should set proper position list in LTR', () => {} );
```

packages/ckeditor5-mention/tests/mentionui.js
```diff
@@ -131,6 +131,41 @@ describe( 'MentionUI', () => {
...
+			model.change( writer => {
+				writer.insertText( '@', doc.selection.getFirstPosition() );
+			} );
+
+			await waitForDebounce();
+		} );
+
+		afterEach( () => {
+			editor.locale.uiLanguageDirection = 'ltr';
+		} );
```
oleq, line 153: I suppose this could be removed. Every test group creates its own editors and the override should not leak.

packages/ckeditor5-mention/tests/mentionui.js
```diff
@@ -131,6 +131,41 @@ describe( 'MentionUI', () => {
...
 
+	describe( 'contextual balloon in rtl', () => {
+		let getBalloonPanelPositionDataSpy;
+
+		beforeEach( async () => {
+			await createClassicTestEditor( { ...staticConfig } );
+
+			setData( model, '<paragraph>foo []</paragraph>' );
+			editor.locale.uiLanguageDirection = 'rtl';
+			getBalloonPanelPositionDataSpy = sinon.spy( mentionUI, '_getBalloonPanelPositionData' );
```
oleq, line 142: Can we spy on `editor.plugins.get( 'ContextualBalloon' ).add()` instead? I'd rather avoid spying on private class members, they could change in an unpredictable way. Spying on the contextual balloon plugin provides a safe e2e test.

oleq (changes requested):
The solution is OK 👍 I suggested some refactoring in tests.

packages/ckeditor5-mention/tests/mentionui.js
```diff
@@ -299,6 +299,40 @@ describe( 'MentionUI', () => {
...
+						contextualBaloonSpy = sinon.spy( contextualBalloon, 'add' );
+
+						model.change( writer => {
+							writer.insertText( '@', doc.selection.getFirstPosition() );
+						} );
+					} )
+					.then( waitForDebounce );
+			} );
+
+			it( 'should set proper position list in rtl', () => {
```
oleq, line 323: We're missing another test next to this one with LTR for symmetry, as per my comment https://github.com/ckeditor/ckeditor5/pull/13443#discussion_r1105700819

oleq (changes requested):

PR #13448, opened 2023-02-10
Rewrite `ckeditor5-special-characters` to TypeScript

### LINKED ISSUE #13031: [TS] Rewrite ckeditor5-special-characters into TypeScript
Part of #11704.

filipsobol:

Other (special-characters): Rewrite `ckeditor5-special-characters` to TypeScript. Closes #13031.


packages/ckeditor5-special-characters/src/ui/charactergridview.ts
```diff
@@ -139,6 +141,8 @@ export default class CharacterGridView extends View {
 			},
 			on: {
 				mouseover: tile.bindTemplate.to( 'mouseover' ),
+				// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+				// @ts-ignore
```
filipsobol, line 145: This is still WIP

packages/ckeditor5-special-characters/src/ui/specialcharactersnavigationview.ts
```diff
@@ -7,81 +7,83 @@
...
 		const locale = this.locale;
-		const t = locale.t;
+		const t = locale!.t;
 		const dropdown = createDropdown( locale );
-		const groupDefinitions = this._getCharacterGroupListItemDefinitions( dropdown, groupNames );
-
-		dropdown.set( 'value', groupDefinitions.first.model.name );
 
-		dropdown.buttonView.bind( 'label' ).to( dropdown, 'value', value => groupNames.get( value ) );
+		dropdown.buttonView.bind( 'label' ).to( this, 'value', value => groupNames.get( value ) );
```
filipsobol, line 86: Previously, the information about currently selected character group was stored in `DropdownView#value`. However, TypeScript wasn't happy about this, because the [DropdownView](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-ui/src/dropdown/dropdownview.ts#L78) doesn't have a `value` property.

I think it makes more sense to store this information in the property of this class anyway (see line 35). Could this have some side effects I'm not aware of besides the need to update the bindings? All tests are green (after a minor update) and I haven't found any problems when manually testing this plugin. @niegowski @arkflpc

arkflpc (approved):

PR #13463, opened 2023-02-14
Move build packages to TypeScript

### LINKED ISSUE #13000: [TS] Rewrite ckeditor5-build-balloon into TypeScript
Part of #11704.

### LINKED ISSUE #13001: [TS] Rewrite ckeditor5-build-balloon-block into TypeScript
Part of #11704.

### LINKED ISSUE #13002: [TS] Rewrite ckeditor5-build-classic into TypeScript
Part of #11704.

### LINKED ISSUE #13003: [TS] Rewrite ckeditor5-build-decoupled-document into TypeScript
Part of #11704.

### LINKED ISSUE #13004: [TS] Rewrite ckeditor5-build-inline into TypeScript
Part of #11704.

filipsobol:

Other (build-balloon): Rewrite `ckeditor5-build-balloon` into Typescript. Closes #13000.

Other (build-balloon-block): Rewrite `ckeditor5-build-balloon-block` into Typescript. Closes #13001.

Other (build-classic): Rewrite `ckeditor5-build-classic` into Typescript. Closes #13002.

Other (build-decoupled-document): Rewrite `ckeditor5-build-decoupled-document` into Typescript. Closes #13003.

Other (build-inline): Rewrite `ckeditor5-build-inline` into Typescript. Closes #13004.


arkflpc (approved):

arkflpc (changes requested):
`tsconfig` files are missing.

Please verify if any changes are needed in `package.json`(s).

arkflpc:
> Please verify if any changes are needed in `package.json`(s).

CC: @pomek

pomek:
Do you think it would work at all? 

https://github.com/ckeditor/ckeditor5/blob/b14f3066fa5eb0285d22010870052b0563fdb252/packages/ckeditor5-build-balloon-block/webpack.config.js#L27

Webpack configuration should be aligned too. Also, I think the `tsconfig.json` will be required. Other than that, it should be fine. Please, let me know when R+, so I will jump here and make a few tests.

pomek:
The `_src` directory is not needed here. CKEditor 5 builds do not provide any documentation.

filipsobol:
I updated the PR and `yarn build` commands in all of these packages seem to be working fine after minor update to the webpack configuration. @arkflpc @pomek

pomek:
We may want to publish typings for builds too.

* I would redesign `tsconfig.json`:

	```json
	{
	  "extends": "../../tsconfig.release.json",
	  "compilerOptions": {
	    "sourceMap": true
	  },
	  "include": [
	    "./src/",
	    "../../typings/"
	  ],
	  "exclude": [
	    "./tests/"
	  ]
	}
	```

	Let's use the release config as a base and generate source maps as builds export them.
* Then, we need to update `package.json`:
	* Add `src/ckeditor.d.ts` to `files`.
	* Add the `types` field pointing to the `src/ckeditor.d.ts` file.

Unfortunately, I am not sure if we go the right direction.

filipsobol:
Today we were discussing how we should publish typings and how other libraries do it. This will require further research, but AFAIK the approach you described is the most popular and the one I used in the past.

Because we are already using webpack to build the package, we can use `emitDeclarationOnly` option, so that `tsc` will only output `.d.ts` files and skip the build step. See this commit for more details: 65de1ce @pomek

packages/ckeditor5-build-balloon-block/tsconfig.release.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../../tsconfig.release.json",
+  "compilerOptions": {
+    "sourceMap": true,
+    "declaration": true,
```
pomek, line 5:
```suggestion
```

I'd move control over typings to the root config.

https://github.com/ckeditor/ckeditor5/blob/bb0f347c09668e6a276dbe374f705db38a87b6b3/tsconfig.release.json#L5

Thanks to that, we need to update a single field to emit typing everywhere.

packages/ckeditor5-build-balloon-block/build/ckeditor.d.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+import BalloonEditorBase from '@ckeditor/ckeditor5-editor-balloon/src/ballooneditor';
+import '../theme/theme.css';
```
pomek, line 6: I do not understand why it lands in typings. It points to a non-existing file, as builds do not expose anything except for a single entry file.

When using the setup from my comment (https://github.com/ckeditor/ckeditor5/pull/13463#issuecomment-1433067922), everything looks good.

filipsobol: I noticed the same, but it actually points to an existing file:

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-build-balloon-block/src/ckeditor.js#L37

I will look into ways to remove this, if it's possible at all.

pomek: Right, I was checking classic build.

packages/ckeditor5-build-balloon-block/package.json
```diff
@@ -77,7 +79,7 @@
     "directory": "packages/ckeditor5-build-balloon-block"
   },
   "scripts": {
-    "build": "webpack --mode production",
+    "build": "webpack --mode production && tsc -p ./tsconfig.release.json",
```
pomek, line 82:
```suggestion
    "build": "webpack --mode production",
    "postbuild": "tsc -p ./tsconfig.release.json"
```

Let's generate typings as post-hook. I would keep the rule "a single task for a single job" as much as possible.

pomek (approved):
Please, update the merge message. We do not convert `html-support`.

LGTM.

PR #13477, opened 2023-02-15
Fixed TS errors in CK_DEBUG code

### LINKED ISSUE #13381: Typescript compiler is removing some important comments.

Some code in `*.ts` files that is marked with `CK_DEBUG` is being removed by Typescript compiler because it's treated as a comment.

### ✔️ Expected result

No lines with `CK_DEBUG` are being removed during compilation.

### ❌ Actual result

Some lines with `CK_DEBUG` are being removed during compilation.

### ❓ Possible solution
Workaround:
```
/**
 * @license 
 */

/**
 * @module 
 */

// ✅ comment followed by an used import

import usedFunc from 'foo'

// ❌ comment followed by an unused import

import unusedFunc from 'bar'

// ❌ comment followed by a type import

import type SomeType from 'sometype'

// ✅ comment followed by a class

/**
 * 
 */
class A {

	// ✅ comment at the beginning of a class 

	constructor() {

	}

	// ✅ comment in the middle of a class 

	/**
	 * doc
	 */
	public method() {
		// ✅ comment inside a method 
		while (true) {
			usedFunc();
			// ✅ comment inside a loop 
		}
	}

	// ❌ comment at the end of a class 
}

// ❌ comment followed by an interface

/**
 * 
 */
interface B {

	/**
	 * doc
	 */
	readonly x: string;
	// ❌ comment inside interface

}

// ✅ comment followed by an export;

export { A };

// ❌ comment followed by a type export;

export type { A as A2 };

// ✅ comment at the end of file
```
Comments with ❌ would be deleted during compilation, and comments with ✅ would be preserved.
All lines in `*.ts` files with `CK_DEBUG` or `CK_DEBUG_<something>` should be investigated whether they are being removed during compilation and moved to a place where they wouldn't.
The same should be done for `*.js` files as the same issue will affect them when they are rewritten to Typescript.

Actual solution:
We should think about a different approach to code that should be available only in debug mode.

## 📃 Other details

* Browser: …
* OS: …

---

przemyslaw-zan:

Internal: Fixed the TypeScript code hidden behind the `CK_DEBUG_*` flags. Closes #13381.

---

### Additional information


pomek:
We need to release https://github.com/ckeditor/ckeditor5-dev/pull/840 and update the version of dev utils everywhere.

PR #13481, opened 2023-02-16
Normalized path to classic editor sources for Windows to use slashes instead of backslashes

psmyrek:

Internal: Normalized path to classic editor sources for content styles to always use slashes instead of backslashes. Otherwise, generating content styles will end with an error on Windows.


pomek (approved):
LGTM.

pomek:
Ups, sorry :D

PR #13486, opened 2023-02-16
Other (engine, core, list): Various typescript type changes.

kbinieda:
Other (engine, core, list): Various typescript type changes.


packages/ckeditor5-engine/src/model/model.ts
```diff
@@ -689,6 +689,7 @@ export default class Model extends ObservableMixin() {
 			doNotResetEntireContent?: boolean;
 			doNotAutoparagraph?: boolean;
 			direction?: 'forward' | 'backward';
+			[ i: string ]: unknown;
```
kbinieda, line 692: Note: There is a need in CF/TC to add additional params eg. `forceDefaultExecution: true` in `deleteContent`.  Change of the contract is influenced by the change of `insertContent` where the function contract was extended.

arkflpc (approved):

PR #13516, opened 2023-02-22
Generate `.d.ts` files for all packages

### LINKED ISSUE #11722: [TS] Decide how to generate type definitions & enable testing them
Part of #11704. This is to achieve #504. See also #11720.

---

*   TODO:
    *   Agree what will we be publishing for each package (ckeditor5, ckeditor5-build-_, ckeditor5-_) on npm → Create an ADR/RFC → Verify with the entire CKE5 team.
    *   Write down how to simulate using the future packages with types.
    *   Decide on whether each package should contain its typings or we should bring them in a separate package (so other packages, like ckeditor5-react, may depend on the typings without depending on the entire CKEditor 5's codebase).
    *   Other random questions
        *   BTW: will this be `src/index.d.ts` only or multiple files?
        *   BTW: separate package with types? is that a common practice, if so, why?
        *   Check out how other similar projects publish their typings. Are there any good practices?
*   This task does not include updating the release process. This will be handled separately.
*   Some ideas: Publish to npm a beta-version? Or find a workflow on how to test it completely locally? E.g. verdaccio? or [https://github.com/wclr/yalc](https://github.com/wclr/yalc)
*   This task will unblock #13471.

filipsobol:

Internal: Generate `.d.ts` files for all packages. Closes #11722.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,5 +81,7 @@ packages/ckeditor5-watchdog/src/**/*.js
 packages/ckeditor5-widget/src/**/*.js
 packages/ckeditor5-word-count/src/**/*.js
 packages/*/src/**/*.d.ts
+packages/*/src/**/*.js.map
 src/**/*.js
+src/**/*.js.map
 src/**/*.d.ts
--- a/packages/ckeditor5-engine/tsconfig.release.json
+++ b/packages/ckeditor5-engine/tsconfig.release.json
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.release.json",
   "include": [
     "./src/",
+    "../../typings/"
   ],
   "exclude": [
     "./tests/"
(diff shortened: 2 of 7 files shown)


pomek (approved):
LGTM.

pomek:
One thing, please add the `src/*.d.ts` entry in an array below:

https://github.com/ckeditor/ckeditor5/blob/446701b6e8862aeead7fc132e8611cc5ae91ec1a/scripts/release/publish.js#L43-L52

It defines what the `ckeditor5` package will contain when releasing the CKEditor 5 project.

filipsobol:
Done :heavy\_check\_mark:

arkflpc (approved):

.gitignore
```diff
@@ -81,5 +81,7 @@ packages/ckeditor5-watchdog/src/**/*.js
 packages/ckeditor5-widget/src/**/*.js
 packages/ckeditor5-word-count/src/**/*.js
 packages/*/src/**/*.d.ts
+packages/*/src/**/*.js.map
```
pomek, line 84: Shouldn't we remove maps too? 

https://github.com/ckeditor/ckeditor5/blob/d88f9d31acbb4be6e3b10124c2c1ecf45bfed78a/scripts/release/clean.js#L39-L42

PR #13532, opened 2023-02-23
Multi-root editor implementation

### LINKED ISSUE #11493: Introduce the multi-root editor build

The multi-root editor is a perfect fit for specific use cases, however, we don't introduce an easy way to integrate it - it requires [creating the custom editor](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/ui/custom-editor-creator.html), which might be difficult, time-consuming and hard to maintain.

It would be nice to introduce a new type of build for the multi-root editor.

---

scofalik:

Feature (multi-root-editor): Introduced multi-root editor type that allows defining multiple editable areas handled by one editor instance. Closes #11493.

Feature (build-multi-root): Introduced new editor build featuring multi-root editor.

Other (core): `EditorConfig#initialData` and `EditorConfig#placeholder` can now be set to `Record<string, string>` where keys are root names and values are settings for related roots.

---

Introducing multi-root-editor editor type + a build, so two new packages.

When implementing, I run into some issues with code/class organization and types. I commented on them. Since that would require more changes that I want to do at this point, I believe that most can be left as it is. Or we may discuss it and create follow-up issues.

But one thing in particular is annoying me and it is `editable` field on `MultiRootEditorUIView` which doesn't make much sense. This is something that we can consider solving but it will require changes in other parts of the code.


docs/examples/builds/multi-root-editor.md
```diff
@@ -0,0 +1,187 @@
...
+		toolbarContainer.appendChild( editor.ui.view.toolbar.element );
+
+		// Make toolbar sticky when the editor is focused.
+		editor.ui.focusTracker.on( 'change:isFocused', () => {
+			if ( editor.ui.focusTracker.isFocused ) {
+				toolbarContainer.classList.add( 'sticky' );
+			} else {
+				toolbarContainer.classList.remove( 'sticky' );
+			}
+		} );
```
scofalik, line 61: I didn't make the toolbar sticky the way it is  done in the classic editor. It is possible and is not that difficult but it would lead to another decisions to be made. E.g. make it always sticky and by default, like in classic, or allow controling it? What if it doesn't make sense in some integrations?

Since `position: sticky` is widely supported, I decided to go this route.

If we enable sticky toolbar in multi-root editor, it should be also enabled for decopled editor, I think, to be consistent.

If it is done like in `ClassicEditor`, it seems it is possible to disable the stickiness through `StickyPanelView#isActive`. So, we can have it configurable. One thing I didn't like, though, is that you have to append `StickyPanelView` element, instead of `ToolbarView` element to the DOM. But maybe you would need toolbar instance for something. So, we may end up with `editor.ui.toolbarPanel` and `editor.ui.toolbar`. Additional confusion. Maybe `editor.ui.attachToolbarToDom( domContainer, isSticky )` would be a good idea.

cc @oleq.

docs/examples/builds/multi-root-editor.md
```diff
@@ -0,0 +1,187 @@
...
+	Wrapping the structure inside a pair of
+	contenteditable="true" + contenteditable="false" elements
+	is required to provide proper caret handling when
+	using arrow keys at the start and end of an editable area.
+	
+	You can skip them if you don't want to move the
+	caret between editable areas using arrow keys.
+!--> 
+<div contenteditable="true">
+	<div contenteditable="false">
```
scofalik, line 86: This doesn't look good but it is necessary to have proper arrows control. But maybe you know better solutions? By adding this, we don't have to introduce logic that would be fired when arrow key is pressed at the beginning or at the end of an editable. The browser handles it by itself.

However, it may add some quirkiness...? OTOH, we use similar structures this inside the editor (e.g. tables).

scofalik: cc @Reinmar

docs/examples/builds/multi-root-editor.md
```diff
@@ -0,0 +1,187 @@
...
+		}
+	);
+```
+</details>
+
+<details>
+<summary>Specify root name when obtaining the data</summary>
+
+```js
+	editor.getData( { rootName: 'leftSide' } ); // -> '<p>Content for left-side box.</p>'
```
scofalik, line 183: As you can see I decided not to change the API for getting the data at this point.

docs/examples/framework/multi-root-editor.md
```diff
@@ -1,481 +0,0 @@
----
```
scofalik, line 1: The guide was moved from framework to builds. cc @Mgsy

docs/framework/guides/custom-editor-creator.md
```diff
@@ -1,460 +0,0 @@
----
```
scofalik, line 1: I decided to remove this guide. It is useful as it describes how to handle `Editor` classes and how to create your own custom editor implementation. But the guide was describing the multi-root editor and it is redundant at this point, might even be confusing. Since I don't have an idea for a different editor (nor time to think on it or write a new guide), I removed this one.

Also think that multi-root editor will solve a lot of custom needs.

scofalik: cc @apadol-cksource @Mgsy

packages/ckeditor5-build-multi-root/README.md
```diff
@@ -0,0 +1,69 @@
+CKEditor 5 multi-root editor build
+==============================================
+
+[![npm version](https://badge.fury.io/js/%40ckeditor%2Fckeditor5-build-multi-root.svg)](https://www.npmjs.com/package/@ckeditor/ckeditor5-build-multi-root)
+[![Coverage Status](https://coveralls.io/repos/github/ckeditor/ckeditor5/badge.svg?branch=master)](https://coveralls.io/github/ckeditor/ckeditor5?branch=master)
+[![Build Status](https://travis-ci.com/ckeditor/ckeditor5.svg?branch=master)](https://app.travis-ci.com/github/ckeditor/ckeditor5)
+![Dependency Status](https://img.shields.io/librariesio/release/npm/@ckeditor/ckeditor5-build-multi-root)
+
+The multi-root editor build for CKEditor 5. Read more about the [multi-root editor build](https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/predefined-builds.html#multi-root-editor) and see the [demo](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/multi-root-editor.html).
```
scofalik, line 9: AFAIK, this misses an image. Can we do it at a later point?

packages/ckeditor5-build-multi-root/tests/ckeditor.js
```diff
@@ -0,0 +1,99 @@
+/**
```
scofalik, line 1: In general, I believe that tests for this package are redundant and unnecessary. Manual test for "cjs version" should be enough. But I added them, as other packages do the same.

packages/ckeditor5-core/src/editor/editorconfig.ts
```diff
@@ -432,9 +432,9 @@ import type Editor from './editor';
 export interface EditorConfig {
 	context?: Context;
 	extraPlugins?: Array<PluginConstructor<Editor>>;
-	initialData?: string;
+	initialData?: string | Record<string, string>; // @TODO doc
 	language?: string | LanguageConfig;
-	placeholder?: string;
+	placeholder?: string | Record<string, string>; // @TODO doc
```
scofalik, line 437: I forgot about that, didn't look for TODOs in other packages :fearful:.

packages/ckeditor5-core/src/editor/editorconfig.ts
```diff
@@ -432,9 +483,9 @@ import type Editor from './editor';
 export interface EditorConfig {
 	context?: Context;
 	extraPlugins?: Array<PluginConstructor<Editor>>;
-	initialData?: string;
+	initialData?: string | Record<string, string>;
 	language?: string | LanguageConfig;
-	placeholder?: string;
+	placeholder?: string | Record<string, string>;
```
scofalik, line 488: This is a change that I needed to make to keep the official multi-root editor API the same as what was shown in the guide (and people use) and to be consistent with the data format for multi-root editors that we already have, e.g. in CS.

Other than that, `DataController#set` already accepts data in this format.

However, this broadens the type, which prompted other changes. I am not sure about docs -- now we need to keep two snippet examples and explain it.

Maybe a better solution would be to have new architecture for editor configs:

*   `EditorConfig` -- without things that are different for multi-root editor and single-root editor,
*   `SingleRootEditorConfig` -- adds `initialData: string` and `placeholder: string`,
*   `MultiRootEditorConfig` -- adds `initialData: Record<string, string>` and same for `placeholder`.

I am not sure if this will not be even more confusing or problematic, though. For example, places that use `EditorConfig` now will have to know which type of config it is. OTOH, `initialData` and `placeholder` aren't used in many places in the code base. Maybe they are encapsulated inside the editor-related classes and maybe there will be no problem with recognizing what type of config interface is passed.

packages/ckeditor5-core/src/editor/utils/dataapimixin.ts
```diff
@@ -19,7 +19,7 @@ import type { Constructor, Mixed } from '@ckeditor/ckeditor5-utils';
  */
 export default function DataApiMixin<Base extends Constructor<Editor>>( base: Base ): Mixed<Base, DataApi> {
 	abstract class Mixin extends base implements DataApi {
-		public setData( data: string ): void {
+		public setData( data: string | Record<string, string> ): void {
```
scofalik, line 22: This goes in line with `EditorConfig#initialData`. Alternatively, `MultiRootEditor` may not extend `DataApiMixin` and implement these methods itself. I don't remember if that caused any problems (whether editor instance is used as `DataApi`).

But in this case, this is more safe, as we broaden the input parameter, and, as I said, `data.set()` already handles this type too.

packages/ckeditor5-core/src/editor/utils/securesourceelement.ts
```diff
@@ -18,16 +18,10 @@ import type { ElementApi } from './elementapimixin';
...
  */
-export default function secureSourceElement( editor: Editor & ElementApi ): void {
-	const sourceElement: any = editor.sourceElement;
-
-	// If the editor was initialized without specifying an element, we don't need to secure anything.
-	if ( !sourceElement ) {
-		return;
-	}
-
+export default function secureSourceElement( editor: Editor, sourceElement: HTMLElement & { ckeditorInstance?: Editor } ): void {
```
scofalik, line 24: I decided to change this implementation because I don't see a reason why not to pass the `sourceElement` as a parameter. AFAIR, all `secureSourceElement` uses were straight after setting `editor.sourceElement`.

For multi-root, though, there's no `editor.sourceElement` -- there can be multiple. So, the new implementation fits the multi-root editor better.

packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
```diff
@@ -0,0 +1,341 @@
...
+ * {@glink installation/getting-started/predefined-builds Builds} are ready-to-use editors with plugins bundled in.
+ *
+ * When using the editor from source you need to take care of loading all plugins by yourself
+ * (through the {@link module:core/editor/editorconfig~EditorConfig#plugins `config.plugins`} option).
+ * Using the editor from source gives much better flexibility and allows for easier customization.
+ *
+ * Read more about initializing the editor from source or as a build in
+ * {@link module:editor-multi-root/multirooteditor~MultiRootEditor.create `MultiRootEditor.create()`}.
+ */
+export default class MultiRootEditor extends DataApiMixin( Editor ) {
```
scofalik, line 58: I decided not to extend `ElementApiMixin`.

packages/ckeditor5-editor-multi-root/src/multirooteditoruiview.ts
```diff
@@ -0,0 +1,107 @@
...
+	 * The main toolbar of the multi-root editor UI.
+	 */
+	public readonly toolbar: ToolbarView;
+
+	/**
+	 * Editable elements used by the multi-root editor UI.
+	 */
+	public readonly editables: Record<string, InlineEditableUIView>;
+
+	public readonly editable: InlineEditableUIView;
```
scofalik, line 34: This property doesn't make much sense for multi-root, where you have multiple editables, as above. This is set to the first defined editable.

I tried not extending `EditorUIView` but this got me into other typing problems (i.e. `EditorUI` requires `EditorUIView`, while `Editor` requires `EditorUI`, etc.).

Another idea would be to remove `editable` from `EditorUIView` and have it defined in particular `EditorUIView` implementations. This would require changes in places that uses `editable`. Alternatively, `editable` could be allowed to be `null`, then in places that requires it, we would have to tell TS that `editable` is set. AFAIR, `editable` is used in the balloon editor.

Of course, yet another idea would be to introduce hierarchy like I proposed for editor configs, so have separate ui view class for single and multi roots.

But I think that `editable` actually may not be needed at all. First of all, we have `getEditableElement()` on `EditorUI` that supports multiple editables (multi-roots). We could have `getEditable()` as well. On `EditorUI` or `EditorUIView`. We could also introduce `getFocusedEditable()` if there's a need to get the focused editable specifically (doesn't matter for single-editable editors, might matter for multi-root).

I do not think that we need to solve `EditorConfig` issues/changes; but this one is something that we might want to solve. `editor.ui.editable` does not make sense for multi-root, it smells. If you have proposals or strong opinion, I listen.

In general, I think there's quite a mess when it comes to which property is on what class and there are multiple ways to access some objects. E.g. we have `sourceElement`, we have `editor.ui.view.editable.element` we have `getEditableElement()` etc. I also don't understand why some things are on `EditorUI` and why other are on `EditorUIView`. However, this is not in a scope of this PR.

godai78 (approved):
LGTM

godai78:
Oi, `master`, of course, merge `master`. Duh.

arkflpc (commented):
Please write `ckeditor5-build-multi-root` in TypeScript. See #13463 For inspiration.

scofalik:
@arkflpc - done.

niegowski (changes requested):
Partial review.

packages/ckeditor5-ui/docs/framework/deep-dive/focus-tracking.md
```diff
@@ -28,8 +28,8 @@ And when many places accept focus, there must be some systems out there to disco
...
 In the following chapters of this guide, we will explain how these systems work, what makes CKEditor "focusable", and how to develop new editor features so they fit into the existing focus management systems and patterns:
 
-- [**The first section**](#focus-in-the-editor-engine) explains how the editor {@link framework/architecture/editing-engine engine} manages focus and what tools exist there to help you out when developing new features.
-- In [**the second section**](#focus-in-the-editor-ui), we will show you how the {@link framework/architecture/ui-library user interface} of the editor tracks focus and how various UI components take advantage of that, for instance, to provide accessibility.
+- [**The first section**](#focus-in-the-editor-engine) explains how the editor {@link framework/guides/architecture/editing-engine engine} manages focus and what tools exist there to help you out when developing new features.
+- In [**the second section**](#focus-in-the-editor-ui), we will show you how the {@link framework/guides/architecture/ui-library user interface} of the editor tracks focus and how various UI components take advantage of that, for instance, to provide accessibility.
```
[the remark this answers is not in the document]
scofalik: This is because in the meantime CCX (IIRC) changed the guides structure. I will fix it.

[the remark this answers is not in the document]
godai78: Just remove all the `/guides` from framework URL. This is a recent change (like, last week) but can be solved with find&replace easily.

packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
```diff
@@ -0,0 +1,340 @@
...
+	 * **Note**: The multi-root editor does not remove the toolbar and editable when destroyed. You can
+	 * do that yourself in the destruction chain, if you need to:
+	 *
+	 * ```ts
+	 * editor.destroy().then( () => {
+	 * 	// Remove the toolbar from DOM.
+	 * 	editor.ui.view.toolbar.element.remove();
+	 *
+	 * 	// Remove the editable from DOM.
+	 * 	editor.ui.view.editable.element.remove();
```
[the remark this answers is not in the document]
scofalik: Good catch.

packages/ckeditor5-editor-multi-root/src/multirooteditorui.ts
```diff
@@ -0,0 +1,186 @@
...
+	/**
+	 * Enable the placeholder text on the editing roots, if any was configured.
+	 */
+	private _initPlaceholder(): void {
+		const editor = this.editor;
+		const editingView = editor.editing.view;
+
+		for ( const editable of Object.values( this.view.editables ) ) {
+			const editingRoot = editingView.document.getRoot( editable.name! )!;
+			const placeholder = editor.config.get( 'placeholder' ) as Record<string, string>;
```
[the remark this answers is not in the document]
scofalik: This could be actually useful, if you just want one, same placeholder for all editables. +1.

packages/ckeditor5-editor-multi-root/src/multirooteditoruiview.ts
```diff
@@ -0,0 +1,107 @@
...
+	 * @param options.shouldToolbarGroupWhenFull When set to `true` enables automatic items grouping
+	 * in the main {@link module:editor-multi-root/multirooteditoruiview~MultiRootEditorUIView#toolbar toolbar}.
+	 * See {@link module:ui/toolbar/toolbarview~ToolbarOptions#shouldGroupWhenFull} to learn more.
+	 */
+	constructor(
+		locale: Locale,
+		editingView: View,
+		editableNames: Array<string>,
+		options: {
+			editableElements?: Record<string, HTMLElement>;
```
niegowski, line 56: There are 2 params (`editableNames` and `options.editableElements`) that could be merged into one (non optional) parameter (with optional values in the record). To make it consistent with [`InlineEditorUIView`](https://github.com/ckeditor/ckeditor5/blob/c8a89f383efcd7d6369012039b2db6409d419467/packages/ckeditor5-editor-inline/src/inlineeditoruiview.ts#L137-L144) I would suggest using top level param (not in the options).

packages/ckeditor5-editor-multi-root/src/multirooteditoruiview.ts
niegowski: Did you checked how `SourceEditing` works with the current solution? I can see the button is active (but should not) but nothing happens when clicked. This looks like related to `editor.ui.view.editable`.

packages/ckeditor5-editor-multi-root/src/multirooteditoruiview.ts
scofalik: This was my initial approach. After a while I came up with this which was easier to use and to describe in docs. It also makes sense to me, when compared to other `EditorUIView`s where `editableElement` is optional and passed in options.

niegowski: :+1:

packages/ckeditor5-editor-multi-root/src/multirooteditoruiview.ts
scofalik: This PR is not supposed to make multi-root editor work with all the features, there will be follow-ups later.

I didn't check how source editing worked with earlier multi-roots but AFAICS in the code, it should work the same as earlier.

scofalik:
@niegowski I fixed everything according to your suggestions, except where we had discussions.

packages/ckeditor5-editor-inline/src/inlineeditorui.ts
```diff
@@ -155,17 +155,20 @@ export default class InlineEditorUI extends EditorUI {
 		const editingRoot = editingView.document.getRoot()!;
 		const sourceElement = ( editor as Editor & ElementApi ).sourceElement;
 
-		const placeholderText = editor.config.get( 'placeholder' ) as string ||
-			sourceElement && sourceElement.tagName.toLowerCase() === 'textarea' && sourceElement.getAttribute( 'placeholder' );
```
niegowski, line 159: Why you removed support for placeholder from `textarea` attribute?

scofalik:
These tests are removed because these editors do not support being created on a `<textarea>` element. The code is never fired except of a theoretical use case in the test. It brings confusion and makes the code messy and may give you false idea that something is possible.

scofalik:
@niegowski Fixed your suggestions + CI looks green.

niegowski (approved):

docs/examples/builds/multi-root-editor.md
oleq: The reason why we went with our custom sticky panel implementation was to prevent the toolbar from staying above the editing root if you scroll past it like this:

```

 ┌───────────────────────────────────┐
 │                                   │
 │                                   │
 │     Editing root                  │
 │                                   │
 │                                   │
 └───────────────────────────────────┘


 ┌───────────────────────────────────┐
 │      Toolbar                      │
 └───────────────────────────────────┘

```

but keep it at a line-height distance from the bottom instead:

```

 ┌───────────────────────────────────┐
 │                                   │
 │                                   │
 │     Editing root                  │
 ├───────────────────────────────────┤
 │      Toolbar                      │
 ├───────────────────────────────────┤
 └───────────────────────────────────┘
```

If it wasn't for that, we'd go with pure CSS approach too. So I think you chose the right direction, at least for this demo. There are unlimited possibilities when it comes to the geometry of the multi-root editor, so hardcoding anything more would be an overkill IMO.

PR #13554, opened 2023-02-24
Add missing config imports to fix CI

filipsobol:

Internal: Add missing config imports to fix CI

### DIFF
--- a/packages/ckeditor5-image/src/image.ts
+++ b/packages/ckeditor5-image/src/image.ts
@@ -12,6 +12,7 @@ import ImageBlock from './imageblock';
 import ImageInline from './imageinline';
 
 import '../theme/image.css';
+import './imageconfig';
 
 /**
  * The image plugin.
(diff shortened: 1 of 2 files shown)


pomek:
We need to apply the same change for the `DocumentList` plugin. Also, I would update the merge message. We do not want to fix CI. We want to support creating custom builds using TypeScript as a source (assuming CKEditor 5 comes from npm).

Reinmar:
Since this PR does not resolve any ticket, a few more words about what issue we're fixing here are needed. Think about a scenario that PK lands in this PR 2 years later and try to figure out what the heck was changed here and why.

pomek:
The following merge message would describe what kind of problem we are trying to resolve:

Fix (image): Load the plugin configuration when using the `Image` plugin imported directly (`@ckeditor/ckeditor5-image/src/image`) instead of using the package entry point. When creating a custom build in a TypeScript project, the editor config didn't resolve the `image` configuration object, resulting in an error TS2322 (property does not exist).

Fix (list): Load the plugin configuration when using the `ListProperties` plugin imported directly (`@ckeditor/ckeditor5-list/src/listproperties`) instead of using the package entry point. When creating a custom build in a TypeScript project, the editor config didn't resolve the `list` configuration object, resulting in an error TS2322 (property does not exist).

Reinmar:
Personally I'm fine with documenting the story in the PR itself. In this case changelog entries could even be marked as "Internal" – I don't think anyone will understand what we're talking about anyway. It's more for us than people outside at this point.

Issue #13565, opened 2023-02-27
Improving module augmentation


filipsobol:
We need to improve how we deal with module augmentation ([explained in more details here](https://github.com/ckeditor/ckeditor5/issues/13433)), to:

*   ensure that module augmentation kicks in when importing individual plugins from packages,
*   all types are available in tests,
*   avoiding arbitrary `import './config'` imports.

To encourage best practices and ensure that the `.d.ts` file with module augmentation is always imported, we will require importing all package contents from the main `index` file instead from individual files in the built package, eg.:

```
// ✔️
import { Table } from '@ckeditor/ckeditor5-table';

// ❌
import Table from '@ckeditor/ckeditor5-table/src/table';
```

filipsobol:
Checklist for migrating packages to the new structure:

*   [ ] Copy `tsconfig.json` and `tsconfig.release.json` from `table` plugin.
*   [ ] Create `src/augmentation.ts` file.
*   [ ] Search for all `declare module '@ckeditor/ckeditor5-core'` declarations and move them to the `src/augmentation.ts` as done in [this file](https://github.com/ckeditor/ckeditor5/pull/13568/files#diff-d5a4789418e06b3d342d850f83c33c3abed982ab24fd88a14059b72e69f68979).
*   [ ] Remove all `import '../xxxconfig'` imports.
*   [ ] Add `import './augmentation';` at the very bottom of the main `index.ts`.
*   [ ] Replace **all imports in the monorepo** referencing individual files in the `src` of the package to use the package name instead. See example and regex below.

Example:

```typescript
// BEFORE
import Table from '@ckeditor/ckeditor5-table/src/table';

// AFTER
import { Table } from '@ckeditor/ckeditor5-table';
```

  
Regex (**commit your changes before trying this**):

```
Search for: import (.*) from '<PACKAGE_NAME>/src/(.*)';
Replace with: import { $1 } from '<PACKAGE_NAME>';

⚠️ Make sure that all files imported this way are actually exported from the main `index.ts` file.
```

filipsobol:
Below is the summary of the changes introduced in PR #13568 to fix module augmentation and tests within monorepo. The same changes (following the checklist above) will be applied to all packages using module augmentation.

### Module augmentation not being loaded in all cases

[...]

PR #13568, opened 2023-02-27
Improve module augmentation in `table` plugin

filipsobol:

Internal: Improve TypeScript configuration in the project and module augmentation in `table` plugin. Related to #13565.


packages/ckeditor5-table/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
-  "extends": "ckeditor5/tsconfig.json",
+  "extends": "../../tsconfig.json",
```
niegowski, line 2: Why such change?

packages/ckeditor5-table/plugin.d.ts
```diff
@@ -0,0 +1,136 @@
+/**
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import type {
```
niegowski, line 6: Why this is `d.ts`? Maybe this could be a standard `ts` file?

niegowski (changes requested):

niegowski:
I'm also concerned about automatic tests for other packages. Those import other features to test integrations and they probably do not import using `index` files. Those could fail if augmentations are not provided.

packages/ckeditor5-table/plugin.d.ts
```diff
@@ -0,0 +1,136 @@
...
+		[ PlainTableOutput.pluginName ]: PlainTableOutput;
+	}
+
+  interface CommandsMap {
+		insertTableColumnLeft: InsertColumnCommand;
+		insertTableColumnRight: InsertColumnCommand;
+    insertTableRowAbove: InsertRowCommand;
+		insertTableRowBelow: InsertRowCommand;
+    insertTable: InsertTableCommand;
+    mergeTableCellRight: MergeCellCommand;
```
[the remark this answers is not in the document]
filipsobol: Fixed

packages/ckeditor5-table/plugin.d.ts
```diff
@@ -0,0 +1,136 @@
...
+    [ TableMouse.pluginName ]: TableMouse;
+    [ TableProperties.pluginName ]: TableProperties;
+		[ TablePropertiesEditing.pluginName ]: TablePropertiesEditing;
+		[ TablePropertiesUI.pluginName ]: TablePropertiesUI;
+    [ TableSelection.pluginName ]: TableSelection;
+    [ TableToolbar.pluginName ]: TableToolbar;
+    [ TableUI.pluginName ]: TableUI;
+    [ TableUtils.pluginName ]: TableUtils;
+		[ PlainTableOutput.pluginName ]: PlainTableOutput;
+	}
```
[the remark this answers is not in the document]
filipsobol: Fixed

packages/ckeditor5-table/src/index.ts
```diff
@@ -29,3 +29,36 @@ export { default as TableSelection } from './tableselection';
 export { default as TableUtils } from './tableutils';
 export { default as TableColumnResize } from './tablecolumnresize';
 export { default as TableColumnResizeEditing } from './tablecolumnresize/tablecolumnresizeediting';
+
+export type { TableConfig } from './tableconfig';
+export type { default as InsertColumnCommand } from './commands/insertcolumncommand';
```
Reinmar, line 34: Why only export types? AFAIK we usually export entire commands. They are a public API of a feature (both, as registered in commands registry and classes that can be reused).

packages/ckeditor5-table/src/plugin.ts
```diff
@@ -0,0 +1,139 @@
+/**
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import type {
```
Reinmar, line 6: This looks like a duplication of index.ts. Do we really need that?

packages/ckeditor5-table/src/plugin.ts
```diff
@@ -0,0 +1,139 @@
+/**
```
Reinmar, line 1: As I commented in the PR thread, I'd consider renaming this to `package.ts` as it contains kind of metadata for a package. But then I'd merge the `fooconfig.ts` file into this one too. Why having configs in one place and plugin and command maps in another?

Reinmar:
My biggest doubt regarding this is: why do we have `plugin.d.ts` and `index.ts`? Why not merging them into one file? And keeping there all the module augmentation too?

packages/ckeditor5-table/tsconfig.json
filipsobol: To make it consistent with `tsconfig.release.json` in the same package.

packages/ckeditor5-table/src/plugin.ts
filipsobol: On today's daily, the team agreed to keep module augmentation in a separate `augmentation.ts` file.

packages/ckeditor5-table/src/plugin.ts
filipsobol: On today's daily, the team agreed to use the `augmentation.ts` file name and to store config in a separate file.

packages/ckeditor5-table/src/index.ts
Reinmar: F2F discussion: It turned out that we didn't expose any commands so far in index.js files. I thought we did expose some already. Hence, let's continue not exposing anything for now (except types if needed).

Reinmar:
Note: Squash all commits in this branch before the merge so to not pollute the history with all the test import changes.

filipsobol:
Adding `"./packages/*/src/augmentation.ts"` to `include` array in `tsconfig.json` didn't work when running `testautomated.js`. However, it worked when we added `"./packages/ckeditor5-table/src/augmentation.ts"` to `files`.

When debugging this with @niegowski and comparing behavior of `tsc` and `testautomated.js`, we found that `tsc` worked with both configurations. After we enabled the `logLevel: 'info'` option in `ts-loader` we noticed that it loads `tsconfig.js` for `paragraph` plugin, despite testing `clipboard` plugin in root of the monorepo project. When we added `"./packages/*/src/augmentation.ts"` to `include` array in `./packages/ckeditor5-paragraph/tsconfig.json`, it started working as expected for `clipboard`.

We need to investigate why seemingly random `tsconfig.json` is being loaded by `ts-loader`. For now, as a quick workaround, we will add `"./packages/*/src/augmentation.ts"`to `include` arrays in all packages.

filipsobol:
Above issue with the wrong \`tsconfig.json\` being loaded will be fixed in ckeditor/ckeditor5-dev#847

Reinmar:
Note to self: I wanted to comment that you changed formatting of JSON files and I remember that we have a special rule for them in `.editorconfig` to use two spaces. But then I checked it's for `package.json`s only :D And we actually use tabs in other JSON files. Case closed :D

niegowski (approved):

niegowski:
Please update the "Suggested merge commit message" to address changes in this PR.

tsconfig.test.json
```diff
@@ -0,0 +1,8 @@
+{
+	"extends": "./tsconfig.json",
+	"include": [
+		"./typings/",
+		"./packages/*/src/augmentation.ts",
+		"./external/*/packages/*/src/augmentation.ts"
```
pomek, line 6:
```suggestion
		"./external/ckeditor5-internal/packages/*/src/**/*.ts",
		"./external/collaboration-features/packages/*/src/**/*.ts",
```

To match https://github.com/ckeditor/ckeditor5/blob/0e5ddd74e5684897d2d2c87a0e65150c287433d4/tsconfig.docs.json#L5-L6. People can have more than two repositories in `external/` and it should. not be loaded.

Btw, we use spaces in all JSON files.

filipsobol: I updated it to:  
 

```
"./external/ckeditor5-internal/packages/*/src/augmentation.ts",
"./external/collaboration-features/packages/*/src/augmentation.ts"
```

scripts/check-manual-tests.sh
```diff
@@ -26,6 +27,11 @@ then
   MANUAL_TEST_SERVER_OPTIONS="$MANUAL_TEST_SERVER_OPTIONS -i $IDENTITY_FILE"
 fi
 
+if [ ! -z "$TS_CONFIG" ]
+then
+  MANUAL_TEST_SERVER_OPTIONS="$MANUAL_TEST_SERVER_OPTIONS --tsconfig $TS_CONFIG"
+fi
```
pomek, line 33: I would add `--tsconfig ./tsconfig.test.json` to L43 and remote support for the `-t` flag. This script is used only for verification on CI and when releasing stuff. We do not have customize it.

pomek (approved):
LGTM. But TBH, too many changes. I will depend on CI results.

PR #13576, opened 2023-02-28
Other (engine, core): Changes in TS typings

kbinieda:
Minor changes in typings needed for CF revision history TS migration.

`documentfragment.markers` - Should this field be `readonly`? While the change in revision history can be fixed using pure TS changes, I'm not sure why it is readonly. From what I can see, it is not possible to pass markers via constructor so setting markers for document fragment (assuming we just want to create document fragment without attaching in to the model) seems difficult.

### DIFF
--- a/packages/ckeditor5-engine/src/index.ts
+++ b/packages/ckeditor5-engine/src/index.ts
@@ -62,6 +62,9 @@ export { default as HtmlDataProcessor } from './dataprocessor/htmldataprocessor'
 // Model / Operation.
 export type { default as Operation } from './model/operation/operation';
 export { default as InsertOperation } from './model/operation/insertoperation';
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-engine/src/model/documentfragment.ts
```diff
@@ -35,7 +35,7 @@ export default class DocumentFragment extends TypeCheckable implements Iterable<
 	 * which will be set as Markers to {@link module:engine/model/model~Model#markers model markers collection}
 	 * when DocumentFragment will be inserted to the document.
 	 */
-	public readonly markers: Map<string, Range> = new Map();
+	public markers: Map<string, Range> = new Map();
```
scofalik, line 38: Can we refactor the part of the code that requires this change, so that this change is not needed anymore?

PR #13604, opened 2023-03-03
check-maunal-tests should not build DLL

pomek:

Internal: The `check-manual-tests` should not produce DLL builds. An integrator should provide them before calling the checker script.


pomek:
Part of CF#5025. Do not merge.

PR #13615, opened 2023-03-06
Fix `ts-loader` configs used in scripts

filipsobol:

Internal: Fix `ts-loader` configs used in scripts. Related to #13565.


scripts/dll/webpack.config.dll.js
```diff
@@ -143,7 +143,14 @@ const webpackConfig = {
 			},
 			{
 				test: /\.ts$/,
-				use: [ 'ts-loader' ]
+				use: [
+					{
+						loader: 'ts-loader',
+						options: {
+							configFile: path.resolve( process.cwd(), 'tsconfig.test.json' )
```
niegowski, line 150: Are you sure that this is the correct file?

scripts/docs/utils.js
```diff
@@ -139,3 +144,28 @@ function writeFile( filePath, data ) {
...
+
+/**
+ * Adds ts-loader with proper configuration to the passed configuration object
+ * only if the `tsconfig.test.json` file exists.
+ *
+ * @param {Object} webpackConfig
+ * @returns {Void}
+ */
+function addTypeScriptLoader( webpackConfig ) {
+	const tsconfigPath = path.resolve( process.cwd(), 'tsconfig.test.json' );
```
niegowski, line 156: Are you sure that this is the correct file? Building production snippets in docs should use `tsconfig.test.json` file?

pomek: Production snippets will not use it (L158). I would rename this configuration from `test` to `dev`. It's a development configuration. Used by us when developing things and checking them by running tests or building docs.

scripts/dll/webpack.config.dll.js
niegowski: This is used to build production dlls so I would suggest to use a separate tsconfig file for this. Maybe `tsconfig.dll.json`? @pomek WDYT?

pomek: https://github.com/ckeditor/ckeditor5/pull/13615#discussion_r1126915268

pomek (approved):
So far, it looks good. I will take a few tests tomorrow before marking it as R+.

scripts/dll/webpack.config.dll.js
niegowski: @pomek comment that you mentioned does not apply to this case. This script is building base DLL and this is not checking whether file exists or not (as in the other comment). What I want to say is that we should not use `dev` tsconfig for production DLL build just to avoid any future WTFs when change in `dev` would affect production DLLs.

pomek: Production DLL is built from development sources, so TBH, it seems to be safe.

Webpack and TypeScript configurations do not land on npm. If somebody wants to refresh the build, it must be done using the `ckeditor5` repository representing our development environment. Perhaps I'm missing something, and I would ask what can go wrong here.

pomek: Btw, what would the `tsconfig.dll` file contain?

packages/ckeditor5-adapter-ckfinder/webpack.config.js
```diff
@@ -7,12 +7,14 @@
...
 const { builds } = require( '@ckeditor/ckeditor5-dev-utils' );
 const webpack = require( 'webpack' );
 
 module.exports = builds.getDllPluginWebpackConfig( webpack, {
 	themePath: require.resolve( '@ckeditor/ckeditor5-theme-lark' ),
 	packagePath: __dirname,
 	manifestPath: require.resolve( 'ckeditor5/build/ckeditor5-dll.manifest.json' ),
-	isDevelopmentMode: process.argv.includes( '--mode=development' )
+	isDevelopmentMode: process.argv.includes( '--mode=development' ),
+	tsconfigPath: path.resolve( __dirname, '..', '..', 'tsconfig.dll.json' )
```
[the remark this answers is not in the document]
niegowski:
```suggestion
	tsconfigPath: require.resolve( 'ckeditor5/tsconfig.dll.json' )
```

niegowski (approved):

pomek:
image

PR #13666, opened 2023-03-13
Remove ?? and ?. operators.

### LINKED ISSUE #13656: Remove Nullish coalescing and safe navigation operators.
Nullish coalescing operator (`??`) and safe navigation (`?.`) operators are nice features, but cause few issues:

- When the TypeScript target is >= ES2020 – they are translated to JavaScript as they are, but the output cannot be parsed by webpack 4.x.x without additional loaders.
- When the TypeScript target is <= ES2019 – they are translated to conditional (ternary) operator with checks for `null` and `undefined` values. Because the result checks both values in separate branches (`x === null || x === void 0`), the code coverage drops unless we test **both** `null` and `undefined` cases.

Because we only have a few places when we use those operators, let's remove them.

arkflpc:

Internal: Remove Nullish coalescing and safe navigation operators. Closes #13656.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-easy-image/src/cloudservicesuploadadapter.ts
+++ b/packages/ckeditor5-easy-image/src/cloudservicesuploadadapter.ts
@@ -85,6 +85,6 @@ class Adapter implements UploadAdapter {
 	}
 
 	public abort() {
-		this.fileUploader?.abort();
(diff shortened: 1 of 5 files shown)


pomek (approved):
LGTM. Let's wait for CI.

PR #13667, opened 2023-03-13
Allow for adding and detaching model roots after the editor was initialized

### LINKED ISSUE #13388: Allow for adding and removing roots from multi-root editor after it was initialized

Currently, the roots' structure for multi-root is static, i.e., new roots cannot be added after the editor was initialized. Similarly, already exisiting roots cannot be removed.

This makes multi-roots very limiting in how they can be used.

Some important notes:

*   This solution should work in real-time collaboration. If a remote client creates a root, all other connected clients should be appropriately updated.
*   Adding / removing roots should work well with undo. In perfect scenario, you would be able to decide if creating a root is undoable or not, as the expected result may vary from application to application.

---

Related ticket / duplicates: #12235, #6291

scofalik:

Feature (engine): Introduced `model.Writer#addRoot()` and `model.Writer#detachRoot()`. Model roots can now be dynamically added or detached from the document. All content gets removed from a detached root, and new content cannot be inserted, as long as the root is detached. Closes #13388.

Feature (engine): Introduced `Differ.getChangedRoots()` which returns roots added or detached since last differ reset. `Differ#isEmpty` and `Differ#hasDataChanges()` will return `true` if a root was added or detached.

Other: Introduced `Observer#stopObserving()` to allow for proper removal of DOM editable elements from the editor.

Other (ui): Introduced `EditorUI#removeEditableElement()`.

Other (engine): `model.Document#getRootNames()` now returns only attached roots by default. `includeDetached` parameter was introduced to enable returning detached roots as well.

Other (engine): Introduced `RootOperation`.

Other (engine): `model.DocumentFragment#isAttached` was introduced for compatibility reasons. It always returns `false`.

Internal (multi-root-editor): Introduced API which allows for adding and removing roots and editable areas.

### DIFF
--- a/packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
+++ b/packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
@@ -24,6 +24,7 @@ import MultiRootEditorUI from './multirooteditorui';
 import MultiRootEditorUIView from './multirooteditoruiview';
 
 import { isElement as _isElement } from 'lodash-es';
+import { type RootElement, type Writer } from 'ckeditor5/src/engine';
 
 /**
  * The {@glink installation/getting-started/predefined-builds#multi-root-editor multi-root editor} implementation.
@@ -64,7 +65,7 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
 	/**
 	 * The elements on which the editor has been initialized.
 	 */
-	public readonly sourceElements: Record<string, HTMLElement> | undefined;
+	public readonly sourceElements: Record<string, HTMLElement>;
 
 	/**
 	 * Creates an instance of the multi-root editor.
@@ -91,6 +92,8 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
 
 		if ( !sourceIsData ) {
 			this.sourceElements = sourceElementsOrData as Record<string, HTMLElement>;
+		} else {
+			this.sourceElements = {};
 		}
 
 		if ( this.config.get( 'initialData' ) === undefined ) {
@@ -123,6 +126,22 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
 		const view = new MultiRootEditorUIView( this.locale, this.editing.view, rootNames, options );
 
 		this.ui = new MultiRootEditorUI( this, view );
+
+		// This can be improved if `$isVirtual` attribute for root is introduced.
+		// Multi-root editor will create (or not) the editable based on the attribute value. The events will be editable-related.
+		this.model.document.on( 'change:data', () => {
+			const changedRoots = this.model.document.differ.getChangedRoots();
+
+			for ( const [ rootName, isAttached ] of changedRoots ) {
+				const root = this.model.document.getRoot( rootName )!;
+
+				if ( isAttached ) {
+					this.fire<AddRootEvent>( 'addRoot', root );
+				} else {
+					this.fire<DetachRootEvent>( 'detachRoot', root );
+				}
+			}
+		} );
 	}
 
 	/**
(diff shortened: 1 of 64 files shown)


packages/ckeditor5-engine/src/model/document.ts
```diff
@@ -213,20 +247,29 @@ export default class Document extends EmitterMixin() {
...
 	public getRoot( name: string = 'main' ): RootElement | null {
 		return this.roots.get( name );
 	}
 
 	/**
-	 * Returns an array with names of all roots (without the {@link #graveyard}) added to the document.
+	 * Returns an array with names of all roots added to the document (except the {@link #graveyard graveyard root}).
+	 *
+	 * Detached roots **are not** returned by this method by default. This is to make sure that all features or algorithms that operate
+	 * on the document data know which roots are still a part of the document and should be processed.
```
scofalik, line 264: Alternatively, we can refrain from changing this method and return all roots. Then, it will be less confusing vs `getRoot()` (which probably has to stay as it is) but it will be additional effort to make sure that all parts of the code that currently use `getRootNames()` will filter out detached roots if needed.

Another idea to avoid confusion would be to not return detached roots in `getRoot()`, however, then, some parts of the code (not sure how many) will have to use `document.roots.get()` directly. For sure it would have to happen when `Position` instance is created, as you may have positions inside detached roots.

I think what I proposed is the safest, although surely a bit confusing.

packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
```diff
@@ -123,6 +126,22 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
...
+			for ( const [ rootName, isAttached ] of changedRoots ) {
+				const root = this.model.document.getRoot( rootName )!;
+
+				if ( isAttached ) {
+					this.fire<AddRootEvent>( 'addRoot', root );
+				} else {
+					this.fire<DetachRootEvent>( 'detachRoot', root );
+				}
+			}
+		} );
```
scofalik, line 144: Next step of multi-root project will include work on roots attributes. I would like to introduce an attribute that would be handled only by the multi-root editor. If the attribute is set, multi-root editor will automatically create or remove editable instance if such root was not virtual (the usual case). After the change is introduced, multi-root editor will fire `addEditable` and `removeEditable` events instead and the integrator will only need to append/remove the DOM element. It will be easier to understand.

scofalik:
One more thing that I do not like is that you need to specify callbacks for `addRoot` and `detachRoot` but I am afraid there's no way around it. It would be easier to just call `const domElement = editor.addRoot();` and append the DOM element to the webpage but there are two scenarios, where roots are created "outside" of the API.

The first scenario is RTC, when roots can be created/removed on remote clients.

The second scenario is undo feature, when a root is created/removed performing undo or redo.

So, the integration could be simpler but only if you do not use RTC and if you make your root changes undoable.

packages/ckeditor5-engine/src/model/operation/rootoperation.ts
```diff
@@ -0,0 +1,154 @@
...
+		// These operations will require root element instance (in operation property or in position instance).
+		// If the root is not created ahead of time, instantiating such operations may fail.
+		if ( !this._document.getRoot( this.rootName ) ) {
+			const root = this._document.createRoot( this.elementName, this.rootName );
+
+			root._isAttached = false;
+		}
+	}
+
+	public override get type(): 'addRoot' | 'removeRoot' {
```
niegowski, line 74: Missing doclets for methods.

packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
```diff
@@ -174,6 +193,181 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
...
+	 * 	rowId++;
+	 * } );
+	 * ```
+	 *
+	 * @param rootName Name of the root to add.
+	 * @param isUndoable Whether creating the root can be undone (using the undo feature) or not.
+	 * @param data Initial data for the root.
+	 * @param elementName Element name for the root element in the model. It can be used to set different schema rules for different roots.
+	 */
+	public addRoot( rootName: string, isUndoable = false, data = '', elementName = '$root' ): void {
```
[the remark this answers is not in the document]
scofalik: Good idea. I actually wanted to do this in the next step (roots meta data) as there will be yet another parameter here.

packages/ckeditor5-engine/src/model/operation/rootoperation.ts
scofalik: I wasn't sure if we actually need to add these if they are just

```
/**
 * @inheritDoc
 */
```

scofalik:
I resolved your suggestions.

packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
```diff
@@ -251,11 +251,11 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
 	 * ```
 	 *
 	 * @param rootName Name of the root to add.
-	 * @param isUndoable Whether creating the root can be undone (using the undo feature) or not.
 	 * @param data Initial data for the root.
 	 * @param elementName Element name for the root element in the model. It can be used to set different schema rules for different roots.
+	 * @param isUndoable Whether creating the root can be undone (using the undo feature) or not.
 	 */
-	public addRoot( rootName: string, isUndoable = false, data = '', elementName = '$root' ): void {
+	public addRoot( { rootName, data, elementName }: RootData, isUndoable: boolean ): void {
```
niegowski, line 258: I'm not sure if this would not confuse integrators, maybe something like this would better suit our needs especially since only `rootName` is a required parameter.

```suggestion
	public addRoot( rootName, { data, elementName, isUndoable } = {} ): void {
```

Btw you should also update examples in docs.

scofalik: > Btw you should also update examples in docs.

Good point.

> maybe something like this

My proposal focuses on providing root data in one object and non-data flags (behaviors) (`isUndoable` for now, may become an object in next PR) separately.

If anything, I'd go with `rootName, RootData, isUndoable` but i would not move `isUndoable` to the data part. I also make an error, `isUndoable` should still be optional.

scofalik: Okay, I played with it a bit and you are right, I will follow your suggestion.

scofalik:
@niegowski ready for final review.

packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
```diff
@@ -255,7 +253,7 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
 	 * @param elementName Element name for the root element in the model. It can be used to set different schema rules for different roots.
 	 * @param isUndoable Whether creating the root can be undone (using the undo feature) or not.
 	 */
-	public addRoot( { rootName, data, elementName }: RootData, isUndoable: boolean ): void {
+	public addRoot( rootName: string, { data = '', elementName = '$root', isUndoable = false } = {} ): void {
```
niegowski, line 256: Is the doclet above building correctly while generating API docs?

packages/ckeditor5-engine/tests/view/observer/arrowkeysobserver.js
```diff
@@ -96,4 +96,10 @@ describe( 'ArrowKeysObserver', () => {
 			observer.observe();
 		} ).to.not.throw();
 	} );
+
+	it( 'should implement empty #stopOvserving() method', () => {
```
niegowski, line 100:
```suggestion
	it( 'should implement empty #stopObserving() method', () => {
```

packages/ckeditor5-engine/tests/view/observer/fakeselectionobserver.js
```diff
@@ -148,6 +148,12 @@ describe( 'FakeSelectionObserver', () => {
 		sinon.assert.notCalled( spy );
 	} );
 
+	it( 'should implement empty #stopOvserving() method', () => {
```
niegowski, line 151:
```suggestion
	it( 'should implement empty #stopObserving() method', () => {
```

packages/ckeditor5-engine/tests/view/observer/tabobserver.js
```diff
@@ -96,4 +96,10 @@ describe( 'TabObserver', () => {
 			sinon.assert.notCalled( tabSpy );
 		} );
 	} );
+
+	it( 'should implement empty #stopOvserving() method', () => {
```
niegowski, line 100:
```suggestion
	it( 'should implement empty #stopObserving() method', () => {
```

packages/ckeditor5-enter/tests/enterobserver.js
```diff
@@ -130,4 +130,10 @@ describe( 'EnterObserver', () => {
 
 		expect( interceptedEventInfo.stop.called ).to.be.undefined;
 	} );
+
+	it( 'should implement empty #stopOvserving() method', () => {
```
niegowski, line 134:
```suggestion
	it( 'should implement empty #stopObserving() method', () => {
```

packages/ckeditor5-typing/tests/deleteobserver.js
```diff
@@ -532,6 +532,12 @@ describe( 'Delete', () => {
 				} );
 			} );
 		} );
+
+		it( 'should implement empty #stopOvserving() method', () => {
```
niegowski, line 536:
```suggestion
		it( 'should implement empty #stopObserving() method', () => {
```

packages/ckeditor5-typing/tests/inserttextobserver.js
```diff
@@ -243,4 +243,10 @@ describe( 'InsertTextObserver', () => {
 			expect( firstCallArgs.selection.isEqual( viewSelection ) ).to.be.true;
 		} );
 	} );
+
+	it( 'should implement empty #stopOvserving() method', () => {
```
niegowski, line 247:
```suggestion
	it( 'should implement empty #stopObserving() method', () => {
```

packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
scofalik: It didn't build well. In docs, the options was `__namedparams` and it didn't have description. I've added a type.

scofalik:
@niegowski Please review.

Inviz:
Great work

PR #13678, opened 2023-03-14
Fix generated `.d.ts` in builds to include types of loaded plugins

### LINKED ISSUE #13676: Fix `.d.ts` in `ckeditor5-build-*` packages to include types for loaded plugins
Currently, the `.d.ts` files in `ckeditor5-build-*` packages don't contain types for loaded plugins:

```ts
import ClassicEditorBase from '@ckeditor/ckeditor5-editor-classic/src/classiceditor';
export default class ClassicEditor extends ClassicEditorBase {
}
```

Because there are no imports from any plugins as [in the source code](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-build-classic/src/ckeditor.ts), the module augmentation doesn't work properly in consuming projects.

filipsobol:

Fix (build-*): Typings from the CKEditor 5 builds contain types of loaded plugins. Closes #13676.


niegowski (approved):

pomek:
I would improve the merge message a bit:

> Fix (build-*): Typings from the CKEditor 5 builds contain types of loaded plugins.

packages/ckeditor5-build-balloon-block/build/ckeditor.d.ts
```diff
@@ -3,9 +3,43 @@
...
+import { MediaEmbed } from '@ckeditor/ckeditor5-media-embed';
+import { Paragraph } from '@ckeditor/ckeditor5-paragraph';
+import { PasteFromOffice } from '@ckeditor/ckeditor5-paste-from-office';
+import { Table, TableToolbar } from '@ckeditor/ckeditor5-table';
+import { TextTransformation } from '@ckeditor/ckeditor5-typing';
+import { CloudServices } from '@ckeditor/ckeditor5-cloud-services';
 /**
  * @internal
  */
 import '../theme/theme.css';
```
pomek, line 29: We wanted to drop it from typings.

packages/ckeditor5-build-balloon-block/src/ckeditor.ts
```diff
@@ -32,82 +32,80 @@ import { CloudServices } from '@ckeditor/ckeditor5-cloud-services';
  */
 import '../theme/theme.css';
 
-export default class BalloonEditor extends BalloonEditorBase {}
+export default class BalloonEditor extends BalloonEditorBase {
+	public static override builtinPlugins = [
```
arkflpc, line 36: Does the trick work because `builtinPlugins` have a tuple type with all plugin types?

packages/ckeditor5-build-balloon-block/build/ckeditor.d.ts
filipsobol: Done

packages/ckeditor5-build-balloon-block/src/ckeditor.ts
filipsobol: Yes, tuple uses types imported from plugin packages and this gives a TypeScript a chance to load module augmentations with them.

pomek (approved):
LGTM.

arkflpc (approved):

Issue #13541, opened 2023-02-24
[TS] Migrate Angular integration to use generated typings


arkflpc:
Part of #13471.

PR #13697, opened 2023-03-16
Add generic type for watchdog data.

mremiszewski:

Feature (watchdog): Add a generic type for watchdog data. See #13541.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-watchdog/src/contextwatchdog.ts
+++ b/packages/ckeditor5-watchdog/src/contextwatchdog.ts
@@ -24,7 +24,7 @@ import type {
 } from 'ckeditor5/src/utils';
 
 import Watchdog, { type WatchdogConfig, type WatchdogState } from './watchdog';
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-watchdog/src/contextwatchdog.ts
```diff
@@ -194,10 +197,15 @@ export default class ContextWatchdog<TContext extends Context = Context> extends
...
+	public getItem( itemId: string ): unknown | undefined {
+		try {
+			const watchdog = this._getWatchdog( itemId );
 
-		return watchdog._item;
+			return watchdog._item;
+		}
+		catch {
+			return undefined;
+		}
```
mremiszewski, line 208: I'm not sure if this change is valid, but docs say it should return `undefined` if not found but it was throwing error instead. If it's valid change, then tests should be adjusted.

PR #13702, opened 2023-03-17
Link doesn't show a change in the link

### LINKED ISSUE #12588: When adding hyperlink without link text, editing the link doesn’t show a change in the link.

1. Click 'link' from the toolbar and type https://www.google.co.in and enter
2. Then click on link to reedit and enter https://about.google/
3. Editing the link doesn’t show a change in the link.

### ✔️ Expected result

On editing the link it should be changed.

### ❌ Actual result

but on editing the link doesn’t show a change in the link.
Screenshot 2022-10-06 at 12 32 36 PM

Browser: Google Chrome latest.
OS: macOS Monterey v12.3.1 

illia-stv:

Fix: Link doesn't show a change in the link. Closes #12588.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-link/src/linkcommand.ts
+++ b/packages/ckeditor5-link/src/linkcommand.ts
@@ -166,10 +166,6 @@ export default class LinkCommand extends Command {
 
 					if ( selection.getAttribute( 'linkHref' ) === linkText ) {
 						linkRange = this._updateLinkContent( model, writer, linkRange, href );
-						writer.setSelection( linkRange.end );
-					} else {
-						// Put the selection at the end of the updated link.
-						writer.setSelection( writer.createPositionAfter( linkRange.end.nodeBefore! ) );
 					}
 
 					writer.setAttribute( 'linkHref', href, linkRange );
@@ -181,6 +177,9 @@ export default class LinkCommand extends Command {
 					falsyManualDecorators.forEach( item => {
 						writer.removeAttribute( item, linkRange );
 					} );
+
+					// Put the selection at the end of the updated link.
+					writer.setSelection( writer.createPositionAfter( linkRange.end.nodeBefore! ) );
 				}
 				// If not then insert text node with `linkHref` attribute in place of caret.
 				// However, since selection is collapsed, attribute value will be used as data for text node.
--- a/packages/ckeditor5-link/src/linkcommand.ts
+++ b/packages/ckeditor5-link/src/linkcommand.ts
@@ -10,7 +10,7 @@
 import { Command } from 'ckeditor5/src/core';
 import { findAttributeRange } from 'ckeditor5/src/typing';
 import { Collection, first, toMap } from 'ckeditor5/src/utils';
-import type { Range } from 'ckeditor5/src/engine';
+import type { Range, DocumentSelection, Model, Writer, Position } from 'ckeditor5/src/engine';
 
 import AutomaticDecorators from './utils/automaticdecorators';
 import { isLinkableElement } from './utils';
@@ -27,8 +27,6 @@ export default class LinkCommand extends Command {
 	 * @readonly
 	 */
 	declare public value: string | undefined;
-
-	declare private _isBind: boolean | null;
 
 	/**
 	 * A collection of {@link module:link/utils/manualdecorator~ManualDecorator manual decorators}
@@ -143,15 +141,7 @@ export default class LinkCommand extends Command {
 	public override execute( href: string, manualDecoratorIds: Record<string, boolean> = {} ): void {
 		const model = this.editor.model;
 		const selection = model.document.selection;
-		const range = selection.getFirstRange( );
-		const linkText = getTextFromRange( range );
-
-		if ( linkText === href && this._isBind === null ) {
-			this._isBind = true;
-		} else {
-			this._isBind = false;
-		}
-
+		const linkText = extractTextFromSelection( selection );
 		// Stores information about manual decorators to turn them on/off when command is applied.
 		const truthyManualDecorators: Array<string> = [];
 		const falsyManualDecorators: Array<string> = [];
@@ -172,7 +162,15 @@ export default class LinkCommand extends Command {
 				// When selection is inside text with `linkHref` attribute.
 				if ( selection.hasAttribute( 'linkHref' ) ) {
 					// Then update `linkHref` value.
-					const linkRange = findAttributeRange( position, 'linkHref', selection.getAttribute( 'linkHref' ), model );
+					let linkRange = findAttributeRange( position, 'linkHref', selection.getAttribute( 'linkHref' ), model );
+
+					if ( selection.getAttribute( 'linkHref' ) === linkText ) {
+						linkRange = this._updateLinkContent( model, writer, linkRange, href );
+						writer.setSelection( linkRange.end );
+					} else {
+						// Put the selection at the end of the updated link.
+						writer.setSelection( writer.createPositionAfter( linkRange.end.nodeBefore! ) );
+					}
 
 					writer.setAttribute( 'linkHref', href, linkRange );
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -277,4 +288,37 @@ export default class LinkCommand extends Command {
 
 		return true;
 	}
+
+	private _updateLinkContent( model: Model, writer: Writer, range: Range, href: string ): Range {
+		const attributes = toMap( {} );
+
+		attributes.set( 'linkHref', href );
+
+		const text = writer.createText( href, attributes );
```
oleq, line 297: How about just `writer.createText( href, { linkHref: href } )`?

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -277,4 +288,37 @@ export default class LinkCommand extends Command {
 
 		return true;
 	}
+
+	private _updateLinkContent( model: Model, writer: Writer, range: Range, href: string ): Range {
```
oleq, line 292: Missing API docs.

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -277,4 +288,37 @@ export default class LinkCommand extends Command {
...
+		attributes.set( 'linkHref', href );
+
+		const text = writer.createText( href, attributes );
+
+		return model.insertContent( text, range );
+	}
+}
+
+function getTextFromRange( range: Range | null ): string {
+	if ( !range || range && !Array.from( range.getItems() ).length ) {
```
oleq, line 304: You could simplify this one

```
const firstNode = first( range.getItems() );

if( !range || range && !firstNode ) { 
	... 
}
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -277,4 +288,37 @@ export default class LinkCommand extends Command {
...
+
+		attributes.set( 'linkHref', href );
+
+		const text = writer.createText( href, attributes );
+
+		return model.insertContent( text, range );
+	}
+}
+
+function getTextFromRange( range: Range | null ): string {
```
oleq, line 303: Missing API docs.

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -277,4 +288,37 @@ export default class LinkCommand extends Command {
...
+		return '';
+	}
+
+	const firstNode = Array.from( range.getItems() )[ 0 ];
+
+	if ( firstNode.is( '$text' ) || firstNode.is( '$textProxy' ) ) {
+		return firstNode.data;
+	}
+
+	return '';
```
oleq, line 314: I'd return `null` in this case. If there's no range or the range is empty or it has no text nodes, there's no text inside (semantics).

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -142,6 +141,7 @@ export default class LinkCommand extends Command {
 	public override execute( href: string, manualDecoratorIds: Record<string, boolean> = {} ): void {
 		const model = this.editor.model;
 		const selection = model.document.selection;
+		const linkText = extractTextFromSelection( selection );
```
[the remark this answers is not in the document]
oleq: This should also simplify the internal logic of the helper. For instance, you can assume there's always some range in the selection.

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -227,14 +231,21 @@ export default class LinkCommand extends Command {
 				}
 
 				for ( const range of rangesToUpdate ) {
-					writer.setAttribute( 'linkHref', href, range );
+					let linkRange = range;
+
+					if ( selection.getAttribute( 'linkHref' ) === linkText ) {
```
oleq, line 236: In the text extraction helper you're checking the text under the first position of the first range in the selection. 
Here, the code is iterating over multiple possible ranges that can accept the link. 
And on top of that, you're obtaining `linkHref`'s value from selection's attribute (which is the attribute value under the selection's first position only).

Check out this use-case:

```js
editor.setData( '<p><a href="foo">foo</a></p><p>bar</p><p>baz</p>' )

editor.model.change( writer => writer.setSelection( editor.model.document.getRoot(), 'in' ) );

editor.execute( 'link', 'foo' );

console.log( editor.getData() );
```

You'd expect the 2nd and 3rd paragraphs to keep their text after linking but their text gets replaced by "foo".

oleq (changes requested):
Some bits of code can be simplified. Applying links to the non-collapsed selection needs work.

packages/ckeditor5-link/src/linkcommand.ts
illia-stv: but how we could get items from range until we didn't check is range null or not?
image

illia-stv:
> * In the text extraction helper you're checking the text under the first position of the first range in the selection.
> * Here, the code is iterating over multiple possible ranges that can accept the link.
> * And on top of that, you're obtaining `linkHref`'s value from selection's attribute (which is the attribute value under the selection's first position only).
> * Result: there will be bugs :P
> 
> Check out this use-case:
> 
> ```js
> editor.setData( '<p><a href="foo">foo</a></p><p>bar</p><p>baz</p>' )
> 
> editor.model.change( writer => writer.setSelection( editor.model.document.getRoot(), 'in' ) );
> 
> editor.execute( 'link', 'foo' );
> 
> console.log( editor.getData() );
> ```
> 
> You'd expect the 2nd and 3rd paragraphs to keep their text after linking but their text gets replaced by "foo".



Regarding this comment, I suggest the next solution. We could just not modify the link even if it is equal to href when the selection is on more than one element. After applying the link's command selection should be the same, but if we would replace the link it brokes the selection and selection.**focus** will be set at the end of inserted content. so we also should handle this case. Then I propose simplifying everything and updating the link's content only in collapsed selection and if the range is only on one text element.

packages/ckeditor5-link/src/linkcommand.ts
oleq: TS is meant to help us but it's not all-knowing. We know that the model document selection in the editor has always at least one range and one position. There's no scenario when document selection has no ranges. We can use this to `!` certain things and simplify code.

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -277,4 +291,46 @@ export default class LinkCommand extends Command {
...
+	return null;
+}
+
+function extractTextFromSelection( selection: DocumentSelection ): string | null {
+	const range = selection.getFirstRange( );
+	const firstPosition = selection.getFirstPosition();
+	const textFromPosition = firstPosition !== null && firstPosition.textNode && firstPosition.textNode.data || null;
+	const textFromRange = getTextFromRange( range );
+
+	return selection.isCollapsed ? textFromPosition : textFromRange;
```
oleq, line 335: If `selection.isCollapsed === true`, you don't have to calculate `textFromRange`. If `false`, you don't have to calculate `textFromPosition`. It's a waste of time and degrades performance.

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -277,4 +291,46 @@ export default class LinkCommand extends Command {
...
+		return firstNode.data;
+	}
+
+	return null;
+}
+
+function extractTextFromSelection( selection: DocumentSelection ): string | null {
+	const range = selection.getFirstRange( );
+	const firstPosition = selection.getFirstPosition();
+	const textFromPosition = firstPosition !== null && firstPosition.textNode && firstPosition.textNode.data || null;
```
oleq, line 332: You can safely assume `firstPosition`[ always exists](https://github.com/ckeditor/ckeditor5/pull/13702/files#r1143111217) and use `const textFromPosition = firstPosition!.textNode && firstPosition!.textNode.data`.

packages/ckeditor5-link/src/linkcommand.ts
oleq: This code could be simplified to

```js
function getTextFromRange( range: Range ): string | null {
	if ( range.isCollapsed ) {
        return null;
    }

	const firstNode = first( range.getItems() );

	if ( firstNode.is( '$text' ) || firstNode.is( '$textProxy' ) ) {
		return firstNode.data;
	}

	return null;
}
```

because:
1. you can safely assume there always will be a range, 
2. you don't have to engage `TreeWalker` to know there are no items in the range. A collapsed range has no items.

oleq (approved):
Waiting for the CI.

PR #13708, opened 2023-03-17
Expose Context, EditorWatchdog, and ContextWatchdog as static editor properties

### LINKED ISSUE #13707: Expose `Context`, `EditorWatchdog`, and `ContextWatchdog` as static editor properties.

To simplify integrations and consistency with editor builds, we should expose `Context`, `EditorWatchdog`, and `ContextWatchdog` as static editor properties. Expose them on editors (not on the base `Editor` class) so we won't have a dependency from the `core` package to `watchdog`.

niegowski:

Other (editor-*): Exposed `Context`, `EditorWatchdog`, and `ContextWatchdog` as static editor properties. Closes #13707.

MINOR BREAKING CHANGE: The `Watchdog` package is no longer available as a standalone [DLL build](https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/alternative-setups/dll-builds.html). It is now included in the main `ckeditor5-dll.js` build. See #13707.

---

### Additional information


packages/ckeditor5-editor-balloon/src/ballooneditor.ts
```diff
@@ -237,6 +240,27 @@ export default class BalloonEditor extends DataApiMixin( ElementApiMixin( Editor
...
+	 * The {@link module:core/context~Context} class.
+	 *
+	 * Exposed as static editor field for easier access in editor builds.
+	 */
+	public static Context = Context;
+
+	/**
+	 * The {@link module:watchdog/editorwatchdog~EditorWatchdog} class.
+	 *
+	 *  Exposed as static editor field for easier access in editor builds.
```
Reinmar, line 254:
```suggestion
	 * Exposed as static editor field for easier access in editor builds.
```

packages/ckeditor5-editor-balloon/src/ballooneditor.ts
```diff
@@ -237,6 +240,27 @@ export default class BalloonEditor extends DataApiMixin( ElementApiMixin( Editor
...
+	 */
+	public static EditorWatchdog = EditorWatchdog;
+
+	/**
+	 * The {@link module:watchdog/contextwatchdog~ContextWatchdog} class.
+	 *
+	 *  Exposed as static editor field for easier access in editor builds.
```
Reinmar, line 261:
```suggestion
	 * Exposed as static editor field for easier access in editor builds.
```

Reinmar:
@pomek I think we should consider adding the watchdog to the core DLL. It's 600SLOC, so not much. And if we expect to use it as a dependency of editor packages, it'd be cleaner if it was imported there via `ckeditor5` instead of becoming a new dependency of those editor packages.

Looking at the arch diagram, watchdog would land for me in the "feature utils", if we agree that an editor creator is a "feature".



PS. As a followup we'd need to update the ESLint rule to expect watchdog to be

pomek (approved):
LGTM.

pomek:
We talked F2F, and there are my proposals:

* Keep the watchdog DLL build (even empty) not to break existing integrations (removing the build results in error 404).
* Add (manually) a `console.warn()` into the build that we will remove in the next major update (`38.0.0`). Also, we can link to an issue where we describe what to do.

arkflpc (approved):

PR #13713, opened 2023-03-20
Color picker implementation

### LINKED ISSUE #2946: Nested lists support
It's mostly ready but requires couple of fixes.

The UX aspect of the feature is described in https://github.com/ckeditor/ckeditor5-list/issues/5.

Dumluregn:

Internal (ui): Add color picker component.

---

Closes: https://github.com/cksource/ckeditor5-internal/issues/2946

### Additional information

Things we should handle in this PR:

- [x] Fixing TS errors
image
- [x] Color picker should be at the bottom of dropdown.
- [x] Unit tests.
- [x] Opt-out config.
- [x] Editor with and without color picker in the manual test.
- [x] Color picker should apply color in the configured format (HSL by default).
- [x] API docs.

Issues that can be extracted:
- [ ] Setting color should be debounced somehow. E.g. triggered only on `mouseup` event. Still, what to do during input typing? Maybe add a small commit button in the same line (input can be narrow).
- [ ] General styling.
- [ ] A11y (keyboard navigation, focus tracking).
- [ ] Maybe selection color should be reflected in the color picker?
- [ ] Maybe color picker input should have some basic placeholder?
- [ ] Maybe picker should reset after dropdown closure?


packages/ckeditor5-font/src/fontconfig.ts
```diff
@@ -153,6 +153,14 @@ export interface FontColorConfig {
 	 * ```
 	 */
 	columns?: number;
+
+	/**
+	 * Color picker gives ability to choose custom color.
+	 *
+	 * By default color picker is set on true.
```
mlewand, line 160:
```suggestion
	 * By default color picker is set on `true`.
```

packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -139,10 +139,18 @@ export default class ColorUI extends Plugin {
...
+			dropdownView.on( 'colorChange', ( evt, data ) => {
+				editor.execute( this.commandName, data );
+			} );
+
 			dropdownView.on( 'change:isOpen', ( evt, name, isVisible ) => {
 				// Grids rendering is deferred (#6192).
 				dropdownView.colorTableView!.appendGrids();
 
+				if ( editor.config.get( this.componentName )!.colorPicker !== false ) {
+					dropdownView.colorTableView!.appendColorPicker();
```
mlewand, line 151: This should be internally done by the `ColorTableView` class, in the constructor.

Since the constructor is [already getting plenty of options](https://github.com/ckeditor/ckeditor5/blob/87f5544a878511d4fabc7faf9108a4b3c0a07472/packages/ckeditor5-font/src/ui/colortableview.ts#L147), you should pass color picker configuration the same way.

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -71,6 +72,12 @@ export default class ColorTableView extends View {
 	 */
 	public documentColors: DocumentColorCollection;
 
+	/**
+	 * Color picker.
+	 *
+	 */
+	public _colorPickerView?: ColorPickerView;
```
mlewand, line 79: Your name indicates a private/protected variable while you marked it as a public, so what is it going to be? 😉

packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -139,10 +139,18 @@ export default class ColorUI extends Plugin {
 				editor.editing.view.focus();
 			} );
 
+			dropdownView.on( 'colorChange', ( evt, data ) => {
```
mlewand, line 142: `colorChange` event on color dropdown is very generic. Note that it should also be fired when selecting regular colors, while it doesn't.

Given that `_colorPickerView` member will become a public property it's better to just listen here on `dropdownView.colorPickerView` directly (with a check if it exists, because it doesn't have to).

packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -0,0 +1,136 @@
...
+		this._debounceInputEvent = debounce( ( color: string ) => {
+			this.fire( 'change', { value: color } );
+			this.setColor( color );
+		}, waitingTime );
+	}
+
+	// Sets color in the color picker.
+	public setColor( color: string | undefined ): void {
+		if ( color && this.picker ) {
+			this.picker.setAttribute( 'color', color );
```
mlewand, line 82: Why not storing the color if the picker is not available? And when it can be unavailable?

If color is held as an observable you could start listening to it changes (reflect `color` value in the picker) right after the picker was created to have a guarantee that picker is available.

packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -0,0 +1,136 @@
...
+		}, waitingTime );
+
+		this._debounceInputEvent = debounce( ( color: string ) => {
+			this.fire( 'change', { value: color } );
+			this.setColor( color );
+		}, waitingTime );
+	}
+
+	// Sets color in the color picker.
+	public setColor( color: string | undefined ): void {
```
mlewand, line 80: It would be nicer to turn it into a observable property (string) called `color`.

Then out of the box we can nicely observe it for state changes, etc.

packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -0,0 +1,136 @@
...
+			tag: 'div',
+			attributes: {
+				class: [ 'ck', 'ck-color-picker' ]
+			},
+			children
+		} );
+
+		const waitingTime = 150;
+
+		this._debouncePickerEvent = debounce( ( color: string ) => {
```
mlewand, line 68: this.\_debounceInputEvent amd this.\_debouncePickerEvent are the same functions. We should have one of these and this function should be called in both cases.

Unless they need can't share debounce timer, but is this needed?

mlewand (changes requested):

mlewand (changes requested):

mlewand (commented):
Couple of more points.

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -293,6 +313,10 @@ export default class ColorTableView extends View {
 			this.items.add( this.documentColorsGrid );
 			this.focusTracker.add( this.documentColorsGrid.element! );
 			this._focusables.add( this.documentColorsGrid );
+
+			if ( this.colorPickerView ) {
+				this.items.add( this.colorPickerView );
```
mlewand, line 318: This logic is:

*   in `if ( this.documentColorsCount ) {` condition - this means that if the configuration `config.fontColor.documentColorsCount = 0` will hide color picker.
*   `appendGrids()` method, which is related to grids.

mlewand (approved):
Added couple of changes.

LGTM 👌

mlewand:
Lets wait for CI the CI to merge.

PR #13714, opened 2023-03-20
Remove return types from 'requires()' methods.

arkflpc:

Internal: Remove return types from 'requires()' methods.

---

Edit by @filipsobol: See comment below for rationale.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -27,6 +27,17 @@ module.exports = {
 	},
 	overrides: [
 		{
+			files: [ '**/*.ts' ],
+			rules: {
+				'@typescript-eslint/explicit-module-boundary-types': [
+					'error',
+					{
+						'allowedNames': [ 'requires' ]
+					}
+				]
+			}
+		},
+		{
 			files: [ '**/tests/**/*.js' ],
 			rules: {
 				'no-unused-expressions': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,7 +32,8 @@ module.exports = {
 				'@typescript-eslint/explicit-module-boundary-types': [
 					'error',
 					{
-						'allowedNames': [ 'requires' ]
+						'allowedNames': [ 'requires' ],
+						'allowArgumentsExplicitlyTypedAsAny': true
 					}
 				]
 			}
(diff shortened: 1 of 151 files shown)


.eslintrc.js
```diff
@@ -26,6 +26,17 @@ module.exports = {
...
+			files: [ '**/*.ts' ],
+			rules: {
+				'@typescript-eslint/explicit-module-boundary-types': [
+					'error',
+					{
+						'allowedNames': [ 'requires' ]
+					}
+				]
+			}
+		},
```
niegowski, line 39: This requires a follow-up to extract into the dedicated package.

packages/ckeditor5-core/src/editor/editor.ts
```diff
@@ -245,7 +245,7 @@ export default abstract class Editor extends ObservableMixin() {
 	 *
 	 * See also {@link module:core/editor/editor~Editor.defaultConfig}.
 	 */
-	public static builtinPlugins?: Array<PluginConstructor<Editor>>;
+	public static builtinPlugins?: ReadonlyArray<PluginConstructor<Editor>>;
```
niegowski, line 248: I'm not sure if this will not break the ability to extend builds? For example like here https://github.com/ckeditor/ckeditor5/blob/29f10a3b67026074b9c8475bec8dc69cb2288f58/docs/_snippets/build-classic-source.js#L12-L14

packages/ckeditor5-special-characters/src/specialcharactersessentials.ts
```diff
@@ -38,7 +38,7 @@ export default class SpecialCharactersEssentials extends Plugin {
 	/**
 	 * @inheritDoc
 	 */
-	public static get requires(): PluginDependencies {
+	public static get requires() {
```
niegowski, line 41: Missing `as const`.

niegowski (approved):

filipsobol:
Rationale for this change, for future reference:

> The generated `.d.ts` file for the `Essential` plugin is missing imports for some of its dependent plugins. Because of that, their `import './augmentation'` statements are never loaded, so the type augmentations do not take effect.
>
> Removing the explicit type from `require()` lets TypeScript infer it from the return value. This makes TypeScript include the necessary imports in the generated `.d.ts` file.

PR #13749, opened 2023-03-24
Accessibility for Font size and Font family dropdowns in toolbar

### LINKED ISSUE #13250: [Accessibility] NVDA screen reader not announcing the selected / not selected state for the 'Fonts' and 'Size' controls present in the CKEditor5 toolbar .

1. Open https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/full-featured-editor.html
2. Enable NVDA screen reader https://www.nvaccess.org/download/
3. Tab to the 'Fonts' or 'Size' control and press Enter to expand. Use Use 'Down Arrow ' key to navigate through the expanded list items and press Enter to select any font or size .

### ✔️ Expected result
NVDA should announce whether a font or size is selected or not selected as the user navigates

### ❌ Actual result
NVDA does not report which item is selected or not selected

### ❓ Possible solution
https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaSelected

## 📃 Other details

* Browser: Edge 108.0.1462.76
* OS: Windows 11 (any build)
* First affected CKEditor version: 4, also repros on 5

---

mmotyczynska:

Fix (font): Screen readers should now announce selected option in dropdown lists for font size and font family.

---

### Additional information


oleq (commented):
While reviewing this PR I attempted to make the heading dropdown accessible similarly because I think this won't end at font\* features. Soon, I discovered this didn't play out so well

The PoC is here [https://github.com/ckeditor/ckeditor5/compare/ck/13250-accessibility-selected-state-for-dropdowns-v2?expand=1.](https://github.com/ckeditor/ckeditor5/compare/ck/13250-accessibility-selected-state-for-dropdowns-v2?expand=1.)

The problem is that while font\* dropdowns have invisible labels, the heading presents the label to the user. As a result, the label is read by the screen reader twice after using the new (correct) aria labels.

*   The first time it is read because `.ck-button` (`ButtonView`) has `aria-labelledby` that points to the label.
*   The second time because the screen reader discovers the "value" of the button (text of the label).



More details:

*   This is happening because of `role="combobox"` set on the `<button>` element. 
*   This proves that my [initial approach to setting labels](https://github.com/ckeditor/ckeditor5/pull/13749/files#diff-9db2ec9c9e2a4e35c61a4cd66f60d3c664d51795ed8255f41fb2ac9b34b30110R71) that inspired Marta could be wrong. It works only as long as the label is invisible.
*   We may need to rethink how dropdowns are labeled, which now that I think about it, totally makes sense:
    *   We need a label for the whole dropdown, for instance "Font family" (this one is missing).
        *   But the API is there: `DropdownView#ariaDescribedById` exists and has been used in UIs such as table properties (as a child of `LabeledFieldView`).
    *   We need a label for the dynamic value of the feature, for instance "Helvetica", unlike what I proposed in the past and unlike what has been done in this PR.
    *   Then we need attributes that couple the dropdown button with the list. This is covered in this PR.
    *   Only then we'll get [as close to the standards](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) as possible.

mmotyczynska:
As we agreed to go with a different solution, I will close this PR. The new version can be found in #13893.

PR #13759, opened 2023-03-26
Allowed to easily set, change, save and load root attributes

### LINKED ISSUE #13395: Allow for storing additional data related to roots

Allow for storing additional data related to roots. We already have `RootAttributeOperation` and it is possible to set an attribute on a root element using `writer.setAttribute()`. However it is hard to use them outside of the editing pipeline. There's no simple API that would allow to store "metadata" that is connected with the document but is not the document data itself.

This is especially important for applications that might want to use multi-root editors. If an application "document" is built from multiple, separated editing areas, it is common that these editing areas may also have their own properties that can be changed by users as they edit the document.

It should be possible to easily save and load such additional data and it also should be propagated in real-time collaborative editing.

scofalik:

Other (engine): `RootAttributeOperation` is now correctly handled by `Differ`. Root attribute changes will be returned in `Differ#getChangedRoots()`.

Internal (editor-multi-root): Introduced API to easily save and load root attributes. Introduced `EditorConfig#rootsAttributes` and `MultiRootEditor#getRootsAttributes()`. Closes #13395.

Internal (editor-multi-root): Introduced `MultiRootEditor#getFullData()`. The method returns document data for all roots in multi-root editor.

Internal (editor-multi-root): `MultiRootEditor#addRoot()` has a new option `attributes`.

Internal (engine): `model.Writer#detachRoot()` will now remove attributes from the detached root.

---

### Additional information

Most things as internal because the changes will be in the changelog as "introduced multi-root editor type".

See https://github.com/ckeditor/ckeditor5/issues/13395#issuecomment-1484150081 for summary of proposed solution.

### DIFF
--- a/packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
+++ b/packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
@@ -283,6 +283,8 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
 	 *
 	 * See also {@link module:core/editor/editorconfig~EditorConfig#rootsAttributes `rootsAttributes` configuration option}.
 	 *
+	 * Note that attributes keys of attributes added in `attributes` option are also included in {@link #getRootsAttributes} return value.
+	 *
 	 * By setting `isUndoable` flag to `true`, you can allow for detaching the root using the undo feature.
 	 *
 	 * Additionally, you can group adding multiple roots in one undo step. This can be useful if you add multiple roots that are
@@ -308,6 +310,7 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
 		{ data = '', attributes = {}, elementName = '$root', isUndoable = false }: AddRootOptions = {}
 	): void {
 		const dataController = this.data;
+		const registeredKeys = this._registeredRootsAttributesKeys;
 
 		if ( isUndoable ) {
 			this.model.change( _addRoot );
@@ -323,6 +326,7 @@ export default class MultiRootEditor extends DataApiMixin( Editor ) {
 			}
 
 			for ( const key of Object.keys( attributes ) ) {
+				registeredKeys.add( key );
 				writer.setAttribute( key, attributes[ key ], root );
 			}
 		}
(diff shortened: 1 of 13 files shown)


scofalik:
See [https://github.com/ckeditor/ckeditor5/issues/13395#issuecomment-1484150081](https://github.com/ckeditor/ckeditor5/issues/13395#issuecomment-1484150081) for summary of proposed solution.

Inviz:
Will this sync attributes set via dom using mutation observer?

scofalik:
This solution is for attributes understood as, hm, roots metadata (additional model root data). You can write converters to convert these roots attributes from model to view and then DOM. We will probably write such guide at some point. But in general, the flow should be:

*   initialize root attributes when initing editor (you can set DOM attributes in DOM/HTML too)
*   custom features allow to change these attributes through a custom UI
*   converters (in custom features) handle these attributes (e.g. change something in DOM in the process)
*   when you save the data, you save the roots attributes as well

Automatic sync through mutation observer in a way that you described (DOM -> model) is against the philosophy of how CKE5 works. Instead, the changes should be done on model and be propagated downwards to DOM. You can introduce a custom UI or any other method to change model roots attributes, but the propagation should always be model -> view -> DOM.

scofalik:
> Is it true now? I remember it was not possible before, for one reason or another it didn't really work. I remember filed issue on that, and i stumbled upon that myself too.

I think it is difficult currently, as root attributes changes were not buffered in `Differ`. After this PR, it will be a bit easier. But in general, even now, you can set attributes on roots. But it is not convenient.

packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
```diff
@@ -602,6 +708,12 @@ export type DetachRootEvent = {
  */
 export type AddRootOptions = {
 	data?: string;
+	attributes?: RootAttributes;
```
niegowski, line 711: In the doclet for `AddRootOptions` type, `@param` is used incorrectly. Object properties should be documented for each property in a separate doclet for each field.

niegowski (commented):
LGTM. Only minor comments.

huzedong2015:
`addRoot` cannot add a new HTML element

PR #13784, opened 2023-03-29
Remove underscore from _updateEditorData method name and make it public.

### LINKED ISSUE #11008: Source editing doesn't save unless going back to WYSIWYG
Maybe this is rather a question than a bug but I couldn't find a bettter place to submit it. 

We're using CKeditor to edit the (mardown) contents of a textarea. Unfortuantely, the textarea isn't part of a submittable form and its contents (among others) are sent to a server after manually saving (and thereby closing the editing area). That's why I use a simple

```js
editor.model.document.on("change:data", (e) => {
    editor.updateSourceElement();
});
```

to ensure that the textarea always contains the user input of the WYSIWYG editor which works fine, except for input made from within the source editor.  There, the `change:data` event never fires and subsequently, nothing is changed.

For CKEditor 4 I found [this](https://stackoverflow.com/questions/36079798/ckeditor-checkdirty-on-source-editing-area). Is there something similar for CKEditor 5?  What am I missing?

1. Use a manual saving setup as above
2. Enter source mode and change something
3. Save (or observe the `change:data` event) and see that nothing happens

### ✔️ Expected result

Changing something while in source editing should also trigger the 

### ❌ Actual result

Only after going back to the WYSIWYG mode, the changes are applied (i.e., the `change:data` event is fired).

## 📃 Other details

* Browser: Firefox 95.0, Chromium 96.0.4664.110
* OS: Arch Linux
* First affected CKEditor version: 31.0.0
* Installed CKEditor plugins: Markdown, SourceEditing

---

mremiszewski:

Feature (source-editing): Remove the underscore from `_updateEditorData` method name and make it public. Closes #11008.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-source-editing/src/sourceediting.ts
+++ b/packages/ckeditor5-source-editing/src/sourceediting.ts
@@ -9,7 +9,7 @@
 
 /* global console */
 
-import { type Editor, Plugin, PendingActions } from 'ckeditor5/src/core';
+import { type ContextPluginDependencies, type Editor, Plugin, PendingActions } from 'ckeditor5/src/core';
 import { ButtonView } from 'ckeditor5/src/ui';
 import { createElement, ElementReplacer } from 'ckeditor5/src/utils';
 import { formatHtml } from './utils/formathtml';
@@ -38,7 +38,7 @@ export default class SourceEditing extends Plugin {
 	/**
 	 * @inheritDoc
 	 */
-	public static get requires() {
+	public static get requires(): ContextPluginDependencies {
 		return [ PendingActions ] as const;
 	}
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-source-editing/src/sourceediting.ts
```diff
@@ -9,7 +9,7 @@
 
 /* global console */
 
-import { type Editor, Plugin, PendingActions } from 'ckeditor5/src/core';
+import { type ContextPluginDependencies, type Editor, Plugin, PendingActions } from 'ckeditor5/src/core';
```
arkflpc, line 12:
```suggestion
import { type Editor, Plugin, PendingActions } from 'ckeditor5/src/core';
```

packages/ckeditor5-source-editing/src/sourceediting.ts
```diff
@@ -38,7 +38,7 @@ export default class SourceEditing extends Plugin {
 	/**
 	 * @inheritDoc
 	 */
-	public static get requires() {
+	public static get requires(): ContextPluginDependencies {
```
arkflpc, line 41:
```suggestion
	public static get requires() {
```

arkflpc (changes requested):

arkflpc (approved):

PR #13796, opened 2023-03-30
Convert the `ckeditor5` package to JavaScript

### LINKED ISSUE #13794: Convert back DLL packages to JavaScript
### Why

* We must prepare DLL builds from JavaScript files to save backward compatibility between them.
	* Example: As a maintainer of a package written for v33.0.0, I would like to have it compatible with the latest version - v37.0.0.
	* If there were no breaking changes in the editor API, my dll feature should work fine with the latest DLL build.
* JS, TS, and typing contain the same value. TS needs to bring something here, and unfortunately, it doesn't.
* Whenever somebody tries to prepare DLL builds, we must prepare JS files first. We tried to solve it using npm scripts (pre hooks), but there is at least one script that does not trigger them ([`scripts/dll/build-dlls.js`](https://github.com/ckeditor/ckeditor5/blob/02fcc03a27e96337b7bcc18980da9140768317b9/scripts/dll/build-dlls.js)).

### Decisions

* Let's convert `ckeditor5` and `ckeditor5-collaboration` packages to JavaScript.
* Make sure to generate typings for them.

pomek:

Internal: Convert the `ckeditor5` package to JavaScript. Closes #13794.


pomek:
https://github.com/ckeditor/ckeditor5/issues/13794#issuecomment-1491610327

PR #13828, opened 2023-04-05
Docs: Add TypeScript support to Angular integration guide.

gorzelinski:

Docs: Add TypeScript option to Angular integration guide. Closes https://github.com/cksource/ckeditor5-internal/issues/2861.

---

### Additional information


gorzelinski:
The PR isn't ready, but I added it to list some problems.

## Using the Document Editor build
The template throws an error: `Argument of type 'Editor' is not assignable to parameter of type 'DecoupledEditor.'` Changing the argument to the `Editor` type causes another problem in the `.component.ts` file: the `Property 'toolbar' does not exist on type 'EditorUIView.'`

## Using a custom CKEditor 5 build
The section links to another guide: [create a custom build guide](https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/quick-start-other.html#building-the-editor-from-source). It starts assuming you generated a `ckeditor.js` file. In a TypeScript project, that won't work (without modifying the config). To make it work, I need to change the linked guide as well, I guess?

## Integrating a build from the online builder
The online builder still uses the `36.0.1` package versions. I have no control over that. I tried to change the build locally and modify the package versions, but the error about missing declarations remains. The error will probably vanish after updating package versions in the online builder. But, at this moment, I don't know if I have a way to test that locally.

apadol-cksource:
> ## Using a custom CKEditor 5 build
> 
> The section links to another guide: [create a custom build guide](https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/quick-start-other.html#building-the-editor-from-source). It starts assuming you generated a `ckeditor.js` file. In a TypeScript project, that won't work (without modifying the config). To make it work, I need to change the linked guide as well, I guess?

Let's leave it as is for now. We may remove it in favor of [integrating from source](https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/alternative-setups/integrating-from-source-webpack.html). TBD

apadol-cksource:
> ## Using the Document Editor build
> 
> The template throws an error: `Argument of type 'Editor' is not assignable to parameter of type 'DecoupledEditor.'` Changing the argument to the `Editor` type causes another problem in the `.component.ts` file: the `Property 'toolbar' does not exist on type 'EditorUIView.'`

We need to figure out a proper way to use the component.  There issue reported here: #13838

Reinmar:
My feedback [https://github.com/ckeditor/ckeditor5/issues/13838#issuecomment-1500346786](https://github.com/ckeditor/ckeditor5/issues/13838#issuecomment-1500346786)

> <table><tbody><tr><td><blockquote><p>Just discovered that we can also set <code>strictTemplates</code> to <code>false</code> in the default AngularCLI configuration and that would be actually more elegant than using <code>any</code>.</p></blockquote><p>This isn't a viable solution for some people. Changing the tsconfig of one's project to loosen the checks might not be accepted by its maintainers as it affects the entire project.</p><p>So, in my opinion this should be reverted: <a href="https://github.com/ckeditor/ckeditor5/commit/d862032311d752d44a6443901304e85dfac5393d">d862032</a></p><p>And this is the correct workaround for now: <a href="https://github.com/ckeditor/ckeditor5/commit/633eb37f362494bbfa81466c09768205be3e5c96">633eb37</a></p></td></tr></tbody></table>

Reinmar (changes requested):
As commented.

PR #13829, opened 2023-04-05
Implement the final dropdowns design for the color picker

illia-stv:

Internal (font): Implement the final dropdown design for the color picker.
Closes https://github.com/cksource/ckeditor5-internal/issues/2980.

---

### Additional information


packages/ckeditor5-font/theme/fontcolor.css
```diff
@@ -3,19 +3,43 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
-.ck .ck-button.ck-color-table__remove-color {
+.ck .ck-button.ck-color-table__remove-color,
+.ck .ck-button.ck-color-table__color-picker {
 	display: flex;
 	align-items: center;
 	width: 100%;
```
pszczesniak, line 10: Please add here also a mixin `ck-dir rtl` and set `justify-content: flex-start;` to make buttons looks ok in RTL mode :)

Screenshot 2023-04-17 at 15 11 16

packages/ckeditor5-font/theme/fontcolor.css
```diff
@@ -3,19 +3,43 @@
...
 
 .ck.ck-color-picker {
+	padding: 8px;
+
 	& .hex-color-picker {
 		height: 100px;
+		margin: 0 0 8px 0;
 
 		&::part(saturation),
 		&::part(hue) {
```
pszczesniak, line 45: :point\_up: those two parts should have border radius set to `--ck-border-radius`; `saturation` part - top left/right, `hue` part bottom left/right.

packages/ckeditor5-font/theme/fontcolor.css
```diff
@@ -3,19 +3,43 @@
...
 label.ck.ck-color-grid__label {
 	font-weight: unset;
 }
 
 .ck.ck-color-picker {
+	padding: 8px;
+
 	& .hex-color-picker {
 		height: 100px;
+		margin: 0 0 8px 0;
```
pszczesniak, line 42: Bottom margin need to be larger, for example `0 0 var(--ck-spacing-large) 0`, because when input below is active - label is almost touching the color picker slider.

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -391,6 +433,81 @@ export default class ColorTableView extends View {
...
+			children
+		} );
+
+		return actionBarRow;
+	}
+
+	/**
+	 * Creates "save" and "cancel" buttons.
+	 */
+	private _createActionButtons() {
```
pszczesniak, line 466: Missing return type.

pszczesniak (changes requested):
It looks quite good :+1: just some polishing needed :)

One thing that i think it would be good to do is to add CSS `flex: 1` on `ck-button-save`, and `ck-button-cancel` - it will make that the buttons will take whole space in the action bar.

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -391,6 +433,81 @@ export default class ColorTableView extends View {
...
+		const locale = this.locale;
+		const saveButtonView = new ButtonView( locale );
+		const cancelButtonView = new ButtonView( locale );
+
+		saveButtonView.set( {
+			icon: icons.check,
+			class: 'ck-button-save',
+			type: 'submit'
```
pszczesniak, line 474: We need a `label` for accessibility. It can be "hidden" to show only the icon - use `withText: false`

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -391,6 +433,81 @@ export default class ColorTableView extends View {
...
+		} );
+
+		cancelButtonView.set( {
+			icon: icons.cancel,
+			class: 'ck-button-cancel'
```
pszczesniak, line 479: We need a `label` for accessibility. It can be "hidden" to show only the icon - use `withText: false`

pszczesniak:
Small code refactor d7a998a : Buttons labels and `CSS` small improvements.

PR #13834, opened 2023-04-06
Created aliases when building docs

### LINKED ISSUE #13835: Snippet adapter should be able to map requests to non-published files

* The nightly and production docs builders use the npm version of packages.
* These packages do not contain the `tests/` nor `docs/` directories.
* Unfortunately, our snippets load configurations or assets from non-published directories.
* Webpack emits errors due to looking for non-existing files.
* It should be possible to redirect these imports using [`resolve.alias`](https://webpack.js.org/configuration/resolve/#resolvealias) option.
* Let's try. The goal is to create aliases for non-existing files. Webpack should load them from the `packages/` directory instead of `node_modules/`.

pomek:

Internal: Created a map for snippet adapter that redirects imports of non-existing files from `node_modules/` to the `packages/` directory. Closes #13835.

---

### Additional information

### DIFF
--- a/scripts/docs/snippetadapter.js
+++ b/scripts/docs/snippetadapter.js
@@ -21,6 +21,41 @@ const DEFAULT_LANGUAGE = 'en';
 const MULTI_LANGUAGE = 'multi-language';
 const SNIPPETS_BUILD_CHUNK_SIZE = 50;
 
+const CKEDITOR5_ROOT_DIRECTORY = path.join( __dirname, '..', '..' );
+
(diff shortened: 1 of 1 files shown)


psmyrek (approved):
LGTM 🎉

martnpaneq (approved):
The only issue I see is that whenever the config changes, or we add a new package that breaks the docs, we would have to update this file.

The perfect solution would be to tell webpack to always look in the source code instead of node_modules if it can't find a specific file. But I don't know if it's possible.

Issue #13849, opened 2023-04-11
When using TypeScript all imports must go through package main file (src/index.js)


Reinmar:
Applies to:

*   [https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/alternative-setups/integrating-from-source-vite.html#typescript-2](https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/alternative-setups/integrating-from-source-vite.html#typescript-2) (imports need to be corrected)
*   [https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/alternative-setups/integrating-from-source-webpack.html#typescript-2](https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/alternative-setups/integrating-from-source-webpack.html#typescript-2) (imports need to be corrected)
*   [https://ckeditor.com/docs/ckeditor5/latest/installation/working-with-typescript.html](https://ckeditor.com/docs/ckeditor5/latest/installation/working-with-typescript.html) (missing section about the rule that all imports must go through the src/index.js – ask the core team for more clarification why if needed)

This rule is mentioned in the 2nd point of this migration guide: [https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-37.html#typescript-typings](https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-37.html#typescript-typings)

![](https://user-images.githubusercontent.com/156149/231086979-9ba419c3-516b-4b6e-bf21-8f148c3a0397.png)

Without adhering to it one may stumble upon weird typing errors (it's actually worth quoting somewhere in the docs the exact error that would be thrown because people may be googling it). Again, consult the core team for more info.

stefanseeger:
Migrating to 27 with typescript is literally impossible at the moment.

As mentioned by @Reinmar we are supposed to not use the src paths anymore and remove all @types. But at the same time not all packages were updated to export the types like `@ckeditor/ckeditor5-special-characters`. https://ckeditor.com/docs/ckeditor5/latest/features/special-characters.html#demo

This should not happen for a paid software.

Issue #13740, opened 2023-03-23
Improve typing of the `Editor.create` method


filipsobol:
I think it's a bad experience and a bad representation of our TypeScript migration if the very first step of using the editor requires the developer to do type casting to make the TS gods happy.

![type-casting](https://user-images.githubusercontent.com/4145208/227185528-44d566b5-cb81-421b-954a-96a86d584c91.jpg)

The examples in our documentation show that the first parameter passed to `Editor.create` should be `document.querySelector('xxx')`, which returns `Element | null`, but the method expects `HTMLElement | string`.

I believe we change the method signature to `HTMLElement | Element | string | null` and check at runtime if the passed element is `instanceof HTMLelement` (e.g. to make sure it's not `SVGElement`) or `typeof string`.

Reinmar:
VERY good point.

This is a perfect example of a case where we need to prioritize DX over the simplicity of our code. This is one of the very most important pieces of our API in terms of the DX for integrators.

So, I'm all for improving this.

Witoso:
I think I can't entirely agree. Types should act as a description of an interface. This may be very confusing if we say we accept `Element` but then reject it in runtime.

Should we check if it's HTMLElement? Sure, we can catch errors early, in the crucial flow. But I wouldn't change the type signature. It's up to the integrator to correctly pass arguments, type cast, or type guard them, their decision.

filipsobol:
After a discussion, we decided to leave this as is for now and see if it becomes a problem for developers. The TypeScript examples in the documentation will have type castings that will silence this error:

```typescript
ClassicBuild.create( document.querySelector<HTMLElement>('#app')! )
```

Reinmar:
When I look at this in the docs, I think this would be better:

> ```ts
> 
> ClassicBuild.create( document.querySelector('#app') as HTMLElement )
> ```

The `!` at the end of the line is harder to spot and the `<>` brackets make this harder for me (TS noob) to read.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #13856, opened 2023-04-11
Docs: Changing webpack guide to adapt it to new import approach

apadol-cksource:

Docs: Changing webpack guide to adapt it to new import approach. See #13849 .

---

### Additional information

### DIFF
--- a/docs/installation/advanced/integrating-from-source-webpack.md
+++ b/docs/installation/advanced/integrating-from-source-webpack.md
@@ -16,7 +16,7 @@ This scenario allows you to fully control the building process of CKEditor 5. Th
 	Similar results to what this method allows can be achieved by {@link installation/getting-started/quick-start-other#building-the-editor-from-source customizing an existing build} and integrating your custom build. This will give faster build times (since CKEditor 5 will be built once and committed), however, it requires maintaining a separate repository and installing the code from that repository into your project (e.g. by publishing a new npm package or using tools like [Lerna](https://github.com/lerna/lerna)). This makes it less convenient than the method described in this scenario.
 </info-box>
 
-First of all, you need to initialize a new project using `npm init`. Then, you need to install source packages that you will use. If you base your integration on one of the existing builds, you can take them from that build's `package.json` file (see e.g. [classic build's `package.json`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-build-classic/package.json)). At this moment you can choose the editor creator and the features you want. Keep in mind, however, that all packages (excluding `@ckeditor/ckeditor5-dev-*`) {@link installation/plugins/installing-plugins#requirements must have the same version as the base editor package}.
+First of all, you need to install source packages that you will use in your existing project. If you base your integration on one of the existing builds, you can take them from that build's `package.json` file (see e.g. [classic build's `package.json`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-build-classic/package.json)). At this moment you can choose the editor type and the features you want. Keep in mind, however, that all packages (excluding `@ckeditor/ckeditor5-dev-*`) {@link installation/plugins/installing-plugins#requirements must have the same version as the base editor package}.
 
 Copy these dependencies to your `package.json` and call `npm install` to install them. You can also install them individually. An example list of plugins may look like this:
 
@@ -253,12 +253,13 @@ If you want to use TypeScript, add the `tsconfig.json` file at the root of your 
 {
   "compilerOptions": {
     "lib": [
-      "ESNext",
-      "dom"
+      "DOM",
+      "DOM.Iterable"
     ],
     "outDir": "lib",
     "removeComments": true,
-    "target": "ES6",
+	"module": "es6",
+    "target": "es2019",
     "baseUrl": "./",
     "esModuleInterop": true,
     "moduleResolution": "node",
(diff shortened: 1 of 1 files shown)


docs/installation/working-with-typescript.md
```diff
@@ -46,10 +47,19 @@ ClassicEditor.create( editorPlaceholder ).catch( error => {
...
+
+```ts
+// Instead of:
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+
+// Do:
+import { Plugin } from '@ckeditor/ckeditor5-core';
+```
 
+Thanks to this approach you will avoid problems with missing types during the compile time of your project.
```
Reinmar, line 62:
```suggestion
This approach ensures that TypeScript correctly loads all module augmentation code necessary to make certain types work. The previous method (importing via `src/*`) still works in most cases, but [it may randomly break](https://github.com/ckeditor/ckeditor5/issues/13433).

```

docs/installation/advanced/integrating-from-source-webpack.md
```diff
@@ -359,32 +360,32 @@ If you want to use TypeScript, the `ckeditor` file looks very similar. However,
...
+import { Italic } from '@ckeditor/ckeditor5-basic-styles';
+import { BlockQuote } from '@ckeditor/ckeditor5-block-quote';
+import { Heading } from '@ckeditor/ckeditor5-heading';
+import { Link } from '@ckeditor/ckeditor5-link';
+import { List } from '@ckeditor/ckeditor5-list';
+import { Paragraph } from '@ckeditor/ckeditor5-paragraph';
+
+export default class CustomEditor extends ClassicEditor {}
+
+CustomEditor.builtinPlugins = [
```
Reinmar, line 376: This isn't consistent with how we define builds today. We had to change the syntax a bit.

docs/installation/advanced/integrating-from-source-webpack.md
```diff
@@ -253,12 +253,13 @@ If you want to use TypeScript, add the `tsconfig.json` file at the root of your
 {
   "compilerOptions": {
     "lib": [
-      "ESNext",
-      "dom"
+      "DOM",
```
Reinmar, line 256: Why do we show TS configuration? Do we have any specific requirements for ts-config to make CKE work?

If not: ts-config is none of our business.

docs/installation/advanced/integrating-from-source-vite.md
```diff
@@ -294,30 +294,30 @@ You can also translate the above code using TypeScript.
...
+import { Bold } from '@ckeditor/ckeditor5-basic-styles';
+import { Italic } from '@ckeditor/ckeditor5-basic-styles';
+import { BlockQuote } from '@ckeditor/ckeditor5-block-quote';
+import { Heading } from '@ckeditor/ckeditor5-heading';
+import { Link } from '@ckeditor/ckeditor5-link';
+import { List } from '@ckeditor/ckeditor5-list';
+import { Paragraph } from '@ckeditor/ckeditor5-paragraph';
 
 ClassicEditor
     .create( document.querySelector<HTMLElement>( '#app')!, {
```
Reinmar, line 309: Use the `as ...` notation to avoid the `!`? I know we had a discussion about this but I'd like to repeat it: https://github.com/ckeditor/ckeditor5/issues/13740#issuecomment-1504184685

docs/installation/advanced/integrating-from-source-vite.md
```diff
@@ -168,32 +168,32 @@ If you scaffolded your project using a Vite template with TypeScript, add the `c
...
+import { Italic } from '@ckeditor/ckeditor5-basic-styles';
+import { BlockQuote } from '@ckeditor/ckeditor5-block-quote';
+import { Heading } from '@ckeditor/ckeditor5-heading';
+import { Link } from '@ckeditor/ckeditor5-link';
+import { List } from '@ckeditor/ckeditor5-list';
+import { Paragraph } from '@ckeditor/ckeditor5-paragraph';
+
+export default class CustomEditor extends ClassicEditor {}
+
+CustomEditor.builtinPlugins = [
```
Reinmar, line 184: Same as for webpack: we changed the syntax.

docs/installation/working-with-typescript.md
```diff
@@ -46,10 +47,19 @@ ClassicEditor.create( editorPlaceholder ).catch( error => {
...
-<info-box warning>
-Custom builds produced by online builder do not provide built-in type definitions yet.
-</info-box>
+### Installing CKEditor 5 packages in your TypeScript project
+
+When you use TypeScript you need to import all modules provided by CKEditor 5 using a package entry point instead of a path to a module.
+
+```ts
+// Instead of:
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
```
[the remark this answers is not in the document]
Reinmar: Ticket: #13862

godai78 (approved):
LGTM

docs/installation/advanced/integrating-from-source-vite.md
Witoso: I would go even further, and a simple check that shows we care about type guarding :)

```ts
const editorElement = document.querySelector( "#app" );

if ( editorElement === null || !(editorElement instanceof HTMLElement ) ) {
  throw new Error( "No editor area!" );
}
```
Or more simple one
```ts
const editorElement = document.getElementById( "app" );

if ( editorElement === null ) {
  throw new Error( "No editor area!" );
}
```

docs/installation/advanced/integrating-from-source-vite.md
Reinmar: I meant this: https://github.com/ckeditor/ckeditor5/blob/29857dbdec67595ab35500d4f99c1883535e40dd/packages/ckeditor5-build-inline/src/ckeditor.ts#L29-L30

docs/installation/advanced/integrating-from-source-webpack.md
Reinmar: I meant this: [https://github.com/ckeditor/ckeditor5/blob/29857dbdec67595ab35500d4f99c1883535e40dd/packages/ckeditor5-build-inline/src/ckeditor.ts#L29-L30](https://github.com/ckeditor/ckeditor5/blob/29857dbdec67595ab35500d4f99c1883535e40dd/packages/ckeditor5-build-inline/src/ckeditor.ts#L29-L30)

AFAIR (cc @niegowski @filipsobol @arkflpc ) we changed this recently for a particular reason. But I don't remember what and whether it applies here too.

docs/installation/advanced/integrating-from-source-webpack.md
Reinmar: Makes sense to limit it and have it consistent with the upgrade guide.

docs/installation/advanced/integrating-from-source-webpack.md
arkflpc: This was changed in builds to allow inferring `builtinPlugins` property type.

Without that, the type would be `Array<PluginConstructor>`.

With type inference, it is something like `Array<typeof Essentials | typeof Autoformat | typeof Bold /* etc. */>`.

Thanks to it, in `.d.ts` of this build, all plugin imports are included (because now they are used as part of `ClassicEditor` type). Because of it, all imports are seen by client code (including `./augmentation.ts` files) and the type inference of `EditorConfig` (and commands) works properly now.

PR #13891, opened 2023-04-17
Change the order of base classes in mixed types

arkflpc:

Docs (utils): Fixes the order of super-classes in mixins.

---

### Additional information


pomek (approved):
It works as expected.

image

PR #13893, opened 2023-04-17
Accessibility for Font size/family and Heading dropdowns in toolbar

mmotyczynska:

Fix (font, heading): Screen readers should now announce selected option in dropdown lists for font size, font family and heading. Closes https://github.com/ckeditor/ckeditor5/issues/13250.

---

### Additional information

There was already [one solution](https://github.com/ckeditor/ckeditor5/pull/13749) in which we tried to implement this but it was not good enough. As described [here](https://github.com/ckeditor/ckeditor5/pull/13749#issuecomment-1507057330), we decided to go with another approach.
Besides font we added here also heading feature to make sure that this time the proposed solution will cover also dropdowns with the selected value shown in the toolbar.


illia-stv (approved):
For me it looks good!!!

packages/ckeditor5-ui/src/button/buttonview.ts
```diff
@@ -140,6 +140,26 @@ export default class ButtonView extends View<HTMLButtonElement> implements Butto
...
 
+	/**
+	 * @inheritDoc
+	 */
+	declare public role: string | undefined;
+
+	/**
+	 * @inheritDoc
+	 */
+	declare public ariaChecked: boolean | undefined;
```
oleq, line 151: Don't we need these in the `Button` interface as well?

packages/ckeditor5-ui/src/button/buttonview.ts
```diff
@@ -209,10 +233,13 @@ export default class ButtonView extends View<HTMLButtonElement> implements Butto
 					bind.if( 'withText', 'ck-button_with-text' ),
 					bind.if( 'withKeystroke', 'ck-button_with-keystroke' )
 				],
+				role: bind.to( 'role' ),
 				type: bind.to( 'type', value => value ? value : 'button' ),
 				tabindex: bind.to( 'tabindex' ),
-				'aria-labelledby': `ck-editor__aria-label_${ ariaLabelUid }`,
+				'aria-label': bind.to( 'ariaLabel' ),
+				'aria-labelledby': bind.to( 'ariaLabelledBy' ),
```
oleq, line 240: The existing test checks if `aria-labelledby` starts with `ck-editor__aria-label`. This is a default value but we also need a test to see if the new observable `ariaLabelledBy` property is actually observable and what it does.

packages/ckeditor5-ui/src/dropdown/utils.ts
```diff
@@ -310,6 +310,7 @@ export function addListToDropdown(
 	itemsOrCallback: Collection<ListDropdownItemDefinition> | ( () => Collection<ListDropdownItemDefinition> ),
 	options: {
 		ariaLabel?: string;
+		role?: string;
```
oleq, line 313: Missing API docs.

packages/ckeditor5-heading/src/headingui.ts
```diff
@@ -83,12 +84,17 @@ export default class HeadingUI extends Plugin {
...
 			const dropdownView = createDropdown( locale );
-			addListToDropdown( dropdownView, itemDefinitions );
+			addListToDropdown( dropdownView, itemDefinitions, {
+				ariaLabel: accessibleLabel,
+				role: 'menu'
+			} );
 
 			dropdownView.buttonView.set( {
+				ariaLabel: accessibleLabel,
+				ariaLabelledBy: undefined,
```
oleq, line 94: These changes have no test coverage.

packages/ckeditor5-ui/src/button/buttonview.ts
mmotyczynska: Test added.

PR #13905, opened 2023-04-18
Remove 2 of 3 ts-ignore flags.

pszczesniak:

Internal (font): 3rd party libs used for color picker should be typed.
Closes https://github.com/cksource/ckeditor5-internal/issues/3070.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ui/src/colorpicker/utils.ts
+++ b/packages/ckeditor5-ui/src/colorpicker/utils.ts
@@ -12,12 +12,18 @@
 // There are no available types for 'color-parse' module.
 // @ts-ignore
 import { default as parse } from 'color-parse';
-import * as convert from 'color-convert';
+import { hex, rgb, hsl, hwb, lab, lch } from 'color-convert';
 
 /**
  * Color formats handled by color converter.
  */
 export type ColorPickerFormat = 'hex' | 'rgb' | 'hsl' | 'hwb' | 'lab' | 'lch';
+
+type colorObject = {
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-ui/src/colorpicker/utils.ts
```diff
@@ -32,6 +38,19 @@ export type ColorPickerConfig = {
 	format?: ColorPickerFormat;
 };
 
+/**
+ * Color converter object that "returns" dedicated function to convert color based on object `key`.
+ */
+
+const convert: any = {
```
illia-stv, line 45: It should be typed properly, without `any`. In worst case, when you can not type something you could use `unknown`.

illia-stv (changes requested):
Looks very nice, but needs tiny improvement.

packages/ckeditor5-ui/src/colorpicker/utils.ts
pszczesniak: Yes, agree - this lib should be typed properly :wink:. For now we're using typings from [community types](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/color-convert) .

`unknown` is not working in this case.

If You got any tips that can help - would love to hear them and use them in this particular case.

oleq (commented):
I refactored the code and improved typings in another branch. Please take a look and cherry-pick changes if you think they make sense https://github.com/ckeditor/ckeditor5/compare/ck/epic/2356-color-picker...ck/epic/2356-color-picker-v2?expand=1.

illia-stv (approved):
Looks good for me

PR #13914, opened 2023-04-19
Create a helper to change views in dropdowns

illia-stv:

Feature: Create a helper to change views in dropdowns. Closes #[3026](https://github.com/cksource/ckeditor5-internal/issues/3026).

---

### Additional information


oleq (changes requested):
LGTM. Refactoring needed, though to decouple the parent component from the children:

*   component that aggregates grid + show picker button
*   component that aggregates the picket

This will make the top component's code way shorter and API simpler and move the complexity to sub-components.

mlewand:
I noticed there's some functional regression:

1.  Enter http://localhost:8125/ckeditor5-font/tests/manual/font-color.html
2.  Select first paragraph.
3.  Open font color dropdown.
4.  Press escape.

### Expected

the toolbar is focused

### Actual

Focus goes somewhere else.

mlewand:
Let's fill-in a follow-up for improved focus styling in color picker (for these circles). Currently the there's some focus indication (I believe shadow of some sort), however it should be more visible. Maybe our blueish color with a relatively thick outline?

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -222,7 +162,59 @@ export default class ColorTableView extends View {
...
+			attributes: {
+				class: colorTableBind.if( 'isColorTableVisible', 'ck-hidden', value => !value )
+			}
+		} );
+
+		this.colorPickerComponent.extendTemplate( {
+			attributes: {
+				class: colorPickerBind.if( 'isColorPickerVisible', 'ck-hidden', value => !value )
+			}
+		} );
```
oleq, line 195: This asks for `#isVisible` observable properties in `this.colorTableComponent` and `this.colorPickerComponent` bound internally to `ck-hidden` class. Then you would bind them here to `#isColorTableVisible` and `#isColorPickerVisible` respectively. 

This would create a chain of data (view model state) instead of cross-layer manipulation of templates. Let children components take care of themselves and encapsulate their internal logic. Communicate with them using a single property change.

https://en.wikipedia.org/wiki/Separation_of_concerns

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -306,79 +251,73 @@ export default class ColorTableView extends View {
...
+		this.colorTableComponent.delegate( 'execute' ).to( this );
+
+		if ( !this._focusables.length ) {
+			this._addColorTablesElementsToFocusTracker();
+			this.focus();
 		}
+	}
 
-		this.staticColorsGrid = this._createStaticColorsGrid();
+	public appendColorPicker( isColorPicker: ColorPickerConfig ): void {
```
oleq, line 265: See my comment in `appendGrids()`.

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -306,79 +251,73 @@ export default class ColorTableView extends View {
 		this.keystrokes.destroy();
 	}
 
-	/**
-	 * Appends {@link #staticColorsGrid} and {@link #documentColorsGrid} views.
-	 */
 	public appendGrids(): void {
```
oleq, line 254: Missing API docs.

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -306,79 +251,73 @@ export default class ColorTableView extends View {
...
+		if ( this.colorTableComponent.colorPickerButtonView ) {
+			this.focusTracker.add( this.colorTableComponent.colorPickerButtonView.element! );
+			this._focusables.add( this.colorTableComponent.colorPickerButtonView );
 		}
+
+		this._addColorPickersElementsToFocusTracker();
+		this._stopPropagationOnArrowsKeys();
+	}
+
+	public updateDocumentColors( model: Model, componentName: string ): void {
```
oleq, line 293: Missing API docs.

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -306,79 +251,73 @@ export default class ColorTableView extends View {
...
+		this.colorTableComponent.updateDocumentColors( model, componentName );
+	}
+
+	public updateSelectedColors(): void {
```
oleq, line 297: Missing API docs.

packages/ckeditor5-font/src/ui/colortableview.ts
```diff
@@ -622,3 +747,173 @@ export default class ColorTableView extends View {
 		}
 	}
 }
+
+class ColorPickerComponentView extends View {
```
oleq, line 751: Missing API docs.

oleq (changes requested):
The refactoring goes in the right direction. The layers (components) are still coupled in a couple of places, though. We can further streamline the architecture here.

oleq (changes requested):
An update on the status of this PR: The code has been refactored during 1:1 sessions and feels to be ready to be shipped.

In the future `ColorTableView` will be moved to ckeditor5-ui and shared between ckeditor5-table and ckeditor5-font.

What we need now is:

*   Missing docs (look for "TODO" in the code)
*   Missing tests after the refactoring.
*   Green CI.

Dumluregn:
Travis fails due to some CKBox validation issue.

Dumluregn (approved):

PR #13918, opened 2023-04-19
Fix (core): Add `licenseKey` property to `EditorConfig` interface

### LINKED ISSUE #13906: [TS] EditorConfig misses `licenseKey`

1. Follow the setup of `ClassicEditor` as, for example, given in [Overview - CKEditor 5 Documentation](https://ckeditor.com/docs/ckeditor5/37.0.1/features/pagination/pagination.html#installation).

### ✔️ Expected result

Given the `licenseKey` as option, it compiles successfully.

### ❌ Actual result

It fails to compile as `licenseKey` is not part of [EditorConfig](https://github.com/ckeditor/ckeditor5/blob/3dc1e52036bc89d488db92d3d440bb360ec1aa69/packages/ckeditor5-core/src/editor/editorconfig.ts).

### More Notes

Possibly related to:

* https://github.com/ckeditor/ckeditor5/issues/7603

Also note, that this is a migration issue, when coming from typings at DefinitelyTyped, where `EditorConfig` had a `licenseKey`:

* https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ckeditor__ckeditor5-core/src/editor/editorconfig.d.ts#L61

(I think in previous versions, it even got marked as "required" property, why we still have it in our code.)

### Workaround for us

We don't need the `licenseKey` (yet) and may safely skip it.

## 📃 Other details

* Browser: Chrome
* OS: Windows
* First affected CKEditor version: 37.0.1
* Installed CKEditor plugins: irrelevant (we don't have any plugins by default, that require a license)

---

filipsobol:

Fix (core): Add `licenseKey` property to `EditorConfig` interface. Closes #13906.

### DIFF
--- a/packages/ckeditor5-core/src/editor/editorconfig.ts
+++ b/packages/ckeditor5-core/src/editor/editorconfig.ts
@@ -488,6 +488,14 @@ export interface EditorConfig {
 	 * {@glink features/html/html-embed HTML embed} features.
 	 */
 	updateSourceElementOnDestroy?: boolean;
+
+	/**
+	 * The license key for the CKEditor 5 premium features.
+	 *
+	 * If you do not have a key yet, please [contact us](https://ckeditor.com/contact/) or
+	 * [order a trial](https://orders.ckeditor.com/trial/premium-features).
+	 */
+	licenseKey?: string;
 }
 
 /**


niegowski (commented):
This `licenseKey` is moved from internal code, we must remove it from other places (API docs will not build properly if there are multiple descriptions for a single property).

arkflpc (approved):

PR #13920, opened 2023-04-19
Update CKE4 plugin and configuration compatibility tables after Productivity Pack is released

AnnaTomanek:

Docs: Updated CKE4 plugin and configuration compatibility tables with templates and format painter. Closes cksource/ckeditor5-internal#3138.

---

### Additional information

_Also:_

*   _Removed WSC from the configuration compatibility table as it's no longer available in CKE4 so the links are invalid._
*   _Added `FullPage` API as an equivalent to CKE4's `fullPage` configuration option._


godai78 (approved):
LGTM

PR #13962, opened 2023-04-24
Ci/3161 selected color propagation

Dumluregn:

Internal(font): Selected color will be propagated to color picker when it is created.

---

### Additional information


mmotyczynska (approved):
LGTM :+1:

Dumluregn:
@mmotyczynska Turns out this PR should also solve cksource/ckeditor5-internal#3065 - could you confirm it?

mmotyczynska:
> @mmotyczynska Turns out this PR should also solve [cksource/ckeditor5-internal#3065](https://github.com/cksource/ckeditor5-internal/issues/3065) - could you confirm it?

  
In case of font color: does not work if the selection is in text with no color set.

In case of background color: does not work at all (unless I use up/down arrow first).

Dumluregn:
Hi @mmotyczynska , I did some actual changes in unit tests (they weren't working as I expected and I noticed they need some refactor - e.g. in `describe( 'appendColorPicker()', () => {` we don't need to create an editor at all). So please make a sanity check here.

mmotyczynska (approved):
LGTM :+1:

Issue #13985, opened 2023-04-26
Resolve Link plugin manual decorators quirks


lamuertepeluda:
## Context

I (am planning to) use the CKEditor5 as the editor of a headless CMS. Content editors need to create hyperlinks setting some attributes such as `target` and `rel`, whose values are especially important for SEO optimization.

To do so, I've configured some manual decorators of the LINK plugin following [the plugin guide](https://ckeditor.com/docs/ckeditor5/latest/features/link.html#configuration).

```js
		link: {
			addTargetToExternalLinks: false,
			decorators: {
				newtab: {
					mode: 'manual',
					label: 'Open in a new tab',
					attributes: {
						target: '_blank',
						rel: 'noopener noreferrer'
					}
				},
				nofollow: {
					mode: 'manual',
					label: 'Set SEO No Follow',
					attributes: {
						rel: 'nofollow'
					}
				},
				sponsored: {
					mode: 'manual',
					label: 'Set SEO Sponsored',
					attributes: {
						rel: 'sponsored'
					}
				}
			}
		}
```

My aim is to allow content editors to (manually) set any of the above, in any possible combination (i.e. I assume they know what they want to do and give them full control over the link).

## 📝 Provide detailed reproduction steps (if any)

1. configure a set of manual decorators such as in the example above
1. create a link in the editor by selecting some text, clicking on the link button in the toolbar and pasting an URL, e.g. "https://ckeditor.com/"
1. enable 2+ switches, e.g. `'Open in a new tab'` and `'Set SEO No Follow'`

### ✔️ Expected result

Given the html

```html
<p>Hello I am a link</p>
```

to which I link `https://ckeditor.com/` and use the first 2 decorators:

I want

```html
<p>Hello <a href="https://ckeditor.com/" target="_blank" rel="noopener noreferrer nofollow">I am a link</a></p>
```

as the editor downcast _and_ downcast output.

### ❌ Actual result

Instead I get

```html
<p>Hello <a href="https://ckeditor.com/" target="_blank" rel="noopener noreferrer"><a rel="nofollow">I am a link</a></a></p>
```

![view_wrong](https://user-images.githubusercontent.com/2249342/234591676-efac1a87-6011-4815-8d66-496844f0b864.png)

as both the editor downcast _and_ downcast output. Which is wrong and results in a broken link. I guess this is the result of the lack of "conflict resolution mechanism" cited in the docs.

Also the model view clearly shows something's off...

[...]

PR #13988, opened 2023-04-27
13985 -  Resolve Link plugin manual decorators quirks

lamuertepeluda:
Fix (link): Resolve Link plugin manual decorators quirks. Attempt to solve #13985.

See issue description

MAJOR BREAKING CHANGE (link): link attributes in decorators for standard html a tag attributes are now treated as strings rather than boolean


Starfox64:
This PR is still relevant, a review would be appreciated.

Witoso:
Work related to this is happening on our side in #17461. This one won't be merged, as we need to handle it a bit differently. Thanks for the proposal!

PR #14017, opened 2023-05-02
Color picker view should handle enter key

### LINKED ISSUE #3177: Should Shift-TAB on selected table go to the last table cell
From @scofalik [comment](https://github.com/ckeditor/ckeditor5-table/pull/10#discussion_r191156342):
> AFAICS, Shift + Tab moves selection to the first cell as well. How about moving it to the last cell in that case? Then, you could traverse the table from the end using Shift + Tab. Looks to me like this could make some sense.

illia-stv:

Internal(ui): Color picker form should accept enter key handling.
Closes [#3177](https://github.com/cksource/ckeditor5-internal/issues/3177).

---

### Additional information


pszczesniak (approved):
LGTM :+1:

PR #14029, opened 2023-05-04
Fix (heading, html-support, minimap, table, ui, word-count): Support multiple roots

### LINKED ISSUE #13596: Prevent plugins from crashing when used with multi-root editor due to using "default root"

Following plugins are not compatible with multi-root editor because they use "default root" (they assume there is just one root named `'main'`):

*   table resize,
*   word count,
*   title,
*   PDF/Word export,
*   block toolbar,
*   mini-map,
*   html comments.

There is also full page editing but I wouldn't consider integrating with it necessary. As for PDF/Word export, this can be integrated using custom data callback but it would be nice to improve it, so providing custom callback is not necessary.

Also, it would be good to search the code for `.getRoot()` and `.getDomRoot()` usages and maybe similar if you will stumble upon any.

kbinieda:

Fix: Added multi-root support for: title, HTML comments, table resize, word count and block toolbar.

### DIFF
--- a/packages/ckeditor5-heading/src/title.ts
+++ b/packages/ckeditor5-heading/src/title.ts
@@ -43,7 +43,7 @@ export default class Title extends Plugin {
 	 * A reference to an empty paragraph in the body
 	 * created when there is no element in the body for the placeholder purposes.
 	 */
-	private _bodyPlaceholder?: null | Element;
+	private _bodyPlaceholder = new Map<string, Element>();
 
 	/**
 	 * @inheritDoc
@@ -65,8 +65,6 @@ export default class Title extends Plugin {
 	public init(): void {
 		const editor = this.editor;
 		const model = editor.model;
-
-		this._bodyPlaceholder = null;
 
 		// To use the schema for disabling some features when the selection is inside the title element
 		// it is needed to create the following structure:
@@ -107,16 +105,16 @@ export default class Title extends Plugin {
 		editor.data.upcastDispatcher.on<UpcastElementEvent>( 'element:h2', dataViewModelH1Insertion, { priority: 'high' } );
 		editor.data.upcastDispatcher.on<UpcastElementEvent>( 'element:h3', dataViewModelH1Insertion, { priority: 'high' } );
 
-		// Take care about correct `title` element structure.
+		// // Take care about correct `title` element structure.
 		model.document.registerPostFixer( writer => this._fixTitleContent( writer ) );
-
-		// Create and take care of correct position of a `title` element.
+		//
+		// // Create and take care of correct position of a `title` element.
 		model.document.registerPostFixer( writer => this._fixTitleElement( writer ) );
-
-		// Create element for `Body` placeholder if it is missing.
+		//
+		// // Create element for `Body` placeholder if it is missing.
 		model.document.registerPostFixer( writer => this._fixBodyElement( writer ) );
-
-		// Prevent from adding extra at the end of the document.
+		//
+		// // Prevent from adding extra at the end of the document.
 		model.document.registerPostFixer( writer => this._fixExtraParagraph( writer ) );
 
 		// Attach `Title` and `Body` placeholders to the empty title and/or content.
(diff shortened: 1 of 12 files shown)


packages/ckeditor5-html-support/src/htmlcomment.ts
```diff
@@ -71,7 +79,7 @@ export default class HtmlComment extends Plugin {
 
 		// Remove comments' markers and their corresponding $root attributes, which are no longer present.
 		editor.model.document.registerPostFixer( writer => {
-			const root = editor.model.document.getRoot()!;
+			const root = editor.model.document.getRoot( editor.model.document.selection.getFirstRange()!.root.rootName! )!;
```
scofalik, line 82: This is incorrect. Every HTML-comment-related attribute should be set on the root that contains the comment.

AFAICS, `root` is used only at one place:

```javascript
writer.removeAttribute( marker.name, root );
```

It looks to me that there's nothing wrong with using `marker.getRange().root`.

packages/ckeditor5-html-support/src/htmlcomment.ts
```diff
@@ -44,7 +44,7 @@ export default class HtmlComment extends Plugin {
 		editor.conversion.for( 'upcast' ).elementToMarker( {
 			view: '$comment',
 			model: ( viewElement, { writer } ) => {
-				const root = this.editor.model.document.getRoot()!;
+				const root = this.editor.model.document.getRoot( this.editor.model.document.selection.getFirstRange()!.root.rootName! )!;
```
scofalik, line 47: This is incorrect. Every HTML-comment-related attribute should be set on the root that contains the comment. `root` here must be a specific model root, the one that holds the comment.

There are three solution candidates for this:

1.  Get the view root from `viewElement` and use `this.editor.editing.mapper` to map view root to model root.
2.  Get the view root from `viewElement` and check if it has `rootName` set correctly. I am not sure that we have a precise rule that respective view and model roots should have the same `rootName`. But that would make sense and if this comes up (i.e. somebody has a problem because their model and view root has different root names) I would recommend fixing the root names.
3.  Change this to "low level" converter and use `modelCursor` value, which is unfortunately not available in this upcast conversion helper. `modelCursor` is a position in model where "conversion takes place", so it would be enough to check `modelCursor.root`. But that would be a bigger change to this part of the code, so I am more inclined to go with solution 2 or 1.

packages/ckeditor5-heading/src/title.ts
```diff
@@ -160,7 +160,7 @@ export default class Title extends Plugin {
 		const editor = this.editor;
 		const data = editor.data;
 		const model = editor.model;
-		const root = editor.model.document.getRoot()!;
+		const root = editor.model.document.getRoot( this.editor.model.document.selection.getFirstRange()!.root.rootName! )!;
```
scofalik, line 163: General comment to whole title feature.

You either didn't check or didn't understand what this feature does, as the way it operates cannot operate on current selection state, as it doesn't make any sense. This feature is a bit more complex so it should raise some questions on your side.

Most importantly - if title plugin is used in multi-root editor, should the title be added to each root? I assumed yes, but maybe that's not so obvious. But for now, for simplification, let's assume yes. Later we may add a config value, where the integrator will specify in array root names, for which the title should be enabled.

So, the title plugin should work for all roots. Every root should have it's own title handling. Every root should manage the title (i.e. always keep title block and body block, this is what the postfixer does).

When `getTitle()` and `getBody()` methods are called, in options, `rootName` should be specified (for multi-root) and the returned content should based on that.

scofalik (changes requested):
Other than what I commented on, please add PDF/Word export integration (iterate through all roots to get the content) and add unit tests that will show that the features work correctly with multiple roots.

packages/ckeditor5-heading/src/title.ts
```diff
@@ -107,16 +105,16 @@ export default class Title extends Plugin {
 		editor.data.upcastDispatcher.on<UpcastElementEvent>( 'element:h2', dataViewModelH1Insertion, { priority: 'high' } );
 		editor.data.upcastDispatcher.on<UpcastElementEvent>( 'element:h3', dataViewModelH1Insertion, { priority: 'high' } );
 
-		// Take care about correct `title` element structure.
+		// // Take care about correct `title` element structure.
```
scofalik, line 108:
```suggestion
		// Take care about correct `title` element structure.
```

arkflpc (approved):

scofalik (approved):

PR #14031, opened 2023-05-04
Release process: prepare JavaScript and typings

### LINKED ISSUE #13950: Redesign the release process
### 🎯 Goals

- Automate the process so it can be done with as little interaction from the Releaser as possible.
- Make the process bulletproof so more people can trigger it. Thus: No need for special planning.
- To enable publishing nightly CKEditor 5 packages.

### 💡 Hints

- Instead of implementing big tasks that do everything and are hard to introduce a new step into it, let's create small utils that can be glued and extendable.
- We can drop the current tools and create new ones from scratch. The old tools can be used as inspiration.

### LINKED ISSUE #13970: Release process: Prepare JS for the `ckeditor5-*` packages

**🎯 The goal is to prepare CKEditor 5 feature packages to be ready to publish on npm.**

- This step should be executed in the release directory.
- We do not want to publish TypeScript sources. Hence, compile them to create JavaScript and declaration files.
- The `"main"` field in a `package.json` should point to a JavaScript file instead of TS.
- TypeScript understands the `"types"` property. Hence, let’s add it. It should point to an entry point of typings.
- Remove TypeScript sources from packages to enable preparing DLLs from the JavaScript code.
- Bonus point: Could we execute the preparation step concurrently to improve building time?
- Possible arguments:
    - a directory where to look for packages (string)
- Ideas:
	- Perhaps this script could land in `ckeditor5-dev-*`
	- It depends on the API. Could we reuse a core of the feature when we prepare DLLs? (#13971)

---

pomek:

Internal: Release process: prepare JavaScript and typings. Closes #13970.

---

### Additional information

- [x] Requires: https://github.com/ckeditor/ckeditor5-dev/pull/869.

### DIFF
--- a/scripts/release/prepare-packages.js
+++ b/scripts/release/prepare-packages.js
@@ -9,20 +9,9 @@
 
 'use strict';
 
-const readline = require( 'readline' );
 const releaseTools = require( '@ckeditor/ckeditor5-dev-release-tools' );
 
 const abortController = new AbortController();
-
-// Windows does not understand CTRL+C attached to `process`.
-if ( process.platform === 'win32' ) {
-	const rl = readline.createInterface( {
-		input: process.stdin,
-		output: process.stdout
-	} );
-
-	rl.on( 'SIGINT', () => process.emit( 'SIGINT' ) );
-}
 
 process.on( 'SIGINT', () => {
 	abortController.abort( 'SIGINT' );
--- a/scripts/release/prepare-packages.js
+++ b/scripts/release/prepare-packages.js
@@ -1,0 +1,82 @@
+#!/usr/bin/env node
+
+/**
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/* eslint-env node */
+
+'use strict';
+
+const readline = require( 'readline' );
+const releaseTools = require( '@ckeditor/ckeditor5-dev-release-tools' );
+// const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
+
+const abortController = new AbortController();
+
+// Windows does not understand CTRL+C attached to `process`.
+if ( process.platform === 'win32' ) {
+	const rl = readline.createInterface( {
+		input: process.stdin,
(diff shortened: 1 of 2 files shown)


scripts/release/prepare-packages.js
```diff
@@ -0,0 +1,81 @@
...
+		packagesDirectory: 'release',
+		processDescription: 'Compiling TypeScript...',
+		signal: abortController.signal,
+		taskToExecute: compileTypeScriptCallback
+	} );
+
+	/**
+	 * @param {String} packagePath
+	 */
+	function compileTypeScriptCallback( packagePath ) {
```
martnpaneq, line 42: I think this function should also be stored in our tools. Typescript compilation is going to be called by 'collaboration-features` and `ckeditor5-internal` packages too.

scripts/release/prepare-packages.js
```diff
@@ -0,0 +1,45 @@
...
+
+// Windows does not understand CTRL+C attached to `process`.
+if ( process.platform === 'win32' ) {
+	const rl = readline.createInterface( {
+		input: process.stdin,
+		output: process.stdout
+	} );
+
+	rl.on( 'SIGINT', () => process.emit( 'SIGINT' ) );
+}
```
psmyrek, line 25: This works weird on Windows:
- I have the impression that I have to press `CTRL+C` several times and wait a few seconds for the script to be terminated.
- Additionally, if there is no `release` directory, the scripts hangs:
   


The good news is we don't need this whole `if` at all 😄 

Question: whether we have the `readline.createInterface()` or not, terminated script throws this error. Is it expected?

obraz

pomek: The unhandled rejection error should be fixed in https://github.com/ckeditor/ckeditor5-dev/pull/869/commits/828a7e214b9688f62650ea739b0b84c4099f4c25.

psmyrek (approved):
LGTM 🎉 

It works as expected when `@ckeditor/ckeditor5-dev-utils` and `@ckeditor/ckeditor5-dev-release-tools` are linked in CKE5 repository.

scripts/release/prepare-packages.js
psmyrek: Thanks, it works fine now.

PR #14037, opened 2023-05-04
Fix output color format in color picker

Dumluregn:

Internal(font): Fix output format in color picker.

---

### Additional information

The flow causing the bug was as follows:

1.  Color is picked from color picker. It's in `hex` format. Property `color` in color picker is set.
2.  On `color` property change, it's converted to desired output format (HSL by default). This fires another nested `.on( 'change:color' )` call. The first one is on hold - the `_hexColor` property is not updated yet.
3.  The second color change is propagated to `colorTableView` and higher. Formatting in HSL format is set in the model.
4.  Afterwards, the first `color` property change is resumed. It re-triggers the changes in `colorTableView`, but this time with the older value - in `hex` format. It overwrites the previously set model property. In color picker everything seems to be fine - `color` property is in HSL format.

To fix this, I put the logic `color` prop logic to `.on( 'set:color' )` listener instead (with minor adjustments). Now we no longer have the duplicated color change events.

### DIFF
--- a/packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
+++ b/packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
@@ -104,16 +104,9 @@ export default class ColorPickerView extends View {
 		} );
 
 		// Sets color in the picker if color was updated.
-		this.on( 'set:color', ( evt, propertyName, newValue, oldValue ) => {
-			const convertedColor = convertColor( newValue, this._format );
-
-			if ( convertedColor != oldValue ) {
-				evt.return = convertedColor;
-
-				return;
-			}
-
-			evt.return = oldValue;
+		this.on( 'set:color', ( evt, propertyName, newValue ) => {
+			// The color needs always to be kept in the output format.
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -104,13 +104,19 @@ export default class ColorPickerView extends View {
 		} );
 
 		// Sets color in the picker if color was updated.
-		this.on( 'change:color', () => {
-			const convertedColor = convertColor( this.color, this._format );
+		this.on( 'set:color', ( evt, propertyName, newValue, oldValue ) => {
+			const convertedColor = convertColor( newValue, this._format );
+
+			if ( convertedColor != oldValue ) {
```
mlewand: Since in `else` the values are equal - the later `evt.return = oldValue;` is equivalent of `evt.return = convertedColor;`. So in 5dda2d7bbf I simplified this path.

mlewand (approved):
Nice fix :+1:

Issue #11738, opened 2022-05-09
Content pasted from WPS renders as blank with PasteFromOffice plugin


slotterbackW:
## 📝 Provide detailed reproduction steps (if any)

1. Paste the following content into an instance of CKEditor which has the [PasteFromOffice](https://ckeditor.com/docs/ckeditor5/latest/api/module_paste-from-office_pastefromoffice-PasteFromOffice.html) plugin included.
```
[...]

Issue #2488, opened 2019-10-09
Support for pasting content from WPS


516821202:
I hope to copy and paste pictures and text in WPS software

songlin51:
Is WPS not currently supported. bad news

jodator:
Another request for WPS support: #7937.

jodator:
@516821202 I've edited the main issue content to describe what is needed to implement this and to allow other users find or understand it better.

xunyidaixu:
Most Chinese use WPS，I hope to support the paste function

add2ws:
> Most Chinese use WPS，I hope to support the paste function

急需这个功能

bamf2077:
5 years later....

leiyangs:
Will this problem still exist in 2026?

PR #14039, opened 2023-05-05
Fix content pasted from WPS renders as blank with PasteFromOffice plugin. See #11738

hechengxixi:

Type: Message. Closes #000.

---

### Additional information


523299338:
dose the pr can be merged?

yoyo837:
The ckeditor teams official seems not to consider compatibility with WPS for the time being. Can we make it compatible by writing our own plug-ins?

yoyo837:
ref: https://github.com/ckeditor/ckeditor5/issues/2488 https://github.com/ckeditor/ckeditor5/issues/11738

yoyo837:
This PR does not fix the image pasting problem.

yoyo837:
WPS does not provide `text/rtf` data to the clipboard, nor does it provide `base64` in `text/html`. `` only has a local file path, and js cannot do anything about it.

huyu456 (approved):

hechengxixi:
> `WPS` does not provide `text/rtf` data to the clipboard, nor does it provide `base64` in `text/html`. `` only has a local file path, and js cannot do anything about it. image
> 
> `text/html` data in clipboard after `WPS` paste: image

Maybe this is another issuse #16397

yoyo837:
> Maybe this is another issuse #16397

Windows 11 only? I can see the same problem on Windows 10.

Reinmar:
Apologies for lack of repsonse from our side. We should have been more active :( 

We'd like to look into this PR. However, without the CLA signed by the author (@hechengxixi) we can't accept this contribution. So that's the prerequisite to include this change.

yoyo837:
ping @hechengxixi

yoyo837:
> Apologies for lack of repsonse from our side. We should have been more active :(
> 
> We'd like to look into this PR. However, without the CLA signed by the author (@hechengxixi) we can't accept this contribution. So that's the prerequisite to include this change.

It seems that this contributor has not made any new actions for a long time. If some other one else creates a new PR contribution with the same content as this PR and signs the CLA, would you consider accepting it?

A-Kun:
Attempted reaching out to the author by email.

PR #14055, opened 2023-05-08
Release date in CKEditor 5

pomek:

Internal (utils): Added the `releaseDate` property containing the latest CKEditor 5 release date.

### DIFF
--- a/packages/ckeditor5-utils/src/version.ts
+++ b/packages/ckeditor5-utils/src/version.ts
@@ -15,6 +15,7 @@ const version = '37.1.0';
 
 export default version;
 
+// The second argument is not a month. It is `monthIndex` and starts from `0`.
 export const releaseDate = new Date( 2023, 3, 19 );
 
 /* istanbul ignore next -- @preserve */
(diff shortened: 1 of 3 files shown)


psmyrek (approved):
LGTM 🎉

packages/ckeditor5-utils/src/version.ts
```diff
@@ -15,6 +15,8 @@ const version = '37.1.0';
 
 export default version;
 
+export const releaseDate = new Date( 2023, 3, 19 );
```
mlewand, line 18: It is a good idea to leave reminder in inline comment that month is 0-based. I personally remember that one of `Date()` arguments is 0-based but I never remember which one 😂

mlewand (approved):

packages/ckeditor5-utils/src/version.ts
pomek, line 18:
```suggestion
// The second argument is not a month. It is `monthIndex` and starts from `0`.
export const releaseDate = new Date( 2023, 3, 19 );
```

PR #14056, opened 2023-05-08
Initial implementation of the PBC

oleq:

Internal: Initial implementation of the powered-by feature. Closes cksource/ckeditor5-internal#3216. Closes cksource/ckeditor5-internal#3219.


packages/ckeditor5-ui/src/editorui/poweredby.ts
```diff
@@ -0,0 +1,292 @@
...
+export default class PoweredBy extends DomEmitterMixin() {
+	/**
+	 * A reference to the view displaying a link with a label and a project logo.
+	 */
+	private [ POWERED_BY_VIEW_SYMBOL ]: PoweredByView | null;
+
+	/**
+	 * A reference to the balloon panel hosting and positioning the "powered by" view.
+	 */
+	private [ POWERED_BY_BALLOON_SYMBOL ]: BalloonPanelView | null;
```
oleq, line 48: These make it quite impossible to test the feature. I'm wondering if we could make these properties public since CSS classes are already there allowing manipulation or disabling of the feature.

mlewand (commented):
Looks good, few questions and we're good to go.

mlewand:
Also pushed a minor local const extraction in cb5f928.

mlewand:
If you could also make the tag unselectable, so that it can't be dragged and dropped onto the content that would be great. Otherwise can be extracted to a followup.

oleq:
> If you could also make the tag unselectable, so that it can't be dragged and dropped onto the content that would be great. Otherwise can be extracted to a followup.

Done in [36c7263](https://github.com/ckeditor/ckeditor5/pull/14056/commits/36c72632b233ac4e2816dc0879d63e093cab4ca5).

> We have to throttle this event. There are cases where it could be called many times, for example during image resize.

Done in [2fe048e](https://github.com/ckeditor/ckeditor5/pull/14056/commits/2fe048ecc61f1c0728149316512541a092986f8c).

mlewand (approved):
All right, looking good.

I took the liberty to make sure the first render happens on the leading edge of UI's update render.

Reason is that we want to have the first render as fast as possible. Pushed in ea0d361.

Issue #3162, opened 2019-10-09
Minimal table support


jodator:
I'd like to propose incremental work on table feature as it will take some time and effort to fully develop even in MVP form.

As the minimal table support I'd like to propose:
- Table conversion (basic upcast/downcast that will not destroy existing data) with:
    - `tbody`/`thead`/`tfoot` parsing,
    - `tr`,`td`,`th` handling,
    - attributes parsing (at least `colspan` & `rowspan`).
- Insert table button (probably temporary) that will insert 2x2 table.
- Insert table command (maybe with parameters already like rows & columns).

And basically that. No cursor support, no better UI, no widget and probably buggy in one way or another but I think that setting small steps along the way might be useful and probably better to review during that time rather one big PR.

PR #14060, opened 2023-05-08
Group changes into one batch

illia-stv:

Feature: Group changes into one batch. Closes [#3162](https://github.com/cksource/ckeditor5-internal/issues/3162).

---

### Additional information

### DIFF
--- a/packages/ckeditor5-font/src/ui/colorui.ts
+++ b/packages/ckeditor5-font/src/ui/colorui.ts
@@ -21,7 +21,7 @@ import {
 	type default as ColorTableView,
 	type ColorTableExecuteEvent,
 	type ColorTableCancelEvent,
-	type ShowColorPickerEvent
+	type ColorTableShowColorPickerEvent
 } from './colortableview';
 import type FontColorCommand from '../fontcolor/fontcolorcommand';
 import type FontBackgroundColorCommand from '../fontbackgroundcolor/fontbackgroundcolorcommand';
@@ -165,7 +165,7 @@ export default class ColorUI extends Plugin {
 				}
 			} );
 
-			this.colorTableView.on<ShowColorPickerEvent>( 'showColorPicker', () => {
+			this.colorTableView.on<ColorTableShowColorPickerEvent>( 'showColorPicker', () => {
 				this._undoStepBatch = editor.model.createBatch();
 			} );
 
--- a/packages/ckeditor5-font/src/ui/colorui.ts
+++ b/packages/ckeditor5-font/src/ui/colorui.ts
@@ -66,7 +66,7 @@ export default class ColorUI extends Plugin {
 	public colorTableView: ColorTableView | undefined;
 
 	/**
-	 * Keeps all changes of color picker in one batch.
+	 * Keeps all changes in color picker in one batch while dropdown is open.
 	 */
 	declare private _undoStepBatch: Batch;
 
--- a/packages/ckeditor5-font/src/ui/colorui.ts
+++ b/packages/ckeditor5-font/src/ui/colorui.ts
@@ -64,7 +64,7 @@ export default class ColorUI extends Plugin {
 	 */
 	public colorTableView: ColorTableView | undefined;
 
-	private _undoStepBatch: Batch;
+	private undoStepBatch: Batch;
 
 	/**
 	 * Creates a plugin which introduces a dropdown with a pre–configured {@link module:font/ui/colortableview~ColorTableView}.
@@ -87,7 +87,7 @@ export default class ColorUI extends Plugin {
 	) {
 		super( editor );
 
-		this._undoStepBatch = editor.model.createBatch();
+		this.undoStepBatch = editor.model.createBatch();
 		this.commandName = commandName;
 		this.componentName = componentName;
 		this.icon = icon;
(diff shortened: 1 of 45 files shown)


packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -63,6 +64,8 @@ export default class ColorUI extends Plugin {
 	 */
 	public colorTableView: ColorTableView | undefined;
 
+	private undoStepBatch: Batch;
```
Dumluregn, line 67: This prop is private (and it should be), so it should start with an underscore. Also, it missed API docs.

packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -84,6 +87,7 @@ export default class ColorUI extends Plugin {
 	) {
 		super( editor );
 
+		this.undoStepBatch = editor.model.createBatch();
```
Dumluregn, line 90: There are 4 places where you create a new batch:

1.  Here, when ColorUI plugin is initialised.
2.  On Save button execution.
3.  On Cancel button execution.
4.  On dropdown closure.

Correct me if I'm wrong, but instead of all that, we could reset a batch just whenever a dropdown is being open (option A), or even only whenever the color picker view is shown (option B). That's much easier to maintain and less error-prone.

To implement option A, we need to create a batch in this tree:

```
dropdownView.on( 'change:isOpen', ( evt, name, isVisible ) => {
				if ( !dropdownContentRendered ) {
					dropdownContentRendered = true;
					dropdownView.colorTableView!.appendUI();
				}

				if ( isVisible ) {
				[create batch here...]
```

And then we can remove all the other calls.

Option B is a bit more complex as we'd need to create a new event (called e.g. `showColorPicker`), fire it by `colorGridsPageView`, delegate it to `ColorTableView`, and then in `ColorUI` plugin listen on `showColorPicker` event fired by `colorTableView` and only then reset the batch.

Dumluregn (changes requested):
Functionally it's ok (solution propagated from table cell properties UI), but let's improve its architecture a bit. Tests will need to be updated probably.

packages/ckeditor5-font/src/ui/colorui.ts
illia-stv: `colorGridsPageView ` doesnt know nothing about when color picker is shown, but we already have in `colorTableView ` method `showColorPicker ` which is called when we need to show color picker so there is no need to delegate it.

packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -63,6 +64,11 @@ export default class ColorUI extends Plugin {
 	 */
 	public colorTableView: ColorTableView | undefined;
 
+	/**
+	 * Keeps all changes of color picker in one batch.
```
Dumluregn, line 68:
```suggestion
	 * Keeps all changes in color picker in one batch while dropdown is open.
```

packages/ckeditor5-font/tests/fontcommand.js
```diff
@@ -224,16 +224,13 @@ describe( 'FontCommand', () => {
...
+			const batch = model.createBatch();
+			const spy = sinon.spy( model, 'enqueueChange' );
 
-			model.change( writer => {
-				expect( writer.batch.operations.length ).to.equal( 0 );
-				command.execute( { value: 'foo' } );
-				expect( writer.batch.operations.length ).to.equal( 1 );
-			} );
-
-			expect( getData( model ) ).to.equal( '<paragraph>a[<$text font="foo">bcfo]obar</$text>xyz</paragraph>' );
```
Dumluregn, line 236: I think having both tests (the old one and new one) makes sense - depending on whether you've passed the batch as a parameter.

packages/ckeditor5-font/tests/ui/colorui.js
```diff
@@ -194,18 +195,50 @@ describe( 'ColorUI', () => {
...
+					source: 'colorPicker'
+				} );
 
 				dropdown.colorTableView.colorPickerPageView.cancelButtonView.fire( 'execute' );
 
-				expect( dropdown.colorTableView.selectedColor ).to.equal( '#ff0000' );
+				sinon.assert.calledOnce( spyUndo );
+			} );
+
+			it( 'should create new batch when color picker is showed', () => {
```
Dumluregn, line 218: I can't find the test that actually counts the undo steps if there were more than one color change. Could you add it?

Dumluregn (changes requested):

packages/ckeditor5-font/src/ui/colorui.ts
Dumluregn: We discussed this and though it admittedly works, it cuts some corners and may be limiting the granularity, so we'll stay with the `colorGridsPageView` being the original emitter.

packages/ckeditor5-font/tests/integration.js
```diff
@@ -20,7 +21,7 @@ describe( 'Integration test Font', () => {
 
 		return ClassicTestEditor
 			.create( element, {
-				plugins: [ Font, ArticlePluginSet ],
+				plugins: [ Font, ArticlePluginSet, Undo ],
```
Dumluregn: `Undo` is included in `Essentials` which are included in `ArticlePluginSet`.

Dumluregn (approved):

PR #14068, opened 2023-05-08
Color picker documentation

Dumluregn:

Internal(font): Add color picker feature documentation.

---

### Additional information

I wasn't able to create a link to [http://192.168.100.6:8080/ckeditor5/37.1.0/api/module\_ui\_colorpicker\_utils-ColorPickerConfig.html#member-format](http://192.168.100.6:8080/ckeditor5/37.1.0/api/module_ui_colorpicker_utils-ColorPickerConfig.html#member-format) - ``{@link module:ui/colorpicker/utils~ColorPickerConfig#format `config.fontColor.colorPicker.format`}`` , which seems correct, remained unconverted. Is there any issue with type members link conversion? I didn't manage to find a similar example in docs unfortunately.

### DIFF
--- a/packages/ckeditor5-font/docs/features/font.md
+++ b/packages/ckeditor5-font/docs/features/font.md
@@ -9,7 +9,7 @@ The font feature lets you change font family, size, and color (including backgro
 
 ## Demo
 
-Use the toolbar dropdowns in the demo below to control the font size {@icon @ckeditor/ckeditor5-font/theme/icons/font-size.svg Font size} and font family {@icon @ckeditor/ckeditor5-font/theme/icons/font-family.svg Font family}. You can also change both the font color {@icon @ckeditor/ckeditor5-font/theme/icons/font-color.svg Font color} and the font background color {@icon @ckeditor/ckeditor5-font/theme/icons/font-background.svg Font background color}.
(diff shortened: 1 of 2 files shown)


godai78:
Follow-up: the bozo link to existing API doclet. Check if it reports in the main branch after merge.

packages/ckeditor5-font/docs/features/font.md
```diff
@@ -355,6 +355,35 @@ ClassicEditor
 	.catch( /* ... */ );
 ```
 
+### Color picker
+
+Colors from outside of the preconfigured palette can be set using the "Color picker" option available at the bottom of the color selection dropdown.
+
+Color picker applies colors in the **HSL** format as it is default for the font color feature. This can be changed using the {@link module:ui/colorpicker/utils~ColorPickerConfig#Format `config.fontColor.colorPicker.format`} option. Available color formats are defined in the {@link module:ui/colorpicker/utils~ColorPickerOutputFormat} type. Note that this change will not affect the color input &ndash; it always accepts only values given in the `hex` format (with or without the `#` sign at the beginning).
```
illia-stv, line 362: > (with or without the `#` sign at the beginning)
To my mind this information is not necessary for developers because it changes nothing. This info is just nice to know for actual users which will use that but not for developers. It just my mind, so it could actually stay.

illia-stv (approved):

illia-stv:
I like👍. I left one tiny note but it’s just a suggestion.

packages/ckeditor5-font/docs/features/font.md
Dumluregn: I think actual users may also sometimes read the documentation, or ask dev to do it for them :stuck\_out\_tongue\_winking\_eye: so let's keep it, it's just a few extra words.

PR #14071, opened 2023-05-09
Color picker for font colors

### LINKED ISSUE #10988: Color Picker

Introduce Color Picker for features where is it is possible to define color or background color so that users can insert color they want:
obraz

---

mlewand:

Feature(font): Introduced color picker feature. Closes #10988.

---

### Additional information


packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	/**
+	 * Creates collection of sliders in color picker.
+	 *
+	 * @private
+	 */
+	private _createSlidersView(): void {
+		const colorPickersChildren = [ ...this.picker.shadowRoot!.children ] as Array<HTMLElement>;
+		const sliders = colorPickersChildren.filter( item => item.role === 'slider' );
```
Reinmar, line 177: This filtering doesn't work on Firefox. It returns an empty array. It causes https://github.com/cksource/ckeditor5-internal/issues/3243

Reinmar: It seems enough to change `.role` to `.getAttribute( 'role' )` to make it work. It works in Chrome and Firefox.

mlewand: Thanks! Indeed it did the trick.

PR #14080, opened 2023-05-10
Update the verify function to cover all possible cases.

pszczesniak:

Internal (utils) : Should cover all edge cases in token verification.

Closes https://github.com/cksource/ckeditor5-internal/issues/3232.

---

### Additional information

All cases are covered 👉 https://github.com/cksource/ckeditor5-internal/issues/3232#issuecomment-1540224260

### DIFF
--- a/packages/ckeditor5-utils/src/madewith.ts
+++ b/packages/ckeditor5-utils/src/madewith.ts
@@ -21,6 +21,14 @@ export type VerifiedKeyStatus = 'VALID' | 'INVALID';
  * @returns String that represents the state of given `token` parameter.
  */
 export default function verify( token: string ): VerifiedKeyStatus {
+	function oldTokenCheck( token: string ): VerifiedKeyStatus {
+		if ( token.match( /^[a-zA-Z0-9+/=$]+$/g ) && ( token.length >= 40 && token.length <= 255 ) ) {
+			return 'VALID';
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-utils/src/madewith.ts
```diff
@@ -36,19 +44,33 @@ export default function verify( token: string ): VerifiedKeyStatus {
...
 	const secondElement = splittedDecryptedData[ 1 ];
 
+	if ( !secondElement ) {
+		return oldTokenCheck( token );
+	}
+
 	try {
-		// Must be a valid format.
-		atob( firstElement );
+		atob( secondElement );
```
mlewand, line 52: Why using second element at all? It's always going to be the last part.

mlewand (approved):
We'll have to simplify this logic, because it's difficult to follow but we'll do that in a followup not to block an issue.

packages/ckeditor5-utils/src/madewith.ts
pszczesniak: Because we're checking old vs new licence key - old licence keys after `atob` can contain `-` and it will split the string into array so it can look like a new licence key format. To verify that we need to check if second element is base64 coded string - if yes it's a new licence key format if not it's and old one.

Issue #3236, opened 2019-10-09
Blockquote applies changes to all cells while table is selected


Mgsy:
## Steps to reproduce
1. Create a new table.
1. Select it.
1. Apply a blockquote.

## Current result
The blockquote is inside all cells. I believe it should be applied to the whole table. 

## GIF
![bug_cke5](https://user-images.githubusercontent.com/20988892/46015781-6b8f8380-c0d3-11e8-98fd-f0067922c2bc.gif)

PR #14089, opened 2023-05-10
Color picker input does not retain its value

illia-stv:

Fix: Color picker input does not retain its value. Closes [#3236](https://github.com/cksource/ckeditor5-internal/issues/3236).

---

### Additional information


illia-stv:
To resolve this issue I suggest reset color when input value is empty, it's seems for logical, because it reflects state of input in the view.

Dumluregn:
As we discussed, this PR doesn't fix the original issue as it can still be reproducible by removing _part_ of the input instead of removing all of it (e.g. leaving only the first character of the hex format). It has an unimpeachable value though regarding the empty input handling. So please report a separate issue for it.

Dumluregn:
Closing in favor of #14153.

Dumluregn:
> As we discussed, this PR doesn't fix the original issue as it can still be reproducible by removing _part_ of the input instead of removing all of it (e.g. leaving only the first character of the hex format). It has an unimpeachable value though regarding the empty input handling. So please report a separate issue for it.

Reported in #14154.

PR #14093, opened 2023-05-10
Added `validateDependenciesVersions()` function as a part of the new release process.

### LINKED ISSUE #13973: Release process: Checks before publishing (validate the release)

**🎯 The goal is to verify if all CKEditor 5 packages are specified and require themselves to use the same version.**

- Example:
    - Release: 38.0.0
        - `engine#version`: 38.0.0, `engine#depenedencies#ckeditor5-core`: ^37.0.0 - NOK
        - `engine#version`: 38.0.0, `engine#depenedencies#ckeditor5-core`: ^38.0.0 - OK
    - Release: 0.0.0-nightly-20230425.0
        - `engine#version`: 0.0.0-nightly-20230425.0, `engine#dependencies#ckeditor5-core`: ^0.0.0-nightly-20230425.0 - OK
        - `engine#version`: 0.0.0-nightly-20230425.0, `engine#dependencies#ckeditor5-core`: `@nightly` - NOK
        - `engine#version`: 0.0.0-nightly-20230425.0, `engine#dependencies#ckeditor5-core`: ^37.0.0 - NOK
- Possible arguments:
	- a version (string)
	- a directory where to look for packages (string)
---

martnpaneq:

Internal: Added `validateDependenciesVersions()` function as a part of the new release process. Closes #13973.

---

### Additional information


psmyrek (changes requested):
It looks like it does not work correctly, because it compares versions strictly:

obraz

martnpaneq:
We agreed that the dependency version should not be an exact match with the argument `version`, but rather we should check if it matches semantically following semver rules.
For example if the argument is `37.0.0`, and the dependency version is `^37.0.0` or `~37.0.0`, no errors should be reported.

scripts/release/validatedependenciesversions.js
```diff
@@ -0,0 +1,78 @@
...
+const chalk = require( 'chalk' );
+const semver = require( 'semver' );
+const { normalizeTrim } = require( 'upath' );
+const isCKEditor5Package = require( './isckeditor5package' );
+
+/**
+ * Validates if the versions of packages and their dependencies in specified directory match the provided version.
+ *
+ * @param {Object} options
+ * @param {String} releaseDirectory Path to directory with packages to validate.
```
psmyrek, line 23:
```suggestion
 * @param {String} options.releaseDirectory Path to directory with packages to validate.
```

scripts/release/validatedependenciesversions.js
```diff
@@ -0,0 +1,78 @@
...
+ * @param {String} options.version Version that all packages and their dependencies need to match.
+ */
+module.exports = async function validateDependenciesVersions( { releaseDirectory, version } ) {
```
psmyrek, line 26: In other scripts we use the `packagesDirectory` name for the place where all packages to process are located, or `packagesDirectory` + `outputDirectory` if both locations are needed. What do you think about changing the name `releaseDirectory` → `packagesDirectory`?

martnpaneq: Renamed :)

psmyrek (approved):
LGTM 🎉

Issue #14135, opened 2023-05-12
Release process: make sure all tasks are asynchronous


pomek:
## 📝 Provide a description of the improvement

To improve the UX layer of the release process (#14134), all pieces (the release process sub-tasks) should be asynchronous code.

### Scope

* Make sure the following steps execute return a promise:
	* ckeditor5-dev: [`updateDependencies()`](https://github.com/ckeditor/ckeditor5-dev/blob/40b3ad50de561f59d34dce4a073a9dbec667eef0/packages/ckeditor5-dev-release-tools/lib/tasks/updatedependencies.js#LL32C27-L32C45)
	* ckeditor5-dev: [`updateVersions()`](https://github.com/ckeditor/ckeditor5-dev/blob/73b30729ebd57402b93aa06c1fc75abaeaf627dc/packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js#L32)
	* ckeditor5-dev: [`executeInParallel()`](https://github.com/ckeditor/ckeditor5-dev/blob/e21ae2544ce0361afa8cffcf4ab71bfdf88ddad4/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js#LL37C1-L37C1)
	* ckeditor5-dev: [`prepareRepository()`](https://github.com/ckeditor/ckeditor5-dev/blob/5cfc654bcc6af45e1b21e0bd35abb78ae2b3f28e/packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js#L26)
	* ckeditor5-dev: [`cleanUpPackages()`](https://github.com/ckeditor/ckeditor5-dev/blob/3758f5fedbc6d0087f6ecfbd06fcc231da814abc/packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js#L30)
	* ckeditor5-dev: [`commitAndTag()`](https://github.com/ckeditor/ckeditor5-dev/blob/d93bab5fb3eeb99dd58c5775a24ef193ede3bab5/packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js#L20)
	* ckeditor5: [`updateVersionReferences()`](https://github.com/ckeditor/ckeditor5/blob/4004bf0d77233b1f51431bef7241d87480c0b5fb/scripts/release/update-version-references.js#L25)
[...]

PR #14157, opened 2023-05-15
 Make release tools asynchronous

pomek:

Internal: Make release tools asynchronous. Closes ckeditor/ckeditor5#14135.


scripts/release/prepare-packages.js
psmyrek, line 1: The following functions are still synchronous:
```
getCKEditor5Dependencies()
compileTypeScriptCallback()
updatePackageEntryPoint()
prepareDllBuildsCallback()
```

psmyrek: Also, there is `await releaseTools.getLastFromChangelog()` but this function is synchronous. Not sure if we should care about this.

PR #14207, opened 2023-05-22
Remove the PoweredBy slightly move on initial editor focus.

pszczesniak:

Fix (ui): Powered by should not moves slightly when the editor is focused for the first time.

Closes https://github.com/cksource/ckeditor5-internal/issues/3289.

---

### Additional information

_There were some problems with correct position calculation when "poweredBy" had 1px border. The move was caused by `throttle` function._


mabryl:
Just tested this in our manual tests and I can confirm that the issue is no longer present.

PR #14210, opened 2023-05-23
`Rect#getVisible()` should not consider ancestors when the target is an element with `position: absolute`

### LINKED ISSUE #14107: Rect#getVisible() does not work well with absolutely positioned elements

Consider the following structure:

```
<div class="parent"> (e.g. body collection container)
    <div class="child" style="position: absolute"> (e.g. a floating balloon element)
```

then call `new Rect( child ).getVisible()`

### ✔️ Expected result

Since the child has `position: absolute` and the parent has no `position: relative`, the child can move freely unobstructed by the parent. It will not get cropped.

But `getVisible()` will consider the parent nevertheless.

### ❌ Actual result

The parent is ignored by `getVisible()` logic.

### ❓ Possible solution

The logic of `getVisible()` should understand whether the child will be cropped or not. See this fiddle to learn more https://jsfiddle.net/4wnxhjf1/47/.

---

oleq:

Fix (utils): `Rect#getVisible()` should not consider ancestors when the target is an element with `position: absolute`. Closes #14107. Closes https://github.com/cksource/ckeditor5-internal/issues/3264.


pszczesniak (approved):
LGTM :+1: 

I did some performance test how fast is `getComputedStyle` and it looks that it works quite fast. I used this piece of code:

```
console.time('getComputedStyle');
for (let i = 0; i < 10000; i++) {
    const cs = window.getComputedStyle( document.body );
}
console.timeEnd('getComputedStyle');
```

For 10k checks it took: `avg 4.5 ms`

For 100k -> `avg 38 ms`

I think that those scores are quite nice, but maybe would be good to check it on a slower machine than M1 :thinking:

oleq (commented):
Self.R-: The logic is still broken. It should compare parent-pairs for the position:absolute/relative/overflow relationship instead of rect source with successive parents.

Dumluregn (changes requested):
See above.

oleq:
Whew, this logic is pretty tricky. I added a few cases to the manual test + an improved implementation.

Dumluregn (approved):
LGTM, the code makes sense. Manual and unit tests are fine too. Well done, it must have been quite a ride to come to this.

I discovered one minor issue regarding Safari and Firefox, but they won't be visible in docs anyway, so I extracted it.

PR #14215, opened 2023-05-23
Fixed key behaviour in color picker

illia-stv:

Fix: Fixed key behaviour in color picker. Closes #[14218](https://github.com/ckeditor/ckeditor5/issues/14218).

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
+++ b/packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
@@ -114,11 +114,8 @@ export default class ColorPickerView extends View {
 		} );
 
 		this.on( 'change:_hexColor', () => {
-			const input = this.hexInputRow!.children.get( 1 ) as LabeledFieldView<InputTextView>;
-
-			// Should update color in color picker when change was caused by input or when
-			// color picker is not visible.
-			if ( input.isFocused || this.picker.offsetParent === null ) {
+			// Should update color in color picker when its not focused
+			if ( document.activeElement !== this.picker ) {
 				this.picker.setAttribute( 'color', this._hexColor );
 			}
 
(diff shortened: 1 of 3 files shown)


illia-stv:
There is no need in setting color attribute in color picker component here 

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts#L117

mlewand (changes requested):
Be sure to add unit tests

Dumluregn (changes requested):
It solves the base issue, but breaks the connection between the input and color palette. Try setting any color by slider and then pasting `69b2b5` in the input - it should update the selected color, but it doesn't.

packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -114,7 +114,13 @@ export default class ColorPickerView extends View {
 		} );
 
 		this.on( 'change:_hexColor', () => {
-			this.picker.setAttribute( 'color', this._hexColor );
+			const input = this.hexInputRow!.children.get( 1 ) as LabeledFieldView<InputTextView>;
+
+			// Should update color in color picker when change was caused by input or when
+			// color picker is not visible.
+			if ( input.isFocused || this.picker.offsetParent === null ) {
```
Dumluregn, line 121: I think we could simplify this check and also make it more intuitive (looking at you, `this.picker.offsetParent === null`). We want to set the color attribute always except the case when the picker sliders are focused. So the following should do the job:

```suggestion
			if ( document.activeElement !== this.picker ) {
```

Please verify it on your end and update the `should not set any color directly to color-picker component when it visible` unit test (it should be much simpler with simply setting the active element instead of creating some nested fixture).

Dumluregn (commented):
I didn't find any regressions with the current solution :+1: Also, thanks for writing the missing tests. 

I see however a possibility to simplify it a bit - please take a look at the inline comment.

Dumluregn (approved):
I slightly adjusted tests and now waiting for CI to merge.

PR #14222, opened 2023-05-23
Allowed the engine View#scrollToTheSelection() to be adjustable

oleq:

Type: TODO. Closes #000.

### DIFF
--- a/packages/ckeditor5-engine/src/view/view.ts
+++ b/packages/ckeditor5-engine/src/view/view.ts
@@ -40,12 +40,14 @@ import {
 	CKEditorError,
 	ObservableMixin,
 	scrollViewportToShowTarget,
-	type ObservableChangeEvent
+	type ObservableChangeEvent,
+	type DecoratedMethodEvent
 } from '@ckeditor/ckeditor5-utils';
 import { injectUiElementHandling } from './uielement';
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-engine/src/view/view.ts
```diff
@@ -776,3 +797,30 @@ export type ViewRenderEvent = {
...
+ *   the scrolling (e.g. include additional `viewportOffset`).
+ * * The second argument corresponds to the original arguments passed to {@link module:utils/dom/scroll~scrollViewportToShowTarget}.
+ *   It allows listeners to re-execute the `scrollViewportToShowTarget()` method with its original arguments if there is such a need,
+ *   for instance, if the integration requires re–scrolling after certain interaction.
+ *
+ * @eventName ~View#scrollToTheSelection
+ */
+export type ViewScrollToTheSelectionEvent = {
+	name: 'scrollToTheSelection';
+	args: [
```
arkflpc, line 816: It looks very ugly in API docs. I'd extract both arguments and define them as separate interfaces. One example: `DowncastReduceChangesEvent` and `DowncastReduceChangesEventData`.

Similarly, `scrollToTheSelection` parameter should be defined separately.

arkflpc (approved):

PR #14231, opened 2023-05-24
Docs: updating configuration.

godai78:

Docs: updating CKBox configuration in restricted editing modes. Closes cksource/ckeditor5-internal#3311

---

### Additional information


gorzelinski (commented):
Overall, the demo works, but I noticed another potential issue. Changing mode from normal to restricted with the balloon displayed causes a warning in the console: `[attachTourBalloon] The target DOM node for the feature tour balloon does not exist.`

godai78:
> `[attachTourBalloon] The target DOM node for the feature tour balloon does not exist.`

Noticed that, too. Only happens when the user changes modes before focusing the editor. I suggest we do a follow-up for this.

gorzelinski:
Works for me 👍

PR #14257, opened 2023-05-25
Update fi.po

habibmy:

Type: Message. Closes #000.

---

### Additional information


pomek:
Hi @habibmy, thanks for the PR. Unfortunately, all the `*.po` files in the CKEditor 5 project are generated automatically. Any change in the repository will be lost while synchronizing the project with Transifex (https://explore.transifex.com/ckeditor/ckeditor5/).

I allowed myself to fix the issue on the Transifex side so that the change would be applied automatically before the next release.

image

However, thank you for preparing the PR. Thanks to that, we could notice that we had a mistake.

PR #14275, opened 2023-05-29
Integrate color picker to table properties

### LINKED ISSUE #14412: Make `ColorTableView` a shareable UI component and reuse it in table package

Let's focus on making the first step, there will be follow-up issues.

---

illia-stv:

Feature (table): Integrate color picker to table properties. Closes #14412.

Feature (ui): Hide input in color picker. Closes [#14288](https://github.com/ckeditor/ckeditor5/issues/14288).

---

### Additional information


packages/ckeditor5-table/src/tableproperties/ui/tablepropertiesview.ts
```diff
@@ -68,6 +69,11 @@ export interface TablePropertiesViewOptions {
 	 * The default table properties.
 	 */
 	defaultTableProperties: TablePropertiesOptions;
+
+	/**
+	 * The default color picker format.
+	 */
```
oleq, line 75: This is a whole config, not just the format.

packages/ckeditor5-table/src/ui/colorinputview.ts
```diff
@@ -245,16 +245,31 @@ export default class ColorInputView extends View {
...
+			if ( data.source === 'saveButton' ) {
+				dropdown.isOpen = false;
+			}
+		} );
+
+		colorGrid.on( 'cancel', () => {
+			this.value = this.options.defaultColorValue || '';
+			this.fire( 'input' );
+			dropdown.isOpen = false;
+		} );
```
oleq, line 271: Are you sure this logic is covered in tests?

packages/ckeditor5-table/src/ui/colorinputview.ts
```diff
@@ -297,42 +312,33 @@ export default class ColorInputView extends View {
...
 
-		removeColorButton.class = 'ck-input-color__remove-color';
-		removeColorButton.withText = true;
-		removeColorButton.icon = icons.eraser;
-		removeColorButton.label = removeColorButtonLabel;
-		removeColorButton.on( 'execute', () => {
-			this.value = defaultColor;
-			this.dropdownView.isOpen = false;
-			this.fire( 'input' );
+		const colorGrid = new ColorTableView( locale, {
```
oleq, line 322: `ColorInputView` tests don't mention `ColorTableView`. We need more tests.

packages/ckeditor5-table/src/utils/ui/table-properties.ts
```diff
@@ -378,13 +379,15 @@ export function getLabeledColorInputCreator(
...
+		colorPickerConfig: false | ColorPickerConfig;
 	}
 ) {
 	return ( labeledFieldView: LabeledFieldView, viewUid: string, statusUid: string ): ColorInputView => {
 		const colorInputView = new ColorInputView( labeledFieldView.locale!, {
 			colorDefinitions: colorConfigToColorGridDefinitions( options.colorConfig ),
 			columns: options.columns,
-			defaultColorValue: options.defaultColorValue
+			defaultColorValue: options.defaultColorValue,
+			colorPickerConfig: options.colorPickerConfig
```
oleq, line 390: We need tests for `getLabeledColorInputCreator()` to check whether this config is accepted and actually passed down to the `ColorInputView`.

packages/ckeditor5-table/src/tableconfig.ts
```diff
@@ -217,6 +217,7 @@ export interface TablePropertiesConfig {
 	borderColors?: TableColorConfig;
 	backgroundColors?: TableColorConfig;
 	defaultProperties?: TablePropertiesOptions;
+	colorPicker?: false | ColorPickerConfig;
```
oleq, line 220: I think we should mention this config somewhere in the doclets of `TableConfig` (`tableProperties?: TablePropertiesConfig;`)

packages/ckeditor5-table/src/tableconfig.ts
```diff
@@ -271,6 +272,7 @@ export interface TableCellPropertiesConfig {
 	borderColors?: TableColorConfig;
 	backgroundColors?: TableColorConfig;
 	defaultProperties?: TableCellPropertiesOptions;
+	colorPicker?: false | ColorPickerConfig;
```
oleq, line 275: I think we should mention this config somewhere in the doclets of `TableConfig` (`tableCellProperties?: TableCellPropertiesConfig;`)

packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -163,6 +166,10 @@ export default class ColorUI extends Plugin {
 				if ( data.source !== 'colorPicker' ) {
 					editor.editing.view.focus();
 				}
+
+				if ( data.source === 'saveButton' ) {
+					dropdownView.isOpen = false;
+				}
```
oleq, line 172: Do we have tests for this behavior?

packages/ckeditor5-font/tests/ui/colortableview.js
```diff
@@ -101,767 +89,6 @@ describe( 'ColorTableView', () => {
...
-
-			const activeStaticDocumentTile = staticColorsGrid.items.find( tile => tile.isOn );
-			const activeDocumentColorTile = documentColorsGrid.items.find( tile => tile.isOn );
-
-			expect( activeStaticDocumentTile ).to.be.undefined;
-			expect( activeDocumentColorTile ).to.be.undefined;
-		} );
-	} );
-
 	describe( 'disabled document colors section', () => {
```
oleq, line 92: What made you leave these tests behind in ckeditor5-font? Shouldn't they land in ckeditor5-ui?

packages/ckeditor5-font/tests/ui/colortableview.js
```diff
@@ -3,28 +3,16 @@
...
+import { ColorTableView } from '@ckeditor/ckeditor5-ui';
 import global from '@ckeditor/ckeditor5-utils/src/dom/global';
-
 import TestColorPlugin from '../_utils/testcolorplugin';
 import testUtils from '@ckeditor/ckeditor5-core/tests/_utils/utils';
 import { setData as setModelData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
 
-import removeButtonIcon from '@ckeditor/ckeditor5-core/theme/icons/eraser.svg';
-
 describe( 'ColorTableView', () => {
```
oleq, line 16: Why are `ColorTableView` tests left behind in ckeditor5-font?

oleq (changes requested):
There's an unnecessary margin left underneath the picker if there's no input configured:

image

We can address thus by switching that bottom margin to the top margin of the input (or labeled field, if you will) using `+` selector only when the input follows the picker.

oleq (changes requested):
Almost there 🙂

packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -58,35 +61,38 @@ export default class ColorPickerView extends View {
 	declare public _hexColor: string;
 
 	/**
-	* Debounced event method. The `colorPickerEvent()` method is called the specified `waitingTime` after
-	* `debouncedPickerEvent()` is called, unless a new action happens in the meantime.
-	*/
```
oleq, line 63: I tried to understand these docs but it appears they reference some non-existing methods or events. Could you please address this one?

oleq:
I also noticed this bug:




We should revert the value the form has been opened with upon clicking "cancel". [ATM we revert to configured default or `''`.](https://github.com/ckeditor/ckeditor5/blob/e2b1efc84fdfa77de4e77ed18194ad5f318722b6/packages/ckeditor5-table/src/ui/colorinputview.ts#L339)

oleq:
Also, can you check if 

```diff
diff --git a/packages/ckeditor5-table/src/ui/colorinputview.ts b/packages/ckeditor5-table/src/ui/colorinputview.ts
index 4f230df2fd..252f97da39 100644
--- a/packages/ckeditor5-table/src/ui/colorinputview.ts
+++ b/packages/ckeditor5-table/src/ui/colorinputview.ts
@@ -16,7 +16,9 @@ import {
          ColorTableView,
          type ColorDefinition,
          type DropdownView,
-         type ColorPickerConfig
+         type ColorPickerConfig,
+         type ColorTableExecuteEvent,
+         type ColorTableCancelEvent
 } from 'ckeditor5/src/ui';
 
 import { FocusTracker, KeystrokeHandler, type Locale } from 'ckeditor5/src/utils';
@@ -320,7 +322,11 @@ export default class ColorInputView extends View {
 
                   colorGrid!.appendUI();
 
-                  colorGrid.on( 'execute', ( _, data ) => {
+                  colorGrid.on<ColorTableExecuteEvent>( 'execute', ( evt, data ) => {
+                           if ( !this.dropdownView.isOpen ) {
+                                    return;
+                           }
+
                            if ( data.source === 'saveButton' ) {
                                     this.dropdownView.isOpen = false;
                                     return;
@@ -335,7 +341,7 @@ export default class ColorInputView extends View {
                            }
                   } );
 
-                  colorGrid.on( 'cancel', () => {
+                  colorGrid.on<ColorTableCancelEvent>( 'cancel', () => {
                            this.value = this.options.defaultColorValue || '';
 
                            this.fire( 'input' );
@@ -343,7 +349,7 @@ export default class ColorInputView extends View {
                            this.dropdownView.isOpen = false;
                   } );
 
-                  // colorGrid.bind( 'selectedColor' ).to( this, 'value' );
+                  colorGrid.bind( 'selectedColor' ).to( this, 'value' );
 
                   return colorGrid;
          }
```

resolves the problem of the feedback loop? I see that some random tests are still failing after this change but I suspect they are failing because the dropdown was not open before they were run (now it must be open for the new value to be applied).

illia-stv:
> Also, can you check if
> 
> ```diff
> diff --git a/packages/ckeditor5-table/src/ui/colorinputview.ts b/packages/ckeditor5-table/src/ui/colorinputview.ts
> index 4f230df2fd..252f97da39 100644
> --- a/packages/ckeditor5-table/src/ui/colorinputview.ts
> +++ b/packages/ckeditor5-table/src/ui/colorinputview.ts
> @@ -16,7 +16,9 @@ import {
>           ColorTableView,
>           type ColorDefinition,
>           type DropdownView,
> -         type ColorPickerConfig
> +         type ColorPickerConfig,
> +         type ColorTableExecuteEvent,
> +         type ColorTableCancelEvent
>  } from 'ckeditor5/src/ui';
>  
>  import { FocusTracker, KeystrokeHandler, type Locale } from 'ckeditor5/src/utils';
> @@ -320,7 +322,11 @@ export default class ColorInputView extends View {
>  
>                    colorGrid!.appendUI();
>  
> -                  colorGrid.on( 'execute', ( _, data ) => {
> +                  colorGrid.on<ColorTableExecuteEvent>( 'execute', ( evt, data ) => {
> +                           if ( !this.dropdownView.isOpen ) {
> +                                    return;
> +                           }
> +
>                             if ( data.source === 'saveButton' ) {
>                                      this.dropdownView.isOpen = false;
>                                      return;
> @@ -335,7 +341,7 @@ export default class ColorInputView extends View {
>                             }
>                    } );
>  
> -                  colorGrid.on( 'cancel', () => {
> +                  colorGrid.on<ColorTableCancelEvent>( 'cancel', () => {
>                             this.value = this.options.defaultColorValue || '';
>  
>                             this.fire( 'input' );
> @@ -343,7 +349,7 @@ export default class ColorInputView extends View {
>                             this.dropdownView.isOpen = false;
>                    } );
>  
> -                  // colorGrid.bind( 'selectedColor' ).to( this, 'value' );
> +                  colorGrid.bind( 'selectedColor' ).to( this, 'value' );
>  
>                    return colorGrid;
>           }
> ```
> 
> resolves the problem of the feedback loop? I see that some random tests are still failing after this change but I suspect they are failing because the dropdown was not open before they were run (now it must be open for the new value to be applied).

@oleq yeah you right, if will add `dropdown.isOpened = true` major of tests will pass but there is one test which fail, this one: https://github.com/ckeditor/ckeditor5/blob/4a01f6abd01581ad315033b85172782c156049d5/packages/ckeditor5-table/tests/ui/colorinputview.js#L190-L209 . On "cancel" button pressing we revert our changes, for example we back to "#ffaaff", when this change is applied we still have dropdown opened then color picker listener will handle this change and will convert it to hsl. Similar case with out color grid, as I know user could define any format of this static grid, for example **hex**, but it still will convert to hsl due this feedback loop, because dropdown is still opened.

Dumluregn (changes requested):
I'd really like to already merge this PR to the epic, but there are still some things that don't really make sense to create separate issues for (they are not bugs, just code improvements), and otherwise we'd only put them away for later (when it may be harder to fix them).

Additionally I'll create a few extra tickets for the further work. For now, just short descriptions for the reference (don't bother with them yet):

*   There's an empty undo step if you set the color in table color picker and click cancel buttons in both color picker and table properties.
*   Selected color is not propagated to the color table view (and color picker) - I guess because `// colorGrid.bind( 'selectedColor' ).to( this, 'value' );` caused some troubles.
*   More thorough docs review + docs for `hideInput` config property.
*   Possibly the guides updates.
*   Should `hideInput` really just hide the input? Shouldn't it stop its creation at all? (other name needed then).

packages/ckeditor5-ui/src/colorpicker/utils.ts
```diff
@@ -33,6 +33,15 @@ export type ColorPickerConfig = {
...
 
+/**
+ * Configuration of the color picker view.
+ *
+ * It can be forced to apply colors in the editor in a particular format and hide input.
+ */
+export type ColorPickerViewConfig = ColorPickerConfig & {
+	hideInput?: boolean;
+};
+
```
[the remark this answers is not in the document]
illia-stv: https://github.com/ckeditor/ckeditor5/pull/14275#discussion_r1218094257

packages/ckeditor5-table/src/ui/colorinputview.ts
```diff
@@ -297,43 +300,66 @@ export default class ColorInputView extends View {
...
-			columns: this.options.columns
+		colorGrid.on( 'execute', ( _, data ) => {
+			if ( data.source === 'saveButton' ) {
+				this.dropdownView.isOpen = false;
+				return;
+			}
+
+			this.value = data.value || defaultColor;
+
+			this.fire( 'input' );
```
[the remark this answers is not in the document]
illia-stv: It just applies the color to property which was defined in **this.value**. It trigers this part of code:
https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-table/src/tablecellproperties/ui/tablecellpropertiesview.ts#L511-L513

packages/ckeditor5-ui/src/colorpicker/utils.ts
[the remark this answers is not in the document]
Dumluregn: Ok, so this is valid if we don't want to expose `hideInput` as a configurable property for the user. But maybe we could just do it? We'll discuss it later.

Dumluregn (approved):
I leave R+ for this PR but let's all be aware there are follow-ups we need to handle until it's the production ready code.

As you may notice, I changed the "Closes" section:

*   The original integration issue was changed to be an epic. This PR only handles part of the job. Therefore, I created #14412 which is meant to indicate the actual status. Epic stays open.
*   I decided this PR doesn't solve #14361. At least not in the way that was requested in [https://github.com/ckeditor/ckeditor5/issues/14361#issuecomment-1582211440.](https://github.com/ckeditor/ckeditor5/issues/14361#issuecomment-1582211440.) Here we've removed the `submit` button type, while it carries the semantic value. To solve #14361, we should restore `submit` type together with its proper handling. BTW, there are two more buttons in the table plugin ("save" buttons in table properties and table cell properties) that use this type, so [let's ensure they work as expected too](https://github.com/ckeditor/ckeditor5/issues/14361#issuecomment-1597299873).

As promised before, there will be follow-ups that I will add to the epic.

Issue #14339, opened 2023-06-05
Bootstrap the show blocks feature plugin


Dumluregn:
## Provide a description of the task

We need the directory, tests etc. prepared for the rest of development.

oleq:
Closed in https://github.com/ckeditor/ckeditor5/pull/14345.

PR #14373, opened 2023-06-12
Implement show blocks feature

### LINKED ISSUE #14348: Implement the block toggling

Pressing the toolbar button should execute the command that will toggle the visibility of the blocks.

Dumluregn:

Internal (show-blocks): Implement toggling the blocks visibility. Closes #14348.

---

### Additional information

Merge #14339 first and rebase the PR to have a clean diff.


packages/ckeditor5-show-blocks/tests/manual/showblocks.js
```diff
@@ -236,6 +240,34 @@ ClassicEditor
 	} )
 	.then( editor => {
 		window.editor = editor;
+
+		editor.execute( 'showBlocks' );
```
oleq, line 244: You probably want to get rid of this and move it to the test description.

packages/ckeditor5-show-blocks/tests/showblockscommand.js
```diff
@@ -40,7 +42,94 @@ describe( 'ShowBlocksCommand', () => {
...
+			expect( command.value ).to.be.false;
+		} );
 	} );
 
-	describe( '#execute()', () => {} );
+	describe( '#execute()', () => {
+		it( 'should set "ck-show-blocks" class on the root when executed for the first time', () => {
+			editor.execute( 'showBlocks' );
+
+			expect( editor.editing.view.document.roots.get( 'main' )._classes.has( 'ck-show-blocks' ) ).to.be.true;
```
oleq, line 59: Please don't access `private` properties in tests unless absolutely necessary. If we had 500 tests like this, a single change in the internal implementation of a view `Element` would require a maaaaaaassive test rewrite. There's a https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_element-Element.html#function-hasClass for that.

oleq (changes requested):
FYI: I changed the way label backgrounds are rendered to give labels more air and to improve readability. I also switched highlighting from borders to outlines to address problems with elements that already have borders (e.g. blockquote). Please make sure I didn't break anything 🙃
image

packages/ckeditor5-show-blocks/theme/showblocks.css
```diff
@@ -2,3 +2,107 @@
...
+
+	& aside {
+		@mixin block-name-background ASIDE;
+	}
+
+	& blockquote {
+		@mixin block-name-background BLOCKQUOTE;
+	}
+
+	& > div {
```
oleq, line 41: I'm not so sure about this. If this was meant to exclude widget internals, this is too wide. For instance, this will hide divs inside some GHS structures (`article > div`) the user would certainly want to notice.

Dumluregn:
> I changed the way label backgrounds are rendered

So it was possible without filters after all 😅 LGTM. 

There's one thing (unrelated to your changes) I wonder about - the lack of the side padding if the nested elements don't add it:
image

In such case the blocks are a bit hard to read. But I suppose it's a minor inconvenience.

pszczesniak (changes requested):
Feature works and looks very nice, but got some things need a polishing:

*   on widgets we shouldn't display the label:



*   Toolbar button should not be "on" in `source editing` mode:



*   SVG background sometimes has some weird "marks", probably it's caused by stroke width (but only on large zoom its visible)



I think that 2nd and 3rd (if needed) can be extracted to a followups - lets fix only the 1st point.

Dumluregn:
I fixed

> *   on widgets we shouldn't display the label:

as requested :wave:

pszczesniak (approved):
LGTM :+1: and new icon rocks :grinning:

Issue #3355, opened 2019-10-09
Create styles for tooltips


oleq:
Related to https://github.com/ckeditor/ckeditor5-ui-default/issues/62.

PR #14379, opened 2023-06-13
Discovery: Export to PDF: text goes beyond its cell when columns are resized by resize handler

illia-stv:

Fix (table): Text will no longer go beyond its cell when columns are resized by resize handler in Firefox. Closes [#14386](https://github.com/ckeditor/ckeditor5/issues/14386).
Fix (export-pdf): Text will no longer go beyond its cell when columns are resized by resize handler in exported PDFs. Closes [#3355](https://github.com/cksource/ckeditor5-internal/issues/3355).

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/theme/table.css
+++ b/packages/ckeditor5-table/theme/table.css
@@ -52,10 +52,6 @@ when content is available outside the editor. See https://github.com/ckeditor/ck
 	text-align: left;
 }
 
-.ck.ck-content .table td {
-	overflow-wrap: break-word;
-}
-
 .ck-editor__editable .ck-table-bogus-paragraph {
 	/*
 	 * Use display:inline-block to force Chrome/Safari to limit text mutations to this element.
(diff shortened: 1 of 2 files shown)


illia-stv:
Let's have a look how Chrome display table with `contenteditable` and without:

Screenshot 2023-06-13 at 16 34 22

And compare it with how Firefox handle it:

Screenshot 2023-06-13 at 16 32 18

So as you see, chrome UA adds property` overflow-wrap: break-word ` in case when its `contenteditable`.

Then it also explains why in firefox this is broken too:

image
 
So we just need to define this property directly in `table` package

Dumluregn:
Let's merge it once QA team approves it and it gets R+ from our side.

mlewand:
Thanks for the writeup @illia-stv - it surfaced one more thing that we didn't noted: **currently** (master branch) table content is rendered differently in Chrome & Firefox - that's also a bug.

This PR addresses this bug, so @illia-stv please search our issues list whether such bug has been already listed or not. If it wasn't reported, please report it.

Finally add this bug to list of closed issues by this PR.

Dumluregn (approved):
LGTM :+1: Let's wait for QA team confirmation.

Dumluregn:
There's a green light from QAs :green\_circle:

PR #14416, opened 2023-06-19
Feedback loop in color picker in tables

### LINKED ISSUE #14389: Feedback loop in color picker in table properties

Visually everything is fine.

Problem actually is in color format which our table property apply to model. When we type **HEX** value into input we expect **HEX** value in model, but color picker listens on input to reflect actual color and it converts into **HSL**. Table property also listen on color picker changes and apply value in model.

### ✔️ Expected result

image

image

### ❌ Actual result

image

image

### ❓ Possible solution

## 📃 Other details

* Browser: …
* OS: …

---

illia-stv:

Fix (table): From now there is no feedback loop from color picker after changes in input. Closes #14389.

---

### Additional information


packages/ckeditor5-table/tests/ui/colorinputview.js
```diff
@@ -205,7 +205,11 @@ describe( 'ColorInputView', () => {
 
 				colorTableView.fire( 'cancel' );
 
-				expect( view.value ).to.be.equal( '#ffaaff' );
+				/**
+				 * TODO, should be HEX format: (https://github.com/ckeditor/ckeditor5/issues/14389)
```
Dumluregn, line 209: You should reference the issue which will handle this TODO, not the one introducing it. Whoever sees this will not get any explanation why this was done.

Dumluregn (changes requested):

packages/ckeditor5-table/tests/ui/colorinputview.js
Dumluregn: This applies to all the other TODOs as well.

Dumluregn (approved):

PR #14446, opened 2023-06-23
Set up block toolbar resize mechanism on given editable after the toolbar button is attached to it, instead of when editor is initialized

### LINKED ISSUE #14445: Block toolbar crashes when there are no roots in the editor

When block toolbar is initialized with multi-root that doesn't have any roots, it crashes, as it expects an editable to exist on the editor init.

This isn't a common case for multi-root per se, however it crashes always for revision history, as the way revision history is initialized with multi-root is that it always starts with no root which are added dynamically.

scofalik:

Internal (ui): Set up block toolbar resize mechanism on given editable after the toolbar button is attached to it, instead of when editor is initialized. Closes #14445.

### DIFF
--- a/packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
+++ b/packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
@@ -381,6 +381,7 @@ export default class BlockToolbar extends Plugin {
 		this.panelView.show();
 
 		const editableElement = this._getSelectedEditableElement();
+
 		this.toolbarView.maxWidth = this._getToolbarMaxWidth( editableElement );
 
 		this.panelView.pin( {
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -390,18 +379,29 @@ export default class BlockToolbar extends Plugin {
 		//
 		// https://github.com/ckeditor/ckeditor5/issues/6449, https://github.com/ckeditor/ckeditor5/issues/6575
 		this.panelView.show();
-		this.toolbarView.maxWidth = this._getToolbarMaxWidth();
+
+		const editableElement = this._getSelectedEditableElement();
+		this.toolbarView.maxWidth = this._getToolbarMaxWidth( editableElement );
```
niegowski, line 384:
```suggestion
		const editableElement = this._getSelectedEditableElement();

		this.toolbarView.maxWidth = this._getToolbarMaxWidth( editableElement );
```

Issue #14451, opened 2023-06-23
[Internal] Improve writing tests using TypeScript


filipsobol:
There are two issues with writing tests using TypeScript:

* `chai` globals like `expect()` are not declared in TypeScript, so it raises an error,
* TypeScript tests use the same eslint rules as the source code, raising issues when importing from `src/..` paths or relatively from other packages.

psmyrek:
Done in:
- https://github.com/ckeditor/ckeditor5-linters-config/pull/23
- https://github.com/ckeditor/ckeditor5-dev/pull/895
- https://github.com/ckeditor/ckeditor5/pull/14452

PR #14452, opened 2023-06-23
Exclude some eslint rules from both JS and TS tests

filipsobol:

Internal: Excluded some ESLint rules from both JavaScript and TypeScript tests. See #14451.


arkflpc (approved):

psmyrek (approved):
LGTM 🎉

PR #14455, opened 2023-06-23
Fix feedback loop in the color picker

### LINKED ISSUE #14415: Selected color is not propagated to the color table and color picker

Only present at [ck/epic/2978-integrate-color-picker-with-tables](https://github.com/ckeditor/ckeditor5/tree/ck/epic/2978-integrate-color-picker-with-tables).

1.  Put selection in a table cell with cell background color set.
2.  Open color picker in the table cell properties.

### ✔️ Expected result

Currently selected color is marked.

### ❌ Actual result

No color is marked, the "Remove color" button is focues.

## 📃 Other details

*   Browser: …
*   OS: …

---

### LINKED ISSUE #14432: Feedback loop while color dropdown is open in table properties

You could reproduce this behaviour on this `epic`: https://github.com/ckeditor/ckeditor5/tree/ck/epic/2978-integrate-color-picker-with-tables

1. Go to http://localhost:8125/ckeditor5-table/tests/manual/tableproperties.html
2. Choose any cell in table
3. Click on `cell properties button` ( image )
4. Type in background color input any **HEX** value e.g. `#ffaaff`
5. Next to input will be button with color panel  (image ), click on it
6. Click on `color picker button` ( image )
7. Start choosing any color
8. Click on `cancel button` ( image )

### ✔️ Expected result

In model on this cell should be applied **HEX** value as it was

image

### ❌ Actual result

image

----------------------------

The same problem with `remove color button`, so when we set default color, e.g. ( `rgb(255,0,0)` ) and after color changes and removing color it will apply in model color in **HSL** format or that format which was defined in color picker as output.

This problem is directly related with **feedback loop** which was partially resolved by this PR: https://github.com/ckeditor/ckeditor5/pull/14416
But it was temporarily fix, we need something more generic and bulletproof solution which cover all cases.  

Dumluregn:

Internal (ui): Fix feedback loop in the color picker. Closes #14432.  
Internal (ui): Propagate the selected color to the color table in table properties. Closes #14415.

---

### Additional information

I requested the QA team assistance, so let's wait with merge until they accept the PR.

### DIFF
--- a/packages/ckeditor5-font/tests/ui/colorui.js
+++ b/packages/ckeditor5-font/tests/ui/colorui.js
@@ -181,7 +181,8 @@ describe( 'ColorUI', () => {
 		describe( 'color picker', () => {
 			it( 'should execute command if the color gets changed when dropdown is open', () => {
 				const spy = sinon.spy( editor, 'execute' );
-				dropdown.colorTableView.colorPickerPageView.colorPickerView.color = '#a37474';
+
+				dropdown.colorTableView.colorPickerPageView.colorPickerView._debounceColorPickerEvent( '#a37474' );
 
 				sinon.assert.calledWithExactly( spy, 'testColorCommand', sinon.match( { value: 'hsl( 0, 20%, 55% )' } ) );
 			} );
@@ -246,7 +247,7 @@ describe( 'ColorUI', () => {
 			it( 'should avoid call the command multiple times', () => {
 				const spy = sinon.spy( editor, 'execute' );
 				// Color format normalization could result with command being called multiple times.
-				dropdown.colorTableView.colorPickerPageView.colorPickerView.color = '#a37474';
+				dropdown.colorTableView.colorPickerPageView.colorPickerView._debounceColorPickerEvent( '#a37474' );
 
 				expect( spy.callCount ).to.equal( 1 );
 			} );
--- a/packages/ckeditor5-font/tests/ui/colorui.js
+++ b/packages/ckeditor5-font/tests/ui/colorui.js
@@ -182,9 +182,9 @@ describe( 'ColorUI', () => {
 			it( 'should execute command if the color gets changed when dropdown is open', () => {
 				const spy = sinon.spy( editor, 'execute' );
 
-				dropdown.colorTableView.colorPickerPageView.colorPickerView._debounceColorPickerEvent( '#a37474' );
+				dropdown.colorTableView.colorPickerPageView.colorPickerView.fire( 'colorSelected', { color: '#a37474' } );
 
-				sinon.assert.calledWithExactly( spy, 'testColorCommand', sinon.match( { value: 'hsl( 0, 20%, 55% )' } ) );
+				sinon.assert.calledWithExactly( spy, 'testColorCommand', sinon.match( { value: '#a37474' } ) );
 			} );
 
 			it( 'should not execute command if the color gets changed when dropdown is closed', () => {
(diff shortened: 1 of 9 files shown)


packages/ckeditor5-table/tests/ui/colorinputview.js
```diff
@@ -195,21 +195,16 @@ describe( 'ColorInputView', () => {
...
+				colorTableView.colorPickerPageView.colorPickerView.fire( 'colorSelected', { color: 'hsl( 250, 100%, 83% )' } );
 
 				colorTableView.fire( 'cancel' );
 
-				/**
-				 * TODO, should be HEX format: (https://github.com/ckeditor/ckeditor5/issues/14432)
-				 */
-				expect( view.value ).to.be.equal( 'hsl( 300, 100%, 83% )' );
-				// expect( view.value ).to.be.equal( '#ffaaff' );
+				expect( view.value ).to.be.equal( '#ffaaff' );
```
illia-stv, line 207: Maybe good idea could be creating similar test with `Remove Color`
Screenshot 2023-06-26 at 13 21 45

packages/ckeditor5-table/tests/ui/colorinputview.js
```diff
@@ -302,12 +293,11 @@ describe( 'ColorInputView', () => {
 				sinon.assert.calledWithExactly( spy.lastCall, 'input' );
 			} );
 
-			it.skip( 'should have #selectedColor bound to the #value', () => {
+			it( 'should set #selectedColor to the #value upon dropdown opening', () => {
```
illia-stv, line 296: Maybe we could also create test which tests the logic of `#selectedColor` when dropdown is closed, then it should not be bound.

illia-stv (changes requested):
I like this solution it works well, I suggested adding only a few additional tests.

packages/ckeditor5-ui/tests/colortable/colortableview.js
```diff
@@ -343,6 +343,13 @@ describe( 'ColorTableView', () => {
 			expect( colorTableView.colorPickerPageView.items.length ).to.equal( 2 );
 		} );
 
+		it( 'should set selected color after changes in color picker', () => {
+			colorTableView._appendColorPicker();
+			colorTableView.colorPickerPageView.colorPickerView._debounceColorPickerEvent( '#113322' );
```
oleq, line 348: Please don't use private properties in tests unless there's absolutely no other choice. This leads to multitude of tests failing in random places if the internal implementation of a class/component changes in the future. 

You can fire an event that belongs to the public API of the `ColorPickerView`

```
colorPickerView.fire( 'colorSelected', { color: '#113322' } );
```

to test this since all we want to know here is that the data propagation between components exists and works fine

```
this.colorPickerView!.on<ColorPickerColorSelectedEvent>( 'colorSelected', ( evt, data ) => {
	// ... 

	this.set( 'selectedColor', data.color );
} );
```

packages/ckeditor5-font/tests/ui/colorui.js
```diff
@@ -181,7 +181,8 @@ describe( 'ColorUI', () => {
 		describe( 'color picker', () => {
 			it( 'should execute command if the color gets changed when dropdown is open', () => {
 				const spy = sinon.spy( editor, 'execute' );
-				dropdown.colorTableView.colorPickerPageView.colorPickerView.color = '#a37474';
+
+				dropdown.colorTableView.colorPickerPageView.colorPickerView._debounceColorPickerEvent( '#a37474' );
```
oleq, line 185: Same here: let's fire a proper event here. All we want to know is that there's a proper response to the public event being fired. 

At this stage, we don't care about the internal state of the `ColorPickerView`. I know that `_debounceColorPickerEvent()` sets both `#color` and fires the event but the latter does not matter here.

oleq (changes requested):
It looks like we finally got it right. Commented on the usage of private properties in tests.

PR #14486, opened 2023-06-28
Additional configuration for context elements in dropdown

DawidKossowski:

Feature (ui): Dropdowns will stay open after clicking on an HTML element added to the dropdown's focus tracker.

### DIFF
--- a/docs/_snippets/examples/bottom-toolbar-editor.js
+++ b/docs/_snippets/examples/bottom-toolbar-editor.js
@@ -12,7 +12,7 @@ import { List } from '@ckeditor/ckeditor5-list';
 import { Alignment } from '@ckeditor/ckeditor5-alignment';
 import { Autoformat } from '@ckeditor/ckeditor5-autoformat';
 import { BlockQuote } from '@ckeditor/ckeditor5-block-quote';
-import { DropdownView, DropdownButtonView, DropdownPanelView, ToolbarView, clickOutsideHandler } from '@ckeditor/ckeditor5-ui';
+import { DropdownView, ToolbarView, createDropdown } from '@ckeditor/ckeditor5-ui';
 import { EasyImage } from '@ckeditor/ckeditor5-easy-image';
 import { Essentials } from '@ckeditor/ckeditor5-essentials';
 import { Heading } from '@ckeditor/ckeditor5-heading';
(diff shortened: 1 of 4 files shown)


scofalik:
I believe that it makes sense to link dropdown focus tracker elements with click outside handler. That is, if an element is in dropdown's focus tracker, then clicking on it, should not trigger click outside handler.

It is very possible that I am missing something but I am having hard time imagining when these two sets of elements should not be the same.

1.  If I have element X in `contextElements` but not in focus tracker, when I click it, the dropdown will be hidden anyway because it lost focus.
2.  If I have element X in focus tracker but not in `contextElements`, the dropdown will be hidden even though it has focus. It may be even difficult to debug.

So, you must remember to add these "extra" `contextElements` to the focus tracker anyway, because just adding them in the parameter doesn't really solve the problem.

The only thing that comes to my mind is that maybe there's an element, that when clicked, should hide the dropdown, but when focused otherwise (keyboard?), should not hide it. But such element would be quite confusing :thinking:.

@oleq I know you discussed this previously with @DawidKossowski, but can we get to this topic back?

oleq:
I suppose what @scofalik said makes sense. I don't see any use cases where `contextElements` and `FocusTracker` work within different elements.

There are two use-cases we must satisfy:
1. A regular dropdown that closes when sb. clicked outside of it (`closeDropdownOnClickOutside()`) or lost focus (`closeDropdownOnBlur()`)
2. A dropdown that stays open as long as the user focus is in the editor (see the formatting dropdown in https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/bottom-toolbar-editor.html).

I think both cases should be fine if we coupled this `contextElements` and `FocusTracker` logic. 

**A sidenote worth checking**: The latter (2.) is the only case in our codebase where we create a dropdown by calling the constructor directly. This is because there was no way to use `createDropdown()` and disable closing the dropdown by clicking outside of it.

I think coupling `contextElements` and `FocusTracker` should streamline this example and allow us to use `createDropdown()` as well. All we'll need to do is add
```js
...[ ...editor.ui.getEditableElementsNames() ].map( name => editor.ui.getEditableElement( name ) ),
document.querySelector( '.ck-body-wrapper' )
```
to the dropdown's focus tracker and that should basically incapacitate the `closeDropdownOnClickOutside()` and `closeDropdownOnBlur()`.

scofalik:
LGTM so far. Can you check the thing which Olek mentioned?

DawidKossowski:
Done

PR #14494, opened 2023-06-29
Fix random selection-releated crashes happening on Firefox for unknown reasons (error code `NS_ERROR_FAILURE`)

### LINKED ISSUE #14493: [FF] NS_ERROR_FAILURE thrown randomly when setting DOM selection

We had problems with `NS_ERROR_FAILURE` thrown on random occasions, some of them seems fixed (closed issues in this repository), and at least one error is not closed (https://github.com/cksource/collaboration-features/issues/4183).

When testing multi-root editor in RTC environment, this happened again, this time, when removing the last root and performing undo. It seems to not be 100%, but it mostly throws when you do it right. It is an edge case but that's not entirely the point.

The point is that the error is not described and it is hard to find any information about when it would be thrown.

The error is thrown in `renderer.ts` in `_updateDomSelection`:

```js
domSelection.collapse( anchor.parent, anchor.offset );
domSelection.extend( focus.parent, focus.offset );
```

I checked and `anchor` and `focus` are set correctly. The DOM elements have same owner document and it is the same as the selection. Everything seems right.

One suggestion I did find when looking for the solution is to use `setBaseAndExtent` instead. After using it, I wasn't able to reproduce this problem anymore.

Maybe we will be able to fix random FF bugs this way?

scofalik:

Fix (engine): Fix random selection-releated crashes happening on Firefox for unknown reasons (error code \`NS\_ERROR\_FAILURE\`). Closes #14493.

### DIFF
--- a/packages/ckeditor5-engine/src/view/renderer.ts
+++ b/packages/ckeditor5-engine/src/view/renderer.ts
@@ -942,8 +942,7 @@ export default class Renderer extends ObservableMixin() {
 		// @if CK_DEBUG_TYPING // 	);
 		// @if CK_DEBUG_TYPING // }
 
-		domSelection.collapse( anchor.parent, anchor.offset );
-		domSelection.extend( focus.parent, focus.offset );
+		domSelection.setBaseAndExtent( anchor.parent, anchor.offset, focus.parent, focus.offset );
 
 		// Firefox–specific hack (https://github.com/ckeditor/ckeditor5-engine/issues/1439).
 		if ( env.isGecko ) {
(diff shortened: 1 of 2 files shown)


scofalik:
Alternative solution would be to put `extend()` and maybe `collapse()` in `try catch`.

niegowski (changes requested):
Looks ok but some tests are failing.

DawidKossowski:
Unit tests have been fixed.

niegowski (approved):
I updated some tests to avoid false positives.

niegowski:
@DawidKossowski could you verify my changes in this PR? https://github.com/ckeditor/ckeditor5/pull/14494/commits/f24669f812e6a76f883dcf13ad0f3f99844dea4d

DawidKossowski (approved):
LGTM

Issue #13866, opened 2023-04-12
insertParagraph command creates additional empty paragraph when used at the end of block element


FilipTokarski:
## 📝 Provide detailed reproduction steps (if any)

1. Open editor
2. Put the caret at the end of a paragraph
3. Use the following code to insert a new empty paragraph:
```js
editor.execute( 'insertParagraph', {
  position: editor.model.document.selection.getLastPosition()
} );
```

### ✔️ Expected result

One empty paragraph is inserted.

### ❌ Actual result

There are two empty paragraphs inserted:

https://user-images.githubusercontent.com/34380544/231429732-9d5bef2f-7c5c-4b2e-8a01-107edf92f4b3.mov

## 📃 Other details

* Browser: …
* OS: …
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

PR #14510, opened 2023-07-02
insertParagraph command creates additional empty paragraph when used at the end of block element

illia-stv:

Fix (paragraph): "insertParagraph" command doesn't insert two paragraphs when position is at the edge of the line. Closes [#13866](https://github.com/ckeditor/ckeditor5/issues/13866).

---

### Additional information

### DIFF
--- a/packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
+++ b/packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
@@ -8,7 +8,7 @@
  */
 
 import { Command, type Editor } from '@ckeditor/ckeditor5-core';
-import type { Element, Position, Text } from '@ckeditor/ckeditor5-engine';
+import type { Element, Position, Text, Writer } from '@ckeditor/ckeditor5-engine';
 
 /**
  * The insert paragraph command. It inserts a new paragraph at a specific
@@ -72,17 +72,20 @@ export default class InsertParagraphCommand extends Command {
 				return;
 			}
 
-			if ( position.isAtStart && position.path.length > 1 && ( position.nodeAfter as Text )?.data ) {
-				// When position is at the start of the line we want to insert paragraph before
+			if ( position.isAtStart && position.isAtEnd && position.path.length > 1 ) {
+				// NOTE: In case when paragraph is empty.
+				// When position is at the start and at the end of the line we want to insert paragraph after.
+				// <paragraph>[]</paragraph> ---> <paragraph></paragraph><paragraph>[]</paragraph>
+				position = createPositionBellow( writer, position, allowedParent );
+				console.log( position );
+			} else if ( position.isAtStart && position.path.length > 1 && ( position.nodeAfter as Text ).data ) {
+				// When position is at the start of the line we want to insert paragraph above.
 				// <paragraph>[]foo</paragraph> ---> <paragraph>[]</paragraph><paragraph>foo</paragraph>
 				position = writer.createPositionFromPath( allowedParent, position.getParentPath() );
-			} else if ( position.isAtEnd && position.path.length > 1 && ( position.nodeBefore as Text )?.data ) {
-				// When position is at the end of the line we want to insert paragraph after
+			} else if ( position.isAtEnd && position.path.length > 1 && ( position.nodeBefore as Text ).data ) {
+				// When position is at the end of the line we want to insert paragraph after.
 				// <paragraph>foo[]</paragraph> ---> <paragraph>foo</paragraph><paragraph>[]</paragraph>
-				const parentPath = position.getParentPath();
-				const length = parentPath.length;
-				const path = [ ...parentPath.slice( 0, length - 1 ), parentPath[ length - 1 ] + 1 ];
-				position = writer.createPositionFromPath( allowedParent, path );
+				position = createPositionBellow( writer, position, allowedParent );
 			} else if ( !model.schema.checkChild( position.parent as Element, paragraph ) ) {
 				// When position is inside the content we want to split it
 				// <paragraph>fo[]o</paragraph>
@@ -99,4 +102,13 @@ export default class InsertParagraphCommand extends Command {
 			writer.setSelection( paragraph, 'in' );
 		} );
 	}
+}
+
+// Creates bellow position where should be inserted new paragraph.
+export function createPositionBellow( writer: Writer, position: Position, allowedParent: Element ): Position {
+	const parentPath = position.getParentPath();
+	const length = parentPath.length;
+	const path = [ ...parentPath.slice( 0, length - 1 ), parentPath[ length - 1 ] + 1 ];
+
+	return writer.createPositionFromPath( allowedParent, path );
 }
--- a/packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
+++ b/packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
@@ -8,7 +8,7 @@
  */
 
 import { Command, type Editor } from '@ckeditor/ckeditor5-core';
-import type { Element, Position, Text, Item } from '@ckeditor/ckeditor5-engine';
+import type { Element, Position, Item } from '@ckeditor/ckeditor5-engine';
 
 /**
  * The insert paragraph command. It inserts a new paragraph at a specific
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-paragraph/tests/insertparagraphcommand.js
```diff
@@ -70,6 +70,59 @@ describe( 'InsertParagraphCommand', () => {
...
+				position: model.document.selection.getLastPosition()
+			} );
+
+			expect( getData( model ) ).to.equal(
+				'<paragraph>[]</paragraph>' +
+				'<paragraph>foo</paragraph>'
+			);
+		} );
+
+		it( 'should insert paragraph when position is at the start of line asd', () => {
```
Dumluregn, line 99: `asd` :stuck\_out\_tongue\_closed\_eyes:

packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
```diff
@@ -82,3 +101,12 @@ export default class InsertParagraphCommand extends Command {
...
 }
+
+// Creates bellow position where should be inserted new paragraph.
+export function createPositionBellow( writer: Writer, position: Position, allowedParent: Element ): Position {
+	const parentPath = position.getParentPath();
+	const length = parentPath.length;
+	const path = [ ...parentPath.slice( 0, length - 1 ), parentPath[ length - 1 ] + 1 ];
+
+	return writer.createPositionFromPath( allowedParent, path );
+}
```
Dumluregn, line 112: There's already an API for it - [`Writer.createPositionAfter()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_model-Model.html#function-createPositionAfter). I think we should be safe with `position =` _`writer`_`.createPositionAfter( position.parent as Item );`.

BTW - [https://dictionary.cambridge.org/dictionary/english/bellow](https://dictionary.cambridge.org/dictionary/english/bellow) :smile:

packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
```diff
@@ -64,15 +64,34 @@ export default class InsertParagraphCommand extends Command {
...
 
+			if ( position.isAtStart && position.isAtEnd && position.path.length > 1 ) {
+				// NOTE: In case when paragraph is empty.
+				// <paragraph>[]</paragraph> ---> <paragraph></paragraph><paragraph>[]</paragraph>
+				position = createPositionBellow( writer, position, allowedParent );
+			} else if ( position.isAtStart && position.path.length > 1 && ( position.nodeAfter as Text ).data ) {
+				// When position is at the start of the line we want to insert paragraph above.
+				// <paragraph>[]foo</paragraph> ---> <paragraph>[]</paragraph><paragraph>foo</paragraph>
+				position = writer.createPositionFromPath( allowedParent, position.getParentPath() );
+			} else if ( position.isAtEnd && position.path.length > 1 && ( position.nodeBefore as Text ).data ) {
```
Dumluregn, line 83: All these conditions contain `position.path.length > 1`. Maybe it's possible to simplify them? What does it mean if it's equal to 0 or 1? Is an earlier condition possible?

packages/ckeditor5-paragraph/tests/insertparagraphcommand.js
```diff
@@ -70,6 +70,59 @@ describe( 'InsertParagraphCommand', () => {
...
+				let expectedPosition = model.createPositionFromPath( root, [ 1 ] );
+				expect( createPositionBellow( writer, position, root ), '[ 0, 2 ] --> [ 1 ]' )
+					.to.eql( expectedPosition );
+
+				position = model.createPositionFromPath( root, [ 1, 2, 3 ] );
+				expectedPosition = model.createPositionFromPath( root, [ 1, 3 ] );
+				expect( createPositionBellow( writer, position, root ), '[ 1, 2, 3 ] --> [ 1, 3 ]' )
+					.to.eql( expectedPosition );
+			} );
+		} );
```
Dumluregn, line 124: This test won't be necessary once you use the API I proposed.

Dumluregn (changes requested):
So, the problem is that currently with the following structure: `<h2>[]</h2>`, `writer.split()` does exactly what we tell it - splits it into `<h2>[]</h2><h2></h2>`, and then we insert a paragraph in the middle (same applies for the `<p>` elements, but I used `<h2>` to make it clearer), ending up with two new blocks instead of one: `<h2></h2><p>[]</p><h2></h2>`. So I agree that we problem lies within the command, not `writer.split()` method.

The proposed solution makes sense. Nice that you've noticed that the issue occurs also at the beginning of the blocks, not just at the end. Let's clean it up a little and we'll be ready to pass the PR to the Core team.

packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
```diff
@@ -82,3 +101,12 @@ export default class InsertParagraphCommand extends Command {
 		} );
 	}
 }
+
+// Creates bellow position where should be inserted new paragraph.
+export function createPositionBellow( writer: Writer, position: Position, allowedParent: Element ): Position {
```
mlewand, line 106:
```suggestion
export function createPositionBelow( writer: Writer, position: Position, allowedParent: Element ): Position {
```

You probably meant that 😉

packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
illia-stv: If **path** contain only 1 value like [ 1 ], `getParentPath()` of that will be empty array ([ ]) then it's not valid path.

illia-stv: Eventually we could replace it with
`
if ( position.path.length < 2 ) {
				model.insertContent( paragraph, position );
				writer.setSelection( paragraph, 'in' );

				return;
			}

			if ( position.isAtStart && position.isAtEnd ) {
				// NOTE: In case when paragraph is empty.
				// <paragraph>[]</paragraph> ---> <paragraph></paragraph><paragraph>[]</paragraph>
				position = writer.createPositionAfter( position.parent as Item );
			} else if ( position.isAtStart && ( position.nodeAfter as Text ).data ) {
				// When position is at the start of the line we want to insert paragraph above.
				// <paragraph>[]foo</paragraph> ---> <paragraph>[]</paragraph><paragraph>foo</paragraph>
				position = writer.createPositionFromPath( allowedParent, position.getParentPath() );
			} else if ( position.isAtEnd && ( position.nodeBefore as Text ).data ) {
				// When position is at the end of the line we want to insert paragraph after.
				// <paragraph>foo[]</paragraph> ---> <paragraph>foo</paragraph><paragraph>[]</paragraph>
				position = writer.createPositionAfter( position.parent as Item );
			} else if ( !model.schema.checkChild( position.parent as Element, paragraph ) ) {
				// When position is inside the content we want to split it
				// <paragraph>fo[]o</paragraph>
				//            |
				//			  ↓
				// <paragraph>fo</paragraph>
				// <paragraph>[]</paragraph>
				// <paragraph>o</paragraph>
				position = writer.split( position, allowedParent ).position;
			}

			model.insertContent( paragraph, position );

			writer.setSelection( paragraph, 'in' );
`

Dumluregn (approved):
LGTM, waiting for CI.

arkflpc (approved):

packages/ckeditor5-paragraph/src/insertparagraphcommand.ts
```diff
@@ -59,20 +59,44 @@ export default class InsertParagraphCommand extends Command {
...
+			// <paragraph>[]</paragraph> ---> <paragraph></paragraph><paragraph>[]</paragraph>
+			const isInEmptyBlock = position.isAtStart && position.isAtEnd;
+
+			// E.g.
+			// <paragraph>foo[]</paragraph> ---> <paragraph>foo</paragraph><paragraph>[]</paragraph>
+			const isAtEndOfTextBlock = position.isAtEnd && position.nodeBefore?.is( '$text' );
+
+			// E.g.
+			// <paragraph>[]foo</paragraph> ---> <paragraph>[]</paragraph><paragraph>foo</paragraph>
+			const isAtStartOfTextBlock = position.isAtStart && position.nodeAfter?.is( '$text' );
```
niegowski, line 91: What about inline elements? For example inline image? Maybe you could check `position.parent.isEmpty` instead of checking `nodeBefore`/`nodeAfter`?

illia-stv: We decided to keep previous behaviour for **non-text** elements, look at this test: https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-widget/tests/widgettypearound/widgettypearound.js#L1056-L1089
here this kind behaviour is expected for **widgets** then we fix issue only for **text** nodes

niegowski: I'm talking about inline widgets, not block widgets. They are allowed everywhere text is allowed.

Dumluregn (approved):

PR #14528, opened 2023-07-04
Hide balloon panel while scrolling out of the limiter element.

### LINKED ISSUE #5328: Balloon panel sticks out of the limiter element while scrolling

pszczesniak:

Fix (ui): Should hide the ballon panel when it's scrolled outside the visible area.

Closes #5328.

---

### Additional information

~**Waiting for QA testing**~ :hourglass\_flowing\_sand: ~cksource/qa-internal#176~  :green\_circle: :point\_right: https://github.com/ckeditor/ckeditor5/pull/14528#issuecomment-1638253635


charlttsie:
### With the outer container smaller than the editor, balloon is visible outside the container when scrolling
**Steps**
1\. Open http://localhost:8125/ckeditor5-ui/tests/manual/tickets/5328/1.html
2\. Use an editor’s instance where the outer container is smaller than the editor
3\. Insert a table/add some text for the horizontal scroll to appear
4\. Scroll up/down

**Result**
Balloons show up outside the container:

https://github.com/ckeditor/ckeditor5/assets/132071257/3a9cd882-a366-4d0a-9cf3-5fb11d851313

charlttsie:
### When scrolling the editor, table toolbar shows up outside of it
**Steps**
1\. Open http://localhost:8125/ckeditor5-ui/tests/manual/tickets/5328/1.html
2\. Insert a table and open its toolbar
3\. Scroll the editor up/down

**Result**
Table toolbar is visible when scrolling out of the editor:

https://github.com/ckeditor/ckeditor5/assets/132071257/80613191-2546-45a2-9453-261e932e5d0c

charlttsie:
### If the target element isn't entirely within the editor's view, balloon toolbar displays below the editor
**Steps**
1\. Open http://localhost:8125/ckeditor5-ui/tests/manual/tickets/5328/1.html
2\. Place the focus on a table or image and scroll up the editor, so the bottom part of the widget is partly hidden but the toolbar is still visible
3\. Scroll down the page

**Result**
The toolbar shows up in an unexpected position:

https://github.com/ckeditor/ckeditor5/assets/132071257/e937667e-3ff5-4541-a184-92f3bf7d5c35

mabryl:
### The entire element with the toolbar has to move outside of the editor container for the toolbar to disappear

#### Steps to reproduce
1. Open http://localhost:8125/ckeditor5-ui/tests/manual/tickets/5328/1.html
2. Place the focus on a table or image and start scrolling horizontally
3. Notice that the balloon disappears only when the entire element with the toolbar moves outside of the editor container

#### Recording:

https://github.com/ckeditor/ckeditor5/assets/72079603/44708b95-5776-4bac-88dc-b1c7bb1c863f

pszczesniak:
[Horizontal scrolling issue](https://github.com/ckeditor/ckeditor5/pull/14528#issuecomment-1630390766) is out of the scope of this.

Follow up can be found [here](https://github.com/ckeditor/ckeditor5/issues/14587).

mabryl:
QA testing completed and it looks good after the latest round of testing. Summary here: https://cksource.slack.com/archives/C056P7FQ8TY/p1689603459626639?thread_ts=1688474081.871299&cid=C056P7FQ8TY

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -1145,6 +1174,20 @@ export function generatePositions( options: {
 					...config
 				}
 			};
+		},
+
+		// ------- Hidden
+
+		viewportHidden: () => {
```
oleq, line 1181: The names of positions were written in such a way that they describe where the balloon will show up. For instance `eastArrowWest` means the balloon will show up east of its target with the arrow poining west. `viewportStickyNorth` means that the balloon will be sticky to the north of the viewport. `viewportHidden` then means...? I think plain `hidden` would make more sense here.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -245,7 +248,8 @@ export default class BalloonPanelView extends View {
 				defaultPositions.northArrowSouthMiddleEast,
 				defaultPositions.northArrowSouthWest,
 				defaultPositions.northArrowSouthEast,
-				defaultPositions.viewportStickyNorth
+				defaultPositions.viewportStickyNorth,
+				defaultPositions.viewportHidden
```
oleq, line 252: See my comment in image/ui/utils.ts. I would try injecting this position in L257 whether there were no positions specified by the user (defaults being used) or they used some specific ones (even custom ones).

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -1145,6 +1174,20 @@ export function generatePositions( options: {
...
 			};
+		},
+
+		// ------- Hidden
+
+		viewportHidden: () => {
+			return {
+				top: OFF_THE_SCREEN_POSITION,
+				left: OFF_THE_SCREEN_POSITION,
+				name: 'noarrow',
```
oleq, line 1185: `viewportStickyNorth` uses `arrowless`. Is there a specific reason for another name here?

packages/ckeditor5-utils/src/dom/position.ts
```diff
@@ -99,12 +100,44 @@ export function getOptimalPosition( { element, target, positions, limiter, fitIn
 
 	let bestPosition: Position;
 
+	// Get intersection of all scrollable ancestors of `target`.
+	const allScrollableAncestors = getScrollableAncestors( target as HTMLElement );
+	const ancestorsIntersection: Rect | null = allScrollableAncestors
+		.reverse()
```
oleq, line 106: Out of curiosity, why reverse and go from the root rect to the leaf rect instead the opposite (default) direction here?

packages/ckeditor5-utils/src/dom/position.ts
```diff
@@ -99,12 +100,44 @@ export function getOptimalPosition( { element, target, positions, limiter, fitIn
...
+		.map( item => new Rect( item as HTMLElement ) )
+		.reduce( ( ancestorIntersectionAccumulator, ancestorRect ): Rect => {
+			if ( !ancestorIntersectionAccumulator || ancestorIntersectionAccumulator.top === undefined ) {
```
oleq, line 109: When would `top` be `undefined`? Which use-case is it?

oleq: Is this to cover cases when `getIntersection()` returns `null` that you ignore by casting `as Rect` in the last line?

packages/ckeditor5-utils/src/dom/position.ts
```diff
@@ -99,12 +100,44 @@ export function getOptimalPosition( { element, target, positions, limiter, fitIn
...
+		} );
+
 	// @if CK_DEBUG_POSITION // RectDrawer.clear();
 	// @if CK_DEBUG_POSITION // RectDrawer.draw( targetRect, { outlineWidth: '5px' }, 'Target' );
 
 	const viewportRect = fitInViewport && getConstrainedViewportRect( viewportOffsetConfig ) || null;
 	const positionOptions = { targetRect, elementRect, positionedElementAncestor, viewportRect };
 
+	if ( ( !ancestorsIntersection ) || ( !targetRect.getVisible() ) ) {
+		// Default last position is `viewportHidden`.
```
oleq, line 127: You don't know that. `getOptimalPosition()` gets an arbitrary set of positions. It is not used by balloon system only but also by dropdowns to position their panels.

packages/ckeditor5-utils/src/dom/position.ts
```diff
@@ -99,12 +100,44 @@ export function getOptimalPosition( { element, target, positions, limiter, fitIn
...
+		return new PositionObject( positions.at( -1 )!, positionOptions );
```
oleq, line 128: `at()` is supported in Safari since Mar 2022 https://caniuse.com/?search=Array.at. Knowing yearly release cycle of MacOS I'm wondering how many "old" installations of MacOS are still in use and how often would this throw. Let's stay on the safe side for a little longer, WDYT?

packages/ckeditor5-utils/src/dom/position.ts
```diff
@@ -99,12 +100,44 @@ export function getOptimalPosition( { element, target, positions, limiter, fitIn
...
+	}
+	const ancestorsIntersectionWindowRect = new Rect( window ).getIntersection( ancestorsIntersection! );
```
oleq, line 130: Get scrollable ancestors will always consider `global.document`. 

https://github.com/ckeditor/ckeditor5/blob/fece6640d3f587646caa42bae8d4dfa1dec6a742/packages/ckeditor5-utils/src/dom/getscrollableancestors.ts#L27

If you got it covered while retrieving rects, you wouldn't need to handle this one separately. It would be just another rect to intersect. See the implementation of the sticky panel to learn more

https://github.com/ckeditor/ckeditor5/blob/fece6640d3f587646caa42bae8d4dfa1dec6a742/packages/ckeditor5-ui/src/panel/sticky/stickypanelview.ts#L417-L429

packages/ckeditor5-utils/src/dom/position.ts
```diff
@@ -119,11 +152,22 @@ export function getOptimalPosition( { element, target, positions, limiter, fitIn
...
+		const properRectLimiter = isNaN( ancestorsIntersectionWindowRect.top ) ? limiterRect : ancestorsIntersectionWindowRect;
+		const limiterViewportIntersection = ( properRectLimiter && !isNaN( properRectLimiter.top ) && viewportRect ) ?
+			properRectLimiter.getIntersection( viewportRect! ) : properRectLimiter;
+
+		Object.assign( positionOptions, {
+			limiterRect, viewportRect: target instanceof Range ? limiterRect : limiterViewportIntersection } );
+
+		// When `limiter` is not set and `getBestPosition` return null lets return the first `position`
+		// from the `positions` list. If the limiter is set - we will take the last `position` (hide).
+		const indexOfPositionToReturn = limiter ? -1 : 0;
```
oleq, line 164: Like I said before, an assumption that the last available position is the one that hides the positioned element is weak.

oleq (changes requested):
In your solution
* you added a hiding position to the `BalloonPanelView#attachTo()` defaults,
* you added a hiding position to the list in widget and image utils responsible for positioning toolbars
* you assumed in `getOptimalPosition()` that the last available position is the hiding one despite `getOptimalPosition()` working as a generic helper (also used by `DropdownView`) and having an open API with the list of positions (also custom ones) to be passed by developers.

This will fail as soon as somebody configures any balloon (or other UI using `getOptimalPosition()`) with their custom positions and they won't put the hiding one as the last. `getOptimalPosition()` will pick the last one in the corner cases and instead of getting hidden, the balloon (UI) will move in some unexpected fashion.

---

The idea that comes to my mind is to have `getOptimalPosition()` return `null` when the target becomes invisible (that ancestor rect intersection check that you did). Then:
* `BalloonPanelView` could say "oh, alright then, I'm gonna use my hidden position in this case because balloons cannot be attached to a void"
* `DropdownView` could say "oh, alright then, I'm gonna use my default position (let's say its south) because you just can't make a dropdown panel disappear like a balloon"
* Integrators UI could say "oh, I'm gonna handle it on my own but the API does not force me to do anything specific, it just gives me options (they hidden position is not hardcoded anywhere)"

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
pszczesniak: Nah, i've used the different name for debugging. It was easier to recognise which one was selected by logging `position.name`

packages/ckeditor5-utils/src/dom/position.ts
pszczesniak: Need to dig deeper in my memory (shame on me that i didn't comment it). TBH was thinking about it when i create this PR.

packages/ckeditor5-utils/src/dom/position.ts
pszczesniak: for example:

```
new Rect( global.document )
------
Rect {top: undefined, right: undefined, bottom: undefined, left: undefined, width: undefined, height: undefined}
```

pszczesniak:
Different approach was chosen - more generic one :point\_right: #14630

PR #14534, opened 2023-07-05
Inline picture should not exceed the width of the parent container

### LINKED ISSUE #14532: Image size attributes and picture (art direction)

Inline image (picture) doesn't look good with current ck-content styles if it is not resized. The output shows a much larger image compared to the editor content.

image

To re-create the issue:
1. Open http://localhost:8125/ckeditor5-image/tests/manual/imagesizeattributesallcases.html manual test.
2. For the first editor set data:
```
<p><picture><source srcset="logo-square.png" media="(max-width: 800px)" type="image/png"><source srcset="logo-wide.png" media="(min-width: 800px)" type="image/png"><img class="image_resized" style="width:250px;" src="logo-wide.png" alt="Alt text of the block image."></picture></p>
```
3. Shrink the browser so the images are small enough to see the difference.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

mmotyczynska:

Task (image): Adds picture use cases to image sizes manual test. Closes #14532.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-image/theme/image.css
+++ b/packages/ckeditor5-image/theme/image.css
@@ -4,6 +4,13 @@
  */
 
 .ck-content {
+	& picture {
+		/* The image should never exceed the width of the parent container */
+		& img {
+			max-width: 100%;
+		}
+	}
+
 	& .image {
 		display: table;
 		clear: both;
--- a/packages/ckeditor5-image/theme/image.css
+++ b/packages/ckeditor5-image/theme/image.css
@@ -4,13 +4,6 @@
  */
 
 .ck-content {
-	& picture {
-		/* The image should never exceed the width of the parent container */
-		& img {
-			max-width: 100%;
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


arkflpc (approved):

packages/ckeditor5-image/theme/image.css
```diff
@@ -4,6 +4,13 @@
  */
 
 .ck-content {
+	& picture {
+		/* The image should never exceed the width of the parent container */
+		& img {
+			max-width: 100%;
+		}
+	}
```
niegowski, line 12: Is this part correct? I was surprised when I saw the `picture` here.

mmotyczynska: It's correct. But actually I am not planning to add this CSS rule at all to be consistent with other inline images. But I'd keep the manual tests introduced in this ticket.

PR #14545, opened 2023-07-07
Added metadata validation script

przemyslaw-zan:
Internal: Added script that is executed on CI to check if all icon paths in plugin metadata are valid.


scripts/ci/validate-metadata-icons.js
```diff
@@ -0,0 +1,88 @@
...
+			}
+
+			if ( missing ) {
+				missingIcons.push( [ `./packages/${ packageName }/ckeditor5-metadata.json`, iconSource ] );
+			}
+		}
+	}
+
+	if ( !missingIcons.length ) {
+		return;
```
psmyrek, line 48: Maybe a confirmation that validations is successful?

scripts/ci/validate-metadata-icons.js
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	console.log( red( 'Detected invalid paths. Check the following paths in following packages:' ) );
```
psmyrek, line 51: "Detected invalid paths" - what paths? The message sent before ("Validating metadata files.") also does not suggest to someone who sees such errors for the first time what paths it is about.

```suggestion
	console.log( red( 'Detected invalid icon paths in the following packages:' ) );
```

scripts/ci/validate-metadata-icons.js
```diff
@@ -0,0 +1,88 @@
...
+	console.log( red( table( missingIcons, { border: getBorderCharacters( 'ramac' ) } ) ) );
```
psmyrek, line 52: Border type = `ramac` is for terminals that does not support Unicode characters. I think we can safely assume that all terminals support nice border characters like `┌` or `┼`.

```suggestion
	console.log( red( table( missingIcons, { border: getBorderCharacters( 'norc' ) } ) ) );
```

scripts/ci/validate-metadata-icons.js
```diff
@@ -0,0 +1,88 @@
...
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/* eslint-env node */
+
+'use strict';
+
+const fs = require( 'fs-extra' );
+const glob = require( 'fast-glob' );
```
psmyrek, line 11: Missing `fast-glob` package in `package.json`, but there is already the `glob` package. Maybe we could use `glob` instead?

scripts/ci/validate-metadata-icons.js
```diff
@@ -0,0 +1,88 @@
...
+const upath = require( 'upath' );
+const { table, getBorderCharacters } = require( 'table' );
```
psmyrek, line 13: Missing `table` package in `package.json`.

scripts/ci/validate-metadata-icons.js
przemyslaw-zan: Travis respectfully disagrees. Switching back to `ramac`

obraz

scripts/ci/validate-metadata-icons.js
przemyslaw-zan: Since we have to add `table` anyway, we might as well add `fast-glob`. This is for CI, so even tiny performance improvement matters.

psmyrek (approved):
LGTM 🎉

PR #14563, opened 2023-07-11
Improving watchdog performance

### LINKED ISSUE #13098: Watchdog autosave feature causes performance problems

One of the most important Watchdog feature is restarting the editor when there is an unhandled error. To prevent losing the editor data, the watchdog internally saves the editor data, so it is able to restore it after the editor is restarted.

The problem is that this can lead to performance problems. The saving is done differently than in autosave feature. Instead of debouncing the save event on editor change (as in autosave), the watchdog throttles it. This means that for big documents, there will be a lag on first typed character and then after every 5 seconds as the user keeps on typing.

This puts further pressure on applications which handle big documents and additionally use autosave feature.

What's more, saving data by watchdog does not make any sense if the editor is used in real-time environment as the data is taken from the remote CS session anyway. But there's no setting to prevent the saves. This can be probably "hacked" by setting `saveInterval` to `2147483647` (this is the biggest value that can be used in JS timeouts). This should be either described in the documentation or even maybe watchdog should overwrite this setting if RTC is used.

One possible way to improve watchdog performance is to save model state (+markers) instead of document data. Since we are restarting the editor using the same configuration, it should be safe to use the model and it should be faster to save and re-init the editor. One problem here is that incorrectly implemented custom features may be saving some important data outside of the model (this can happen in non-RTC integrations, as in RTC applications this kind of bug should can show up faster as the non-model data is not shared among users).

DawidKossowski:

Fix (watchdog): Improved the performance of the Watchdog mechanism which works in the background to keep the data updated. Closes #13098.

---

### Additional information


packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -323,6 +342,79 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+		Object.entries( this._data! ).forEach( ( [ root, { content, markers, attributes } ] ) => {
+			const parsedNodes = JSON.parse( content );
+			const children = [];
+
+			if ( this._isMultiRootEditor( editor ) ) {
+				editor.addRoot( root );
+			}
+
+			for ( const [ key, value ] of attributes ) {
+				editor.model.document.getRoot( root )!._setAttribute( key, value );
```
scofalik, line 391: We can do this through `writer`.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -165,15 +172,21 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+				// 	const updatedConfig = Object.assign( {}, this._config, {
+				// 		initialData: this._data
+				// 	} );
+				//
+				// 	return this.create( this._elementOrData, updatedConfig, updatedConfig.context );
+				// }
 
-					return this.create( this._elementOrData, updatedConfig, updatedConfig.context );
-				}
+				EditorWatchdogInitPlugin.editorData = this._data!;
```
scofalik, line 185: You can pass this in config (define internal/temp config property just for this plugin).

scofalik (changes requested):

scofalik:
Okay, so, the `Watchdog`, if not used with RTC, seems to have now a problem with:

1.  Revision history (when re-inited, the data is set as initial, so the revisions may get messed up).
2.  Comments and track changes and revision history if integrated without adapters (and also if integrated with adapters but if the data has not been sent on time).

For now, let's just switch from setting data through HTML to setting data using the model. In next step, let's try to fix problems above.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -284,14 +306,30 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+					affectsData: marker._affectsData
+				};
+			}
 
-		for ( const rootName of this._editor!.model.document.getRootNames() ) {
-			data[ rootName ] = this._editor!.data.get( { rootName } );
-		}
+			data[ root ] = {
+				content: JSON.stringify( Array.from( editor.model.document.getRoot( root )!.getChildren() ) ),
+				attributes: Array.from( editor.model.document.getRoot( root )!.getAttributes() ),
```
scofalik, line 329:
```suggestion
				content: JSON.stringify( Array.from( root.getChildren() ) ),
				attributes: Array.from( root.getAttributes() ),
```

scofalik: I will this comment even though I noticed that you "roots" is actually "root names". The name for the variable is incorrect. I leave it to show you that it is important to correctly name stuff. My first reaction was that you have an error somewhere.

BTW. `Document#getRoots()` will be available very soon (hopefully in this release as well).

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -284,14 +306,30 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
-		const data: Record<string, string> = {};
+	private _getData(): EditorData {
+		const editor = this.editor!;
+		const roots = editor.model.document.getRootNames();
+		const data: EditorData = {};
+
+		roots.forEach( root => {
+			const markersData = {} as any;
+
+			for ( const marker of editor.model.markers ) {
```
scofalik, line 319: This seems incorrect as you will get all markers while you need to only get markers for a given root. Unless you will save data as:

```
{
	roots: {
		rootA: {
			data: ...,
			attributes: ...,
		}, ...
	},
	markers: []
}
```

It seems there's not much difference what will you choose here as this is internal data structure that is used almost right-away. So, I'd propose the one that I proposed :D as it seems it will requre less code (you don't need to filter markers).

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -323,6 +361,89 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
 
+class EditorWatchdogInitPlugin extends Plugin {
+	private _data: EditorData;
+
+	constructor( editor: Editor ) {
+		super( editor );
+
+		this._data = editor.config.get( '_watchdogInitialData' )!;
+
+		editor.config.set( '_watchdogInitialData', null );
```
scofalik, line 372: Why this is needed? If it is important, write that in inline comment.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -323,6 +361,89 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public init(): void {
+		this.editor.data.on( 'init', evt => {
+			evt.stop();
+
+			this.editor.model.enqueueChange( writer => {
```
scofalik, line 382: Use proper batch type same as in `DataController#set`.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -323,6 +361,89 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+					children.push( Element.fromJSON( child ) );
+				} else {
+					// Otherwise, it is a Text node.
+					children.push( Text.fromJSON( child ) );
+				}
+			}
+
+			Object.entries( markers ).forEach( ( [ markerName, markerOptions ] ) => {
+				writer.addMarker( markerName, markerOptions );
+			} );
```
scofalik, line 427: Add content first, then markers. For a while, these markers are created on non-existing model tree. If there's a piece of code that would check something with markers right away, it will crash.

scofalik (changes requested):

packages/ckeditor5-watchdog/src/editorwatchdog.ts
DawidKossowski: I forgot to refactor it, sorry

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -165,12 +164,24 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+					return acc;
+				}, {} as Record<string, string> );
+
+				const updatedConfig = {
+					...this._config,
+					plugins: this._config!.plugins || [],
+					_watchdogInitialData: this._data
+				};
+
+				updatedConfig.plugins!.push( EditorWatchdogInitPlugin as any );
```
scofalik, line 179: I would push it to `extraPlugins`. AFAIR, if you `config.plugins` is not set, then setting it to `[ WatchdogInitPlugin ]` will result in an editor that has no plugins. And it is possible to leave `config.plugins` empty if you use editor build. So, do the same, but use `extraPlugins`.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -165,12 +164,24 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+
 				if ( typeof this._elementOrData === 'string' ) {
-					return this.create( this._data, this._config, this._config!.context );
+					return this.create( existingRoots, updatedConfig, updatedConfig!.context );
 				} else {
-					const updatedConfig = Object.assign( {}, this._config, {
-						initialData: this._data
-					} );
+					updatedConfig.initialData = existingRoots;
```
scofalik, line 184: I am not sure if this will work correctly for multi-root editor, if you add some roots before there's editor restart. It may be that there will be letfovers in DOM after the crashed editor is destroyed. And then, after you restart, you will use only DOM elements for the initial roots (as they were saved in `this._elementOrData`). For the new roots, you will again create new DOM elements. Please check if that really happens.

In this case, you will need to update `this._elementOrData` with proper DOM elements as well :thinking:. Or fix it in a different way :D.

scofalik: Anyway, please remember, that `this._elementOrData` may still be an object with strings or with DOM elements, and it is supported for both multi-root and single-root editors.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -284,13 +295,31 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+			roots: {},
+			markers: {}
+		};
+
+		rootNames.forEach( rootName => {
+			const root = editor.model.document.getRoot( rootName )!;
+
+			data.roots[ rootName ] = {
+				content: JSON.stringify( Array.from( root.getChildren() ) ),
+				attributes: Array.from( root.getAttributes() )
```
scofalik, line 313: One thing I am worried about is that if there is an object saved as an attribute, you will get a reference to this object. I am not sure if `getAttributes()` or setting attributes through writer handles it in any way. I am afraid that a reference to this object may cause:

1.  Unexpected changes if the original object got changed somewhere.
2.  Memory leaks, if the original object had a reference to something.

In general, people should not use objects as attributes and surely not objects that keep non-primitve data types (other than maybe more plain objects, you get the point). But maybe to be perfectly safe, we should also stringify/parse this data?

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -323,6 +352,110 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
 	}
 }
 
+class EditorWatchdogInitPlugin {
```
scofalik, line 355: API docs or inline comments at least for this class, even though it is private internal thingy.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -323,6 +352,110 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+		this.editor = editor;
+
+		this._data = editor.config.get( '_watchdogInitialData' )!;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public init(): void {
+		this.editor.data.on( 'init', evt => {
```
scofalik, line 369: Explain in a few sentences what is going on here.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -323,6 +352,110 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+				this._restoreEditorData( writer );
+			} );
+
+			this.editor.data.fire<EditorReadyEvent>( 'ready' );
+
+			// Keep priority `'high' - 1` to be sure that RTC initialization will be first.
+		}, { priority: 1000 - 1 } );
+	}
+
+	private _createElement( writer: Writer, node: any ): Text | Element {
```
scofalik, line 382: I'd rename to `_createNode` :grimacing:

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -323,6 +352,110 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+				writer.setAttribute( key, value, rootElement );
+			}
+
+			for ( const child of parsedNodes ) {
+				children.push( this._createElement( writer, child ) );
+			}
+
+			frag._appendChild( children );
+
+			writer.insert( frag, rootElement );
```
scofalik, line 423: At this point, if we are creating the nodes one after another, I think that we can simply add them directly to the root (using writer). Did you encounter any problems with that?

scofalik (changes requested):

packages/ckeditor5-watchdog/src/editorwatchdog.ts
DawidKossowski: Yes, you are right and it doesn't work after adding/removing roots in the multi-root editor. As we discussed, it will be fixed in the next step (operations approach).

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -284,13 +295,31 @@ export default class EditorWatchdog<TEditor extends Editor = Editor> extends Wat
...
+			data.roots[ rootName ] = {
+				content: JSON.stringify( Array.from( root.getChildren() ) ),
+				attributes: JSON.stringify( Array.from( root.getAttributes() ) )
+			};
+		} );
 
-		for ( const rootName of this._editor!.model.document.getRootNames() ) {
-			data[ rootName ] = this._editor!.data.get( { rootName } );
+		for ( const marker of editor.model.markers ) {
+			data.markers[ marker.name ] = {
```
scofalik, line 318: Filter out markers that do not affect the data (we talked about it for the "next step" but AFAIU it makes sense here as well).

scofalik (changes requested):
Hopefully final changes.

PR #14564, opened 2023-07-11
CKE5 properly reflects table alignments from Microsoft Word 

### LINKED ISSUE #8752: Support for table alignment when copying from Word

1. In a Word document, create a table and left-align it
2. Copy the table and paste it into CK Editor

### ✔️ Expected result

The table should stay left-aligned in CK Editor.

### ❌ Actual result

The table was centre-aligned.

## 📃 Other details

The issue can be reproduced here: https://ckeditor.com/docs/ckeditor5/latest/features/pasting/paste-from-word.html

---

illia-stv:

Fix (paste from word): CKE5 properly reflects table alignments from Microsoft Word . Closes #8752.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-paste-from-office/src/filters/table.ts
+++ b/packages/ckeditor5-paste-from-office/src/filters/table.ts
@@ -29,7 +29,7 @@ export function tableAlignmentFilter( documentFragment: ViewDocumentFragment, wr
 		const align = item.getAttribute( 'align' );
 		const child = item.getChild( 0 );
 
-		// Element should have alignemt attribute and should have child and should be div.
+		// Element should have alignment attribute and should have child and should be div.
 		if ( !align || !child || item.name !== 'div' ) {
 			continue;
 		}
--- a/packages/ckeditor5-paste-from-office/src/filters/table.ts
+++ b/packages/ckeditor5-paste-from-office/src/filters/table.ts
@@ -15,6 +15,10 @@ import type { UpcastWriter, ViewDocumentFragment } from 'ckeditor5/src/engine';
  */
 export function tableAlignmentFilter( documentFragment: ViewDocumentFragment, writer: UpcastWriter ): void {
 	for ( const item of documentFragment.getChildren() ) {
+		if ( !item.is( 'element' ) ) {
+			continue;
+		}
+
 		// If table is not wrapped in div[align] it should be alligned left.
 		// More details:(https://github.com/ckeditor/ckeditor5/issues/8752#issuecomment-1623507171).
 		if ( item.is( 'element', 'table' ) ) {
@@ -22,25 +26,18 @@ export function tableAlignmentFilter( documentFragment: ViewDocumentFragment, wr
 			continue;
 		}
 
-		const align = item.is( 'element' ) && item.getAttribute( 'align' );
-		const child = item.is( 'element' ) && item.getChild( 0 );
+		const align = item.getAttribute( 'align' );
+		const child = item.getChild( 0 );
 
 		// Element should have alignment attribute and should have child and should be div.
 		if ( !align || !child || item.name !== 'div' ) {
 			continue;
 		}
 
-		// If table is wrapped in div[right|left] table should be alligned right or left.
+		// If table is wrapped in div[align] table should be alligned with defined alignment value.
 		// More details:(https://github.com/ckeditor/ckeditor5/issues/8752#issuecomment-1629065676)
-		if ( align !== 'center' && child.is( 'element', 'table' ) ) {
-			writer.setAttribute( 'align', align, child );
-			continue;
-		}
-
-		// If table is wrapped in div[center] table should be centered.
-		// More details:(https://github.com/ckeditor/ckeditor5/issues/8752#issuecomment-1629065676)
-		if ( align === 'center' && child.is( 'element', 'table' ) ) {
-			writer.setAttribute( 'align', 'none', child );
+		if ( child.is( 'element', 'table' ) ) {
+			writer.setAttribute( 'align', align === 'center' ? 'none' : align, child );
 		}
 	}
 }
--- a/packages/ckeditor5-paste-from-office/src/filters/table.ts
+++ b/packages/ckeditor5-paste-from-office/src/filters/table.ts
@@ -6,14 +6,16 @@
 /**
  * @module paste-from-office/filters/parse
  */
-import type { UpcastWriter, ViewDocumentFragment } from 'ckeditor5/src/engine';
+import { UpcastWriter, type ViewDocumentFragment } from 'ckeditor5/src/engine';
 
 /**
  * Set alignment for table pasted from MS Word.
  *
  * @param documentFragment The view structure to be transformed.
  */
-export function tableAlignmentFilter( documentFragment: ViewDocumentFragment, writer: UpcastWriter ): void {
+export function setTableAlignment( documentFragment: ViewDocumentFragment ): void {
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-paste-from-office/src/filters/table.ts
```diff
@@ -0,0 +1,46 @@
...
+		const child = item.is( 'element' ) && item.getChild( 0 );
+
+		// Element should have alignment attribute and should have child and should be div.
+		if ( !align || !child || item.name !== 'div' ) {
+			continue;
+		}
+
+		// If table is wrapped in div[right|left] table should be alligned right or left.
+		// More details:(https://github.com/ckeditor/ckeditor5/issues/8752#issuecomment-1629065676)
+		if ( align !== 'center' && child.is( 'element', 'table' ) ) {
```
mlewand, line 35: Can't the following lines: [https://github.com/ckeditor/ckeditor5/blob/2b5758c590983b9bf06812c828662f060487a55f/packages/ckeditor5-paste-from-office/src/filters/table.ts#L33-L44](https://github.com/ckeditor/ckeditor5/blob/2b5758c590983b9bf06812c828662f060487a55f/packages/ckeditor5-paste-from-office/src/filters/table.ts#L33-L44)

Be simplified down to:

```javascript
if ( child.is( 'element', 'table' ) ) {
	writer.setAttribute( 'align', align === 'center' ? 'none' : align, child );
}
```

I don't have the option to run tests at hand.

packages/ckeditor5-paste-from-office/src/filters/table.ts
```diff
@@ -0,0 +1,46 @@
...
+export function tableAlignmentFilter( documentFragment: ViewDocumentFragment, writer: UpcastWriter ): void {
+	for ( const item of documentFragment.getChildren() ) {
+		// If table is not wrapped in div[align] it should be alligned left.
+		// More details:(https://github.com/ckeditor/ckeditor5/issues/8752#issuecomment-1623507171).
+		if ( item.is( 'element', 'table' ) ) {
+			writer.setAttribute( 'align', 'left', item );
+			continue;
+		}
+
+		const align = item.is( 'element' ) && item.getAttribute( 'align' );
```
mlewand, line 25: You're using `item.is( 'element' )` check extensively and AFAICS this logic indeed only applies to item that is an element.

In that case it make sense to do a check:

```javascript
if ( !item.is( 'element' ) ) {
	continue;
}
```

At the beginning of for loop.

It will also simplify `item.is( 'element', 'table' )` check to => `item.name === 'table'`.

mlewand (commented):
Just a few notes.

packages/ckeditor5-paste-from-office/src/normalizers/mswordnormalizer.ts
```diff
@@ -42,11 +42,12 @@ export default class MSWordNormalizer implements Normalizer {
 	 * @inheritDoc
 	 */
 	public execute( data: NormalizerData ): void {
+		const writer = new UpcastWriter( this.document );
 		const { body: documentFragment, stylesString } = data._parsedData;
 
 		transformListItemLikeElementsIntoLists( documentFragment, stylesString );
 		replaceImagesSourceWithBase64( documentFragment, data.dataTransfer.getData( 'text/rtf' ) );
-
+		tableAlignmentFilter( documentFragment, writer );
```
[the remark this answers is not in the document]
Dumluregn: Above we have two functions with nice command-like names, starting with the verb:

*   **transform**ListItemLikeElementsIntoLists
*   **replace**ImagesSourceWithBase64

Let's stick to this convention.

Additionally, with the current name the function can be misunderstood (filtering may also be understood as removing something). Something like `setTableAlignment(`) tells more about it's true nature.

packages/ckeditor5-paste-from-office/src/filters/table.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+import type { UpcastWriter, ViewDocumentFragment } from 'ckeditor5/src/engine';
+
+/**
+ * Set alignment for table pasted from MS Word.
+ *
+ * @param documentFragment The view structure to be transformed.
+ */
+export function tableAlignmentFilter( documentFragment: ViewDocumentFragment, writer: UpcastWriter ): void {
+	for ( const item of documentFragment.getChildren() ) {
```
Dumluregn, line 17: With the shallow iterator like this, the tables nested deeper than in one container will not be caught. I don't have an Office license now, but I imagine things like table inside the blockquote or just a table inside a table are possible. So we need to make this look as deep as possible.

You can get inspired by how it's done in other filters: https://github.com/ckeditor/ckeditor5/blob/cdebba2da7fe5746833dcd3e51e72115bb99c5f6/packages/ckeditor5-paste-from-office/src/filters/image.ts#L67-L75

packages/ckeditor5-paste-from-office/src/filters/table.ts
```diff
@@ -0,0 +1,43 @@
...
+
+		// Element should have alignment attribute and should have child and should be div.
+		if ( !align || !child || item.name !== 'div' ) {
+			continue;
+		}
+
+		// If table is wrapped in div[align] table should be alligned with defined alignment value.
+		// More details:(https://github.com/ckeditor/ckeditor5/issues/8752#issuecomment-1629065676)
+		if ( child.is( 'element', 'table' ) ) {
+			writer.setAttribute( 'align', align === 'center' ? 'none' : align, child );
```
Dumluregn, line 40: I'm unsure about setting the `align` attribute on view `table` elements. As per [https://github.com/ckeditor/ckeditor5/issues/8752#issuecomment-1629065676:](https://github.com/ckeditor/ckeditor5/issues/8752#issuecomment-1629065676:)

> *   that has a parent element `div[align=left]` or `div[align=right]` then the model's `table` element should get `tableAlignment` set to `left` or `right`.

If you apply it as you did, I'm not sure it will work with the table properties as it uses the mentioned `tableAlignment` property in model and styling (`float: right/left`) in view. Pasted table should have the correct alignment selected.

Dumluregn (commented):
I don't have the Office license, so unfortunately my review is entirely code-based. Still, I noticed some things to fix.

packages/ckeditor5-paste-from-office/src/filters/table.ts
illia-stv: With `tableAlignment` it doesn't work, I guess because in upcast you could just apply DOM-like attributes.

packages/ckeditor5-paste-from-office/src/filters/table.ts
illia-stv: There is no need indeed. MS Word doesn't allow you to have table in other blocks as list for example. Blockquote is not a block in the MS Word, it just spacing between corner of document and your paragraph or other elements. How to add blockquote in MS Word could check it here https://www.howtogeek.com/697529/how-to-add-block-quotes-in-microsoft-word/

Dumluregn: > MS Word doesn't allow you to have table in other blocks

I confirmed it with the converters team, so the only scenario where you can have a table inside a block element is a table inside a table. I agree it would be a rare scenario, so we can go with the performance micro-optimisation you've proposed. I'll extract an issue for the mentioned case so we stay aware of it.

Dumluregn: Reported in #14609.

PR #14577, opened 2023-07-13
Set image width and height on image change

### LINKED ISSUE #14557: Set image width and height on resize (and possibly other image changes)

Image width and height should be added when the user edits the image.
Currently, image sizes are added after switching from block to inline (and vice versa).
This should also happen when the image is resized and, if possible, when other changes are made to the image (e.g. alignment).

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

mmotyczynska:

Task (image): Image width and height should be set on image change (resize, alignment, etc.). Closes #14557.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-image/src/imageutils.ts
+++ b/packages/ckeditor5-image/src/imageutils.ts
@@ -19,12 +19,15 @@ import type {
 	ViewDocumentFragment,
 	DowncastWriter,
 	Model,
-	Position
+	Position,
+	ViewContainerElement
 } from 'ckeditor5/src/engine';
 import { Plugin, type Editor } from 'ckeditor5/src/core';
 import { findOptimalInsertionRange, isWidget, toWidget } from 'ckeditor5/src/widget';
 import { determineImageTypeForInsertionAtSelection } from './image/utils';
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-image/src/imageutils.ts
```diff
@@ -211,6 +214,13 @@ export default class ImageUtils extends Plugin {
 		return this.isImage( selectedElement ) ? selectedElement : selection.getFirstPosition()!.findAncestor( 'imageBlock' );
 	}
 
+	/**
+	 * Returns an image widget editing view based on the passed image view.
+	 */
+	public getImageWidgetFromImageView( imageView: ViewElement ): ViewContainerElement | null {
+		return imageView.findAncestor( { classes: IMAGE_WIDGETS_CLASSES_MATCH_REGEXP } ) as ViewContainerElement;
```
arkflpc, line 221:
```suggestion
		return imageView.findAncestor( { classes: IMAGE_WIDGETS_CLASSES_MATCH_REGEXP } ) as ( ViewContainerElement | null );
```
It doesn't make a difference, though. But the cast is to the wrong type. Another alternative is to cast to `any` in order  to not repeat yourself.

arkflpc (approved):

PR #14591, opened 2023-07-15
Account for rowspan in table headingColumns

bendemboski:
When computing `headingColumns` during upcasting, the logic wasn't factoring in `<th>`s with `rowspan` attributes that "add" a heading column to subsequent rows. See tests and comments in the code for more detail.

Closes: #14826


Witoso:
Thanks for the PR! I created an issue for it #14826. We will take a look as soon as possible! :blush:

PR #14599, opened 2023-07-17
Make the `scrollAncestorsToShowRect()` function take the limiter element

### LINKED ISSUE #14598: Enable scrolling the ancestors up to target limiter

Make the [`scrollAncestorsToShowRect()` function](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/dom/scroll.ts#L295) accept the limiter element, which will stop it from scrolling further ancestors.

---

Dumluregn:

Feature(utils): Made the `scrollAncestorsToShowRect()` function take the limiter element, which will stop it from scrolling further ancestors. Closes #14598.

---

### Additional information

The code coverage was essentially provided in `scrollAncestorsToShowTarget()` function tests.

### DIFF
--- a/packages/ckeditor5-utils/src/dom/scroll.ts
+++ b/packages/ckeditor5-utils/src/dom/scroll.ts
@@ -292,26 +292,30 @@ function scrollWindowToShowRect<T extends boolean, U extends IfTrue<T>>(
  * @param options.forceScroll When set `true`, the `rect` will be aligned to the top of scrollable ancestors
  * whether it is already visible or not. This option will only work when `alignToTop` is `true`
  */
-function scrollAncestorsToShowRect<T extends boolean, U extends IfTrue<T>>(
+export function scrollAncestorsToShowRect<T extends boolean, U extends IfTrue<T>>(
 	{
 		parent,
 		getRect,
 		alignToTop,
 		forceScroll,
-		ancestorOffset = 0
+		ancestorOffset = 0,
+		limiterElement
 	}: {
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-utils/src/dom/scroll.ts
```diff
@@ -292,26 +292,30 @@ function scrollWindowToShowRect<T extends boolean, U extends IfTrue<T>>(
...
-function scrollAncestorsToShowRect<T extends boolean, U extends IfTrue<T>>(
+export function scrollAncestorsToShowRect<T extends boolean, U extends IfTrue<T>>(
 	{
 		parent,
 		getRect,
 		alignToTop,
 		forceScroll,
-		ancestorOffset = 0
+		ancestorOffset = 0,
+		limiterElement
```
oleq, line 302: Shouldn't we add this option to `scrollAncestorsToShowTarget()` and use it in the document outline instead of making this helper public?

I mean, this `limiterElement` argument would stay there (it needs docs, BTW) but we wouldn't need to make this low-level API exportable.

Dumluregn: TBH this choice was purely caused by the functions signatures.

`scrollAncestorsToShowTarget()` has 2 params of which one is optional. I didn't want to enforce passing the nullish `ancestorOffset` to use `limiterElement`. However DO was the only place that used this function and it was passing the `ancestorOffset` anyways, so I guess we can go with your proposal.

PR #14603, opened 2023-07-18
Added script that ensures that dependencies in CKE5 are "pinned" and …

martnpaneq:

Internal: Added a script that ensures that dependencies in CKEditor 5 packages are "pinned" and use the same version. 

### Ticket to close

Closes cksource/ckeditor5-internal#3463.

---

### Additional information

### DIFF
--- a/scripts/ci/check-dependencies-versions-match.js
+++ b/scripts/ci/check-dependencies-versions-match.js
@@ -10,15 +10,21 @@
 'use strict';
 
 // This script ensures that all "dependencies" in package JSONs listed below, use the same versions of
-// dependencies. It also checks that all versions and are pinned, and they don't use the caret operator "^".
-// If you provide "--fix" argument, the script will automatically fix the errors for you.
+// dependencies. It also checks that all versions are pinned, and they don't use the caret operator "^".
+// If you provide the "--fix" argument, the script will automatically fix the errors for you.
 
+const chalk = require( 'chalk' );
 const semver = require( 'semver' );
 const { globSync } = require( 'glob' );
 const fs = require( 'fs-extra' );
 const upath = require( 'upath' );
+const { execSync } = require( 'child_process' );
 
+const versionsCache = {};
 const shouldFix = process.argv[ 2 ] === '--fix';
+
+console.log( chalk.blue( '🔍 Starting checking dependencies versions...' ) );
+
 const [ packageJsons, pathMappings ] = getPackageJsons( [
 	'package.json',
 	'packages/*/package.json',
@@ -54,6 +60,8 @@ function fixDependenciesVersions( expectedDependencies, packageJsons, pathMappin
 
 			fs.writeJsonSync( pathMappings[ packageJson.name ], packageJson, { spaces: 2 } );
 		} );
+
+	console.log( chalk.green( '✅  All dependencies fixed!' ) );
 }
 
 /**
@@ -73,9 +81,12 @@ function checkDependenciesMatch( expectedDependencies, packageJsons ) {
 		);
 
 	if ( errors.length ) {
-		console.error( errors );
+		console.error( chalk.red( '❌  Errors found. Run this script with an argument: `--fix` to resolve the issues automatically:' ) );
+		console.error( chalk.red( errors.join( '\n' ) ) );
 		process.exit( 1 );
 	}
+
+	console.log( chalk.green( '✅  All dependencies are correct!' ) );
 }
 
 /**
@@ -99,7 +110,7 @@ function getExpectedDepsVersions( packageJsons ) {
 		.filter( Boolean )
 		.reduce( ( expectedDependencies, dependencies ) => {
 			Object.entries( dependencies ).forEach( ( [ dependency, version ] ) => {
-				expectedDependencies[ dependency ] = getNewestVersion( version, expectedDependencies[ dependency ] );
+				expectedDependencies[ dependency ] = getNewestVersion( version, expectedDependencies[ dependency ], dependency );
 			} );
 
 			return expectedDependencies;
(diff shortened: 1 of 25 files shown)


scripts/ci/check-dependencies-versions-match.js
```diff
@@ -0,0 +1,135 @@
...
+ * @param {String|undefined} versionA
+ * @param {String|undefined} versionB
+ * @return {String}
+ */
+function getNewestVersion( versionA, versionB ) {
+	const versionAStripped = semver.valid( semver.coerce( versionA ) ) || '0.0.0';
+	const versionBStripped = semver.valid( semver.coerce( versionB ) ) || '0.0.0';
+
+	return semver.gt( versionAStripped, versionBStripped ) ? versionAStripped : versionBStripped;
+}
```
psmyrek, line 122: I don't know if it's worth adding in this PR, but I'll describe the scenario I've been thinking about.

I'm not sure if we really should just truncate the range character `^` and take the rest as the target version of the package.

Let's assume we have the `lodash-es` package in `package.json` defined like this: `"lodash-es": ^4.17.11"`. The latest version on npm is 4.17.21, which may introduce significant bug fixes or vulnerabilities fixes comparing to 4.17.11. Previously, using the range, version 4.17.21 would be installed. Now, the exact version 4.17.11 will be installed - the version that someone once (probably a long time ago) entered into `package.json`.

So I have another idea: the function `getNewestVersion()` could first get the available versions for a given package and then calculate the newest one that matches the range:

```js
const { execSync } = require( 'child_process' );
const semver = require( 'semver' );

const versions = execSync( `npm view lodash-es versions --json`, { encoding: 'utf8' } );

// Returns 4.17.21, as it is the latest version that satisfies the ^4.17.11 range.
return semver.maxSatisfying( JSON.parse( versions ), '^4.17.11' );
```

As the npm calls are expensive, the script could cache the results for each package.

scripts/ci/check-dependencies-versions-match.js
```diff
@@ -0,0 +1,135 @@
...
+				if ( version !== expectedDependencies[ dependency ] ) {
+					return getWrongVersionErrorMsg( dependency, packageJson.name, version, expectedDependencies );
+				}
+			} )
+			.filter( Boolean )
+		);
+
+	if ( errors.length ) {
+		console.error( '❌ Errors found. Run this script with `--fix` to resolve the issues automatically.' );
+		console.error( errors );
```
psmyrek, line 79: In case of errors, it prints the array. It would look better with just strings:

errors

If there are no errors, the scripts just ends without any info that everything is fine.

scripts/ci/check-dependencies-versions-match.js
```diff
@@ -0,0 +1,135 @@
...
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/* eslint-env node */
+
+'use strict';
+
+// This script ensures that all "dependencies" in package JSONs listed below, use the same versions of
+// dependencies. It also checks that all versions and are pinned, and they don't use the caret operator "^".
```
psmyrek, line 13:
```suggestion
// dependencies. It also checks that all versions are pinned, and they don't use the caret operator "^".
```

scripts/ci/check-dependencies-versions-match.js
psmyrek, line 1: Do we want to expose this script in `package.json` in the root of `ckeditor5` repository? Now, probably no one will even know about it.

psmyrek (commented):
It works as expected, but I found a few things to consider.

scripts/ci/check-dependencies-versions-match.js
martnpaneq: This was actually a pretty good catch. Thanks for this point! Fixed!

scripts/ci/check-dependencies-versions-match.js
martnpaneq: Added a `script` named `"check-dependencies:versions-match"` in the root `package.json`.

scripts/ci/check-dependencies-versions-match.js
```diff
@@ -0,0 +1,162 @@
...
+ * @param {Array.<Object>} packageJsons
+ * @return {Object.<String, String>} expectedDependencies
+ */
+function getExpectedDepsVersions( packageJsons ) {
+	return packageJsons
+		.map( packageJson => packageJson.dependencies )
+		.filter( Boolean )
+		.reduce( ( expectedDependencies, dependencies ) => {
+			Object.entries( dependencies ).forEach( ( [ dependency, version ] ) => {
+				expectedDependencies[ dependency ] = getNewestVersion( version, expectedDependencies[ dependency ], dependency );
```
psmyrek, line 113: Optional arguments should be defined at the end:
```suggestion
				expectedDependencies[ dependency ] = getNewestVersion( dependency, version, expectedDependencies[ dependency ] );
```

scripts/ci/check-dependencies-versions-match.js
```diff
@@ -0,0 +1,162 @@
...
+		}, {} );
+}
+
+/**
+ * @param {String|undefined} newVersion
+ * @param {String|undefined} currentMaxVersion
+ * @param {String} packageName
+ * @return {String}
+ */
+function getNewestVersion( newVersion = '0.0.0', currentMaxVersion = '0.0.0', packageName ) {
```
psmyrek, line 126: Optional arguments should be defined at the end:
```suggestion
 * @param {String} packageName
 * @param {String} [newVersion='0.0.0']
 * @param {String} [currentMaxVersion='0.0.0']
 * @return {String}
 */
function getNewestVersion( packageName, newVersion = '0.0.0', currentMaxVersion = '0.0.0' ) {
```

scripts/ci/check-dependencies-versions-match.js
```diff
@@ -0,0 +1,162 @@
...
+	if ( newVersion.includes( '^' ) ) {
```
psmyrek, line 127: This only covers one type of a range. Instead, we can use `semver` to check if the version is valid. The `semver.valid()` method returns `null` for any range.

```suggestion
	if ( !semver.valid( newVersion ) ) {
```

PR #14614, opened 2023-07-19
Handling collaboration data in watchdog

### LINKED ISSUE #14585: Handling comments and suggestions data after restarting by Watchdog

Watchdog does not work well with non-RTC integrations (comments, suggestions). The markers are set in the content, however, all comments and suggestions data are not passed to the new editor instance which causes errors. 

The improvement should keep all comments and suggestion data that were created after initialization and restore it during restarting.

DawidKossowski:

Feature (watchdog): Watchdog will now restore the collaboration data (comment threads and suggestions) after the crash. Closes #14585.

### DIFF
--- a/packages/ckeditor5-watchdog/src/augmentation.ts
+++ b/packages/ckeditor5-watchdog/src/augmentation.ts
@@ -3,7 +3,7 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
-import type { EditorData } from './editorwatchdog';
+import type { CollaborationData, EditorData } from './editorwatchdog';
 
 declare module '@ckeditor/ckeditor5-core' {
 	interface EditorConfig {
@@ -13,6 +13,6 @@ declare module '@ckeditor/ckeditor5-core' {
 		 *
 		 * @internal
 		 */
-		_watchdogInitialData?: EditorData | null;
+		_watchdogInitialData?: EditorData & CollaborationData;
 	}
 }
--- a/packages/ckeditor5-watchdog/src/augmentation.ts
+++ b/packages/ckeditor5-watchdog/src/augmentation.ts
@@ -3,7 +3,7 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
-import type { CollaborationData, EditorData } from './editorwatchdog';
+import type { EditorData } from './editorwatchdog';
 
 declare module '@ckeditor/ckeditor5-core' {
 	interface EditorConfig {
@@ -13,6 +13,6 @@ declare module '@ckeditor/ckeditor5-core' {
 		 *
 		 * @internal
 		 */
-		_watchdogInitialData?: EditorData & CollaborationData;
+		_watchdogInitialData?: EditorData;
 	}
 }
--- a/packages/ckeditor5-watchdog/src/editorwatchdog.ts
+++ b/packages/ckeditor5-watchdog/src/editorwatchdog.ts
@@ -434,6 +434,7 @@ class EditorWatchdogInitPlugin {
 			}
 		} );
 
+		console.log( this._data!.markers );
 		Object.entries( this._data!.markers ).forEach( ( [ markerName, markerOptions ] ) => {
 			const { document } = editor.model;
 			const {
--- a/packages/ckeditor5-watchdog/src/editorwatchdog.ts
+++ b/packages/ckeditor5-watchdog/src/editorwatchdog.ts
@@ -27,7 +27,7 @@ import type { Node, Text, Element, Writer } from 'ckeditor5/src/engine';
 import type { CommentsRepository } from '@ckeditor/ckeditor5-comments';
 
 // eslint-disable-next-line ckeditor5-rules/no-cross-package-imports
-import type { TrackChanges, SuggestionJSON } from '@ckeditor/ckeditor5-track-changes';
+import type { TrackChanges, TrackChangesEditing, SuggestionJSON } from '@ckeditor/ckeditor5-track-changes';
 
 // eslint-disable-next-line ckeditor5-rules/no-cross-package-imports,ckeditor5-rules/allow-imports-only-from-main-package-entry-point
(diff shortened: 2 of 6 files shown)


packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -23,6 +23,15 @@ import { throttle, cloneDeepWith, isElement, type DebouncedFunc } from 'lodash-e
...
 import type { Node, Text, Element, Writer } from 'ckeditor5/src/engine';
 
+// eslint-disable-next-line ckeditor5-rules/no-cross-package-imports
+import type { CommentsRepository } from '@ckeditor/ckeditor5-comments';
+
+// eslint-disable-next-line ckeditor5-rules/no-cross-package-imports
+import type { TrackChanges, SuggestionJSON } from '@ckeditor/ckeditor5-track-changes';
+
+// eslint-disable-next-line ckeditor5-rules/no-cross-package-imports,ckeditor5-rules/allow-imports-only-from-main-package-entry-point
+import type { CommentThreadDataJSON } from '@ckeditor/ckeditor5-comments/src/comments/commentsrepository';
```
DawidKossowski, line 33: I will add a separate PR in the CF repo to export this type

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -381,6 +424,7 @@ class EditorWatchdogInitPlugin {
 			evt.stop();
 
 			this.editor.model.enqueueChange( { isUndoable: true }, writer => {
+				this._restoreCollaborationData();
```
scofalik, line 427: Not sure if that makes any difference, but in guides we put initialization straight in `init()` of integration plugins, instead of in `'init'` event. Same when it comes to the RTC adapters (they connect before `'init'` event).

scofalik: Maybe there's one difference -- handling RTC, actually. If we would want to prevent using watchdog restore in RTC, it is easier to put `this._restoreCollaborationData()` in the `'init'` callback as it will simply won't get fired.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -453,6 +497,27 @@ class EditorWatchdogInitPlugin {
 			} );
 		} );
 	}
+
+	/**
+	 * Restores the editor collaboration data - comment threads and suggestions.
+	 */
+	private _restoreCollaborationData() {
```
scofalik, line 504: We need to make sure that we don't add the same comment or suggestion twice. How it can happen? If there were starting comments/suggestions added in the integration plugin, like we describe in guides (look at load&save integration specifically).

scofalik: So, we need to either listen to which comments/suggestion were added, or we need save which were initially added, or here we need to skip these, which already exists (they were added earlier in `init()` methods of the integration plugins).

packages/ckeditor5-watchdog/tests/editorwatchdog.js
```diff
@@ -10,6 +10,35 @@ import Editor from '@ckeditor/ckeditor5-core/src/editor/editor';
 import ClassicTestEditor from '@ckeditor/ckeditor5-core/tests/_utils/classictesteditor';
 import CKEditorError from '@ckeditor/ckeditor5-utils/src/ckeditorerror';
 import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
+import { Comments } from '@ckeditor/ckeditor5-comments';
+import { TrackChanges } from '@ckeditor/ckeditor5-track-changes';
+import Suggestion from '@ckeditor/ckeditor5-track-changes/src/suggestion';
+
+class UsersInit {
```
scofalik, line 17: I've been wondering if we don't have to also "copy users" (the same way as we copy suggestions and comments) to the newly created editor. But for RTC that is not needed and for non-RTC we should not need more users than initially provided.

packages/ckeditor5-watchdog/src/augmentation.ts
```diff
@@ -0,0 +1,18 @@
...
+
+declare module '@ckeditor/ckeditor5-core' {
+	interface EditorConfig {
+
+		/**
+		 * The temporary property that is used for passing data to the plugin which restores the editor state.
+		 *
+		 * @internal
+		 */
+		_watchdogInitialData?: EditorData & CollaborationData;
```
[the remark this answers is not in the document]
DawidKossowski: I refactored it in the next PR

packages/ckeditor5-watchdog/tests/editorwatchdog.js
DawidKossowski: Yes, I was also wondering about it but I realized that we always initialize users by using separate plugins and also it's not a "fresh" issue.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
DawidKossowski: It caused errors. Fixed in the newest commit

packages/ckeditor5-watchdog/src/augmentation.ts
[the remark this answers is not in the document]
DawidKossowski: Done in the newest commit

packages/ckeditor5-watchdog/src/editorwatchdog.ts
DawidKossowski: PR in CF repo is awaiting.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
scofalik: Now as I think about it... I think that we actually should overwrite the already existing comment threads and suggestions as the saved data may be more up-to-date than the initial data (e.g. comments has been added to the thread).

scofalik: We also should remove threads and suggestions if needed.

PR #14623, opened 2023-07-20
CI: Added the "skipCompilingFramework" flag for testing script

martnpaneq:

Internal: CI: Added the `skipCompilingFramework` flag for the testing script to avoid compiling the CKEditor 5 framework twice when executing the same command in different scenarios.

See cksource/collaboration-features#5338.

---

### Additional information


pomek (commented):
As we do not add the `skipCoverage` option, I believe we can leave the old function name.

PR #14627, opened 2023-07-20
Update Placeholder after initialization

### LINKED ISSUE #9925: Update Placeholder after initialization
Hello,

I'd like to have the ability, for an Inline editor, to update the placeholder at runtime (i.e. after the editor has been created). Is there any way to do that? Currently the placeholder is set in the configuration options, so I can't change it later. I found this thread https://github.com/ckeditor/ckeditor5/issues/7383 but I think nothing has changed yet.
Thanks!

---

illia-stv:

MINOR BREAKING CHANGE (engine): placeholder could be be updated after initialization. Closes #9925.

---

### Additional information


packages/ckeditor5-engine/src/view/placeholder.ts
```diff
@@ -60,16 +60,32 @@ export function enablePlaceholder( { view, element, text, isDirectHost = true, k
...
+				hostElement: isDirectHost ? element : null
+			} );
+
+			// Update the placeholders right away.
+			view.change( writer => updateDocumentPlaceholders( doc, writer ) );
+		} );
+	}
+
+	if ( element.placeholder ) {
+		documentPlaceholders.get( doc )!.set( element, {
```
mlewand, line 79: Following logic is duplicated, please make a function that will be reused:

```
		element.on( 'change:placeholder', ( evtInfo, evt, text ) => {
			// Store information about the element placeholder under its document.
			documentPlaceholders.get( doc )!.set( element, {
				text,
				isDirectHost,
				keepOnFocus,
				hostElement: isDirectHost ? element : null
			} );

			// Update the placeholders right away.
			view.change( writer => updateDocumentPlaceholders( doc, writer ) );
```

packages/ckeditor5-engine/src/view/placeholder.ts
```diff
@@ -37,10 +38,9 @@ const documentPlaceholders = new WeakMap<Document, Map<Element, PlaceholderConfi
  * editable root elements.
  * @param options.keepOnFocus If set `true`, the placeholder stay visible when the host element is focused.
  */
-export function enablePlaceholder( { view, element, text, isDirectHost = true, keepOnFocus = false }: {
+export function enablePlaceholder( { view, element, isDirectHost = true, keepOnFocus = false }: {
```
mlewand, line 41: You have removed `options.text` property, yet it remains in the API docs:

```
* @param options.text Placeholder text.
```

It should be removed, since it's no longer there.

packages/ckeditor5-engine/src/view/editableelement.ts
```diff
@@ -42,6 +42,13 @@ export default class EditableElement extends ObservableMixin( ContainerElement )
 	 */
 	declare public isFocused: boolean;
 
+	/**
+	 * Placeholder of current editable element.
```
mlewand, line 46: "**current** editable" isn't fitting in this context.

By using `EditableElement` instance you're working only with that particular editable element. Yo're working only with this instance. There's no relation to other editables.

Also you could add a code listing in here showing how to change it in runtime, including the entire call chain, like you did in sample: `editor.editing.view.document.getRoot( 'main' ).placeholder = 'New placeholder';`.

mlewand (changes requested):
Couple of remarks, there was also no breaking change information - be sure to add it as this is a backward-incompatible change.

Also, I don't see a PR in collaboration features repo. You must also add a related PR in there as at least commenteditor.ts file uses the `enablePlaceholder` function.

mlewand (changes requested):
The following docs are no longer true:

[https://github.com/ckeditor/ckeditor5/blob/54a683fbd8af8dbf8482bcfc69c082bcc52060b3/packages/ckeditor5-engine/src/view/placeholder.ts#L27](https://github.com/ckeditor/ckeditor5/blob/54a683fbd8af8dbf8482bcfc69c082bcc52060b3/packages/ckeditor5-engine/src/view/placeholder.ts#L27)

You should update it so that it describes how one should change the placeholder from here on.

---

[https://github.com/ckeditor/ckeditor5/blob/ce44c03034b486bd57e65d3f3c6de173b70d8abf/packages/ckeditor5-editor-balloon/src/ballooneditorui.ts#L106](https://github.com/ckeditor/ckeditor5/blob/ce44c03034b486bd57e65d3f3c6de173b70d8abf/packages/ckeditor5-editor-balloon/src/ballooneditorui.ts#L106)

Be sure to check the surrounding API docs when changing something.

> Enable the placeholder text on the editing root, if any was configured.

"if any was configured" part is no longer true after your changes.

---

Logic like that: [https://github.com/ckeditor/ckeditor5/blob/ce44c03034b486bd57e65d3f3c6de173b70d8abf/packages/ckeditor5-editor-balloon/src/ballooneditorui.ts#L108-L128](https://github.com/ckeditor/ckeditor5/blob/ce44c03034b486bd57e65d3f3c6de173b70d8abf/packages/ckeditor5-editor-balloon/src/ballooneditorui.ts#L108-L128)

Is repeated in multiple places. It would be a good idea to make a common helper reused by all the editor UIs.

mlewand:
I'll handle the review comments to conclude this PR.

PR #14648, opened 2023-07-24
Focus observer should be flushed even if the new selection is equal to the document selection.

### LINKED ISSUE #14569: Editor has a reverse typing effect on slower machines

First reported in the #13926.

> It's not that the typing is "reversed", but more like the cursor is "stuck" and isn't updating -- so text just keeps getting added behind the cursor. Seems to be something with setting the DOM selection. The behavior seems to be more obvious when the JS thread is very busy and the user is typing during this time. Funny enough, when things settle slightly, the cursor is able to update again, and the user doesn't really need to do anything.
> 
> I was able to actually repro it a couple times in the inline editor demo here: [https://ckeditor.com/docs/ckeditor5/latest/examples/builds/inline-editor.html](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/inline-editor.html)
> 
> Simulating throttling of CPU by 4x/6x on DevTools, and loading the page before the content has settled allowed me to briefly reproduce the issue at some point.

### ✔️ Expected result

There's no reverse typing effect.

### ❌ Actual result

On slower machines, the text keeps getting added behind the cursor.

## 📃 Other details

*   Browser: …
*   OS: …

---

niegowski:

Fix (engine): Fixed reverse typing issue on editor gaining focus. Closes #14569.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/view/observer/selectionobserver.ts
+++ b/packages/ckeditor5-engine/src/view/observer/selectionobserver.ts
@@ -253,6 +253,9 @@ export default class SelectionObserver extends Observer {
 
 		this.view.hasDomSelection = true;
 
+		// Mark the latest focus change as complete (we got new selection after the focus so the selection is in the focused element).
+		this.focusObserver.flush();
+
 		if ( this.selection.isEqual( newViewSelection ) && this.domConverter.isDomSelectionCorrect( domSelection ) ) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-typing/src/inserttextobserver.ts
```diff
@@ -38,12 +39,20 @@ const TYPING_INPUT_TYPES = [
...
+	 */
+	public readonly focusObserver: FocusObserver;
+
 	/**
 	 * @inheritDoc
 	 */
 	constructor( view: View ) {
 		super( view );
 
+		this.focusObserver = view.getObserver( FocusObserver )!;
```
arkflpc, line 54: It makes this class depend on the external order of initialization. Which is the reason of all the ugly `!` we were forced to add everywhere.
I'd rather add the paranoid check (throw an error) about not having `FocusObserver` in `View`.

niegowski: The `FocusObserver` is part of the `engine` package and always added by the `View`. The `InsertTextObserver` is in the plugin in the `typing` package so `FocusObserver` is always there.

arkflpc (approved):

PR #14670, opened 2023-07-26
Prevent closing dropdown view on scrollbar click.

### LINKED ISSUE #14364: Don't close DropdownView when interacting with scrollbars

Currently, if the content of a `DropdownView` exceeds its size and scrollbars are shown, trying to interact with those scrollbars using the mouse will close the dropdown (without any action being triggered).

The behaviour seems to have been added in commit 633dead6b6d1a2d8deedd08ae47ea6d0de83092e – I assume for accessibility, because closing the dropdown on `blur` ensures that it will close if focus is moved away from it e.g. using the Tab key. If I remove the listener on the `blur` event of the dropdown, it stays open when using the Tab key to move away the focus. On the other hand, even with the `blur` event listener disabled, the dropdown will still close correctly if the user clicks outside of the dropdown with the mouse, or selects one of the elements of the dropdown.

The improvement I'd like to see is for mouse interaction with a scollbar that is part of the `DropdownView` to not trigger the `blur` handler, so the list can be proberly scrolled by dragging the scrollbar slider or clicking one of the scrollbar buttons. Currently, a dropdown view's content can only be scrolled using the keyboard arrow keys, or the scroll wheel of the mouse.

Here's a screencast to visualise. In it, I first scroll the dropdown using the cursor keys (work fine), then using the mouse wheel (works fine), then by trying to drag the scrollbar slider (closes the dropdown), and finally by trying to click a scrollbar stepper (closes the dropdown).
https://github.com/ckeditor/ckeditor5/assets/517549/0a5735a9-e020-4dc9-a5bc-a547a1ca255c

## 📃 Other details

* Browser: tested on Firefox 114.0, Chrome 114.0.5735.110, and Edge 114.0.1823.41
* OS: Windows 10 x64 (10.0.19044 Build 19044)
* CKEditor version: 38.0.1
* Installed CKEditor plugins: Track Changes, Comments, some custom ones (including custom toolbar elements)

Since our custom UI dropdowns can contain lists whose length exceeds the height of the element containing CKEditor (and well exceeds common screen size), we allow scrollbars in dropdowns through CSS:

```
.ck.ck-reset.ck-list {
    overflow-y: auto;
}
```

I assume that this is a pretty common requirement and setup.

Currently, since for our users it's much more imporant that the scrollbars can be interacted with properly, than keyboard focus changes behaving properly, I worked around the issue by simply removing the `closeDropdownOnBlur()` function from `ckeditor5-ui/src/dropdown/utils.ts` in our fork of CKEditor code, but of course this is not ideal.

---

mmotyczynska:

Fix (ui): Dropdown view should not be closed when interacting with a scrollbar. Closes #14364.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/dropdown/dropdown.css
+++ b/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/dropdown/dropdown.css
@@ -83,7 +83,7 @@
 		&.ck-off:active,
 		&.ck-on:active {
 			box-shadow: none;
-			
+
 			&:focus {
 				@mixin ck-box-shadow var(--ck-focus-outer-shadow);
 			}
@@ -118,5 +118,9 @@
 
 	&.ck-dropdown__panel_nw {
 		border-bottom-right-radius: 0;
+	}
+
+	&:focus {
+		outline: none;
 	}
 }
--- a/packages/ckeditor5-ui/src/dropdown/dropdownpanelview.ts
+++ b/packages/ckeditor5-ui/src/dropdown/dropdownpanelview.ts
@@ -69,7 +69,8 @@ export default class DropdownPanelView extends View implements DropdownPanelFocu
 					'ck-dropdown__panel',
 					bind.to( 'position', value => `ck-dropdown__panel_${ value }` ),
 					bind.if( 'isVisible', 'ck-dropdown__panel-visible' )
-				]
+				],
+				tabindex: '-1'
 			},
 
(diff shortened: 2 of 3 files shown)


niegowski:
You should disable native focus outline on focused dropdown panel, sth like here: 
https://github.com/ckeditor/ckeditor5/blob/e8f49348ffd05b1b6afb0d85477d35b5bc79b150/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/toolbar/toolbar.css#L109-L111

PR #14699, opened 2023-07-31
Wrong tooltip when hovering toolbar buttons on Chrome on iOS

### LINKED ISSUE #13812: Wrong tooltip when hovering toolbar buttons on Chrome on iOS

1. Connect external keyboard to an iPad.
2. Open Chrome browser.
3. Go to [online demo](https://ckeditor.com/ckeditor-5/demo/feature-rich/).
4. Focus the toolbar and navigate buttons until you reach `bold`.

### ✔️ Expected result

Tooltip shows `Bold (⌘B)`.

### ❌ Actual result

Tooltip shows `Bold (Ctrl+B)`.

### ❓ Possible solution

Extend checks in utils package:
- https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/keyboard.ts#L124
- https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/keyboard.ts#L205

`env.isMac || env.isiOS` instead of `env.isMac`.

## 📃 Other details

* Browser: …
* OS: …

---

illia-stv:

Fix (utils): Wrong tooltip when hovering toolbar buttons on Chrome on iOS. Closes #13812.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-utils/tests/keyboard.js
+++ b/packages/ckeditor5-utils/tests/keyboard.js
@@ -85,14 +85,60 @@ describe( 'Keyboard', () => {
 
 	describe( 'parseKeystroke', () => {
 		const initialEnvMac = env.isMac;
+		const initialEnviOS = env.isiOS;
 
 		afterEach( () => {
 			env.isMac = initialEnvMac;
+			env.isiOS = initialEnviOS;
 		} );
 
 		describe( 'on Macintosh', () => {
 			beforeEach( () => {
 				env.isMac = true;
+			} );
+
+			it( 'parses string', () => {
+				expect( parseKeystroke( 'ctrl+a' ) ).to.equal( 0x880000 + 65 );
+			} );
+
+			it( 'parses string without modifier', () => {
+				expect( parseKeystroke( '[' ) ).to.equal( 91 );
+			} );
+
+			it( 'allows spacing', () => {
+				expect( parseKeystroke( 'ctrl +   a' ) ).to.equal( 0x880000 + 65 );
+			} );
+
+			it( 'is case-insensitive', () => {
+				expect( parseKeystroke( 'Ctrl+A' ) ).to.equal( 0x880000 + 65 );
+			} );
+
+			it( 'works with an array', () => {
+				expect( parseKeystroke( [ 'ctrl', 'a' ] ) ).to.equal( 0x880000 + 65 );
+			} );
+
+			it( 'works with an array which contains numbers', () => {
+				expect( parseKeystroke( [ 'shift', 33 ] ) ).to.equal( 0x220000 + 33 );
+			} );
+
+			it( 'works with two modifiers', () => {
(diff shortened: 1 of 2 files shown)


mlewand (changes requested):
Functional: it works as intended on Safari & Firefox on iPad, however it's not fixed on Google Chrome (still shows ctrl+b there). How do we get it fixed on Chrome (on iPad/iPhone) as well?

Also, there is no test coverage for this change. Any change needs always to be backed up by unit tests that will fail against current `master` branch. This is what gives us confidence that we fixed an issue and that it will not appear in the future.

illia-stv:
> Functional: it works as intended on Safari & Firefox on iPad, however it's not fixed on Google Chrome (still shows ctrl+b there). How do we get it fixed on Chrome (on iPad/iPhone) as well?
> 
> Also, there is no test coverage for this change. Any change needs always to be backed up by unit tests that will fail against current `master` branch. This is what gives us confidence that we fixed an issue and that it will not appear in the future.

That's interesting because on my side on Iphone with chrome it works well. 

IMG_1519

For clarification I don't have Ipad, maybe it matters.

packages/ckeditor5-utils/tests/keyboard.js
```diff
@@ -232,6 +280,52 @@ describe( 'Keyboard', () => {
 			} );
 		} );
 
+		describe( 'on iOS', () => {
```
mlewand, line 283: The tests you added won't protect us from regressions. You can test it on mac.

Open terminal in your main ckeditor5 directory. Execute following 2 commands to undo your fix for logic (but to keep your tests).

```
git fetch --all
git checkout origin/master packages/ckeditor5-utils/src/keyboard.ts
```

Run tests, you can run just this single file to make it simpler:

```
yarn test -f utils/keyboard -w
```

## Expected

The tests should fail ❌ because there's no fix from this PR applied.

## Actual

The tests are passing.

## Reason

When running tests on mac, the `env.isMac` is set to true so it works due to the old logic:

[https://github.com/ckeditor/ckeditor5/blob/a8b1e91ba6855507b9d703a7e55076a05b2159be/packages/ckeditor5-utils/src/keyboard.ts#L124](https://github.com/ckeditor/ckeditor5/blob/a8b1e91ba6855507b9d703a7e55076a05b2159be/packages/ckeditor5-utils/src/keyboard.ts#L124)

That checks isMac.

In order to fix it you need to stub both `isMac` and `isiOS` (making sure that mac is false for ios tests, **and that ios is false for mac tests**).

mlewand (commented):
I can verify that indeed it works on chrome, so it must have been a cache issue most probably. Sorry about that!  
  
The only one issue remaining is related to unit tests and we're good for merging it :ok\_hand:

arkflpc (approved):

filipsobol:
LGTM, but I have no way to verify Marek's claims. Please make sure this is checked (and fixed if necessary) before merging.

PR #14630, opened 2023-07-21
Ballon panel sticks out of the limiter - generic approach.

pszczesniak:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Fix (ui): Should hide the ballon panel when it's scrolled outside the visible area.

Closes #5328.

---

### Additional information

_~Code coverage = W.I.P.~_

Different approach than done here 👉 #14528

PR #14713, opened 2023-08-02
Fix for balloon positioning where it would stick outside of an editable

mlewand:

Fix (ui,utils): A balloon panel should hide when its target (anchor) becomes invisible due to scrolling or cropping. Closes [#5328](https://github.com/ckeditor/ckeditor5/issues/5328).

MINOR BREAKING CHANGE (utils): The `getOptimalPosition()` helper will now return `null` (previously first available position) if the positioning criteria cannot be met, for instance, if the `target` is off the visible viewport.

---

### Additional information

This PR is based on #14630.


mlewand:
Closing in favor of #14755.

Issue #14586, opened 2023-07-14
Why do we have the `engine` property in pkg.json?


pomek:
## Provide a description of the task

As in the title. AFAIR, we have it from day 1.

https://github.com/ckeditor/ckeditor5/blob/26214b329d3c2904836d7c8657979349f9b4f9c4/package.json#L86-L89

However, I do not understand why need it. CKEditor 5 packages do not use Node features at all. Also, we do not use npm, except for publishing packages.

I believe we can drop the entire object, or for sure, the `npm` description if for some reason keeping `node` is valuable.

Reinmar:
I think we could drop this, indeed.

I don't remember any reasons for including it. It might've been our eagerness to be safer. But we don't rely on any of the npm/node features ourselves. Our dependencies might (e.g. if you need webpack to build CKE5). But that should be included in their checks.

pomek:
### Scope

* We agree to remove the `engine` property from all `ckeditor5-*` packages.
* Keeping the property in the development environment makes sense, as it points to a minimal version that a developer needs to run the CKEditor 5 project.
	* Drop the property While publishing the `ckeditor5` package on npm.

PR #14715, opened 2023-08-02
Removed `engines` from `package.json` in `ckeditor5-*` packages.

martnpaneq:

Internal: Removed `engines` from `package.json` in `ckeditor5-*` packages and removed `engines.npm` from the root package. See #14586.

---

### Additional information

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,8 +148,7 @@
     "webpack": "^5.58.1"
   },
   "engines": {
-    "node": ">=16.0.0",
-    "npm": ">=5.7.1"
+    "node": ">=16.0.0"
   },
   "author": "CKSource (http://cksource.com/)",
   "license": "GPL-2.0-or-later",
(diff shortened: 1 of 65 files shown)


pomek (changes requested):
Please, let's drop the [`engine.npm`](https://github.com/ckeditor/ckeditor5/blob/b82334152d0f6a4ba1a334d8a33bb15b81c82d43/package.json#L152) property from the root pkg.json. Also, we should make sure [it is not published to npm](https://github.com/ckeditor/ckeditor5/blob/b82334152d0f6a4ba1a334d8a33bb15b81c82d43/scripts/release/utils/getckeditor5packagejson.js#L28).

martnpaneq:
> Please, let's drop the [`engine.npm`](https://github.com/ckeditor/ckeditor5/blob/b82334152d0f6a4ba1a334d8a33bb15b81c82d43/package.json#L152) property from the root pkg.json. Also, we should make sure [it is not published to npm](https://github.com/ckeditor/ckeditor5/blob/b82334152d0f6a4ba1a334d8a33bb15b81c82d43/scripts/release/utils/getckeditor5packagejson.js#L28).

Removed `engines.npm` from the root package.
Also, after adding a callback to `packageJsonFieldsToRemove`, the editor is able to display the default values:
image

psmyrek (approved):
LGTM 🎉

PR #14718, opened 2023-08-02
CKBox command should (allow to) set focus in the opened dialog

### LINKED ISSUE #14312: CKBox command should (allow to) set focus in the opened dialog

1.  Open [v38.0.1 sample with CKBox](https://ckeditor.com/docs/ckeditor5/38.0.1/examples/builds-custom/full-featured-editor.html) 
2.  Evaluate following code in js console: `window.setTimeout( () => editor.execute('ckbox'), 5000 )`.
3.  make sure to move back your browser focus back the editor
4.  wait till dialog opens
5.  type random words

### ✔️ Expected result

Your window focus should go to CKBox dialog (your keyboard actions should be handled by CKBox).

### ❌ Actual result

Your focus is still in editor, thus you're changing the content of the editor.

### ❓ Possible solution

## 📃 Other details

*   Browser: …
*   OS: …

---

illia-stv:

Fix (ckbox): CKBox dialog is focused after initialization. Closes #14312.

---

### Additional information


packages/ckeditor5-ckbox/src/ckboxcommand.ts
```diff
@@ -387,6 +389,25 @@ function getAssetUrl( asset: CKBoxRawAssetDefinition ) {
...
+ *
+ * @param wrapper The element in DOM which wrap CKBox component.
+ */
+function focusDialog( wrapper: Element | null ) {
+	// If the DOM is not fully loaded and rendered by the time the focus() method is invoked,
+	// the element might not get focused. So we call the focus() when the thread becomes idle.
+	setTimeout( () => {
+		// Wrapper cannot be focused by default because it's covered by it's children entirely.
+		// Then we getting to the nearest focusable child( dialog container ) inside wrapper.
+		const dialogContainer = wrapper && wrapper.children[ 0 ] && wrapper.children[ 0 ].children[ 0 ];
```
Dumluregn, line 403: This is a very unstable way of selecting the element. You don't have any control over the component you are integrating and with such binding any change in its structure that you won't be aware of (like adding an extra wrapper) will break your code.

Unfortunately I don't think the component exposes some parts through it's API, but plugging into an element with a class `ckbox-gallery` seems to be much safer (it will work no matter where it will be moved unless the class name changes - but it looks pretty generic, so I think it shouldn't). And actually I'd focus it's first child - the first image in the gallery (as I think that's the most common intention when opening the dialog).

packages/ckeditor5-ckbox/tests/ckboxcommand.js
```diff
@@ -159,6 +159,26 @@ describe( 'CKBoxCommand', () => {
 				expect( document.body.appendChild.args[ 0 ][ 0 ] ).to.equal( wrapper );
 			} );
 
+			it( 'should focus dialog after initialization', () => {
```
Dumluregn, line 162: The test will need adjustments after changing the target accessing method.

Dumluregn (changes requested):
It doesn't work for me - the focused element becomes

```
<div class="ckbox ckbox-theme-default ckbox--dialog ckbox-theme-default" tabindex="-1">
```

and I still can't control the CKBox dialog. Tested with both toolbar button in http://localhost:8125/ckeditor5-ckbox/tests/manual/ckbox.html and slash command in http://192.168.100.6:8080/ckeditor5/38.1.1/examples/builds-custom/full-featured-editor.html.

arkflpc (approved):

PR #14753, opened 2023-08-04
enablePlaceholder() API should remain backward compatible for some time

### LINKED ISSUE #14743: enablePlaceholder() API should remain backward compatible for some time

`enablePlaceholder()` is used more commonly than we anticipated.

It's possible to introduce our change in a backward-compatible way. We just need to keep support text option passed to this helper which will set the `.placeholder` property on a passed `element`.

However, using the old API should leave a notice in the dev console that you're using old API which will be replaced soon.

Note, that this console log must not appear more than once during a given page reload.

## 📃 Other details

*   Browser: …
*   OS: …
*   CKEditor version: …

---

illia-stv:

Fix (engine): enablePlaceholder() API remain backward compatible for some time. Closes #14743.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/view/placeholder.ts
+++ b/packages/ckeditor5-engine/src/view/placeholder.ts
@@ -20,6 +20,8 @@ import type { ObservableChangeEvent } from '@ckeditor/ckeditor5-utils';
 // Each document stores information about its placeholder elements and check functions.
 const documentPlaceholders = new WeakMap<Document, Map<Element, PlaceholderConfig>>();
 
+let hasDisplayedWarning = false;
+
 /**
  * A helper that enables a placeholder on the provided view element (also updates its visibility).
  * The placeholder is a CSS pseudo–element (with a text content) attached to the element.
@@ -37,10 +39,11 @@ const documentPlaceholders = new WeakMap<Document, Map<Element, PlaceholderConfi
  * editable root elements.
  * @param options.keepOnFocus If set `true`, the placeholder stay visible when the host element is focused.
  */
-export function enablePlaceholder( { view, element, isDirectHost = true, keepOnFocus = false }: {
+export function enablePlaceholder( { view, element, text, isDirectHost = true, keepOnFocus = false }: {
 	view: View;
 	element: PlaceholderableElement | EditableElement;
 	isDirectHost?: boolean;
+	text?: string;
 	keepOnFocus?: boolean;
 } ): void {
 	const doc = view.document;
@@ -67,6 +70,12 @@ export function enablePlaceholder( { view, element, isDirectHost = true, keepOnF
 
 	if ( element.placeholder ) {
 		setPlaceholder( element.placeholder );
+	} else if ( text ) {
+		setPlaceholder( text );
+	}
+
+	if ( text ) {
+		showDeprecationWarning();
 	}
 
 	function setPlaceholder( text: string ) {
@@ -297,6 +306,19 @@ function getChildPlaceholderHostSubstitute( parent: Element ): Element | null {
 	}
 
 	return null;
+}
+
+/**
+ * Displays a deprecation warning message in the console, but only once per page load.
+ */
+function showDeprecationWarning() {
+	if ( !hasDisplayedWarning ) {
+		console.warn( 'Deprecation Warning: The "text" argument in the "enableProperty" function will be deprecated ' +
+		'in the next version. Please update your code. Refer to the documentation for alternative usage: ' +
+		'https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-39.html#view-element-placeholder' );
+	}
+
+	hasDisplayedWarning = true;
 }
 
 /**
--- a/packages/ckeditor5-engine/tests/view/placeholder.js
+++ b/packages/ckeditor5-engine/tests/view/placeholder.js
@@ -3,6 +3,8 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
+/* globals console */
+
 import {
 	enablePlaceholder,
 	disablePlaceholder,
@@ -618,8 +620,7 @@ describe( 'placeholder', () => {
 
 			enablePlaceholder( {
 				view,
-				element: viewRoot,
-				text: 'foo bar baz'
+				element: viewRoot
 			} );
 			viewRoot.placeholder = 'new placeholder';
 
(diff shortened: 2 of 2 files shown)


packages/ckeditor5-engine/src/view/placeholder.ts
```diff
@@ -20,6 +20,8 @@ import type { ObservableChangeEvent } from '@ckeditor/ckeditor5-utils';
 // Each document stores information about its placeholder elements and check functions.
 const documentPlaceholders = new WeakMap<Document, Map<Element, PlaceholderConfig>>();
 
+let hasDisplayedWarning = false;
```
mlewand, line 23: The name could be a bit more specific in a sense to tell what kind of warning is it.

packages/ckeditor5-engine/src/view/placeholder.ts
```diff
@@ -37,10 +39,11 @@ const documentPlaceholders = new WeakMap<Document, Map<Element, PlaceholderConfi
  * editable root elements.
  * @param options.keepOnFocus If set `true`, the placeholder stay visible when the host element is focused.
  */
-export function enablePlaceholder( { view, element, isDirectHost = true, keepOnFocus = false }: {
+export function enablePlaceholder( { view, element, text, isDirectHost = true, keepOnFocus = false }: {
```
mlewand, line 42: You missed documenting it in API docs. If it's in the API, it should be listed as a param - however, with clear information that it is deprecated.

packages/ckeditor5-engine/src/view/placeholder.ts
```diff
@@ -299,6 +308,19 @@ function getChildPlaceholderHostSubstitute( parent: Element ): Element | null {
 	return null;
 }
 
+/**
+ * Displays a deprecation warning message in the console, but only once per page load.
+ */
+function showDeprecationWarning() {
+	if ( !hasDisplayedWarning ) {
+		console.warn( 'Deprecation Warning: The "text" argument in the "enableProperty" function will be deprecated ' +
```
mlewand, line 316: Semantic: it is deprecated already (will become with your change). What will happen in the future is the removal of the function.  
  
To avoid saying exactly when we'll remove it, let's just say that it will be removed in a future release (without pointing out precisely when). I assume that it's going to be few months (iterations) but noting his written in stone.

packages/ckeditor5-engine/src/view/placeholder.ts
```diff
@@ -67,6 +70,12 @@ export function enablePlaceholder( { view, element, isDirectHost = true, keepOnF
 
 	if ( element.placeholder ) {
 		setPlaceholder( element.placeholder );
+	} else if ( text ) {
+		setPlaceholder( text );
+	}
+
+	if ( text ) {
```
mlewand, line 77: You added a second if statement for the exact same condition that is preceding this condition (`} else if ( text ) {`).

This should be combined into one to be logically consistent.

packages/ckeditor5-engine/src/view/placeholder.ts
```diff
@@ -299,6 +308,19 @@ function getChildPlaceholderHostSubstitute( parent: Element ): Element | null {
 	return null;
 }
 
+/**
+ * Displays a deprecation warning message in the console, but only once per page load.
+ */
+function showDeprecationWarning() {
```
mlewand, line 314:
```suggestion
function showPlaceholderTextDeprecationWarning() {
```

Similar as for the `hasDisplayedWarning` const name.

packages/ckeditor5-engine/tests/view/placeholder.js
```diff
@@ -632,12 +633,62 @@ describe( 'placeholder', () => {
...
+			enablePlaceholder( {
+				view,
+				element: viewRoot,
+				text: 'placeholder'
+			} );
+
+			expect( viewRoot.getAttribute( 'data-placeholder' ) ).to.equal( 'placeholder' );
+		} );
+
+		it( 'should use element\'s placeholder as more prior value', () => {
```
mlewand, line 680:
```suggestion
        it( 'should prefer element\'s placeholder value over text parameter', () => {
```

mlewand: Good test :+1: we need to assert that new style is indeed preferred over deprecated format.

packages/ckeditor5-engine/tests/view/placeholder.js
```diff
@@ -632,12 +633,62 @@ describe( 'placeholder', () => {
...
 				element: viewRoot,
-				text: 'foo bar baz',
 				isDirectHost: false
 			} );
 			viewRoot.placeholder = 'new placeholder';
 
 			expect( viewRoot.getChild( 0 ).getAttribute( 'data-placeholder' ) ).to.equal( 'new placeholder' );
 		} );
+
+		it( 'should through warning once if "text" is used as argument', () => {
```
mlewand, line 643: What did you mean by "should through warning once"?

mlewand (changes requested):
Couple of adjustments ad we're good to go :+1: 

Be sure to check tests with `--production` CLI flag for testing as it _might_ forbid using `window.console`.

mlewand:
And my assumption was correct, the test are failing on CI: [https://app.travis-ci.com/github/ckeditor/ckeditor5/jobs/607359829](https://app.travis-ci.com/github/ckeditor/ckeditor5/jobs/607359829)

packages/ckeditor5-engine/tests/view/placeholder.js
illia-stv: Warning should be throughn only once in dev console

packages/ckeditor5-engine/src/view/placeholder.ts
illia-stv: But what if someone has simultaneously an element placeholder's value and passes text as an argument, so the element's value will have higher priority and will be applied but the text is also passed as the argument so we should inform about it, don't we?

illia-stv: But what if someone has simultaneously an element placeholder's value and passes text as an argument, so the element's value will have higher priority and will be applied but the text is also passed as the argument so we should inform about it, don't we?

illia-stv:
@mlewand There is one additional suggestion by @Dumluregn to perform displaying warning by `logWarning`. 

image

Is worth to mention when you open your dev console you will see only error name with link to the documentation.

image

`logWarning` creates a section in error code list.

image

What do you think?

packages/ckeditor5-engine/src/view/placeholder.ts
mlewand: Ok sounds fair, a rare case - but indeed we would be on a safe side with this approach 👍

Issue #14801, opened 2023-08-16
Missing name `global` when scrolling ckeditor - header does not stick


FreexD:
## 📝 Provide detailed reproduction steps (if any)

1. Trrying to scroll ckedtor-5 on version 39.0.1.

### ✔️ Expected result

The header sticks properly and no error is in the console.

### ❌ Actual result

The header does not stick and it errors in the console - missing name `global`.

### ❓ Possible solution

Missin global import  `import global from './global';` in https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/dom/getelementsintersectionrect.ts .

## 📃 Other details

* Browser: Chrome
* OS: MacOS
* First affected CKEditor version: 39.0.1

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Witoso:
Thanks for reporting! cc @filipsobol

Witoso:
We noticed it doesn't work in the Vite setup (ckeditor/vite-plugin-ckeditor5#17). @FreexD could you share the details about your setup?

FreexD:
@Witoso I was testing on a Vite setup as well.

filipsobol:
This issue will be fixed by PR #14808.

filipsobol:
The fix for this issue has been merged (PR #14808) and will be available in the next release.

PR #14808, opened 2023-08-17
fix: load only specified typings for TypeScript

### LINKED ISSUE #14173: Disallow the use of `global` in the JS code intended for the browser
We've just stumbled upon an issue with the accidental use of `global` in our code: dc25921

Which resulted in an error only when the editor was built with Vite (and probably other bundlers that don't try to polyfill the CJS env):

image

The issue in our code was hard to spot during the review because as you can see in dc25921 we actually have a module named `global` too and when you copy&paste code between files you may copy the usage of it but miss the import.

---

filipsobol:

Fix: Don't rely on `global` object available only in Node. Fixes https://github.com/ckeditor/vite-plugin-ckeditor5/issues/17 and #14801.

Internal: Load only specified typings for TypeScript. Closes #14173.

### DIFF
--- a/packages/ckeditor5-engine/src/view/observer/mutationobserver.ts
+++ b/packages/ckeditor5-engine/src/view/observer/mutationobserver.ts
@@ -55,7 +55,7 @@ export default class MutationObserver extends Observer {
 	/**
 	 * Native mutation observer.
 	 */
-	private _mutationObserver: InstanceType<typeof global.MutationObserver>;
+	private _mutationObserver: InstanceType<typeof window.MutationObserver>;
 
 	/**
 	 * @inheritDoc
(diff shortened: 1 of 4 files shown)


arkflpc (approved):
LGTM. However, I'd seek approval from @ckeditor/ckeditor-5-devops  team as well.

pomek (approved):
LGTM.

I like the proposed empty array as it still loads needed types, but a developer does not have to remember to add new types when using some external library.

pomek:
I believe the proposed message should be _Internal_, or if it really matters for integrators, _Other_.

filipsobol:
> I believe the proposed message should be _Internal_, or if it really matters for integrators, _Other_.

In the `packages/ckeditor5-utils/src/dom/getelementsintersectionrect.ts` file we're fixing code error, so I believe that `Fix` is appropriate.

PR #14832, opened 2023-08-22
Switch to CircleCI

pomek:

Internal: Switch the CI provider from TravisCI to CircleCI. 

---

Closes cksource/ckeditor5-internal#3559.

---

### Additional information


pomek:
Example builds:

* Fails when CC is not 100%: https://app.circleci.com/pipelines/github/ckeditor/ckeditor5/174/workflows/229f3cd7-75ad-44b6-a25b-f2069022678b/jobs/188

pomek:
Highlights:

* The code coverage job is executed when a PR targets the `#master` branch.
	* If a PR comes from the community, the job starts, but it does not send anything. It exists as fast as we know the source of the PR.
	* The similar behavior is applied to the notifier and triggers the Uber CI processes.
* Tests (+ checking the CC) are split into two phases: framework and features. ([example when fails](https://app.circleci.com/pipelines/github/ckeditor/ckeditor5/174/workflows/229f3cd7-75ad-44b6-a25b-f2069022678b/jobs/188))
* When a PR updates the documentation only, tests and checking the release steps are skipped.
* To avoid executing a JavaScript script via `node -e require(...);` I converted all scripts to self-executable.

scripts/ci/should-run-short-flow.js
```diff
@@ -26,56 +28,53 @@ const filePatterns = [
...
-	if ( !changedFilesPaths ) {
 		changedFilesPaths = execSync( `git diff --name-only ${ diffTargets }`, { cwd } ).toString();
 	}
 
-	return doAllFilesMatchPattern(
-		changedFilesPaths.trim().split( '\n' )
+	const files = changedFilesPaths.trim().split( '\n' );
+
+	return process.exit(
+		doAllFilesMatchPattern( files ) ? 0 : 1
```
martnpaneq, line 75: I know that it is not in the scope of the PR, but while reading the code I noticed that it was hard to understand what `filePatterns` meant. What do you think about renaming:
1. `filePatterns` -> `shortFlowFilePatterns`,
2. `doAllFilesMatchPattern` -> `allFilesMatchShortFlow`
Or some other names that have similar meaning.

pomek: `s/doAllFilesMatchPattern/areAllFilesMatchShortFlow` ?

scripts/ci/is-community-pr.js
```diff
@@ -0,0 +1,45 @@
...
+const minimist = require( 'minimist' );
+
+const {
+	// The number of the associated GitHub or Bitbucket pull request. Only available on forked PRs.
+	CIRCLE_PR_NUMBER
+} = process.env;
+
+module.exports = main();
+
+function main() {
```
martnpaneq, line 21: I feel like this function is a bit confusing.
What about just exporting a const: `IS_COMMUNITY_PR`?
Then in the node command execution we could just use: 
image

pomek: It's a compromise between having a single file that can be executed as CJS (`require()`) and a node script.

I am not glad of it, but it resolves issues with complicated bash ifs if we want to break the build.

scripts/ci/should-run-short-flow.js
martnpaneq: `doAllFilesMatchShortFlow` sound good.

pomek: Addressed in 2ece7f8b912888ba69d22d964190069fa503d549.

scripts/ci/is-community-pr.js
martnpaneq: What do you think about splitting it into two files? The const could live in some `constants` directory/file. And the script could just use the const.

pomek: That was the goal: having single file for this micro util :D

I will not block this change and I will update the scripts soon.

psmyrek (approved):
LGTM 🎉

PR #14843, opened 2023-08-22
Fix RangeError when handling big tables

### LINKED ISSUE #14785: Loading very long table created in CKEditor 4 freezes the editor

1.  Use the following content as the initial data - [https://gist.github.com/Mgsy/7a8cb809e34c1595d5bec788804091e5](https://gist.github.com/Mgsy/7a8cb809e34c1595d5bec788804091e5) (it's a table with ~3000 rows) or copy/paste it to the editor.

### ✔️ Expected result

_The table loads._

### ❌ Actual result

_The browser freezes. It might also throw errors after some time:_

## 📃 Other details

The same table loads almost immediately in CKEditor 4.

---

filipsobol:

Fix (table): Fix crashing when handling tables with over 2500 rows. Closes #14785.

Other (table): Improve performance when handling large tables. See #14785.

---

**See the comments below for more details.**


filipsobol:
## Fix crashing when handling tables with over 2500 rows

The exception reported in #14785 was thrown when `TableWalker`, created with the `row` argument, iterated over a large table ([code](https://github.com/ckeditor/ckeditor5/blob/6bbde59ef34a9aa2ce4613f50dff74f24dec7c8b/packages/ckeditor5-table/src/converters/downcast.ts#L102)).

To downcast each cell, `TableWalker` traversed the table from the beginning to the last cell in `row`. This means that if the table has 2000 rows and 3 columns, it walked through all 6000 cells to downcast a cell in a last row. Now repeat the same process for all the other 5999 cells 😱.

When the table was large enough, browsers reached their limits and threw the `RangeError: Maximum call stack size exceeded` error.

This is best illustrated by this flamegraph (downcasting table with 2000 rows and 3 columns):

Flamegraph taking about 5 screen heights

### Solution

The approach described above was used to get a list of all cells that span multiple rows or columns, since they could affect the `row`.

However, this process could be optimized. `TableWalker` will now check the number of columns in the table and see if it matches the number of cells in the `row`. If it does, it sets it as the current row and skips all rows before it. Otherwise, it checks the rows before it until it finds one that matches this criteria. See the comment above the `_jumpToNonSpannedRowClosestToStartRow` method for an example.

Flamegraph for this approach (downcasting table with 2000 rows and 3 columns):

Much smaller flamegraph taking less than 1 screen height

filipsobol:
## Improve performance when handling large tables

As described in the comment above, the browser has to do much less work to downcast a large table. According to my tests, adding a table with 2000 rows and 3 columns took on average:
* **about 5526ms before the changes**,
* **about 4123ms after the changes**.

The above numbers were taken from the latest Chrome and only include Scripting (ignoring Rendering, Painting, etc.).

packages/ckeditor5-table/src/tablewalker.ts
```diff
@@ -423,6 +432,67 @@ export default class TableWalker implements IterableIterator<TableSlot> {
...
+	 * 4.) Check if the 2nd row contains 3 cells. It does, so set the current row to that row.
+	 *
+	 * Setting the current row this way is necessary to let the `next()`  method loop over the cells
+	 * spanning multiple rows or columns and update the `this._spannedCells` property.
+	 */
+	private _jumpToNonSpannedRowClosestToStartRow(): void {
+		const firstRow = this._table.getChild( 0 ) as Element;
+		const firstRowLength = this._getRowLenght( firstRow );
+
+		for ( let i = this._startRow!; i >= 0; i-- ) {
```
arkflpc, line 475: Is it possible that the loop condition will stop the loop? Or it will always be terminated by `break`?
It should be, because the loop goes from some element in the direction to the first one and the first one has length equal to `firstRowLength`.
If so, what is the purpose of `i >= 0`?

filipsobol: You are right, iterating over element `0` would compare it to itself. I changed it to `i > 0` and fixed the typo in method name.

arkflpc (approved):

PR #14893, opened 2023-08-30
[TODO] Improved and added new UI components

oleq:

* https://github.com/ckeditor/ckeditor5/tree/list-view-groups-poc
  * Feature (ui): Implemented the `ListItemGroupView` component. Allowed grouping `ListView` items.
* https://github.com/ckeditor/ckeditor5/pull/14904
  * Feature (ui): Implementing the `TextareaView` component with the same base as the `InputTextView` component.
  * Feature (ui): Implementing the `SpinnerView` component.
* https://github.com/cksource/ckeditor5-commercial/issues/5478 
  * Feature (ui, theme-lark): Created the `SearchView` UI component. Created the `AutocompleteView` component. Introduced a `HighlightedTextView` component for better search results presentation. Made the `ButtonView` label logic open for extension.
 
---

### Additional information


packages/ckeditor5-ui/src/autocomplete/autocompleteview.ts
```diff
@@ -0,0 +1,232 @@
...
+
+		const toPx = toUnit( 'px' );
+
+		this.extendTemplate( {
+			attributes: {
+				class: [ 'ck-autocomplete' ]
+			}
+		} );
+
+		const resultsView = this.resultsView as AutocompleteResultsView;
```
arkflpc, line 47: Add `declare public resultsView: AutocompleteResultsView;` to this class and get rid of those casts. The reason is to make it explicit that `this.resultsView` in this class always conforms to `AutocompleteResultsView` instead of assuring the compiler with the implicit knowledge at every use site with casts.

packages/ckeditor5-ui/src/autocomplete/autocompleteview.ts
```diff
@@ -0,0 +1,232 @@
...
+		resultsView.on( 'change:isVisible', () => {
+			this._updateResultsViewWidthAndPosition();
+		} );
+	}
+
+	/**
+	 * Updates the position of the results view on demand.
+	 */
+	private _updateResultsViewWidthAndPosition() {
+		const resultsView = ( this.resultsView as AutocompleteResultsView );
```
arkflpc, line 132: See my comment to line 47.

packages/ckeditor5-ui/src/autocomplete/autocompleteview.ts
```diff
@@ -0,0 +1,232 @@
...
+
+		const optimalResultsPosition = AutocompleteView._getOptimalPosition( {
+			element: this.resultsView.element!,
+			target: this.queryView.element!,
+			fitInViewport: true,
+			positions: AutocompleteView.defaultResultsPositions
+		} );
+
+		// _getOptimalPosition will return null if there is no optimal position found (e.g. target is off the viewport).
+		resultsView._position = optimalResultsPosition ? optimalResultsPosition.name as string : 's';
```
arkflpc, line 148: I know that `AutocompleteResultsView._position` is never `undefined` in your code. But if it would be typed as such, you wouldn't need to cast to `string` (nor use `!`) here and everything would compile anyway.

packages/ckeditor5-ui/src/autocomplete/autocompleteview.ts
```diff
@@ -0,0 +1,232 @@
...
+	 * * `n` - above the search field.
+	 */
+	public static defaultResultsPositions: Array<PositioningFunction> = [
+		( fieldRect => {
+			return {
+				top: fieldRect.bottom,
+				left: fieldRect.left,
+				name: 's'
+			};
+		} ) as PositioningFunction,
```
arkflpc, line 174:
```suggestion
		} ) satisfies PositioningFunction,
```
Supported since TS 4.9. It checks the function conforms to `PositioningFunction` (and makes contextual typing work), but doesn't _force_ it to `PositioningFunction`.

packages/ckeditor5-ui/src/autocomplete/autocompleteview.ts
```diff
@@ -0,0 +1,232 @@
...
+		( ( fieldRect, resultsRect ) => {
+			return {
+				top: fieldRect.top - resultsRect.height,
+				left: fieldRect.left,
+				name: 'n'
+			};
+		} ) as PositioningFunction
```
arkflpc, line 181:
```suggestion
		} ) satisfies PositioningFunction
```

packages/ckeditor5-ui/src/highlightedtext/highlightedtextview.ts
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * A class representing a view that displays a text which subset can be highlighted using the
+ * {@link #highlightText} method.
+ */
+export default class HighlightedTextView extends View {
+	/**
+	 * The text that can be highlighted using the {@link #highlightText} method.
+	 *
+	 * **Note:** When this property changes, the previous highlighting is removed.
+	 */
```
arkflpc, line 24:
```suggestion
     * @observable
	 */
```

packages/ckeditor5-ui/src/highlightedtext/highlightedtextview.ts
```diff
@@ -0,0 +1,126 @@
...
+	 * Updates element's innerHTML with the passed content.
+	 *
+	 * @param newText
+	 */
+	private _updateInnerHTML( newInnerHTML: string | undefined ) {
+		this.element!.innerHTML = newInnerHTML || '';
+	}
+}
+
+// Replaces RegExp occurrences with <mark> tags in a text.
```
arkflpc, line 73: Use `/** ... */` instead of `//`. This function is not exported, so it will not land in API docs. But it would enable hinting in IDE.

packages/ckeditor5-ui/src/highlightedtext/highlightedtextview.ts
```diff
@@ -0,0 +1,126 @@
...
+//
+// @param text A text to get marked.
+// @param [regExp] An optional RegExp. If not passed, this is a pass-through function.
```
arkflpc, line 76: Not sure if `[]` is supported by TypeDoc. If the argument is optional, mark it with `?` in the argument list (probably in addition to being `| null`).

packages/ckeditor5-ui/src/highlightedtext/highlightedtextview.ts
```diff
@@ -0,0 +1,126 @@
...
+		return escape( text );
+	}
+
+	const textParts: Array<{ text: string; isMatch: boolean }> = [];
+	let lastMatchEnd = 0;
+	let matchInfo;
+
+	// Iterate over all matches and create an array of text parts. The idea is to mark which parts are query matches
+	// so that later on they can be highlighted.
+	while ( ( matchInfo = regExp.exec( text ) ) !== null ) {
```
arkflpc, line 89: Just learned from @Reinmar yesterday: assignment in conditions is not readable. Consider refactoring this loop.

packages/ckeditor5-ui/src/search/searchinfoview.ts
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * @module ui/search/searchinfoview
+*/
+
+import View from '../view';
+
+/**
+ * A view displaying an information text related to different states of {@link module:ui/search/text/searchtextview~SearchTextView}.
+ *
+ * @private
```
arkflpc, line 15: Did you mean `internal`? I don't really know the impact of having `prrivate` on API docs. https://typedoc.org/tags/private/ describes it only in the meaning of `private` in TypeScript (i.e. class memebers).

packages/ckeditor5-ui/src/search/text/searchtextqueryview.ts
```diff
@@ -0,0 +1,151 @@
...
+import { createLabeledInputText } from '../../labeledfield/utils';
+import type InputBase from '../../input/inputbase';
+import type { Locale } from '@ckeditor/ckeditor5-utils';
+
+import { icons } from '@ckeditor/ckeditor5-core';
+
+/**
+ * A search input field for the {@link module:ui/search/text/searchtextview~SearchTextView} component.
+ *
+ * @private
```
arkflpc, line 22: `internal`? See my comment for `searchinfoview.ts`.

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,388 @@
...
+
+	/**
+	 * The view that allows the user to enter the search query.
+	 */
+	public queryView: SearchTextQueryView<TQueryFieldView>;
+
+	/**
+	 * Controls whether the component is in read-only mode.
+	 *
+	 * @default true
```
arkflpc, line 83:
```suggestion
	 * @default true
	 * @observable
```

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,388 @@
...
+	 */
+	declare public isEnabled: boolean;
+
+	/**
+	 * The number of results found for the current search query. Updated upon the {@link #search} event.
+	 *
+	 * @default 0
```
arkflpc, line 90:
```suggestion
	 * @default 0
	 * @observable
```

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,388 @@
...
+	 */
+	declare public resultsCount: number;
+
+	/**
+	 * The number of the items that can be searched in the {@link #filteredView}. Updated upon the {@link #search} event.
+	 *
+	 * @default 0
```
arkflpc, line 97:
```suggestion
	 * @default 0
	 * @observable
```

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,388 @@
...
+			actions: {
+				// Navigate form fields backwards using the Shift + Tab keystroke.
+				focusPrevious: 'shift + tab',
+
+				// Navigate form fields forwards using the Tab key.
+				focusNext: 'tab'
+			}
+		} );
+
+		this.on( 'search', ( evt, { resultsCount, totalItemsCount } ) => {
```
arkflpc, line 180:
```suggestion
		this.on<SearchTextViewSearchEvent>( 'search', ( evt, { resultsCount, totalItemsCount } ) => {
```
Currently, `resultsCount` and `totalItemsCount` are `any`.

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,388 @@
...
+
+		return queryView;
+	}
+
+	/**
+	 * Initializes the default {@link #infoView} behavior with default text labels when no custom info view
+	 * was specified in the view config.
+	 */
+	private _enableDefaultInfoViewBehavior(): void {
+		const t = this.locale!.t;
```
arkflpc, line 281: If you require `locale` to be non-optional in the constructor, you could ensure `this.locale` is non-optional as well using `declare public locale: Locale` in the class.

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	/**
+	 * Creates a search field view based on configured creator..
+	 *
+	 * @param locale The localization services instance.
+	 * @param label The label of the search field.
+	 */
+	private _createSearchTextQueryView(): SearchTextQueryView<TQueryFieldView> {
+		const queryView = new SearchTextQueryView<TQueryFieldView>( this.locale!, this._config.queryView );
```
arkflpc, line 264: See my comment to line 281.

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,388 @@
...
+	 * Initializes the default {@link #infoView} behavior with default text labels when no custom info view
+	 * was specified in the view config.
+	 */
+	private _enableDefaultInfoViewBehavior(): void {
+		const t = this.locale!.t;
+		const infoView = this.infoView as SearchInfoView;
+
+		this.on<SearchTextViewSearchEvent>( 'search', ( evt, data ) => {
+			if ( !data.resultsCount ) {
+				const defaultTextConfig = this._config.infoView?.text;
```
arkflpc, line 286: I hope this will work. We had problems before. See [ADR 18](https://github.com/cksource/ckeditor5-commercial/blob/master/adr/0018-avoid-using-certain-operators.md).

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,388 @@
...
+export type SearchTextViewDefaultInfoText = string | ( ( query: string, resultsCount: number, totalItemsCount: number ) => string );
+
+/**
+ * An event fired when the search query changes fired by {@link module:ui/search/text/searchtextview~SearchTextView#search}.
+ *
+ * @eventName ~SearchTextView#search
+ */
+export type SearchTextViewSearchEvent = {
+	name: 'search';
+	args: [ {
```
arkflpc, line 383: I'm pretty sure, API Docs for this will not be useful. Consider extracting this object type as a separate (and documented) `interface`. If you think this is too much, at least mention the fields in the event description (the doclet above)).

arkflpc (changes requested):

packages/ckeditor5-ui/src/search/searchinfoview.ts
arkflpc: On the other hand, `internal` is supported by Umberto and the class can be omitted from `.d.ts` files (although, I'm not sure if it's enable in this repo's scripts).

packages/ckeditor5-ui/src/autocomplete/autocompleteview.ts
DawidKossowski: The current version of Typescript is `"^4.8.4"`. My IDE doesn't work correctly with the `satisfies` and I'm not sure if it will compile properly. For now, I will stay with `as`.

arkflpc (approved):

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,414 @@
...
+	 */
+	declare public resultsCount: number;
+
+	/**
+	 * The number of the items that can be searched in the {@link #filteredView}. Updated upon the {@link #search} event.
+	 *
+	 * @default 0
+	 * @observable
+	 */
+	declare public totalItemsCount: number;
```
scofalik, line 102: Shouldn't `totalItemsCount` be refreshed when the items in `filteredView` change instead of on search? Shouldn't we display a proper info if `filteredView` becomes empty separately from processing search requests?

If you wanted to exclude that from the scope to deliver the component faster, let's create a followup issue for this.

packages/ckeditor5-ui/src/search/text/searchtextview.ts
```diff
@@ -0,0 +1,414 @@
...
+		queryView.bind( 'isEnabled' ).to( this );
+
+		return queryView;
+	}
+
+	/**
+	 * Initializes the default {@link #infoView} behavior with default text labels when no custom info view
+	 * was specified in the view config.
+	 */
+	private _enableDefaultInfoViewBehavior(): void {
```
scofalik, line 282: As mentioned in comment below, I think that the component should not be searchable if it is empty. "Empty" and "no results" should be handled separately. But we can move this to a followup issue.

scofalik (changes requested):

packages/ckeditor5-ui/src/textarea/textareaview.ts
```diff
@@ -0,0 +1,211 @@
...
+		);
+
+		if ( shouldScroll ) {
+			viewElement.scrollTop = viewElement.scrollHeight;
+		}
+
+		singleLineContentClone.remove();
+		fullTextValueClone.remove();
+
+		this.fire<TextareaViewAutoGrowEvent>( 'autoGrow' );
```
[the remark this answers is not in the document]
oleq: Yes, it's consumed in another project. This is a bit sketchy, though because the aim of this event was to know when anything happens to the content of textarea, either by user input or by declarative input change. The auto-grow logic acts on both events (even if the height didn't change) so it was convenient to fire an event here.

I think we could move this `fire()` to the right place (somewhere in the constructor) but I'm at a loss when it comes to naming this kind of event. 

Context: we use this event to determine whether the scrollbar showed up or disappeared. This is not necessarily associated with the change of height (e.g. pressing enter when a number of lines === maxRows does not change height but introduces a scrollbar).

packages/ckeditor5-ui/src/textarea/textareaview.ts
```diff
@@ -0,0 +1,211 @@
...
+	 */
+	private _validateMinMaxRows() {
+		if ( this.minRows > this.maxRows ) {
+			/**
+			 * The minimum number of rows is greater than the maximum number of rows.
+			 *
+			 * @error ui-textarea-view-min-rows-greater-than-max-rows
+			 * @param textareaView The misconfigured textarea view instance.
+			 */
+			console.warn( 'ui-textarea-view-min-rows-greater-than-max-rows', { textareaView: this } );
```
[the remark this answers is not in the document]
oleq: TBH, I'd go the other way around and throw instead of patching the misconfiguration.

packages/ckeditor5-ui/src/labeledfield/utils.ts
```diff
@@ -120,7 +113,56 @@ export function createLabeledInputNumber(
...
+ * ```
+ *
+ * @param labeledFieldView The instance of the labeled field view.
+ * @param viewUid An UID string that allows DOM logical connection between the
+ * {@link module:ui/labeledfield/labeledfieldview~LabeledFieldView#labelView labeled view's label} and the textarea.
+ * @param statusUid An UID string that allows DOM logical connection between the
+ * {@link module:ui/labeledfield/labeledfieldview~LabeledFieldView#statusView labeled view's status} and the textarea.
+ * @returns The textarea view instance.
+ */
+const createLabeledTextarea: LabeledFieldViewCreator<TextareaView> = ( labeledFieldView, viewUid, statusUid ) => {
```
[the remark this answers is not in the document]
oleq: Agreed. This is a mess.

packages/ckeditor5-ui/src/list/listitemgroupview.ts
```diff
@@ -0,0 +1,120 @@
...
+			},
+
+			children: this.children
+		} );
+	}
+
+	/**
+	 * Creates a label for the group.
+	 */
+	private _createLabel( groupLabelId: string ): View {
```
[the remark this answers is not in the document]
oleq: I think a labelless group contradicts the idea of a grouping. It does not make sense visually. It does not make sense accessibility-wise either. 

If it wasn't for the fact that we don't provide values via constructors but use observables (I mentioned this in a previous comment that we should really figure it out one day), I'd either block the possibility of passing an empty string or even throw/warn if that ever happened.

packages/ckeditor5-ui/src/search/searchinfoview.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @internal
+ */
+export default class SearchInfoView extends View {
+	/**
+	 * Controls whether the view is visible.
+	 *
+	 * @observable
+	 * @default false
+	 */
+	declare public isVisible: boolean;
```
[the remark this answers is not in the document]
oleq: We figured out the same with @DawidKossowski the other day 🙂 The nasty thing is though, that base `View` class does not provide its own template. And if it suddenly did (with this handy `isVisible` binding), every child view would have to use `View#extendTemplate()` instead of `View#setTempate()` which would be a maaaaaaaaaaaasive BC. Plus, on top of that, `View#extendTempate()` does not support adding new children, so that would be useless anyway.

So long story short, with the current architecture, that would have to be... a utility that creates an observable property and then calls `extendTemplate()` itself with the correct binding? Something akin to `CssTransitionDisablerMixin`. Or maybe some very very late `View#render` listener that `extendsTextmplate()` at the last moment?

packages/ckeditor5-ui/src/search/text/searchtextview.ts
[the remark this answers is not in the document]
oleq: We did this on purpose. We know there will be an abstract class called `SearchView` in the future. We didn't have time to make it so, though. 

For one thing, that `SearchView` class will make some TS a lot easier for sub-classes such as `SearchTextView` and `AutocompleteView` (it's a long story) and shorten the sub-classes to the case-specific code only.

Another thing is that it's very likely that more `SearchView` derivatives will come: for instance, a search-by-tags feature, that uses a toolbar (with tag buttons) to filter an arbitrary filterable view.

Organizing the code like this and keeping this naming convention now will spare us a lot of BCs in the future. We'll be able to extract `SearchView` from `SearchTextView` silently and bring other search types without any fuss.

packages/ckeditor5-ui/src/search/text/searchtextview.ts
oleq: It makes sense but it would mean `FilteredView` will have to be broader than that. For instance, it will have to expose some (observable) properties such as `totalItemsCount` itself. I don't think we have time to address that now.

packages/ckeditor5-ui/src/search/text/searchtextview.ts
oleq: Let's discuss this in a follow-up please.

packages/ckeditor5-ui/src/search/searchinfoview.ts
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+/**
+ * @module ui/search/searchinfoview
+*/
+
+import View from '../view';
+
+/**
+ * A view displaying an information text related to different states of {@link module:ui/search/text/searchtextview~SearchTextView}.
```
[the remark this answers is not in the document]
oleq: `SearchInfoView` is a really dummy view. It displays two labels according to two observable strings and can be hidden. It does not understand any `SearchTextView` logic nor should pretend it does. I think this kind of documentation does not belong here.

packages/ckeditor5-ui/src/search/text/searchtextview.ts
scofalik: Would you please add an issue for that with very short description? Then we can resolve these comments.

packages/ckeditor5-ui/src/list/listitemgroupview.ts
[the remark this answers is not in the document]
scofalik: Visually it makes a bit of sense because the groups can be separated by, well, some visual element, like a border. Do you see maybe a use for `ListItemGroupView` in UIs like nested dropdowns or application menu? Or will it be just an array of list items anyway?

I was more afraid about a11y issues.

But maybe if you want to have a group, you should always have a name for it (a11y, something else?) and then if you do not want to display it, you can have a proper template or even use CSS.

I was coming just from the visual POV -- in case if someone wants to have the options seprated but not to include a label. Well, similar thing as in toolbar, you also create groups but don't name them.

packages/ckeditor5-ui/src/textarea/textareaview.ts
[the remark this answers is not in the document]
scofalik: I believe that the only time we will get here is through a typo, I don't believe these values will be taken from another place or dynamically provided. So, throwing works for me. The goal is to not have bugs because `minRows > maxRows` and it glitches somewhere down the road.

packages/ckeditor5-ui/src/search/searchinfoview.ts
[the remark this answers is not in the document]
scofalik: Ah, I didn't know about the need to change `setTemplate()` to `extendTemplate()`. That's a bummer.

> Or maybe some very very late `View#render` listener that `extendsTextmplate()` at the last moment?

I think we should keep it in mind during refactoring UI. We will see what direction we will take and how many BCs we will introduce. Maybe it will be easier to make a decision then.

packages/ckeditor5-ui/src/search/searchinfoview.ts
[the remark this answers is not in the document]
scofalik: I guess the thing here is whether we approach the view semantically or functionally. In other words, whether we have "info view" -- a view to display information or "no results view" -- a view displayed when there are no results.

I think the first is more useful, especially since we display "no results" and "empty list" messages.

So, I agree with you.

packages/ckeditor5-ui/src/textarea/textareaview.ts
```diff
@@ -0,0 +1,211 @@
...
+	 * @observable
+	 * @default 5
+	 */
+	declare public maxRows: number;
+
+	/**
+	 * Specifies the value of HTML attribute that indicates whether the user can resize the element.
+	 *
+	 * @observable
+	 * @default 'both'
```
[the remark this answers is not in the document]
oleq: Let's go with `none` for none and discuss auto-grow later in a follow-up.

packages/ckeditor5-ui/src/textarea/textareaview.ts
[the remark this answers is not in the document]
oleq: Decision: Let's name the even update and fire it in related listeners that may change the geometry of the component (`input`, `change:value` + `reset()`)

packages/ckeditor5-ui/src/textarea/textareaview.ts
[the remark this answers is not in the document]
oleq: Decision: Let's throw.

packages/ckeditor5-ui/src/list/listitemgroupview.ts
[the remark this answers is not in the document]
oleq: There's a `ListItemSeparator` component in out codebase for that purpose.

packages/ckeditor5-ui/src/textarea/textareaview.ts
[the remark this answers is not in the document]
oleq: [70c8e55](https://github.com/ckeditor/ckeditor5/pull/14893/commits/70c8e554f7b45c57805bb91fa1973268956059cd)

packages/ckeditor5-ui/src/search/text/searchtextview.ts
oleq: https://github.com/ckeditor/ckeditor5/issues/15059

packages/ckeditor5-ui/src/search/text/searchtextview.ts
oleq: https://github.com/ckeditor/ckeditor5/issues/15059

packages/ckeditor5-ui/src/textarea/textareaview.ts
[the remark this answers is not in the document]
oleq: https://github.com/ckeditor/ckeditor5/issues/15061

scofalik (approved):

PR #14904, opened 2023-08-31
Implementing textarea and spinner component

DawidKossowski:

Feature (ui): Implementing the Textarea component with the same base as the Input component.
Feature (ui): Implementing the Spinner component.


packages/ckeditor5-ui/src/textarea/textareaview.ts
```diff
@@ -0,0 +1,65 @@
...
+				'aria-invalid': bind.if( 'hasError', true ),
+				'aria-describedby': bind.to( 'ariaDescribedById' )
+			},
+			on: {
+				input: bind.to( ( ...args ) => {
+					this.fire( 'input', ...args );
+					this._updateIsEmpty();
+				} ),
+				change: bind.to( this._updateIsEmpty.bind( this ) )
+			}
```
oleq, line 51: I think you can move most of these to `InputBase` and use `this.extendTemplate( { ... } )` here and add sub-class properties and behaviors only.

packages/ckeditor5-ui/src/textarea/textareaview.ts
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+/**
+ * Fired when the user types in the textarea. Corresponds to the native
+ * DOM `input` event.
+ *
+ * @eventName ~TextareaView#input
+ */
+export type TextareaInputEvent = {
```
oleq, line 62: Do you think we need two input event types for input text and textarea? They look the same.

packages/ckeditor5-ui/src/spinner/spinnerview.ts
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @observable
+	 * @default false
+	 */
+	declare public isVisible: boolean;
+
+	/**
+	 * @inheritDoc
+	 */
+	constructor( locale?: Locale ) {
```
oleq, line 31: I'm not sure if `Locale` will be of any use for this view ever. There's nothing to translate here and it's unlikely to get any children that will need translations.

packages/ckeditor5-ui/theme/components/spinner/spinner.css
```diff
@@ -0,0 +1,37 @@
...
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+:root {
+	--ck-toolbar-spinner-size: 18px;
+}
+
+.ck.ck-spinner-container {
+	display: block;
+	position: relative;
```
oleq, line 12: I know this is weird our system works like this since there's only one theme and we never went on from that place but can you leave the structure (display, position) here and move the presentation (border, width, margin) to ckeditor5-theme-lark? Just in case one day we change our minds.

oleq (changes requested):
I'd also love how textarea view works out with `LabeledFieldView` (in a manual latest). Without that integration, the component will not make any sense, it must be labeled at some point.

packages/ckeditor5-ui/theme/components/spinner/spinner.css
oleq: And no... there's no clear guide on this 😃

PR #14917, opened 2023-09-01
Removed the short-flow on CI

### LINKED ISSUE #14916: CI: Short-flow detection does not work in PRs

In this PR – https://github.com/ckeditor/ckeditor5/pull/14914 – we changed docs and package sources.

Unfortunately, the last commit (https://github.com/ckeditor/ckeditor5/pull/14914/commits/955049354ed33b078ddb68bf6d5f07705427271b) modified only docs files. Hence, the short-flow was active. In PRs, we should check all updated files (all commits) instead.

I think, something is wrong here:

https://github.com/ckeditor/ckeditor5/blob/2ece7f8b912888ba69d22d964190069fa503d549/scripts/ci/should-run-short-flow.js#L60-L70

pomek:

Internal: Removed the short flow on CI. Closes #14916.


pomek:
I am unsure about the change, as the problem occurs due to the flow of how developers work. Before creating a pull request, a developer needs to push a branch. And CircleCI builds the latest commit on the branch. If it updated docs files only, the short flow is enabled. Perhaps we should altogether remove this feature, as CircleCI processed a build within ~15 minutes, which is acceptable compared to the previous Travis results (~40 min).

PR #14926, opened 2023-09-04
Fixes usage of MentionAttribute.

### LINKED ISSUE #14923: TS2345: Argument of type  { userId: unknown; }  is not assignable to parameter of type  MentionAttribute

I am using custom build of CKEditor 5 with Angular. Here is the function I use, almost the same as example provided in the documentation:

```
private mentionCustomization(editor: CustomCkeEditor.default) {
    // The upcast converter will convert view <a class="mention" href="" data-user-id="">
    // elements to the model 'mention' text attribute.
    editor.conversion.for('upcast').elementToAttribute({
      view: {
        name: 'span',
        key: 'data-mention',
        classes: 'mention',
        attributes: {
          'data-user-id': true,
        },
      },
      model: {
        key: 'mention',
        value: (viewItem: Element) => {
          // The mention feature expects that the mention attribute value
          // in the model is a plain object with a set of additional attributes.
          // In order to create a proper object use the toMentionAttribute() helper method:
          return editor.plugins.get('Mention').toMentionAttribute(viewItem, {
            // Add any other properties that you need.
            userId: viewItem.getAttribute('data-user-id'),
          } as unknown as MentionAttribute);
        },
      },
      converterPriority: 'high',
    });

    // Downcast the model 'mention' text attribute to a view <span> element.
    editor.conversion.for('downcast').attributeToElement({
      model: 'mention',
      view: (modelAttributeValue: { [key: string]: string }, { writer }: any) => {
        // Do not convert empty attributes (lack of value means no mention).
        if (!modelAttributeValue) {
          return;
        }

        return writer.createAttributeElement(
          'span',
          {
            class: 'mention',
            'data-mention': modelAttributeValue.id,
            'data-user-id': modelAttributeValue.userId,
          },
          {
            // Make mention attribute to be wrapped by other attribute elements.
            priority: 20,
            // Prevent merging mentions together.
            id: modelAttributeValue.uid,
          }
        );
      },
      converterPriority: 'high',
    });
  }
```

### ✔️ Expected result

It should not show the error `Argument of type  { userId: unknown; }  is not assignable to parameter of type  MentionAttribute  Object literal may only specify known properties, and  userId  does not exist in type  MentionAttribute`

### ❌ Actual result
It shows `Argument of type  { userId: unknown; }  is not assignable to parameter of type  MentionAttribute  Object literal may only specify known properties, and  userId  does not exist in type  MentionAttribute`

### ❓ Possible solution
Example listed [here](https://ckeditor.com/docs/ckeditor5/latest/features/mentions.html#customizing-the-output) says ` // Add any other properties that you need.` But the type `MentionAttribute` is preventing that. MentionAttribute type should allow me to define the properties I need, like it says in the docs example.

## 📃 Other details

* Browser: Chrome
* OS: Windows 10
* First affected CKEditor version: Not sure, probably since typescript is introduced to custom build.
* Installed CKEditor plugins: Autoformat, AutoImage, Autosave, BlockQuote, Bold, Essentials, FontSize, Heading, Image, ImageUpload, Indent, Italic, Link, List, Mention, Paragraph, PasteFromOffice, SimpleUploadAdapter, TextTransformation, 

---

arkflpc:

Fix (mention): Fixes typing of `Mention#toMentionAttribute` method. Closes #14923.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-mention/src/mention.ts
+++ b/packages/ckeditor5-mention/src/mention.ts
@@ -22,7 +22,7 @@ import '../theme/mention.css';
  */
 export default class Mention extends Plugin {
 	/**
-	 * Creates a mention attribute value from the provided view element and optional data.
+	 * Creates a mention attribute value from the provided view element and additional data.
 	 *
 	 * ```ts
 	 * editor.plugins.get( 'Mention' ).toMentionAttribute( viewElement, { userId: '1234' } );
@@ -40,7 +40,7 @@ export default class Mention extends Plugin {
 	): ( MentionAttribute & MentionData ) | undefined;
 
 	/**
-	 * Creates a mention attribute value from the provided view element and optional data.
+	 * Creates a mention attribute value from the provided view element.
 	 *
 	 * ```ts
 	 * editor.plugins.get( 'Mention' ).toMentionAttribute( viewElement );
--- a/packages/ckeditor5-mention/src/mention.ts
+++ b/packages/ckeditor5-mention/src/mention.ts
@@ -35,7 +35,28 @@ export default class Mention extends Plugin {
 	 * @param viewElement
 	 * @param data Additional data to be stored in the mention attribute.
 	 */
-	public toMentionAttribute( viewElement: Element, data?: MentionAttribute ): MentionAttribute | undefined {
+	public toMentionAttribute<MentionData extends object>(
+		viewElement: Element,
+		data: MentionData
+	): ( MentionAttribute & MentionData ) | undefined;
+
+	/**
+	 * Creates a mention attribute value from the provided view element and optional data.
+	 *
+	 * ```ts
+	 * editor.plugins.get( 'Mention' ).toMentionAttribute( viewElement );
+	 *
+	 * // For a view element: <span data-mention="@joe">@John Doe</span>
+	 * // it will return:
+	 * // { id: '@joe', uid: '7a7bc7...', _text: '@John Doe' }
+	 * ```
+	 *
+	 * @param viewElement
+	 * @param data Additional data to be stored in the mention attribute.
+	 */
+	public toMentionAttribute( viewElement: Element ): MentionAttribute | undefined;
+
+	public toMentionAttribute( viewElement: Element, data?: object ): MentionAttribute | undefined {
 		return _toMentionAttribute( viewElement, data );
 	}
 
@@ -71,7 +92,7 @@ export type MentionAttribute = {
 	 * A unique ID of this mention instance. Should be passed as an `option.id` when using
 	 * {@link module:engine/view/downcastwriter~DowncastWriter#createAttributeElement writer.createAttributeElement()}.
 	 */
-	uid?: string;
+	uid: string;
 
 	/**
 	 * Helper property that stores the text of the inserted mention. Used for detecting a broken mention
@@ -79,5 +100,5 @@ export type MentionAttribute = {
 	 *
 	 * @internal
 	 */
-	_text?: string;
+	_text: string;
 };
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-mention/src/mention.ts
```diff
@@ -35,7 +35,28 @@ export default class Mention extends Plugin {
 	 * @param viewElement
 	 * @param data Additional data to be stored in the mention attribute.
 	 */
-	public toMentionAttribute( viewElement: Element, data?: MentionAttribute ): MentionAttribute | undefined {
+	public toMentionAttribute<MentionData extends object>(
```
filipsobol, line 38: Consider using `{ [ key: string ]: unknown }` like in `packages/ckeditor5-mention/src/mentioncommand.ts`) instead of `object`.

[TypeScript Playground](https://www.typescriptlang.org/play?#code/DYUwLgBAhgXBDeEDaBrEBPOBnMAnAlgHYDmAunAK6EqED2A7oRAL4QC8CzA3ALABQUAHQAzWrV59+oSACM4tGQCsQAY0gd43fjJFiJU8BBVwASqtq4AJgB4cBEgBoIVGg0IA+dpwkrd4oA).

This also applies to the line 59 in this file and lines 88 and 103 in `packages/ckeditor5-mention/src/mentionediting.ts`.

arkflpc: Why? The data is just consumed, not produced.

filipsobol: It's always better to be explicit about what type you expect. The `object` type means **any** object, including `() => {}` and `[]`, since both are objects in JavaScript. See the [TypeScript Playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABHARgKwKbQCoE8AOGAFKmgFzLpZQCUFAbnDACaIDeAvgLABQvp1PISKcaAbkSSpUgPQzEAQQDOiDAA9C0DM35UcBYkRqIAvAD52HcbPmBeDcClO7sz7hAbQC616dLmIHvXlBIWAREACcsODDmIWJSCgAlSOiAHiUoMJgwAHMAGkRwAGswOAB3MDM6REYWSwCeCIgomIMRKwlvG0UVdU0obV5G5tiiI1MLUQ7EXwB1AEMwsCy8xDmVUoxVDWpB5Ja3TynO2YWlnPy1xA2tvvqg6HgkARcMGZgoAAsAZQzlgGkMLglCR0BQ2IhXIhCoCKOlMjlEO4KEUSuVEFYGExWJx6s8oLE3p8fvDsgCgW0vJ1pvJlDdqAMeHiCe9vr8cmTgWNzJZKSdFssLutNr0doy9PiDITWSSOUQPJTjvJ5vzzqshXToEA) example.

In this case, to be more explicit and only allow objects with string keys, as we [show in the documentation](https://ckeditor.com/docs/ckeditor5/latest/features/mentions.html#customizing-the-output), we should use `{ [ key: string ]: unknown }` or `Record<string, unknown>` (I'm a fan of `Record` because it's more concise, but in this case it doesn't matter).

packages/ckeditor5-mention/src/mention.ts
```diff
@@ -32,10 +32,27 @@ export default class Mention extends Plugin {
...
 	 * @param data Additional data to be stored in the mention attribute.
 	 */
-	public toMentionAttribute( viewElement: Element, data?: MentionAttribute ): MentionAttribute | undefined {
+	public toMentionAttribute<MentionData extends Record<string, unknown>>(
+		viewElement: Element,
+		data: MentionData
+	): ( MentionAttribute & MentionData ) | undefined;
+
+	/**
+	 * Creates a mention attribute value from the provided view element and optional data.
```
mmotyczynska, line 43: For both signatures of the `toMentionAttribute` method, we should change descriptions, because both have the phrase "optional data". In the first one `data` parameter is mandatory and in the second one there is no `data` parameter.

mmotyczynska (commented):
Besides descriptions mentioned in a comment, LGTM 👍

filipsobol (approved):

mmotyczynska (approved):
Looks good 👍

PR #14970, opened 2023-09-11
Improve the horizontal drop target line.

### LINKED ISSUE #14645: [Drag & drop, intro] Improve the drop target line
We want to improve the drop target line for both horizontal and vertical lines:

*   Symmetrical approach – arrow should also be added in the horizontal line, 
    * RTL, the arrow should be on the other side.
*   For darker background
    * The lines vertical and horizontal by default should be surrounded by the white stroke (0.5-1px).
    * Outline may be tricky to style, intersections, etc., SVG?
    * _Possible scope cut._
*   I should be possible to style both line color and the surrounding stroke by the CSS.

[Internal Figma file.](https://www.figma.com/file/B20B2kvXTDzzO0zvNZmPE4/Enhanced-drag-and-drop-of-block?type=design&node-id=13-755&mode=design&t=9mSNImGgnYAmLhXL-4)
---

pszczesniak:

Other (theme-lark): Improve the drag and drop target line.

Closes #14645.

---

### Additional information

One thing is not done - white outline for darken backgrounds.

`CSS` outlines and box shadows wont work cause we're using two elements combined `triangle` a `pseudo element` to imitate the arrow and the line (vertical or horizontal). Those CSS properties works on each of them - for example a shadow from the arrow covers the line.

2nd approach was to use `filter: drop-shadow` and this solution was looking promising (Chrome):

https://github.com/ckeditor/ckeditor5/assets/7074833/9f007bd5-541a-4c02-812b-f293a098ab89

BUT it completely not working on Safari (i found that dragging elements with `filter: drop-shadow` issue is there for years):

https://github.com/ckeditor/ckeditor5/assets/7074833/ef162018-8c84-46d7-b797-92473c5f25ff

Using ASCII/Unicode signs as the arrows also wasn't working as expected.

---
### Idea 💡 

Maybe in future we can think about different solution/approach - get the background color of the editing area and based on this set some light/dark colors (it's different than the `prefers-color-scheme`). [Simple example](https://codepen.io/ikarium/pen/yXPQpo) of what i have on my mind.


Witoso:
FYI @dagdzi, we needed to cut the white outline for now.

PR #14975, opened 2023-09-12
Readded short flow on the stable branch and through commit message.

### LINKED ISSUE #14928: Restore short flow on CI by explicitly triggering it via a commit message

In #14916, we have dropped the short-flow functionality on CI due to issues determining when to run full or short-flow.

We can restore the short-flow option executed on demand. If a developer included the `[short-flow]` or `[short flow]` suffix in a commit message, CIs could run a short-flow.

martnpaneq:

Internal: Readded short flow on the stable branch and through commit message. Closes #14928.

---

### Additional information

In order to execute short flow add "[short-flow]" in the commit message.

### DIFF
--- a/.circleci/template.yml
+++ b/.circleci/template.yml
@@ -13,11 +13,12 @@ commands:
   halt_if_short_flow:
     steps:
       - run:
-          name: Check Commit Message
+          name: ⭐ Short flow breakpoint - Check if the build should continue
           command: |
             COMMIT_MESSAGE=$(git log -n 1 --pretty=format:%s ${CIRCLE_SHA1})
-            if [[ "$COMMIT_MESSAGE" == *"[short-flow]"* ]]; then
-              echo "Skipping, because the commit message contains '[short-flow]'."
+            
+            if [[ "$COMMIT_MESSAGE" == *"[short flow]" ]]; then
+              echo "Skipping, because the commit message contains '[short flow]'."
               circleci step halt
             fi
 
@@ -141,7 +142,6 @@ jobs:
     steps:
       - community_verification_command
       - checkout_command
-      - halt_if_short_flow
       - bootstrap_repository_command
       - run:
           name: Trigger the Uber CI
@@ -226,8 +226,6 @@ workflows:
             branches:
               only:
                 - master
-              ignore:
-                - stable
       - notify_ci_failure:
           filters:
             branches:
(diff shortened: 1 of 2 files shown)


.circleci/template.yml
```diff
@@ -128,6 +141,7 @@ jobs:
     steps:
       - community_verification_command
       - checkout_command
+      - halt_if_short_flow
```
pomek, line 144:
```suggestion
```

Short flow should trigger Uber CI, as it checks docs.

.circleci/template.yml
```diff
@@ -10,6 +10,17 @@ parameters:
     default: false
 
 commands:
+  halt_if_short_flow:
+    steps:
+      - run:
+          name: Check Commit Message
```
pomek, line 16:
```suggestion
          name: ⭐ Short flow breakpoint - Check if the build should continue
```

Let's describe "what it checks" instead of "what it does". The latter can be read from the script.

pomek: The _star_ symbol is used on flow charts and Notion. Hence, I would use it as a common symbol when explaining different formats.

.circleci/template.yml
```diff
@@ -193,6 +226,8 @@ workflows:
             branches:
               only:
                 - master
+              ignore:
+                - stable
```
pomek, line 230: It seems not to be needed due to `only` instruction above.

.circleci/template.yml
```diff
@@ -10,6 +10,17 @@ parameters:
     default: false
 
 commands:
+  halt_if_short_flow:
+    steps:
+      - run:
+          name: Check Commit Message
+          command: |
+            COMMIT_MESSAGE=$(git log -n 1 --pretty=format:%s ${CIRCLE_SHA1})
+            if [[ "$COMMIT_MESSAGE" == *"[short-flow]"* ]]; then
```
pomek, line 19:
```suggestion
            COMMIT_MESSAGE=$(git log -n 1 --pretty=format:%s ${CIRCLE_SHA1})

            if [[ "$COMMIT_MESSAGE" == *"[short-flow]"* ]]; then
```

A blank line would help separate variables from instructions.

pomek, line 19: WDYT about supporting only a suffix `[short flow]`? Example commit:

```
Abc. [short flow]
```

Similar to [`[skip ci]`](https://circleci.com/docs/skip-build/#skip-jobs).

pomek:
Other than mentioned, we need to remember to update docs.

pomek (approved):
LGTM.

PR #14977, opened 2023-09-13
Improved images preview while dragging them.

### LINKED ISSUE #14744: [Drag & drop] Improve images preview
We want to improve the image preview (especially for large images) so that they do not cover the content during using drag and drop feature.

We have chosen the following approach: **Image preview cannot be bigger than 50% of editor area.** 
The preview is smaller when the image is bigger than 50% of editor area, and the user drags the image by clicking on it.

In the Balloon block editor, when the user uses the drag handle to drag an image, the preview is the same size as the image size. 

[Figma prototype ](https://www.figma.com/proto/B20B2kvXTDzzO0zvNZmPE4/Enhanced-drag-and-drop-of-block?type=design&node-id=191-484&t=6RLoxFqczYmzUYzf-1&scaling=min-zoom&page-id=191%3A428&starting-point-node-id=191%3A484&show-proto-sidebar=1&mode=design)(click on the image to see the preview).

---

pszczesniak:

Fix(clipboard): Images should be now optimal scaled while dragging them.. Closes #14744 .

---

### Additional information

_Optional PR_


pszczesniak:
Different approach was chosen.

Issue #14924, opened 2023-09-04
Upgrade the minimal required Node.js version to 18


pomek:
## Why

- [Node@16 ends its LTS in 2023-09-11.](https://github.com/nodejs/Release)

## What

- Upgrade the minimal requested version of Node.js.
- Currently, we require Node@16. Upgrade it to Node@18.
- In CKEditor 5, it's a change for developers integrating the ckeditor/ckeditor5 repository only.
	- A related thread: https://github.com/ckeditor/ckeditor5/issues/14586.
- In repositories using Node.js (ckeditor5-dev, mrgit, etc.), it's a breaking change requiring bumping the major version.
- In the docs repository, do not forget about bumping the Node.js version in `Dockerfile`.
- Could we drop the `node-fetch` package in favor of the built-in function?

## Before start

* Talk to CERD (@jacekbogdanski) about whether they support Node@18. If you need more clarification, ask to verify, as the documentation builder will require Node 18 soon. Perhaps @vokiel could help.

pomek:
> - Could we drop the `node-fetch` package in favor of the built-in function?

We decided to leave the current scripts as they are to avoid potential issues using an experimental API.

Source: https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch.

pomek:
The `ckeditor5-dev-*` packages are released as https://github.com/ckeditor/ckeditor5-dev/releases/tag/v39.0.0.

Bump in CKE5: https://github.com/ckeditor/ckeditor5/commit/877a98e427114d1ade0ab8f6732f96f1a002b4a6. CC is covered too.

pomek:
I am closing this issue as all CKEditor 5 repositories/packages are aligned to Node@18.

The remaining PRs are tracked in an internal issue.

PR #14989, opened 2023-09-13
Node version bump

przemyslaw-zan:
Other: Updated the required version of Node.js to 18 for the development environment. See #14924.


pomek (approved):
LGTM.

Issue #14995, opened 2023-09-14
Add "ck ck-content" classes to all snippets in the docs


filipsobol:
Currently, the snippets in the documentation don't have the `"ck ck-content"` classes, so they look different before and after loading the editor, which is not an ideal representation of the What You See Is What You Get (WYSIWYG) editor.

To prevent layout shifting and to display the same content in both scenarios, we should add the `"ck ck-content"` classes to all snippets in the documentation.

godai78:
Actually it seems like there is this class present in the snippets style sheet:

https://github.com/ckeditor/ckeditor5/blob/dd2e92dd25ed169221ccabc287afc01240fa26de/docs/assets/snippet-styles.css#L20

pszczesniak:
It has to be added into `HTML` , for example:

```
<div id="snippet-classic-editor">
```

:point\_down: 

```
<div class="ck ck-content" id="snippet-classic-editor">
```

PR #14998, opened 2023-09-14
Docs: updating demos CSS.

godai78:

Docs: updating demos CSS. Closes https://github.com/ckeditor/ckeditor5/issues/14995

---

### Additional information

Merge together with https://github.com/cksource/ckeditor5-commercial/pull/5514


Witoso:
*   The inline editor example is still jumping, I think classes need to be added to all direct containers of the editor, e.g. `header`. Not sure, as maybe a different CSS styles it.
*   Classic behaves perfectly, full-featured is also a classic but the image jumps left to right (better than right now, anyway).


https://github.com/ckeditor/ckeditor5/assets/9881379/b6ae0c79-8931-41ea-8030-2355197db9ea

godai78:
This one is hard, adding the class to the very container in question made things even worse.

Adding a loader class (like we do in CF) to postpone display before loading did nothing at all.

I changed the image alignment to the left, so it jumps a little less, but I can come up with no handy solution for that.

We might consider putting it into a follow up, if we want global changes fast.

Witoso:
Good for me :+1:

PR #15011, opened 2023-09-15
Fix triple click issue in paragraph with widget

### LINKED ISSUE #11130: Not able to select whole paragraph with triple click when there is a table right after it

1. Add some text in paragraph
2. Add a table right after the paragraph
3. Triple click on the paragraph to select it

### ✔️ Expected result

Whole paragraph is selected.

### ❌ Actual result

Paragraph is not selected. Selection collapses and jumps to the beginning of the paragraph.

### Screencast

https://user-images.githubusercontent.com/34380544/149969165-4aad13d7-6d15-418b-be0d-0112ca44a3ab.mp4

## 📃 Other details

* First affected CKEditor version: `27.0.0`

---

illia-stv:

Fix ( widget ): Not able to select whole paragraph with triple click when there is a table right after it. Closes #11130.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-widget/src/widget.ts
+++ b/packages/ckeditor5-widget/src/widget.ts
@@ -207,20 +207,17 @@ export default class Widget extends Plugin {
 			// But at least triple click inside nested editable causes broken selection in Safari.
 			// For such event, we select the entire nested editable element.
 			// See: https://github.com/ckeditor/ckeditor5/issues/1463.
+
 			if ( ( env.isSafari || env.isGecko ) && domEventData.domEvent.detail >= 3 ) {
-				const mapper = editor.editing.mapper;
-				const viewElement = element.is( 'attributeElement' ) ?
-					element.findAncestor( element => !element.is( 'attributeElement' ) )! : element;
-				const modelElement = mapper.toModelElement( viewElement )!;
-
-				domEventData.preventDefault();
-
-				this.editor.model.change( writer => {
-					writer.setSelection( modelElement, 'in' );
-				} );
+				setSelectionInAncestorElement();
 			}
 
 			return;
+		}
+
+		// If triple click should select entire paragraph.
+		if ( domEventData.domEvent.detail >= 3 ) {
+			setSelectionInAncestorElement();
 		}
 
 		// If target is not a widget element - check if one of the ancestors is.
@@ -247,6 +244,21 @@ export default class Widget extends Plugin {
 		const modelElement = editor.editing.mapper.toModelElement( element );
 
 		this._setSelectionOverElement( modelElement! );
+
+		function setSelectionInAncestorElement() {
+			if ( element ) {
+				const mapper = editor.editing.mapper;
+				const viewElement = element.is( 'attributeElement' ) ?
+					element.findAncestor( element => !element.is( 'attributeElement' ) )! : element;
+				const modelElement = mapper.toModelElement( viewElement )!;
+
+				domEventData.preventDefault();
+
+				editor.model.change( writer => {
+					writer.setSelection( modelElement, 'in' );
+				} );
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


arkflpc (approved):

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -247,6 +244,21 @@ export default class Widget extends Plugin {
...
+				const viewElement = element.is( 'attributeElement' ) ?
+					element.findAncestor( element => !element.is( 'attributeElement' ) )! : element;
+				const modelElement = mapper.toModelElement( viewElement )!;
+
+				domEventData.preventDefault();
+
+				editor.model.change( writer => {
+					writer.setSelection( modelElement, 'in' );
+				} );
+			}
```
filipsobol, line 260: Two tiny formatting suggestions.

1. To reduce unnecessary nesting, I like to use the [early return pattern](https://gomakethings.com/the-early-return-pattern-in-javascript/). In this case, we can reduce the nesting by one level by returning early if `element` is falsy.

```js
if ( !element ) {
  return;
}

// The rest of the code
```

2. Conditional (ternary) operator is a little bit more readable when true and false expressions are in separate lines:

```js
const viewElement = element.is( 'attributeElement' ) ?
  element.findAncestor( element => !element.is( 'attributeElement' ) )! :
  element;
```

filipsobol (approved):
The above comment is nitpicky and optional. I'd like to see it included, but I'm fine with PR as is.

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -247,6 +244,21 @@ export default class Widget extends Plugin {
 		const modelElement = editor.editing.mapper.toModelElement( element );
 
 		this._setSelectionOverElement( modelElement! );
+
+		function setSelectionInAncestorElement() {
```
niegowski, line 248: Why this is a nested function? I'd suggest extracting to a private method or a helper function.

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -207,22 +207,19 @@ export default class Widget extends Plugin {
...
+				setSelectionInAncestorElement();
 			}
 
 			return;
 		}
 
+		// If triple click should select entire paragraph.
+		if ( domEventData.domEvent.detail >= 3 ) {
+			setSelectionInAncestorElement();
+		}
```
niegowski, line 221: Maybe we should narrow it to just affected cases with widgets? This looks like first we check if in nested editable on FF or Safari and then we do it everywhere.

niegowski (changes requested):

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -207,22 +207,25 @@ export default class Widget extends Plugin {
...
+		const parentElement = element.parent;
+
+		// If triple click should select entire paragraph.
+		if ( parentElement &&
+			parentElement.is( 'element' ) &&
+			!isWidget( parentElement ) &&
+			domEventData.domEvent.detail >= 3
+		) {
+			this._setSelectionInAncestorElement( element, domEventData );
+		}
```
niegowski, line 227: Why are checking the parent element? I'm not sure what this condition is checking.

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -249,6 +244,38 @@ export default class Widget extends Plugin {
 		this._setSelectionOverElement( modelElement! );
 	}
 
+	/**
+	 * Selects entire block content, e.g. on triple click it selects entire paragraph.
+	 */
+	private _selectBlockContent( element: ViewElement ): boolean {
```
niegowski, line 250: This code could be simpler, we could use some existing helper methods. Let's talk about it f2f.

niegowski (changes requested):

Witoso:
I would release the first iteration with the white background, it improves a bad experience. All other edge cases could be done in a follow-up. @illia-stv check how the preview of tables would look with your changes with balloon block.

niegowski (approved):

PR #15012, opened 2023-09-15
Don't show drop marker in places where the dragged elements are not allowed

### LINKED ISSUE #14709: [Drag & drop] Prevent dropping in the

Don’t show the drop target inside the element that is selected or where you cannot place it, examples:

1.  Table into itself,
2.  Block quote into itself.
3.  Inline image to block's image caption.

### ✔️ Expected result

Drop target line doesn't show.

### ❌ Actual result

Drop target line does show.

---

filipsobol:

Fix (clipboard): Don't show drop marker in places where the dragged elements are not allowed. Closes #14709.


illia-stv (approved):
Looks very good ✅, I have added only few tiny suggestions.

niegowski (changes requested):
I think this behavior is not the expected one (you can drop table after a heading but visually the drop target is not updated:

filipsobol:
> I think this effect is still very weird:
> 
>  [  ](https://user-images.githubusercontent.com/1232187/270365791-211e3b96-77c6-413b-b426-19e804080a20.gif) [ ](https://user-images.githubusercontent.com/1232187/270365791-211e3b96-77c6-413b-b426-19e804080a20.gif)
> 
> I think the solution should be inside `findDropTargetRange()`, not after it.

Dropping the table on the image will place it before or after the image. However, dropping the table on the `<caption>` element does nothing because it's not allowed.

Drop target now reflects this.

filipsobol:
> > Drop target now reflects this.
> 
> Unfortunately on the attached GIF, you can see that a drop on the caption was possible (no drop target visible) but it is inserted below the image.

Fixed

packages/ckeditor5-clipboard/src/dragdroptarget.ts
```diff
@@ -122,16 +123,23 @@ export default class DragDropTarget extends Plugin {
 		targetViewRanges: Array<ViewRange> | null,
 		clientX: number,
 		clientY: number,
-		blockMode: boolean
+		blockMode: boolean,
```
arkflpc, line 126: Nitpick: There are so many arguments. I'd consider wrapping it with `option` argument. That way on call site the arguments would be named, not positional and less error-prone.

packages/ckeditor5-clipboard/src/dragdrop.ts
```diff
@@ -345,7 +345,7 @@ export default class DragDrop extends Plugin {
 
 			const { clientX, clientY } = ( data as DomEventData<DragEvent> ).domEvent;
 
-			dragDropTarget.updateDropMarker( data.target, data.targetRanges, clientX, clientY, this._blockMode );
+			dragDropTarget.updateDropMarker( data.target, data.targetRanges, clientX, clientY, this._blockMode, this._draggedRange );
```
arkflpc, line 348: This line could look more readable, see my comment in `dragdroptarget.ts`.

filipsobol: This would be a breaking change because this method is public and this PR will land in the next release (after the one where we'll introduce the `DragDropTarget` class).

Also, to be consistent, we should also change the params in the `getFinalDropRange` method, which would introduce a lot of noise in the PR.

Unless this is a hard blocker, I would skip this change.

packages/ckeditor5-clipboard/src/dragdroptarget.ts
filipsobol: This would be a breaking change because this method is public and this PR will land in the next release (after the one where we'll introduce the `DragDropTarget` class).

Also, to be consistent, we should also change the params in the `getFinalDropRange` method, which would introduce a lot of noise in the PR.

Unless this is a hard blocker, I would skip this change.

packages/ckeditor5-clipboard/src/dragdrop.ts
filipsobol: See my response in other comment.

niegowski:
While playing with latest version I got an exception:

```
schema.ts:184 Uncaught TypeError: Cannot use 'in' operator to search for 'is' in undefined
    at Schema.getDefinition (schema.ts:184:1)
    at Schema.isBlock (schema.ts:225:1)
    at canDropOnRange (dragdroptarget.ts:517:1)
    at DragDropTarget.updateDropMarker (dragdroptarget.ts:138:1)
    at listenTo.priority (dragdrop.ts:348:1)
    at Mixin.fire (emittermixin.ts:241:1)
    at fireListenerFor (bubblingemittermixin.ts:207:1)
    at Document.fire (bubblingemittermixin.ts:118:1)
    at Mixin.<anonymous> (clipboardobserver.ts:68:1)
    at Mixin.fire (emittermixin.ts:241:1)
```

filipsobol:
> While playing with latest version I got an exception:
> 
> ```
> schema.ts:184 Uncaught TypeError: Cannot use 'in' operator to search for 'is' in undefined
> ```

Fixed

niegowski (changes requested):
F2F live review

packages/ckeditor5-clipboard/src/dragdroptarget.ts
```diff
@@ -347,19 +375,26 @@ function findDropTargetRange(
...
+				if ( targetViewRanges ) {
+					const targetViewPosition = targetViewRanges[ 0 ].start;
+					const targetModelPosition = mapper.toModelPosition( targetViewPosition );
+					const canDropOnPosition = !draggedRange || Array
+						.from( draggedRange.getItems() )
+						.every( item => model.schema.checkChild( targetModelPosition, item as Node ) );
+
+					if ( canDropOnPosition ) {
+						if ( model.schema.checkChild( targetModelPosition, '$text' ) ) {
+							return model.createRange( targetModelPosition );
```
filipsobol, line 387: `while...if...if...if...if...if` - this method becomes unmaintainable.

This is outside the scope of this task, but we need to refactor this method to make it more readable, but also keep it performant, because it's called every frame when something is dragged.

Thoughts @niegowski @arkflpc ?

packages/ckeditor5-clipboard/src/dragdroptarget.ts
arkflpc: It is not public, but internal. My comment was a nitpick - safe to be ignored.

arkflpc (approved):

PR #15016, opened 2023-09-18
Enable block drag and drop by default

### LINKED ISSUE #14734: [Drag & drop] Enable block drag and drop by default
1.  Add drag and drop of blocks to work by default.
2.  Add drag and drop by block toolbar handle to work by default when block toolbar is used.
3.  Remove the old implementation of Drag & drop.

illia-stv:

Feature (clipboard): Block drag and drop should be enabled by default. Closes #14734.

---

### Additional information


packages/ckeditor5-clipboard/src/clipboard.ts
```diff
@@ -10,7 +10,7 @@
 import { Plugin } from '@ckeditor/ckeditor5-core';
 
 import ClipboardPipeline from './clipboardpipeline';
-import DragDrop from './dragdrop';
+import DragDropExperimental from './dragdropexperimental';
```
arkflpc, line 13: `DragDropExperimental` should replace old `DragDrop` plugin.

arkflpc (changes requested):

niegowski (changes requested):
Missing code coverage.

niegowski (approved):

PR #15025, opened 2023-09-19
Autocomplete improvements

oleq:

The `AutocompleteView` should react to its `#filteredView`'s `execute` event. Moved the history icon from ckeditor5-revision-history to ckeditor5-core. `TextareaView` should fire an event whenever it calculates its height.


packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/autocomplete/autocomplete.css
```diff
@@ -20,11 +20,13 @@
 		&.ck-search__results_n {
 			border-bottom-left-radius: 0;
 			border-bottom-right-radius: 0;
+			margin-bottom: -1px;
```
DawidKossowski, line 23: Why do we need it?

DawidKossowski (approved):
Generally, it looks good

packages/ckeditor5-ui/src/textarea/textareaview.ts
```diff
@@ -193,3 +195,13 @@ function getTextareaElementClone( element: HTMLTextAreaElement, value: string ):
 
 	return clone;
 }
+
+/**
+ * Fired when the textarea view is resized.
```
DawidKossowski, line 200: I'm wondering if `resized` is causing any confusion. In the context of textarea, `resize` is typically used when a user has changed the size.

packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/autocomplete/autocomplete.css
oleq: When the results panel is displayed next to the input via `bottom: 100%;` or `top: 100%`, you get a double border: one pixel from the input, another from the panel. This mitigates the duplication. We don't have a custom prop for border-width, so I cannot use `calc( -1 * var(--some-custom-prop ) )` here.

Issue #11656, opened 2022-04-21
[Document list] Avoid breaking a document list when pasting into an item


oleq:
## 📝 Provide a description of the improvement

There are two cases that prove we should avoid breaking lists while pasting. The bottom line is it's very unlikely anyone wants the list to be broken on purpose on paste.

## Case 1

Paste into a flat list. When it gets broken, selecting the pasted content and clicking list creates billions of list items. It's hard to merge them into a single item.

![2022-04-21 10 42 19](https://user-images.githubusercontent.com/1099479/164416469-ea41f710-9417-4855-b901-09ec10a32958.gif)

## Case 2

Pasting into a deeply indented list. Breaking the list alters the structure that follows the pasted content: it gets flattened because lists cannot begin with a deep indentation and then successively decrease it. It's pretty much unfixable (in a few steps) and I consider this a content loss.

![2022-04-21 10 48 17](https://user-images.githubusercontent.com/1099479/164417423-7c315642-6e99-4fa7-bd02-ab6aeb72c89c.gif)

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

oleq:
This should also fix #11547.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity over the last year. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

mabryl:
Re-opening and adding a support label, since a client reached out to us via Zendesk inquiring about this issue.

Witoso:
Current behavior (that was a part of #15042 fix).

### CASE 1

We still don't detect pasting inside, but the list is not broken.

https://github.com/ckeditor/ckeditor5/assets/9881379/a0a146af-4a28-4911-aa3a-b0af7da5a0fc

### CASE 2

This works much better than before. Still no paste inside but the list structure is kept.

https://github.com/ckeditor/ckeditor5/assets/9881379/e1fe8b0f-4554-4fa6-9723-9b313d5aae3b

PR #15042, opened 2023-09-21
[Document lists] Improve pasting plain-text lists and lists of different type into document list

### LINKED ISSUE #13826: List breaks when you paste content as plain text

1. Go to [demo](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/classic-editor.html).
2. Create a list with two nested list items
3. Copy two list items from some other list
4. Put the cursor at the beginning of the first nested list item
5. Paste as plain text (`cmd+shift+v`)

### ✔️ Expected result

Content gets pasted as new list items or at least inside the list item. Example from Google Docs.
  
First paste is normal, second is pasted as plain tex:

https://user-images.githubusercontent.com/34380544/230026525-f03b6eb8-9078-4a90-b8aa-2c8c00baf8eb.mov

### ❌ Actual result

List breaks (happens for both normal and document lists).
  
First paste is normal, second is pasted as plain text:

https://user-images.githubusercontent.com/34380544/230026602-5f2d86b5-a00c-46b7-81d9-5e9e1272754d.mov

### ❓ Possible solution

## 📃 Other details

* Browser: …
* OS: …

---

filipsobol:

Fix (list): Pasting plain text content should not break the lists. Closes #13826.

Fix (list): Pasting one list into another should preserve the target list type. See #13826. 

---

#### Disclaimer

The issue reported in #13826 affects both Lists and DocumentLists. However, since we will soon be deprecating Lists, this PR only fixes DocumentLists.

When discussing this internally, we came to the conclusion that when pasting content into a list, we should do our best to **not** break it, even if the pasted content is not itself a list.

#### Pasting plain text

Copying and pasting plain text lists no longer breaks them.

[plain_text_paste.webm](https://github.com/ckeditor/ckeditor5/assets/4145208/e9825f97-8569-40a2-9a02-d318b4c26496)

#### Pasting paragraphs

Since plain text list items are just paragraphs, we can now also paste regular paragraphs without breaking the list too.

[pasting_paragraphs.webm](https://github.com/ckeditor/ckeditor5/assets/4145208/638b779c-95fc-48e9-8c69-0ca70ba733ec)

This also works when pasting lists separated by paragraphs.

[pasting_paragraph_mixed_with_lists.webm](https://github.com/ckeditor/ckeditor5/assets/4145208/f6c513ff-7c9c-41a6-ba29-dccd041de1ba)

This partially fixes issue reported in #11656, as pasting paragraphs no longer breaks the list. However, paragraphs pasted into a multi-line list item should place the content within it, not convert the paragraphs into new list items.

#### Preserving list style

When pasting both plain text and regular lists into another list, the original list type is preserved.

[preserving_list_style.webm](https://github.com/ckeditor/ckeditor5/assets/4145208/1c4286d6-ba0f-4c82-aa01-2f1498759d65)


arkflpc:
> Don't break the list when pasting plain text content into it.
Please rephrase it, so it doesn't sound like an imperative to the changelog reader. Like "Fixes pastsing plaing text..." or "Pasting plaing text content should not break the list..."

arkflpc (approved):

packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
```diff
@@ -21,41 +21,61 @@ const smallPaddingElements = [ 'figcaption', 'li' ];
...
+function newLinePadding(
+	element: ViewElement,
+	previous: ViewElement | null
+): string {
+	if ( !previous ) {
+		// Don't add padding to first elements in a level.
+		return '';
+	}
+
+	if ( element.is( 'element', 'li' ) && !element.getChild( 0 )!.is( '$text' ) ) {
```
niegowski, line 64: I'm not sure if this is safe. We should check if there is any child before accessing it.

packages/ckeditor5-list/src/documentlist/documentlistediting.ts
```diff
@@ -767,72 +768,68 @@ function modelChangePostFixer(
...
+
+				for ( const child of content.getChildren() ) {
+					const isListItem = isListItemBlock( child );
+
+					writer.setAttributes( {
+						listIndent: ( isListItem ? child.getAttribute( 'listIndent' ) : 0 ) + indentDiff,
+						listItemId: isListItem ? child.getAttribute( 'listItemId' ) : ListItemUid.next(),
+						listType: refType
+					}, child );
+				}
```
niegowski, line 832: This part looks strange. You should not rely on the `content` type. It's just a different way of passing parameters to this function. The logic that alters the content should be unified.

niegowski (changes requested):
Also, CI is failing on some tests.

packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
filipsobol: I added check to ensure that `element` is not empty

packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
```diff
@@ -21,41 +21,61 @@ const smallPaddingElements = [ 'figcaption', 'li' ];
...
+): string {
+	if ( !previous ) {
+		// Don't add padding to first elements in a level.
+		return '';
+	}
+
+	if ( element.is( 'element', 'li' ) && !element.isEmpty && !element.getChild( 0 )!.is( '$text' ) ) {
+		// Separate document list items with empty lines.
+		return '\n\n';
+	}
```
niegowski, line 67: If I understand this correctly, it's supposed to detect list items with blocks inside. This would fail on such a structure:

```html
<ul>
  <li>UL List item 1</li>
  <li><strong>UL List item 2</strong></li>
</ul>
```
The second LI would be detected as DocumentList but it's not in legacy lists.

Also, the code above would not detect properly simple lists in document lists.

filipsobol: I pushed fixes to:

* properly detect if element is a document list (https://github.com/ckeditor/ckeditor5/pull/15042/commits/8c16e9eecef64094a869060ae27c8a9472bac3b6),
* avoid breaking simple lists when pasting paragraphs or plain text lists (https://github.com/ckeditor/ckeditor5/pull/15042/commits/906ee8060383c1a3723a4043de78d9ab1df1fa39).

packages/ckeditor5-list/src/documentlist/documentlistediting.ts
```diff
@@ -790,72 +788,73 @@ function modelChangePostFixer(
...
+
+			for ( const item of items ) {
+				const isListItem = isListItemBlock( item );
+
+				if ( refItem.is( 'element', 'listItem' ) && item.is( 'element', 'paragraph' ) ) {
+					/**
+					 * When paragraphs or a plain text list is pasted into a simple list, convert
+					 * the `<paragraphs>' to `<listItem>' to avoid breaking the target list.
+					 *
+					 * See https://github.com/ckeditor/ckeditor5/issues/13826
```
niegowski, line 848:
```suggestion
					 * See https://github.com/ckeditor/ckeditor5/issues/13826.
```

niegowski (changes requested):
Just a small bug while copying mixed lists:

PR #15050, opened 2023-09-22
Implemented `FocusCycler#forwardCycle` and `#backwardCycle` events

oleq:

Feature (ui): Implemented `FocusCycler#forwardCycle` and `#backwardCycle` events.
Internal (ui): Made `SearchResultsView` a focus-friendly container. Made `SearchInfoView` focusable.

### DIFF
--- a/packages/ckeditor5-ui/src/focuscycler.ts
+++ b/packages/ckeditor5-ui/src/focuscycler.ts
@@ -274,11 +274,9 @@ export default class FocusCycler extends EmitterMixin() {
 	 * @returns
 	 */
 	private _focus( view: FocusableView | null, direction: 1 | -1 ) {
-		if ( !view ) {
-			return;
+		if ( view ) {
+			view.focus( direction );
 		}
-
-		view.focus( direction );
 	}
 
 	/**
--- a/packages/ckeditor5-ui/src/focuscycler.ts
+++ b/packages/ckeditor5-ui/src/focuscycler.ts
@@ -140,6 +140,9 @@ export default class FocusCycler extends EmitterMixin() {
 				}
 			}
 		}
+
+		this.on<FocusCyclerForwardCycleEvent>( 'forwardCycle', () => this.focusFirst(), { priority: 'low' } );
+		this.on<FocusCyclerBackwardCycleEvent>( 'backwardCycle', () => this.focusLast(), { priority: 'low' } );
 	}
 
 	/**
@@ -149,7 +152,7 @@ export default class FocusCycler extends EmitterMixin() {
 	 * **Note**: Hidden views (e.g. with `display: none`) are ignored.
 	 */
 	public get first(): FocusableView | null {
-		return ( this.focusables.find( isPlainFocusableView ) || null ) as FocusableView | null;
+		return ( this.focusables.find( isFocusable ) || null ) as FocusableView | null;
 	}
 
 	/**
@@ -159,7 +162,7 @@ export default class FocusCycler extends EmitterMixin() {
 	 * **Note**: Hidden views (e.g. with `display: none`) are ignored.
 	 */
 	public get last(): FocusableView | null {
-		return ( this.focusables.filter( isPlainFocusableView ).slice( -1 )[ 0 ] || null ) as FocusableView | null;
+		return ( this.focusables.filter( isFocusable ).slice( -1 )[ 0 ] || null ) as FocusableView | null;
 	}
 
 	/**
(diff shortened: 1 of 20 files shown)


packages/ckeditor5-ui/src/focuscycler.ts
```diff
@@ -237,14 +246,34 @@ export default class FocusCycler {
...
+			return;
+		}
+
+		if ( isViewWithFocusableChildren( view ) ) {
+			if ( childrenFocusDirection === 1 ) {
+				view.focusFirst();
+			} else if ( childrenFocusDirection === -1 ) {
+				view.focusLast();
+			}
+		} else {
```
scofalik, line 276: What if I have a view that can be focused AND has focusable children? Shouldn't we let the view decide what it does upon being focused?

packages/ckeditor5-ui/src/focuscycler.ts
```diff
@@ -302,11 +339,40 @@ export interface FocusCyclerActions {
...
+};
+
 /**
  * Checks whether a view is focusable.
  *
  * @param view A view to be checked.
  */
-function isFocusable( view: View & { focus?: unknown } ) {
-	return !!( view.focus && isVisible( view.element ) );
+function isPlainFocusableView( view: View ): view is FocusableView {
```
scofalik, line 369: What is a "plain focusable view"? AFAICS, the implementation did not change. If this is to differentiate "focusable" from "focusable with focusable children", I wouldn't make a change. To me, it felt like maybe "plain focsuable view" is a focusable that is not a "focusable with focusable children". But that's not true. I might not get something, but at the first glance, the change and the new name is confusing.

packages/ckeditor5-ui/src/focuscycler.ts
```diff
@@ -228,7 +231,13 @@ export default class FocusCycler {
 	 * **Note**: Hidden views (e.g. with `display: none`) are ignored.
 	 */
 	public focusNext(): void {
-		this._focus( this.next );
+		const next = this.next;
+
+		this._focus( next, 1 );
+
+		if ( next == this.first ) {
+			this.fire<FocusCyclerForwardCycleEvent>( 'forwardCycle' );
```
scofalik, line 239: I am no expert in the UI space, so maybe you'll be able to explain it all to me, but...

AFAICS, the way it works is that after reaching the last focusable element and firing `focusNext()` I will first focus the first element and then fire `forwardCycle` (and possibly immediately focus something else).

This doesn't seem correct, it seems unexpected and I am afraid of side-effects (or hard to discover bugs).

Instead, I'd assume that if we reach the end of focus cycle, `forwardCycle` is fired (instead of straight focusing the view). Then, there's a default listener on low priority that focuses the first element. Finally, to chain focus cyclers, you would overwrite `forwardCycle` and stop the event. We could have a helper method for that, or have it somehow automated, if possible. Same for `backwardCycle`.

This would be also more in-line with how we do things in CKE5.

Alternatively, we could have something like this:

```javascript
this.fire( 'forwardCycle' );

if ( !somethingHappenedOnForwardCycle ) {
    this._focus( first, 1 );
}
```

That is, instead of having a default action on low priority listener, we could focus the first element if `forwardCycle` has not been handled. In this way, you would not need to stop `forwardCycle` event, which would make the handling a little bit easier. But I am afraid that checking if `fowardCycle` was handled in a custom way may be difficult :thinking:. So maybe that's overengineering. Especially since, as I mentioned, in most places in CKE5 we expect people to stop the event after handling it in a custom way. We could also just put it in the API docs for the event.

scofalik (changes requested):

oleq:
@scofalik I went with your suggestions and moved the decision-making process down to implementers of the `FocusableView` interface (and the `focus()` method). 
* `focus()` has an optional parameter that gives the view extra information about the direction of a traveling focus and it's up to the view to handle this information (or ignore it)
* I made the events cancellable, thus preventing unnecessary element focus in DOM if a custom logic takes over and focuses something else instead.
* I reverted the new interface for a view with focusable children. We don't need it (yet).

oleq:
FYI: Recent changes to this PR (mainly: `FocusCycler#focusNext()` will no longer re-focus the same view over and over again if there's just one) revealed in unit tests that the focus management in the `ColorInputView` is totally accidental (but working from the UX standpoint). This is probably due to the fact that the component underwent refactoring this year and nobody realized that. 

I addressed this issue and I took the liberty of using the new FC events in table forms to handle focus properly. Changes in https://github.com/ckeditor/ckeditor5/pull/15050/commits/5f9bc87e73967a546414dc295bc1303dd0d474c8.

Issue #5152, opened 2019-10-09
Handle image pasted/dropped from various sources (external websites, base64-encoded, etc.)


mjadobson:
When copy-pasting a local image into the editor, it will be uploaded as expected.

However, if you copy an image from another website and paste it into the editor, it will use the previous url as the source, rather than uploading it.

Is this the intended behaviour? This can sometimes cause issues with regards to hotlinking, dead links,  security, etc.

dirkjf:
@taocz, I tried exactly what you suggested: 

``` 
let content = dataTransfer.getData('text/html');
const elem = document.createElement("div");
elem.innerHTML = content;

let images = elem.getElementsByTagName('img');

(async() => {
    for (let img of images) {
        let response = await fetch(img.src);
        let blob = await response.blob();
        await editor.execute('imageUpload', {file: new File([blob], 'image', response)} )
    }
})();
``` 

For copying multiple images though only one is ever uploaded. Making the image names unique did not matter. Any ideas?

CanVeeD:
Any updates/solutions? Trying to upload a pasted image, searching for any solution

astro2049:
I implemented the lazy upload mentioned by [@taocz](https://github.com/taocz) with Axios, here are the additional logics to go through for a newly created *imageElement*:

1. Check if the image is on server, *return* if it is
2. Download the image from internet
3. Upload the image to server

```javascript
const additionalUploadImage = (imageElement) => {
    let url = imageElement.getAttribute('src')

    // return if the image is already on server
    if (url.includes(process.env.SERVER_ADDRESS)) {
    return
    }

    ;(async () => {
    let filename = url.split('/').pop().split('#')[0].split('?')[0] // extracts image's filename, see https://stackoverflow.com/a/36756650/15630163

    let response = await axios.get(url, {
        responseType: 'blob',
    })
    let blob = response.data

    let formData = new FormData()
    formData.append('upload', blob, filename) // form name and file name can be customized

    let uploadResult = await axios.post(
        process.env.SERVER_ADDRESS + '/files/upload', // route can be customized
        formData,
    )

    // replace the image's url if it's successfully uploaded
    if (uploadResult.status === 200) {
        imageElement._attrs.set('src', uploadResult.data.url)
    }
    })()
}
```

So, since these are lazy upload logics to run through, we'll execute these logics in the *for* loop that iterates through changed image elements in the *doc.on( 'change', ... )* listener.

```javascript
[...]

PR #15052, opened 2023-09-25
Add support for uploading images included in pasted HTML content

felixpackard:

Feature (image): Add support for uploading images included in pasted HTML content. Closes: #5152

### DIFF
--- a/packages/ckeditor5-image/src/autoimage.ts
+++ b/packages/ckeditor5-image/src/autoimage.ts
@@ -13,8 +13,10 @@ import { LivePosition, LiveRange } from 'ckeditor5/src/engine';
 import { Undo } from 'ckeditor5/src/undo';
 import { Delete } from 'ckeditor5/src/typing';
 import { global } from 'ckeditor5/src/utils';
+import { type FileLoader } from 'ckeditor5/src/upload';
 
 import ImageUtils from './imageutils';
+import { fetchImage } from './imageupload/utils';
 
 // Implements the pattern: http(s)://(www.)example.com/path/to/resource.ext?query=params&maybe=too.
 const IMAGE_URL_REGEXP = new RegExp( String( /^(http(s)?:\/\/)?[\w-]+\.[\w.~:/[\]@!$&'()*+,;=%-]+/.source +
@@ -151,6 +153,15 @@ export default class AutoImage extends Plugin {
 				return;
 			}
 
+			// If the FileRepository plugin is available, we should upload the image rather than hotlinking it.
+			let loader: FileLoader | null = null;
+
+			if ( editor.plugins.has( 'FileRepository' ) ) {
+				const fileRepository = editor.plugins.get( 'FileRepository' );
+				const imagePromise = fetchImage( src );
+				loader = fileRepository.createLoader( imagePromise );
+			}
+
 			editor.model.change( writer => {
 				this._timeoutId = null;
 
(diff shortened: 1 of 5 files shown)


Witoso:
Hey! Thanks for the contribution. Before jumping into reviewing the code in depth, it would be great to get some explanation of what was introduced. The [#5152](https://github.com/ckeditor/ckeditor5/issues/5152) has a very long discussion of different scenarios. 

If you could provide:

*   Which scenario are you trying to cover, and are most interested in?
*   What decisions have you made and why (e.g., `image.setAttribute( 'crossorigin', 'anonymous' );` looks interesting).
*   Info on the lack of configuration, as far as I can see, this is enabled by default, which could be problematic for some users, as it may lead to some automatic copyright abuses. Curious about your thoughts.
*   Fail scenarios, is there any path in which this won't work, what is a recovery path when an image is not downloadable?

felixpackard:
Hey @Witoso!

The main scenario we're interested in is avoiding hotlinking images. We are migrating to CKEditor from Froala, which had built in support for uploading images to a server, rather than hotlinking images, which is generally considered bad practice.

This PR adds support for:

- Pasting HTML that contains one or more  tags and automatically uploading them using the FileRepository. This includes images copied using the "Copy image" option in your browser, since that copies the image as both a file and an HTML  tag.
- Pasting the URL of an image and automatically uploading it using the FileRepository.

Some remote servers may include CORS headers that prevent the image from being downloaded, so the `crossorigin="anonymous"` attribute is added to attempt to circumvent that. This appears to work in my manual testing so far.

You're correct in thinking this is enabled by default. I think it would make sense for this to be the default functionality, due to hotlinking images being generally considered a bad practice as mentioned before. If it would be considered a breaking change, or if there are any other backwards compatibility concerns, I'd understand if enabling it via a config option was preferred.

As for copyright concerns, as far as I'm aware copyright is based on the domain the image is _displayed_ on rather than the domain it's _hosted_ on - but I could be wrong here.

The PR may require some additional error handling for images that can't be downloaded, as I haven't tested this scenario yet.

felixpackard:
> The PR may require some additional error handling for images that can't be downloaded, as I haven't tested this scenario yet.

After some more testing it looks like any issues fetching the image are handled by existing error handling logic for the loader, since I'm just passing the `fetchImage` Promise into the loader. The same error handling logic that's used if a local file can't be fetched for some reason will apply to remote images.

If an image can't be fetched, the `` element will be removed from the pasted content.

Witoso:
I think both act of hot linking, and the download action **could be** illegal (depending on country, jurisdiction, fair use, etc., not a legal advice :wink:). So I think we could get this argument out of the way.  
  
I think giving users an option is most likely a must-have, we need to discuss internally what's the default, etc. I will get back to you.

> Some remote servers may include CORS headers that prevent the image from being downloaded, so the `crossorigin="anonymous"` attribute is added to attempt to circumvent that.

Gotcha, most likely needs to be removed afterward when you replace the `src`, as it's no longer needed.

felixpackard:
> I think giving users an option is most likely a must-have, we need to discuss internally what's the default, etc. I will get back to you.

No worries, makes sense.

> Gotcha, most likely needs to be removed afterward when you replace the `src`, as it's no longer needed.

The attribute is only set on the `img` element that's created in order to download the image and convert it to a `Blob`, not the actual `img` element that's inserted into the editor content, so this shouldn't be a concern.

Witoso:
As we discussed it, we should keep by default the current behavior, and the configuration option should be added, I guess to the `AutoImage`, that will switch the hot linking to the upload. @felixpackard we will take it on ourselves.

PR #15118, opened 2023-10-03
Introduce Vale linter for documentation

AnnaTomanek:

Internal: Introduces Vale linter for documentation. Closes [cksource/ckeditor5-internal#3395](https://github.com/cksource/ckeditor5-internal/issues/3395).

---

### Additional information

_This PR adds Vale to CKEditor 5's internal tools installed with `yarn install`._

_It introduces a set of base style guides (Microsoft, proselint, Readability, write-good) and a custom CKEditor rules set. In the initial configuration Vale lints Markdown files only: all guides, the changelog, `CONTRIBUTING.md`, `LICENSE.md`, `README.md`, and `SECURITY.md` files._

_Use:_

*   _`yarn docs:lint` to lint all guides_
*   _`yarn docs:lint -d/--directory some-directory` to lint files that contain `some-directory` in their path_
*   _`yarn docs:lint "**/README.md" "packages/*/LICENSE.md"` to lint selected files using a custom glob pattern_
*   _`yarn docs:lint -s/--save` to save Vale results (errors, warnings and readability metrics) to a file_


Reinmar:
One thing that I'd change is moving the output from e.g. `scripts/vale/results/2023-10-25--10-34-47.json` to a directory that's not tracked by git. E.g. to `build/` which is ignored.

Reinmar:
Also, let's remove the initial vale results from the repo. We cannot add those results as with time the repo will grow.

I know we want to store the initial results somewhere, so it can be kept here for the time being. But then, once we figure out where to upload this data, let's remove this file: `scripts/vale/results/2023-10-24--11-47-01.json`.

package.json
```diff
@@ -113,19 +113,23 @@
     "@ckeditor/ckeditor5-slash-command": "40.0.0",
     "@ckeditor/ckeditor5-template": "40.0.0",
     "@ckeditor/ckeditor5-track-changes": "40.0.0",
+    "@ocular-d/vale-bin": "2.29.1",
```
Reinmar, line 116: Why is this file locked to the specific version while other dev deps use caret ranges here?

package.json
```diff
@@ -175,6 +179,8 @@
     "docs:api": "node ./scripts/docs/build-api-docs.js",
     "docs:serve": "http-server ./build/docs/",
     "docs:verify": "node ./scripts/web-crawler/index.js --docs",
+    "docs:lint": "node scripts/vale/vale.js",
+    "docs:vale": "vale",
```
Reinmar, line 183: Why do we have both scripts here?

Reinmar:
Other thank the 2 comments that I left and the proposal to change the output directory to something we don't track (not necessarily now, can be done in the future if it was intentionally set to something we track), I have no other comments and this looks good to me :+1:

package.json
przemyslaw-zan: `"docs:vale"` is used internally by `"docs:lint"`, as it otherwise cannot access the `vale` script. Additionally it allows executing `vale` without our wrapper, if anyone has the need for direct access.

package.json
przemyslaw-zan, line 116: It seems that only recent changes in this package are Vale version bumps, so it should be safe to follow them.

```suggestion
    "@ocular-d/vale-bin": "^2.29.1",
```

pomek (approved):

PR #15121, opened 2023-10-03
Make possible get function through get() method in config

### LINKED ISSUE #12835: It's impossible to get functions from config using nested names

1. Add a custom object to the editor config with functions (see the snippet below). 
2. Call `editor.config.get()` with a nested name to get such functions from the config.

```
// Example config:
custom: {
	fn: () => 'foo',
	promise: async () => 'bar',
	string: 'test',
	number: 5,
	obj: {
		foo: true,
		bar: false,
		baz: () => 'foobar'
	}
}


// Get config:
console.log( editor.config.get( 'custom.fn' ) );
console.log( editor.config.get( 'custom.promise' ) );
console.log( editor.config.get( 'custom.string' ) );
console.log( editor.config.get( 'custom.number' ) );
console.log( editor.config.get( 'custom.obj' ) );
console.log( editor.config.get( 'custom.obj.baz' ) );
```

### ✔️ Expected result

Calling [`config.get()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_utils_config-Config.html#function-get) with the nested name should always return the element from the config.

### ❌ Actual result

If the nested name points to a function it returns an empty object instead :
image

(console.logs used above are the same as provided in the reproduction steps)

---

### LINKED ISSUE #14804: Using callbacks as tokenUrl value in ckbox is not possible

1. Use CKEditor 5 with CKBox plugin.
2. Provide [tokenUrl](https://ckeditor.com/docs/ckeditor5/latest/api/module_ckbox_ckboxconfig-CKBoxConfig.html#member-tokenUrl) as a callback. 

Example:
```
ckbox: {
  tokenUrl: () => new Promise( ( r ) => { setTimeout( ( ) => r( 'token_string' ), 500 ) } )
}
```

### ✔️ Expected result

The provided callback is executed and the value from the resolved promise is used as a token.

### ❌ Actual result

The editor tries to fetch the token from `./[object%20Object]` url

### ❓ Possible solution

Could be caused by https://github.com/ckeditor/ckeditor5/issues/12835 as there is nested name used in `editor.config.get` in the plugin source code: 
https://github.com/ckeditor/ckeditor5/blob/a8b1e91ba6855507b9d703a7e55076a05b2159be/packages/ckeditor5-ckbox/src/ckboxediting.ts#L74

## 📃 Other details

* Not a regression

---

illia-stv:

Fix (utils): Config's `get()` method should be able return a function. 
Closes #14804.
Closes #12835.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-utils/src/config.ts
+++ b/packages/ckeditor5-utils/src/config.ts
@@ -250,6 +250,11 @@ export default class Config<Cfg> {
 			source = source[ part ];
 		}
 
+		// Return actual value if it's a function.
+		if ( source && typeof source[ name ] === 'function' ) {
+			return source[ name ];
+		}
+
 		// Always returns undefined for non existing configuration.
 		return source ? cloneConfig( source[ name ] ) : undefined;
 	}
--- a/packages/ckeditor5-utils/tests/config.js
+++ b/packages/ckeditor5-utils/tests/config.js
@@ -29,7 +29,8 @@ describe( 'Config', () => {
 					{ bar: 'a' },
 					{ bar: 'z' }
 				]
-			}
+			},
+			callback: () => null
 		} );
 	} );
 
@@ -363,6 +364,11 @@ describe( 'Config', () => {
 			expect( config.get( 'resize.icon.path' ) ).to.equal( 'xyz' );
 		} );
 
+		it( 'should return a function', () => {
+			expect( typeof config.get( 'callback' ) ).to.equal( 'function' );
+			expect( config.get( 'callback' )() ).to.equal( null );
+		} );
+
 		it( 'should retrieve an object of the configuration', () => {
 			const resize = config.get( 'resize' );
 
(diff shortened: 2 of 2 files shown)


packages/ckeditor5-utils/src/config.ts
```diff
@@ -250,6 +250,11 @@ export default class Config<Cfg> {
 			source = source[ part ];
 		}
 
+		// Return actual value if it's a function.
+		if ( source && typeof source[ name ] === 'function' ) {
+			return source[ name ];
+		}
```
niegowski, line 256: Maybe we could handle functions the same way as DOM node references (inside `cloneConfig()`)?

niegowski (changes requested):
I'd add a test that accesses function from nested config, for example, `config.get( 'options.callback' )`

niegowski (approved):

PR #15124, opened 2023-10-04
Fix(ckbox): Improve error handling.

### LINKED ISSUE #13062: CKBox(Editing) should warn if the library has not been loaded

While working on docs (configuring the build) I came across the following scenario:

* I loaded `CKBox` in the `plugins: []`.
* Added `'ckbox'` to the toolbar config.
* Then I got `toolbarview-item-unavailable {item: 'ckbox'} ` warning in the console. Weird, right?
* I couldn't understand why, though. I checked and `CKBoxUI` was loaded. 
* I started digging into the code and I found out `CKBoxUI` [will not register the button unless the command has been registered](https://github.com/ckeditor/ckeditor5/blob/cdaa1d2190d5afdfad0f264392c70cd50ab4c881/packages/ckeditor5-ckbox/src/ckboxui.js#L34-L37).
* OK, not the most common pattern across our codebase but it makes sense. What does not make sense is that `CKBoxEditing` is loaded but the command is not there 🤔 
* Then I searched the editing plugin and I found out it [will not register command unless there's config and the lib already loaded](https://github.com/ckeditor/ckeditor5/blob/cdaa1d2190d5afdfad0f264392c70cd50ab4c881/packages/ckeditor5-ckbox/src/ckboxediting.js#L48-L52).
* Finally I found out I forgot to load the lib `<script src="https://cdn.ckbox.io/CKBox/1.2.1/ckbox.js"></script>` 🤦 

It took me a while to debug this and I had access to the source code and knowledge about the architecture. For a developer who tries to integrate this feature and knows very little about the editor this is going to be a nightmare.

What we really need is two warnings:
* about a missing config
* about a missing lib

---

gorzelinski:

Fix(ckbox): Improve error handling. Closes #13062.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ckbox/src/ckboxediting.ts
+++ b/packages/ckeditor5-ckbox/src/ckboxediting.ts
@@ -77,12 +77,12 @@ export default class CKBoxEditing extends Plugin {
 
 		if ( !hasConfiguration ) {
 			/**
-			 * The configuration required by the ckbox plugin is missing.
+			 * The configuration required by the CKBox plugin is missing.
 			 * Follow the instructions provided in {@glink features/file-management/ckbox#configuration the CKBox feature documentation}
 			 * to initialize the plugin properly.
 			 *
 			 * @error ckbox-config-not-found
-			 * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the ckbox.
+			 * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the CKBox.
 			 */
 			logWarning( 'ckbox-config-not-found', editor.config );
 		}
--- a/packages/ckeditor5-ckbox/src/ckboxediting.ts
+++ b/packages/ckeditor5-ckbox/src/ckboxediting.ts
@@ -84,7 +84,7 @@ export default class CKBoxEditing extends Plugin {
 			 * @error ckbox-config-not-found
 			 * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the ckbox.
 			 */
-			logWarning( 'ckbox-config-not-found', editor );
+			logWarning( 'ckbox-config-not-found', editor.config );
 		}
 
 		if ( !hasConfiguration && !isLibraryLoaded ) {
--- a/packages/ckeditor5-ckbox/src/ckboxediting.ts
+++ b/packages/ckeditor5-ckbox/src/ckboxediting.ts
@@ -23,7 +23,7 @@ import {
 	type ViewElement,
 	type Writer
 } from 'ckeditor5/src/engine';
-import { CKEditorError, logError } from 'ckeditor5/src/utils';
+import { CKEditorError, logError, logWarning } from 'ckeditor5/src/utils';
 
 import type { CKBoxAssetDefinition } from './ckboxconfig';
 
@@ -64,6 +64,29 @@ export default class CKBoxEditing extends Plugin {
 
 		// Proceed with plugin initialization only when the integrator intentionally wants to use it, i.e. when the `config.ckbox` exists or
 		// the CKBox JavaScript library is loaded.
+		if ( !isLibraryLoaded ) {
+			/**
+			 * The CKBox library is not loaded.
+			 * Follow the instructions provided in {@glink features/file-management/ckbox#installation the CKBox feature documentation}
+			 * to load the library properly.
+			 *
+			 * @error ckbox-library-not-loaded
+			 */
+			logWarning( 'ckbox-library-not-loaded' );
+		}
+
+		if ( !hasConfiguration ) {
+			/**
+			 * The configuration required by the ckbox plugin is missing.
+			 * Follow the instructions provided in {@glink features/file-management/ckbox#configuration the CKBox feature documentation}
+			 * to initialize the plugin properly.
+			 *
+			 * @error ckbox-config-not-found
+			 * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the ckbox.
+			 */
+			logWarning( 'ckbox-config-not-found', editor );
+		}
+
 		if ( !hasConfiguration && !isLibraryLoaded ) {
 			return;
 		}
--- a/packages/ckeditor5-ckbox/tests/ckbox.js
+++ b/packages/ckeditor5-ckbox/tests/ckbox.js
@@ -3,7 +3,7 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
-/* global document */
+/* global document, window */
 
 import LinkEditing from '@ckeditor/ckeditor5-link/src/linkediting';
 import PictureEditing from '@ckeditor/ckeditor5-image/src/pictureediting';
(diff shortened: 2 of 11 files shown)


gorzelinski:
The current implementation works, but I'm not sure about it. I considered some alternatives.

```ts
if ( !isLibraryLoaded ) {
    /**
     * The CKBox library is not loaded.
     * Follow the instructions provided in {@glink features/file-management/ckbox#installation the CKBox feature documentation}
     * to load the library properly.
     *
     * @error ckbox-library-not-loaded
     */
    logWarning( 'ckbox-library-not-loaded' );
    return;
}

if ( !hasConfiguration ) {
    /**
     * The configuration required by the ckbox plugin is missing.
     * Follow the instructions provided in {@glink features/file-management/ckbox#configuration the CKBox feature documentation}
     * to initialize the plugin properly.
     *
     * @error ckbox-config-not-found
     * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the ckbox.
     */
    logWarning( 'ckbox-config-not-found', editor );
    return;
}
```

The code logs warnings correctly, but it's not equivalent to the original code. Tests failed as expected, but I tested it anyway.

---

```ts
if ( !isLibraryLoaded ) {
    /**
     * The CKBox library is not loaded.
     * Follow the instructions provided in {@glink features/file-management/ckbox#installation the CKBox feature documentation}
     * to load the library properly.
     *
     * @error ckbox-library-not-loaded
     */
    logWarning( 'ckbox-library-not-loaded' );

    if ( !hasConfiguration ) {
        /**
         * The configuration required by the ckbox plugin is missing.
         * Follow the instructions provided in {@glink features/file-management/ckbox#configuration the CKBox feature documentation}
         * to initialize the plugin properly.
         *
         * @error ckbox-config-not-found
         * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the ckbox.
         */
        logWarning( 'ckbox-config-not-found', editor );
        return;
    }
}
```

The code is equivalent to the original. Tests pass. But there is an edge case - nested warning depends on the outer if.

---

```ts
if ( !isLibraryLoaded ) {
    /**
     * The CKBox library is not loaded.
     * Follow the instructions provided in {@glink features/file-management/ckbox#installation the CKBox feature documentation}
     * to load the library properly.
     *
     * @error ckbox-library-not-loaded
     */
    logWarning( 'ckbox-library-not-loaded' );
}

if ( !hasConfiguration ) {
    /**
     * The configuration required by the ckbox plugin is missing.
     * Follow the instructions provided in {@glink features/file-management/ckbox#configuration the CKBox feature documentation}
     * to initialize the plugin properly.
     *
     * @error ckbox-config-not-found
     * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the ckbox.
     */
    logWarning( 'ckbox-config-not-found', editor );
}

if( !hasConfiguration && !isLibraryLoaded) {
    return;
}
```

Committed implementation works. Warnings are displayed. Tests pass. But it looks a little redundant - I don't know if that's the most elegant solution.

gorzelinski:
The last commit fixed issues in CKBox unit tests. I also saw warnings in build unit tests. I guess that builds use CKBox. So, I need to handle them as well.

gorzelinski:
I don't know what to do with failing unit tests, tbh. Added warnings cause [CI to crash](https://app.circleci.com/pipelines/github/ckeditor/ckeditor5/968/workflows/b36fbdfd-4fae-4dc2-b60c-68d1aab2af41/jobs/4900). It's because the [builds per se](https://github.com/ckeditor/ckeditor5/blob/ck/13062/packages/ckeditor5-build-balloon/src/ckeditor.ts) have `ckbox` in the list of plugins but no configuration. I tried to mock the config in the [unit tests](https://github.com/ckeditor/ckeditor5/blob/ck/13062/packages/ckeditor5-build-balloon/tests/ckeditor.js) but without results. I'm not sure if that's the right solution in the first place. If I understand correctly, CKBox won't work without config, and our builds don't have one. I could delete the CKBox plugin from the list of plugins - that would solve the issue, but we probably don't want that. I could also add an example config in our builds, but I don't know if we have something like this. Fake config doesn't work. Or I'm missing something. Long story short, currently, I see two solutions:

1. Deleting CKBox from the builds.
2. Adding an example CKBox config to the builds.
3. ?

gorzelinski:
@pszczesniak found a token used in cloud services tests. In the end, I reused it here. It solved the above issue with the CI. However, our builds still don't have an example config. I don't know if we want to change that.

packages/ckeditor5-ckbox/tests/ckbox.js
```diff
@@ -20,14 +20,17 @@ import CKBoxUI from '../src/ckboxui';
...
-	let editorElement, editor;
+	let editorElement, editor, originalCKBox;
 
 	testUtils.createSinonSandbox();
 
 	beforeEach( async () => {
 		editorElement = document.createElement( 'div' );
 		document.body.appendChild( editorElement );
 
+		originalCKBox = window.CKBox;
```
arkflpc, line 31: Why?

packages/ckeditor5-ckbox/tests/ckboxediting.js
```diff
@@ -232,13 +237,16 @@ describe( 'CKBoxEditing', () => {
 				theme: 'default',
 				tokenUrl: 'http://cs.example.com'
 			} );
+			expect( console.warn.callCount ).to.equal( 1 );
```
arkflpc, line 240: It would be nice, to check the warning text as well. I.e. if it contains `ckbox-config-not-found`.

packages/ckeditor5-ckbox/tests/ckboxediting.js
```diff
@@ -78,13 +78,16 @@ describe( 'CKBoxEditing', () => {
...
+		sinon.stub( console, 'warn' );
+
 		const editor = await createTestEditor( {
 			ckbox: {
 				tokenUrl: 'http://cs.example.com'
 			}
 		} );
 
 		expect( editor.commands.get( 'ckbox' ) ).to.be.undefined;
+		expect( console.warn.callCount ).to.equal( 1 );
```
arkflpc, line 90: It would be nice, to check the warning text as well. I.e. if it contains `ckbox-library-not-loaded`.

arkflpc (changes requested):

arkflpc:
Let's wait for https://github.com/ckeditor/ckeditor5/issues/13062#issuecomment-1759497032 and https://github.com/ckeditor/ckeditor5/issues/13062#issuecomment-1759515299 to be addressed.

packages/ckeditor5-ckbox/src/ckboxediting.ts
```diff
@@ -64,6 +64,29 @@ export default class CKBoxEditing extends Plugin {
...
+			 * to load the library properly.
+			 *
+			 * @error ckbox-library-not-loaded
+			 */
+			logWarning( 'ckbox-library-not-loaded' );
+		}
+
+		if ( !hasConfiguration ) {
+			/**
+			 * The configuration required by the ckbox plugin is missing.
```
pszczesniak, line 80:
```suggestion
			 * The configuration required by the CKBox plugin is missing.
```

packages/ckeditor5-ckbox/src/ckboxediting.ts
```diff
@@ -64,6 +64,29 @@ export default class CKBoxEditing extends Plugin {
...
+			 * Follow the instructions provided in {@glink features/file-management/ckbox#configuration the CKBox feature documentation}
+			 * to initialize the plugin properly.
+			 *
+			 * @error ckbox-config-not-found
+			 * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the ckbox.
```
pszczesniak, line 85:
```suggestion
			 * @param {module:ckbox/ckboxconfig~CKBoxConfig} config Configuration of the CKBox.
```

pszczesniak (changes requested):

PR #15131, opened 2023-10-04
CKBox multiple files insertion in paragraphs.

### LINKED ISSUE #15094: Fix multiple files insertion via CKBox
When we select in CKBox files, and add them, they become joined together:
image

### Scope
Add each file in a separate paragraph, this way it may be easier to create a list.

pszczesniak:

Other (ckbox): When multiple files are selected add each one of them in a separate paragraph. Closes #15094.

---

### Additional information


arkflpc:
I'm not sure if it should be `Fix`. Maybe `Other`?

packages/ckeditor5-ckbox/src/ckboxcommand.ts
```diff
@@ -299,6 +303,25 @@ export default class CKBoxCommand extends Command {
 		if ( selection.isCollapsed ) {
 			const selectionAttributes = toMap( selection.getAttributes() );
 			const textNode = writer.createText( linkName, selectionAttributes );
+
+			if ( writer && assetsCount > 1 ) {
```
arkflpc, line 307: When can `writer` be `null`? I'm surprised there is 100% for this PR.

arkflpc, line 307: I don't think, `_insertLink` is the proper place to check `assetsCount > 1`.
For instance, `_insertAssets` takes the parameter `isLastAsset` - it does not check itself if it is the last one. It could take `isSingleAsset` as well and pass it here.

packages/ckeditor5-ckbox/src/ckboxcommand.ts
```diff
@@ -203,17 +203,18 @@ export default class CKBoxCommand extends Command {
 				isImageAllowed: imageCommand.isEnabled,
 				isLinkAllowed: linkCommand.isEnabled
 			} );
+			const assetsCount = assetsToProcess.length;
```
arkflpc, line 206:
```suggestion

			const assetsCount = assetsToProcess.length;
```
It just looks odd.

arkflpc (changes requested):

packages/ckeditor5-ckbox/src/ckboxcommand.ts
pszczesniak: `writer` is passed only when asset is not an `image` :point\_right: https://github.com/ckeditor/ckeditor5/blob/3cbdb2e85e92b4392d70e1943a3efab72591e7b7/packages/ckeditor5-ckbox/src/ckboxcommand.ts#L261

pszczesniak: `writer` is passed only when asset is not an `image` :point\_right: https://github.com/ckeditor/ckeditor5/blob/3cbdb2e85e92b4392d70e1943a3efab72591e7b7/packages/ckeditor5-ckbox/src/ckboxcommand.ts#L261

filipsobol: In the `if` statement you linked, the `_insertImage` method is called if the asset is an image. Otherwise, the `_insertLink` method is called. These method have different signatures - the latter method accepts the `writer` object, while the former does not.

In both `_insertAsset` and `_insertLink`, the `writer` type indicates that it's a required parameter. If that's really the case, then there's no need to check if `writer` exists. Otherwise, the type of the `writer` should be updated to `writer?: Writer`.

pszczesniak: > In the `if` statement you linked, the `_insertImage` method is called if the asset is an image. Otherwise, the `_insertLink` method is called. These method have different signatures - the latter method accepts the `writer` object, while the former does not.

Yes, you have right! My bad, didn't spot it that there are called different functions :see\_no\_evil:

pszczesniak:
> I'm not sure if it should be `Fix`. Maybe `Other`?

Yes, `Other` fits perfectly in this case.

arkflpc (approved):

PR #15135, opened 2023-10-05
Remove some attributes in responsive images from ckbox

### LINKED ISSUE #15093: Fix the URL editing of the CKBox/Responsive image
Replacing the URL of the image, should replace the image itself. Looks like this was already planned but is not working:

https://github.com/ckeditor/ckeditor5/blob/257e0f0e02148985499bad6b709aa6e7780465e0/packages/ckeditor5-image/src/image/replaceimagesourcecommand.ts#L48-L49

### Scope

Updating the URL of the responsive image (`srcset`, `sources`), should change it to a normal image. We should also clean the `ckboxImageId` attribute.

illia-stv:

Fix (image): Should correctly replace the image while changing the URL of it. Closes #15093.

---

### Additional information


packages/ckeditor5-image/src/image/replaceimagesourcecommand.ts
```diff
@@ -47,6 +47,13 @@ export default class ReplaceImageSourceCommand extends Command {
 			writer.setAttribute( 'src', options.source, image );
 			writer.removeAttribute( 'srcset', image );
 			writer.removeAttribute( 'sizes', image );
+
+			/**
+			 * In case responsive images some attributes should be cleaned up.
+			 * Check: https://github.com/ckeditor/ckeditor5/issues/15093
+			 */
+			writer.removeAttribute( 'ckboxImageId', image );
+			writer.removeAttribute( 'sources', image );
```
pomek, line 56: Shouldn't the CKBox command listen to `ReplaceImageSourceCommand#execute()`? I do not think mixing logic between two different plugins is a good idea. The CKBox integration adds attributes, and it should take care of removing them. It would require more code, but I think it's worth it.

arkflpc: The question: does it have to be in the same block or not? If yes, that would complicate it much more.

pomek: Fair point.

packages/ckeditor5-image/src/image/replaceimagesourcecommand.ts
```diff
@@ -34,6 +43,15 @@ export default class ReplaceImageSourceCommand extends Command {
 		this.value = this.isEnabled ? element.getAttribute( 'src' ) as string : null;
 	}
 
+	/**
+	 * Register callback which will be executed after command execution.
+	 *
+	 * @param callback Callback which will be called after command execution.
+	 */
+	public registerImageCallback( callback: ( writer: Writer, image: Element ) => void ): void {
```
arkflpc, line 51: It sounds too generic. The method name should include the purpose of those callbacks.

packages/ckeditor5-image/src/image/replaceimagesourcecommand.ts
```diff
@@ -34,6 +43,15 @@ export default class ReplaceImageSourceCommand extends Command {
 		this.value = this.isEnabled ? element.getAttribute( 'src' ) as string : null;
 	}
 
+	/**
+	 * Register callback which will be executed after command execution.
```
arkflpc, line 47: It is not executed after command execution but during the execution. I would document the purpose of this (it cleans attributes not related to new source URL) and give some code example (i.e. how to remove `myId` attribute).

packages/ckeditor5-image/src/image/replaceimagesourcecommand.ts
```diff
@@ -22,6 +23,14 @@ import type ImageUtils from '../imageutils';
 export default class ReplaceImageSourceCommand extends Command {
 	declare public value: string | null;
 
+	declare private _imageCallbacks: Array<( writer: Writer, image: Element ) => void>;
+
+	constructor( editor: Editor ) {
+		super( editor );
+
+		this._imageCallbacks = [];
```
arkflpc, line 31: The name is too generic.

packages/ckeditor5-image/tests/image/replaceimagesourcecommand.js
```diff
@@ -44,6 +44,33 @@ describe( 'ReplaceImageSourceCommand', () => {
 
 			expect( element.getAttribute( 'src' ) ).to.equal( 'bar/foo.jpg' );
 		} );
+
+		it( 'should clean up some attributes in responsive image', () => {
+			setModelData( model, `[<imageBlock
+				src="foo/bar.jpg"
+				ckboxImageId="id"
```
arkflpc, line 51: This test data does not really belong to `ckeditor5-image`. I would test the specific case in `ckbox` and add here (in `image`) tests of the cleanup mechanism.

arkflpc (changes requested):

packages/ckeditor5-image/src/image/replaceimagesourcecommand.ts
```diff
@@ -43,10 +61,30 @@ export default class ReplaceImageSourceCommand extends Command {
...
-			writer.removeAttribute( 'sizes', image );
+
+			this._cleanupImage( writer, image );
 		} );
 	}
+
+	/**
+	 * Cleanup some image attributes.
+	 */
+	private _cleanupImage( writer: Writer, image: Element ) {
```
arkflpc, line 75: If this was `public` and `decorate`d, you wouldn't need to maintain `_imageCallbacks`.

arkflpc (approved):

packages/ckeditor5-ckbox/src/ckboxediting.ts
```diff
@@ -23,12 +23,13 @@ import {
 	type ViewElement,
 	type Writer
 } from 'ckeditor5/src/engine';
-import { CKEditorError, logError } from 'ckeditor5/src/utils';
+import { CKEditorError, type DecoratedMethodEvent, logError } from 'ckeditor5/src/utils';
```
filipsobol, line 26: `type` imports should be at the end :)

packages/ckeditor5-image/src/image/replaceimagesourcecommand.ts
```diff
@@ -43,10 +50,42 @@ export default class ReplaceImageSourceCommand extends Command {
...
+
 		this.editor.model.change( writer => {
 			writer.setAttribute( 'src', options.source, image );
-			writer.removeAttribute( 'srcset', image );
-			writer.removeAttribute( 'sizes', image );
+
+			this.cleanupImage( writer, image );
 		} );
+
+		imageUtils.setImageNaturalSizeAttributes( image );
```
niegowski, line 61: This should be inside the above change block to use a single undo step.

packages/ckeditor5-ckbox/src/ckboxediting.ts
```diff
@@ -62,8 +63,19 @@ export default class CKBoxEditing extends Plugin {
...
 
+		if ( replaceImageSourceCommand ) {
+			this.listenTo<DecoratedMethodEvent<ReplaceImageSourceCommand, 'cleanupImage'>>(
+				replaceImageSourceCommand,
+				'cleanupImage',
+				( _, [ writer, image ] ) => {
+					writer.removeAttribute( 'ckboxImageId', image );
+				}
+			);
+		}
```
pomek, line 77: Shouldn't this code be placed within the `_initConversion()` or `_initFixers()` to keep it closer to its source?

pomek (approved):

niegowski (approved):

PR #15139, opened 2023-10-05
Fix (ckeditor5-core): Added optional to PluginInterface.destroy

Marrikulus:

Fix (ckeditor5-core): Adds optional to the PluginInterface.destroy function.

---

### Additional information

* TextTransformation plugin does not implement destroy function and therefore is not assignable to variable of type PluginConstructor.

### DIFF
--- a/packages/ckeditor5-core/src/plugin.ts
+++ b/packages/ckeditor5-core/src/plugin.ts
@@ -211,7 +211,7 @@ export interface PluginInterface {
 	 *
 	 * **Note:** This method is optional. A plugin instance does not need to have it defined.
 	 */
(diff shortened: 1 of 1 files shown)


Witoso:
Thanks, @Marrikulus!

@arkflpc, could you take a look?

@vokiel, could you recheck the CLA?

arkflpc:
`TextTransformation` extends `Plugin` class, so it has `destroy()` method.
However, this method should be indeed optional. It's documented as such, and there are tests covering this scenario.

arkflpc (approved):

Witoso:
Let's merge this :blush:

PR #15176, opened 2023-10-12
Stop calling function `data.get()` when it's not needed while destroying the editor instance.

### LINKED ISSUE #15009: [Performance] Do not call data.get() when it's not used on destroy()
Flame graph of `editor.destroy()`:

image

## Findings

Because of this line:

https://github.com/ckeditor/ckeditor5/blob/fe95a500c3c3940b0096f196ed1eefe07887e4d4/packages/ckeditor5-core/src/editor/utils/elementapimixin.ts#L26

We do the entire data stringification regardless of whether we actually use it later on:

https://github.com/ckeditor/ckeditor5/blob/fe95a500c3c3940b0096f196ed1eefe07887e4d4/packages/ckeditor5-core/src/editor/utils/elementapimixin.ts#L49

Which, is 90%+ of cases because `config.updateSourceElementOnDestory` is `false` by default.

## Bonus

https://ckeditor.com/docs/ckeditor5/latest/api/module_editor-classic_classiceditor-ClassicEditor.html#function-updateSourceElement

The `data` param has no documentation. I guess it was added there because of the source mode or something (curious as well) and the use case is most likely very narrow but let's have it documented anyway.

---

pszczesniak:

Other(core): Should stop calling function `data.get()` when it's not needed while destroying the editor instance. Closes #15009 .

---

### Additional information

### DIFF
--- a/packages/ckeditor5-core/src/editor/utils/elementapimixin.ts
+++ b/packages/ckeditor5-core/src/editor/utils/elementapimixin.ts
@@ -52,7 +52,7 @@ export default function ElementApiMixin<Base extends Constructor<Editor>>( base:
 				return;
 			}
 
-			const dataToSet = data || this.data.get();
+			const dataToSet = typeof data === 'string' ? data : this.data.get();
 
 			setDataInElement( this.sourceElement, dataToSet );
 		}
--- a/packages/ckeditor5-core/src/editor/utils/elementapimixin.ts
+++ b/packages/ckeditor5-core/src/editor/utils/elementapimixin.ts
@@ -23,7 +23,7 @@ export default function ElementApiMixin<Base extends Constructor<Editor>>( base:
 	abstract class Mixin extends base implements ElementApi {
 		public sourceElement: HTMLElement | undefined;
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-core/src/editor/utils/elementapimixin.ts
```diff
@@ -44,15 +44,17 @@ export default function ElementApiMixin<Base extends Constructor<Editor>>( base:
...
+			// secure. This behavior could be changed by setting the `updateSourceElementOnDestroy`
 			// configuration option to `true`.
 			if ( !shouldUpdateSourceElement && !isSourceElementTextArea ) {
 				setDataInElement( this.sourceElement, '' );
 
 				return;
 			}
 
-			setDataInElement( this.sourceElement, data );
+			const dataToSet = data || this.data.get();
```
arkflpc, line 55: What if `data` is empty string and `this.data.get()` is not?

And please add test coverage for that case.

arkflpc (changes requested):

arkflpc (approved):

PR #15190, opened 2023-10-16
Implemented running vale with parameters

przemyslaw-zan:

Internal: `docs:lint` will now accept arbitrary amount of glob patterns as arguments. Added `--directory`/`-d` CLI option for `docs:lint`, that allows matching files that contain provided directory name anywhere in their path. This option overwrites default file list and globs provided as arguments.

---

### Additional information


pomek:
@AnnaTomanek, would you like to review it?

AnnaTomanek:
I played a bit with the command, any combination that I tried worked correctly :heart\_eyes: (files separated by a comma, directories, directories with `*`, etc.).

What I noticed, the "Matched N files." message returns weird results. Examples:

```shell
yarn docs:lint -f=docs/
...
Matched 1 files.
```

```shell
yarn docs:lint -f=docs/*

Matched 12 files.
```

And for both, the result returned by Vale is:

```shell
✖ 531 errors, 0 warnings and 0 suggestions in 127 files.
```

scripts/vale/vale.js
```diff
@@ -0,0 +1,115 @@
...
+	shell: true
+};
+
+const globOptions = {
+	cwd: CKEDITOR5_ROOT,
+	ignore: '**/node_modules/**'
+};
+
+const itemsToCheckInsidePackages = [
+	'README.MD',
```
AnnaTomanek, line 41:
```suggestion
	'README.md',
```

przemyslaw-zan:
> What I noticed, the "Matched N files." message returns weird results. Examples:

I've looked at that, and the reason is that in those cases, provided patterns resolved directories, and the message simply counted those results.

- `-f=docs/` resolved `docs`
- `-f=docs/*` resolved all 12 directories and files directly inside `docs`.

Vale then proceeds to lint entire content of given directory or directories.

I suppose that the behaviour is technically correct, just the message is not. Calculating number of files Vale will resolve would be complex. I could perhaps correct the message to say `Resolved N files and M directories.`, or just remove it altogether, as I'm not sure whether its needed.

scripts/vale/vale.js
```diff
@@ -0,0 +1,135 @@
...
+
+function main() {
+	const args = minimist( process.argv.slice( 2 ), minimistOptions );
+	const patterns = args.directory ?
+		[ `**/${ args.directory }/**/*.md` ] :
+		args.files.split( ',' ).filter( Boolean );
+	const valeArgs = [ 'run', 'docs:vale' ];
+
+	if ( patterns.length ) {
+		const files = patterns.flatMap( pattern => globSync( pattern, globOptions ) );
```
psmyrek, line 70: `glob` supports array of patterns:

```suggestion
		const files = globSync( patterns, globOptions );
```

scripts/vale/vale.js
```diff
@@ -0,0 +1,135 @@
...
+		process.exit( code );
+	} );
+}
+
+function getPathsFromPackages( packagesDirPath ) {
+	return itemsToCheckInsidePackages.flatMap( item => {
+		const globPattern = upath.join( packagesDirPath, '*', item );
+
+		return globSync( globPattern, globOptions );
+	} );
```
psmyrek, line 134: `glob` supports array of patterns:

```suggestion
	const globPatterns = itemsToCheckInsidePackages.map( item => upath.join( packagesDirPath, '*', item ) );

	return globSync( globPatterns, globOptions );
```

scripts/vale/vale.js
```diff
@@ -0,0 +1,135 @@
...
+		console.log( chalk.blue( `\nMatched ${ files.length } files.` ) );
+
+		valeArgs.push( ...files );
+	} else {
+		valeArgs.push( ...defaultPatterns );
+	}
+
+	console.log( chalk.blue( '\nExecuting vale...\n' ) );
+
+	const vale = spawn( 'yarn', valeArgs, spawnOptions );
```
psmyrek, line 87: This part is tricky. To correctly pass glob pattern to the script, you need to escape them before the shell processes them.

**Example**: Executing `yarn docs:lint -f *.md` causes that shell automatically expands the `*.md` pattern into `CHANGELOG.md CONTRIBUTING.md LICENSE.md README.md SECURITY.md` files separated by a space. Then, this list of files is passed to the script in the following way: `node scripts/vale/vale.js -f CHANGELOG.md CONTRIBUTING.md LICENSE.md README.md SECURITY.md`. So, this script get `CHANGELOG.md` as the glob pattern, so Vale gets only the `CHANGELOG.md` file to process.



To make it work as we expect, glob pattern must be escaped.

**Example**: Executing `yarn docs:lint -f "*.md"` is not expanded by the shell, but the glob pattern is passed "as is" to this script.



The problem is when the `"**"` pattern is correctly escaped: the `ENAMETOOLONG` error is thrown:



To fix that, probably `spawn()` has to be wrapped inside `try ... catch`.

scripts/vale/vale.js
```diff
@@ -0,0 +1,135 @@
...
+		if ( output.trim() === 'The command line is too long.' ) {
+			console.log( chalk.red( [
+				'Provided pattern matched too many files',
+				'and exceed the command character limit.',
+				'Please use a more specific pattern.',
+				''
+			].join( '\n' ) ) );
+
+			return;
+		}
```
psmyrek, line 115: In case there are too many files matched, maybe the files list could be split into smaller chunks?

pomek:
As we discussed on the standup today, I would eliminate the `-f` option. Instead, let me pass glob patterns or resolved expressions directly in the command:

```bash
yarn run docs:lint packages/*/docs docs/features
```

There is no comma or a `-f` modifier. Thanks to that, we do not have to care about the bash resolving mechanism before executing the command.

pomek:
Then, if `yarn run docs:lint` is called without additional arguments, we can assume that the entire project should be processed.

Also, I would keep the `--directory` (`-d`) modifier that allows checking particular directories in all packages and repos.

scripts/vale/vale.js
```diff
@@ -0,0 +1,164 @@
...
+
+'use strict';
+
+const chalk = require( 'chalk' );
+const upath = require( 'upath' );
+const minimist = require( 'minimist' );
+const { globSync } = require( 'glob' );
+const { spawn } = require( 'child_process' );
+
+const ARGS_CHAR_LIMIT = 4000;
```
psmyrek, line 16: I think we can safely increase the limit to 8000.

- MacOS limit is 262144.
   - https://serverfault.com/questions/69430/what-is-the-maximum-length-of-a-command-line-in-mac-os-x
- Linux limit is 131072.
   - https://unix.stackexchange.com/a/120652
- Windows limit is 8191.
   - https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation

scripts/vale/vale.js
```diff
@@ -0,0 +1,164 @@
...
+const globOptions = {
+	cwd: CKEDITOR5_ROOT,
+	ignore: '**/node_modules/**'
+};
+
+const defaultPatterns = [
+	'*.md',
+	'docs',
+	'packages/*/docs',
+	'packages/*/README.md',
```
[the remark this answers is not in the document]
przemyslaw-zan: @AnnaTomanek WDYT?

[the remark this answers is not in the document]
AnnaTomanek: Fine by me :+1:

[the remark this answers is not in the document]
AnnaTomanek: A quick note, let's make sure to also ignore the tests because they are full of `.md` files :)

psmyrek (approved):
LGTM 🎉

AnnaTomanek:
Thanks a lot @przemyslaw-zan @pomek @psmyrek :heart\_eyes\_cat: !

PR #15211, opened 2023-10-20
Docs: Drupal guide rewrite.

godai78:

Docs: Drupal integration guide SEO-oriented rewrite. Closes https://github.com/cksource/ckeditor5-internal/issues/3604

---

### Additional information


Witoso:
Some thoughts after reading the SEO guide:

*   Keywords, [quick search](https://ahrefs.com/keyword-generator) real-time editing drupal vs collaboration drupal, looks like a collaboration is used more (although the volume is pretty small). We could change the meta-title to real-time collaboration but the volume is so low in both that I don't think it will make a difference. 
*   URL, although it is a small factor in ranking, we could change it to something more like a title.
*   Theoretically, headings have quite a significant value when used together with keywords.
    *   I would add a better structure for:
        *   Track changes, Comments, Revision History as `h3`, after each 2-3 sentences of description
        *   other significant keywords that could be headings and small sections: templates, full screen, paste from office

Example:

```
## Real-time collaboration in Drupal with CKEditor 5
2-3 general sentences.

### Adding comments
2-3 sentences.

### Track changes

## Productivity improvements
2-3 general sentences.

### Templates

### Full screen

### Others
* slash commands

## Converters
2-3 general sentences.

### Import from Word or bullets
```

*   Let's resize those images, they are too large. I would tweak them to 70% of the current width of the section.

Witoso (changes requested):
As the comment in the PR.

godai78:
I don't find putting plenty of headers in a docs guide the best idea, actually. While I understand that SEO is a god, this still should be human-oriented stuff ;)

Witoso:
IMHO, headers are for people too, not just for bots :blush:   Well-structured headers can act as signposts that guide the reader through the content, making it easier to skim and find relevant information quickly.

I'm against sections like:

```
## Real-time collaboration in Drupal with CKEditor 5

### Comments in Drupal

### Track changes in Drupal
```

That beats the purpose and is too repeatable.

With a table of contents, headers can significantly improve the usability of a document, allowing users to jump to the section (or copy a link to it).

godai78:
K, I think this is a good compromise between what we do and what we could've wanted.

Witoso (approved):
LGTM

PR #15221, opened 2023-10-23
Added option to save vale data

przemyslaw-zan:

Internal: Added `-s`/`--save` option to `docs:lint` script that allows saving readability scores as well as error and warning counts from Vale.

---

### Additional information


scripts/vale/results/1698071908279.json
pomek, line 1: Could we update the file name to match: `YYYY-MM-DD-HH-MM`?

Also, to store some meta-data in the future, I would keep results in a nested object or even, an array.

```json5
{
	"results": [
		{
			"file": "...",
			/* rest of properties */
		}
	]
}
```

scripts/vale/vale.js
```diff
@@ -146,21 +247,58 @@ function runVale( files ) {
...
 				text: text
-					.replace( /^\$ vale .+\n/m, '' )
+					.replace( valeCommandPattern, '' )
 					.replace( valeFooterPattern, '' )
 			} );
 		} );
 	} );
 }
+
+function prepareConfigFiles() {
```
pomek, line 277: I can see that some steps of the entire process are async. Why couldn't it be consistent across the entire script? In other words, if we already use async code, I would substitute all sync calls in favor of async.

scripts/vale/vale.js
```diff
@@ -146,21 +247,58 @@ function runVale( files ) {
...
+	fs.writeFileSync( VALE_CONFIG_PATH, newValeConfigContent, 'utf-8' );
+
+	globSync( READABILITY_FILES_GLOB, globOptions )
+		.map( upath.toUnix )
+		.forEach( readabilityFilePath => {
+			originalFileContents[ readabilityFilePath ] = fs.readFileSync( readabilityFilePath, 'utf-8' );
+
+			const newReadabilityFileContent = originalFileContents[ readabilityFilePath ]
+				.replace( /(?<=\nmessage: ").+(?="\n)/, '%s' )
+				.replace( /(?<=\ncondition: ").+(?="\n)/, '> -999999' );
```
psmyrek, line 292: Maybe it would be better to use a library to parse YML files instead of regular expressions?

pomek (approved):
LGTM.

psmyrek (approved):
LGTM 🎉

AnnaTomanek:
Thank you so much :heart:. Now let's merge it and I cannot wait to bring the error count to 0 :muscle: and make our readability skyrocket :chart\_with\_upwards\_trend:!

PR #15229, opened 2023-10-24
Added support for verifying if all SVG files are optimized and enabled validator on CI

### LINKED ISSUE #15142: Add validator on CI to ensure all icons are already optimized

It happened several times that we accidentally broke some icons before the release, because the SVG optimizer removed too much content from the SVG file.

To prevent this, the following rules could be established:
1. Each newly added icon to the project must be added in an optimized form.
2. On CI we would then verify, whether all icons are already optimized. If not, the build would be marked as failed.
3. When preparing the release, we would no longer have to run the `clean-up-svg-icons` script anymore.

---

psmyrek:

Other: Added support for the `--verify-only` flag in the `yarn run clean-up-svg-icons` script that checks if all SVG files are optimized.

Internal: Added SVG files validator on CI. Closes #15142.

---

### Additional information

Currently all SVG files are optimized.

If we remove files added to exceptions, the `yarn clean-up-svg-icons --verify-only "packages/ckeditor5-ui/theme/**"` command will return the following results:

obraz

### DIFF
--- a/.circleci/template.yml
+++ b/.circleci/template.yml
@@ -24,7 +24,7 @@ commands:
               echo "Running full flow due to nightly build, despite detecting '[short flow]'."
               exit 0
             fi
-            
+
             if [[ "$COMMIT_MESSAGE" == *"[short flow]" ]]; then
               echo "Skipping, because the commit message contains '[short flow]'."
               circleci step halt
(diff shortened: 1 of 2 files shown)


martnpaneq (approved):
Works correctly:
image

pomek (approved):
LGTM.

PR #15285, opened 2023-10-31
Dialogs system implementation

### LINKED ISSUE #14973: Dialog/modal system for the CKEditor 5

CKEditor 5 lacks a dialog system that could be used to implement more complex and larger forms. This is something that CKEditor 4 had, and we want to gather interest for this feature.

---

oleq:

Feature (ui): Implemented the [`Dialog`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_dialog_dialog-Dialog.html) plugin that allows for displaying dialog windows in the UI of the editor. [Learn more](https://ckeditor.com/docs/ckeditor5/latest/framework/architecture/ui-library.html#dialogs-and-modals) about using dialogs. Closes #14973.
Feature (find-and-replace): The UI of the feature has been migrated to a dialog for a better user experience. See #14973.

Fix (ui): [`TextareaView`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_textarea_textareaview-TextareaView.html) component should update its size if its value has been changed while it was invisible. See #14973.
Other (ui): The [`FocusCycler#focusables`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusCycler.html#member-focusables) collection should only contain [`FocusableView`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusableView.html) instances. See #14973.
Other (ui): The [`#next`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusCycler.html#member-next) and [`#previous`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusCycler.html#member-previous) properties of a [`FocusCycler`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusCycler.html) instance should point to the same view if there is only one focusable view registered in the [`focusables`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusCycler.html#member-focusables) collection. See #14973.

BREAKING CHANGE (find-and-replace): From this release on, the UI of the Find and replace feature is displayed by default in a dialog instead of a dropdown. To bring the previous user experience back, you can use the [`config.findAndReplace.uiType`](https://ckeditor.com/docs/ckeditor5/latest/api/module_core_editor_editorconfig-EditorConfig.html#member-findAndReplace) configuration option. See #14973.

MINOR BREAKING CHANGE (ui): The `--ck-z-modal` CSS custom property has been renamed to `--ck-z-panel`. We recommend updating custom CSS and integrations that use this custom property to avoid presentation issues. See #14973.
MINOR BREAKING CHANGE (find-and-replace): The layout of the UI has been changed. Please keep in mind that customizations based on certain CSS selectors may not work anymore because of a different DOM structure in the UI. [Learn more](https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-41.html) about the scope of changes. See #14973.
MINOR BREAKING CHANGE (ui): The view collection (`focusables`) required by [`FocusCycler#constructor()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusCycler.html#function-constructor) must only contain views implementing the [`FocusableView`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusableView.html) interface. Failing to do so will result in a TypeScript error. If your custom code creates `FocusCycler` instances, please make sure that all views passed in `focusables` implement the `focus()` method. See #14973.
MINOR BREAKING CHANGE (ui): The font size of the `FormHeaderView` component has been increased. This change affects the look of [Find and replace](https://ckeditor.com/docs/ckeditor5/latest/features/find-and-replace.html) and [Table styling](https://ckeditor.com/docs/ckeditor5/latest/features/tables/tables-styling.html) features as well as custom user interfaces that use this component. See #14973.
MINOR BREAKING CHANGE (ui): The type of `AriaLiveAnnouncerPoliteness` has been changed (previously `enum`, now a constant `object`). See #14973.
MINOR BREAKING CHANGE (ui): The `#next` and `#previous` properties of a [`FocusCycler`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusCycler.html) will now point to the same view if there is only one focusable view (previously `null`). This change may affect integrations that use this helper to manage advanced focus navigation in dynamic UIs. See #14973.

---

### Additional information

Requires https://github.com/cksource/ckeditor5-commercial/pull/5720.


packages/ckeditor5-ui/src/focuscycler.ts
```diff
@@ -237,6 +237,9 @@ export default class FocusCycler extends EmitterMixin() {
 		const next = this.next;
 
 		if ( next && this.focusables.getIndex( next ) === this.current ) {
+			// TBD
+			this.fire<FocusCyclerForwardCycleEvent>( 'forwardCycle' );
```
oleq, line 241: I don't like this but this is necessary for views with just a single focusable child in their focus cycler to emit the right event and let the outside world know that they would cycle back the same focusable child (they won't, see `return` below) and maybe some parent wants to take over from here. I don't have a better idea, though.

scofalik (changes requested):
Partial review.

Dumluregn:
Features integration (besides the find and replace which is useful for testing) will be handled in the separate PR. The work done here was temporarily left behind as [https://github.com/ckeditor/ckeditor5/tree/ck/14973-dialogs-integration.](https://github.com/ckeditor/ckeditor5/tree/ck/14973-dialogs-integration.)

Commit reverting the changes: 2c9b3e1.

packages/ckeditor5-ui/src/focuscycler.ts
```diff
@@ -274,7 +280,8 @@ export default class FocusCycler extends EmitterMixin() {
 	 * @returns
 	 */
 	private _focus( view: FocusableView | null, direction: 1 | -1 ) {
-		if ( view ) {
+		// TBD
+		if ( view && this.focusTracker.focusedElement !== view.element ) {
```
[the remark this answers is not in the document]
Dumluregn: Yup, in particular it is used in the following scenario:

1.  There's only one focusable item that is already focused.
2.  `focusNext()` is executed.
3.  It recognizes there's nothing else to focus, so it returns (see [https://github.com/ckeditor/ckeditor5/pull/15285#discussion_r1395747487).](https://github.com/ckeditor/ckeditor5/pull/15285#discussion_r1395747487).) But it triggers the `forwardCycle` event.
4.  On `forwardCycle` event, `focusCycler` executes `focusFirst()`, which in turn executes `_focus()`. 
5.  So even though we cut the rope in one place, the element is still about to be focused. It would break the scenario discussed in the mentioned comment (when someone may want to move the focus to the outer cycler).

Unit test: `'does not refocus if there is only one focusable item'`.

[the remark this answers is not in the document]
scofalik: Would you add a short inline comment about that?

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,129 @@
...
+		} );
+
+		editor.ui.view.body.add( this.view );
+		editor.ui.focusTracker.add( this.view.element! );
+	}
+
+	/**
+	 * TODO
+	 */
+	public show( {
```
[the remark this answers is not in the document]
scofalik: After few days I still think it would be great to have an event we could listen to. Decorated method (or something that acts like it) would be best as you could possibly change the incoming parameters.

Both show and hide should fire events.

Additionally, the passed `onShow` and `onHide` should also be included in that event. I am still not sure whether we will want `onShow` and `onHide` happen before or after the regular code defined in `show()` and `hide()`. For now let's stick to what we have now. I expect that `onShow` will not be popular (I'd even say it is not needed) as you can do same stuff just before firing `dialog.show()`. As for `onHide` I assume it won't matter.

If `show()` and `hide()` are decorated (or if they fire events directly), then the outside integrator will be able to add callbacks on `'high'` or `'low'` priority, having more flexibility. Additionally, they will be able to change the parameters.

But, if there's a dialog opened, I'd like `hide` event to be called first and `show` event to be called then. And this prevents us from decorating `show()` as it will fire the events in a different order...

So maybe it should be like this:

```javascript
show( params ) {
	if ( dialog opened ) {
		this.hide();
	}
	
	this.fire( 'show', ... ); // Pass `params` here so it works like a decorated method.
}

hide() {
	this.fire( 'hide' );
}
```

The regular code from these methods will be called with default priority.

```javascript
this.on( 'show', ( evt, args ) => {
	this._show( ...args );
} );
```

We could fire `onShow` and `onHide` as an callback to `show` and `hide` events too:

```javascript
this.on( 'show', ( evt, args ) => {
	args[ 0 ].onShow( ... ); // Not sure if this is correct, writing from top of my head.
}, { priority: 'low' } );
```

This will give extra flexibility so you will be able to add custom callback from outside to be called between `_show()` and `onShow`.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,143 @@
...
```
scofalik, line 61: Different matter - one thing that I am missing in this whole dialog thing is a way to identify what dialog is/was opened. This is especially important if you want to plug in a custom behavior but only for certain dialogs. I propose introducing `id: string` property in `show()` and keep it in `Dialog` after a dialog was opened. It will be also passed to the events as one of parameters, so the integrator will be able to check it.

We could go even further and fire `show:<id>` and `hide:<id>` events.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,143 @@
...
+import '../../theme/components/dialog/dialog.css';
+
+/**
+ * TODO
+ */
+export default class Dialog extends Plugin {
+	/**
+	 * TODO
+	 */
+	public readonly view: DialogView;
```
scofalik, line 24: We should either have `isOpen` property or the `view` should be `DialogView | null` type and it should be created/destroyed each time the dialog is opened/closed.

packages/ckeditor5-ui/src/bindings/draggableviewmixin.ts
```diff
@@ -0,0 +1,175 @@
...
+			}
+
+			this._attachDragDurationListeners();
+
+			const x = ( domEvt instanceof TouchEvent ? domEvt.touches[ 0 ] : domEvt ).clientX;
+			const y = ( domEvt instanceof TouchEvent ? domEvt.touches[ 0 ] : domEvt ).clientY;
+
+			this._currentCoordinates = { x, y };
+
+			this._isDragging = true;
```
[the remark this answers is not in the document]
Dumluregn: It would be useful if we e.g. had some CSS bound to the dragging process, but I don't see such plans so it seems we can safely remove it and restore later if necessary.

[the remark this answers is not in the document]
scofalik: Hmm.. good call.. It could be useful for "outside world" or for some custom behaviors/styling. But in this case, it should be an observable public property, right? I am not sure what is our approach to mixed-in properties, i.e. if we specify them as private or public if we expect them to be used by the host class.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,143 @@
...
+
+		this.view.set( {
+			position,
+			isVisible: true,
+			className,
+			isDraggable: !isModal
+		} );
+
+		if ( title ) {
+			this.setTitle( title );
```
scofalik, line 96: I don't like that we have `Dialog#setTitle()` but we don't have `Dialog#setActionButtons()`. Since `setTitle()` is the only "hoisted" method from the `DialogView`, I propose getting rid of `Dialog#setTitle()` and just assume that we operate on the view directly if there's some custom stuff to do.

packages/ckeditor5-ui/src/dialog/dialogactionsview.ts
```diff
@@ -0,0 +1,159 @@
...
+import type Button from '../button/button';
+import ButtonView from '../button/buttonview';
+import View from '../view';
+import ViewCollection from '../viewcollection';
+import FocusCycler from '../focuscycler';
+
+/**
+ * TODO
+ */
+export default class DialogActionsView extends View {
```
scofalik, line 25: Please check which properties of this class must be public. E.g. I am not sure about `keystrokes`. But maybe this is required by some type/interface.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,635 @@
...
+	CURRENT_ROOT_NE = 'current-root-ne',
+	CURRENT_ROOT_NW = 'current-root-nw'
+}
+
+const toPx = toUnit( 'px' );
+
+/**
+ * TODO
+ */
+export default class DialogView extends DraggableViewMixin( View ) implements DraggableView {
```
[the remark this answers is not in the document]
scofalik: Same as with `DialogsActionsView`, please check which properties must be public.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,635 @@
...
+		this._updateFocusCycleableItems();
+
+		this.keystrokes.set( 'Esc', ( data, cancel ) => {
+			this.fire<DialogViewCloseEvent>( 'close' );
+			cancel();
+		} );
+
+		// Support for dragging the modal.
+		// TODO: Don't allow dragging beyond the edge of the viewport.
+		// TODO: Disable dragging when the mobile view is on.
```
scofalik, line 245: I am not sure how we can check "if the mobile view is on" right now + I think we decided that for now we don't address mobile -- as long as the UX does not degrade.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,635 @@
...
+	 * @inheritDoc
+	 */
+	constructor( locale: Locale, getCurrentDomRoot: () => HTMLElement, getViewportOffset: () => EditorUI[ 'viewportOffset' ] ) {
+		super( locale );
+
+		const bind = this.bindTemplate;
+
+		this.set( 'isVisible', false );
+		this.set( 'className', '' );
+		this.set( 'isDraggable', false );
```
scofalik, line 162: Rename to `isModal`.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,635 @@
...
+			actions: {
+				// Navigate form fields backwards using the Shift + Tab keystroke.
+				focusPrevious: 'shift + tab',
+
+				// Navigate form fields forwards using the Tab key.
+				focusNext: 'tab'
+			}
+		} );
+
+		this.setTemplate( {
```
scofalik, line 195: One thing that I don't dig is that the dialog is always inside this overlay wrapper that stretches on the whole viewport. I'd imagine that these are two seperate elements that are next to each other:

```
<div class="overlay"></div>
<div class="dialog">
	...
</div>
```

Since the overlay does not have background currently, I guess you can still interact with things that are below it. But the structure gets complicated for anyone that would like to somehow customize this solution. OTOH, I am not sure how you can easily do it anyway, so...

@oleq let's discuss it after you are back.

packages/ckeditor5-ui/src/dialog/dialogview.ts
scofalik, line 1: I feel there's a mess in this class with `children` and `parts`. They are both similar things/names and I was really confused where and why we use this or that. I feel that something got overcomplicated here. It could be due to two reasons:

1.  This view is not re-instantiated currently, it is created once and need to be re-set.
2.  Probably @oleq wanted to be able to "pass" multiple views to the dialog, so the dialog is: optional header + one or more "main" views + optional action buttons.

I'd like this to get simplified.

First, I am not sure if there is any gain in keeping a one view instance and re-setting it instead. If you need to do something custom with it, you will need to do it anyway the first time it is created. If you need to clear something custom after you did it, it is easier if the dialog instance is simply discarded. When it comes to performance, this is just one action, there's no performance difference.

Second, I think there's no need to allow to "compose" multiple views like this. If this need will arise, we can easily add it. There will be a workaround too -- you can always make a new, bare view that will include these views that you want to compose.

packages/ckeditor5-ui/src/bindings/draggableviewmixin.ts
```diff
@@ -0,0 +1,175 @@
...
+			if ( !this.isDraggable || !this._isHandleEventTarget( domEvt ) ) {
+				return;
+			}
+
+			this._attachDragDurationListeners();
+
+			const x = ( domEvt instanceof TouchEvent ? domEvt.touches[ 0 ] : domEvt ).clientX;
+			const y = ( domEvt instanceof TouchEvent ? domEvt.touches[ 0 ] : domEvt ).clientY;
+
+			this._currentCoordinates = { x, y };
```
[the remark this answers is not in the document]
Dumluregn: I did some wider renamings in this matter, including etc. `_lastDraggingCoordinates`, `deltaX` and `newX`. We'll see if you like them more.

packages/ckeditor5-ui/src/bindings/draggableviewmixin.ts
```diff
@@ -0,0 +1,175 @@
...
+				this.on( 'render', () => {
+					this._attachListeners();
+				} );
+			}
+		}
+
+		/**
+		 * TODO
+		 */
+		private _attachListeners() {
```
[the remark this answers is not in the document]
Dumluregn: It's done in [https://github.com/ckeditor/ckeditor5/blob/925589645eced920fbd4c358fe59c576cba1655d/packages/ckeditor5-ui/src/view.ts#L512](https://github.com/ckeditor/ckeditor5/blob/925589645eced920fbd4c358fe59c576cba1655d/packages/ckeditor5-ui/src/view.ts#L512) (`DraggableViewMixin` extends `View`).

packages/ckeditor5-ui/src/dialog/dialog.ts
Dumluregn: For now, I've settled for `_isOpen` private prop. It may turn out useful to turn it into a public observable in the future, but it wasn't necessary yet, so I kept is simple.

packages/ckeditor5-ui/src/dialog/dialogview.ts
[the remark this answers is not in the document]
Dumluregn: Regarding the first comment - I think it was made on purpose. `extends DraggableViewMixin( View )` part provides the dragging mechanisms (and the whole `View` core class) to the class and doesn't expect anything in return. OTOH the `implements DraggableView` indicates that there are some additional constraints (namely, some props/methods that need to be defined) in the `DialogView` class. And I think @oleq wanted to emphasise this.

I think it looks a bit strange right now as this interface is rather small:

[https://github.com/ckeditor/ckeditor5/pull/15285/files#diff-2434be37d8a50dd90e47ad6ed9e54254b4745e7e951b9968b03cb4c83e1ff093R160-R164](https://github.com/ckeditor/ckeditor5/pull/15285/files#diff-2434be37d8a50dd90e47ad6ed9e54254b4745e7e951b9968b03cb4c83e1ff093R160-R164)

with `isDraggable` to be gone, and `resetDrag()` not even implemented (btw I wonder why it works :thinking: ). But I think it may still make sense to keep these two constructs separate.

packages/ckeditor5-ui/src/bindings/draggableviewmixin.ts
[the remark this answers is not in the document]
Dumluregn: I think it does make sense to make it a public observable. I don't see why that could be an issue for now :thinking:

packages/ckeditor5-ui/src/dialog/dialog.ts
Dumluregn: I'm not sure if ids will make the dialogs traceable (AFAIU id would be some random hash). If we want to make it usable for the integrators, maybe the dialog name/title would be better than id? We could use the `title` parameter passed with the `show()` method.

Still, I'm not entirely sure it's necessary. Custom behaviour can already be implemented within `onShow()`/\`onHide()\` callbacks while defining the dialog, so maybe let's discuss how it could be used.

scofalik: By id I meant a `string` that is set by a given feature that uses dialog. So more like a name for the dialog, rather than UID.

So, for example, F&R feature would call `dialog.show( { id: 'findAndReplace', ... }`.

packages/ckeditor5-ui/src/dialog/dialog.ts
scofalik: Private prop does not give us what I wanted to achieve - I want external code to be able to check if there's any dialog open.

packages/ckeditor5-ui/src/bindings/draggableviewmixin.ts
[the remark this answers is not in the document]
scofalik: Okay, so let's make that a public observable.

packages/ckeditor5-ui/src/dialog/dialogview.ts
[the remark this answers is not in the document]
Dumluregn: As for the properties accessibility, I'm rather cautious about it due to [https://github.com/ckeditor/ckeditor5/pull/15285#discussion_r1400475486](https://github.com/ckeditor/ckeditor5/pull/15285#discussion_r1400475486) - access to stuff like `keystrokes` property may be expected for the integrator creating their own dialogs. And some properties that could be private are observables (e.g. `DialogView#_top`) so they have to stay public. So I did some changes but only those that seemed safe (both in `DialogView` and `DialogActionsView`).

packages/ckeditor5-ui/src/dialog/dialog.ts
scofalik: > Still, I'm not entirely sure it's necessary. Custom behaviour can already be implemented within `onShow()`/\`onHide()\` callbacks while defining the dialog, so maybe let's discuss how it could be used.

They cannot be implemented using `onShow` and `onHide` if you don't control the code that opens the dialog (and sets them). Which would be true for 3rd party customizations.

packages/ckeditor5-ui/src/dialog/dialogview.ts
Dumluregn: I think parts and children actually make sense in this class, but as you noticed there's a bit of cleaning up necessary.

The `children` prop should only contain a div with overlay, which contains the dialog, which contains the parts. We provide three possible parts: header, contents and action buttons, each with a convenient classes for styling. In `Dialog#show` (actually `Dialog#_show` after changes) we create those parts that were defined in the client plugin. 

Currently it's confusing because the contents are not added to the `contents` part, but directly as the dialog view children:

https://github.com/ckeditor/ckeditor5/pull/15285/files#diff-fc2f2fe7006390236bdc68ffd44aed385d08a50dcd0a986686aa69f80ecb56baR99-R108

but I believe @oleq planned to fix it by passing the `content` to `this.view.addContentPart();`, which would keep the `DialogView#children` clean and untouched.

So let me try to fix the current approach, maybe it will make more sense to you then.

As for the other point (not destroying the view upon hiding), this seems like a pattern in our codebase (I think the same is true for panels, balloons etc), so maybe it's better to be consistent here - I'd ask @oleq about the possible setbacks of reinstantiation once he's back.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,635 @@
...
+
+import cancelIcon from '@ckeditor/ckeditor5-core/theme/icons/cancel.svg';
+import type EditorUI from '../editorui/editorui';
+
+export enum DialogViewPosition {
+	SCREEN_CENTER = 'screen-center',
+	CURRENT_ROOT_CENTER = 'current-root-center',
+	CURRENT_ROOT_NE = 'current-root-ne',
+	CURRENT_ROOT_NW = 'current-root-nw'
+}
```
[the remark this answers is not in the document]
Dumluregn: I implemented the requested positions. As for the `ABOVE` and `BELOW`, there's already the partial logic responsible for keeping the dialog visible within `DialogView#moveTo()` method. It's missing just the bottom edge part. Maybe finishing it would be enough? See how it works for the `ABOVE` position in the first demo.

packages/ckeditor5-source-editing/src/sourceediting.ts
```diff
@@ -369,6 +370,15 @@ export default class SourceEditing extends Plugin {
 		// Checks, if the editor's editable belongs to the editor's DOM tree.
 		return editable && !editable.hasExternalElement;
 	}
+
+	/**
+	 * If any {@link module:ui/dialog/dialogview~DialogView editor dialog} is currently visible, hide it.
+	 */
+	private _hideVisibleDialog(): void {
+		if ( this.editor.plugins.has( 'Dialog' ) && this.editor.plugins.get( 'Dialog' ).isOpen ) {
```
arkflpc, line 378: See my other comment about type inference in `plugins.get`. TL;DR have typed variable.

packages/ckeditor5-ui/src/bindings/draggableviewmixin.ts
```diff
@@ -0,0 +1,181 @@
...
+ */
+
+import type View from '../view.js';
+import { type Constructor, type Mixed, global, type EventInfo } from '@ckeditor/ckeditor5-utils';
+
+/**
+ * A mixin that brings the possibility to observe dragging of the view element.
+ */
+export default function DraggableViewMixin<Base extends Constructor<View>>( view: Base ): Mixed<Base, DraggableView> {
+	abstract class DraggableMixin extends view {
```
arkflpc, line 17: Nitpick: if you added `implements DraggableView`, the compiler could check the contract. Especially that you have cast to `any` at the bottom.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,284 @@
...
+	 */
+	public view?: DialogView;
+
+	/**
+	 * The currently dialog plugin instance controlling the currently visible dialog view.
+	 */
+	public static visibleDialogPlugin?: Dialog;
+
+	/**
+	 * A flag indicating whether the dialog is currently visible.
```
arkflpc, line 36: I think `@observable` is missing. Probably in others `declare`s as well.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,284 @@
...
+	 * Initiates listeners for the `show` and `hide` events emitted by this plugin.
+	 */
+	private _initShowHideListeners() {
+		this.on<DialogShowEvent>( 'show', ( evt, args ) => {
+			this._show( args );
+		} );
+
+		// 'low' priority allows to add custom callback between `_show()` and `onShow()`.
+		this.on<DialogShowEvent>( 'show', ( evt, args ) => {
+			args.onShow?.( this );
```
arkflpc, line 75: https://github.com/cksource/ckeditor5-commercial/blob/master/adr/0018-avoid-using-certain-operators.md

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,284 @@
...
+		}, { priority: 'low' } );
+
+		this.on<DialogHideEvent>( 'hide', () => {
+			this._hide();
+		} );
+
+		// 'low' priority allows to add custom callback between `_hide()` and `onHide()`.
+		this.on<DialogHideEvent>( 'hide', () => {
+			this._onHide?.( this );
```
arkflpc, line 84: https://github.com/cksource/ckeditor5-commercial/blob/master/adr/0018-avoid-using-certain-operators.md

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,284 @@
...
+			}
+		} );
+	}
+
+	/**
+	 * Shows the dialog. If another dialog is currently visible, it will be hidden.
+	 * This method is decorated to enable interacting on the `show` event.
+	 */
+	public show( dialogDefinition: DialogDefinition ): void {
+		Dialog.visibleDialogPlugin?.hide();
```
arkflpc, line 136: https://github.com/cksource/ckeditor5-commercial/blob/master/adr/0018-avoid-using-certain-operators.md

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,699 @@
...
+import DialogActionsView, { type DialogActionButtonDefinition } from './dialogactionsview.js';
+import DialogContentView from './dialogcontentview.js';
+import type EditorUI from '../editorui/editorui.js';
+
+import '../../theme/components/dialog/dialog.css';
+// @if CK_DEBUG_DIALOG // const RectDrawer = require( '@ckeditor/ckeditor5-utils/tests/_utils/rectdrawer' ).default;
+
+import cancelIcon from '@ckeditor/ckeditor5-core/theme/icons/cancel.svg';
+
+export enum DialogViewPosition {
```
arkflpc, line 40: See my other concern about `enum` type.

packages/ckeditor5-ui/src/focuscycler.ts
```diff
@@ -256,6 +262,9 @@ export default class FocusCycler extends EmitterMixin() {
 		const previous = this.previous;
 
 		if ( previous && this.focusables.getIndex( previous ) === this.current ) {
+			// TBD
```
arkflpc, line 265: Added comment so you don't forget about TBD.

arkflpc (changes requested):

packages/ckeditor5-editor-classic/src/classiceditorui.ts
```diff
@@ -208,5 +215,36 @@ export default class ClassicEditorUI extends EditorUI {
...
+		if ( !this.editor.plugins.has( 'Dialog' ) ) {
+			return;
+		}
+
+		const stickyPanel = this.view.stickyPanel;
+
+		this.editor.plugins.get( 'Dialog' ).on( 'show', () => {
+			const dialogView = this.editor.plugins.get( 'Dialog' ).view!;
+
+			dialogView.on<DialogViewMoveToEvent>( 'moveTo', ( evt, data ) => {
```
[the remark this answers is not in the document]
Dumluregn: The view is re-created, so we need to re-attach the listeners every time.

packages/ckeditor5-find-and-replace/src/findandreplaceconfig.ts
```diff
@@ -0,0 +1,39 @@
...
+	/**
+	 * The type of the find and replace UI opened by the `'findAndReplace'` button registered in the
+	 * editor's {@link module:ui/componentfactory~ComponentFactory component factory}
+	 *
+	 * The default value is `'dialog'`.
+	 */
+	uiType?: FindAndReplaceUIType;
+}
+
+export enum FindAndReplaceUIType {
```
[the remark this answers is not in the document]
Dumluregn: Done at [ck/14973-dialogs-docs](https://github.com/ckeditor/ckeditor5/tree/ck/14973-dialogs-docs) in a7448f1 - we ended up with the using `object as const`.

packages/ckeditor5-ui/src/dialog/dialogview.ts
Dumluregn: Done at [ck/14973-dialogs-docs](https://github.com/ckeditor/ckeditor5/tree/ck/14973-dialogs-docs) in a7448f1 - we ended up with the using `object as const`.

packages/ckeditor5-ui/src/textarea/textareaview.ts
```diff
@@ -107,12 +124,44 @@ export default class TextareaView extends InputBase<HTMLTextAreaElement> {
...
+
+				this._updateAutoGrowHeight();
+				this.fire<TextareaViewUpdateEvent>( 'update' );
 			} );
 		} );
+
+		// This logic handles cases when textarea has been hidden while changes were made to its content but they could not
+		// be reflected to its height (auto-grown) because it was invisible or detached from DOM. The auto-grow logic
+		// gets executed as soon as the textarea becomes visible again.
+		this._resizeObserver = new ResizeObserver( this.element!, evt => {
```
scofalik, line 141: I tried to figure out why we ended up with this solution and why we need `ResizeObserver` for this. It is used only for the case described in the comment, which made me even more confusing.

In principle, what we want to do is: when the textarea becomes visible, we want to auto-resize it (as it's value may have changed while it was not visible).

But why we need `ResizeObserver` for that? AFAIR -- we don't have a precise way to observe when given element becomes visible/hidden if it is done outside of editor's control. However, I assume that `ResizeObserver` detects that, because when something gets hidden/visible its size becomes 0/non-0.

Is this correct?

If yes, then please change the comment so it describes what we want to achieve in simpler words and why we need `ResizeObserver` for that.

scofalik (changes requested):

packages/ckeditor5-find-and-replace/src/findandreplaceui.ts
```diff
@@ -52,79 +68,168 @@ export default class FindAndReplaceUI extends Plugin {
...
+		return buttonView;
 	}
 
 	/**
 	 * Sets up the form view for the find and replace.
 	 *
 	 * @param formView A related form view.
 	 */
-	private _setupFormView( formView: FindAndReplaceFormView ) {
+	private _createFormView(): FindAndReplaceFormView & ViewWithCssTransitionDisabler {
```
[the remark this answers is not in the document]
Dumluregn: The `formView` is only created once, both for the [dialog](https://github.com/ckeditor/ckeditor5/blob/ddc72262642b6519eb054fe5c363a9c70e00ed1d/packages/ckeditor5-find-and-replace/src/findandreplaceui.ts#L191) and for the [dropdown](https://github.com/ckeditor/ckeditor5/blob/ddc72262642b6519eb054fe5c363a9c70e00ed1d/packages/ckeditor5-find-and-replace/src/findandreplaceui.ts#L129). 

It isn't destroyed when form is hidden to keep the state of inputs etc. See the [`FindAndReplaceFormView#reset`](https://github.com/ckeditor/ckeditor5/blob/6e895281e33e7f9fbccd514b59129c3849bcd28a/packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.ts#L379-L382) method docs for details.

packages/ckeditor5-ui/lang/contexts.json
```diff
@@ -28,5 +28,6 @@
 	"HEX": "Label of an input field for typing colors in the HEX color format.",
 	"Accept": "Label of the button closing the color picker and confirming the changes done in the color selector component.",
 	"No results found": "The main text of the message shown to the user when given query does not match any results.",
-	"No searchable items": "The main text of the message shown to the user when no results are available."
+	"No searchable items": "The main text of the message shown to the user when no results are available.",
+	"Editor dialog": "A default label of a dialog window displayed on top the editor."
```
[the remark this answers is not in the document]
Dumluregn: It's a context to get more accurate translations, and the `Editor dialog` is a label to be read by screen readers. Not sure which part would you like to give up 🤔

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,286 @@
...
+		model.document.on<DocumentChangeEvent>( 'change:data', () => {
+			if ( !this.view ) {
+				return;
+			}
+
+			const changedRoots = model.document.differ.getChangedRoots();
+
+			for ( const changes of changedRoots ) {
+				if ( changes.state ) {
+					this.view.updatePosition();
```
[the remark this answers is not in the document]
Dumluregn: If we detach more than one root (e.g. all of them) in the same change block, we can't stop at the first changed one.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,723 @@
...
+				// @if CK_DEBUG_DIALOG // }
+
+				if ( domRootRect ) {
+					const leftCoordinate = this.locale!.contentLanguageDirection === 'ltr' ?
+						domRootRect.right - dialogRect.width - defaultOffset :
+						domRootRect.left + defaultOffset;
+
+					this.moveTo( leftCoordinate, domRootRect.top + defaultOffset );
+				} else {
+					this._moveOffScreen();
```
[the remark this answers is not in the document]
Dumluregn: It's consistent with other UI parts like dropdowns or balloons, but I don't really know if there's a reason for it - cc @oleq

packages/ckeditor5-find-and-replace/src/findandreplaceui.ts
```diff
@@ -52,79 +68,168 @@ export default class FindAndReplaceUI extends Plugin {
...
-			// Dropdown should be disabled when in source editing mode. See #10001.
-			const findCommand: FindCommand = editor.commands.get( 'find' )!;
-			dropdown.bind( 'isEnabled' ).to( findCommand );
+			if ( isUiUsingDropdown ) {
+				view = this._createDropdown();
 
-			dropdown.once( 'change:isOpen', () => {
-				this.formView = new ( CssTransitionDisablerMixin( FindAndReplaceFormView ) )( editor.locale );
+				// Button should be disabled when in source editing mode. See #10001.
+				view.bind( 'isEnabled' ).to( findCommand );
```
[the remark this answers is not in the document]
Dumluregn: Unfortunately, not according to TS:



IIRC we spent a while trying to solve this with @oleq and didn't come to the satisfying solution, hence we left this code duplicated.

packages/ckeditor5-find-and-replace/src/findandreplaceui.ts
```diff
@@ -52,79 +68,168 @@ export default class FindAndReplaceUI extends Plugin {
...
+			}
 
-				this._setupFormView( this.formView );
-			} );
+			editor.keystrokes.set( 'Ctrl+F', ( data, cancelEvent ) => {
+				if ( !findCommand.isEnabled ) {
+					return;
+				}
+
+				if ( view instanceof DropdownView ) {
```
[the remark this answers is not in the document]
Dumluregn: Also here TS objects: `if ( isUiUsingDropdown ) {` doesn't narrow the type, so there are still two incompatible types:



Attempts to fix it look worse than the current code.

packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/dialog/dialog.css
```diff
@@ -0,0 +1,42 @@
...
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+@import "../../../mixins/_rounded.css";
+@import "../../../mixins/_shadow.css";
+
+:root {
+	--ck-dialog-overlay-background-color: hsla( 0, 0%, 0%, .5 );
+	--ck-dialog-drop-shadow: 0px 0px 6px 2px hsl(0deg 0% 0% / 15%);
+	--ck-dialog-max-width: min(90vw, 800px);
```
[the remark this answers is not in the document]
Dumluregn: cc @oleq

packages/ckeditor5-ui/lang/contexts.json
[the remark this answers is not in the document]
scofalik: Yes, I know what it :).

Honestly, I didn't check where this translation is used. From the context description, I thought this is a default value for the dialog header. And I thought that we don't need it. If it is actually used somewhere, then OK.

packages/ckeditor5-find-and-replace/src/findandreplaceui.ts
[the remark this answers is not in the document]
scofalik: OMG...

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,723 @@
...
+		const viewportRect = this._getViewportRect();
+		const defaultOffset = DialogView.defaultOffset;
+
+		// @if CK_DEBUG_DIALOG // RectDrawer.clear();
+		// @if CK_DEBUG_DIALOG // RectDrawer.draw( viewportRect, { outlineColor: 'blue' }, 'Viewport' );
+
+		switch ( configuredPosition ) {
+			case DialogViewPosition.EDITOR_TOP_SIDE: {
+				const domRootRect = this._getVisibleDomRootRect( viewportRect );
+				const dialogRect = this._getDialogRect();
```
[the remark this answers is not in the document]
scofalik: Happy that TS does not object here, lol.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -28,7 +28,12 @@ export default class Dialog extends Plugin {
 	public view?: DialogView;
 
 	/**
-	 * The currently dialog plugin instance controlling the currently visible dialog view.
+	 * The dialog plugin instance controlling the currently visible dialog view.
+	 *
+	 * Only one dialog can be visible at once, even if there are many editor instances on the page.
+	 * If one editor wants to show a dialog, it should first check if there is no other visible dialog already.
+	 * But only the plugin that showed the dialog should be able to hide it
+	 * as it stores the {@link #_onHide()} callback and the proper editor reference.
```
scofalik, line 38: Ummm... I am not sure what you meant here.

I can call `Dialog#hide()` whenever I want, from any code, if I feel need to close the dialog. So either this comment is unnecessary, or is not clear enough.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -0,0 +1,354 @@
...
+	 */
+	public view?: DialogView;
+
+	/**
+	 * The dialog plugin instance controlling the currently visible dialog view.
+	 *
+	 * Only one dialog can be visible at once, even if there are many editor instances on the page.
+	 * If one editor wants to show a dialog, it should first check if there is no other visible dialog already.
+	 * But only the plugin that showed the dialog should be able to hide it
+	 * as it stores the {@link #_onHide()} callback and the proper editor reference.
```
scofalik, line 38:
```suggestion
	/**
	 * The `Dialog` plugin instance which most recently showed the dialog.
	 *
	 * Only one dialog can be visible at once, even if there are many editor instances on the page.
	 * If an editor wants to show a dialog, it should first hide the dialog that is already opened.
	 * But only the `Dialog` instance that showed the dialog is able able to properly hide it.
	 * This is why we need to store it in a globally available space (static property).
	 * This way every `Dialog` plugin in every editor is able to correctly close any open dialog window.
```

scofalik: As I was re-writing this docs, I had three observations:

*   This property should be private, it is an implementation detail and there should be no need for external code to reach for `visibleDialogPlugin`.
*   The name is not the best but I don't have a better proposal for now and since it is private, I wouldn't spend much time on it.
*   And most important: `Dialog#hide()` should also use `Dialog.visibleDialogPlugin`. This way I can easily close all dialogs without thinking how many editors I have on page. Honestly, using `Dialog#hide()` would be incorrect if it just operated on "itself". Especially in our own plugins, where we don't know in what editor configuration they will be used. We would have to always go for `Dialog.visibleDialogPlugin.hide()`. So, let's have it baked in.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -88,7 +88,9 @@ export default class Dialog extends Plugin {
 		}, { priority: 'low' } );
 
 		this.on<DialogHideEvent>( 'hide', () => {
-			this._hide();
+			if ( Dialog._visibleDialogPlugin ) {
+				Dialog._visibleDialogPlugin._hide();
```
scofalik, line 92: AFAIU, the way currently implemented it leads to `hide` event being fired for _this_ `Dialog` instance, while _other_ `Dialog` instance is actually hidden. This may lead to errors if someone does something on hide.

Instead, `Dialog#hide()` should cause `hide` event be fired for `_visibleDialogPlugin` instance (same for `Dialog#show()` of course -- and right now this is not consistent, AFAICS).

packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
```diff
@@ -98,7 +99,7 @@ export default class ListPropertiesView extends View {
 	/**
 	 * A collection of views that can be focused in the properties view.
 	 */
-	public readonly focusables: ViewCollection = new ViewCollection();
+	public readonly focusables: ViewCollection<FocusableView> = new ViewCollection();
```
niegowski, line 102: FYI this could be simplified (here and in other places):
```suggestion
	public readonly focusables = new ViewCollection<FocusableView>();
```

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,723 @@
...
+	 * @observable
+	 */
+	declare public className: string | undefined;
+
+	/**
+	 * The position of the dialog view.
+	 *
+	 * @observable
+	 */
+	declare public position: DialogViewPosition;
```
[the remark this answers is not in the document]
Dumluregn: It does, otherwise it won't be possible to change the default position of e.g. AI Assistant in `show` event listeners.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -0,0 +1,723 @@
...
+	 */
+	declare public isVisible: boolean;
+
+	/**
+	 * A flag indicating if dialog is transparent. It is used to prevent the dialog from being visible
+	 * before its position is calculated.
+	 *
+	 * @observable
+	 */
+	declare public isTransparent: boolean;
```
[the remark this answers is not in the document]
Dumluregn: Both props need to be observable as they are bound with CSS classes. And we can't have a private observable. I marked them as `@internal` and added the `_` at the beginning to show they should not be treated as public ones. Same treatment have already `_top` and `_left` props.

packages/ckeditor5-ui/src/textarea/textareaview.ts
Dumluregn: For the context - the root cause for the whole change was to simplify the code in AI Assistant (https://github.com/cksource/ckeditor5-commercial/pull/5832#discussion_r1426461711). We didn't want to make people use the additional listener to reset their components before dialog is hidden, and `onHide` effectively works as `afterHide`. Hence we're making the `Textarea` responsible for detecting when it can resize itself.

I added more elaborate comment explaining the generic issue.

PR #15290, opened 2023-10-31
Return `undefined` to `Schema#addChildCheck()` calls

Kocal:

Docs: Return `undefined` to `Schema#addChildCheck()` calls

Other: Return `undefined` to `Schema#addChildCheck()` calls

---

### Additional information

Hi everyone, sorry maybe I should have opened an issue first, but I thought this will be faster by directly opening the PR.

So, I was following the [Implementing a Block Widget tutorial](https://ckeditor.com/docs/ckeditor5/latest/tutorials/widgets/implementing-a-block-widget.html) and I got stuck on an incomprehensible issue (sorry I don't remember it, I'm not on my professional PC, I will edit this message tomorrow) and I was completely lost about what was happening.

After a lot of debbuging, trying many things, I found the problem was caused by my `schema.addChildCheck()` call.

I use ESLint, with the [consistent-return](https://eslint.org/docs/latest/rules/consistent-return) rule, and it was screaming that `return` calls were not consistent: 
```js
schema.addChildCheck( ( context, childDefinition ) => {
    if ( /* ... */ ) {
        return false;
    }

    // ESLint scream about no `return` here
} );
```

So I've updated the code to add a `return true` after the `if`, I was like "since we can return `false`, maybe we should return `true` otherwise? :shrug:", but that's what caused my downfall 😬 (because I didn't read the method description)

After many tries, I've found that using `return undefined` resolved my issue and makes ESLint happy.

I really think this PR can be a nice improvment, at least for the documentation, I'm sure it will save a lot of time and headache for others.

WDYT? Thanks :)

### DIFF
--- a/docs/_snippets/framework/tutorials/block-widget.js
+++ b/docs/_snippets/framework/tutorials/block-widget.js
@@ -108,6 +108,8 @@ class SimpleBoxEditing extends Plugin {
 			if ( context.endsWith( 'simpleBoxDescription' ) && childDefinition.name == 'simpleBox' ) {
 				return false;
 			}
+
+			return undefined;
 		} );
 	}
 
--- a/docs/examples/how-tos.md
+++ b/docs/examples/how-tos.md
@@ -367,6 +367,8 @@ class Forms extends Plugin {
 			) {
 				return false;
 			}
+
+			return undefined;
 		} );
 
 		// Data upcast. Convert a single element loaded by the editor to a structure of model elements.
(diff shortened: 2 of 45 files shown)


arkflpc:
Thank you for your contribution and the details you've shared. While I understand the challenges presented by ESLint rules, it's important to recognize that the CKEditor 5 framework adheres to its own coding standards. After careful consideration, we have decided that the current code aligns more closely with the framework's overall style and approach, and we will maintain it as is.

However, your insights regarding the return type in the `addChildCheck` method (and `addAttributeCheck` as well) are indeed valuable. A more specific callback type than `unknown` could enhance clarity and the developer experience. I am going to address this through a separate pull request.

PR #15331, opened 2023-11-08
PoC of new installation method

filipsobol:

Type: Message. Closes #000.

---

### Additional information


packages/ckeditor5-engine/src/model/node.ts
```diff
@@ -17,9 +17,6 @@ import type Element from './element';
 
 import { CKEditorError, compareArrays, toMap } from '@ckeditor/ckeditor5-utils';
 
-// To check if component is loaded more than once.
-import '@ckeditor/ckeditor5-utils/src/version';
-
```
filipsobol, line 22: I'm not sure why it's here, and I couldn't find a ticket related to this commit. Version should already be checked by the `utils' package, so do we need this?

I also plan to add a static `version` property to the `Plugin` and `Editor` classes so that they can be compared to each other to specifically detect which plugins are causing the `ckeditor-duplicated-modules` error, so the highlighted check may not be needed.

packages/ckeditor5-theme-lark/theme/theme.css
```diff
@@ -2,3 +2,83 @@
  * Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
+
+/* TODO: Order */
```
filipsobol, line 6: The order of the imports is important here. For example, with the current order, the "Powered by CKEditor" banner isn't positioned correctly if there's content after the editor.

I need to spend some more time to find the optimal order, but maybe @oleq can point out any obvious problems with this order?

packages/ckeditor5-ui/src/dropdown/utils.ts
```diff
@@ -38,7 +38,6 @@ import {
 } from '@ckeditor/ckeditor5-utils';
 
 import '../../theme/components/dropdown/toolbardropdown.css';
-import '../../theme/components/dropdown/listdropdown.css';
```
filipsobol, line 41: We no longer need to import the empty CSS files, because the entire theme-lark is built as one CSS file. This increases the base CSS size by about 10kB gzipped, but we no longer need CKEditor specific plugins for webpack or Vite.

scripts/build.mjs
```diff
@@ -0,0 +1,133 @@
...
+				minimize: true,
+				sourceMap
+			} ),
+			typescriptPlugin( {
+				tsconfig: `${ cwd }/tsconfig.json`,
+				typescript,
+				compilerOptions: {
+					declarationDir: `${ cwd }/dist/types`,
+					declaration: true,
+					declarationMap: false, // TODO
```
filipsobol, line 83: This option is probably not needed, but if we have problems with VSCode or other editors navigating to `.d.ts' files instead of the source `.ts' files, then we need to enable this option during development.

scripts/build.mjs
```diff
@@ -0,0 +1,133 @@
...
+			} ),
+			typescriptPlugin( {
+				tsconfig: `${ cwd }/tsconfig.json`,
+				typescript,
+				sourceMap: false
+			} ),
+			terser( {
+				format: {
+					// Remove all comments except third-party licenses and the license banner from above (starting with `!`).
+					comments: ( node, comment ) => /@license/.test( comment.value ) && ( /^!/.test( comment.value ) || !/CKSource/.test( comment.value ) )
```
filipsobol, line 128: I'm not a lawyer, but I'm guessing that we need to keep licenses of 3rd party libraries even in minified CDN builds. Is it safe to assume that these license comments will contain `@license'?

Witoso:
What are the reasons behind the TS bump?

filipsobol:
> What are the reasons behind the TS bump?

I updated TS to the latest version to be able to use `"moduleResolution": "bundler"` added in TypeScript 5.0. It tells TypeScript to ignore lack of file extensions in relative imports, because we will bundle the source code anyway.

https://github.com/ckeditor/ckeditor5/blob/d548fed598e21de857eb103c68411b5210db15a5/tsconfig.kowalski.json#L20

More information: https://www.typescriptlang.org/tsconfig#moduleResolution

packages/ckeditor5-ui/src/dropdown/utils.ts
filipsobol: Note that we need to keep these empty CSS files if we want to still support old installation method.

packages/ckeditor5-markdown-gfm/src/html2markdown/html2markdown.ts
```diff
@@ -11,7 +11,7 @@
 
 // Importing types for this package is problematic, so it's omitted.
 // @ts-ignore
-import TurndownService from 'turndown';
+import TurndownService from 'turndown/lib/turndown.browser.es';
```
filipsobol, line 14: The default `turndown` bundle contains `require('jsdom')`, so I updated the import to use the browser version instead.

scripts/build.mjs
```diff
@@ -0,0 +1,141 @@
...
+	{
+		input: inputPath,
+		output: {
+			format: 'esm',
+			file: path.join( cwd, 'dist', 'index.js'),
+			assetFileNames: '[name][extname]',
+			sourcemap: sourceMap,
+			banner
+		},
+		external,
```
filipsobol, line 65: Change to `external: id => external.some( name => id.startsWith( name ) )`

PR #15336, opened 2023-11-10
Add Next.js integration guide

gorzelinski:

Docs: Add Next.js integration guide. Closes https://github.com/cksource/ckeditor5-commercial/issues/5671

---

### Additional information

### DIFF
--- a/docs/installation/integrations/next-js.md
+++ b/docs/installation/integrations/next-js.md
@@ -1,10 +1,10 @@
 ---
 menu-title: Next.js
 meta-title: Compatibility with Next.js | CKEditor 5 documentation
-meta-description: Learn how to set up real-time collaboration in the Drupal editing platform with the CKEditor 5 Premium Features module.
+meta-description: "Learn how to integrate the Next.js framework with the rich text editor - CKEditor 5 - using both routing strategies: App Router and Pages Router."
 category: integrations
 order: 100
-modified_at: 2023-11-13
+modified_at: 2023-11-14
 ---
 
 # Compatibility with Next.js
@@ -13,15 +13,15 @@ Next.js is a React meta-framework that helps create full-stack web apps for prod
 
 Next.js 13 introduced a new App Router as an alternative to the previous Pages Router. App Router supports server components and is more server-centric than Pages Router, which is client-side oriented.
 
-CKEditor 5 doesn't support server-side rendering yet, but it's possible to integrate it with the Next.js framework. In this guide, we will add the editor to the Next.js project using both routing paradigms. For this purpose, we will need [Next.js CLI](https://nextjs.org/docs/pages/api-reference/create-next-app), CKEditor [online builder](https://ckeditor.com/ckeditor-5/online-builder/), and the official CKEditor {@link installation/integrations/react React component}.
+CKEditor&nbsp;5 doesn't support server-side rendering yet, but it's possible to integrate it with the Next.js framework. In this guide, we will add the editor to the Next.js project using both routing paradigms. For this purpose, we will need [Next.js CLI](https://nextjs.org/docs/pages/api-reference/create-next-app), CKEditor [online builder](https://ckeditor.com/ckeditor-5/online-builder/), and the official CKEditor {@link installation/integrations/react React component}.
 
 ## Preparing a build
 
-In this guide, we will use the [online builder](https://ckeditor.com/ckeditor-5/online-builder/). It is a web UI that lets you create a custom build of CKEditor 5 and download the code as a zip package.
+In this guide, we will use the [online builder](https://ckeditor.com/ckeditor-5/online-builder/). It is a web UI that lets you create a custom build of CKEditor&nbsp;5 and download the code as a zip package.
 
 The online builder is a powerful tool that lets you effortlessly create a rich text editor personalized to your needs. With the online builder, you can choose the desired editor type and plugins, configure the toolbar, and choose the UI language for your editor.
 
-You can learn more about creating custom builds of CKEditor 5 with the online builder in our {@link installation/getting-started/quick-start-other#creating-custom-builds-with-online-builder Customized installation} guide.
+You can learn more about creating custom builds of CKEditor&nbsp;5 with the online builder in our {@link installation/getting-started/quick-start-other#creating-custom-builds-with-online-builder Customized installation} guide.
 
 ## Setting up the project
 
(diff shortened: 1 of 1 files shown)


godai78 (approved):
LGTM

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
+---
+menu-title: Next.js
+meta-title: Compatibility with Next.js | CKEditor 5 documentation
+meta-description: Learn how to integrate the Next.js framework with the rich text editor - CKEditor 5 - using both the App Router and Pages Router routing strategies.
+category: integrations
+order: 100
+modified_at: 2023-11-14
+---
+
+# Compatibility with Next.js
```
Witoso, line 10: I would use the H1 for a SEO bump, something like `Integrate CKEditor 5 with Next.js`.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
+---
+menu-title: Next.js
+meta-title: Compatibility with Next.js | CKEditor 5 documentation
+meta-description: Learn how to integrate the Next.js framework with the rich text editor - CKEditor 5 - using both the App Router and Pages Router routing strategies.
```
Witoso, line 4: I would move CKEditor a bit more to the beginning of the sentence to be more noticeable.

App Router and Pages Router are exclusive, so I would add `or`.

> Learn how to integrate CKEditor 5  with the Next.js framework using the App Router or a Pages Router routing strategies.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+category: integrations
+order: 100
+modified_at: 2023-11-14
+---
+
+# Compatibility with Next.js
+
+Next.js is a React meta-framework that helps create full-stack web apps for production. It offers different rendering strategies like server-side rendering (SSR), client-side rendering (CSR), or static site generation (SSG). Additionally, it provides file-based routing, automatic code splitting, and other handy features out of the box.
```
Witoso, line 12: ~for production~ not needed IMHO.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+
+You can learn more about creating custom builds of CKEditor&nbsp;5 with the online builder in the {@link installation/getting-started/quick-start-other#creating-custom-builds-with-online-builder Customized installation} guide.
+
+## Setting up the project
+
+This guide assumes you will use the official CLI tool for Next.js &ndash; `create-next-app`. Refer to the [Next.js documentation](https://nextjs.org/docs/pages/api-reference/create-next-app) to learn how to set up your project.
+
+The CLI will ask you some questions. Depending on your choices, the folder structure in your project may differ.
+
+```plain
```
Witoso, line 32: I'm not sure if this is needed. It may change in the future, and we would need to update it / track it.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+│   └── vercel.svg
+├── jsconfig.json
+├── next.config.js
+├── package.json
+└── ...
+```
+
+## Integrating the build in your Next.js project
+
+With the Next.js setup done, let's move to the actual integration. First, put your unzipped custom build at the project root in the `ckeditor5` folder. Then, add it as a local dependency using the yarn command:
```
Witoso, line 88: I would somehow use the file structures you have above to point out where the `ckeditor5` folder should go, to better reference this visually.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+
+```bash
+yarn add file:./ckeditor5
```
Witoso, line 91: I would love to have an npm equivalent here (as a comment), as it's more popular.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+The dependency should be visible in the `package.json` file as `ckeditor5-custom-build`. Additionally, we need an external dependency &ndash; the CKEditor&nbsp;5 React component. Next.js extends React, and hence, the component is valid here. You can install it using the following command:
+
+```bash
+yarn add @ckeditor/ckeditor5-react
```
Witoso, line 97: I would love to have an npm equivalent here (as a comment), as it's more popular.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+Next, we will use the installed dependencies in a React component. Create a new component in the components directory, for example, `components/custom-editor.js`. Inside the component file, import all necessary dependencies. Then, create a functional component that returns the CKEditor&nbsp;5 React component with a proper configuration.
+
+```jsx
+// components/custom-editor.js
+
+import React from 'react';
+import { CKEditor } from "@ckeditor/ckeditor5-react";
+import Editor from "ckeditor5-custom-build";
+
+const editorConfiguration = {
```
Witoso, line 109: Theoretically, this is not needed, as toolbar was configured in the online builder. I guess you want to show the possibility of extension. Maybe a comment mentioning this?

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+
+const CustomEditor = dynamic( () => {
+  return import( '../components/custom-editor' );
+}, { ssr: false } );
+
+function Home() {
+  return (
+    <CustomEditor 
+      initialData='<h1>Hello from CKEditor in Next.js!</h1>'
+      onChange={ ( e ) => console.log( e ) }
```
Witoso, line 168: Obsolete, you're not using this prop.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
```
Witoso, line 196: Obsolete, you're not using this prop.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+    />
+  );
+}
+
+export default Home;
+```
+
+Finally, you can run your project. Regardless of your router choice, type `yarn run dev` to see your app in the browser. If you have trouble seeing the editor, remember that the Next.js project ships with CSS files that can interfere with the editor. You can remove them or add your styling.
```
Witoso, line 204: `yarn run dev` or `npm ...`

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,204 @@
...
+    <CustomEditor 
+      initialData='<h1>Hello from CKEditor in Next.js!</h1>'
+      onChange={ ( e ) => console.log( e ) }
+    />
+  );
+}
+
+export default Home;
+```
+
```
Witoso, line 203: App router and pages are so similar that I wonder if we shouldn't have one code block like:

```javascript
'use client' // only in App router.

import React from 'react';
import dynamic from 'next/dynamic';

const CustomEditor = dynamic( () => {
  return import( '../components/custom-editor' );
}, { ssr: false } );
// . . .
```

And join both sections together for brevity. Up to you, I see pros and cons for both approaches.

Witoso (changes requested):
Very good guide! Added some nitpicks, and comments.

docs/installation/integrations/next-js.md
gorzelinski: Good idea!

docs/installation/integrations/next-js.md
gorzelinski: Next.js was marketing itself as a "React framework for production" - that was a pretty good summary, IMO. But we can omit it.

docs/installation/integrations/next-js.md
gorzelinski: Yeah, that's always a possibility. The guide will be fine without this snippet, as I'm looking at it.

docs/installation/integrations/next-js.md
gorzelinski: Yeah, exactly. I'll add a comment.

docs/installation/integrations/next-js.md
gorzelinski: Eventually, I decided to merge these snippets. The difference in the folder structure is visible by showing the final state for both routers. I think it adds more value than the initial state. WDYT?

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,171 @@
...
+This guide assumes you will use the official CLI tool for Next.js &ndash; `create-next-app`. Refer to the [Next.js documentation](https://nextjs.org/docs/pages/api-reference/create-next-app) to learn how to set up your project.
+
+The CLI will ask you some questions. Depending on your choices, the folder structure in your project may differ.
+
+## Integrating the build in your Next.js project
+
+With the Next.js setup done, let's move to the actual integration. First, put your unzipped custom build at the project root in the `ckeditor5` folder. Then, add it as a local dependency using the `npm install --save ./ckeditor5` or yarn command:
+
+```bash
+yarn add file:./ckeditor5
```
Witoso, line 37: I would still add a bash comment :grin: 

```shell
yarn add file:./ckeditor5 # or npm install file:./ckeditor5
```

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,171 @@
...
+    <CustomEditor 
+      initialData='<h1>Hello from CKEditor in Next.js!</h1>'
+    />
+  );
+}
+
+export default Home;
+```
+
+### Page Router
```
Witoso, line 121: I would move this higher just after.

> First, put your unzipped custom build at the project root in the \`ckeditor5\` folder.

Unfortunately, this step is usually very unclear for devs (this is not popular nor best practice currently). And we need some visual guide there. If it's too long, we can cut out some next.js nested files/folders.

docs/installation/integrations/next-js.md
```diff
@@ -0,0 +1,171 @@
...
+export default CustomEditor;
+```
+
+Our `CustomEditor` component is ready to be used inside a page. The usage will differ depending on the chosen routing strategy.
+
+You can create a second page inside the `pages` directory or replace the content of the existing `index.js` page &ndash; we will go with the second option. CKEditor&nbsp;5 is a client-side text editor and relies on the browser APIs, so we need to disable SSR for our custom component. We can lazily load the component using the `dynamic()` function built into Next.js.
+
+App Router, by default, uses server components. It means we need to mark a component as client-side explicitly. We can achieve that by using the `'use client'` directive at the top of a file, above your imports. Also, pay attention to the folder structure, which is different now. The `page.js` file should go into the `app` directory.
+
+```jsx
```
Witoso, line 100: I would add here the comment

```javascript
// app/page.js (App Router)
// pages/page.js (Pages Router)
```

Witoso (commented):
Added comments :)

PR #15340, opened 2023-11-13
Added metadata plugin export validation

### LINKED ISSUE #14725: Validate if plugins described in ckeditor5-metadata are available in a package entry point

The goal is to verify if plugins defined by a `ckeditor5-metadata.json` file are exported via a package entry point (`index.ts`).

For the following JSON file, the validator should check if the statement `import { ShowBlocks } from '@ckeditor/ckeditor5-show-blocks';` is correct.

```json
{
	"plugins": [
		{
			"name": "Show blocks",
			"className": "ShowBlocks",
			"description": "Allows to display the HTML element names of content blocks.",
			"docs": "features/show-blocks.html",
			"path": "src/showblocks.js",
			"uiComponents": [
				{
					"type": "Button",
					"name": "showBlocks",
					"iconPath": "theme/icons/show-blocks.svg"
				}
			]
		}
	]
}
```

Instead of creating a new tool, I suggest redesigning the existing one.

https://github.com/ckeditor/ckeditor5/blob/b82334152d0f6a4ba1a334d8a33bb15b81c82d43/scripts/ci/validate-metadata-icons.js

przemyslaw-zan:

Internal: Updated the CI metadata validation script to check whether plugins in the package metadata are available in a package entry point. Closes #14725.

---

### Additional information


scripts/ci/validate-metadata-files.js
```diff
@@ -0,0 +1,184 @@
...
+
+	const metadataPath = upath.join( packagePath, 'ckeditor5-metadata.json' );
+	const indexPath = upath.join( packagePath, 'src', 'index.ts' );
+
+	let metadata = null;
+	let index = null;
+
+	if ( await fs.exists( metadataPath ) ) {
+		metadata = await fs.readJSON( metadataPath );
+	}
```
pomek, line 99: To avoid `let`, I suggest creating a helper function:

```js
async function readJSON( filePath ) {
	if ( await fs.exists( filePath ) ) {
		return fs.readJSON( filePath );
	}

	return null;
}

const metadata = await readJSON( metadataPath ); 
```

scripts/ci/validate-metadata-files.js
```diff
@@ -0,0 +1,184 @@
...
+	for ( const pluginData of packageData.metadata.plugins ) {
+		if ( !pluginData.uiComponents ) {
+			continue;
+		}
+
+		const iconPaths = pluginData.uiComponents.map( uiComponent => uiComponent.iconPath ).filter( Boolean );
+
+		if ( !iconPaths.length ) {
+			continue;
+		}
```
pomek, line 144:
```suggestion
	const pluginsToCheck = packageData.metadata.plugins
		.filter( ( { uiComponents } ) ) => !uiComponents )
		.filter( ( { uiComponents } ) => {
			return uiComponents.map( ( { iconPath } ) => iconPath )
				.filter( Boolean )
				.length;
		} );

	for ( const pluginData of pluginsToCheck ) {
```

This would simplify the `for..of` loop. This loop could look even nicer after simplifying the `try/catch`. Not to mention that it's easier to modify and understand what's going on here.

pomek (commented):
I proposed a few syntax changes that would make the script more readable, I think.

scripts/ci/validate-metadata-files.js
przemyslaw-zan: If the issue is to avoid `let`, then I've used ternary operator instead. New `readJSON()` function does not make much sense as it would only be used once, since the second file we read here is not JSON.

scripts/ci/validate-metadata-files.js
```diff
@@ -0,0 +1,166 @@
...
+		} );
+	}
+
+	if ( !invalidMetadata.length ) {
+		console.log( green( 'Validation successful.' ) );
+
+		return;
+	}
+
+	for ( const x of invalidMetadata ) {
```
pomek, line 63: `x`? If cannot find a proper variable name, I propose a destructing assignment. 

```suggestion
	for ( const { name, missingExports, missingIcons } of invalidMetadata ) {
```

przemyslaw-zan: Oops, forgot to replace this one

scripts/ci/validate-metadata-files.js
```diff
@@ -0,0 +1,166 @@
...
+
+	process.exit( 1 );
+}
+
+/**
+ * Gets plugins from metadata file at provided path.
+ * Additionally, adds `packageName` of the plugin to returned data.
+ */
+async function getPackageData( packagePath ) {
+	const name = packagePath.split( '/' ).at( -1 );
```
psmyrek, line 89:
```suggestion
	const name = upath.basename( packagePath );
```

scripts/ci/validate-metadata-files.js
```diff
@@ -0,0 +1,166 @@
...
+
+	const pkgJsonPath = upath.join( packagePath, 'package.json' );
+	const pkgJsonContent = await fs.readJSON( pkgJsonPath );
+
+	const metadataPath = upath.join( packagePath, 'ckeditor5-metadata.json' );
+	const indexPath = upath.join( packagePath, pkgJsonContent.main );
```
psmyrek, line 95: If there is no `main` field in any package, then this script throws:

obraz

scripts/ci/validate-metadata-files.js
```diff
@@ -0,0 +1,166 @@
...
+	const metadata = await fs.exists( metadataPath ) ? await fs.readJSON( metadataPath ) : null;
+	const index = await fs.exists( indexPath ) ? await fs.readFile( indexPath, 'utf-8' ) : null;
+
+	return { name, metadata, index };
+}
+
+function getMissingExports( packageData ) {
+	if ( !packageData.index ) {
+		throw new Error( 'missing index' );
+	}
```
psmyrek, line 106: Do we want to throw an error from this script? It would be better to collect packages without index file and print error message at the end.

psmyrek (commented):
Some cosmetic ideas. Other than that, it works as expected.

scripts/ci/validate-metadata-files.js
pomek: What do you think we should do in this case?

pomek (approved):
LGTM. Works as expected. Let's fix issues mentioned by Paweł and we can merge it.

scripts/ci/validate-metadata-files.js
psmyrek: In this script we check the metadata of packages and if the package contains a file with metadata, it must also contain an index file to know where the plugins announced in the metadata are available.

Not all packages have an entry point and that seems OK, because apparently they don't have to have one.

So, in case a package contains a metadata file but it does not have an entry point, I would expect the script not to explode and the package to be marked as incorrect, because it has missing exports (because we do not know the index file where these exports are available).

psmyrek:
In https://github.com/ckeditor/ckeditor5/pull/15340/commits/b5d86b04c8f97b7a7f706137b9fa84e0770b8cea I pushed small fix for the `--cwd` option, because the `cwd` should be always an absolute path. Without this change, executing the script with `--cwd` does not work as expected and it reports multiple missing icons.

psmyrek (approved):
LGTM 🎉

PR #15383, opened 2023-11-21
`docs:lint` script will now treat values from `.vale.ini` as defaults

### LINKED ISSUE #15382: `docs:lint` script should treat values from `.vale.ini` as defaults
See https://github.com/ckeditor/ckeditor5/pull/15316#discussion_r1399224084

przemyslaw-zan:

Internal: `docs:lint` script will now use the value of `MinAlertLevel` from `.vale.ini` file as default value of the `alert` option. Previously, if the `-a`/`--alert` option was not specified `warning` would be used. `warning` is still a fallback value. Closes #15382.

---

### Additional information


pomek (changes requested):
Please, from #stable, to #stable.

pomek (approved):
LGTM.

PR #15393, opened 2023-11-22
Add paste from markdown experimental

### LINKED ISSUE #2321: Support converting pasted markdown content
## Is this a bug report or feature request?

🆕 Feature request

## 💻 Version of CKEditor

CKEditor v5 @ 12.4.0

## 📋 Steps to reproduce

Currently there's no integration with the clipboard plugin (pasting).

For example use the following scenario:

1. Copy following text to your clipboard: `foo **bar** [baz](https://ckeditor.com).`.
1. Open our [Markdown output](https://ckeditor.com/docs/ckeditor5/latest/features/markdown.html) sample.
1. Select all the content.
1. Delete the content.
1. Paste using `ctrl/cmd+v`.

## ✅ Expected result

Pasted markdown gets formatted.

## ❎ Actual result

Plain text is being pasted.

I pushed a [t/24](https://github.com/ckeditor/ckeditor5-markdown-gfm/tree/t/27) branch with a PoC.

---

martnpaneq:

Feature (markdown-gfm): Added an experimental support for pasting from markdown. Closes #2321.

---

### Additional information

This is an experimental implementation dedicated for further testing by the end user.
The tests provide 100% code coverage, but there are missing integration tests, and tests for different clipboard contents in Firefox, Edge and Safari.

pasteFromMarkdown


packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+		} );
+	}
+
+	/**
+	 * Determines if code copied from a website in `text/html` type can be parsed as markdown.
+	 * It removes any OS specific HTML tags e.g. <meta> on Mac OS and <!--StartFragment--> on Windows.
+	 * Then removes a single wrapper HTML tag or wrappers for sibling tags, and if there are no more tags left,
+	 * returns the remaining text. Returns null, if there are any remaining HTML tags detected.
+	 *
+	 * @private
```
arkflpc, line 93: The method is already `private` on TS-level. This tag is not needed.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+			return null;
+		}
+
+		return this._replaceHtmlReservedEntitiesWithCharacters( withoutWrapperTag );
+	}
+
+	/**
+	 * Removes OS specific tags.
+	 *
+	 * @private
```
arkflpc, line 113: The method is already `private` on TS-level. This tag is not needed.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+		return withoutBodyTag.replace( /^<!--StartFragment-->/, '' ).replace( /<!--EndFragment-->$/, '' ).trim();
+	}
+
+	/**
+	 * Removes a single HTML wrapper tag from string.
+	 *
+	 * @private
+	 * @param htmlString Clipboard content without any OS specific tags.
+	 */
+	private _removeWrapperTag( htmlString: string ): string {
```
arkflpc, line 134: The method is already `private` on TS-level. This tag is not needed.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+		return htmlString.replace( /^<(\w+)\b[^>]*>\s*([\s\S]*?)\s*<\/\1>/, '$2' ).trim();
+	}
+
+	/**
+	 * Removes multiple HTML wrapper tags from a list of sibling HTML tags.
+	 *
+	 * @private
```
arkflpc, line 141: The method is already `private` on TS-level. This tag is not needed.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+			br.replaceWith( '\n' );
+		}
+
+		return tempDiv.innerHTML;
+	}
+
+	/**
+	 * Determines if sting contains sibling HTML tags at root level.
+	 *
+	 * @private
```
arkflpc, line 166: The method is already `private` on TS-level. This tag is not needed.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+		const tempDiv = document.createElement( 'div' );
+		tempDiv.innerHTML = htmlString;
+
+		return tempDiv.children.length > 1;
+	}
+
+	/**
+	 * Determines if string contains any HTML tags.
+	 *
+	 * @private
```
arkflpc, line 179: The method is already `private` on TS-level. This tag is not needed.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+	 * @param str
```
arkflpc, line 180: Not particularly descriptive. I'd remove this tag.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+	 */
+	private _containsAnyRemainingHtmlTags( str: string ): boolean {
+		return /<[^>]+>[\s\S]*<[^>]+>/.test( str );
+	}
+
+	/**
+	 * Replaces the reserved HTML entities with the actual characters.
+	 *
+	 * @private
```
arkflpc, line 189: The method is already `private` on TS-level. This tag is not needed.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+	/**
+	 * Determines if code copied from a website in `text/html` type can be parsed as markdown.
+	 * It removes any OS specific HTML tags e.g. <meta> on Mac OS and <!--StartFragment--> on Windows.
+	 * Then removes a single wrapper HTML tag or wrappers for sibling tags, and if there are no more tags left,
+	 * returns the remaining text. Returns null, if there are any remaining HTML tags detected.
+	 *
+	 * @private
+	 * @param htmlString Clipboard content in `text/html` type format.
+	 */
+	private _parseMarkdownFromHtml( htmlString: string ): string | null {
```
arkflpc, line 96: I like the variable naming in this (and other) methods. It's very descriptive.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+		const withoutOsSpecificTags = this._removeOsSpecificTags( htmlString );
+
+		const withoutWrapperTag = this._isHtmlList( withoutOsSpecificTags ) ?
+			this._removeListWrapperTagsAndBrs( withoutOsSpecificTags ) :
+			this._removeWrapperTag( withoutOsSpecificTags );
+
+		if ( this._containsAnyRemainingHtmlTags( withoutWrapperTag ) ) {
```
arkflpc, line 103: Valid Markdown can contain HTML tags (for instance, complicated tables). It seems like we don't support them here. Is it intentional?

martnpaneq: We had a lot of discussions about this topic, and our current goal is to provide full paste support in format `text/plain` - code copied from online markdown editors and code blocks. We decided to provide limited support for pasting code in format `text/html` - copied from a website but not in a code block.

We added support for simple html structures, so pasting most code from here should work fine: https://www.markdownguide.org/basic-syntax/

Unfortunately tables are a bit more complex, and the algorithm detecting them would be much more complicated.

We tried to follow how Notion handles pasting markdown, and it behaves almost exactly the same as our implementation - including tables.

martnpaneq: Plus, this plugin is released as experimental. If there are any requests from the community for paste markdown table from HTML website support, we can get back to it then.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	/**
+	 * Removes a single HTML wrapper tag from string.
+	 *
+	 * @private
+	 * @param htmlString Clipboard content without any OS specific tags.
+	 */
+	private _removeWrapperTag( htmlString: string ): string {
+		return htmlString.replace( /^<(\w+)\b[^>]*>\s*([\s\S]*?)\s*<\/\1>/, '$2' ).trim();
```
niegowski, line 132: Can this lead to content loss? Maybe this should be more precise?

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+		return tempDiv.innerHTML;
+	}
+
+	/**
+	 * Determines if sting contains sibling HTML tags at root level.
+	 *
+	 * @private
+	 * @param htmlString Clipboard content without any OS specific tags.
+	 */
+	private _isHtmlList( htmlString: string ): boolean {
```
niegowski, line 164: The name of this method is confusing.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	/**
+	 * Determines if string contains any HTML tags.
+	 *
+	 * @private
+	 * @param str
+	 */
+	private _containsAnyRemainingHtmlTags( str: string ): boolean {
+		return /<[^>]+>[\s\S]*<[^>]+>/.test( str );
```
niegowski, line 175: Looks like this pattern could be simpler. Wouldn't search for `<` be enough?

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	/**
+	 * Replaces the reserved HTML entities with the actual characters.
+	 *
+	 * @private
+	 * @param htmlString Clipboard content without any tags.
+	 */
+	private _replaceHtmlReservedEntitiesWithCharacters( htmlString: string ) {
+		return htmlString.replace( '&gt;', '>' ).replace( '&lt;', '<' );
```
niegowski, line 184: Are other entities not needed? What about `&nbsp;`?

niegowski (changes requested):

packages/ckeditor5-markdown-gfm/docs/_snippets/features/paste-from-markdown.js
```diff
@@ -0,0 +1,83 @@
...
+import { CKBox } from '@ckeditor/ckeditor5-ckbox';
+import { PictureEditing, ImageResize, AutoImage } from '@ckeditor/ckeditor5-image';
+import { LinkImage } from '@ckeditor/ckeditor5-link';
+import { Font } from '@ckeditor/ckeditor5-font';
+
+// Umberto combines all `packages/*/docs` into the `docs/` directory. The import path must be valid after merging all directories.
+import ClassicEditor from '../build-classic';
+
+const plugins = ClassicEditor.builtinPlugins
+	// Remove the `List` plugin as in a single demo we want to use the Document list feature.
```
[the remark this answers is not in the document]
pomek: It is. Please, see https://github.com/ckeditor/ckeditor5/blob/b6c56e5ea7d31456590d9aeb5e4833810197294a/docs/_snippets/build-classic.js#L56.

packages/ckeditor5-markdown-gfm/docs/features/paste-markdown.md
```diff
@@ -0,0 +1,87 @@
...
+	This feature is not available in any of the {@link installation/getting-started/predefined-builds predefined builds}.
+</info-box>
+
+To enable this data processor in your editor, install the [`@ckeditor/ckeditor5-markdown-gfm`](https://www.npmjs.com/package/@ckeditor/ckeditor5-markdown-gfm) package:
+
+```
+npm install --save @ckeditor/ckeditor5-markdown-gfm
+```
+
+Then add the {@link module:markdown-gfm/markdown~Markdown} plugin to the editor configuration, which will change the default {@link module:engine/dataprocessor/dataprocessor~DataProcessor data processor} to the {@link module:markdown-gfm/gfmdataprocessor~GFMDataProcessor}:
```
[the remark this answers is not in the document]
pomek: No, `Markdown` is not required. I simplified this sentence.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
martnpaneq: I removed the complicated regex, because this logic can be handled by this function: `_removeFirstLevelWrapperTagsAndBrs()`.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
martnpaneq: This method is no longer needed as I'm handling single HTML tag and list of tags with a single same function.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
martnpaneq: Good catch.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
martnpaneq: Added `&nbsp;`

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -8,14 +8,14 @@
  */
 
 import { type Editor, Plugin } from 'ckeditor5/src/core';
-import { Clipboard, type ClipboardInputTransformationEvent, type ClipboardPipeline } from 'ckeditor5/src/clipboard';
+import { type ClipboardInputTransformationEvent, ClipboardPipeline } from 'ckeditor5/src/clipboard';
```
niegowski, line 11:
```suggestion
import { ClipboardPipeline, type ClipboardInputTransformationEvent } from 'ckeditor5/src/clipboard';
```

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -8,14 +8,14 @@
  */
 
 import { type Editor, Plugin } from 'ckeditor5/src/core';
```
niegowski, line 10:
```suggestion
import { Plugin, type Editor } from 'ckeditor5/src/core';
```

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -125,23 +121,12 @@ export default class PasteFromMarkdownExperimental extends Plugin {
...
-
 	/**
 	 * Removes multiple HTML wrapper tags from a list of sibling HTML tags.
 	 *
-	 * @private
 	 * @param htmlString Clipboard content without any OS specific tags.
 	 */
-	private _removeListWrapperTagsAndBrs( htmlString: string ): string {
+	private _removeFirstLevelWrapperTagsAndBrs( htmlString: string ): string {
 		const tempDiv = document.createElement( 'div' );
```
niegowski, line 130: Is this safe in the context of interactive content? Maybe we should use a separate DOM document like here: https://github.com/ckeditor/ckeditor5/blob/7a5a239de7a46080626d02a27572f5199c1839b9/packages/ckeditor5-engine/src/dataprocessor/basichtmlwriter.ts#L21-L28

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,166 @@
...
+	/**
+	 * Replaces the reserved HTML entities with the actual characters.
+	 *
+	 * @param htmlString Clipboard content without any tags.
+	 */
+	private _replaceHtmlReservedEntitiesWithCharacters( htmlString: string ) {
+		return htmlString
+			.replace( new RegExp( '&gt;', 'g' ), '>' )
+			.replace( new RegExp( '&lt;', 'g' ), '<' )
+			.replace( new RegExp( '&nbsp;', 'g' ), ' ' );
```
niegowski, line 164:
```suggestion
			.replace( /&gt;/g, '>' )
			.replace( /&lt;/g, '<' )
			.replace( /&nbsp;/g, ' ' );
```

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
niegowski: Or maybe we should use the `DOMParser`? https://github.com/ckeditor/ckeditor5/blob/7a5a239de7a46080626d02a27572f5199c1839b9/packages/ckeditor5-engine/src/dataprocessor/htmldataprocessor.ts#L122C3-L122C72

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,166 @@
...
+	 * Removes multiple HTML wrapper tags from a list of sibling HTML tags.
+	 *
+	 * @param htmlString Clipboard content without any OS specific tags.
+	 */
+	private _removeFirstLevelWrapperTagsAndBrs( htmlString: string ): string {
+		const tempDiv = document.createElement( 'div' );
+		tempDiv.innerHTML = htmlString;
+
+		const outerElements = tempDiv.querySelectorAll( ':not(:empty)' );
+		const brElements = tempDiv.querySelectorAll( 'br' );
```
niegowski, line 134: This is a non-live `NodeList`. Is this safe to gather those elements before other content manipulations?

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -0,0 +1,166 @@
...
+	/**
+	 * Removes multiple HTML wrapper tags from a list of sibling HTML tags.
+	 *
+	 * @param htmlString Clipboard content without any OS specific tags.
+	 */
+	private _removeFirstLevelWrapperTagsAndBrs( htmlString: string ): string {
+		const tempDiv = document.createElement( 'div' );
+		tempDiv.innerHTML = htmlString;
+
+		const outerElements = tempDiv.querySelectorAll( ':not(:empty)' );
```
niegowski, line 133: This selector is contrary to the method name. It collects all the non-empty elements, not only the 'first-level'. This looks like it would unwrap all inline tags also. I'm not sure if this is the desired behavior.

niegowski (changes requested):

arkflpc (approved):

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
martnpaneq: Changed to 
```js
const outerElements = tempElement.querySelectorAll( ':scope > *' );
```

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
martnpaneq: Good catch. In the new approach I'm performing manipulations on `<br>` elements first, and then using `querySelectorAll` for remaining elements.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
martnpaneq: using `DOMParser`.

packages/ckeditor5-markdown-gfm/src/pastefrommarkdownexperimental.ts
```diff
@@ -7,11 +7,13 @@
...
  */
 
-import { type Editor, Plugin } from 'ckeditor5/src/core';
-import { type ClipboardInputTransformationEvent, ClipboardPipeline } from 'ckeditor5/src/clipboard';
+import { Plugin, type Editor } from 'ckeditor5/src/core';
+import { ClipboardPipeline, type ClipboardInputTransformationEvent } from 'ckeditor5/src/clipboard';
 import GFMDataProcessor from './gfmdataprocessor';
 import type { ViewDocumentKeyDownEvent } from 'ckeditor5/src/engine';
 
+const FORMATTING_TAGS = [ 'B', 'STRONG', 'I', 'EM', 'MARK', 'SMALL', 'DEL', 'INS', 'SUB', 'SUP' ];
```
niegowski, line 15: I'm not sure if this is enough. Maybe we could go in the opposite direction and try to filter out only wrappers that we know are just wrappers without meaning for the content, and then just check if there is no `<` in the filtered content. I'm afraid that the current approach could filter out too many tags. Is this safe for pasting multiple paragraphs or inline images inside the text?

niegowski (changes requested):

Issue #15014, opened 2023-09-17
integrating editor with vite and pnpm workspaces gives ambiguous indirect export: default error


Joshuajrodrigues:
## 📝 Provide detailed reproduction steps (if any)

my project layout looks like this
![image](https://github.com/ckeditor/ckeditor5/assets/54748294/d3471477-6af0-479e-ad24-91f509b1572d)
Ckeditor5 is a direct build from your online build page.
```
{
  "name": "@workspace/ckeditor5-custom-build",
  "author": "CKSource",
  "description": "A custom CKEditor 5 build made by the CKEditor 5 online builder.",
  "version": "0.0.1",
  "license": "SEE LICENSE IN LICENSE.md",
  "private": true,
  "main": "./build/ckeditor.js",
  "types": "./build/ckeditor.d.ts",
  "dependencies": {
    "@ckeditor/ckeditor5-autoformat": "39.0.2",
    "@ckeditor/ckeditor5-basic-styles": "39.0.2",
    "@ckeditor/ckeditor5-block-quote": "39.0.2",
    "@ckeditor/ckeditor5-code-block": "39.0.2",
    "@ckeditor/ckeditor5-editor-classic": "39.0.2",
    "@ckeditor/ckeditor5-essentials": "39.0.2",
    "@ckeditor/ckeditor5-heading": "39.0.2",
    "@ckeditor/ckeditor5-image": "39.0.2",
    "@ckeditor/ckeditor5-indent": "39.0.2",
    "@ckeditor/ckeditor5-link": "39.0.2",
    "@ckeditor/ckeditor5-list": "39.0.2",
    "@ckeditor/ckeditor5-paragraph": "39.0.2",
    "@ckeditor/ckeditor5-table": "39.0.2",
    "@ckeditor/ckeditor5-typing": "39.0.2"
  },
  "devDependencies": {
    "@ckeditor/ckeditor5-dev-translations": "^32.1.2",
    "@ckeditor/ckeditor5-dev-utils": "^32.1.2",
    "@ckeditor/ckeditor5-theme-lark": "^39.0.2",
    "css-loader": "^5.2.7",
    "postcss": "^8.4.29",
    "postcss-loader": "^4.3.0",
    "raw-loader": "^4.0.2",
    "style-loader": "^2.0.0",
    "terser-webpack-plugin": "^4.2.3",
    "ts-loader": "^9.4.4",
    "typescript": "^4.9.5",
    "webpack": "^5.88.2",
[...]

Issue #9807, opened 2021-06-01
Using Vite instead of Webpack


lchrennew:
## 📝 Provide an introduction on How to build from source using Vite.js instead of Webpack.

_My code, which build editor from source, has been migrated to Vue 3+ and Vite.js, and an introduction on How to config the `vite.config.js` is nessesary, because the editor not works after migration.

---

⚠️ **EDIT #2** by @Witoso : New installation methods shipped in [v42.0.0](https://github.com/ckeditor/ckeditor5/releases/tag/v42.0.0) work seamlessly with Vite.

---
⚠️ **EDIT** by @Reinmar: We shipped an experimental integration for Vite. Read more in https://github.com/ckeditor/ckeditor5/issues/9807#issuecomment-1363284224 🚀

lchrennew:
any feedback?

pierre-H:
Any news ? 😃

brucewar:
waiting...

pierre-H:
ping @oleq @Reinmar 😃

lchrennew:
waiting...

lchrennew:
The current simplest approach: modify and use the build js file 
![image](https://user-images.githubusercontent.com/1522365/128314229-be3d60ea-ea54-4211-b3a3-61a321dcb5b5.png)
@pierre-H @brucewar

pomek:
Guys, could you describe your problem? Steps you have already done and results that have been produced.  Please, share your configuration that allows reproducing the issue on our side. 

Vite is a new thing that we never worked on before.

lchrennew:
> Could you describe your problem in more detail? Please include the steps you have already done and the results that have been produced. Also please share your configuration that allows reproducing the issue on our side.
> 
> Vite is a new thing that we have never worked on before so the more details you are able to provide, the easier it will be for us to help you.

More vue developer starts to use Vite.js to replace heavy weight Vue CLI / Webpack, and the following document doesn't help those using Vite.js.
https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/frameworks/vuejs-v3.html#configuring-vueconfigjs

jsmnbom:
ckeditor5 seems to build fine for me in a pure vite project, with one key exception. I had to add a plugin like below to transform the SVG files properly. It also seems to break in SSR mode, but for me that's not a big deal as I can just exclude it on the server side.
```
((options) => {
      return {
        name: 'vite-ckeditor-svg-raw-plugin',
[...]

PR #15405, opened 2023-11-23
Docs: Add context for integrating Vite with the custom CKEditor build.

### LINKED ISSUE #15396: Add context for setup: React + Vite + CKEditor 5 Custom build.
Users have problems with the mentioned setup, so we should provide some helpful information and context to the [existing guide](https://ckeditor.com/docs/ckeditor5/latest/installation/integrations/react.html).

gorzelinski:

Docs: Add context for integrating Vite with the custom CKEditor build. Closes #15396.

---

### Additional information

Related issues:
- https://github.com/ckeditor/ckeditor5/issues/9807
- https://github.com/ckeditor/ckeditor5/issues/15014

### DIFF
--- a/docs/installation/integrations/react.md
+++ b/docs/installation/integrations/react.md
@@ -54,9 +54,8 @@ class App extends Component {
 						// You can store the "editor" and use when it is needed.
 						console.log( 'Editor is ready to use!', editor );
 					} }
-					onChange={ ( event, editor ) => {
-						const data = editor.getData();
-						console.log( { event, editor, data } );
+					onChange={ ( event ) => {
+						console.log( event );
 					} }
 					onBlur={ ( event, editor ) => {
 						console.log( 'Blur.', editor );
@@ -232,7 +231,7 @@ class App extends Component {
                             this.editor.ui.view.toolbar.element.remove();
                         }
                     } }
-                    onChange={ ( event, editor ) => console.log( { event, editor } ) }
+                    onChange={ ( event ) => console.log( event ) }
                     editor={ DecoupledEditor }
                     data="<p>Hello from CKEditor&nbsp;5's decoupled editor!</p>"
                     config={ /* the editor configuration */ }
@@ -268,13 +267,13 @@ This guide assumes that you have created a zip archive with the editor built usi
 
 The directory with the editor's build cannot be placed inside the `src/` directory because Node could return an error:
 
-```
+```plain
 FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 ```
 
 Because of that, we recommend placing the directory next to the `src/` and `node_modules/` folders:
 
-```
+```plain
 ├── ckeditor5
 │   ├── build
 │   ├── sample
@@ -319,9 +318,8 @@ class App extends Component {
 						// You can store the "editor" and use when it is needed.
 						console.log( 'Editor is ready to use!', editor );
 					} }
-					onChange={ ( event, editor ) => {
-						const data = editor.getData();
-						console.log( { event, data } );
+					onChange={ ( event ) => {
+						console.log( event);
 					} }
 					onBlur={ ( event, editor ) => {
 						console.log( 'Blur.', editor );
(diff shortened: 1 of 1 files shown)


docs/installation/integrations/react.md
```diff
@@ -321,7 +321,7 @@ class App extends Component {
 					} }
 					onChange={ ( event, editor ) => {
 						const data = editor.getData();
-						console.log( { event, editor, data } );
+						console.log( { event, data } );
```
Witoso, line 324: Data can we large as well, `editor.getData` can significantly delay typing, I would comment out both lines.

docs/installation/integrations/react.md
```diff
@@ -599,6 +599,10 @@ import { createRequire } from 'node:module';
 const require = createRequire( import.meta.url );
 ```
 
+<info-box warning>
```
Witoso, line 602: I think this should appear higher, near the: Configuring `vite.config.js` / `vite.config.ts`

Witoso (changes requested):

Witoso:
One of the most popular landing pages, we need to make sure what we add makes sense.

docs/installation/integrations/react.md
gorzelinski: I read that @FilipTokarski couldn't replicate this lag. I also tried and failed - I couldn't copy this behavior even though I was typing like a mad lad with additional content. However, I believe it can slow down the editor for somebody, so I remove this.

docs/installation/integrations/react.md
gorzelinski: I'll probably add another section for Vite + custom build, so we don't have to link to comments, etc.

docs/installation/integrations/react.md
FilipTokarski: Yeah, let's remove those lines. Unfortunately using `getData()` inside `onChange` is a known case and it is problematic for some users. So it's something that we definitely do not recommend.

docs/installation/integrations/react.md
gorzelinski: Eventually, I added a new section under the Integration a build from the online builder header. Configuring the `vite.config.js` header belongs to the build from the source path, and from my testing, it works fine. The problem is with the custom build from the online builder, so I put the info there.

docs/installation/integrations/react.md
```diff
@@ -338,11 +336,36 @@ class App extends Component {
...
+  optimizeDeps: {
+    include: ['@workspace/ckeditor5-custom-build'],
+  },
+  build: {
+    commonjsOptions: {
+      include: [/@workspace\/ckeditor5-custom-build/, /node_modules/],
+    }
+  }
+})
+```
```
Witoso, line 356: I think I get it, but the online builder route doesn't require our plugin, right?

gorzelinski:
Yeah, there is a mess in this guide. But I would also leave it as it is for now. With new installation methods, it will need an update anyway.

docs/installation/integrations/react.md
```diff
@@ -338,11 +336,36 @@ class App extends Component {
...
+
+Also, remember to use the correct import style for the CKEditor React component:
+
+```js
+import { CKEditor } from '@ckeditor/ckeditor5-react';
+```
```
filipsobol, line 362: I don't think we need this. It seemed like a minor oversight by the developer who reported this issue and the documentation above is showing this import correctly.

docs/installation/integrations/react.md
filipsobol: Yes, bundle generated by the Online Builder project is ready to use and doesn't require CKEditor-specific plugins.

docs/installation/integrations/react.md
```diff
@@ -319,9 +318,8 @@ class App extends Component {
 						// You can store the "editor" and use when it is needed.
 						console.log( 'Editor is ready to use!', editor );
 					} }
-					onChange={ ( event, editor ) => {
-						const data = editor.getData();
-						console.log( { event, editor, data } );
+					onChange={ ( event ) => {
+						console.log( event);
```
pszczesniak, line 322:
```suggestion
						console.log( event );
```

docs/installation/integrations/react.md
gorzelinski: I wasn't sure about it either. I'll remove it.

docs/installation/integrations/react.md
gorzelinski: Good eye!

PR #15417, opened 2023-11-27
Dialogs - single instance, `isOn` button state, recreating instead of resetting the view

Dumluregn:

*   Allow only one dialog instance across all the editor instances (opening a dialog in another editor closes and destroys the old one).
*   Introduce `isOn` state for the buttons using the dialog.
*   Destroy and recreate the `DialogView` instead of resetting it.
    *  But make sure the integrator's view is not destroyed in the process and can be re-used next time the dialog shows up.
* Ensured the correct part of the dialog gets focused when the dialog shows up. 
* Maintained the correct focus cycling order after code refactoring.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ui/src/dialog/dialog.ts
+++ b/packages/ckeditor5-ui/src/dialog/dialog.ts
@@ -7,7 +7,7 @@
  * @module ui/dialog/dialog
  */
 
-import View from '../view';
+import type View from '../view';
 import { type Editor, Plugin } from '@ckeditor/ckeditor5-core';
 import DialogView, { type DialogViewCloseEvent, DialogViewPosition } from './dialogview';
 import type { DialogActionButtonDefinition } from './dialogactionsview';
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -115,13 +103,35 @@ export default class Dialog extends Plugin {
...
+
+		// If the dialog was closed by opening in another editor instance,
+		// we need to update the state in the previous instance manually.
+		Dialog.view.on( 'destroy', () => {
+			this.id = '';
+			this.isOpen = false;
+		} );
+
+		editor.ui.view.body.add( Dialog.view );
+		editor.ui.focusTracker.add( Dialog.view.element! );
```
oleq, line 126: ATM there are references stored to previous dialog elements in the focusTracker. We should remove elements when views are destroyed.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -293,6 +298,15 @@ export default class DialogView extends DraggableViewMixin( View ) implements Dr
 		}
 	}
 
+	/**
+	 * TODO
+	 */
+	public override destroy(): void {
+		this.element?.remove();
```
oleq, line 305: We've never done anything like this. It's the parent view or the controller that are responsible for removing views from DOM. I'd mirror `editor.ui.view.body.add( Dialog.view );` behavior in `Dialog` class and clean up things there.

PR #15423, opened 2023-11-28
Editing images not in CKBox

arkflpc:

Type: Message. Closes #000.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ckbox/src/ckboximageedit/ckboximageeditcommand.ts
+++ b/packages/ckeditor5-ckbox/src/ckboximageedit/ckboximageeditcommand.ts
@@ -21,6 +21,7 @@ import { prepareImageAssetAttributes } from '../ckboxcommand';
 import type { CKBoxRawAssetDefinition, CKBoxRawAssetDataDefinition } from '../ckboxconfig';
 
 import type { ImageUtils } from '@ckeditor/ckeditor5-image';
+import { createEditabilityChecker, getImageEditorMountOptions } from './utils';
 
 /**
  * The CKBox edit image command.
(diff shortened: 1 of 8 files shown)


pszczesniak (changes requested):
Works nice 🎉  just few small things:
1. In manual test i suggest to remove `data-ckbox-resource-id="example-id-for-image"` from example image (throws an error while we want to edit it cause it's not a real CKBox image).
2. After adding an image from `imgur` i can click `edit` but if i resign (back arrow button) the `edit` button is still active (video)

https://github.com/ckeditor/ckeditor5/assets/7074833/31d6e452-3bb8-47ea-9dd3-46faa2bfadc5

pszczesniak (approved):
LGTM :+1:

One thing (not a blocker) - would be good to add some explanation in manual test about possibilities in image editing.

I like how you made code cleaner and more readable! :clap:

packages/ckeditor5-ckbox/src/ckboximageedit/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+	if ( Array.isArray( allowExternalImagesEditing ) ) {
+		const urlMatchers = allowExternalImagesEditing.map( createUrlChecker );
+
+		return src => urlMatchers.some( matcher => matcher( src ) );
+	}
+
+	if ( allowExternalImagesEditing == 'origin' ) {
+		const origin = global.window.location.origin;
+
+		return src => src.startsWith( origin + '/' );
```
niegowski, line 56: Will this work correctly with relative URLs?

packages/ckeditor5-ckbox/src/ckboximageedit/ckboximageeditcommand.ts
```diff
@@ -169,6 +210,16 @@ export default class CKBoxImageEditCommand extends Command {
 		return states;
 	}
 
+	private _checkIfElementIsBeingProcessed( selectedElement: ModelElement ) {
```
niegowski, line 213: Missing docs.

mmotyczynska (commented):
The feature works as expected 👍 
Some docs missing (I added comments).

packages/ckeditor5-ckbox/src/ckboximageedit/ckboximageeditcommand.ts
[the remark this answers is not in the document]
arkflpc: It's private and the name is long enough.

PR #15434, opened 2023-11-30
Add the missing type for the list view.

### LINKED ISSUE #15298: Missing `ListSeparatorView` type in `ListView` view collection

`ViewCollection` in `ListView` is missing `ListSeparatorView` type, as this view could be one of the child views.

## 📃 Other details

Commit: https://github.com/ckeditor/ckeditor5/commit/27996929c64b044cc9d1331af38f35b5de504113#diff-26c902f9c4eff860ec9dc0755aea9160ff7a24d82122970e4f0c2f90379ce0c7R42

gorzelinski:

Add the missing type for the list view. Closes #15298.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ui/src/list/listview.ts
+++ b/packages/ckeditor5-ui/src/list/listview.ts
@@ -12,6 +12,7 @@ import FocusCycler from '../focuscycler';
 
 import type ListItemView from './listitemview';
 import ListItemGroupView from './listitemgroupview';
+import type ListSeparatorView from './listseparatorview';
 import type DropdownPanelFocusable from '../dropdown/dropdownpanelfocusable';
 import ViewCollection from '../viewcollection';
 
@@ -39,7 +40,7 @@ export default class ListView extends View<HTMLUListElement> implements Dropdown
 	/**
 	 * Collection of the child list views.
 	 */
-	public readonly items: ViewCollection<ListItemView | ListItemGroupView>;
+	public readonly items: ViewCollection<ListItemView | ListItemGroupView | ListSeparatorView>;
 
 	/**
 	 * Tracks information about DOM focus in the list.


packages/ckeditor5-ui/src/list/listitemgroupview.ts
```diff
@@ -108,7 +109,7 @@ export default class ListItemGroupView extends View {
 	 * Focuses the list item.
 	 */
 	public focus(): void {
-		if ( this.items.first ) {
+		if ( this.items.first && !( this.items.first instanceof ListSeparatorView ) ) {
 			this.items.first.focus();
 		}
```
mmotyczynska, line 114: By using the `focus()` on the `ListItemGroupView` we want to focus the first item. What if the first item is a `ListSeparatorView`? I think that maybe we should then focus the first item that is not a list separator, instead of not focusing anything. Then we should also add a unit test to verify this behaviour.

packages/ckeditor5-ui/src/list/listview.ts
```diff
@@ -141,7 +142,7 @@ export default class ListView extends View<HTMLUListElement> implements Dropdown
 		for ( const item of this.items ) {
 			if ( item instanceof ListItemGroupView ) {
 				this._registerFocusableItemsGroup( item );
-			} else {
+			} else if ( !( item instanceof ListSeparatorView ) ) {
 				this._registerFocusableListItem( item );
 			}
```
mmotyczynska, line 147: What if we add another type to this collection in the future? Then unfortunately this check would not be sufficient. So it might be better to explicitly check the needed class in `_registerFocusableListItem`, which is `ListItemView`. So instead of checking that it's not a separator, let's check if it's a list item.

mmotyczynska (changes requested):
I'd introduce small improvements to your solution, check the comments please 🙂

mmotyczynska (approved):
LGTM 👍

illia-stv (approved):

PR #15438, opened 2023-12-01
Unit tests for the dialogs system

oleq:

Tests (ui): Added unit tests for the dialogs system. 

---

FYI: The changes made to the find and replace in the main PR https://github.com/ckeditor/ckeditor5/pull/15285 have not been reflected in tests yet because we're waiting for design changes in the feature.

### DIFF
--- a/packages/ckeditor5-ui/tests/dialog/dialog.js
+++ b/packages/ckeditor5-ui/tests/dialog/dialog.js
@@ -1,0 +1,368 @@
+/**
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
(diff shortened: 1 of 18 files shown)


packages/ckeditor5-ui/tests/dialog/dialog.js
```diff
@@ -0,0 +1,368 @@
...
+						sinon.assert.calledOnce( _hideSpy );
+						sinon.assert.calledOnce( onHideSpy );
+					}, { priority: 'lowest' } );
+
+					dialogPlugin.fire( 'hide' );
+				} );
+			} );
+
+			describe( 'multiroot integration', () => {
+				it( 'should update the dialog view position whenever a root state is changed', () => {
```
oleq, line 127: We're missing a test that verifies that this does not happen upon root attribute change.

Dumluregn (approved):

Issue #14944, opened 2023-09-06
Support nested lists with the AdjacentListsSupport plugin


star-szr:
## 📝 Provide a description of the new feature

In #13945 it mentions that the initial scope of `AdjacentListsSupport` does not support nested lists.

For the project I'm working on, we have a requirement for adjacent nested lists. I have been digging to find where or how to update this since I could potentially contribute this feature.

I'm filing this issue to express interest in this feature and to see if anyone can point me in the right direction. From what I have seen so far code changes may need to happen outside of the scope of the `AdjacentListsSupport` plugin for this to work.

Using the manual tests, you can test the `AdjacentListsSupport` plugin using `documentlist-separator.html`. Example code to test:

```html
<ol>
  <li>One
    <ol>
      <li>1.1</li>
    </ol>
    <ol>
      <li>1.2</li>
    </ol>
  </li>
  <li>Two</li>
</ol>
```

Current result:

```html
<ol>
  <li>One
    <ol>
      <li>1.1</li>
    </ol>
  </li>
</ol>
<ol>
  <li>1.2</li>
</ol>
<ol>
  <li>Two</li>
</ol>
```

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

wimleers:
Drupal issue: https://www.drupal.org/project/drupal/issues/3402481

Witoso:
Let me change it to a bug.

Issue #10118, opened 2021-07-09
HTML comments: comments located between some HTML tags are repositioned


psmyrek:
## Provide a description of the task

Steps to reproduce:

1. Open `HtmlComment` manual test: http://localhost:8125/ckeditor5-html-support/tests/manual/htmlcomment.html.
2. Set the document content to:

```html
<figure class="table">
    <!-- C1 -->
    <table>
        <!-- C2 -->
        <tbody>
            <!-- C3 -->
            <tr>
                <!-- C4 -->
                <td>
                    <!-- C5 -->Table cell<!-- C6 -->
                </td>
                <!-- C7 -->
            </tr>
            <!-- C8 -->
        </tbody>
        <!-- C9 -->
    </table>
    <!-- C10 -->
</figure>
```

3. Call `editor.getData()`. A few comments, originally located between `<figure>` and `<table>`, `<table>` and `<tbody>` or `<tbody>` and `<tr>` , are repositioned at the end of the `</table>` tag.

```html
<figure class="table">
    <table>
        <tbody>
            <tr>
                <!-- C4 -->
                <td>
                    <!-- C5 -->Table cell<!-- C6 -->
                </td>
                <!-- C7 -->
            </tr>
        </tbody>
    </table>
    <!-- C10 --><!-- C1 --><!-- C9 --><!-- C8 --><!-- C3 --><!-- C2 -->
</figure>
```

It is also easily reproducible using `SourceEditing`:

https://user-images.githubusercontent.com/56868128/125051559-5aa53680-e0a3-11eb-9e48-06bd48553ea8.mp4

psmyrek:
Similar issue is with the `List` plugin, but in this case some comments are totally lost.

Example: after initializing the editor with the following content:

```html
<!-- c1 -->
<ol>
    <!-- c2 -->
    <li>
        <!-- c3 -->
        Ordered list item
        <!-- c4 -->
    </li>
    <!-- c5 -->
</ol>
<!-- c6 -->
<ul>
    <!-- c7 -->
    <li>
        <!-- c8 -->
        Unordered list item
        <!-- c9 -->
    </li>
    <!-- c10 -->
</ul>
<!-- c11 -->
```

the `editor.getData()` returns:

```html
<!-- c1 -->
<ol>
    <li>
        <!-- c3 -->
        Ordered list item
        <!-- c4 -->
    </li>
</ol>
<!-- c6 -->
<ul>
    <li>
        <!-- c8 -->
[...]

Issue #14767, opened 2023-08-07
📣 Upcoming changes to the lists feature


Witoso:
We wanted to inform you about some essential changes coming to the list feature in CKEditor 5.

## TL;DR

*   Over the years, we developed two independent list feature implementations.
    *   [`List`](https://ckeditor.com/docs/ckeditor5/latest/features/lists/lists.html) (later called List v1) 
    *   [`DocumentList`](https://ckeditor.com/docs/ckeditor5/latest/features/lists/document-lists.html#demo) (later called List v2).
*   List v2 is more future-proof and feaure-rich.
*   We need to sunset List v1. But there are small gaps in feature-parity, which we are fixing now.
*   This will be a breaking change to the majority of you, but the migration path will be trivial.
*   Stay tuned. No action needed yet.

## Context

We currently maintain two list features: [`List`](https://ckeditor.com/docs/ckeditor5/latest/features/lists/lists.html) (later called List v1) and [`DocumentList`](https://ckeditor.com/docs/ckeditor5/latest/features/lists/document-lists.html#demo) (later called List v2).

The List v1 feature was implemented in the early days of CKEditor 5. It supports “plain lists” – lists where `<li>` cannot contain block content (paragraphs, headings, tables, block images). It supports todo lists. But it does not support extending list markup via the [General HTML Support](https://ckeditor.com/docs/ckeditor5/latest/features/html/general-html-support.html) (GHS) feature.

The List v2 feature was implemented in 2022 to add support for block content in list items. It supports extending list markup via GHS. It does not support todo lists.

All development effort goes into List v2. The time has come to sunset List v1. Keeping two features that are almost the same is a waste of time and causes confusion for integrators. At the same time, we want to share new list features with a larger user base!

## Plan overview

*   \[DONE\] Stage 1. **Ensure feature completeness of List v2**
    *   Scope:
        *   Add support for todo lists.
        *   Add an option to configure List v2 to behave like List v1 (allow only inline content). It’s an edge case, but we aim at making sure that the migration path is trivial.
        *   Issue: #14632.
*   Stage 2. **Replacement (beginning of 2024, to be confirmed)**
    *   Scope:
        *   `List` plugin with be replaced with `DocumentList` implementation.
            *   Effect: after the update, without any action, you will start using List v2.
        *   `LegacyList` plugin will be released as a new separate plugin if someone wants to use the old lists implementation and have more time to migrate.
        *   `DocumentList` will still work, but will issue a deprecation warning.
[...]

PR #15443, opened 2023-12-01
Fix(list): Preserve HTML comments inside lists

star-szr:

Fix(list): Preserve HTML comments inside lists. See #10118.

---

### Additional information

This does not address repositioning at all, but does resolve the data loss portion of #10118, unless there are other similar cases that have not been found yet.

The behaviour with nested lists is not ideal (HTML comments will be nested inside a new paragraph outside of the list indentation), I believe this is related to #14944. In my testing this behaviour is consistent whether or not you have the AdjacentListsSupport plugin enabled.

Remaining TODO:
- [x] Write automated test coverage

---

### Credits

This work was sponsored by @ongov.

### DIFF
--- a/packages/ckeditor5-html-support/tests/manual/htmlcomment.html
+++ b/packages/ckeditor5-html-support/tests/manual/htmlcomment.html
@@ -10,14 +10,16 @@
 		</tr>
 	</table>
 	<ol>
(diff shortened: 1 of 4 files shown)


Witoso:
@star-szr thanks for taking a look at this! No need to add implementation for Lists v1 (they will be deprecated early next year, see: #14767)

PR #15481, opened 2023-12-07
Add a11y to the dialogs

Dumluregn:

Internal(ui): Improve dialogs accessibility.

---

### Additional information

As a side effect, the editor keystrokes are active while dialog is focused (e.g. `cmd+b`). This is coherent with the dropdown behaviour.


packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -173,9 +178,11 @@ export default class DialogView extends DraggableViewMixin( View ) implements Dr
 		super( locale );
 
 		const bind = this.bindTemplate;
+		const t = locale.t;
 
 		this.set( 'isVisible', false );
 		this.set( 'className', '' );
+		this.set( 'ariaLabel', t( 'Editor dialog' ) );
```
oleq, line 185: We're missing an entry in contexts.json.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -86,6 +87,27 @@ export default class Dialog extends Plugin {
...
 	}
 
+	/**
+	 * Initiates keystroke handler for toggling the focus between the editor and dialog view.
+	 */
+	private _initFocusToggler() {
+		const editor = this.editor;
+
+		editor.keystrokes.set( 'Ctrl+F6', ( data, cancel ) => {
+			if ( !this.isOpen || !this.view ) {
```
oleq, line 97: In what cases would a dialog be open but has no #view?

oleq (changes requested):
Looks good. I noticed that the keystroke keeps working for modals, though. I think moving the focus back to the editing root should not be possible in such a case.

Issue #5731, opened 2019-11-12
addTargetToLinks


KyooMario:
I tried config setting below but checked or not the target is _blank.
Please advise:

`link: {
        // Automatically add target="_blank" and rel="noopener noreferrer" to all external links.
        //addTargetToExternalLinks: true,

        // Let the users control the "download" attribute of each link.
        decorators: {
            addTargetToLinks: {
                mode: 'manual',
                label: 'Open in a new tab',
                attributes: {
                    target: '_blank',
                    rel: 'noopener noreferrer'
                }
            }
        }
    }`

Mgsy:
Hi! Can you describe your issue more precisely? If you ask about links in a link balloon (inside the editor) it will always be `target=_blank` and it's expected behaviour. 

If you call `editor.getData()`, you'll see that those attributes have been added properly.

KyooMario:
Thanks for the reply. We found out that in Edit mode the target is always _blank but in read mode it follows the expected behavior _self when disabled and _blank when enabled.

PR #15530, opened 2023-12-14
Internal: Fix error caused by eslint rule not having `ruleId`

filipsobol:

Internal: Fix error caused by eslint rule not having `ruleId`. Related to #5731.

---

### Additional information

### DIFF
--- a/scripts/eslint-formatter.js
+++ b/scripts/eslint-formatter.js
@@ -29,7 +29,7 @@ module.exports = results => {
 		}
 
 		return item.messages.some( message => {
-			return message.ruleId && message.ruleId.startsWith( 'ckeditor5-rules' );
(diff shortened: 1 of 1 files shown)


scripts/eslint-formatter.js
```diff
@@ -29,7 +29,7 @@ module.exports = results => {
 		}
 
 		return item.messages.some( message => {
-			return message.ruleId.startsWith( 'ckeditor5-rules' );
+			return message.ruleId && message.ruleId.startsWith( 'ckeditor5-rules' );
```
pomek, line 32: As we require Node@18, could we use optional chaining (available since @14)?

pomek (approved):
LGTM.

PR #15553, opened 2023-12-18
Proper keystrokes handling for F&R dialog

Dumluregn:

Internal(find-and-replace): Changed the handling of the `ctrl/cmd+F` keystroke to never close the find and replace dialog.


oleq (changes requested):
Keystroke support for in-dropdown integration needs to be reverted to the behavior on master. I provided Darek with details on Slack.

PR #15577, opened 2023-12-21
Dialogs: make the `close` button optional

Dumluregn:

Internal(ui): Made the `close` button in the header optional. Added the `source` information to the `DialogViewCloseEvent`.

---

### Additional information


packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -325,16 +325,21 @@ export default class DialogView extends DraggableViewMixin( View ) implements Dr
...
 	} ): void {
 		if ( title ) {
 			this.headerView = new FormHeaderView( this.locale, { icon } );
-			this.closeButtonView = this._createCloseButton();
-			this.headerView.children.add( this.closeButtonView );
+
+			if ( hasCloseButton ) {
+				this.closeButtonView = this._createCloseButton();
+				this.headerView.children.add( this.closeButtonView );
+			}
```
oleq, line 341: This breaks (shift)tab navigation in the modal. Or rather: it reveals a preexisting issue. The culprit is here and it does not have much in common with the close button:

https://github.com/ckeditor/ckeditor5/blob/f3e817576a858cad2c0e62174edd70234c4c3d21/packages/ckeditor5-ui/src/dialog/dialogview.ts#L634-L642

I suppose the thing is that if there's nothing else besides the actions view, this logic hijacks a standard (shift)tab cycling that would happen in the actions view (or any view with `FocusCycler`). It says: "OK you're on the last action button and you're pressing tab so you're about to focus the first action button but my forwardCycle logic wants to focus the next thing in the whole dialog (usually the close button)". The thing is: there's no next thing besides actions. So it does not do anything according to this code https://github.com/ckeditor/ckeditor5/blob/f3e817576a858cad2c0e62174edd70234c4c3d21/packages/ckeditor5-ui/src/focuscycler.ts#L249-L251 and because it `evt.stop();` it disallows the actions view to handle it on its own.

I think the solution would be checking whether `this._focusCycler.next` is the same view as `#current` and at lest preventing `evt.stop();`.

Let me know if you want to handle this in this PR or in a separate one. I guess this does not block the main functionality.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -189,6 +190,7 @@ export default class Dialog extends Plugin {
 		view.setupParts( {
 			icon,
 			title,
+			hasCloseButton,
```
oleq, line 193: I think we're missing a unit test that verifies that this property gets passed down from `show()` to `setupParts()`. Somewhere in or around `it( 'should setup the view parts with the passed arguments', () => {`?

oleq (changes requested):

PR #15606, opened 2023-12-29
Use serviceOrigin in image editing

### LINKED ISSUE #5834: Tab in the last editable area or after it should leave the editor
### ✔️ Expected result

Tab should not be completely blocked. I think it should allow you to move to the next field in the form.

### ❌ Actual result

Tab is blocked in the last editable area.

illia-stv:

Internal (ckbox): In on-premises should use serviceOrigin in image-editing. Closes [#5834](https://github.com/cksource/ckeditor5-commercial/issues/5834).

---

### Additional information


michalbilik:
I've checked this fix and it looks good 👍🏼.
Image editing with CKBox manual connected to on-prem `token` and `serviceOrigin` works correctly now.
SaaS Image editing also works without any issues.

arkflpc (approved):

PR #15623, opened 2024-01-03
FocusCycler's #next and #previous should point to the same view even if that's the only focusable view 

oleq:

Internal (ui): `FocusCycler#next` and `#previous` should point to the same view even if that's the only focusable view to allow for easier nested cyclers' integration. Previously they were set to `null`.


scofalik:
If I understand correctly, earlier, if there were multiple items, `next` and `previous` were cyclic (but also fired an extra event to chain cyclers). But with no items, it was always `null`. If that's the case, then I think the change is correct.

I want to understand if this works correctly with scenarios with chained focus cyclers. AFAIR, there could be, theoretically, three cases. I have a view + focus cycler, and:

1.  it has no focusable views + another view(s) with focus cycler(s);
2.  it has one focusable view + another view(s) with focus cycler(s);
3.  it has multiple focusable views + another view(s) with focus cycler(s).

If I understand correctly:

*   Case 1 is not applicable - we will focus something inside one of internal views and we will go around the internal focus cycler (or multiple of them if they are chained).
*   Case 2 would not work earlier because `previous` and `next` are null so appropriate event would not be sent?
*   Case 2 and 3 will now work correctly and behave the same.

Do we have any cases where we have a focusable view that also has focusable views? :thinking:

Do you believe this could be a breaking change for someone? Could somebody be checking for `next` or `previous` being `null`? Also, can `next` and `previous` still be `null` (I guess only if there are no views in the cycler).

oleq:
> If I understand correctly, earlier, if there were multiple items, next and previous were cyclic (but also fired an extra event to chain cyclers). But with no items, it was always null. If that's the case, then I think the change is correct.

Yes, this is how it worked.

> I want to understand if this works correctly with scenarios with chained focus cyclers. AFAIR, there could be, theoretically, three cases. I have a view + focus cycler, and:

I'm not sure if I understand correctly:

> 1. it has no focusable views + another view(s) with focus cycler(s);
> Case 1 is not applicable - we will focus something inside one of internal views and we will go around the internal focus cycler (or multiple of them if they are chained).

If at a certain point of its lifecycle, a view:
* has a focus cycler of its own,
* no views that have `focus()` method,
* but Another view that has focus cycler of its own,
 
then 

* that view has to have `focus()` method that executes its own `focusCycler#focusFirst/last()` 
* ...which in turn focuses that Another view (because this is the only one that also must have `focus()` method).
* in other words that view is `ViewWithFocusCycler` (also `FocusableView`) and the case is still valid. 
  * there's no way for a view to have a focus cycler and have no `focus()` method
* and view's `focusCycler#next` and `#previous` will always point to Another view after this PR

> it has one focusable view + another view(s) with focus cycler(s);
> Case 2 would not work earlier because previous and next are null so appropriate event would not be sent?

This case will not work after this PR but for a different reason. And the reason is: there's no automatic focus cycler chaining; it has never been developed and we agreed upon it. If you want chaining, the parent view must openly attach `forwardCycle` and `backwardCycle` listeners on the child's focus tracker and react accordingly. It works in dialogs because the dialog implementation has these listeners. It will not work anywhere else out of the box.

This PR is about something else: It's about control logic in `focusNext()` and `focusPrevious()` https://github.com/ckeditor/ckeditor5/blob/b4b57cacdf4663a76cc0be7b43ee392aea32099c/packages/ckeditor5-ui/src/focuscycler.ts#L236-L254. 

And mostly about this logic that does not fire `forwardCycle` if `next` is `null` but attempts to focus `null` which silently fails instead:

```
if ( next === this.first ) {
	this.fire<FocusCyclerForwardCycleEvent>( 'forwardCycle' );
} else {
	this._focus( next, 1 );
}
```

I could've rewritten it to something like `if ( next === this.first || next === null ) { this.fire( ... ) }` but I figured this would be magical. Instead, I decided that `next` should always have a value (whether this is the only focusable view, already focused or not) and I didn't touch this bit at all. 

Now because the even is fired correctly, the logic of `DialogView` can chain properly whether there's one view in there or more

https://github.com/ckeditor/ckeditor5/pull/15285/files#diff-b5ac09f46ee844cd7dc01643c39b3b37e95cd998b1843ae48ef770ad655ffcc6R629-R651

scofalik:
Thanks for claryfing the case 1.

As for case 2 -- yes, I know it won't happen automatically and we base on events. But earlier, the events would not have been fired, and now they are, right?

oleq:
> As for case 2 -- yes, I know it won't happen automatically and we base on events. But earlier, the events would not have been fired, and now they are, right?

Yes, they should be.

packages/ckeditor5-ui/src/focuscycler.ts
```diff
@@ -327,7 +327,7 @@ export default class FocusCycler extends EmitterMixin() {
 			index = ( index + collectionLength + step ) % collectionLength;
 		} while ( index !== current );
 
-		return null;
+		return this.focusables.get( this.current! ) as FocusableView;
```
scofalik, line 330: We had some F2F discussion about this loop, whether it can go infinite, etc. We also did some changes to types in other PR and we discussed that `focusables` should indeed always have only `FocusableView`s.

I wanted to drop `index !== current` condition from the loop so that `current` is also considered so this return will not need to change.

However, the loop can go infinite if all focusables are hidden.

But in this case, if all focusables are hidden (including current) should we return current or `null`?

We could change this to return `null` if current is not focusable but I am not 100% sure if that's correct -- I am not sure if that would work correctly with some chained focus cyclers etc. So I leave this thought to you, please proceed as you like.

oleq: I feel an instant aversion to unrestricted loops 🙂 So instead I went with 

```ts
let index = ( current + collectionLength + step ) % collectionLength;

for ( let i = 0; i++ < collectionLength ) {
	const view = this.focusables.get( index )!;

	if ( isFocusable( focusableItemCandidate ) ) {
		return view;
	}

	index = ( index + collectionLength + step ) % collectionLength;
}

// TS needs this...
return null
```

which was OK if not for the fact that TS does not understand that in this loop we'll always get back to the current (initial) view that is focusable. I had to add `null` below just to satisfy TS it and I think it's misleading for anyone who'll read this code in the future.

Finally I went with the `do {} while ()` implementation but with slight refactoring and some comments [780a93e](https://github.com/ckeditor/ckeditor5/pull/15623/commits/780a93e1ba4bc75eddbe290d8c35e6bb66211445).

PR #15629, opened 2024-01-04
Docs for the dialogs

Dumluregn:

Docs: Added the entries about the dialog component.

---

### Additional information

### DIFF
--- a/docs/framework/architecture/ui-components.md
+++ b/docs/framework/architecture/ui-components.md
@@ -2,7 +2,7 @@
 category: framework-architecture
 meta-title: UI Components | CKEditor 5 Framework Documentation
 order: 50
-modified_at: 2023-03-06
+modified_at: 2024-01-03
 ---
 
 # UI components
@@ -469,6 +469,221 @@ disabledDropdown.render();
 
 document.getElementById( 'dropdown-disabled' ).append( disabledDropdown.element );
 ```
+
+## Dialog
+
+{@snippet framework/ui/ui-dialog}
+
+A dialog window is a draggable popup that can be displayed on top of the editor contents and remains open while user interacts with the editing area. You can use it to display any detached UI. Dialogs are brought by the {@link module:ui/dialog/dialog~Dialog `Dialog`} plugin.
+
+```js
+// Necessary imports. Remember to install the packages first.
+import {
+	ButtonView,
+	Dialog,
+	View
+} from '@ckeditor/ckeditor5-ui';
+import { Plugin } from '@ckeditor/ckeditor5-core';
+import { ClassicEditor } from '@ckeditor/ckeditor5-editor-classic';
+import { Paragraph } from '@ckeditor/ckeditor5-paragraph';
+import { Essentials } from '@ckeditor/ckeditor5-essentials';
+import { Bold, Italic } from '@ckeditor/ckeditor5-basic-styles';
+
+// Create a plugin that brings a button that toggles the visibility of a dialog window.
+// Read more about creating the plugins here: https://ckeditor.com/docs/ckeditor5/latest/installation/plugins/plugins.html.
+class MinimalisticDialog extends Plugin {
+	// Make sure the Dialog plugin is loaded.
+	get requires() {
+		return [ Dialog ];
+	}
+
+	init() {
+		// Add a button to the component factory so it is available for the editor.
+		this.editor.ui.componentFactory.add( 'showDialog', locale => {
+			const buttonView = new ButtonView( locale );
+
+			buttonView.set( {
+				label: 'Show a dialog',
+				tooltip: true,
+				withText: true
+			} );
+
+			// Define button behavior on press.
+			buttonView.on( 'execute', () => {
+				const dialog = this.editor.plugins.get( 'Dialog' );
+
+				// If button is turned on, hide the dialog.
+				if ( buttonView.isOn ) {
+					dialog.hide();
+					buttonView.isOn = false;
+
+					return;
+				}
+
+				buttonView.isOn = true;
+
+				// Otherwise, show the dialog.
+				// Create a view with some simple content. It will be displayed as a dialog's body.
+				const textView = new View( locale );
+
+				textView.setTemplate( {
+					tag: 'div',
+					attributes: {
+						style: {
+							padding: 'var(--ck-spacing-large)',
+							whiteSpace: 'initial',
+							width: '100%',
+							maxWidth: '500px'
+						},
+						tabindex: -1
+					},
+					children: [
+						'This is the content of the dialog.',
+						'You can put here text, images, inputs, buttons, etc.'
+					]
+				} );
+
+				// Tell the plugin to display a dialog with title, content and one action button.
+				dialog.show( {
+					title: 'Dialog with text',
+					content: textView,
+					actionButtons: [
+						{
+							label: 'OK',
+							class: 'ck-button-action',
+							withText: true,
+							onExecute: () => dialog.hide()
+						}
+					],
+					onHide() { buttonView.isOn = false; }
+				} );
+			} );
+
+			return buttonView;
+		} );
+	}
+}
+
+// Create an editor instance. Remember to have an element with `[id="editor"]` attribute in the document.
+ClassicEditor
+	.create( document.querySelector( '#editor' ), {
+		plugins: [ Essentials, Paragraph, Bold, Italic, MinimalisticDialog, Dialog ],
+		toolbar: [ 'bold', 'italic', '|', 'showDialog' ]
+	} )
+	.catch( error => {
+		console.error( error.stack );
+	} );
+```
+
+See the guide about the {@link framework/architecture/ui-library#dialogs-and-modals API of the dialogs system}.
+
+## Modal
+
+{@snippet framework/ui/ui-modal}
+
+Modal is a specific kind of a dialog window which while open, doesn't allow to interact with the editor content - it has to be closed first. It can be used to enforce the user interaction or interrupt them in some important situations. To create a modal, use the optional {@link module:ui/dialog/dialog~DialogDefinition#isModal `isModal`} property of the {@link module:ui/dialog/dialog~Dialog#show `Dialog#show()`} method.
+
+```js
+// Necessary imports. Remember to install the packages first.
+import {
+	ButtonView,
+	Dialog,
+	View
+} from '@ckeditor/ckeditor5-ui';
+import { Plugin } from '@ckeditor/ckeditor5-core';
+import { ClassicEditor } from '@ckeditor/ckeditor5-editor-classic';
+import { Paragraph } from '@ckeditor/ckeditor5-paragraph';
+import { Essentials } from '@ckeditor/ckeditor5-essentials';
+import { Bold, Italic } from '@ckeditor/ckeditor5-basic-styles';
+
+// Create a plugin that brings a button that toggles the visibility of a modal window.
+// Read more about creating the plugins here: https://ckeditor.com/docs/ckeditor5/latest/installation/plugins/plugins.html.
+class MinimalisticModal extends Plugin {
+	// Make sure the Dialog plugin is loaded.
+	get requires() {
+		return [ Dialog ];
+	}
+
+	init() {
+		// Add a button to the component factory so it is available for the editor.
+		this.editor.ui.componentFactory.add( 'showModal', locale => {
+			const buttonView = new ButtonView( locale );
+
+			buttonView.set( {
+				label: 'Show a modal',
+				tooltip: true,
+				withText: true
+			} );
+
+			// Define button behavior on press.
+			buttonView.on( 'execute', () => {
+				const dialog = this.editor.plugins.get( 'Dialog' );
+
+				// If button is turned on, hide the modal.
+				if ( buttonView.isOn ) {
+					dialog.hide();
+					buttonView.isOn = false;
+
+					return;
+				}
+
+				buttonView.isOn = true;
+
+				// Otherwise, show the modal.
+				// First, create a view with some simple content. It will be displayed as a dialog's body.
+				const textView = new View( locale );
+
+				textView.setTemplate( {
+					tag: 'div',
+					attributes: {
+						style: {
+							padding: 'var(--ck-spacing-large)',
+							whiteSpace: 'initial',
+							width: '100%',
+							maxWidth: '500px'
+						},
+						tabindex: -1
+					},
+					children: [
+						'This is a sample content of the modal.',
+						'You can put here text, images, inputs, buttons, etc.'
+					]
+				} );
+
+				// Tell the plugin to display a modal with title, content and one action button.
+				dialog.show( {
+					isModal: true,
+					title: 'Modal with text',
+					content: textView,
+					actionButtons: [
+						{
+							label: 'OK',
+							class: 'ck-button-action',
+							withText: true,
+							onExecute: () => dialog.hide()
+						}
+					],
+					onHide() { buttonView.isOn = false; }
+				} );
+			} );
+
+			return buttonView;
+		} );
+	}
+}
+
+// Create an editor instance. Remember to have an element with `[id="editor"]` attribute in the document.
+ClassicEditor
+	.create( document.querySelector( '#editor' ), {
+		plugins: [ Essentials, Paragraph, Bold, Italic, MinimalisticDialog, Dialog ],
+		toolbar: [ 'bold', 'italic', '|', 'showModal' ]
+	} )
+	.catch( error => {
+		console.error( error.stack );
+	} );
+```
+
+See the guide about the {@link framework/architecture/ui-library#dialogs-and-modals API of the dialogs system}.
 
 ## Icons
 
--- a/docs/framework/architecture/ui-library.md
+++ b/docs/framework/architecture/ui-library.md
@@ -2,6 +2,8 @@
 category: framework-architecture
 meta-title: UI library | CKEditor 5 Framework Documentation
 order: 40
+toc-limit: 3
+modified_at: 2024-01-03
 ---
 
 # UI library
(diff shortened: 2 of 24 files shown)


docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,159 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+dialog.show( {
+	icon: pencilIcon,
+	title: 'My first dialog',
+	hasCloseButton: false
+} );
+```
+
+##### Contents
+
+This part can be a single [view](#views) or a collection of views. They will be displayed directly inside a body of a dialog.
```
oleq, line 527: What I'm missing here is an example of how to inject any view into the dialog. The example in the first section is not helpful because it does not bring any content. I'd either extend the initial example or create another one with some content (e.g. simple paragraph view with some text) here.

The whole point of docs is to give developers a minimal snippet to achieve something at least somehow relateable to their goals. One thing I'm sure they'll want to display _something_ in their dialog.

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,159 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+Dialogs provide full keyboard accessiblity. While a dialog is open, press the <kbd>Ctrl</kbd>+<kbd>F6</kbd> combination to move the focus between the editor and the dialog. It can also be closed at any time by pressing the <kbd>Esc</kbd> key. To navigate through the dialog, use the <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd> keystrokes. The contents of the dialog are also available for screen readers.
+
+#### API
+
+The dialog's lifecycle (creating and destroying) is managed by the {@link module:ui/dialog/dialog Dialog plugin}. It provides two public methods: {@link module:ui/dialog/dialog~Dialog#show `show()`} and {@link module:ui/dialog/dialog~Dialog#hide `hide()`}. Both of them fire the respective events ({@link module:ui/dialog/dialog~DialogShowEvent `show`} and {@link module:ui/dialog/dialog~DialogHideEvent `hide`}), so it's possible to hook after or before them.
+
+##### `show()` method
+
+This method accepts a number of configuration options that allow to shape the structure and behaviour of the dialog.
+
```
oleq, line 581: DialogDefinition is worth mentioning here.

docs/framework/architecture/ui-components.md
```diff
@@ -470,6 +470,205 @@ disabledDropdown.render();
...
+
+function hideDialog( dialog ) {
+	dialog.contentView.reset();
+	dialog.destroy();
+	document.body.removeChild( dialog.element );
+	dialogButton.isOn = false;
+}
+```
+
+There are a few configurable positions for the dialog display:
```
oleq, line 565: ~This should link to API.~

See my next comment.

docs/framework/architecture/ui-components.md
```diff
@@ -470,6 +470,205 @@ disabledDropdown.render();
...
+dialogButton.on( 'execute', () => {
+	if ( dialogButton.isOn ) {
+		hideDialog( dialog );
+
+		return;
+	}
+
+	dialogButton.isOn = true;
+
+	dialog = new DialogView( locale, {
```
oleq, line 504: I'm not sure if this whole snippet makes any sense. `DialogView` has never meant to be used without the `Dialog` plugin. It's quite obvious it's missing specific editor couplings such as viewport offset or DOM roots. We never meant developers to use dialogs this way.

The examples (snippets) presented in this guide should be copy-pasteable pieces of code developers can use in their integrations. What you presented here is more or less how to build your own dialogs system using some pieces of our dialogs system :D Not saying it's impossible, but this is not the way go.

Either we create an editor and showcase how to use the dialog plugin properly here or I say we don't do anything in this guide. Notice how silent it is on the contextual balloon plugin (that also makes no sense without an editor). `BalloonPanelView`, however, unlike `DialogView`, can be used standalone and does not need contextual balloon plugin to be useful (that's why it got a section here).

packages/ckeditor5-find-and-replace/docs/_snippets/features/find-and-replace-dropdown.js
```diff
@@ -0,0 +1,55 @@
...
+			forceDemoLabel: true
+		},
+		findAndReplace: {
+			uiType: 'dropdown'
+		}
+	} )
+	.then( editor => {
+		window.editor = editor;
+
+		window.attachTourBalloon( {
```
oleq, line 43: The first (in-dialog) demo is missing a tour balloon. Both editor should have one.

oleq (changes requested):

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,170 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+This method accepts a number of configuration options that allow to shape the structure and behavior of the dialog.
+
+##### `show:[id]` event
+
+When the {@link module:ui/dialog/dialog~Dialog#show `Dialog#show()`} function is called and the {@link module:ui/dialog/dialog~DialogDefinition#id `id`} parameter was provided, the fired event will be namespaced, allowing to customize some of the default dialog's properties. For example, you can change the default position of the "Find and replace" dialog from the editor corner to the bottom with the following code:
+
+```js
+dialog.on( 'show:findAndReplace', ( evt, data ) => {
+	return Object.assign( data, { position: DialogViewPosition.EDITOR_CENTER } );
```
scofalik, line 585: `return` is not needed here right? :thinking:

scofalik: Description says "to the bottom" but used position moves it to the center ;).

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,170 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+
+Dialogs provide full keyboard accessibility. While a dialog is open, press the <kbd>Ctrl</kbd>+<kbd>F6</kbd> combination to move the focus between the editor and the dialog. It can also be closed at any time by pressing the <kbd>Esc</kbd> key. To navigate through the dialog, use the <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd> keystrokes. The contents of the dialog are also available for screen readers.
+
+#### API
+
+The dialog's lifecycle (creating and destroying) is managed by the {@link module:ui/dialog/dialog Dialog plugin}. It provides two public methods: {@link module:ui/dialog/dialog~Dialog#show `show()`} and {@link module:ui/dialog/dialog~Dialog#hide `hide()`}. Both of them fire the respective events ({@link module:ui/dialog/dialog~DialogShowEvent `show`} and {@link module:ui/dialog/dialog~DialogHideEvent `hide`}), so it's possible to hook after or before them.
+
+##### `show()` method
+
+This method accepts a number of configuration options that allow to shape the structure and behavior of the dialog.
```
scofalik: Same way as you mentioned in `hide()` description, it's good to add here that this each time creates a new instance of a dialogview and shows it.

docs/framework/architecture/ui-library.md
scofalik: I'd say that an example is worth mentioning here.

In general, I am missing one, two or three nice examples. Maybe something simple, then the one with checkbox that I mentioned earlier.

packages/ckeditor5-find-and-replace/docs/features/find-and-replace.md
```diff
@@ -11,14 +11,33 @@ The find and replace feature lets you find and replace any text in your document
...
 
 <info-box info>
 	This demo only presents a limited set of features. Visit the {@link examples/builds/full-featured-editor feature-rich editor example} to see more in action.
 </info-box>
 
+## Configuration
+
+### Configuring the UI type
+
+By default, the find and replace form is displayed inside a dialog. That allows for keeping it open while editing the document at the same time. It is still possible to use the feature in a dropdown. To do this, use the {@link module:find-and-replace/findandreplaceconfig~FindAndReplaceConfig `config.findAndReplace.uiType`} configuration option:
```
scofalik, line 26:
```suggestion
By default, the find and replace form is displayed inside a dialog. That allows for keeping it open while editing the document at the same time. Alternatively, it still possible to display the feature in a dropdown. To do this, use the {@link module:find-and-replace/findandreplaceconfig~FindAndReplaceConfig `config.findAndReplace.uiType`} configuration option:
```

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,170 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+```js
+import { Dialog, ButtonView } from '@ckeditor/ckeditor5-ui';
+
+class MyPluginWithDialog extends Plugin {
+	public static get requires() {
+		return [ Dialog ] as const;
+	}
+
+	public init(): void {
+		this.editor.ui.componentFactory.add( 'aiAssistant', locale => {
```
scofalik: Change `aiAssistant` to something else.

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,170 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
 );
 ```
 
+### Dialogs and modals
+
+Another UI component provided by the framework is a dialog. The dialog system in CKEditor 5 is brought by the {@link module:ui/dialog/dialog~Dialog `Dialog` plugin} that offers API for displaying [views](#views) in dialogs. In a sense, this plugin corresponds to another that manages views in balloons (popovers) across the UI ({@link module:ui/panel/balloon/contextualballoon~ContextualBalloon `ContextualBalloon` plugin}).
+
+Dialog is a popup window that does not close when the user clicks outside of it. It allows for interacting with the editor and its content while being open (unless it is a modal, which blocks the interaction with the rest of the page until it is closed). When configured to display a [header](#header), a dialog is also {@link module:ui/bindings/draggableviewmixin~DraggableViewMixin draggable} using mouse or touch. Only one dialog can be open at a time - opening another one closes the previously visible one.
+
+A minimal plugin creating a toolbar button component showing an empty dialog can look like this:
```
scofalik:
```suggestion
A minimal plugin that adds a toolbar button component which shows an empty dialog could look like this:
```

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,170 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+
+##### Header
+
+A header may consist of any combination of three elements:
+
+* the icon,
+* the title,
+* the "Close" button.
+
+The "Close" button is added automatically as long as you provide an icon or a title. The code below shows how to create display a dialog with a header that comes with an icon and a title, but without the "Close" button.
```
scofalik, line 478:
```suggestion
By default, the "Close" button ("X") is added to the header as long as you provide an icon or a title. To hide it, set `hasCloseButton` flag to `false`:
```

scofalik: Do we want to add:

*   some kind of info that if you hide the close button you should add a way to close the dialog (kind of obvious, I know),
*   somewhere information that the dialog can be closed using `Esc` key (and whether it still true if `hasCloseButton` is false),
    *   okay I see we have the info about `Esc` key in a11y section.
*   info how to prevent closing on `Esc`?
    *   or how to prevent closing in general (especiall in regards to modals)
    *   of info that this is a bad practice
*   info if model can be closed using `Esc`?
    *   this and above points are related

docs/framework/architecture/ui-library.md
scofalik: +1. At least add some simple view with some text, like in manual tests.

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,170 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+			withText: true,
+			onExecute: () => {
+				dialog.view!.headerView!.label = 'New title';
+			}
+		},
+		{
+			label: 'Drag detector'
+			withText: true,
+			onCreate: buttonView => {
+				buttonView.bind( 'isEnabled' ).to( dialog.view!, 'wasMoved', wasMoved => !wasMoved );
```
scofalik, line 555: In other PR I asked whether `wasMoved` must be observable. I wonder if they are some real use cases where we need to react to the dialog `wasMoved` change (outside of the dialog itself).

scofalik: If you want to change this example, you can add an example where button is disabled for some time after the dialog is opened.

Or even better example would be that the user has to click some kind of "I agree" checkbox and the button is active only then. But this will make this snippet less clear.

docs/framework/architecture/ui-library.md
scofalik: I'd also add an information, here or somewhere, that the dialog size will depend on the view passed to it. Dialog itself does not have any settings related to size, so it is up to the integrator to create a view with a proper size.

Also, if the internal view size has changed it may be necessary to call `updatePositon()` -- I'd mention that. Unless we make this automatic. Which would be great.

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,170 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+```
+
+#### Accessibility
+
+Dialogs provide full keyboard accessibility. While a dialog is open, press the <kbd>Ctrl</kbd>+<kbd>F6</kbd> combination to move the focus between the editor and the dialog. It can also be closed at any time by pressing the <kbd>Esc</kbd> key. To navigate through the dialog, use the <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd> keystrokes. The contents of the dialog are also available for screen readers.
+
+#### API
+
+The dialog's lifecycle (creating and destroying) is managed by the {@link module:ui/dialog/dialog Dialog plugin}. It provides two public methods: {@link module:ui/dialog/dialog~Dialog#show `show()`} and {@link module:ui/dialog/dialog~Dialog#hide `hide()`}. Both of them fire the respective events ({@link module:ui/dialog/dialog~DialogShowEvent `show`} and {@link module:ui/dialog/dialog~DialogHideEvent `hide`}), so it's possible to hook after or before them.
+
```
scofalik, line 574: It could be just in API docs but maybe it would be worth mentioning that the dialog is opened on `normal` priority and that `onShow` and `onHide` callbacks are called on `low` priority.

Also, now as I think of it, do we even need `onShow` and `onHide` if we have `show` and `hide` events? I guess they are simpler to use but I am getting redundancy smell. And then somebody reads guide and API docs and is confused when they should use this and when that. WDYT? cc @oleq

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,170 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+		},
+		{
+			label: 'Cancel',
+			withText: true,
+			onExecute: () => dialog.hide()
+		}
+	]
+} );
+```
+
```
scofalik, line 566: I'd add a section talking about `onShow` and `onHide`. `onShow` could be with views actually, because you will usually set up something in the view after it is created, but not always, so maybe it can be in separate section.

scofalik (changes requested):

docs/framework/architecture/ui-library.md
oleq: At some point, we decided to have both:
* callbacks for convenience: 90% of users will be satisfied by them and they won't need to dive into our events system, they are straightforward and correspond to the API in other dialog libraries.
* events for power users who want more

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,355 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+<info-box>
+	If you decide to hide the "Close" button, remember to leave some other way(s) to close the dialog (like pressing the <kbd>Esc</kbd> key).
+</info-box>
+
+##### Content
+
+This part can be a single [view](#views) or a collection of views. They will be displayed directly inside a body of a dialog. You can find an example of how to insert a block of text into the dialog.
```
scofalik, line 511:
```suggestion
This part can be any single [view](#views) or a collection of views. They will be displayed directly inside a body of a dialog. Below, you can find an example of how to insert a block of text into the dialog.
```

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,355 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+} );
+```
+
+##### Action buttons
+
+The last piece of the dialog is the actions area, where the buttons are displayed. Their behaviour can be fully customized using the `onCreate()` and `onExecute()` callbacks. Below you can find an example of the configuration for the four custom buttons:
+
+* The "OK" button that closes the dialog and has a custom CSS class set,
+* The "Set custom title" button that changes the dialog's title,
+* The "Drag detector" button that changes its state once the dialog was dragged by the user.
```
scofalik, line 547: Outdated description - the button was changed to something else.

docs/framework/architecture/ui-library.md
```diff
@@ -439,6 +441,355 @@ dropdownView.bind( 'isEnabled' ).toMany( buttons, 'isEnabled',
...
+		},
+		{
+			label: 'Set custom title',
+			withText: true,
+			onExecute: () => {
+				dialog.view!.headerView!.label = 'New title';
+			}
+		},
+		{
+			label: 'This button will be disabled in 3...'
```
scofalik, line 569: I'd go the other way -- make it enabled in 3, 2, 1. It used to be a pattern in dialogs some time ago (so you won't click by mistake). And when enabled, change label to something like "Enabled after 3 seconds" or simply "Button".

scofalik (changes requested):
I haven't checked the whole docs again, here are some of my notes from earlier suggestions that were marked as resolved. Let's make sure that they are either included or we consciously decided to skip them:

> Do we want to add:
> 
> *   some kind of info that if you hide the close button you should add a way to close the dialog (kind of obvious, I know),
> *   somewhere information that the dialog can be closed using Esc key (and whether it still true if hasCloseButton is false),
>     *   okay I see we have the info about Esc key in a11y section.
> *   info how to prevent closing on Esc?
>     *   or how to prevent closing in general (especiall in regards to modals)
>     *   of info that this is a bad practice
> *   info if model can be closed using Esc?
>     *   this and above points are related

> I'd also add an information, here or somewhere, that the dialog size will depend on the view passed to it. Dialog itself does not have any settings related to size, so it is up to the integrator to create a view with a proper size.
> 
> Also, if the internal view size has changed it may be necessary to call updatePositon() -- I'd mention that. Unless we make this automatic. Which would be great.

oleq:
> Do we want to add:
> 
> *   some kind of info that if you hide the close button you should add a way to close the dialog (kind of obvious, I know),

Covered by 
* https://github.com/ckeditor/ckeditor5/blob/97ab0c352432563e4fe3831c7b3a88f5ac9c984f/docs/framework/architecture/ui-library.md#L505-L507
* https://github.com/ckeditor/ckeditor5/blob/46dfb74dde13928064b4efcfc749ba76581b365d/packages/ckeditor5-ui/src/dialog/dialog.ts#L381-L387

> *   somewhere information that the dialog can be closed using Esc key (and whether it still true if hasCloseButton is false),

Same as above

>     *   okay I see we have the info about Esc key in a11y section.
> *   info how to prevent closing on Esc?

Covered by https://github.com/ckeditor/ckeditor5/blob/46dfb74dde13928064b4efcfc749ba76581b365d/docs/framework/architecture/ui-library.md#L697-L705

>     *   or how to prevent closing in general (especiall in regards to modals)
>     *   of info that this is a bad practice

Handled by https://github.com/ckeditor/ckeditor5/blob/46dfb74dde13928064b4efcfc749ba76581b365d/docs/framework/architecture/ui-library.md#L722-L724

https://github.com/ckeditor/ckeditor5/blob/46dfb74dde13928064b4efcfc749ba76581b365d/docs/framework/architecture/ui-library.md#L472-L474

> *   info if model can be closed using Esc?

I'm not sure if this is relevant to integrators.

>     *   this and above points are related

> I'd also add an information, here or somewhere, that the dialog size will depend on the view passed to it. Dialog itself does not have any settings related to size, so it is up to the integrator to create a view with a proper size.

Addressed in https://github.com/ckeditor/ckeditor5/blob/bab8a8e54d628eaedad91e7abb6cba99f8a7cc0d/docs/framework/architecture/ui-library.md#L541

> 
> Also, if the internal view size has changed it may be necessary to call updatePositon() -- I'd mention that. Unless we make this automatic. Which would be great.

Addressed in https://github.com/ckeditor/ckeditor5/blob/69437a8639e4baa8c5c8392da66aa926742efd9c/docs/framework/architecture/ui-library.md#L795-L799.

oleq (approved):

PR #15632, opened 2024-01-04
FocusCycler's #focusables collection should only contain FocusableView instances

### LINKED ISSUE #14973: Dialog/modal system for the CKEditor 5

oleq:

Other (ui): `FocusCycler`'s `#focusables` collection should only contain `FocusableView` instances. Closes #14973.

MINOR BREAKING CHANGE (ui): The view collection required by [`FocusCycler#constructor()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusCycler.html#function-constructor) must only contain views implementing the [`FocusableView`](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_focuscycler-FocusableView.html) interface. Failing to do so will result in a TypeScript error. Please make sure each and every view passed to this collection implements the `focus()` method.

---

### Additional information

**Requires https://github.com/cksource/ckeditor5-commercial/pull/5886**

### DIFF
--- a/packages/ckeditor5-image/src/imageinsert/ui/imageinsertformview.ts
+++ b/packages/ckeditor5-image/src/imageinsert/ui/imageinsertformview.ts
@@ -14,7 +14,8 @@ import {
 	FocusCycler,
 	CollapsibleView,
 	type FocusCyclerForwardCycleEvent,
-	type FocusCyclerBackwardCycleEvent
+	type FocusCyclerBackwardCycleEvent,
+	type FocusableView
 } from 'ckeditor5/src/ui.js';
 import { FocusTracker, KeystrokeHandler, type Locale } from 'ckeditor5/src/utils.js';
 
@@ -39,7 +40,7 @@ export default class ImageInsertFormView extends View {
 	/**
 	 * A collection of views that can be focused in the form.
 	 */
-	protected readonly _focusables: ViewCollection;
+	protected readonly _focusables: ViewCollection<FocusableView>;
 
 	/**
 	 * Helps cycling over {@link #_focusables} in the form.
@@ -49,7 +50,7 @@ export default class ImageInsertFormView extends View {
 	/**
 	 * A collection of the defined integrations for inserting the images.
 	 */
-	private readonly children: ViewCollection;
+	private readonly children: ViewCollection<FocusableView>;
 
 	/**
 	 * Creates a view for the dropdown panel of {@link module:image/imageinsert/imageinsertui~ImageInsertUI}.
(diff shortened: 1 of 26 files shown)


packages/ckeditor5-ui/src/collapsible/collapsibleview.ts
scofalik, line 1: Does it make sense to add `implements FocusableView` to `CollapsibleView`? Or in that case we should add it to many more files and you didn't want to do that? :D.

packages/ckeditor5-image/src/imageinsert/ui/imageinsertformview.ts
```diff
@@ -49,15 +50,15 @@ export default class ImageInsertFormView extends View {
...
+	private readonly children: ViewCollection<FocusableView>;
 
 	/**
 	 * Creates a view for the dropdown panel of {@link module:image/imageinsert/imageinsertui~ImageInsertUI}.
 	 *
 	 * @param locale The localization services instance.
 	 * @param integrations An integrations object that contains components (or tokens for components) to be shown in the panel view.
 	 */
-	constructor( locale: Locale, integrations: Array<View> = [] ) {
+	constructor( locale: Locale, integrations: Array<FocusableView> = [] ) {
```
scofalik, line 61: Is there a scenario where integrations are not `FocusableView`s? If yes, then maybe we should filter `integrations` when we add them to `this._focusables`?

scofalik: I see that you already applied such filtering in some places in this PR.

scofalik (approved):
LGTM. I added some questions.

Do we want to merge this as MINOR BREAKING CHANGE?

packages/ckeditor5-ui/src/collapsible/collapsibleview.ts
oleq: > Does it make sense to add implements FocusableView to CollapsibleView? Or in that case we should add it to many more files and you didn't want to do that? :D.

Collapsible is a generic container. It can contain both focusable and non-focusable children. In this matter, it is very similar to the toolbar that hosts both focusables (buttons) and non-focusables (e.g. separators).

packages/ckeditor5-image/src/imageinsert/ui/imageinsertformview.ts
oleq: We require these views to be focusable already on master. My changes only reflect this  

https://github.com/ckeditor/ckeditor5/blob/b5c31df8ca7f2032256513532449f424bbc023dd/packages/ckeditor5-image/src/imageinsert/imageinsertui.ts#L252-L257

https://github.com/ckeditor/ckeditor5/blob/f0a702cf32fa47d245b03512417d89abd7f45da7/packages/ckeditor5-image/src/imageinsert/imageinsertui.ts#L183-L188

oleq:
> Do we want to merge this as MINOR BREAKING CHANGE?

I think so. Before these changes, developers were able to populate FocusCycler's focusable collection with non-focusable views. It worked because the system has JS mechanisms built-in to work around this. After this PR, their TS projects will show errors and they'll need to stop adding these (invalid) views there. It may involve some conditions or some filtering but still - there will be some action required.

PR #15649, opened 2024-01-05
Introduced `DiffItemInsert#element` and `DiffItemRemove#element`

scofalik:

Internal (engine): Introduced `DiffItemInsert#_element` and `DiffItemRemove#_element`.

---

### Additional information

Introduced `element` properties in diff items to allow more precise changes detection. In particular, we need to be able to check whether given change stored in `Differ` was an actual insert/remove or was it a rename. This is important for the correct behavior of the new attribute suggestions mechanism. Currently, there's no way to say with 100% certainty that given element was renamed or maybe one element was removed and another was inserted at the same place.

The change will also make it easier if you want to track that given particular element (that you observe, for any reason) was removed. Earlier you would need to check changes in the graveyard and check it by `DiffItemInsert#position`.


scofalik:
One controversial change is to add `element` to `DiffItemRemove` which we might want to avoid as it may lead to incorrect use. I added docs there.

A different approach would be to only change `DiffItemInsert` and add `details` property, e.g. like this:

```
details: {
    type: 'insert' | 'rename' | 'refresh' (possibly also move, split and merge?),
    before: {
    	name: string,
    	attributes:
    }
}
```

Where before is the snapshot of the item before changes (the same what is now available at `DiffItemRemove#name` and `DiffItemRemove#attributes`).

niegowski:
> One controversial change is to add `element` to `DiffItemRemove` which we might want to avoid as it may lead to incorrect use. I added docs there.

We should not add an element reference there.
1. This could lead to misuse and potential graveyard modifications. I know there is a code comment but developers might not notice this.
2. It would be inconsistent, we expose the name and attributes map to avoid directly accessing the model element
3. It's unknown when the `element` field will be available. What about text nodes?

The problem this PR is trying to resolve is the rename case. I remember there are problems with renamed elements. The one I can recall is the `Mapper` that keeps mappings to the view element based on the model element reference.

I'd suggest using the simplest possible approach (without any object references) and marking this new property as an internal API.

niegowski (approved):

PR #15658, opened 2024-01-08
Font color and font background color throw error while having buttons in the main toolbar and the balloon toolbar 

### LINKED ISSUE #15580: Font color and Font background color throw error while having buttons in the main toolbar and the balloon toolbar

1.  Create a classic editor with the [Balloon toolbar plugin](https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_toolbar_balloon_balloontoolbar-BalloonToolbar.html).
2.  Load Font color and Font background color plugins.
3.  Add `fontColor` and `fontBackgroundColor` buttons to the main toolbar and the balloon toolbar configuration.
4.  Run the editor and select some text.
5.  Click on the font color or font background color buttons within the balloon toolbar.

### ✔️ Expected result

_There is no error in the console._

### ❌ Actual result

_The error is thrown._

```
ckeditorerror.js:90 Uncaught TypeError: Cannot set properties of undefined (setting 'selectedColor')
    at Dm.updateSelectedColors (colorgridsfragmentview.js:104:26)
    at Mm.updateSelectedColors (colorselectorview.js:226:37)
    at ss.<anonymous> (colorui.js:121:44)
    at ss.fire (emittermixin.js:146:47)
    at ss.set [as isOpen] (observablemixin.js:72:30)
    at cs.<anonymous> (dropdownview.js:120:18)
    at cs.fire (emittermixin.js:146:47)
    at O (emittermixin.js:425:17)
    at cs.fire (emittermixin.js:165:25)
    at Ui.<anonymous> (buttonview.js:105:30)
```

---

illia-stv:

Fix (font): Font color and font background should work in both the main toolbar and the balloon toolbar. Closes #15580.

MINOR BREAKING CHANGE (font): `colorSelectorView` property will no longer accessible from `ColorUI` plugin in `@ckeditor/ckeditor5-font/src/ui/colorui.ts` .

---

### Additional information


illia-stv:
You could easily reproduce the problem by following next steps:

- in `font-color.js` in manual test in `Font` package import BalloonToolbar from `@ckeditor/ckeditor5-ui`
- add into plugin's array `BalloonToolbar`
- add into config also `balloonToolbar: [ 'bold', 'italic', 'fontColor' ]`
- run manual test and change color by contextual balloon and error will be reproduced

pszczesniak (commented):
Could you create a manual test that covers that case? You can place it in `tests/manual/tickets/ISSUE_ID` - it will be easier to test for everyone and issue will have reference in manual test.

pszczesniak (approved):
:+1:

mmotyczynska (changes requested):
Unfortunately, this is not the correct way to fix this bug.

The `staticColorsGrid` in our case is `undefined` for a reason. We need to fix it so that it's not `undefined`.

What happens (current implementation):
If we add more than one toolbar button with the font color (can be in the same toolbar, no need to create a separate toolbar), then in the `colorui` only the last reference to `colorSelectorView` is stored in this line: https://github.com/ckeditor/ckeditor5/blob/3bfed6dd313d5be18e94f18280a7a71a1d67f834/packages/ckeditor5-font/src/ui/colorui.ts#L126

In other words, if we have two font color buttons in a toolbar, then the first `colorSelectorView` is created and stored in `this.colorSelectorView`, and then the second one is created and overrides the first one.

In this case, we have two views, but only one has a reference in `colorui` (which is wrong). This is an old regression introduced in https://github.com/ckeditor/ckeditor5-font/pull/39. I don't think anyone noticed it until now, because it's a very rare use case to have the same button twice.

The solution would be to change the line above back to `const colorSelectorView` and update the rest of the code accordingly, along with the tests that fail.

arkflpc (approved):

packages/ckeditor5-font/tests/manual/tickets/15580/1.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
```
mmotyczynska, line 2:
```suggestion
 * @license Copyright (c) 2003-2024, CKSource Holding sp. z o.o. All rights reserved.
```

packages/ckeditor5-font/tests/manual/tickets/15580/1.md
```diff
@@ -0,0 +1,5 @@
+## Font color and font background color throw error while having buttons in the main toolbar and the balloon toolbar [#8233](https://github.com/ckeditor/ckeditor5/pull/15658)
+
+1. Select text, should appear balloon with font color button.
+2. Click in the font color button and change color.
+3. In console should no be any errors.
```
mmotyczynska, line 5:
```suggestion
3. There should be no errors in the console.
```

packages/ckeditor5-font/tests/ui/colorui.js
```diff
@@ -535,4 +535,61 @@ describe( 'ColorUI', () => {
...
+			element = document.createElement( 'div' );
+			document.body.appendChild( element );
+
+			return ClassicTestEditor
+				.create( element, {
+					plugins: [ Paragraph, TestColorPlugin, Undo ],
+					testColor: testColorConfig,
+					toolbar: [
+						'testColor',
+						'testColor'
```
mmotyczynska, line 555: This is useless if in the test you create a new dropdown with `dropdown = editor.ui.componentFactory.create( 'testColor' );`. If you want to create a dropdown this way, there is no need of adding these toolbar buttons (and maybe no need of creating a separate `describe` block, but it's up to you). Check the next comment about how to handle a case with 2 dropdowns created manually.

packages/ckeditor5-font/tests/ui/colorui.js
```diff
@@ -535,4 +535,61 @@ describe( 'ColorUI', () => {
...
+			return editor.destroy();
+		} );
+
+		describe( 'testColor Dropdown', () => {
+			let dropdown;
+
+			beforeEach( () => {
+				command = editor.commands.get( 'testColorCommand' );
+				dropdown = editor.ui.componentFactory.create( 'testColor' );
+				dropdown.isOpen = true;
```
mmotyczynska, line 577: Unfortunately if you test it like this, then this test will not actually verify the changes you've made in the `colorui.ts`. You can check this by undoing your changes in `colorui.ts` and running this test - it still passes.

To check that your changes fix the problem, you have to either keep two dropdowns in the toolbar and open the first one (instead of creating here a third dropdown) or create here 2 dropdowns and try to open the first one.

Your test does not verify the fix because in these 2 lines you create a dropdown and open the same, so in `colorui.ts` this last reference would be stored and in the console there would be no error. For the error to be displayed you'd have to try to open a dropdown that was created not as the last one.

mmotyczynska (changes requested):
Please correct the test so that it actually verifies the fix.

mmotyczynska (approved):

pszczesniak (approved):

PR #15676, opened 2024-01-11
Docs: adding new merging lists behavior to update guide.

### LINKED ISSUE #15675: Documenting different behavior of list vs. document lists.

In the old lists, now `LegacyList` it was possible to create the same list type but with different properties, and they didn't merge.

https://github.com/ckeditor/ckeditor5/assets/9881379/37fe1026-bc05-46b5-86b5-3fbbdfff8061

Document list **merges such lists items**. It's possible to load `AdjacentListsSupport` but it will only work on the data load or paste, not from the UI.

We tried to collect feedback #14478, but we haven't decided yet.

The impact of replacing the lists' implementation (#14767) may be visible if users had such lists (data change). Keep in mind we also switched transparently the implementation of predefined builds.

But this could serve us as a valuable moment of feedback gathering, as after numerous discussion we haven't reached final verdict.

Action items:
* inform in the update guide about that the change happened in the predefined builds.
* link in the update guide the issue for adjacent lists to collect feedback.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

godai78:

Docs: adding new merging lists behavior to update guide. Closes #15675

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/docs/features/lists-editing.md
+++ b/packages/ckeditor5-list/docs/features/lists-editing.md
@@ -59,6 +59,8 @@ ClassicEditor
 	.catch( /* ... */ );
 ```
 
+This feature only works for pasted contents or on data load, it {@link updating/update-to-41#changes-to-list-merging does not support entering adjacent lists via the editor UI}.
+
 ## Indenting lists
 
 Besides controlling {@link features/indent text block indentation}, the indent {@icon @ckeditor/ckeditor5-core/theme/icons/indent.svg Indent} and outdent {@icon @ckeditor/ckeditor5-core/theme/icons/outdent.svg Outdent} buttons allow for indenting list items (nesting them).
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-list/docs/features/lists-editing.md
```diff
@@ -59,6 +59,8 @@ ClassicEditor
 	.catch( /* ... */ );
 ```
 
+This feature only works for pasted contents or on data load, it {@link updating/update-to-41#changes-to-list-merging does not support entering adjacent lists via the editor UI}.
+
```
Witoso, line 63: I think this is a good place to link the issue from GH as well. Someone may miss the update guide and navigate from the feature guide.

Witoso (changes requested):

PR #15686, opened 2024-01-12
Update the find and replace demo content

AnnaTomanek:

Docs: Updated the find and replace demo content. Closes cksource/ckeditor5-internal#3654.

---

### Additional information

Also noticed that the article is missing the Configuration section where the new `aiType` config could be explained. We could probably add it in this PR as well.


godai78 (approved):
LGTM

PR #15702, opened 2024-01-16
Created the accessibility help dialog.

### LINKED ISSUE #1014: Re-add the "accessibility instructions" modal

🆕 Feature request

## 💻 Version of CKEditor

CKEditor 5 builds

## 📋 Steps to reproduce

1.  Press alt+0 while in the text area (which used to bring up the accessibility instructions modal in CKE4)

## ✅ Expected result

Accessibility instructions modal pops up

## ❎ Actual result

I get a degrees symbol on my Mac inserted in the text area 😅

---

oleq:

Feature (ui): Implemented the `AccessibilityHelp` plugin that brings a dialog displaying keyboard shortcuts available in the editor. Closes #1014.

Feature (core): Brought the `editor.accessibility` namespace to the base `Editor` class as a container for accessibility-related features and systems. See #1014.

Other (essentials): Enabled the `AccessibilityHelp` plugin by default. See #1014.

Fix (utils): The exported `keyCodes` object should contain correct codes for keys related to punctuation, brackets, braces, etc. See #1014.

---

**ℹ️ Comes with https://github.com/cksource/ckeditor5-commercial/pull/5941**

---

https://github.com/ckeditor/ckeditor5/assets/1099479/d266ffe0-203c-4c01-b983-33b467227195


packages/ckeditor5-essentials/src/essentials.ts
```diff
@@ -36,7 +37,7 @@ export default class Essentials extends Plugin {
 	 * @inheritDoc
 	 */
 	public static get requires() {
-		return [ Clipboard, Enter, SelectAll, ShiftEnter, Typing, Undo ] as const;
+		return [ Clipboard, Enter, SelectAll, ShiftEnter, Typing, Undo, AccessibilityHelp ] as const;
```
oleq, line 40: I don't like the way it gets injected. `Essentials` is just a plugin not every integration will use it. OTOH I couldn't just use the same [approach](https://github.com/ckeditor/ckeditor5/blob/e3b93bfdd65b8756519812576884434362add1d1/packages/ckeditor5-ui/src/editorui/editorui.ts#L138-L139) as with `PoweredBy` or `Tooltip` because these are not plugins (`AccessibilityHelp` is a plugin and it requires `Dialog`).

I'm wondering if it would be OK to force-inject `AccessibilityHelp` into `editor.config.plugins` at some point? [`BalloonEditor` does that](https://github.com/ckeditor/ckeditor5/blob/e3b93bfdd65b8756519812576884434362add1d1/packages/ckeditor5-editor-balloon/src/ballooneditor.ts#L91-L92) (but it's an editor class). Optimally that would take place in `EditorUI` but there are a few gotchas:
* not sure if the timing is right (would need to check when `EditorUI` gets created)
* not sure if all integrators want the `AccessibilityHelp` feature (?), if only because it requires `Dialog`.

packages/ckeditor5-utils/src/keyboard.ts
```diff
@@ -264,9 +282,19 @@ function generateKnownKeyCodes(): { readonly [ keyCode: string ]: number } {
...
+		'-': 109,
+		'.': 110,
+		'/': 111,
+		';': 186,
+		'=': 187,
+		'[': 219,
+		'\\': 220,
+		']': 221,
+		'`': 223
+	} );
```
oleq, line 297: I'm leaving a comment for the record because this is quite interesting. 

The former implementation was wrong and we hadn't noticed it for years ([not only us](https://stackoverflow.com/questions/27861696/javascript-keycode-clash-right-arrow-and-single-quote)). The thing is, a DOM event's `keyCode` not always aligns with the ASCII table. Sometimes it does ("a", "1", etc.) but for characters like "'" `"'".charCodeAt( 0 )` returns 39 while the actual keystroke code would be 222. 

Because of that, our exported [`keyCodes`](https://ckeditor.com/docs/ckeditor5/latest/api/module_utils_keyboard.html#constant-keyCodes) had duplicated key values ("'" was 39, and "arrowright" was 39), which then became a problem for `keyCodeNames` used by [`getEnvKeystrokeText()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_utils_keyboard.html#function-getEnvKeystrokeText) that reverses the mapping. For instance, as a result, 39 in `keyCodeNames` pointed to "'". So `getEnvKeystrokeText( 'arrowright' )` returned "'". That looked funny when I printed some keystrokes in the `AccessibilityHelp` dialog. ↑, ←, and ↓ looked fine but instead of → there was "'" 😅

We failed to notice this for so long because AFAIR we've never used any of these characters for any feature's keystroke. They wouldn't work because `KeystrokeHandler.set()` also uses [`keyCodes`](https://ckeditor.com/docs/ckeditor5/latest/api/module_utils_keyboard.html#constant-keyCodes) and that would be obvious.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
```diff
@@ -0,0 +1,382 @@
...
+
+					writer.setAttribute(
+						'aria-label',
+						`${ currentAriaLabel }. ${ t( 'Press %0 for help.', [ getEnvKeystrokeText( 'Alt+0' ) ] ) }`,
+						viewRoot!
+					);
+				}
+			} );
+		} );
+	}
```
oleq, line 279: I wish there was an API for that. It's not pretty, and besides, if there were more places like this one, maintaining order would become an issue.

packages/ckeditor5-utils/src/keyboard.ts
oleq: A side comment: we should really drop [`keyCode`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). It's officially deprecated and there are better ways of discovering keys in events. I suppose that would be a BC because `KeystrokeHandler.set()` and `ButtonView#keystroke` format would change.

scofalik:
When I press `Alt+0` on Mac, a character "ľ" is inserted into the editor. Please note, that I have remapped keys on my Mac, so I actually press `Cmd+0` but it shouldn't matter. It looks to me that maybe you didn't prevent the event.

Also, most of the time, the dialog content is not focused. Inserting a character may be a reason but IDK. It works randomly, most of the time it doesn't and someties it does.

Also, it looks that because the character is inserted, the Mac VoiceOver gets confused. It usually does not read the content of the dialog. Instead it reads what was just inserted. Then, even if it read the "introduction" part ("Below are...") it didn't read the table and so forth, I am not sure how this is supposed to work.

scofalik:
> Also, most of the time, the dialog content is not focused. Inserting a character may be a reason but IDK. It works randomly, most of the time it doesn't and someties it does.

Sorry for confusion - actually, the dialog is focused but there is no outline. Since I have hard time making VoiceOver read contents, I thought it is not focused at all. But when I use arrow down it scrolls and the outline shows.

packages/ckeditor5-basic-styles/src/bold/boldui.ts
```diff
@@ -55,4 +55,20 @@ export default class BoldUI extends Plugin {
 			return view;
 		} );
 	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public afterInit(): void {
```
DawidKossowski, line 62: I'm not a fan of this approach of registering new keystrokes. I remember suggesting the use of a static field that could be utilized by `AccessibilityDialog` to minimize the 'connections' between plugins. Anyway, we can still proceed with this solution.

@scofalik, do you have any preferences?

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
```diff
@@ -0,0 +1,382 @@
...
+			buttonView.on( 'execute', () => this._showDialog() );
+
+			return buttonView;
+		} );
+
+		editor.keystrokes.set( 'Alt+0', () => this._showDialog() );
+
+		this._setupRootLabels();
+	}
+
```
DawidKossowski, line 123: I think this API is quite messy. On the one hand, we have a method to register categories; on the other hand, we have a method to register groups, keystrokes, or both. Maybe we should provide 3 elementary methods to handle each level separately. So, `registerKeystrokeCategory` should remain as it is. `registerKeystrokeGroup` should be added to handle the options parameter:
```
options: {
    category: string;
    group: string;
    groupLabel: string;
    keystrokes?: Array<AccessibilityHelpKeystrokeDefinition>;
}
```
When keystrokes are provided, the next method to be called will be `registerKeystrokes`. And when it is not necessary to register a new group, we could simply call `registerKeystrokes`.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
```diff
@@ -0,0 +1,382 @@
...
+	 * 		},
+	 * 		{
+	 * 			label: t( 'Move focus to the visible contextual balloon' ),
+	 * 			keystroke: 'Tab'
+	 * 		}
+	 * 	]
+	 * } );
+	 * ```
+	 */
+	public registerKeystrokes( options: {
```
DawidKossowski, line 201: Please move this parameter to a separate type.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
```diff
@@ -0,0 +1,382 @@
...
+		category?: string;
+		group?: string;
+		groupLabel?: string;
+		keystrokes: Array<AccessibilityHelpKeystrokeDefinition>;
+	} | Array<AccessibilityHelpKeystrokeDefinition> | AccessibilityHelpKeystrokeDefinition ): void {
+		if ( options instanceof Array ) {
```
DawidKossowski, line 207: `Array.isArray( options )` is safer.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
```diff
@@ -0,0 +1,382 @@
...
+			options = {
+				keystrokes: options
+			};
+		} else if ( typeof options === 'object' && 'keystroke' in options ) {
```
DawidKossowski, line 211: I think there is no need to check `typeof options === 'object'` since there is no other possibility due to the first `if` statement.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
DawidKossowski, line 1: Should not we move those files (`accessibilityhelp.ts` and `accessibilityhelpcontentview.ts`) to the dedicated folder: `editorui/accesibilityhelp`?

packages/ckeditor5-autoformat/src/autoformat.ts
```diff
@@ -201,6 +202,22 @@ export default class Autoformat extends Plugin {
 			blockAutoformatEditing( this.editor, this, /^---$/, 'horizontalLine' );
 		}
 	}
+
+	/**
+	 * TODO
```
DawidKossowski, line 207: Remember to write API docs.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
```diff
@@ -0,0 +1,382 @@
...
+	private _showDialog() {
+		const editor = this.editor;
+		const dialog = editor.plugins.get( 'Dialog' );
+		const t = editor.locale.t;
+
+		if ( !this.contentView ) {
+			this.contentView = new AccessibilityHelpContentView( editor.locale, this._keystrokes );
+		}
+
+		dialog.show( {
```
DawidKossowski, line 293: I believe dialog is of type `PluginInterface`. To prevent TS errors, we need to import types.

DawidKossowski (changes requested):

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
oleq: That's one thing. Another thing is, shouldn't it land in ckeditor5-core instead?

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
oleq: WFM 
image

oleq:
> When I press Alt+0 on Mac, a character "ľ" is inserted into the editor. Please note, that I have remapped keys on my Mac, so I actually press Cmd+0 but it shouldn't matter. It looks to me that maybe you didn't prevent the event.

> Also, most of the time, the dialog content is not focused. Inserting a character may be a reason but IDK. It works randomly, most of the time it doesn't and someties it does.

> Also, it looks that because the character is inserted, the Mac VoiceOver gets confused. It usually does not read the content of the dialog. Instead it reads what was just inserted. Then, even if it read the "introduction" part ("Below are...") it didn't read the table and so forth, I am not sure how this is supposed to work.

> Sorry for confusion - actually, the dialog is focused but there is no outline. Since I have hard time making VoiceOver read contents, I thought it is not focused at all. But when I use arrow down it scrolls and the outline shows.

> EDIT: Buuut, when I use VoiceOver, arrow down press is usually captured and does not scroll the dialog, instead some random element is selected.

> IDK how it all works with other screen readers.

You're right. I forgot to cancel the event. I addressed that. I also added more meaningful focus styles that correspond to our design system for visual clarity.

packages/ckeditor5-autoformat/src/autoformat.ts
oleq: I left that on purpose. If the API changes, this would be a waste of time.

packages/ckeditor5-basic-styles/src/bold/boldui.ts
oleq: For the record, whatever solution we choose the API of the plugin (methods) must remain in case sb. wants to register some new entries on the fly (e.g. via some external integration).

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
DawidKossowski: Weird...
Zrzut ekranu 2024-01-18 o 16 37 46

packages/ckeditor5-essentials/src/essentials.ts
scofalik: If we split this functionality between `AccessibilityDialog` and some registry for the accessibility help info, maybe it will make more sense to add `AccessibilityDialog` to the `Essentials` plugin. Buuuut, AFAICS, neither of essentials plugin requires our UI (or UI at all). So maybe we shouldn't change this.

I am not happy with force-adding the plugin to the `Editor` class or somewhere through the `EditorUI` class. It doesn't feel good.

Side note: requiring the `Dialog` plugin just in itself is not a big problem but yes, it makes other things into plugins. Maybe it is a mistake to have dialog functionality as a plugin and it should have been available through `editor.ui`?

Back to the question. Do we actually need to force the a11y dialog upon people? Maybe let's just have it as any other plugin?

If we want to automate and we are worried about different kinds of editors, we could add a11y dialog plugin to the specific editor classes that we have. At least we know that all of them have UIs.

packages/ckeditor5-basic-styles/src/bold/boldui.ts
scofalik: Each time I return to this subject, I feel that gathering a11y data should be separated from displaying it. I wish so much it was possible to automate it somehow but I guess this not possible currently, without a broad refactor in many many places.

I propose two solutions:

1.  Something like `editor.accessibilityHelp` where we would add registering methods. Then, `AccessibiltyDialog` will read from it and plugins will add to it. It will be always available and we will decouple plugins from the a11y dialog plugin.
2.  Put this functionality into `editor.keystrokes`. We could have something like `editor.keystroke.declare()` that would not actually create a keystroke but would save the information about it. We could have partial automation. We could read from the keystrokes that already exist, and where needed we would add `declare()`. This solution is better if we want to semi-automate it, but it is far less straightforward than `editor.accessibilityHelp.registerKeystroke()`. Especially if we will go deeper with accessibility help and we will start adding some more data there, and it will be clear that "accessibility help" is its own "data space" not just a list of keystrokes.

packages/ckeditor5-list/src/list/listediting.ts
```diff
@@ -598,6 +599,22 @@ export default class ListEditing extends Plugin {
...
+	 * TODO
+	 */
+	private _setupAccessibilityHelp(): void {
+		const editor = this.editor;
+
+		if ( editor.plugins.has( 'AccessibilityHelp' ) ) {
+			const t = editor.t;
+
+			editor.plugins.get( 'AccessibilityHelp' ).registerKeystrokes( {
+				label: t( 'Nest the current list item (when in a list)' ),
```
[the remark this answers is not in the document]
scofalik: Aaand where's shift+tab? :D

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
scofalik: Or maybe we will switch it to `AccessibilityDialog` and treat it as any other plugin (so it will get a mini package :)).

I agree with @oleq, it will depend on the injection method. I feel we need to discuss it one more time.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
```diff
@@ -0,0 +1,395 @@
...
+					label: t( 'Move focus to the visible contextual balloon' ),
+					keystroke: 'Tab'
+				},
+				{
+					label: t( 'Open the accessibility help dialog' ),
+					keystroke: 'Alt+0'
+				},
+				{
+					label: t( 'Move focus between fields (inputs and buttons) in balloons and dialogs' ),
+					keystroke: 'Tab'
```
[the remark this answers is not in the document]
scofalik: Do we need to mention "in balloons and dialogs" here? Feels like unnecessary noise. Unless you don't want to confuse with things in toolbar?

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
scofalik: Make sure that it is possible to overwrite a keystroke. I think that when adding a keystroke we could use a label as a key/id.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
scofalik: I'll take a look at it but I want to submit the review, so no comment here now.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
scofalik: +1 for making it straightforward.

packages/ckeditor5-essentials/src/essentials.ts
oleq: F2F decision:
* Let's go with `Essentials` because, for instance, unlike the balloon toolbar in `BalloonEditor`, this is not a critical functionality for every editor, and the coupling would be too tight if we force-injected it in editor constructors. Besides, this option is still on the table.
* Let's keep in mind that we need to have `Essentials` or `AccessibilityHelp` in snippets (guides). This will be a follow-up.
* Let's check the compatibility with context plugins (a follow-up).

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
oleq: F2F decision, let's go with 3 methods.

DawidKossowski:
I checked the newest changes in API and LGTM.

packages/ckeditor5-list/src/list/listediting.ts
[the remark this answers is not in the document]
scofalik: What about this comment?

docs/features/keyboard-support.md
```diff
@@ -195,6 +195,15 @@ Use the following keystrokes for more efficient navigation in the CKEditor&nbsp;
 }
 </style>
 
+## Displaying keyboard shortcuts in the editor
```
scofalik, line 198: Did you have a chance to check if it is feasible to display the a11y dialog contents automatically, instead of the table above? I think it was a cool idea.

scofalik (changes requested):
A few comments from my side in the review. I also added some comments to already existing "threads" so I hope you won't miss them.

Other than that -- looks like you skipped comments from my previous review. If you don't agree with them, please reply to them.

When it comes to the API (also relates to `accessibilityHelpMetadata`) -- I thought that you will separately register category, group, and then keystrokes in them. But I can live with what is right now. Might be even better.

Finally a very minor suggestion that you don't have to agree with it, `accessibilityHelpMetadata` -> `accessibilityMetadata`?

BTW. the format of `accessibilityHelpMetadata` will allow adding further types of metadata in future (not related to keystrokes). I like that - was that on purpose, were you thinking about something?

oleq:
@scofalik 

> When it comes to the API (also relates to accessibilityHelpMetadata) -- I thought that you will separately register category, group, and then keystrokes in them. But I can live with what is right now. Might be even better.

👍

> Finally a very minor suggestion that you don't have to agree with it, accessibilityHelpMetadata -> accessibilityMetadata?

Good idea.

> BTW. the format of accessibilityHelpMetadata will allow adding further types of metadata in future (not related to keystrokes). I like that - was that on purpose, were you thinking about something?

Yes, I wanted to keep this open-ended in case properties arrive in the future. This will avoid breaking changes.

docs/features/keyboard-support.md
oleq: Yes, I considered that. This would butcher SEO, though as the editor must run first for the content to generate. So I figured out duplication is safer.

packages/ckeditor5-basic-styles/lang/contexts.json
```diff
@@ -5,5 +5,8 @@
 	"Code": "Toolbar button tooltip for the Code feature.",
 	"Strikethrough": "Toolbar button tooltip for the Strikethrough feature.",
 	"Subscript": "Toolbar button tooltip for the Subscript feature.",
-	"Superscript": "Toolbar button tooltip for the Superscript feature."
+	"Superscript": "Toolbar button tooltip for the Superscript feature.",
+	"Italic text": "Accessibility help dialog keystroke description: keystroke for making text italic.",
+	"Move out of an inline code style": "Accessibility help dialog keystroke description: keystroke for moving selection out of an inline code style.",
+	"Bold text": "Accessibility help dialog keystroke description: keystroke for making text bold."
```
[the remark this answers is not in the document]
oleq: And case change!

docs/_snippets/features/keyboard-support.js
```diff
@@ -0,0 +1,538 @@
+/**
```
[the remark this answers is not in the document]
oleq: This is, I'm afraid, an issue with any sample. 

Anyway, I tried `full-featured-editor.js` but its main purpose is displaying all features in a multi-row toolbar (+word count...). It's an ugly sample with huge (and quite random) content that overwhelms the a11y keyboard guide. 

In Umberto, I can inject it as a whole only and I have no way of customizing the config so I cannot display the a11y help button in the toolbar along with the tour balloon (same with initial data).

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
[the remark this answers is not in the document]
oleq: 785464f

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -72,6 +72,26 @@ export default class Dialog extends Plugin {
...
+	public afterInit(): void {
+		const editor = this.editor;
+
+		if ( editor.plugins.has( 'AccessibilityHelp' ) ) {
+			const t = editor.t;
+
+			editor.plugins.get( 'AccessibilityHelp' ).registerKeystrokes( {
+				category: 'navigation',
+				keystrokes: [ {
+					label: t( 'Move focus in and out of an active dialog window' ),
```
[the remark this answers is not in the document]
oleq: This is technical jargon. The dialog is meant to help all kinds of users.

image

image

packages/ckeditor5-list/src/list/listediting.ts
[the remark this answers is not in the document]
oleq: Addressed in 9ad9c8f.

packages/ckeditor5-ui/src/editorui/accessibilityhelp.ts
oleq: I addressed this in 6c4df0b1352bb02da3a95344020ec7daab5df77e.

It turned out the `add` event on `editor.editing.view.document.roots` fires too early to inject this kind of label (`InlineEditableUIView` hasn't been rendered yet).

For clarity, I don't like this piece of code either. AFAIR I proposed a better solution a couple of years ago but it was rejected. There's a lot of magic going on around this process:

1. first, `InlineEditableUIView` is created along the engine root editable element, and you can write its attributes using downcast writer (if you hook into the process too early),
2. But whatever you do, it will get erased as soon as the related `InlineEditableUIView` renders and [sets the attribute via UI library API](https://github.com/ckeditor/ckeditor5/blob/6c4df0b1352bb02da3a95344020ec7daab5df77e/packages/ckeditor5-ui/src/editableui/inline/inlineeditableuiview.ts#L66-L70)
3. Only after `InlineEditableUIView#render()`, you can extend its attributes for real using downast writer.

We really need to streamline this at some point.

docs/_snippets/features/keyboard-support.js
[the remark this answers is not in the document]
godai78: We probably should add the help button to feature rich, but all being sad is true, actually.

DawidKossowski (commented):
Generally, everything looks fine. I left only several random thoughts.

packages/ckeditor5-ui/src/editorui/accessibilityhelp/accessibilityhelpcontentview.ts
```diff
@@ -0,0 +1,146 @@
...
+			return createElement( document, 'section', {}, elements );
+		} );
+	}
+
+	/**
+	 * Creates `[<h4>Optional label</h4>]<dl>...</dl>` elements for each group of keystrokes in a category.
+	 */
+	private _createGroup( groupDefinition: KeystrokeInfoGroupDefinition ): Array<HTMLElement> {
+		const elements: Array<HTMLElement> = [
+			createElement( document, 'dl', {}, Array.from( groupDefinition.keystrokes )
```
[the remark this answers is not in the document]
oleq: Addressed in https://github.com/ckeditor/ckeditor5/pull/15702/commits/421b18712b0370c5e0a29a8a047494e12ff6b06d.

PR #15784, opened 2024-01-31
Add the mutation observer to `TooltipManager`

Dumluregn:

Feature(ui): Tooltip manager will now detect changes in the tooltip text and position in the element with the currently visible tooltip.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ui/src/tooltipmanager.ts
+++ b/packages/ckeditor5-ui/src/tooltipmanager.ts
@@ -123,6 +123,11 @@ export default class TooltipManager extends DomEmitterMixin() {
 	private _resizeObserver: ResizeObserver | null = null;
 
 	/**
+	 * An instance of the mutation observer that keeps track on target element attributes changes.
+	 */
+	private _mutationObserver: MutationObserverWrapper | null = null;
+
+	/**
 	 * A debounced version of {@link #_pinTooltip}. Tooltips show with a delay to avoid flashing and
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-ui/src/tooltipmanager.ts
```diff
@@ -453,3 +468,29 @@ function getTooltipData( element: HTMLElement ): TooltipData {
 		cssClass: element.dataset.ckeTooltipClass || ''
 	};
 }
+
+// Creates a simple MutationObserver instance wrapper that observes changes in the tooltip-related attributes of the given element.
+// Used instead of the MutationObserver from the engine for simplicity.
```
scofalik, line 473:
```suggestion
// Creates a simple `MutationObserver` instance wrapper that observes changes in the tooltip-related attributes of the given element.
// Used instead of the `MutationObserver` from the engine for simplicity.
```

scofalik (commented):
With this PR, there were two options how we can handle the changes in the data attribute:

*   Hide tooltip if it becomes empty.
*   Dynamically change tooltip text as the attribute changes.

You chose the first one. My question is: was it a conscious decision, i.e., did you guys decided that the second choice is a bad UX?

Dumluregn:
I made this decision consciously to avoid distraction for the user and because it seemed to be simpler, but I'm not sure if we discussed it explicitly :thinking:

PR #15791, opened 2024-02-02
Fix ability to change found text while the dialog is open

### LINKED ISSUE #15680: Ability to change found text while the dialog is open

1.  Open `Find and Replace` feature guide (http://127.0.0.1:8080/ckeditor5/latest/features/find-and-replace.html)
2.  Open `Find and Replace` dialog, type a word that exist in content and click button `Find`.
3.  Type something in `replace with...` input (to better see the final "effect").
4.  Now, for exaple, place selection in the middle of the marked word in content and start typing (or like on the video, select part of the word and move it somewhere else).
5.  Hit `Replace All` button.

### ✔️ Expected result

If searched words !== word in text, should marked text dissapeared and `Replace` should not affect this place.

### ❌ Actual result

https://github.com/ckeditor/ckeditor5/assets/9881379/6731fb8a-2c58-4667-8811-0d11da077c4c

### ❓ Possible solution

🤔 Maybe it should not be possible to change the marked text or it can be changed but searching should be refreshed?

## 📃 Other details

*   Browser: …
*   OS: …

Mati365:

Fix(find-and-replace): Highlighted search result is no longer highlighted if the user has modified part of the found phrase in the text editor. 

---

### Additional information

Closes #15680

### DIFF
--- a/packages/ckeditor5-find-and-replace/src/findandreplace.ts
+++ b/packages/ckeditor5-find-and-replace/src/findandreplace.ts
@@ -58,7 +58,7 @@ export default class FindAndReplace extends Plugin {
 			// Data is contained only for the "find" button.
 			if ( data ) {
 				state.searchText = data.searchText;
-				this.editor.execute( 'find', data.searchText, data );
+				findAndReplaceEditing.find( data.searchText, data );
 			} else {
 				// Find next arrow button press.
 				this.editor.execute( 'findNext' );
@@ -67,7 +67,7 @@ export default class FindAndReplace extends Plugin {
 
 		ui.on<FindPreviousEvent>( 'findPrevious', ( event, data ) => {
 			if ( data && state.searchText !== data.searchText ) {
-				this.editor.execute( 'find', data.searchText );
+				findAndReplaceEditing.find( data.searchText );
 			} else {
 				// Subsequent calls.
 				this.editor.execute( 'findPrevious' );
@@ -76,7 +76,7 @@ export default class FindAndReplace extends Plugin {
 
 		ui.on<ReplaceEvent>( 'replace', ( event, data ) => {
 			if ( state.searchText !== data.searchText ) {
-				this.editor.execute( 'find', data.searchText );
+				findAndReplaceEditing.find( data.searchText );
 			}
 
 			const highlightedResult = state.highlightedResult;
@@ -89,7 +89,7 @@ export default class FindAndReplace extends Plugin {
 		ui.on<ReplaceAllEvent>( 'replaceAll', ( event, data ) => {
 			// The state hadn't been yet built for this search text.
 			if ( state.searchText !== data.searchText ) {
-				this.editor.execute( 'find', data.searchText );
+				findAndReplaceEditing.find( data.searchText );
 			}
 
 			this.editor.execute( 'replaceAll', data.replaceText, state.results );
--- a/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
+++ b/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
@@ -7,8 +7,8 @@
  * @module find-and-replace/findandreplaceediting
  */
 
-import { Plugin, type Editor } from 'ckeditor5/src/core.js';
-import type { DiffItem, DiffItemAttribute, Element, Item, Node } from 'ckeditor5/src/engine.js';
+import { Plugin } from 'ckeditor5/src/core.js';
+import type { DiffItem, DiffItemAttribute, Element, Node } from 'ckeditor5/src/engine.js';
 import {
 	scrollViewportToShowTarget,
 	type Collection,
@@ -16,12 +16,12 @@ import {
 	type ObservableChangeEvent
 } from 'ckeditor5/src/utils.js';
 
-import FindCommand from './findcommand.js';
+import FindCommand, { type FindAttributes } from './findcommand.js';
 import ReplaceCommand from './replacecommand.js';
 import ReplaceAllCommand from './replaceallcommand.js';
 import FindNextCommand from './findnextcommand.js';
 import FindPreviousCommand from './findpreviouscommand.js';
-import FindAndReplaceState from './findandreplacestate.js';
+import FindAndReplaceState, { type FindCallback } from './findandreplacestate.js';
 import FindAndReplaceUtils from './findandreplaceutils.js';
 import type { ResultType } from './findandreplace.js';
 
@@ -30,66 +30,6 @@ import { debounce } from 'lodash-es';
 import '../theme/findandreplace.css';
 
 const HIGHLIGHT_CLASS = 'ck-find-result_selected';
-
-/**
- * Reacts to document changes in order to update search list.
- */
-function onDocumentChange(
-	results: Collection<ResultType>,
-	editor: Editor,
-	searchCallback: ( ( { item, text }: { item: Item; text: string } ) => Array<ResultType> )
-) {
-	const changedNodes = new Set<Node>();
-	const removedMarkers = new Set<string>();
-	const model = editor.model;
-
-	const changes = model.document.differ.getChanges() as Array<Exclude<DiffItem, DiffItemAttribute>>;
-
-	// Get nodes in which changes happened to re-run a search callback on them.
-	changes.forEach( change => {
-		if ( change.name === '$text' || model.schema.isInline( change.position.nodeAfter! ) ) {
-			changedNodes.add( change.position.parent as Element );
-
-			[ ...model.markers.getMarkersAtPosition( change.position ) ].forEach( markerAtChange => {
-				removedMarkers.add( markerAtChange.name );
-			} );
-		} else if ( change.type === 'insert' ) {
-			changedNodes.add( change.position.nodeAfter! );
-		}
-	} );
-
-	// Get markers from removed nodes also.
-	model.document.differ.getChangedMarkers().forEach( ( { name, data: { newRange } } ) => {
-		if ( newRange && newRange.start.root.rootName === '$graveyard' ) {
-			removedMarkers.add( name );
-		}
-	} );
-
-	// Get markers from the updated nodes and remove all (search will be re-run on these nodes).
-	changedNodes.forEach( node => {
-		const markersInNode = [ ...model.markers.getMarkersIntersectingRange( model.createRangeIn( node as Element ) ) ];
-
-		markersInNode.forEach( marker => removedMarkers.add( marker.name ) );
-	} );
-
-	// Remove results & markers from the changed part of content.
-	model.change( writer => {
-		removedMarkers.forEach( markerName => {
-			// Remove the result first - in order to prevent rendering a removed marker.
-			if ( results.has( markerName ) ) {
-				results.remove( markerName );
-			}
-
-			writer.removeMarker( markerName );
-		} );
-	} );
-
-	// Run search callback again on updated nodes.
-	changedNodes.forEach( nodeToCheck => {
-		const findAndReplaceUtils: FindAndReplaceUtils = editor.plugins.get( 'FindAndReplaceUtils' );
-		findAndReplaceUtils.updateFindResultFromRange( model.createRangeOn( nodeToCheck ), model, searchCallback, results );
-	} );
-}
 
 /**
  * Implements the editing part for find and replace plugin. For example conversion, commands etc.
@@ -110,27 +50,25 @@ export default class FindAndReplaceEditing extends Plugin {
 	}
 
 	/**
-	 * The collection of currently highlighted search results.
-	 *
-	 * @private
-	 * @member {module:utils/collection~Collection} #_activeResults
-	 */
-	private _activeResults?: Collection<ResultType> | null;
-
-	/**
 	 * An object storing the find and replace state within a given editor instance.
-	 *
-	 * @member {module:find-and-replace/findandreplacestate~FindAndReplaceState} #state
 	 */
 	public state?: FindAndReplaceState;
+
+	/**
+	 * A flag that indicates that the user has started a search and the editor is listening for changes
+	 * to the text on which it will perform an automatic search. Among other things, the mode is activated
+	 * when the user first clicks 'Find' button and then later deactivated when the modal or search dropdown is closed.
+	 *
+	 * @internal
+	 */
+	public declare _isSearchActive: boolean;
 
 	/**
 	 * @inheritDoc
 	 */
 	public init(): void {
-		this._activeResults = null;
-
 		this.state = new FindAndReplaceState( this.editor.model );
+		this.set( '_isSearchActive', false );
 
 		this._defineConverters();
 		this._defineCommands();
@@ -179,40 +117,32 @@ export default class FindAndReplaceEditing extends Plugin {
 		// It's possible that the editor will get destroyed before debounced call kicks in.
 		// This would result with accessing a view three that is no longer in DOM.
 		this.listenTo( this.editor, 'destroy', debouncedScrollListener.cancel );
+
+		this.on<ObservableChangeEvent<boolean>>( 'change:_isSearchActive', ( evt, name, isSearchActive ) => {
+			if ( isSearchActive ) {
+				this.listenTo( this.editor.model.document, 'change:data', this._onDocumentChange );
+			} else {
+				this.stopListening( this.editor.model.document, 'change:data', this._onDocumentChange );
+			}
+		} );
 	}
 
 	/**
 	 * Initiate a search.
 	 */
-	public find(
-		callbackOrText: string | ( ( { item, text }: { item: Item; text: string } ) => Array<ResultType> )
-	): Collection<ResultType> {
-		const { editor } = this;
-		const { model } = editor;
+	public find( callbackOrText: string | FindCallback, findAttributes?: FindAttributes ): Collection<ResultType> {
+		this._isSearchActive = true;
+		this.editor.execute( 'find', callbackOrText, findAttributes );
 
-		const { findCallback, results } = editor.execute( 'find', callbackOrText );
-
-		this._activeResults = results;
-
-		// @todo: handle this listener, another copy is in findcommand.js file.
-		this.listenTo( model.document, 'change:data', () => onDocumentChange( this._activeResults!, editor, findCallback ) );
-
-		return this._activeResults;
+		return this.state!.results;
 	}
 
 	/**
 	 * Stops active results from updating, and clears out the results.
 	 */
 	public stop(): void {
-		if ( !this._activeResults ) {
-			return;
-		}
-
-		this.stopListening( this.editor.model.document );
-
 		this.state!.clear( this.editor.model );
-
-		this._activeResults = null;
+		this._isSearchActive = false;
 	}
 
 	/**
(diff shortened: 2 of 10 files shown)


packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -192,27 +192,28 @@ export default class FindAndReplaceEditing extends Plugin {
 
 		const { findCallback, results } = editor.execute( 'find', callbackOrText );
 
-		this._activeResults = results;
+		this._inFilteringMode = true;
 
-		// @todo: handle this listener, another copy is in findcommand.js file.
-		this.listenTo( model.document, 'change:data', () => onDocumentChange( this._activeResults!, editor, findCallback ) );
+		this.listenTo( model.document, 'change:data', () => {
+			if ( this.state?.results ) {
```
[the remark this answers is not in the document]
Mati365: Fixed

Dumluregn (commented):
There's a bug in the scenario when we're adding searched phrase to the content:

1.  Clear document content.
2.  Find anything, there should be 0 occurrences.
3.  Add the searched phrase to the document.

The find counter shows `0 of 1`. The "Find next" and "Find previous" buttons are disabled (because there's only one result). The "Replace" and "Replace all" buttons are enabled, but the former doesn't do anything when it's clicked. 

I think we should either select the new result (so apply the "selected" highlight and make the counter display `1 of 1`), or enable "Find next/previous" buttons in such case (but it's less intuitive IMO and requires to disable them again once used). CC @DawidKossowski

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -80,7 +80,11 @@ function onDocumentChange(
 				results.remove( markerName );
 			}
 
-			writer.removeMarker( markerName );
+			// This condition is checked in several tests anyway and seems to be ignored by instabul.
```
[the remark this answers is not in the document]
Mati365: It's not falsy in this scenario. It crashes when `results` object contain marker that no longer exists in document (because it was removed by the user in editor).

packages/ckeditor5-find-and-replace/src/findandreplace.ts
```diff
@@ -76,20 +76,21 @@ export default class FindAndReplace extends Plugin {
...
 			if ( state.searchText !== data.searchText ) {
-				this.editor.execute( 'find', data.searchText );
+				findAndReplaceEditing.find( data.searchText );
 			}
 
 			const highlightedResult = state.highlightedResult;
 
 			if ( highlightedResult ) {
 				this.editor.execute( 'replace', data.replaceText, highlightedResult );
+				findAndReplaceEditing.find( data.searchText );
```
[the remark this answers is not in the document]
Mati365: You are right. After fixes it is no longer needed.

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -110,12 +114,12 @@ export default class FindAndReplaceEditing extends Plugin {
...
 	/**
-	 * The collection of currently highlighted search results.
+	 * Flag that indicates that user started using search
 	 *
 	 * @private
-	 * @member {module:utils/collection~Collection} #_activeResults
+	 * @member {module:utils/collection~Collection} #_inFilteringMode
 	 */
-	private _activeResults?: Collection<ResultType> | null;
+	private _inFilteringMode?: boolean = false;
```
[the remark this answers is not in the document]
Mati365: I removed this boolean completely and adjusted assertions. Works completely fine.

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -110,12 +114,12 @@ export default class FindAndReplaceEditing extends Plugin {
 	}
 
 	/**
-	 * The collection of currently highlighted search results.
+	 * Flag that indicates that user started using search
```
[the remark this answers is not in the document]
Mati365: Flag has been removed, not more relevant.

Mati365:
@Dumluregn Can you take a look again? I fixed mentioned issues.

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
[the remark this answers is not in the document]
DawidKossowski: I agree with @Dumluregn, we should remove `istanbul ignore`.

Currently, the tests do not fall within this `if` statement, and adding the `istanbul ignore` comment should not be the solution. Perhaps you will need to prepare mocks or additional API calls to simulate this scenario in unit tests. If you can replicate this scenario in manual tests, it should also be feasible in unit tests.
Please take some time to determine how you can prepare unit tests to cover this case.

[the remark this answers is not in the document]
Mati365: @DawidKossowski The problem is that tests[1] actually fall into this condition and if we remove it then they will simply fail. I added to the `Writer` class specified method[2] that actually does the same as current line and coverage is back equal 100% in this file. I suspect that instabul incorrectly handles the exception raised by `writer.removeMarker` and does not include this line into coverage at all. 

[1] `hitting "Find" with no result should watch document modifications and update highlighted item if not present`

[2] 
```ts 
public removeMarkerIfPresent( markerOrName: string | Marker ): void {
		this._assertWriterUsedCorrectly();

		const name = typeof markerOrName == 'string' ? markerOrName : markerOrName.name;

		if ( !this.model.markers.has( name ) ) {
			return;
		}

		const marker = this.model.markers.get( name )!;

		if ( !marker.managedUsingOperations ) {
			this.model.markers._remove( name );

			return;
		}

		const oldRange = marker.getRange();

		applyMarkerOperation( this, name, oldRange, null, marker.affectsData );
	}
```

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -80,15 +80,33 @@ function onDocumentChange(
...
-		findAndReplaceUtils.updateFindResultFromRange( model.createRangeOn( nodeToCheck ), model, searchCallback, results );
+		const changedNodeSearchResults = findAndReplaceUtils.updateFindResultFromRange(
+			model.createRangeOn( nodeToCheck ), model, searchCallback, results
+		);
+
+		// Sometimes `changedNodeSearchResults` contains elements that are already present on `changedSearchResults`
+		// list so do not return them duplicated.
+		changedNodeSearchResults.forEach( changedNodeResult => {
+			changedSearchResults.add( changedNodeResult );
+		} );
```
Dumluregn, line 104: Could you give an example?

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -80,15 +80,33 @@ function onDocumentChange(
 				results.remove( markerName );
 			}
 
-			writer.removeMarker( markerName );
+			// This condition is checked in several tests anyway and seems to be ignored by instabul.
+			// istanbul ignore next -- @preserve
+			if ( model.markers.has( markerName ) ) {
+				writer.removeMarker( markerName );
+			}
```
Dumluregn, line 87: After the research I can say this code is unnecessary. I pushed a branch with some added logs to better show what is going on here: `ck/15680-with-logs`.

Just a few lines above https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts#L82-L84 we update the find and replace state by removing the deleted results. This triggers the collection listener in [https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplacestate.ts#L78.](https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplacestate.ts#L78.) There the collection has its own logic removing the markers: [https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplacestate.ts#L90-L93.](https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplacestate.ts#L90-L93.) Remembering that this code is executed before the control is returned to `onDocumentChange()` callback in [https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts#L88-L91,](https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts#L88-L91,) the markers are already gone and there's nothing to remove.

Hence, AFAICS this whole block [https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts#L78-L93](https://github.com/ckeditor/ckeditor5/blob/00fdb2ccfb9912a5d1379650f1c458a25e6048e1/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts#L78-L93) doesn't have to be inside a `model.change()` call - we can only remove the results.

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -190,29 +198,25 @@ export default class FindAndReplaceEditing extends Plugin {
...
 
-		this._activeResults = results;
+		this.listenTo( model.document, 'change:data', () => {
+			const { changedSearchResults } = onDocumentChange( this.state!.results, editor, findCallback );
 
-		// @todo: handle this listener, another copy is in findcommand.js file.
-		this.listenTo( model.document, 'change:data', () => onDocumentChange( this._activeResults!, editor, findCallback ) );
+			if ( changedSearchResults.length && !this.state!.highlightedResult ) {
+				this.state!.highlightedResult = changedSearchResults[ 0 ];
+			}
```
Dumluregn, line 208: I think it would be suitable to move these lines to the `onDocumentChange()` since it's a continuation of that logic.

Dumluregn (changes requested):
Besides the inline comments, there's another bug to handle: the `Match case` and `Whole words only` options are not respected:

Dumluregn (commented):
One more thing I noticed: every time `findAndReplaceEditing.find()` is executed, it applies a new `onDocumentChange()` callback. So if you use find 10 times, you get 10 callbacks. We obviously don't want that. The previous idea with having a flag indicating that we're tracking the document content was good, but the flag was checked in the cleanup instead of when the listener is added. Let's revisit this in the improved form, and with unit test for sure (execute `findAndReplaceEditing.find()` a number of times, then change something in the document and check how many times `onDocumentChange()` was called).

Dumluregn:
Sorry for throwing small bits at you, but I also discussed with @DawidKossowski the architecture a bit and we'd like to get rid of `findAndReplaceEditing.find()` calls and instead be coherent and use editor commands (`editor.execute( ... )`) everywhere in the `findandreplace.ts` file listeners. That means that we'll apply the change listener next to these command execute calls instead of inside the `findAndReplaceEditing.find()` method. The good place for the flag I mentioned in the comment before seems to be a `FindAndReplaceState` class, both semantically and it will be handy to use it here.

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
Mati365: Removed. Thx!

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
Mati365: Moved to that function

Mati365:
@Dumluregn / @DawidKossowski I have refactored the use of `findCallback` and how callbacks listening for change are assembled. Can you take a look at it?

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -269,4 +206,89 @@ export default class FindAndReplaceEditing extends Plugin {
...
+			);
+
+			// Sometimes `changedNodeSearchResults` contains elements that are already present on `changedSearchResults`
+			// list so do not return them duplicated.
+			changedNodeSearchResults.forEach( changedNodeResult => {
+				changedSearchResults.push( changedNodeResult );
+			} );
+		} );
+
+		// If there is no selection on found phrase then select first
```
Dumluregn, line 271:
```suggestion
		// If there are found phrases but none is selected, select the first one.
```

Mati365: changed

Mati365:
@DawidKossowski / @Dumluregn I added some changes that add support for "find" event overloading in some plugins (e.g. track-changes).

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -269,4 +197,106 @@ export default class FindAndReplaceEditing extends Plugin {
...
+				const changedNodeSearchResults = findAndReplaceUtils.updateFindResultFromRange(
+					markerToCheck.data.newRange, model, this.state!.lastSearchCallback!, results
+				);
+
+				console.info( 'new', [ ...markerToCheck.data.newRange ] );
+
+				changedSearchResults.push( ...changedNodeSearchResults );
+
+				console.info( 'new found', changedNodeSearchResults );
+			} else if ( markerToCheck.data.oldRange &&
```
scofalik, line 282: I think this functionality will be difficult to implement and it might not be possible to do it here (in `change` event callback). We do know about `oldRange` but it was the range before any change happened. The model might have changed greatly, and the `oldRange` might be quite outdated and point to incorrect paths.

Which you already observed, but as we've seen, it crashed in another scenario and it still doesn't work other scenarios.

I'd drop this requirement for now. That is, do not react to deletion suggestion marker changes (or do not react to deletion suggestion markers being removed, as it seems it is easy to react to such markers being added). Let's move this to a separate ticket.

Other than that it seems to me that this whole callback would use some refactor but let's close what we have since the QA tests went fine.

scofalik (changes requested):

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
scofalik: We can add followup for refactor as well, although I don't assume that we will find time for it.

packages/ckeditor5-find-and-replace/src/findandreplaceutils.ts
```diff
@@ -52,6 +52,21 @@ export default class FindAndReplaceUtils extends Plugin {
 	): Collection<ResultType> {
 		const results = startResults || new Collection();
 
+		const checkIfResultAlreadyOnList = ( marker: Marker ) => results.find(
```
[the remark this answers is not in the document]
Mati365: > I guess there may be a situation where we removed something that already had a search marker and we re-add both the search marker and the new search result?
 
Yes, there can be such case. It happens quite often in T&C mode when you remove multiple paragraphs at once and then undo it. You are absolutely right about performance, it will affect it, but at this moment it's still better than having the feature broken at all. I suspect that this function and huge part of implementation of `_onDocumentChange` will be changed anyway, and it'll be no longer needed after fixing changes detection.

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
Mati365: I added your comment to newly created issue about this problem.

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -110,27 +50,27 @@ export default class FindAndReplaceEditing extends Plugin {
...
+	 * @member {module:find-and-replace/findandreplacestate~FindAndReplaceState} #state
 	 */
-	private _activeResults?: Collection<ResultType> | null;
+	public state?: FindAndReplaceState;
 
 	/**
-	 * An object storing the find and replace state within a given editor instance.
+	 * A flag that indicates that the user has started a search and the editor is listening for changes
+	 * to the text on which it will perform an automatic search. Among other things, the mode is activated
+	 * when the user first clicks 'search' and then later deactivated when the modal or search dropdown is closed.
```
Dumluregn, line 62:
```suggestion
	 * when the user first clicks 'Find' button and then later deactivated when the modal or search dropdown is closed.
```

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -110,27 +50,27 @@ export default class FindAndReplaceEditing extends Plugin {
 	}
 
 	/**
-	 * The collection of currently highlighted search results.
+	 * An object storing the find and replace state within a given editor instance.
 	 *
-	 * @private
-	 * @member {module:utils/collection~Collection} #_activeResults
+	 * @member {module:find-and-replace/findandreplacestate~FindAndReplaceState} #state
```
Dumluregn, line 55: Let's remove this deprecated \`@member\` tag.

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -269,4 +201,90 @@ export default class FindAndReplaceEditing extends Plugin {
...
+				[ ...model.markers.getMarkersAtPosition( change.position ) ].forEach( markerAtChange => {
+					removedMarkers.add( markerAtChange.name );
+				} );
+			} else if ( change.type === 'insert' && change.position.nodeAfter ) {
+				changedNodes.add( change.position.nodeAfter );
+			}
+		} );
+
+		// Get markers from removed nodes also.
+		model.document.differ.getChangedMarkers().forEach( ( { name, data: { newRange } } ) => {
```
Dumluregn, line 235:
```suggestion
		changedMarkers.forEach( ( { name, data: { newRange } } ) => {
```

packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
```diff
@@ -269,4 +201,90 @@ export default class FindAndReplaceEditing extends Plugin {
...
+
+		changedNodes.forEach( nodeToCheck => {
+			const changedNodeSearchResults = findAndReplaceUtils.updateFindResultFromRange(
+				model.createRangeOn( nodeToCheck ), model, this.state!.lastSearchCallback!, results
+			);
+
+			changedSearchResults.push( ...changedNodeSearchResults );
+		} );
+
+		changedMarkers.forEach( markerToCheck => {
```
Dumluregn, line 273: Looks like this loop could be merged with the previous one (from the line 235).

packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
```diff
@@ -124,3 +134,7 @@ export default class FindAndReplaceState extends ObservableMixin() {
 	}
 }
 
+/**
+ * The callback function used to find matches in the document.
+ */
+export type FindCallback = ( ( { item, text }: { item: Item; text: string } ) => Array<ResultType> );
```
Dumluregn, line 140: I think I'd keep this type in `findcommand.ts`, but I won't persist on it.

Dumluregn (commented):
I noticed one incorrect behavior with replace and undo:

https://github.com/ckeditor/ckeditor5/assets/10883184/d06eaea4-a301-43c8-89d4-66d757d939f8

Please take a look if it's something that can be easily fixed. Besides that, only some small nitpicks here and there.

There were quite a few changes in the code since the last testing, so I'd re-send this PR to the QA team once it's done on our side.

Mati365:
@Dumluregn Can you check again? I pushed fixes and rebased PR into one commit.

Dumluregn (approved):
LGTM, well done :+1: We'll handle the track changes followups separately.

PR #15818, opened 2024-02-08
Docs: Write the Quick start CDN guide.

gorzelinski:

Docs: Write the Quick start CDN guide. Closes https://github.com/cksource/ckeditor5-commercial/issues/5975.

---

### Additional information

### DIFF
--- a/docs/getting-started/quick-start.md
+++ b/docs/getting-started/quick-start.md
@@ -1,7 +1,7 @@
 ---
 category: installation
 order: 10
-menu-title: Quick Start
+menu-title: Quick Start (NPM)
 meta-title: Quick Start | CKEditor 5 documentation
 meta-description: Learn the fastest way to install and use CKEditor 5 - the powerful, rich text WYSIWYG editor in your web application using npm.
 modified_at: 2024-02-05
@@ -72,7 +72,7 @@ import { FormatPainter, SlashCommand } from 'ckeditor5-premium-features';
 
 import 'ckeditor5/dist/styles.css';
 
-ClassicEditor
+await ClassicEditor
 	.create( document.querySelector( '#editor' ), {
 		plugins: [ Essentials, FormatPainter, SlashCommand ],
 		toolbar: [ 'formatPainter' ],
@@ -81,6 +81,63 @@ ClassicEditor
 	.catch( err => {
 		console.error( err );
 	} );
+```
+
+## Using CKEditor&nbsp;5 from CDN
+
+// Will there be 2 CDN links? Which one we want to promote?
+
+Short description about this method.
+
+// If the premium one, what about the license key?
+
+### Obtain a license key
+
+* Trial?
+* Buy?
+
+// Do we want to show steps on how to obtain a license key?
+Steps with screenshots how to obtain a license key.
+
+### CDN setup
+
+1. Add a link for CSS
+
+```html
+<link rel="stylesheet" href="<CDN_LINK>/ckeditor5/dist/styles.css">
+```
+
+2. Add scripts with import maps
+
+```html
+<script type="importmap">
+	{
+	"imports": {
+		"ckeditor5": "<CDN_LINK>/ckeditor5/dist/index.min.js",
+		"ckeditor5/": "<CDN_LINK>/ckeditor5/",
+	}
+	}
+</script>
+```
+
+3. Add a script with editor initialization
+
+```html
+<script type="module">
+	import { ClassicEditor, Essentials, Paragraph } from 'ckeditor5';
+	import translations from 'ckeditor5/dist/translations/pl.js';
+
+	await ClassicEditor.create( document.querySelector( '#editor' ), {
+	plugins: [
+		Essentials,
+		Paragraph,
+	],
+	toolbar: {
+		items: [ 'undo', 'redo' ]
+	},
+	translations
+	} );
+</script>
 ```
 
 ## Next steps
--- a/docs/getting-started/quick-start-cdn.md
+++ b/docs/getting-started/quick-start-cdn.md
@@ -1,0 +1,66 @@
+---
+category: installation
+order: 12
+menu-title: Quick Start (CDN)
+meta-title: Quick Start (CDN) | CKEditor 5 documentation
+meta-description: Learn the fastest way to install and use CKEditor 5 - the powerful, rich text WYSIWYG editor in your web application using npm.
+modified_at: 2024-02-05
+---
+
+# Quick start (CDN)
+
+Short description about this method.
+
+## Obtain a license key
+
+* Trial
+* Buy
+
+Steps with screenshots how to obtain a license key.
+
+## CDN setup
+
+1. Add a link for CSS
+
+```html
+<link rel="stylesheet" href="<CDN_LINK>/ckeditor5/dist/styles.css">
+```
+
+2. Add scripts with import maps
+
+```html
+<script type="importmap">
+	{
+	"imports": {
+		"ckeditor5": "<CDN_LINK>/ckeditor5/dist/index.min.js",
+		"ckeditor5/": "<CDN_LINK>/ckeditor5/",
+	}
+	}
+</script>
+```
+
+3. Add a script with the editor initialization
+
+```html
+<script type="module">
+	import { ClassicEditor, Essentials, Paragraph } from 'ckeditor5';
+	import translations from 'ckeditor5/dist/translations/pl.js';
+
+	await ClassicEditor.create( document.querySelector( '#editor' ), {
+	plugins: [
+		Essentials,
+		Paragraph,
+	],
+	toolbar: {
+		items: [ 'undo', 'redo' ]
+	},
+	translations
+	} );
+</script>
+```
+
+## Next steps
+
+* See how to manipulate the editor's data in the {@link getting-started/getting-and-setting-data Getting and setting data} guide.
+* Refer to the {@link getting-started/setup/configuration setup section} to see how to customize your editor further.
+* Check the {@link features/index features category} to learn more about individual features.


godai78 (approved):

docs/getting-started/quick-start.md
```diff
@@ -72,15 +67,107 @@ import { FormatPainter, SlashCommand } from 'ckeditor5-premium-features';
...
+```
+
+Then, you need to attach the script with the JavaScript code. To simplify imports, you can use the feature available in browsers &ndash; the [import map](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap). It maps the text used as the module specifier and the value that will replace the text when resolving the specifier. This way, import statements do not need to contain the full CDN URL.
+
+```html
+<script type="importmap">
+	{
+		"imports": {
+			"ckeditor5": "<CDN_LINK>/ckeditor5/dist/index.min.js",
+			"ckeditor5/": "<CDN_LINK>/ckeditor5/"
```
Witoso, line 103: @filipsobol is this always needed?

Witoso (commented):
LGTM.

docs/getting-started/quick-start.md
```diff
@@ -25,38 +25,33 @@ npm install ckeditor5
 
 Now you can import all the modules from the `ckeditor5` package. Additionally, you have to import CSS styles separately.
 
-**Importing and registering UI translations is optional for the English language.** For any other language, use imported translations, as in the example below.
+**Importing and registering UI translations is optional for the English language.** For any other language, use imported translations, as in the {@link getting-started/setup/configuration setup section}.
```
filipsobol, line 28:
```suggestion
**Importing and registering UI translations is optional for the American English.** To use the editor in any other language, import translations as shown in the {@link getting-started/setup/configuration setup section}.
```

docs/getting-started/quick-start.md
```diff
@@ -72,15 +67,107 @@ import { FormatPainter, SlashCommand } from 'ckeditor5-premium-features';
...
 		licenseKey: 'your-license-key'
 	} )
-	.catch( err => {
-		console.error( err );
-	} );
+```
+
+## Using CKEditor&nbsp;5 from CDN
+
+CDN is an effortless method of running CKEditor 5. You can start using it in just a few steps and with a few tags.
```
filipsobol, line 80: I'd be cautious about using the word `effortless`, as - depending on the setup - it may require more work than the NPM setup. I'd suggest replacing `effortless` with `alternative`.

filipsobol, line 80: I'd be cautious about using the word `effortless`, as - depending on the setup - it may require more work than the NPM setup. I'd suggest replacing `effortless` with `alternative`.

```suggestion
CDN is an alternative method of running CKEditor 5. You can start using it in just a few steps.
```

docs/getting-started/quick-start.md
```diff
@@ -72,15 +67,107 @@ import { FormatPainter, SlashCommand } from 'ckeditor5-premium-features';
...
+
+### Obtain a license key
+
+To activate CKEditor&nbsp;5 premium features, you will need a commercial license. If you wish to purchase a commercial CKEditor&nbsp;5 license or a license to one of the premium features, [contact us](https://ckeditor.com/contact/?sales=true#contact-form) to receive an offer tailored to your needs. To obtain an activation key, please follow the {@link support/license-key-and-activation License key and activation} giuide.
```
filipsobol, line 84:
```suggestion
To activate CKEditor&nbsp;5 premium features, you will need a commercial license. If you wish to purchase a commercial CKEditor&nbsp;5 license or a license to one of the premium features, [contact us](https://ckeditor.com/contact/?sales=true#contact-form) to receive an offer tailored to your needs. To obtain an activation key, please follow the {@link support/license-key-and-activation License key and activation} guide.
```

docs/getting-started/quick-start.md
```diff
@@ -72,15 +67,107 @@ import { FormatPainter, SlashCommand } from 'ckeditor5-premium-features';
...
+
+<!-- Not sure we should duplicate this in QS: TODO: Describe the steps to obtain a license key, attaching some screenshots. -->
+
+### CDN setup
+
+Start by attaching a link to the style sheet. It contains all styles for the editor's UI and content. You can also include your styles if you like. Refer to the content styles guide for more information.
```
filipsobol, line 90: Add link to the `content styles` guide.

docs/getting-started/quick-start.md
```diff
@@ -72,15 +67,107 @@ import { FormatPainter, SlashCommand } from 'ckeditor5-premium-features';
...
+
+```html
+<link rel="stylesheet" href="<CDN_LINK>/ckeditor5/dist/styles.css" />
+```
+
+Then, you need to attach the script with the JavaScript code. To simplify imports, you can use the feature available in browsers &ndash; the [import map](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap). It maps the text used as the module specifier and the value that will replace the text when resolving the specifier. This way, import statements do not need to contain the full CDN URL.
```
filipsobol, line 96: This feature is not just used to make imports look nicer, so I would rephrase that so as not to tempt anyone to skip using it.

```suggestion
Next, we need to add an [import map](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) that allows us to map an easy-to-remember specifier (like `ckeditor5`) to the full URL of the file from the CDN. We use this browser feature to share an editor engine code between plugins.
```

docs/getting-started/quick-start.md
filipsobol: It's only needed when importings translations, but I'm fine with having this here.

docs/getting-started/quick-start.md
```diff
@@ -72,15 +67,107 @@ import { FormatPainter, SlashCommand } from 'ckeditor5-premium-features';
...
+	{
+		"imports": {
+			"ckeditor5": "<CDN_LINK>/ckeditor5/dist/index.min.js",
+			"ckeditor5/": "<CDN_LINK>/ckeditor5/"
+		}
+	}
+</script>
+```
+
+After you  set up the import map, you can add plugins. In the following script tag, import the desired plugins and add them to the plugins array. Notice that both script tags have the corresponding `type` values.
```
filipsobol, line 109:
```suggestion
Once you've added the import map, you can access editor and its plugins using the `ckeditor5` specifier. In the following script tag, we import the desired plugins and add them to the `plugin' array. Note that this and the previous script tags have the appropriate `type` values.
```

docs/getting-started/quick-start.md
godai78: I'd suggest sticking to "you" as this is the dominant across the whole guide.

docs/getting-started/quick-start.md
godai78: No contractions.

docs/getting-started/quick-start.md
godai78: Unless we have separate translations for American and British (if so, we should reflect it in the docs immediately), this change makes very little sense.

docs/getting-started/quick-start.md
gorzelinski: Yeah, you're right. That was marketing jabber foremost. Will do.

docs/getting-started/quick-start.md
gorzelinski: Sure, but we don't have this guide in the new structure yet.

docs/getting-started/quick-start.md
gorzelinski: Good point. But I would leave "we" in the last sentence because there is a rationale for why we recommend this feature.

godai78: Sure.

gorzelinski:
To consider: if we want to add steps to obtain a license key. On the one hand, the guide will cover all steps. On the other, we duplicate content.

docs/getting-started/quick-start.md
filipsobol: As far as I can see, we have `en` (default), `en-au` and `en-gb`. That's why I suggested explicitly saying that **American English** is the default.

gorzelinski:
To do: add missing links when they are available.

docs/getting-started/quick-start.md
```diff
@@ -25,7 +25,7 @@ npm install ckeditor5
 
 Now you can import all the modules from the `ckeditor5` package. Additionally, you have to import CSS styles separately.
 
-**Importing and registering UI translations is optional for the English language.** For any other language, use imported translations, as in the {@link getting-started/setup/configuration setup section}.
+**Importing and registering UI translations is optional for American English.** To use the editor in any other language, use imported translations, as shown in the {@link getting-started/setup/configuration setup section}.
```
godai78, line 28:
```suggestion
**Importing and registering UI translations is optional for English.** To use the editor in any other language, use imported translations, as shown in the {@link getting-started/setup/configuration setup section}.
```

godai78: I insist. We do not have separate translations, do we?

gorzelinski: It seems that we have:

> As far as I can see, we have en (default), en-au and en-gb. That's why I suggested explicitly saying that American English is the default.

I also asked @pszczesniak, and he confirmed.

godai78: Why don't we have this documented and only list English in out langauge guide? o\_O This is a pretty meaningful rewrite.  
  
OK\< Keep American English, then. I'll split the issue on that.

gorzelinski: I looked through the language guides, and you're right - there is only "English" without dialects.

godai78: We actually do have community translations for these two bad boys: \`au\` and `gb` https://github.com/ckeditor/ckeditor5/tree/master/packages/ckeditor5-core/lang/translations:

PR #15833, opened 2024-02-12
Adjust typings in headings options

### LINKED ISSUE #15790: Types issues in headings

Taken from #15785

### Type definitions to support custom heading models

Currently the typescript definition for heading configs, uses this union type:

https://github.com/ckeditor/ckeditor5/blob/89c183174fe22cf0ec76dda530dc4b344ad36c93/packages/ckeditor5-heading/src/headingconfig.ts#L75

These options explicitly specify heading1-6, and paragraph models. However, according to the [docs](https://ckeditor.com/docs/ckeditor5/latest/features/headings.html#configuring-custom-heading-elements), we're able to specify our own models. (Docs example is `headingFancy`).

https://github.com/ckeditor/ckeditor5/blob/89c183174fe22cf0ec76dda530dc4b344ad36c93/packages/ckeditor5-heading/src/headingconfig.ts#L82

https://github.com/ckeditor/ckeditor5/blob/89c183174fe22cf0ec76dda530dc4b344ad36c93/packages/ckeditor5-heading/src/headingconfig.ts#L110

Could we add some `HeadingCustomOption` type to the union so we don't have to use `@ts-ignore`?

### UpcastAlso in typescript definitions?

I believe the heading options support `upcastAlso` as a property. For example in heading tests:

https://github.com/ckeditor/ckeditor5/blob/89c183174fe22cf0ec76dda530dc4b344ad36c93/packages/ckeditor5-heading/tests/headingediting.js#L93-L106

Can this be added to the typescript definitions for heading options?

Mati365:

Feat (heading): Adjust typings in headings options. Closes #15790.

### DIFF
--- a/packages/ckeditor5-heading/src/headingconfig.ts
+++ b/packages/ckeditor5-heading/src/headingconfig.ts
@@ -75,12 +75,18 @@ export interface HeadingConfig {
  */
 export type HeadingOption = HeadingElementOption | HeadingParagraphOption;
 
+/**
+ * Custom name of the model specified by the user.
+ */
+export type HeadingCustomModel = `heading${ 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H' | 'I' | 'J' | 'K'
+| 'L' | 'M' | 'N' | 'O' | 'P' | 'Q' | 'R' | 'S' | 'T' | 'U' | 'V' | 'W' | 'X' | '-' | '_' | ':' }${ string }`;
+
 export interface HeadingElementOption {
 
 	/**
 	 * Name of the model element to convert.
 	 */
-	model: 'heading1' | 'heading2' | 'heading3' | 'heading4' | 'heading5' | 'heading6' | `heading-${ string }`;
+	model: 'heading1' | 'heading2' | 'heading3' | 'heading4' | 'heading5' | 'heading6' | HeadingCustomModel;
 
 	/**
 	 * Definition of a view element to convert from/to.
--- a/packages/ckeditor5-heading/src/headingconfig.ts
+++ b/packages/ckeditor5-heading/src/headingconfig.ts
@@ -80,7 +80,7 @@ export interface HeadingElementOption {
 	/**
 	 * Name of the model element to convert.
 	 */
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-heading/src/headingconfig.ts
```diff
@@ -74,12 +75,27 @@ export interface HeadingConfig {
...
+ *
+ * @internal
+ */
+type HeadingCustomModelSeparator = 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H' | 'I' | 'J' | 'K'
+	| 'L' | 'M' | 'N' | 'O' | 'P' | 'Q' | 'R' | 'S' | 'T' | 'U' | 'V' | 'W' | 'X' | '-' | '_' | ':';
+
+/**
+ * Custom name of the model specified by the user.
+ *
+ * @internal
```
arkflpc, line 89: Why is it internal if it's used in non-internal type (line 98)?

Mati365: @arkflpc It should be exported and public available but we have issue with documentation generation. Umberto incorrectly serializes union in template string literal and in API reference we have something like it:

obraz

I tried to split it between types or use it inline directly in `HeadingCustomModel` but it always result in `[object Set]` in docs.

DawidKossowski (changes requested):
I think we should change this solution a bit. Adding the new type (`HeadingCustomModelSeparator`) seems redundant, and it makes our API docs look confusing.
Zrzut ekranu 2024-02-19 o 09 18 08

Here's my idea: let's create a separate interface for the custom element. We could call it `HeadingCustomElementOption` or `HeadingCustomOption`, and it would basically be a copy of `HeadingElementOption`, but with the model property changed to `heading${ string }`. You can also try to prepare a basic type and extend it, but I am not sure if it will be correctly compiled by Umberto.

Mati365:
@DawidKossowski Have you tested this solution? I'm quite sure that it'll lead to collapse of heading union.

Mati365:
@DawidKossowski I pushed modifications.

PR #15839, opened 2024-02-12
Source editing should not add white spaces to pre-formatted code lines

### LINKED ISSUE #15084: Source code view causes <pre><code> to gain leading whitespace if inside <li>

1. Create `<li><pre><code>` markup, for example:
```
<h2>
    Good:
</h2>
<pre><code class="language-plaintext"># From any path:
something --flag</code></pre>
<h2>
    <strong>Bad:</strong>
</h2>
<ol>
    <li>
        <p>
            preformatted code block in list gains leading whitespace 😬
        </p>
        <pre><code class="language-plaintext"># From any path:
something --flag</code></pre>
    </li>
</ol>
```
2. Enable source view
3. Change *anything at all*, e.g. change "Good:" to "Good!"
4. Disable source view

### ✔️ Expected result

The only change is the one you made.

### ❌ Actual result

The `something --flag` has been indented with additional spaces:
li pre code

## 📃 Other details

* Browser: Chrome
* OS: macOS
* First affected CKEditor version: Unknown.
* Installed CKEditor plugins: Reproduced at https://ckeditor.com/ckeditor-5/demo/html-support/

---

### LINKED ISSUE #15810: Source editing mode adds leading whitespaces to code block

1. Open https://ckeditor.com/docs/ckeditor5/41.1.0/features/source-editing.html
2. In source editing mode paste
```
<h2>
    Good:
</h2>
<pre><code class="language-plaintext"># From any path:
something --wflag</code></pre>
<h2>
    <strong>Bada:</strong>
</h2>
<ol>
    <li>
        <p>
            preformatted code block in list gains leading whitespace
        </p>
        <pre><code class="language-plaintext"># From any path:
something --flag</code></pre>
    </li>
</ol>
```
3. Close source editing mode and open it again
4. Change something outside the code blocks and close source editing mode

### ✔️ Expected result

Code block shouldn't change

### ❌ Actual result

Code in list got some additional whitespace beside the first line

### ❓ Possible solution

Ignore extra whitespaces on code block upcast or do not add them in code block downcast.

## 📃 Other details

* Browser: …
* OS: …

---

oleq:

Fix (source-editing): Source editing should not add white spaces to pre-formatted code lines. Closes #15084.

---

### Additional information

I don't like our approach in this helper. Using RegExp to manipulate HTML was not the best idea in the first place and there's always a risk of the whole thing breaking up. A parser would be a better choice but that would mean a re-write of the helper. I briefly researched how `DOMParser` could help us and this only confirmed my suspicion that a lot of work is necessary. We could also use an external library to prettify the source output.

### DIFF
--- a/packages/ckeditor5-source-editing/src/utils/formathtml.ts
+++ b/packages/ckeditor5-source-editing/src/utils/formathtml.ts
@@ -79,7 +79,7 @@ export function formatHtml( input: string ): string {
 		.split( '\n' );
 
 	let indentCount = 0;
-	let isPreformattedLine: 'first' | 'last' | 'middle' | false = false;
+	let isPreformattedLine: ReturnType<typeof isPreformattedBlockLine> = false;
 
 	return lines
 		.filter( line => line.length )
--- a/packages/ckeditor5-source-editing/src/utils/formathtml.ts
+++ b/packages/ckeditor5-source-editing/src/utils/formathtml.ts
@@ -79,16 +79,23 @@ export function formatHtml( input: string ): string {
 		.split( '\n' );
 
 	let indentCount = 0;
+	let isPreformattedLine: 'first' | 'last' | 'middle' | false = false;
 
 	return lines
 		.filter( line => line.length )
 		.map( line => {
+			isPreformattedLine = isPreformattedBlockLine( line, isPreformattedLine );
+
 			if ( isNonVoidOpeningTag( line, elementsToFormat ) ) {
 				return indentLine( line, indentCount++ );
 			}
 
 			if ( isClosingTag( line, elementsToFormat ) ) {
 				return indentLine( line, --indentCount );
+			}
+
+			if ( isPreformattedLine === 'middle' || isPreformattedLine === 'last' ) {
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-source-editing/src/utils/formathtml.ts
```diff
@@ -79,10 +79,13 @@ export function formatHtml( input: string ): string {
 		.split( '\n' );
 
 	let indentCount = 0;
+	let isPreformattedLine: ReturnType<typeof isPreformattedBlockLine> = false;
```
Dumluregn, line 82: Nice typing :+1:

Dumluregn (approved):
Agree that regexp is not great but for now it does the work. I think we could switch to the external lib (like [https://www.npmjs.com/package/js-beautify](https://www.npmjs.com/package/js-beautify) ) if at some point we decide to have syntax highlight (as that would probably also use some existing tooling). Or maybe we'll just use some more complex thing like Monako :thinking: anyway, for now LGTM and it also resolves #15084.

PR #15843, opened 2024-02-13
Add check if differ change node is not null.

mremiszewski:

Fix (table): Prevent errors when multiple changes occur while resizing a table. 

---

### Additional information


oleq:
Fixed in https://github.com/ckeditor/ckeditor5/pull/15854.

PR #15847, opened 2024-02-13
Not able to paste image from Word/Excel when there is a chart or image at the end of the document

### LINKED ISSUE #15758: Not able to paste image from Word/Excel when there is a chart or image at the end of the document

[Sample Word document.](https://github.com/ckeditor/ckeditor5/files/14050410/Sample_image_issue.docx)

1. Open sample Word document
2. Select all content
3. Copy
4. [Paste in the demo](https://ckeditor.com/docs/ckeditor5/latest/features/pasting/paste-from-office-enhanced.html#demo)

### ✔️ Expected result

Image is pasted properly.

### ❌ Actual result

Pasted image is broken. Error in the console:
```js
Not allowed to load local resource
```

When you remove the chart from the file and repeat the steps, image pastes ok.

### Similar issues:
- https://github.com/ckeditor/ckeditor5/issues/9077
- https://github.com/ckeditor/ckeditor5/issues/13998
- https://github.com/ckeditor/ckeditor5/issues/13786

## 📃 Other details

* Browser: …
* OS: …

---

illia-stv:

Fix (paste-from-office): Should be able to paste charts from Microsoft Word. Closes #15758.

---

### Additional information


illia-stv:
### ⛔️ Problem explanation:

First of all, we need to remember that we can not paste Microsoft Word shapes into our editor, as proof Tiny MCE, CKE4, Google docs can not as well (btw in safari is possible). It's impossible because we can not extract shapes from [RTF](https://en.wikipedia.org/wiki/Rich_Text_Format).

Here is where shapes come from.

image

Shapes usually have `v:shapetype` attribute in MC Word markup.

Here is a function which purpose is finding shapes in document:

https://github.com/ckeditor/ckeditor5/blob/869b87fa9163654e0bf27bdc24ecaffd526d3dad/packages/ckeditor5-paste-from-office/src/filters/image.ts#L66-L88

Here is a place where shapes are removed:

https://github.com/ckeditor/ckeditor5/blob/869b87fa9163654e0bf27bdc24ecaffd526d3dad/packages/ckeditor5-paste-from-office/src/filters/image.ts#L99-L129

However charts also have that `v:shapetype` attribute but we can extract it from RTF file, so that's where problem is occur.

Look at this line:

https://github.com/ckeditor/ckeditor5/blob/869b87fa9163654e0bf27bdc24ecaffd526d3dad/packages/ckeditor5-paste-from-office/src/filters/image.ts#L302

We comparing images from document:

https://github.com/ckeditor/ckeditor5/blob/869b87fa9163654e0bf27bdc24ecaffd526d3dad/packages/ckeditor5-paste-from-office/src/filters/image.ts#L39

and also images extracted from RTF

https://github.com/ckeditor/ckeditor5/blob/869b87fa9163654e0bf27bdc24ecaffd526d3dad/packages/ckeditor5-paste-from-office/src/filters/image.ts#L42

We removed chart which is also shape, but we could extract it from RTF, so number of images which is extracted from RTF and found in document is not equal, so thats why images are not rendered at all.

### ✅ Solution explanation:

So if charts could be extracted from RTF we could rendered it, everything we need it's only distinguish regular shapes and chart. I found out that charts have `Chart` in there ids, than we could use that.

In this PR you could also notice huge amount of changes and also raise a question where are the tests. Well in `paste-from-office`package things looks a little bit different. We have files, with MC Word markup as clipboard input and then we have files which represent our expectations in our model, with different browsers (especially difference between safari and other browsers). 

In this file we have our core logic due iterating throw files and verifying input with expected result:

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-paste-from-office/tests/_utils/utils.js

So in context of this PR we have changed a little bit MC Word file here:

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-paste-from-office/tests/_data/image/shapes-online-offline/shapes-online-offline.docx

by adding chart image, and updating another files with updated expectations (chart image should be in model).

packages/ckeditor5-paste-from-office/tests/_data/image/shapes-online-offline/~$apes-online-offline.docx
mmotyczynska, line 1: I guess this file should not be added?

mmotyczynska (changes requested):
Please check comments (only 2 😉 ).
Additionally please add somewhere information why [the second reported file (Excel)](https://github.com/ckeditor/ckeditor5/issues/15758#issuecomment-1910251884) will not work with this fix.
Please also change the PR commit message to be more precise: "It should be possible to paste charts from Microsoft Word." because it still does not work in Excel.

packages/ckeditor5-paste-from-office/src/filters/image.ts
```diff
@@ -80,9 +80,18 @@ function findAllShapesIds( documentFragment: ViewDocumentFragment, writer: Upcas
...
+		const exceptionIds = [ 'Chart' ];
 
 		// If shape element have 'o:gfxdata' attribute and is not directly before `<v:shapetype>` element it means it represent Word shape.
-		if ( shapeElementsMatcher.match( el ) && el.getAttribute( 'o:gfxdata' ) && prevSiblingName !== 'v:shapetype' ) {
+		if (
+			shapeElementsMatcher.match( el ) &&
+			el.getAttribute( 'o:gfxdata' ) &&
+			prevSiblingName !== 'v:shapetype' &&
+			!exceptionIds.some( item => el.getAttribute( 'id' )!.includes( item ) )
+		) {
```
pszczesniak, line 94: The `if` statement becomes too big and it's almost unreadable without analyzing each condition. Lets make it better to read :book: 

I suggest to extract every condition to a `const` and name it this way so the comment above won't be needed.

Suggestion code:

```
const isElementAShape = shapeElementsMatcher.match( el );
const hasElementGfxdataAttribute = el.getAttribute( 'o:gfxdata' );
const isPreviousSiblingAShapeType = prevSiblingName === 'v:shapetype';
const isElementIdInExcpetionsArray = exceptionIds.includes( el.getAttribute('id')! );

if (
	isElementAShape &&
	hasElementGfxdataAttribute &&
	!isPreviousSiblingAShapeType &&
	!isElementIdInExcpetionsArray
)
```

I've change also `some` to `includes` array method - also should increase the readability.

packages/ckeditor5-paste-from-office/src/filters/image.ts
```diff
@@ -80,9 +80,18 @@ function findAllShapesIds( documentFragment: ViewDocumentFragment, writer: Upcas
 		const el = value.item as ViewElement;
 		const previousSibling = el.previousSibling;
 		const prevSiblingName = previousSibling && previousSibling.is( 'element' ) ? previousSibling.name : null;
+		// If shape element has 'o:gfxdata' attribute and is not directly before
```
pszczesniak, line 83: Is this comment needed (line 83 and 84)? Same is in 88th line.

pszczesniak (changes requested):
Two things from me. Other than that it works as expected.

packages/ckeditor5-paste-from-office/src/filters/image.ts
illia-stv: Good point, I duplicated it accidentally😅

packages/ckeditor5-paste-from-office/src/filters/image.ts
```diff
@@ -84,12 +84,19 @@ function findAllShapesIds( documentFragment: ViewDocumentFragment, writer: Upcas
 		// https://github.com/ckeditor/ckeditor5/pull/15847#issuecomment-1941543983
 		const exceptionIds = [ 'Chart' ];
 
-		// If shape element have 'o:gfxdata' attribute and is not directly before `<v:shapetype>` element it means it represent Word shape.
+		const isElementAShape = shapeElementsMatcher.match( el );
+		const hasElementGfxdataAttribute = el.getAttribute( 'o:gfxdata' );
+		const isPreviousSiblingAShapeType = prevSiblingName === 'v:shapetype';
+		const isElementIdInExcpetionsArray = hasElementGfxdataAttribute &&
```
mmotyczynska, line 90:
```suggestion
		const isElementIdInExceptionsArray = hasElementGfxdataAttribute &&
```

mmotyczynska (commented):
1 typo, other than that it works as expected.

pszczesniak (approved):
Please apply @mmotyczynska [suggestion](https://github.com/ckeditor/ckeditor5/pull/15847/files#r1492400155), besides this :+1:

mmotyczynska (approved):

Issue #15850, opened 2024-02-13
Improve support for keys in KeystrokesHandler


map-r:
## 📝 Provide a description of the improvement

Currently, the `KeystrokesHandler` detects modifier keys, alphanumeric keys a-z A-Z 0-9, and is even said to support some special characters. But there are certain problems to it:

- problem 1: the special characters are not actually detected properly, because they are detected via comparing `keyCode`, but the list of supported special characters has been generated using `charCodeAt` which gives completely different values which are not being detected (`=` charCode is 61 while its keyCode is 187),
- problem 2: checking only for keyCodes isn't optimal as well, if we planned to check of uppercase letters, or even better, the uppercase special characters (like #$%^& or _+:? in qwerty keyboards) - the keystrokes are returning the very same keyCode for, example given, `+` and `=`.
- problem 3: the **numpad** keyCodes for the same characters like all digits, but also `-`, `+` etc. are also different, while the keystrokes for digits only detects the non-numpad digit keys.

## 📃 Other details

I understand the keystrokes handler is used mostly for key shortcuts like Ctrl+B, Shift+C etc., but this will probably become a limitation, while this is a fairly easy fix. If not, then there' still a minimum set of problems to be addressed, like problem 1 of bad detection.

If you feel this is more of a bugfix issue, please let me know.

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

Issue #15849, opened 2024-02-13
List properties type is not selected when first opening after the editor document loads


map-r:
## 📝 Provide detailed reproduction steps (if any)

1. Open a new instance of the ckeditor5 with list properties plugin active, preferably in collaboration mode (e.g. try this https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/collaborative-document-editor.html?docId=ee3f09f428a93f93baf4ba8ae1c74719b or create a new one)
2. Make a numbered list, write an entry or two in it,
3. Change the list type to other one, A.B.C. for instance,
4. Reload the page/editor
5. When the editor content reloads, select the list you previously created,
6. Open list properties dropdown

### ✔️ Expected result

- the correct list type in the dropdown is selected from the start

### ❌ Actual result

- no list type is selected
- only after clicking on anything then back to the list type we had it is correct and highlighted in the dropdown 

## 📃 Other details

* Browser: Chrome, Firefox, Safari, probably else

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #15851, opened 2024-02-13
Fixed list properties input behaviour when invalid values typed

### LINKED ISSUE #14939: Typing "e" in the numeric field causes overlap input content with its label

1. Go to numeric field in List Properties (Numbered List)
2. Type "e"
3. Leave the field
4. Go back
5. The entered content and label will overlap in the input

https://github.com/ckeditor/ckeditor5/assets/135701960/fe94d08d-5bf6-4c19-aa55-658f1ebaf957

### ✔️ Expected result

The content of a numeric field should not overlap with its label

### ❌ Actual result

The entered content and label overlap in the input

### ❓ Possible solution

The user probably shouldn't be able to write "e" in numeric fields because it's not needed for numbered lists

## 📃 Other details

* Browser: Chrome
* OS: Ventura 13.4

---

### LINKED ISSUE #15848: Updating isEmpty state on InputBase is getting called multiple times

1. Have a log or debugger to check the `InputBase._updateIsEmpty` (from ui package) (fn that updates isEmpty state of the input view)
2. Make a numbered list
3. open the List properties plugin dropdown while on that list
4. Click on the start index input and type in any character

### ✔️ Expected result

The mentioned function should only get called once per valid input value change

### ❌ Actual result

Unnecessary calls of the function, `_updateIsEmpty` is called two or three times, which adds only a very tiny load to the  editor's performance, but a big disaster is a sum of many small ones isn't it.

### ❓ Possible solution

Seen that the double calls emerge from two-way kind of binding between `value` in the view and the DOM value of input, and the other way around happens probably after the DOM value propagates its event up.
Adding same value checks in one of the "secondary" events that trigger updateIsEmpty might be an option.

---

map-r:

Fix (list): Added new error message in case of an invalid value format in the input. Closes #14939 .
Other (theme-lark): Disabled scaling up of the label onto an empty input if there is an error (`LabeledFieldView`).

---

### Findings
During working on a fix, some side potential issues have been discovered, some of them are already documented under issue tickets:
- https://github.com/ckeditor/ckeditor5/issues/15849
- https://github.com/ckeditor/ckeditor5/issues/15850

### The error text:
The error was still required to both inform about a pasted value being a NaN (like `e`).
This error text can be merged with the original error message instead of having two messages, but in this situation all translations also have to adjust to the change, so either way we require some translation updates.

### DIFF
--- a/packages/ckeditor5-list/lang/contexts.json
+++ b/packages/ckeditor5-list/lang/contexts.json
@@ -24,6 +24,7 @@
 	"Upper-latin": "The tooltip text of the button that toggles the \"upper–latin\" list style.",
 	"List properties": "The label of the button that toggles the visibility of additional numbered list property UI fields.",
 	"Start at": "The label of the input allowing to change the start index of a numbered list.",
+	"Invalid start index value.": "The error message displayed when the numbered list is not a valid number",
 	"Start index must be greater than 0.": "The error message displayed when the numbered list start index input value is invalid.",
 	"Reversed order": "The label of the switch button that reverses the order of the numbered list."
 }
--- a/packages/ckeditor5-list/lang/contexts.json
+++ b/packages/ckeditor5-list/lang/contexts.json
@@ -24,7 +24,7 @@
 	"Upper-latin": "The tooltip text of the button that toggles the \"upper–latin\" list style.",
 	"List properties": "The label of the button that toggles the visibility of additional numbered list property UI fields.",
 	"Start at": "The label of the input allowing to change the start index of a numbered list.",
-	"Invalid start index value.": "The error message displayed when the numbered list is not a valid number.",
+	"Invalid start index value.": "The error message displayed when the numbered list start index input value is not a valid number.",
 	"Start index must be greater than 0.": "The error message displayed when the numbered list start index input value is invalid.",
 	"Reversed order": "The label of the switch button that reverses the order of the numbered list."
 }
--- a/packages/ckeditor5-list/lang/contexts.json
+++ b/packages/ckeditor5-list/lang/contexts.json
@@ -24,7 +24,7 @@
 	"Upper-latin": "The tooltip text of the button that toggles the \"upper–latin\" list style.",
 	"List properties": "The label of the button that toggles the visibility of additional numbered list property UI fields.",
 	"Start at": "The label of the input allowing to change the start index of a numbered list.",
-	"Invalid start index value.": "The error message displayed when the numbered list is not a valid number",
+	"Invalid start index value.": "The error message displayed when the numbered list is not a valid number.",
 	"Start index must be greater than 0.": "The error message displayed when the numbered list start index input value is invalid.",
 	"Reversed order": "The label of the switch button that reverses the order of the numbered list."
 }
--- a/packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
+++ b/packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
@@ -221,9 +221,6 @@ export default class ListPropertiesView extends View {
 			this.keystrokes.set( 'arrowleft', stopPropagation );
 			this.keystrokes.set( 'arrowup', stopPropagation );
 			this.keystrokes.set( 'arrowdown', stopPropagation );
-
-			this.listenTo( this.startIndexFieldView.fieldView.element!, 'keydown',
-				( evInfo, event ) => this._preventSpecialNumericValues( event ) );
 		}
 
 		if ( this.reversedSwitchButtonView ) {
@@ -400,16 +397,6 @@ export default class ListPropertiesView extends View {
 
 		return reversedButtonView;
 	}
-
-	/**
-	 * Cuts out the `e`, `E`, `+` and `-` characters from the numeric input.
-	 * @internal
-	 */
-	private _preventSpecialNumericValues = ( event: KeyboardEvent ) => {
-		if ( /^[e+-]$/i.test( event.key ) ) {
-			event.preventDefault();
-		}
-	};
 }
 
 export type StylesView = View & {
--- a/packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
+++ b/packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
@@ -367,7 +367,7 @@ export default class ListPropertiesView extends View {
 
 			if ( Number.isNaN( startIndex ) ) {
 				// Number inputs allow for the entry of characters that may result in NaN,
-				// such as 'e', '123e', '2-' (ending with a special character).
+				// such as 'e', '+', '123e', '2-'.
 				startIndexFieldView.errorText = t( 'Invalid start index value.' );
 
 				return;
--- a/packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
+++ b/packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
@@ -366,6 +366,7 @@ export default class ListPropertiesView extends View {
 			const startIndex = inputElement.valueAsNumber;
 
 			if ( Number.isNaN( startIndex ) ) {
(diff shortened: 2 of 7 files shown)


packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
```diff
@@ -366,6 +369,7 @@ export default class ListPropertiesView extends View {
 			const startIndex = inputElement.valueAsNumber;
 
 			if ( Number.isNaN( startIndex ) ) {
+				startIndexFieldView.errorText = t( 'Invalid start index value.' );
```
map-r, line 372: second-thought question: do we have any core translation for just 'invalid format' or something like this? we could avoid the new key.

packages/ckeditor5-list/lang/contexts.json
```diff
@@ -24,6 +24,7 @@
 	"Upper-latin": "The tooltip text of the button that toggles the \"upper–latin\" list style.",
 	"List properties": "The label of the button that toggles the visibility of additional numbered list property UI fields.",
 	"Start at": "The label of the input allowing to change the start index of a numbered list.",
+	"Invalid start index value.": "The error message displayed when the numbered list is not a valid number",
```
DawidKossowski, line 27: There is no `.` at the end of the message.

packages/ckeditor5-list/tests/listproperties/ui/listpropertiesview.js
```diff
@@ -567,6 +567,76 @@ describe( 'ListPropertiesView', () => {
...
+				it( 'should not prevent from typing digits in the start index field', () => {
+					const keyEvent = {
+						key: '0',
+						keyCode: 48,
+						shiftCode: false
+					};
+					view.startIndexFieldView.fieldView.element.dispatchEvent( new KeyboardEvent( 'keydown', keyEvent ) );
+					const args = preventSpecialNumericValuesSpy.args[ 0 ];
+
+					expect( args[ 0 ].isTrusted ).to.equal( false );
```
DawidKossowski, line 587: This check is useless in our case.

packages/ckeditor5-ui/src/input/inputbase.ts
```diff
@@ -149,8 +149,10 @@ export default abstract class InputBase<TElement extends HTMLInputElement | HTML
 		// Bind `this.value` to the DOM element's value.
 		// We cannot use `value` DOM attribute because removing it on Edge does not clear the DOM element's value property.
 		this.on<ObservableChangeEvent>( 'change:value', ( evt, name, value ) => {
-			this._setDomElementValue( value );
-			this._updateIsEmpty();
+			if ( value?.toString() !== this.element!.value ) {
```
DawidKossowski, line 152: We do not use `?.` operator since it could generate additional branches in unit tests.
I will send you ADR later.

packages/ckeditor5-list/tests/listproperties/ui/listpropertiesview.js
```diff
@@ -567,6 +567,76 @@ describe( 'ListPropertiesView', () => {
 				view.keystrokes.press( keyEvtData );
 				sinon.assert.callCount( keyEvtData.stopPropagation, 4 );
 			} );
+
+			describe( 'special keys keydown event', () => {
```
DawidKossowski, line 571: The tests do not accurately assess the true impact of  the `_preventSpecialNumericValues` function. Even if I remove the `event.preventDefault();` logic from this function, all tests still pass.

DawidKossowski (changes requested):
After reviewing the entire code, the issue, and how this solution works, I think we need to adjust our approach slightly. Currently, there's still a possibility of pasting incorrect content into the number input, resulting in an error display. We would need to add further checks for such cases (on the `input` event) and handle them accordingly, which would again complicate our code.

Therefore, I believe the best approach is to withdraw the code responsible for blocking not allowed keys (`keydown` event) and leave only the error messages you've added. I've discussed this approach with Tomek, and it aligns with WCAG (error notification when user input contains incorrect value). Remember to write several unit tests.

I understand that we're changing concepts and approaches now, which is why it's crucial to consult on issues and work effects continuously.

packages/ckeditor5-ui/src/input/inputbase.ts
DawidKossowski: ADR: https://github.com/cksource/ckeditor5-commercial/blob/master/adr/0018-avoid-using-certain-operators.md

packages/ckeditor5-list/tests/listproperties/ui/listpropertiesview.js
map-r: yes, it was intentionally left for reviewer with a comment of mine (above) to explain context.

packages/ckeditor5-ui/src/input/inputbase.ts
map-r: thanks.

packages/ckeditor5-list/tests/listproperties/ui/listpropertiesview.js
map-r: Did you get to my comment about tests? If so, feel free to comment on that as at this point I'm not sure which approach would be better in terms of code quality.
https://github.com/ckeditor/ckeditor5/pull/15851#discussion_r1487968936

packages/ckeditor5-list/lang/contexts.json
map-r: I don't get it, I did the same as the other existing example below:
image

map-r: Please explain what do you expect here

packages/ckeditor5-list/tests/listproperties/ui/listpropertiesview.js
map-r: actually I see now another approach to this, simply extracting only the regexp check, thus making this way simpler in tests. But , given general comment to shift approach, I should rather revert all tests and the logic addition.

map-r:
> Therefore, I believe the best approach is to withdraw the code responsible for blocking not allowed keys (keydown event) and leave only the error messages you've added.
Just FYI, going with this change.

packages/ckeditor5-list/lang/contexts.json
map-r: okay, not the id but the message, now I get you I think:)

packages/ckeditor5-list/tests/listproperties/ui/listpropertiesview.js
map-r: closing as outdated (removed)

packages/ckeditor5-list/tests/listproperties/ui/listpropertiesview.js
```diff
@@ -708,6 +708,17 @@ describe( 'ListPropertiesView', () => {
 				sinon.assert.notCalled( spy );
 				expect( view.startIndexFieldView.errorText ).to.equal( 'Start index must be greater than 0.' );
 			} );
+
+			it( 'should not fire #listStart upon #input but display an error if the numeric value is NaN', () => {
```
DawidKossowski, line 712: There are still no tests or assertions to verify the disappearance of error messages after typing the correct value.

packages/ckeditor5-list/src/listproperties/ui/listpropertiesview.ts
```diff
@@ -366,6 +366,7 @@ export default class ListPropertiesView extends View {
...
```
DawidKossowski, line 369: I would add a comment explaining the cases/values that could result in this error.

DawidKossowski (changes requested):

packages/ckeditor5-list/tests/listproperties/ui/listpropertiesview.js
map-r: okay, but I don't see them in all other tests, anyway good idea

PR #15863, opened 2024-02-14
Fixes #15862: Handle newline-suppressed scenarios (e.g. limit elements) in AutoLink plugin

### LINKED ISSUE #15862: AutoLink plugin throws errors and doesn't create links after pressing enter in an inline limit element

https://codepen.io/jonscheiding/pen/NWJejvY

1. Place cursor after the url in the black block
2. Press Enter

### ✔️ Expected result

Either of the following would be acceptable:

* The url is converted to a link
* Nothing happens

### ❌ Actual result

Nothing happens in the editor, but there is an error on the console.

### ❓ Possible solution

The AutoLink plugin [calls `model.createRangeIn`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-link/src/autolink.ts#L247) on a `Node` without first checking if it's a valid node type for that operation. If you're inside a limit element, the newline on enter is suppressed, so `position.parent.previousSibling` might be a text node, and the node you actually want to look at is `position.parent`.

This is often benign because:
* stock CKE plugin schemas don't contain elements that can trigger this set of circumstances
* console errors by default don't cause any symptoms

## 📃 Other details

* Browser: all
* OS: all
* Installed CKEditor plugins: `AutoLink`, any custom plugin that defines a schema item that's `isLimit`, can appear next to text, and allows typing inside

---

jonscheiding:

Fix: Handle newline-suppressed scenarios (e.g. limit elements) in AutoLink plugin. Closes #15862.

### DIFF
--- a/packages/ckeditor5-link/src/autolink.ts
+++ b/packages/ckeditor5-link/src/autolink.ts
@@ -240,11 +240,13 @@ export default class AutoLink extends Plugin {
 		enterCommand.on( 'execute', () => {
 			const position = model.document.selection.getFirstPosition()!;
 
-			if ( !position.parent.previousSibling ) {
-				return;
-			}
+			let rangeToCheck: Range;
 
-			const rangeToCheck = model.createRangeIn( position.parent.previousSibling as Element );
+			if ( position.parent.previousSibling?.is( 'element' ) ) {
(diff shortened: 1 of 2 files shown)


jonscheiding:
What are the next steps to get a review on this?

PR #15868, opened 2024-02-15
GHS should collect styles using StylesMap.

### LINKED ISSUE #10399: Border style reducer does not return all possible shorthands for CSS properties

Assuming this GHS pattern:

```js
styles: {
  'border-left': /.*/
}
```

and markup like below:

```html
<p style="border-left: 1px solid;">
  foobar
</p>
```

There is a bug where border style reducer returns only style properties like ` border-left-width: 1px` and `border-left-style: solid`, but not `border-left: 1px solid` which prevents Matcher from correctly matching pattern to the style of the element.

The source of this behavior can be found here: https://github.com/ckeditor/ckeditor5/blob/e6cd6e4/packages/ckeditor5-engine/src/view/styles/border.js#L331-L362 .

This is a follow-up to [this comment](https://github.com/ckeditor/ckeditor5/issues/10282#issuecomment-903706092).

---

### LINKED ISSUE #15757: Conflict between GeneralHtmlSupport and FontBackgroundColor

 On the branch [ssimonsen/general-html-support](https://github.com/sunesimonsen/ckeditor-bugs/tree/ssimonsen/general-html-support) the editor renders the background color for the heading:

Screenshot 2024-01-25 at 10 49 58

But when the [BackgroundFontColor](https://ckeditor.com/docs/ckeditor5/41.0.0/features/font.html) plugin is added the background color is no longer rendered:

Screenshot 2024-01-25 at 10 53 32

I tracked the problem do to [this line](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-font/src/fontbackgroundcolor/fontbackgroundcolorediting.ts#L105), when removed everything works again.

Minimal reproduction can be found [here](https://github.com/sunesimonsen/ckeditor-bugs/pull/1).

Notice that the table plugin also introduces the bug, as it also makes use of [addBackgroundRules](https://github.com/ckeditor/ckeditor5/blob/b4a74134b58042cb5e394f7a95dad9c4be00993f/packages/ckeditor5-engine/src/view/styles/background.ts#L37).

This bug is currently blocking Zendesk customers from applying background styles.

### ✔️ Expected result

That the background color is maintained.

### ❌ Actual result

The background inline style is stripped.

### ❓ Possible solution

Make [addBackgroundRules](https://github.com/ckeditor/ckeditor5/blob/b4a74134b58042cb5e394f7a95dad9c4be00993f/packages/ckeditor5-engine/src/view/styles/background.ts#L37) compatible with the general HTML support plugin.

## 📃 Other details

* First affected CKEditor version: probably all
* Installed CKEditor plugins: `@ckeditor/ckeditor5-basic-styles`, `@ckeditor/ckeditor5-essentials`, `@ckeditor/ckeditor5-paragraph`, `@ckeditor/ckeditor5-html-support`, `@ckeditor/ckeditor5-font`

---

niegowski:

Fix (html-support): Background color style should be properly preserved by GHS while the `FontBackgroundColor` plugin is enabled. It should be able to preserve a partly defined style. Closes #15757. Closes #10399.

---

### Additional information


arkflpc (approved):

packages/ckeditor5-html-support/src/datafilter.ts
```diff
@@ -805,129 +807,133 @@ export interface DataFilterRegisterEvent {
...
+	return attributesObject;
 }
 
 /**
- * Converts the given iterable object into an object.
+ * Returns list of styles as a normalized object (without redundant entries).
  */
-function iterableToObject( iterable: Set<string>, getValue: ( s: string ) => any ) {
-	const attributesObject: Record<string, any> = {};
+function normalizeStyles( viewElement: ViewElement, styles: Iterable<string> ): Record<string, string> {
```
pszczesniak, line 924: If the function returns a list we can add a prefix `get` to the name of function, WDYT?

So the name would be `getNormalizedStyles`.

packages/ckeditor5-html-support/src/datafilter.ts
```diff
@@ -805,129 +807,133 @@ export interface DataFilterRegisterEvent {
...
-	attributes: Set<string>;
-	styles: Set<string>;
-	classes: Set<string>;
-} {
-	const matchResult = {
-		attributes: new Set<string>(),
-		classes: new Set<string>(),
-		styles: new Set<string>()
-	};
+function normalizeAttributes( viewElement: ViewElement, attributes: Iterable<string> ): Record<string, string> {
```
pszczesniak, line 907: If the function returns a list we can add a prefix `get` to the name of function, WDYT?

So the name would be `getNormalizedAttributes`.

pszczesniak (approved):
Added two nitpicks and i'm curious of your opinion.

Other than that 👍

mmotyczynska (approved):

packages/ckeditor5-html-support/src/datafilter.ts
niegowski: Changed to `getReducedStyles()`.

packages/ckeditor5-html-support/src/datafilter.ts
niegowski: Changed to `getAttributes()`.

PR #15893, opened 2024-02-21
Application menu bar

### LINKED ISSUE #15894: Application menu bar

The editor deserves a proper application menu because our current toolbar setup gets too crowded to navigate and manage.

**Current approach:**

image

**Proposed approach:**

image

---

oleq:

Feature (ui,editor-classic,editor-decoupled): Implemented the application menu bar that contains various options and commands for controlling and navigating the editor. Closes #15894.

MINOR BREAKING CHANGE (upload): The `FileDialogButtonView` class has been moved from `ckeditor5-upload` to `ckeditor5-ui`. Please update your import paths accordingly (was: `import { FileDialogButtonView } from 'ckeditor5/src/upload.js';`, is: `import { FileDialogButtonView } from 'ckeditor5/src/ui.js';`).

MINOR BREAKING CHANGE (theme-lark): The default vertical spacing around `ButtonView` in `ListItemView` (`--ck-list-button-padding`) has been reduced for better readability. This affects the presentation of various editor features that use this type of UI (headings, font size, font family, etc.). You can restore the previous value by setting `--ck-list-button-padding: calc(.2 * var(--ck-line-height-base) * var(--ck-font-size-base)) calc(.4 * var(--ck-line-height-base) * var(--ck-font-size-base));` in your custom styles sheet.

---

### Additional information

**Requires https://github.com/cksource/ckeditor5-commercial/pull/6072**


packages/ckeditor5-ui/src/menubar/menubarmenupanelview.ts
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+}
+
+/**
+ * The names of the positions of the {@link module:ui/menubar/menubarmenupanelview~MenuBarMenuPanelView}.
+ *
+ * They are reflected as CSS class suffixes on the panel view element.
+ */
+export type MenuBarMenuPanelPosition = 'se' | 'sw' | 'ne' | 'nw' | 'w' | 'e';
```
scofalik, line 108: Didn't we have these registered in a different place?

packages/ckeditor5-alignment/src/alignmentui.ts
```diff
@@ -122,50 +191,62 @@ export default class AlignmentUI extends Plugin {
...
-			editor.editing.view.focus();
-		} );
+			for ( const option of options ) {
+				const listItemView = new MenuBarMenuListItemView( locale, menuView );
+				const buttonView = new MenuBarMenuListItemButtonView( locale );
 
-		return buttonView;
+				buttonView.extendTemplate( {
+					attributes: {
+						'aria-checked': buttonView.bindTemplate.to( 'isOn' )
```
scofalik, line 224: I wonder if this could be just default behavior of a button. But then `isOn` would need to be able to take `undefined` state.

packages/ckeditor5-list/src/list/utils.ts
```diff
@@ -8,43 +8,65 @@
...
-		buttonView.bind( 'isOn', 'isEnabled' ).to( command, 'value', 'isEnabled' );
+		return buttonView;
+	} );
 
-		// Execute command.
-		buttonView.on<ButtonExecuteEvent>( 'execute', () => {
-			editor.execute( commandName );
-			editor.editing.view.focus();
-		} );
+	editor.ui.componentFactory.add( `menuBar:${ commandName }`, () =>
```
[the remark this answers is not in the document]
scofalik: Or ... is it overwritten on purpose? If so, let's add comment in `listpropertiesui.ts` because it is confusing.

Oh, I see there is a comment there - I didn't notice it because it used to work like that already. I read the comment and I dislike it. It explains it backwards. I'd rewrite this comment, so it is more clear. 

E.g.: "Below we overwrite components registered in `ListUI` plugin. This way, depending on editor configuration, toolbar and menubar will either use the simple component from `ListUI` plugin, or extended component that manges list properties, defined here."

scofalik (commented):
I went through the PR and left some documentation-related comments and other loose thoughts, that we don't need to improve unless you strongly agree.

But, there is one thing that I don't like and it is the number of `MenuBar*` components that exist and need to be used to craft a UI component for menu bar. E.g. `MenuBarMenuListItemView` seems like an unnecessary complication for a plugin developer. If the component is really needed, maybe we could at least hide it from the developer.

Instead, the way I see it is that we have a small, defined set of items that can end up in the menu bar lists. Contrary to the toolbar, we don't want flexibility here. We agree that not everything makes sense in the menu bar. So, as I said, we have small lists of components to choose from.

Instead of crafting the components from the building blocks, it would be better (and safer!) to just give the developers the components. AFAICS we have four components now:

*   button -- a "leaf" in the menus structure, you click it and it makes an action and closes the menu bar, some of them are "toggles" (can be enabled, and have aria-checked attribute), this could be a parameter in the constructor,
*   file button -- probably it must be a separate component from the regular button,
*   sub-menu -- a button that opens a new menu, all I should care about is providing components for the menu, not creating the menu bar list view by hand and link it with other components (I mean things like event delegations),
*   dropdown -- a button that opens an arbitrary view in a dropdown, for things like color picker or table insertion grid, all I should care about is to provide the view.

So, we should have four "top-level" classes to be used by a plugin developer, with precise parameters in constructors and standarized behaviors.

I know that it sounds easy but then may get complicated in the details, especially I believe focus handling and keystrokes handling may be difficult for the dropdown component. But we should strive for the above, we should make it easy for plugin developers to provide the components. Right now, if I were to provide such component, I would be all in fear that I forgot to set some property, bind some observable, delegate an event, etc. Basically it is impossible to create a correct component without copy pasting existing, similar one.

(PS. this would not be a breaking change - we can build it over the existing components and update integrations as we go, depending on how much time we have.)

PR #15901, opened 2024-02-22
Tests in classic editor

mremiszewski:

Tests (classic-editor): Add tests for menu bar. Related to #15894.

---

### Additional information


packages/ckeditor5-editor-classic/tests/classiceditorui.js
```diff
@@ -756,31 +756,141 @@ describe( 'Focus handling and navigation between editing root and editor toolbar
...
+
+			// Try Alt+F9 again.
+			pressAltF9( editor );
+
+			sinon.assert.calledOnce( menuBarFocusSpy );
+			sinon.assert.notCalled( domRootFocusSpy );
+		} );
+	} );
+
+	describe( 'Restoring focus on Esc key press', () => {
```
oleq, line 833: **This is a follow-up**: We're going to need this kind of logic for the menu bar as well. 

https://github.com/ckeditor/ckeditor5/blob/76919d55c7825bd4401d4a473d6415922f80b49e/packages/ckeditor5-ui/src/editorui/editorui.ts#L437-L464

I coded the prototype as fast as possible so I didn't consider cases where the focus must get back to some other place (e.g. source editing). 

I suppose `EditorUI#addMenuBar` corresponding to `addToolbar` would be a good idea. Because currently, the keystroke handling code for the menu bar is in `ClassicEditorUI`, and if sb wanted to build their editor, they wouldn't be able to add a menu bar without duplicating the code.

cc @DawidKossowski

packages/ckeditor5-editor-classic/tests/classiceditoruiview.js
```diff
@@ -51,12 +52,12 @@ describe( 'ClassicEditorUIView', () => {
...
 			} );
 
-			it( 'is given a locate object', () => {
+			it( 'is given a locale object', () => {
 				expect( view.toolbar.locale ).to.equal( locale );
 			} );
 
 			it( 'is put into the "stickyPanel.content" collection', () => {
-				expect( view.stickyPanel.content.get( 0 ) ).to.equal( view.toolbar );
+				expect( view.stickyPanel.content.has( view.toolbar ) ).to.be.true;
```
oleq, line 60: But in which order? We should assert that too.

packages/ckeditor5-editor-classic/tests/classiceditoruiview.js
```diff
@@ -82,12 +83,26 @@ describe( 'ClassicEditorUIView', () => {
...
+			it( 'is created', () => {
+				expect( view.menuBarView ).to.be.instanceof( MenuBarView );
+			} );
+
+			it( 'is given a locale object', () => {
+				expect( view.menuBarView.locale ).to.equal( locale );
+			} );
+
+			it( 'is put into the "stickyPanel.content" collection', () => {
+				expect( view.stickyPanel.content.has( view.menuBarView ) ).to.be.true;
```
oleq, line 96: But in which order? We should assert that too.

oleq (changes requested):

PR #15903, opened 2024-02-22
Allow registering translations in the `defaultConfig`

### LINKED ISSUE #15902: The `translations` object is ignored when used in `defaultConfig`
The newly added `translations` configuration option only works when used in the second parameter of the `Editor.create()` method. However, it should also be possible to define it in the `defaultConfig` property of the `Editor` class:

```ts
export default class Editor extends ClassicEditor {
  public static override builtinPlugins = [ /* ... */ ];

  public static override defaultConfig = {
    /* ... */
    translations,
  };
}
```

filipsobol:

Fix (core): Use `translations` from the `defaultConfig` if they were not provided in the `create` method. Closes #15902.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-core/src/editor/editor.ts
+++ b/packages/ckeditor5-core/src/editor/editor.ts
@@ -280,7 +280,10 @@ export default abstract class Editor extends ObservableMixin() {
 
 		// We don't pass translations to the config, because its behavior of splitting keys
 		// with dots (e.g. `resize.width` => `resize: { width }`) breaks the translations.
-		const { translations, ...rest } = config;
+		const {
+			translations = ( constructor.defaultConfig && constructor.defaultConfig.translations ),
+			...rest
+		} = config;
 
 		this._context = config.context || new Context( { language, translations } );
 		this._context._addEditor( this, !config.context );


packages/ckeditor5-core/src/editor/editor.ts
```diff
@@ -268,19 +268,20 @@ export default abstract class Editor extends ObservableMixin() {
 	 *
 	 * Usually, not to be used directly. See the static {@link module:core/editor/editor~Editor.create `create()`} method.
 	 *
-	 * @param config The editor configuration.
+	 * @param providedConfig The editor configuration.
```
arkflpc, line 271: Do we want to change the API documentation because we wanted to use the `config` name for local variable?

illia-stv (approved):
LGTM ✅

packages/ckeditor5-core/src/editor/editor.ts
filipsobol: I wasn't sure about this change too. I reverted it, so that we don't change our API docs.

pszczesniak (approved):

PR #15904, opened 2024-02-22
Add menu bar integration to block quote.

mremiszewski:

Feature (block-quote): Add menu bar integration to block quote package. Related to #15894.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-block-quote/src/blockquoteui.ts
+++ b/packages/ckeditor5-block-quote/src/blockquoteui.ts
@@ -8,7 +8,7 @@
  */
 
 import { Plugin, icons } from 'ckeditor5/src/core.js';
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-block-quote/src/blockquoteui.ts
```diff
@@ -33,29 +33,51 @@ export default class BlockQuoteUI extends Plugin {
 	 */
 	public init(): void {
 		const editor = this.editor;
-		const t = editor.t;
+		const command: BlockQuoteCommand = editor.commands.get( 'blockQuote' )!;
```
niegowski, line 36: This change might require adding `requires` getter to ensure that the feature's editing part is loaded before the UI part. @oleq

PR #15912, opened 2024-02-23
Docs: fixing the snippet.

### LINKED ISSUE #15733: Docs issue report from "installation/getting-started/getting-and-setting-data.html"
### Origin URL

https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/getting-and-setting-data.html

### Project version

41.0.0

### Is the information outdated? How?

_No response_

### Is there something missing in the guide? What is it?

The PHP example is wrong. It seems the encoding charactes which should be shown in the documentation got encoded wrong. Now it shows exactly what should not happen.

### Is there anything else you would like to add?

_No response_

### User agent

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36 Edg/120.0.0.0

godai78:

Docs: fixing the snippet. Closes #15733

---

### Additional information

### DIFF
--- a/docs/installation/getting-started/getting-and-setting-data.md
+++ b/docs/installation/getting-started/getting-and-setting-data.md
@@ -125,7 +125,7 @@ When you print the data from the database to a `<textarea>` element in an HTML p
 Thanks to that, the `<textarea>` will be printed out like this:
 
 ```html
-&lt;textarea&lt;p>This is some sample content.&lt;/p>&lt/textarea>
+&lt;textarea>&lt;p>This is some sample content.&lt;/p>&lt;/textarea>
 ```
 
 Instead of being printed like this:
--- a/docs/installation/getting-started/getting-and-setting-data.md
+++ b/docs/installation/getting-started/getting-and-setting-data.md
@@ -112,31 +112,29 @@ Please note that the replaced `<textarea>` element is updated automatically by C
 If you need to get the actual data from CKEditor at any moment using JavaScript, use the {@link module:editor-classic/classiceditor~ClassicEditor#getData `editor.getData()`} method as described in the next section.
 </info-box>
 
-<info-box>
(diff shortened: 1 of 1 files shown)


gorzelinski:
I tried different methods of getting this worked, like this:

```html
<info-box>
  ```html
    // HTML entities
  \```
</info-box>
```
or

```html
<info-box>
  <pre><code> // HTML entities </code></pre>
</info-box>
```

etc.

No matter what, HTML entities are just rendered when they are inside of an info-box. Rendered snippets look the same.

Screenshot 2024-02-23 at 12 12 17

Because of that, I would advise to **avoid using code snippets inside info-boxes.** It can mess up the markup, and it adds unnecessary nesting.

In general, displaying HTML entities instead of rendering them in a markdown is [tricky](https://stackoverflow.com/questions/47982386/html-entities-in-markdown-block-code).

I've also just noticed that the info box messes up the PHP snippet. Replacing '&' with '&' doesn't make sense. 

Screenshot 2024-02-23 at 12 41 46

godai78:
It actually works my way.  
  
 



But I think we may drop this infobox anyway and just have it as text.

gorzelinski:
That's interesting... Hmm...
Still, as far as I see, the PHP snippet doesn't work as expected.

gorzelinski:
The problem is a little tricky. I'll try to suggest a solution. Stay tuned.

godai78:
This is after the infobox was dropped:

godai78:
I'll push it and you can tak up from here.

gorzelinski:
Now, it looks as it should. Probably.

gorzelinski (commented):
Now, it looks as it should. Probably.

PR #15925, opened 2024-02-26
Updating using two editors docs

filipsobol:

Docs: Update "Using two editors" docs.

---

### Additional information

### DIFF
--- a/docs/installation/advanced/using-two-editors.md
+++ b/docs/installation/advanced/using-two-editors.md
@@ -38,7 +38,7 @@ Now it is time to add the missing editor package and install it:
 yarn add -D @ckeditor/ckeditor5-editor-inline
 ```
 
-Once all the dependencies are installed, you'll have tochange the `src/ckeditor.ts` file which for now only exports only a single class. The first step is to move all plugins and configuration to a variables, so they can be reused by both editors:
+Once all the dependencies are installed, you'll need to modify the `src/ckeditor.ts` file, which currently only exports a single class. The first step is to move all plugins and configuration to a variables so they can be reused by both editors:
 
 ```js
 import { ClassicEditor as ClassicEditorBase } from '@ckeditor/ckeditor5-editor-classic';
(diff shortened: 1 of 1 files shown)


godai78 (approved):

docs/installation/advanced/using-two-editors.md
```diff
@@ -29,91 +29,88 @@ You can start by forking (or copying) an existing build like in the {@link insta
...
 Now it is time to add the missing editor package and install it:
 
-```
-npm install --save-dev @ckeditor/ckeditor5-editor-inline
+```bash
+yarn add -D @ckeditor/ckeditor5-editor-inline
 ```
 
-Once all the dependencies are installed, change the webpack's entry point which is the `src/ckeditor.js` file. For now, it was exporting just a single class:
+Once all the dependencies are installed, you will need to modify the `src/ckeditor.ts` file, which currently only exports a single class. The first step is to move all plugins and configuration to variables so they can be reused by both editors:
```
gorzelinski, line 41:
```suggestion
After installing all the dependencies, modify the `src/ckeditor.ts` file, which currently only exports a single class. The first step is to move all plugins and configuration to variables so they can be reused by both editors:
```

Maybe we could avoid passive voice?

docs/installation/advanced/using-two-editors.md
```diff
@@ -29,91 +29,88 @@ You can start by forking (or copying) an existing build like in the {@link insta
...
+ClassicEditor.builtinPlugins = plugins;
+ClassicEditor.defaultConfig = config;
+
+export {
+  ClassicEditor
 };
 ```
 
-Let's make it export an object with two classes: `ClassicEditor` and `InlineEditor`. To make both constructors work in the same way (load the same plugins and default configuration) you also need to assign `builtinPlugins` and `defaultConfig` static properties to both of them:
+Now you can add the `InlineEditor` class to the file, add the same plugins and configuration to it and export it:
```
gorzelinski, line 66:
```suggestion
Now, you can add the `InlineEditor` class to the file. Add the same plugins and configuration to the class and then export it:
```

I would change the punctuation a little.

gorzelinski (commented):
I added some suggestions, but merging without them would be ok.

filipsobol:
I like these suggestions. I will make another PR @gorzelinski

PR #15930, opened 2024-02-26
Update "Using two editors" docs

filipsobol:

Docs: Update "Using two editors" docs.

---

### Additional information

Follow-up to the https://github.com/ckeditor/ckeditor5/pull/15925

### DIFF
--- a/docs/installation/advanced/using-two-editors.md
+++ b/docs/installation/advanced/using-two-editors.md
@@ -38,7 +38,7 @@ Now it is time to add the missing editor package and install it:
 yarn add -D @ckeditor/ckeditor5-editor-inline
 ```
 
-Once all the dependencies are installed, you will need to modify the `src/ckeditor.ts` file, which currently only exports a single class. The first step is to move all plugins and configuration to variables so they can be reused by both editors:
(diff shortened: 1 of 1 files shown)


docs/installation/advanced/using-two-editors.md
```diff
@@ -38,7 +38,7 @@ Now it is time to add the missing editor package and install it:
 yarn add -D @ckeditor/ckeditor5-editor-inline
 ```
 
-Once all the dependencies are installed, you will need to modify the `src/ckeditor.ts` file, which currently only exports a single class. The first step is to move all plugins and configuration to variables so they can be reused by both editors:
+Next, modify the `src/ckeditor.ts` file, which currently only exports a single editor class. Start by moving all plugins and configuration to variables so they can be reused by both editors:
```
gorzelinski, line 41:
```suggestion
Next, modify the `src/ckeditor.ts` file, which only exports a single editor class. Start by moving all plugins and configuration to variables so both editors can reuse them:
```
What about removing passive voice?

gorzelinski (approved):
Just one nitpick - nothing major.

PR #16014, opened 2024-03-12
Add a `mergeTag` element to `inlineObjectElements` array in the `DomConverter` class

Dumluregn:

Internal (merge-tags): Added a `mergeTag` element to the `inlineObjectElements` array in the `DomConverter` class.

---

### Additional information

This allows to preserve the whitespace after the `mergeTag` element. At first I came up with a more complex and specific solution in https://github.com/ckeditor/ckeditor5/pull/16014/commits/d1fe0757c65dbeb75209191a92469d99c77d4a0c, but later it turned out there's a simpler one.


DawidKossowski (approved):
For now, I don't have a better idea of how we could improve it. Nevertheless, it's something that @scofalik could think about.

DawidKossowski:
And of course, unit tests are missing. So you could add them before merging.

PR #16024, opened 2024-03-14
Add the visual label to the beginning of the programmatic heading dropdown aria label

Mati365:

Type: Add the visual label to the beginning of the programmatic aria label in heading dropdown. Closes https://github.com/cksource/ckeditor5-commercial/issues/6029

### DIFF
--- a/packages/ckeditor5-heading/src/headingui.ts
+++ b/packages/ckeditor5-heading/src/headingui.ts
@@ -15,7 +15,7 @@ import {
 	type ButtonExecuteEvent,
 	type ListDropdownItemDefinition
 } from 'ckeditor5/src/ui.js';
-import { Collection } from 'ckeditor5/src/utils.js';
+import { Collection, type ObservableChangeEvent } from 'ckeditor5/src/utils.js';
 import type { ParagraphCommand } from 'ckeditor5/src/paragraph.js';
 
 import { getLocalizedOptions } from './utils.js';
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-heading/src/headingui.ts
```diff
@@ -124,6 +124,21 @@ export default class HeadingUI extends Plugin {
 				return titles[ whichModel ];
 			} );
 
+			dropdownView.buttonView.bind( 'ariaLabel' ).to( headingCommand, 'value', paragraphCommand, 'value', ( value, para ) => {
+				const whichModel = value || para && 'paragraph';
```
oleq, line 128: Since we're visiting this code for the first time since 2018, can you please rewrite it to human-readable code using if/else or at least the ternary operator? 🙂 (I know you copied it from `dropdownView.buttonView.bind( 'label' )...`, you can change it there too)

oleq (changes requested):
The solution is correct. What we're missing:

* unit tests,

and as per [original issue](https://github.com/cksource/ckeditor5-commercial/issues/6029)

>The same issue exists in other dropdown controls. 

* an integration with text part language (the aria-label is "Language" but it displays "French")
image

packages/ckeditor5-heading/src/headingui.ts
Mati365: I reduced this one line to `const whichModel = paragraph ? 'paragraph' : heading;`. Seems to be working correctly.

packages/ckeditor5-language/src/textpartlanguageui.ts
```diff
@@ -106,6 +106,16 @@ export default class TextPartLanguageUI extends Plugin {
...
 			} );
 
+			dropdownView.buttonView.bind( 'ariaLabel' ).to( languageCommand, 'value', value => {
+				const selectedLanguageTitle = value && titles[ value ];
+
+				if ( !selectedLanguageTitle ) {
+					return accessibleLabel;
+				}
+
+				return `${ accessibleLabel }, ${ selectedLanguageTitle }`;
```
oleq, line 116: In the heading feature you went with "Current value, Feature name" ("Heading 2, Heading") (as per [Tomek's comment](https://github.com/cksource/ckeditor5-commercial/issues/6029)). Here, you have the opposite: "Feature name, Current value" ("Language, French"). I think text part language should be aligned to headings in this matter.

oleq (changes requested):

packages/ckeditor5-language/src/textpartlanguageui.ts
Mati365: Agh, you are correct. I'm fixing it.

Mati365: Fixed.

PR #16029, opened 2024-03-14
Unable to remove an image placed inside of an editable field in restricted editing mode

### LINKED ISSUE #15521: Unable to remove an image placed inside of an editable field in restricted editing mode

1. Create some content in which you specify editable and non-editable areas via restricted editing
2. Enable the `'insertImage'` and/or `'uploadImage'` commands in `restrictedEditing`
3. Turn on restricted editing mode
4. Place the caret at the beginning or end of the editable field
5. Insert an inline image in this position
6. Try removing the image once it's inserted

### ✔️ Expected result

You should be able to remove the image, as is the case when it's inserted somewhere in the middle of the editable field:

https://github.com/ckeditor/ckeditor5/assets/72079603/9da9b4ff-d717-4eb4-9b4a-b79f91c9e805

### ❌ Actual result

If the caret is at the beginning or end of the editable field, the image cannot be removed:

https://github.com/ckeditor/ckeditor5/assets/72079603/7526efbc-da7b-4c4c-ac7a-fa5906f841bc

### ❓ Possible solution

When the caret is placed at the beginning or end of the editable field, the image gets inserted outside of the `span` with the `restricted-editing-exception` class. In this case, the markup can look like this:

```
<p>Non-editable</p><p><picture><source srcset="https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/ZD-j86KZtZQ8/images/80.webp 80w,https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/ZD-j86KZtZQ8/images/160.webp 160w,https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/ZD-j86KZtZQ8/images/240.webp 240w,https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/ZD-j86KZtZQ8/images/256.webp 256w" sizes="(max-width: 256px) 100vw, 256px" type="image/webp"><img data-ckbox-resource-id="ZD-j86KZtZQ8" src="https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/ZD-j86KZtZQ8/images/256.png" width="256" height="256"></picture><span class="restricted-editing-exception">Editable content here</span></p><p>Non-editable</p>
```

When the caret is placed somewhere in the middle of the editable field, the image is placed inside of the span and the markup can look like this:

```
<p>Non-editable</p><p><span class="restricted-editing-exception">Editable content&nbsp;</span><picture><source srcset="https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/5wtlAcRPSgai/images/80.webp 80w,https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/5wtlAcRPSgai/images/160.webp 160w,https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/5wtlAcRPSgai/images/240.webp 240w,https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/5wtlAcRPSgai/images/256.webp 256w" sizes="(max-width: 256px) 100vw, 256px" type="image/webp"><img data-ckbox-resource-id="5wtlAcRPSgai" src="https://ckbox.cloud/4NCAhHInx1m9uHJjNYDz/assets/5wtlAcRPSgai/images/256.png" width="256" height="256"></picture><span class="restricted-editing-exception">here</span></p><p>Non-editable</p>
```

## 📃 Other details

* Browser: cross-browser
* OS: cross-OS

---

illia-stv:

Fix (restricted-editing): Should be able to remove an image placed inside of an editable field in restricted editing mode. Closes #15521.

---

### Additional information


packages/ckeditor5-restricted-editing/src/restrictededitingmode/converters.ts
```diff
@@ -108,7 +108,7 @@ export function extendMarkerOnTypingPostFixer( editor: Editor ): ModelPostFixer
 		let changeApplied = false;
 
 		for ( const change of editor.model.document.differ.getChanges() ) {
-			if ( change.type == 'insert' && change.name == '$text' ) {
+			if ( change.type == 'insert' && ( change.name == '$text' || change.name == 'imageInline' ) ) {
```
niegowski, line 111: This would not work for other inline widgets. Please use schema to check if it is an inline object allowed where text.

packages/ckeditor5-restricted-editing/tests/restrictededitingmodeediting.js
```diff
@@ -476,6 +482,62 @@ describe( 'RestrictedEditingModeEditing', () => {
...
+
+			model.change( writer => {
+				writer.addMarker( 'restrictedEditingException:1', {
+					range: writer.createRange( writer.createPositionAt( firstParagraph, 4 ), writer.createPositionAt( firstParagraph, 7 ) ),
+					usingOperation: true,
+					affectsData: true
+				} );
+			} );
+
+			command.execute( { source: imgSrc } );
```
niegowski, line 499: Why not `editor.execute(...)`?

niegowski (changes requested):

packages/ckeditor5-restricted-editing/src/restrictededitingmode/converters.ts
illia-stv: Is it possible to verify element in `DiffItemInsert` with schema?
image

packages/ckeditor5-restricted-editing/tests/restrictededitingmodeediting.js
illia-stv: Because there is no `insertImage` command in editor. So we would need to add `Image` plugin and when we add `Image`,  tests are crash. For some reason editor created with `VirtualTestEditor` has no `ui` property so it causes this kind of errors "**_Cannot read property of undefined_**". In the code base there is only one example of usage `insertImage` with editor.**execute()** and it was used with `ClassicEditor` so final decision was to go with simple `new InsertImageCommand()`.

image

packages/ckeditor5-restricted-editing/src/restrictededitingmode/converters.ts
illia-stv: Nevermind, indeed there is a way, `schema.isInline( change.position.nodeAfter! )`

packages/ckeditor5-restricted-editing/tests/restrictededitingmodeediting.js
illia-stv: Unfortunately, still the same problem when I add `Image` to plugins.

packages/ckeditor5-restricted-editing/src/restrictededitingmode/converters.ts
```diff
@@ -106,9 +106,10 @@ export function extendMarkerOnTypingPostFixer( editor: Editor ): ModelPostFixer
 	// This post-fixer shouldn't be necessary after https://github.com/ckeditor/ckeditor5/issues/5778.
 	return writer => {
 		let changeApplied = false;
+		const schema = editor.model.schema;
 
 		for ( const change of editor.model.document.differ.getChanges() ) {
-			if ( change.type == 'insert' && change.name == '$text' ) {
+			if ( change.type == 'insert' && schema.checkChild( '$block', change.name ) ) {
```
mmotyczynska, line 112: Why does this work? Are there other cases that are handled in addition to the inline image? Asking because I'm not that familiar with schema and want to learn something new 🙂

packages/ckeditor5-restricted-editing/tests/manual/restrictedediting.js
```diff
@@ -90,4 +111,11 @@ async function reloadEditor( config ) {
 	}
 
 	window.editor = await ClassicEditor.create( document.querySelector( '#editor' ), config );
+	window.editor.plugins._plugins.delete( 'ImageBlockEditing' );
```
mmotyczynska, line 114: This looks like a hack. You could, for example, adjust the list of loaded plugins and use [removePlugins](https://ckeditor.com/docs/ckeditor5/latest/api/module_core_editor_editorconfig-EditorConfig.html#member-removePlugins) instead.

mmotyczynska (commented):
The fix is working as expected and the tests are doing their job 👍 I added a few comments.

packages/ckeditor5-restricted-editing/tests/manual/restrictedediting.js
illia-stv: Yeah, its look a hack. But unfortunetely we can not use `removePlugins`, because `Image` requires `ImageBlockEditing`. `Image` is in the `ArticlePluginSet` then its difficult to exclude this plugin. So the simplest way is just remove `ImageBlockEditing` after building. Or we could just import every package from `ArticlePluginSet` except `Image`.

packages/ckeditor5-restricted-editing/src/restrictededitingmode/converters.ts
illia-stv: Our schema registers relation between various types of elements in editor. For example `$block` elements e.g. `paragraph` allows `$text` inside or other inline elements. `schema.checkChild(...)` verifies does element allowed be inside `$block` or not.

packages/ckeditor5-restricted-editing/tests/manual/restrictedediting.js
```diff
@@ -51,6 +57,10 @@ async function startStandardEditingMode() {
 				'mergeTableCells'
 			]
 		},
+		ckfinder: {
```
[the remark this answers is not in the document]
illia-stv: Interesting, because I have no errors.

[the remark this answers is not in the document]
mmotyczynska: Let's wait for another reviewer then. I've reinstalled packages and still encounter this error 🤔

niegowski (approved):

PR #16031, opened 2024-03-14
Do not regenerate name of markers that are not copyable

Mati365:

Fix (clipboard): The imported Word document that contains suggestions will no longer cause errors. 

Closes https://github.com/cksource/ckeditor5-commercial/issues/6081.


charlttsie:
I've verified the fix and the initial issue doesn't reproduced. I also haven't found any regressions, so looks good 👍

DawidKossowski (approved):

arkflpc (approved):

PR #13984, opened 2023-04-26
Change color only when color dropdown is opened

Dumluregn:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Internal(font): Change color only when color dropdown is opened.

---

### Additional information

_For example – encountered issues, assumptions you had to make, other affected tickets, etc._

PR #16034, opened 2024-03-14
Add menu bar integration to font.

mremiszewski:

Internal (font): Add menu bar integration. Related to https://github.com/ckeditor/ckeditor5/issues/15894.

---

### Additional information


packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -208,5 +210,77 @@ export default class ColorUI extends Plugin {
...
+					} );
+				}
+
+				if ( data.source === 'colorPicker' ) {
+					evt.stop();
+				}
+
+				if ( data.source === 'colorPickerSaveButton' ) {
+					menuView.isOpen = false;
+				}
```
oleq, line 261: I think this is obsolete since we don't have a picker here.

packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -208,5 +210,77 @@ export default class ColorUI extends Plugin {
...
+				if ( menuView.isOpen ) {
+					editor.execute( this.commandName, {
+						value: data.value,
+						batch: this._undoStepBatch
+					} );
+				}
+
+				if ( data.source === 'colorPicker' ) {
+					evt.stop();
+				}
```
oleq, line 257: I think this is obsolete since we don't have a picker here.

packages/ckeditor5-font/src/ui/colorui.ts
```diff
@@ -208,5 +210,77 @@ export default class ColorUI extends Plugin {
...
+				documentColorsLabel: documentColorsCount !== 0 ? t( 'Document colors' ) : '',
+				documentColorsCount: documentColorsCount === undefined ? this.columns : documentColorsCount,
+				colorPickerViewConfig: false
+			} );
+
+			colorSelectorView.bind( 'selectedColor' ).to( command, 'value' );
+
+			colorSelectorView.delegate( 'execute' ).to( menuView );
+			colorSelectorView.on<ColorSelectorExecuteEvent>( 'execute', ( evt, data ) => {
+				if ( menuView.isOpen ) {
```
oleq, line 248: In the current shape, there's no way to execute a collor without opening the menu. This condition was introduced for odd cases where the color picker propagated the `execute` event when the user changed the selection in the document but the dropdown was closed (see https://github.com/ckeditor/ckeditor5/pull/13984).

packages/ckeditor5-font/src/fontfamily/fontfamilyui.ts
```diff
@@ -73,6 +83,46 @@ export default class FontFamilyUI extends Plugin {
...
+
+			const listView = new MenuBarMenuListView( locale );
+
+			for ( const definition of listOptions ) {
+				const listItemView = new MenuBarMenuListItemView( locale, menuView );
+				const buttonView = new MenuBarMenuListItemButtonView( locale );
+
+				buttonView.extendTemplate( {
+					attributes: {
+						'aria-checked': buttonView.bindTemplate.to( 'isOn' )
```
oleq, line 103: I think at this point we really need `ButtonView#ariaChecked` and:
* An update here,
* An update in `FontSizeUI`,
* An update in `HeadingUI` (it's missing),
* An update in dropdown utils (`bindViewCollectionItemsToDefinitions()`).

oleq (changes requested):
I noticed that focus is not moved to the editing root (it's lost) upon applying a color/backgrund color. We're missing some `editor.editing.view.focus()` somewhere.

PR #16035, opened 2024-03-14
Add menu bar integration to highlight.

mremiszewski:

Internal (highlight): Add menu bar integration. Related to https://github.com/ckeditor/ckeditor5/issues/15894.

---

### Additional information


oleq (changes requested):
Everything works fine except for some visual issues:

1. We need a separator (`ListSeparatorView`) between markers and the remove button.
2. We need transparent fills for the main and active icons:

Screenshot 2024-03-15 at 16 53 17

so they become this

Screenshot 2024-03-15 at 17 01 47

I know a transparent fill is not the best UX for the active item but it's better than this. It will be resolved once we address the problem of checkboxes displayed next to active list items in the future.

PR #16039, opened 2024-03-15
Fix (ckeditor5-engine): use content.appendChild for templates

not-paul-v:

Type: Message. Closes #000.

---

### Additional information

We are using CKEditor5 to create templates for documents. We use the editor's data to render the document with AlpineJS to do additional logic like inserting variable values or conditionally rendering blocks.

For certain features like the latter we need to use template elements.

We used the following converter to downcast to a template tag based on an attribute being present:
```typescript
conversion.for('downcast').attributeToElement({
  model: 'condition',
  view: (value, { writer }) => {
    return writer.createAttributeElement('template', {
      'x-if': value,
    });
  },
});
```

This works fine for the view, elements with the attribute (e.g. a paragraph) get wrapped with a `template` tag as expected.

But when we try to get the editors data using `editor.getData()` the children of the `template` are always missing. If we replace the `template` tag with other tags like a `div`, it works as expected.

After looking into it and it seems like the cause is inside `domconverter.ts` which always appends the children using `appendChild`. This does not work when the dom element is a `template` since you need to append the children to `content`.

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of different DOM elements in the editor to improve content structuring.
	- Improved data parsing in the editor for templates with nested elements.

### DIFF
--- a/packages/ckeditor5-engine/src/view/domconverter.ts
+++ b/packages/ckeditor5-engine/src/view/domconverter.ts
@@ -460,7 +460,11 @@ export default class DomConverter {
 
 			if ( options.withChildren !== false ) {
 				for ( const child of this.viewChildrenToDom( viewElementOrFragment, options ) ) {
-					domElement.appendChild( child );
+					if ( domElement instanceof HTMLTemplateElement ) {
+						domElement.content.appendChild( child );
+					} else {
+						domElement.appendChild( child );
+					}
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-engine/src/view/domconverter.ts
```diff
@@ -734,8 +738,16 @@ export default class DomConverter {
 		options: Parameters<DomConverter[ 'domToView' ]>[ 1 ] = {},
 		inlineNodes: Array<ViewNode> = []
 	): IterableIterator<ViewNode> {
-		for ( let i = 0; i < domElement.childNodes.length; i++ ) {
-			const domChild = domElement.childNodes[ i ];
+		// Get child nodes from content document fragment if element is template
+		let childNodes: Array<ChildNode> = [];
+		if ( domElement instanceof HTMLTemplateElement ) {
```
scofalik, line 743:
```suggestion
		let childNodes: Array<ChildNode> = [];

		if ( domElement instanceof HTMLTemplateElement ) {
```

Code style issue.

scofalik (approved):
Just one simple code style issue. Other than that LGTM. Thanks for checking the upcast and writing extra test.

PR #16041, opened 2024-03-15
Fix (ckbox): use safer way for stringifying numbers

### LINKED ISSUE #16040: SyntaxError: Invalid or unexpected token 50.toString()

Some algorithms of minification turn code like this `ITEMS_PER_REQUEST.toString()` to `50.toString()` - substitute the variable's value. This leads to the error: `SyntaxError: Invalid or unexpected token`.

The place where I personally have this problem
https://github.com/ckeditor/ckeditor5/blob/54eae6a1bb1a72d42f00b43fcf7f7f0b167507ba/packages/ckeditor5-ckbox/src/ckboxutils.ts#L228

### ❓ Possible solution

Following the specification of `Number.prototype.toString()`, using another way of stringifying numbers will be safer. I suggest using `String(number)`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toString

---

maxmartynov:

Fix (ckbox): use safer way for stringifying numbers. Closes #16040

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ckbox/src/ckboxutils.ts
+++ b/packages/ckeditor5-ckbox/src/ckboxutils.ts
@@ -225,8 +225,8 @@ export default class CKBoxUtils extends Plugin {
 		function fetchCategories( offset: number ): Promise<{ totalCount: number; items: Array<AvailableCategory> }> {
 			const categoryUrl = new URL( 'categories', serviceOrigin );
 
-			categoryUrl.searchParams.set( 'limit', String(ITEMS_PER_REQUEST) );
-			categoryUrl.searchParams.set( 'offset', String(offset) );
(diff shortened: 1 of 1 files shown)


bobby5892:
This pull request does need tested and merged in.

The risk of not having this is that when ckeditor is used behind minifcation (cloudflare etc) this causes
50 .toString()  <--- valid
to become
50.toString() <--- invalid JS

This caused a production outage for us to track down.

Mati365:
@maxmartynov I picked your modification to new [PR](https://github.com/ckeditor/ckeditor5/pull/16726) and added regression test to our test suite. I'm closing this particular PR.

PR #16043, opened 2024-03-15
Expose shared components to the ui library and add scroll reset to the `SearchTextView#reset()` method

Dumluregn:

Feature (ui): Add the `LabelWithHighlightView` and `ButtonLabelWithHighlightView` components to the common UI library.
Feature (ui): Add the `filterGroupAndItemNames()` helper to the common UI library. 
Feature (ui): The `SearchTextView#reset()` method will also reset the scroll of its `filteredView` to the top.

---

### Additional information


DawidKossowski (approved):
LGTM. We can only confirm with Olek whether there is anything here that should be private or is not ready for being public API.

PR #16047, opened 2024-03-17
Internal (ckeditor5): Upgrade eslint to v8.

fedemp:
Change how eslint formatter is created so it's compatible with eslint v8. This should allow to update eslint in ckeditor5-linters-config later. More info at
https://github.com/eslint/eslint/blob/main/docs/src/use/migrate-to-8.0.0.md#remove-cliengine

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "chalk": "^4.1.0",
     "coveralls": "^3.1.0",
     "date-fns": "^2.30.0",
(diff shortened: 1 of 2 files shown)


pomek:
Closing the PR as ESLint@8 became a part of the CKEditor 5 environment while we were migrating `ckeditor5-dev` packages to ESM. See #17132.

Issue #16053, opened 2024-03-18
Add aria-label to code block


CaitBarnard:
## 📝 Provide a description of the improvement

When you add a code block to the editor, there is a small label on the right that says which language the code block is. This is inaccessible as an assistive technology user may not necessarily be able to pick that label up.

Is it possible to have some sort of event listener on the code blocks which updates the aria-label on this element?

![Screenshot 2024-03-18 at 13 19 16](https://github.com/ckeditor/ckeditor5/assets/54268863/1b1531ca-f8dc-4eea-9fa7-fa22fd76b106)

```
<pre data-language="Python" spellcheck="false" aria-label="python"><code class="language-python">Test code block</code></pre>
```

## 📃 Other details

This is a snippet of how I'm solving the issue currently. However it only fires on change and so doesn't provide an aria-label when loading a page that already has an editor with a code block. It's also not sensible to have a setTimeout which is there as the change trigger takes a moment to generate pre elements and prevents it being null before it's had a moment to update.
```
for (let id in editors) {
        editors[id].model.document.on('change', () => {
          setTimeout(() => {
            const preElements = document.querySelectorAll('pre');

            preElements.forEach(function (pre) {
              const code = pre.querySelector('code[class^="language-"]');
              if (code) {
                const language = code.className.replace('language-', '');
                pre.setAttribute('aria-label', language);
              }
            });
          }, 300);
        });
      }
```

* Browser: Chrome
* OS: Mac OS
* CKEditor version: 5
* Installed CKEditor plugins: …

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

Mati365:
Hi @CaitBarnard! Thanks for reporting the issue, I added `aria-label` attribute to the `code` element in this PR: https://github.com/ckeditor/ckeditor5/pull/16056. Can you take a look and let us know if this solves the reported problem?

CaitBarnard:
Hi @Mati365. That solves the issue perfectly. Thank you for sorting that.

Mati365:
@CaitBarnard We checked docs, and it seems that `aria-label` is ignored in most of non-interactive elements (including `code` tag).   Are you using these `aria-labels` with specific screen reader that supports it?

CaitBarnard:
@Mati365 My bad, that's a good point that I completely overlooked. I believe the pre element can have an aria label. Is it possible to add it there instead? I've seen a few examples where a pre tag has an aria-label but I've yet to test it myself. 

[...]

PR #16056, opened 2024-03-19
 Add support for screen reader announcements in code block document elements. 

Mati365:

Feature (code-block):  Introduced screen reader announcements for entering or exiting code block elements in the document editor. Closes https://github.com/ckeditor/ckeditor5/issues/16053.


packages/ckeditor5-code-block/src/codeblockediting.ts
```diff
@@ -19,18 +19,21 @@ import {
 	type DowncastInsertEvent,
 	type UpcastElementEvent,
 	type UpcastTextEvent,
-	type Element
+	type Element,
+	type SelectionChangeRangeEvent
 } from 'ckeditor5/src/engine.js';
 
 import type { ListEditing } from '@ckeditor/ckeditor5-list';
+import type { CodeBlockLanguageDefinition } from './codeblockconfig.js';
```
oleq, line 27: This is obsolete.

packages/ckeditor5-code-block/src/codeblockediting.ts
```diff
@@ -277,6 +280,43 @@ export default class CodeBlockEditing extends Plugin {
...
+	 *
+	 * @internal
+	 */
+	private _initAriaAnnouncements( ) {
+		const { model, ui, t } = this.editor;
+		const languageDefs = getNormalizedAndLocalizedLanguageDefinitions( this.editor );
+
+		let lastFocusedCodeBlock: Element | null = null;
+
+		model.document.selection.on<SelectionChangeRangeEvent>( 'change:range', () => {
```
oleq, line 299: This works fine as long as the user moves between a code block and a non-code block context. If they move between two code blocks (create two code blocks, put the caret in the last line of the first one, then press arrow down), only the "Entering..." part is read out, while it should be "Leaving..." followed by "Entering...".

It's the same story as with the comment markers.

packages/ckeditor5-code-block/src/utils.ts
```diff
@@ -258,3 +259,32 @@ export function canBeCodeBlock( schema: Schema, element: Element ): boolean {
...
+		}
+
+		return t( 'Leaving %0 code snippet', language );
+	}
+
+	if ( direction === 'enter' ) {
+		return t( 'Entering code snippet' );
+	}
+
+	return t( 'Leaving code snippet' );
```
oleq, line 289: I think we can safely assume there will never be a code snippet without the language upcasted in the editor. See https://github.com/ckeditor/ckeditor5/blob/84d1ada4e42bbdd3df3f2b4623a77a77db5d863f/packages/ckeditor5-code-block/src/converters.ts#L205-L207. I'd assume that `element.getAttribute( 'language' )!`.

oleq (changes requested):

packages/ckeditor5-code-block/src/utils.ts
Mati365: It can happen if `language` passed to `codeBlock` has no language definition specified in config (like `FooBar`). I'd stay with a bit more verbose but safer approach.

packages/ckeditor5-code-block/src/codeblockediting.ts
Mati365: Fixed

Mati365:
@oleq Can you take a look again?

PR #16064, opened 2024-03-20
Introduced restricted editing integration with the menu bar

DawidKossowski:

Internal (restricted-editing): Introduced restricted editing integration with the menu bar.


packages/ckeditor5-ui/src/menubar/utils.ts
```diff
@@ -467,6 +467,30 @@ export const DefaultMenuBarItems: DeepReadonly<MenuBarConfigObject[ 'items' ]> =
 				items: [
 					'menuBar:showBlocks'
 				]
+			},
+			{
+				groupId: 'restrictedEditing',
+				items: [
+					{
+						menuId: 'restrictedEditing',
+						label: 'Restricted editing',
```
oleq, line 476: I think `DefaultMenuBarItems` should be a helper that accepts `Editor` insstance, so we can use `t()` to localize this one. This is the first time we need that but I think it's not reeleasable unless we address this issue.

oleq (commented):
I'm on a fence with this one

image

because there will never be more items in this menu in this editor mode. Maybe this will make more sense?

```
View
   ...
   🔒 Disable editing
   Restricted editing
       Previous editable region
       Next editable region
```

"Disable/Enable editing"  will be next to other menu items then and IMO it will make more sense 
image

I'm not sure this will stay in "View", just giving you an example.

OTOH, your approach is logical and maybe it's better for users that everything stays in one place. I don't know :D 

WDYT?

--- 

On a side note: we need to add a margin-left to icon-less buttons because they look misaligned when an icon-less menu shows next to them. Let's address this later on.

DawidKossowski:
I had very similar doubts to yours and was also considering something like this:
```
View
   ...
   🔒 Disable editing
      Previous editable region
      Next editable region
```
However, your proposal worked for me, so I pushed the changes.

To avoid blocking this PR, I will address the two additional issues as separate tasks:
1. Add a margin-left to icon-less buttons because they look misaligned when an icon-less menu shows next to them.
2. `DefaultMenuBarItems` should be a helper that accepts Editor instance, so we can use `t()`

Issue #14702, opened 2023-08-01
[Safari] Reverse typing effect after focus change, e.g., outdent button causing text to insert backwards.


david-merritt:
## 📝 Instructions

1. Open Safari
2. Navigate to: https://ckeditor.com/ckeditor-5/demo/feature-rich/
3. Remove all text from CKEditor input
4. Click into the CKEditor input
5. Click the Outdent button
6. Start typing

### ✔️ Expected result
_Text should input normally_

### ❌ Actual result
_The text is input in reverse_

## 📃 Other details
* Browser: Safari
* OS: macOS
* First affected CKEditor version: Latest

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Witoso:
I cannot reproduce it, could you share a video?

https://github.com/ckeditor/ckeditor5/assets/9881379/e71164d5-5385-4672-bb47-16701fb84b45

urbanspr1nter:
@Witoso I believe this may be another case of "backwards" typing. We've also seen this type of behavior in Mac where no performance issue needs to be present for it to reproduce. This may be one of the cases I called out in the bug: https://github.com/ckeditor/ckeditor5/issues/14569#issuecomment-1650106287 where I believed there could be additional cases in this platform (Mac).

Internally, we've had a number of users claim similar things where performance issue has not been present (again, Mac!) but I've yet to actually reproduce it myself. 

From my general understanding, the CMD+TAB action back into the editable will actually begin to insert text backwards (sticky cursor) -- however, based on my testing with @niegowski 's change, I was unable to reproduce it further. I personally don't use Mac often so it may just be me not being thorough enough. 

Could there be another area where things like indent/outdent could affect selection not being updated?

Witoso:
The demos on the page are not updated yet, so there's a chance it will help with this case. @niegowski WDYT about the indent?

niegowski:
This case is different. While clicking on the disabled button (in Safari) it gets focused but the selection is still in the editable:

![2023-08-01 15 22 01](https://github.com/ckeditor/ckeditor5/assets/1232187/b3218162-a52b-453b-b0e5-9263952871ec)

But since editable is not focused, we do not touch the DOM selection so the selection is not updated.

Witoso:
Oh, it's about the outdent :facepalm:

urbanspr1nter:
> This case is different. While clicking on the disabled button (in Safari) it gets focused but the selection is still in the editable:
[...]

PR #16066, opened 2024-03-21
Fix sticky cursor resulting in reverse typing if the editable is not the activeElement in the document.

### LINKED ISSUE #15182: "Reverse typing" effect in some instances where selection may have been lost

Hi, I'm building off of these bugs here, as the problem still persists:
1. https://github.com/ckeditor/ckeditor5/issues/13926#issuecomment-1761548452
2. https://github.com/ckeditor/ckeditor5/issues/14702#issuecomment-1660318476

For some background information, the original issue was first presented and solved in this bug: 
* https://github.com/ckeditor/ckeditor5/issues/14569

This "reverse typing" effect unfortunately has many causes and manifestations which eventually lead to the same effect. The original bug #14569 had intended to solve the issue where selection was not updated in time due to heavy CPU usage while user was typing. This caused some reverse typing as the cursor would "stick" at the current selection, and not update. 

This had been fixed with the PR: #14648 - https://github.com/ckeditor/ckeditor5/pull/14648 

This fix has worked quite well.

Now, we run into issues where if the editor is somehow taken out of focus in a "forceful" manner, typing will result in a reverse order due to the cursor never updating its position.

For example, one can synthetically reproduce this by running this piece of JavaScript code while the cursor is within the editor:

```js
// In 5 seconds, this will cause reverse typing
setTimeout(() => {
  const editable = document.querySelector("div.ck-editor__editable");
  editable.dispatchEvent(new FocusEvent("blur"));

  console.log("Now, try to type!");
}, 5000);
```
backwards

### ✔️ Expected result

Typing should never result in characters being visually placed in reverse order when the editor is not truly in focus.

### ❌ Actual result

Typing results in characters being injected in reversed order.

### ❓ Possible solution

To get around this, what we can do as integrators is actually listen in on the `beforeinput` event on the `ViewDocument`, at the `low` event priority. There's an interesting behavior that happens in this event, upon typing, if we _manually dispatch_ a new `FocusEvent` with `focus` to the editable, the selection is restored. 

For example:

```js
editor.editing.view.document.on('beforeinput', () => {
  const editable = getEditableHelper(editor);
  editable.dispatchEvent(new FocusEvent("focus"));
});
```

Or building off our synthetic repro example as a demonstration:

```js
setTimeout(() => {
  const editable = document.querySelector("div.ck-editor__editable");
  editable.dispatchEvent(new FocusEvent("blur"));

  console.log("Now, try to type!");
  
  // In 3 seconds, there will be normal typing
  setTimeout(() => {
      editable.dispatchEvent(new FocusEvent("focus"));
      console.log("Now, we are typing normally again.");
  }, 3000);
}, 5000);
```

Though this is a known workaround, for this specific scenario, I wonder what happens if some _focus stealing_ of some other DOM element occurs, while the cursor is active in the inline editor will result in? For sure it's going to be reverse typing, but there are times where a specific JS library plays bad and maintains/traps focus no matter what. 

The fact is that there is a cursor in the editable indicating that it is "focused" while it is not which is the tricky aspect to this all. I think if the editor truly does not have focus, it should not have a cursor. That's one scenario solved... maybe???? 🤔

## 📃 Other details
This happens in Chromium based browsers, Safari, etc. 

---

urbanspr1nter:
This PR attempts to fix an issue with cases where "reverse typing" effect still occurs. This is caused by a selection not being updated when an input event occurs. The selection seems to not be updated because the `document.activeElement` in DOM is not the correct element. 

There are cases where the editable visually looks focused, but in reality the active element is some other element as it may have stole focus while the editable was active at some point. This can be caused by a manual dispatch of a `blur` event for example. 

We can actually simulate it like this:
```js
setTimeout(() => { 
	const editable = document.querySelector("div[contenteditable=true]");
	editable.dispatchEvent(new FocusEvent("blur"));

	console.log("BLUR!")
}, 5000);
```

If the editable is currently in focus, but the `blur` happens later, the active element in the `document` is not updated properly, and can end up having a selection which is out of sync with the characters being rendered into the editable. (It just sticks and never moves forward).

An example: 
274956696-3bd80cd7-cf19-4776-ab76-d1a026b4dd5f

Something like this can also be seen in the Safari web browser where a a `button` click can cause the issue to occur:
257527157-b3218162-a52b-453b-b0e5-9263952871ec
[Source](https://github.com/ckeditor/ckeditor5/issues/14702#issuecomment-1660318476)

After fix:
fix

Browsers tested on web (manual):
* Google Chrome
* Safari
* Firefox

Type: (engine): Force the editable to be focused if the active element is not the editable when an input event occurs. Closes #15182, #14702 .

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/view/observer/inputobserver.ts
+++ b/packages/ckeditor5-engine/src/view/observer/inputobserver.ts
@@ -39,6 +39,7 @@ export default class InputObserver extends DomEventObserver<'beforeinput'> {
 		// DOM `activeElement`, but the editable is somehow still in focus.
 		// In other words, the editable visually appears to be in focus, but it is not.
 		if ( !this.view.document.isFocused ) {
+			// The current target is assumed to be the editable
 			domEvent.currentTarget?.dispatchEvent( new FocusEvent( 'focus' ) );
 		}
 		const domTargetRanges = domEvent.getTargetRanges();
--- a/packages/ckeditor5-engine/src/view/observer/inputobserver.ts
+++ b/packages/ckeditor5-engine/src/view/observer/inputobserver.ts
@@ -35,6 +35,12 @@ export default class InputObserver extends DomEventObserver<'beforeinput'> {
 		// @if CK_DEBUG_TYPING // 	);
 		// @if CK_DEBUG_TYPING // }
 
+		// Handles reverse typing cases where the current document is not seen as the
+		// DOM `activeElement`, but the editable is somehow still in focus.
+		// In other words, the editable visually appears to be in focus, but it is not.
+		if ( !this.view.document.isFocused ) {
+			domEvent.currentTarget?.dispatchEvent( new FocusEvent( 'focus' ) );
+		}
 		const domTargetRanges = domEvent.getTargetRanges();
 		const view = this.view;
 		const viewDocument = view.document;
--- a/packages/ckeditor5-engine/tests/view/observer/inputobserver.js
+++ b/packages/ckeditor5-engine/tests/view/observer/inputobserver.js
@@ -59,6 +59,34 @@ describe( 'InputObserver', () => {
 		sinon.assert.calledOnce( beforeInputSpy );
 	} );
 
+	it( 'should not dispatch focus event if the document has focus', () => {
+		viewDocument.isFocused = true;
+
+		const mockDispatchEvent = sinon.spy();
+		fireMockNativeBeforeInput( {
+			inputType: 'foo',
+			currentTarget: {
+				dispatchEvent: mockDispatchEvent
+			}
+		} );
+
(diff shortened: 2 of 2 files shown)


urbanspr1nter:
Thoughts on this?

Witoso:
@urbanspr1nter thanks for the PR, and sorry for the late reply. A bit hectic down here :blush:  @niegowski will soon take a look at a couple of problems regarding typing. We will take a look at this as well!

Witoso:
Hey @urbanspr1nter, we prepared a separate PR, as we are merging all typing improvements PRs to one branch. Thanks for the contribution, we used your solution but placed it in a slightly different place. Check #16267 for details. :)

urbanspr1nter:
Thanks! Of course, I'll always leave it to the experts. :)

PR #16067, opened 2024-03-21
Docs: moving the Support section.

godai78:

Docs: moving the Support section. Closes cksource/ckeditor5-commercial#6102

---

### Additional information


Witoso:
IMHO, the browser compatibility doesn't fit the Set up section, and it's too visible. I think we need to move it back to support, but I'm curious what others think.

godai78:
I'm fine with it.

PR #16070, opened 2024-03-21
Pressing the tab key in restricted editing mode enters a table thats non editable

### LINKED ISSUE #15506: Pressing the `Tab` key in restricted editing mode enters a table that's non-editable

1. Create a document that has some editable areas and a non-editable table
2. Place the caret in the first editable area
3. Keep pressing the `Tab` key

### ✔️ Expected result

The non-editable table should not be focusable when pressing `Tab`. At the very least, it should be possible to exit the table when pressing `Tab` again or when pressing `Shift + Tab`

### ❌ Actual result

Eventually, the caret will enter the table and get stuck there:

https://github.com/ckeditor/ckeditor5/assets/72079603/e0bed056-93ce-4fb5-992d-f34ad0dff70d

## 📃 Other details

* Browser: cross-browser
* OS: cross-OS

---

illia-stv:

Fix (widget): Pressing tab should not focus widget editable. Closes #15506.

---

### Additional information


packages/ckeditor5-restricted-editing/src/restrictededitingmodeediting.ts
```diff
@@ -13,14 +13,15 @@ import {
...
-import type {
-	DocumentSelection,
-	Marker,
-	ModelDeleteContentEvent,
-	ModelPostFixer,
-	Range,
-	SchemaAttributeCheckCallback,
-	SchemaChildCheckCallback
+import {
+	UpcastWriter,
```
arkflpc, line 17: This is unused in this file.

packages/ckeditor5-widget/src/utils.ts
```diff
@@ -287,6 +287,7 @@ export function toWidgetEditable(
 	writer.addClass( [ 'ck-editor__editable', 'ck-editor__nested-editable' ], editable );
 
 	writer.setAttribute( 'role', 'textbox', editable );
+	writer.setAttribute( 'tabindex', '-1', editable );
```
arkflpc, line 290: How would it behave if the whole editor has other `tabindex` set by integrators? For instance `tablindex=-2`. Was it tested (in manual test or otherwise)?

arkflpc (changes requested):

packages/ckeditor5-widget/src/utils.ts
niegowski: From MDN: 

> A negative value (the exact negative value doesn't actually matter, usually tabindex="-1") means that the element is not reachable via sequential keyboard navigation.
> Note: tabindex="-1" may be useful for elements that should not be navigated to directly using the Tab key, but need to have keyboard focus set to them. Examples include an off-screen modal window that should be focused when it comes into view, or a form submission error message that should be immediately focused when an errant form is submitted.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

niegowski (approved):

PR #16082, opened 2024-03-22
Docs: moving Tutorials around.

godai78:

Docs: moving Tutorials around. Closes cksource/ckeditor5-commercial#6113

---

### Additional information

### DIFF
--- a/docs/getting-started/integrations/angular.md
+++ b/docs/getting-started/integrations/angular.md
@@ -24,7 +24,7 @@ Currently, the CKEditor&nbsp;5 component for Angular supports integrating CKEdit
 </info-box>
 
 <info-box hint>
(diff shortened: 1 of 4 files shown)


gorzelinski (commented):
Everything works fine, but I wonder if we don't want to nest the tutorial folder into the framework folder. It would match the URL structure better.

godai78:
> Everything works fine, but I wonder if we don't want to nest the tutorial folder into the framework folder. It would match the URL structure better.

Not sure about it, actually. I believe we could even unnest some parts of this sections rather.

pszczesniak (approved):

godai78:
Follow-up: cksource/ckeditor5-commercial#6120

PR #16087, opened 2024-03-23
 Introduced CKBox integration with the menu bar

DawidKossowski:

Internal (ckbox): Introduced CKBox integration with the menu bar.

### DIFF
--- a/packages/ckeditor5-ckbox/src/ckboxui.ts
+++ b/packages/ckeditor5-ckbox/src/ckboxui.ts
@@ -8,11 +8,9 @@
  */
 
 import { icons, Plugin } from 'ckeditor5/src/core.js';
-import { ButtonView } from 'ckeditor5/src/ui.js';
+import { ButtonView, MenuBarMenuListItemButtonView } from 'ckeditor5/src/ui.js';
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-ckbox/src/ckboxui.ts
```diff
@@ -97,4 +86,28 @@ export default class CKBoxUI extends Plugin {
...
+	 */
+	private _createButton<T extends typeof ButtonView | typeof MenuBarMenuListItemButtonView>( ButtonClass: T ): InstanceType<T> {
+		const editor = this.editor;
+		const locale = editor.locale;
+		const view = new ButtonClass( locale ) as InstanceType<T>;
+		const command = editor.commands.get( 'ckbox' )!;
+		const t = locale.t;
+
+		view.set( {
+			label: t( 'Open file manager' ),
```
DawidKossowski, line 101: I'm wondering if we should unify the labels with `CKFinder`.

oleq: That sounds like a task for another day 🙂

PR #16100, opened 2024-03-26
Docs: rearranging plugin guides.

godai78:

Docs: rearranging plugin guides. Closes cksource/ckeditor5-commercial#6120

---

### Additional information

_Merge with https://github.com/cksource/ckeditor5-commercial/pull/6122_


docs/updating/update-to-33.md
gorzelinski, line 1: Do we want to change links retrospectively?

gorzelinski (commented):
Everything seems to work fine - no errors in links. I just added one question.

docs/updating/update-to-33.md
godai78: I'll check this one. Most of the times we should not.

godai78: We have to, in this case, there is no legacy version of this guide. Or we may delink it.

PR #16104, opened 2024-03-27
Add resize table column balloon editor

Mati365:

Feature (table): Add "Resize Column" option in the table editor toolbar, offering an alternative to drag-and-drop resizing.

---

### Additional information

Closes https://github.com/cksource/ckeditor5-commercial/issues/6023
Part of https://github.com/cksource/ckeditor5-commercial/pull/6133
Related to https://github.com/ckeditor/ckeditor5/pull/16140

Screens:
image
image

## Summary by CodeRabbit

- **New Features**
  - Added functionality for resizing table columns in the CKEditor interface.
  - New UI components for adjusting table column widths, including labeled input fields and buttons.
  - Enhanced form validation and handling for table column adjustments.

- **Bug Fixes**
  - Improved precision in calculating table cell widths using `parseFloat`.

- **Refactor**
  - Updated module paths and reorganized code for better clarity and separation of concerns in table column resizing features.

- **Style**
  - Introduced new styles for table column resizing forms.
  - Updated table cell border color to use CSS variables for better theme consistency.

- **Documentation**
  - Enhanced user interface documentation related to new table column resizing features.

### DIFF
--- a/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeutils.ts
+++ b/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeutils.ts
@@ -1,0 +1,561 @@
+/**
+ * @license Copyright (c) 2003-2024, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module table/tablecolumnresize/tablecolumnresizeutils
+ */
+
+import { isEqual } from 'lodash-es';
+
+import { Plugin, type Editor } from 'ckeditor5/src/core.js';
+
+import type { DowncastWriter, Element, ViewElement, ViewNode } from 'ckeditor5/src/engine.js';
+import TableUtils from '../tableutils.js';
+
+import { COLUMN_MIN_WIDTH_IN_PIXELS } from './constants.js';
+import {
+	clamp, getColumnEdgesIndexes, getColumnGroupElement, getDomCellOuterWidth,
+	getElementWidthInPixels,
+	getTableColumnsWidths, getTableWidthInPixels, sumArray, toPrecision
+} from './utils.js';
+
+import TableWalker from '../tablewalker.js';
+
+/**
+ * The table column resize utils plugin.
+ */
+export default class TableColumnResizeUtils extends Plugin {
+	/**
+	 * @inheritDoc
+	 */
+	public static get requires() {
+		return [ TableUtils ] as const;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public static get pluginName() {
+		return 'TableColumnResizeUtils' as const;
+	}
+
+	/**
+	 * Performs resize of specified by resizer column with specified width.
+	 * It composes steps from {@link #prepareColumnResize}, {@link #assignColumnWidth} and {@link #endResize}.
+	 *
+	 * @param resizerElement Resize column handle view element.
+	 * @param newColumnWidth New column width in pixels.
+	 * @param compensateWidthWhenMoveCenteredTable
+	 */
+	public resizeColumnUsingResizer(
+		resizerElement: ViewElement,
+		newColumnWidth: number,
+		compensateWidthWhenMoveCenteredTable?: boolean
+	): void {
+		const resizingData = this.prepareColumnResize( resizerElement );
+
+		if ( resizingData ) {
+			this.assignColumnWidth( resizingData, newColumnWidth, compensateWidthWhenMoveCenteredTable );
+			this.endResize( resizingData );
+		}
+	}
+
+	/**
+	 * In this scenario:
+	 *
+	 *  +---+---+---+
+	 *  | a         |
+	 *  +---+---+---+
+	 *  | b | c | d |
+	 *  +---+---+---+
+	 *  | e | f | g |
+	 *  +---+---+---+
+	 *
+	 * When user selects column that contains `a`, `b`, `e` cells this function returns the first smallest
+	 * column resize view element in that selection (`b` column). The resize view element is the blue draggable
+	 * border element on the right side of selected column.
+	 *
+	 * @returns View element of column resizer DOM node.
+	 */
+	public getSmallestSelectedColumnResizer(): ViewElement | null {
+		const sortedElements = getAllSelectedResizersDOMNodes( this.editor )
+			.sort( ( a, b ) => Math.sign( a.getBoundingClientRect().left - b.getBoundingClientRect().left ) );
+
+		if ( !sortedElements.length ) {
+			return null;
+		}
+
+		const viewElement = this.editor.editing.view.domConverter.domToView( sortedElements[ 0 ] )!;
+
+		return viewElement as ViewElement;
+	}
+
+	/**
+	 * Starts resizing process.
+	 *
+	 *  * assigns resize attributes that will be used to resize table in next steps,
+	 *  * calculates resizing data based on resizer element,
+	 *  * applies the attributes to the `<table>` view element.
+	 *
+	 * @param resizerElement Resize column handle view element.
+	 */
+	public prepareColumnResize( resizerElement: ViewElement ): ResizingData | null {
+		const { editing } = this.editor;
+
+		const viewTable = resizerElement.findAncestor( 'table' )!;
+		const editingView = editing.view;
+		const columnWidthsInPx = calculateResizerColumnWidth( this.editor, resizerElement );
+
+		if ( !columnWidthsInPx ) {
+			return null;
+		}
+
+		// Insert colgroup for the table that is resized for the first time.
+		if ( !Array.from( viewTable.getChildren() ).find( viewCol => viewCol.is( 'element', 'colgroup' ) ) ) {
+			editingView.change( viewWriter => {
+				insertColgroupElement( viewWriter, columnWidthsInPx, viewTable );
+			} );
+		}
+
+		const resizingData = getResizingData( this.editor, resizerElement, columnWidthsInPx );
+
+		// At this point we change only the editor view - we don't want other users to see our changes yet,
+		// so we can't apply them in the model.
+		this.editor.editing.view.change( writer => applyResizingAttributesToTable( writer, viewTable, resizingData ) );
+
+		return resizingData;
+	}
+
+	/**
+	 * This function returns the maximum and minimum dimensions to which a table column can be expanded.
+	 * It also provides the current dimension of the column.
+	 *
+	 * @param resizingDataOrElement Resizing data or view element.
+	 */
+	public getPossibleResizeColumnRange( resizingDataOrElement: ResizingData | ViewElement ): PossibleResizeColumnRange | null {
+		const resizingData = ( () => {
+			if ( 'startDraggingPosition' in resizingDataOrElement ) {
+				return resizingDataOrElement;
+			}
+
+			const columnWidthsInPx = calculateResizerColumnWidth( this.editor, resizingDataOrElement );
+
+			if ( !columnWidthsInPx ) {
+				return null;
+			}
+
+			return getResizingData( this.editor, resizingDataOrElement, columnWidthsInPx );
+		} )();
+
+		if ( !resizingData ) {
+			return null;
+		}
+
+		const {
+			flags: {
+				isRightEdge
+			},
+			widths: {
+				viewFigureParentWidth,
+				tableWidth,
+				leftColumnWidth,
+				rightColumnWidth
+			}
+		} = resizingData;
+
+		const dxUpperBound = isRightEdge ?
+			viewFigureParentWidth - tableWidth :
+			rightColumnWidth! - COLUMN_MIN_WIDTH_IN_PIXELS;
+
+		return {
+			current: leftColumnWidth,
+			lower: COLUMN_MIN_WIDTH_IN_PIXELS,
+			upper: leftColumnWidth + Math.max( 0, dxUpperBound )
+		};
+	}
+
+	/**
+	 * Update column width using provided resizing data.
+	 *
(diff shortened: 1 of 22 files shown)


charlttsie:
## Table size balloon is left out after removing the table
### Steps
1. Open e.g. http://localhost:8125/ckeditor5/tests/manual/all-features.html
2. Insert a table
3. Select the “Resize column” option from the table toolbar
4. Press Cmd + A to remove the content
### Result
The balloon with table size is not removed:

https://github.com/ckeditor/ckeditor5/assets/132071257/e2311bca-0705-48bd-83ca-f48537ca6a45

charlttsie:
https://github.com/ckeditor/ckeditor5/pull/16104#issuecomment-2031549361 and https://github.com/ckeditor/ckeditor5/pull/16104#issuecomment-2031551359 look fixed 👍

charlttsie:
## It's not possible to resize a column right after table insertion
### Steps
1. Open e.g. http://localhost:8125/ckeditor5/tests/manual/all-features.html (or any other manual test with table resize)
2. Insert a table
3. Look for the “Resize column” option in the table toolbar

### Result
The option is greyed out. Resizing the table using resize handler activates the option:

https://github.com/ckeditor/ckeditor5/assets/132071257/42b2efda-922f-4208-a5ce-280d2585d5e9

charlttsie:
## Table column width changes when the column is selected
### Steps
1. Open e.g. http://localhost:8125/ckeditor5/tests/manual/all-features.html
2. Insert a table 
3. Select a table column (either with the toolbar option or manually)
4. Deselect the column

### Result
The column gets narrower when selected:

https://github.com/ckeditor/ckeditor5/assets/132071257/8b37575d-90ed-4b7c-894a-bc4e92ebba28

### Note
It's a regression - not reproducible on master

kubaklatt:
## The error when entering an incorrect number is misleading

### The error suggests that it is possible to enter a number with a comma/dot (depends on the browser) - but this is not possible.

### Steps:
1. Open e.g. [all-features-rtc.html](http://localhost:8125/ckeditor5/external/ckeditor5-commercial/packages/ckeditor5-collaboration/tests/manual/all-features-rtc.html)
2. Try entering a high number to trigger an error that shows the maximum possible value
3. Enter suggested value

### Result:

An error appears that the entered value is invalid

https://github.com/ckeditor/ckeditor5/assets/90427796/cc21ce9e-942b-4de6-9adb-3eb6994772ec

### Notes:

When entering an incorrect value and switching focus from the input to something else, the input header overlaps with the incorrect value

https://github.com/ckeditor/ckeditor5/assets/90427796/b5bb8b18-7c2f-4008-9d7e-f15313478205

Mati365:
@kubaklatt / @charlttsie All reported issues should be fixed. Please switch to `ck/6023` on commercial repo.

charlttsie:
We've finished testing with @kubaklatt and @marcelmroz and the PR looks good. We haven't found anything new, while the previously reported issues don't reproduce anymore. Thanks @Mati365 for the quick fixes! 🎉

Dumluregn (changes requested):
1. Changing the size of the furthest right column sets different result:

https://github.com/ckeditor/ckeditor5/assets/10883184/01a54c4a-3b73-4b5a-9681-da0e185cb230

2. The "Resize column" button is always visible in the dropdown, even if the table column resize feature is not loaded in the editor.

image

3. I'd try to implement a live preview when changing the column widths. Without it user has to guess a few times until they find the expected size. It should work like the other table properties, so apply only view changes until the change is applied and after committing the new value, propagate change to the model - this way we only have one undo step.

Mati365:
@Dumluregn 
1. Fixed
2. Fixed
3. I added validation with proper messages instead (so it's normalized to common behavior of other such forms according to other tickets related to VPAT). I believe that solution that you described is better in terms of UX, but it's also much harder to implement knowing that there will be a lot of edge cases with incorrect positioning of the balloon. Let's keep this in mind in further UX improvements.

packages/ckeditor5-table/src/tablecolumnresize/commands/tableresizecolumncommand.ts
```diff
@@ -0,0 +1,101 @@
...
+			const { last, first } = tableUtils.getColumnIndexes(
+				tableUtils.getSelectionAffectedTableCells( this.editor.model.document.selection )
+			);
+
+			if ( last === first ) {
+				this.isEnabled = true;
+				this.possibleRange = resizeUtils._getPossibleResizeColumnRange( smallestResizer );
+				return;
+			}
+		}
```
Dumluregn, line 70: This part requires a couple comments answering questions like: "why the heck we get the smallest resizer?" etc.

packages/ckeditor5-table/src/tablecolumnresize/utils/getsmallestcolumnresizer.ts
```diff
@@ -0,0 +1,59 @@
...
+ *  | e | f | g |
+ *  +---+---+---+
+ *
+ * When user selects column that contains `a`, `b`, `e` cells this function returns the first smallest
+ * column resize view element in that selection (`b` column).
+ *
+ * @param editor Editor instance.
+ * @returns View element of column resizer DOM node.
+ */
+export function _getSmallestColumnResizer( editor: Editor ): ViewElement | null {
```
Dumluregn, line 30: It's only used in the command, so it may as well be a local function defined there. Or if you think it may be useful somewhere else, we have `utils.ts` or even `TableColumnResizeUtils` plugin now.

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeutils.ts
```diff
@@ -0,0 +1,517 @@
...
+	getElementWidthInPixels,
+	getTableColumnsWidths, getTableWidthInPixels, sumArray, toPrecision
+} from './utils.js';
+
+import TableWalker from '../tablewalker.js';
+
+/**
+ * The table column resize editing plugin.
+ */
+export default class TableColumnResizeUtils extends Plugin {
```
Dumluregn, line 29: I think this needs a bit of a cleanup.

There's already a `utils.ts` file with helpers dedicated for column resize feature. At the first glance, I don't see how functions added as methods of `TableColumnResizeUtils` plugin are different than those helpers in terms of semantics. So either all of them should be just helper functions not collected in a plugin, or all of them be kept in plugin. The situation where there's no clear distinction will raise confusion. Unless of course I don't see something, please educate me then 😄 

If you go with a plugin, please make it similar to `TableUtils` one - a real collection of utils and nothing more, so all the methods should be public and independent. Helpers for helpers can be kept outside of the plugin.

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeutils.ts
```diff
@@ -0,0 +1,517 @@
...
+	/**
+	 * Performs resize of specified by resizer column with specified width.
+	 * It composes steps from {@link #_prepareColumnResize}, {@link #_assignColumnWidth} and {@link #_endResize}.
+	 *
+	 * @internal
+	 * @param resizerElement Resize column handle view element.
+	 * @param newColumnWidth New column width in pixels.
+	 * @param compensateWidthWhenMoveCenteredTable
+	 */
+	public _resizeColumnUsingResizer(
```
Dumluregn, line 53: `@internal` flag and `_` at the name's start are not necessary since this meant to be a public helper. The same applies to all the other methods in the plugin.

Dumluregn (changes requested):
Overall, it's really cool it's all working and that you did quite a lot of cleaning in the `TableColumnResizeEditing` plugin 👏🏻 

Besides some necessary cleanup that I wrote about below, I noticed two more issues:

1. When a selected cell spans through a few columns, the width of the last one is shown in the form and resized if you change the width. I think it should be the first one.

https://github.com/ckeditor/ckeditor5/assets/10883184/f14bab0e-b317-491f-8df4-4a2f40c34e2b

2. A problem with undo. If you change the width twice and undo once, then open the resize form, then you'll see the wrong value.

https://github.com/ckeditor/ckeditor5/assets/10883184/49e19079-7a15-4fa8-91ad-fc5f1fd9867a

packages/ckeditor5-table/src/tablecolumnresize/utils/getsmallestcolumnresizer.ts
Mati365: Moved to `TableColumnResizeUtils`.

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeutils.ts
Mati365: @Dumluregn I used the same naming convention as it is in the core and many other places:
https://github.com/ckeditor/ckeditor5/blob/ck/6023/packages/ckeditor5-engine/src/model/element.ts#L219-L252 
https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-engine/src/model/position.ts#L877-L901

Are you sure?

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeutils.ts
Mati365: I kept `TableColumnResizeUtils` and moved all private methods outside this plugin.

packages/ckeditor5-table/src/tablecolumnresize/commands/tableresizecolumncommand.ts
Mati365: Adjusted docs in `_getSmallestSelectedColumnResizer` util method. `resizer` term is used in many other places in this particular plugin, so imo it does not need further explanation.

packages/ckeditor5-table/src/tablecolumnresize/commands/tableresizecolumncommand.ts
```diff
@@ -0,0 +1,101 @@
...
+ * ```
+ */
+export default class TableResizeColumnCommand extends Command {
+	/**
+	 * The command value: Current size of column (in pixels) and possible resize range of such column..
+	 *
+	 * @readonly
+	 * @observable
+	 */
+	declare public possibleRange: PossibleResizeColumnRange | null;
```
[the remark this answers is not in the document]
Mati365: It can be, but it simplifies a lot of the code due to the fact that the currently selected column width is calculated alongside with range, and knowing the fact that the calculation of column widths is quite expensive it's good enough to stay as it is.

Mati365:
@Dumluregn 
1. Actually, it's correct behavior according to column resize implementation. It's because when you start to resize the right corner of such colspan you actually resize the top row column element with much smaller width. It's counterintuitive, but at this moment it's not possible to change this behavior without breaking a lot of other edge cases. 
2. Fixed. Table resize operations seem to trigger `refresh()`  only on selection change (which occurs quite often in mouse operations). I added `refresh` call before loading initial values to form. Seems to be working correctly.

Rest of CR remarks should be fixed. Can you check again?

packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeutils.ts
Mati365: @Dumluregn done

Mati365:
No changes, rebased.

charlttsie:
## Maximum column size allowed value is 1px higher from the size that can be set
### Steps
1. Open e.g. http://localhost:8125/ckeditor5-table/tests/manual/tablecolumnresize.html
2. In the table column resize balloon, insert the highest possible value and save
3. Open table column resize balloon once again 

### Result
The allowed value was 1px higher from the actual size that's been set:

https://github.com/ckeditor/ckeditor5/assets/132071257/aa80f5bd-5833-41e1-acf5-e36509d55c96

charlttsie:
## For a single-column table, minimum column size allowed is 1px higher from the size that can be set
### Steps
1. Open e.g. http://localhost:8125/ckeditor5-table/tests/manual/tablecolumnresize.html
2. Insert a table that has one column and at least 2 rows
2. In the table column resize balloon, insert the lowest possible value and save
3. Open table column resize balloon once again

### Result
The column used to have 40px but now has 39px:

https://github.com/ckeditor/ckeditor5/assets/132071257/4408ed94-ea32-4358-9356-1ca36726ffa3

charlttsie:
## For a single-column table, the value is sometimes set to 1px lower from what's been inserted
Seems related to: https://github.com/ckeditor/ckeditor5/pull/16104#issuecomment-2063933056, but this scenario doesn't involve the validation issue
### Steps
1. Open e.g. http://localhost:8125/ckeditor5-table/tests/manual/tablecolumnresize.html
2. Insert a table that has one column and at least 2 rows
3. Open table column resize balloon and change the column size

### Result
Sometimes, but not always (see 460px at 00:12 in the screencast), the value after reopening the resize balloon differs from what's been inserted:

https://github.com/ckeditor/ckeditor5/assets/132071257/25077fff-0509-4834-a200-74c662f7f494

charlttsie:
FYI: https://github.com/ckeditor/ckeditor5/issues/16251 - Clicking on a resize handle in a nested table extends the tables to the full editor width

The above issue also reproduces when resizing using the balloon:

https://github.com/ckeditor/ckeditor5/assets/132071257/bc1b1dd7-7781-4bc5-9d15-b319a990fc41

charlttsie:
@Mati365 
We've finished this round of tests with @marcelmroz and @kubaklatt and have found a few issue, which are reported above. Please let us know when ready for a retest, and will take another look 👀

Mati365:
Blocked, there were discussions about that in the past with @scofalik / @DawidKossowski / @oleq. From what I remember, it should be discussed on architecture guild.

PR #16105, opened 2024-03-27
Change configuration handling to make menu bar opt-in in classic editor.

mremiszewski:

Internal (ui,editor-classic): Streamline menu bar normalization. Made the feature an opt-in in the `ClassicEditor`. Related to https://github.com/ckeditor/ckeditor5/issues/15894.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-editor-classic/src/classiceditor.ts
+++ b/packages/ckeditor5-editor-classic/src/classiceditor.ts
@@ -74,6 +74,10 @@ export default class ClassicEditor extends ElementApiMixin( Editor ) {
 
 		super( config );
 
+		this.config.define( 'menuBar', {
+			isVisible: false
+		} );
+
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-editor-classic/src/classiceditor.ts
```diff
@@ -85,8 +85,10 @@ export default class ClassicEditor extends ElementApiMixin( Editor ) {
 		this.model.document.createRoot();
 
 		const shouldToolbarGroupWhenFull = !this.config.get( 'toolbar.shouldNotGroupWhenFull' );
+		const useMenuBar = !!this.config.get( 'menuBar' );
```
oleq, line 88: The problem with this is that we're going to use `config.define( 'menuBar', ... )` for the default menu bar configuration soon. This will always be true. Either we don't use `config.define()` or we have to come up with another config.

Issue #15502, opened 2023-12-11
🦸 Improving CKEditor 5 installation methods


Witoso:
_Imagine you are setting up this cool editor for your app. You copied snippets from the docs, installed the packages, and quickly embedded the editor in your app. Everything works, you can type, and play with it. But now you start thinking about your specific use case. There’s just this one plugin that you want to add. You do it… and all hell breaks loose_ 🔥🔥🔥 _Wall of errors, misleading docs. “I copied it straight from the docs! What’s wrong?!_ 😖 _” bangs in your head._

This is just one scenario of misleading onboarding to CKEditor 5. Which, you ask? The one where you extend a predefined build with additional plugins. :face\_exhaling:

Over the years, we learned a lot about the issues that you are facing. We resolved some and worked around some others, but the core issues remained there. We feel that it's now the right time to get to the root of it. 💪

For the past few months, we have researched the topic even further, focusing on these key issues:

*   We have **several methods of installation**. Each has its specific quirks that make using the documentation difficult.
*   The source packages that we release to npm need **a specific setup**, like webpack scripts, or a Vite plugin.
*   Certain popular environments require **various workarounds** to set up CKEditor 5 (e.g. server-side rendered (SSR) applications like Next.js, Nuxt.js, or SvelteKit).

The ultimate goal is to make **setting up of CKEditor a pleasure, not a nightmare.** It should feel like adding another library to your application — nothing more, nothing less. And we came up with a plan for how to do that.

**Now, we need your help**!

Share your experiences, ask us questions, correct our assumptions, and show your unique use cases. Finally, let us know if you like the proposed direction. 🙏

**The RFC with the details is in the first comment below**. Join us on this exciting ride!🎢

filipsobol:
# Setup methods RFC

Hello, CKEditor 5 Community! 👋

As we explained in the first comment, our current setup methods are not optimal. In this RFC, we will dive deeper and expand on our thought process and proposal.

## Problems 😓

WYSIWYG editors are one of the most complex frontend projects and this affects current CKEditor 5 installation methods. Moreover, we have added some complexity due to the lack of standardization in the JavaScript ecosystem when we created this project, browser limitations, and most importantly, technical debt.

We feel it is time to remove some of that complexity.

### The build step

Currently, the editor code requires a special build step to resolve non-standard CSS and SVG imports and to automatically add required theme styles and translations. Also, the editor itself does not allow dynamic plugin registration.

[...]

PR #16107, opened 2024-03-27
Generate bundles for new installation methods

filipsobol:

Feature: Add bundles for new installation methods. See #15502.

Fix: Change various exports of types and interfaces to type-only exports.

Fix (utils): Allow `Translations.getPluralForm` type to be `null`.

---

### Additional information


packages/ckeditor5-ui/src/index.ts
```diff
@@ -17,7 +20,7 @@ export { default as AccessibilityHelp } from './editorui/accessibilityhelp/acces
 
 export { default as BodyCollection } from './editorui/bodycollection.js';
 
-export { default as Button, type ButtonExecuteEvent } from './button/button.js';
+export type { default as Button, ButtonExecuteEvent } from './button/button.js';
```
arkflpc, line 23: Why? Isn't it breaking change?

packages/ckeditor5-utils/src/locale.ts
```diff
@@ -201,6 +201,6 @@ function interpolateString( string: string, values: ReadonlyArray<any> ): string
 export type Translations = {
 	[ language: string ]: {
 		dictionary: { [ messageId: string ]: string | ReadonlyArray<string> };
-		getPluralForm?: ( n: number ) => number;
+		getPluralForm: ( ( n: number ) => number | boolean ) | null;
```
arkflpc, line 204: Where is it used that way?

arkflpc (commented):
LGTM,
You can merge if you're certain that we want to demote some exports from value exports to type-only ones.

filipsobol:
@coderabbitai review

packages/ckeditor5-ui/src/index.ts
filipsobol: No, because `Button` is an interface and `ButtonExecuteEvent` is a type.

packages/ckeditor5-utils/src/locale.ts
filipsobol: The English translation generated by the new CLI contains `getPluralForm: null`. The same translation generated by the old tool skips this field entirely, so I reverted the `?` symbol.

filipsobol:
> LGTM, Are you certain that we want to demote some exports from value exports to type-only ones?

Yes, I added the `type` only to the exports of types or interfaces.

package.json
```diff
@@ -203,7 +204,9 @@
     "clean-up-svg-icons": "node ./scripts/clean-up-svg-icons.js",
     "collect-svg-icons": "node scripts/collect-svg-icons.js",
     "check-dependencies": "ckeditor5-dev-dependency-checker",
-    "check-dependencies:versions-match": "node ./scripts/ci/check-dependencies-versions-match.js"
+    "check-dependencies:versions-match": "node ./scripts/ci/check-dependencies-versions-match.js",
+    "check-theme-lark-imports": "node ./scripts/check-theme-lark-imports.js",
+    "build:dist": "node ./scripts/build-ckeditor5.mjs"
   },
   "lint-staged": {
     "**/*.{js,ts}": [
```
psmyrek, line 212: Since we have `*.mjs` files now, it would be good to add them to the `lint-staged`.

filipsobol: Done :heavy_check_mark:

psmyrek (approved):
LGTM 🎉

package.json
pszczesniak, line 1: Shouldn't we add `@ckeditor/ckeditor5-premium-features` to `devDependencies`?

package.json
```diff
@@ -90,9 +90,10 @@
     "@ckeditor/ckeditor5-ai": "41.3.1",
     "@ckeditor/ckeditor5-case-change": "41.3.1",
     "@ckeditor/ckeditor5-comments": "41.3.1",
+    "@ckeditor/ckeditor5-dev-build-tools": "^39.7.0-alpha.3",
     "@ckeditor/ckeditor5-dev-bump-year": "^39.0.0",
     "@ckeditor/ckeditor5-dev-ci": "^39.0.0",
-    "@ckeditor/ckeditor5-dev-dependency-checker": "^39.0.0",
+    "@ckeditor/ckeditor5-dev-dependency-checker": "^39.7.0-alpha.3",
```
pomek, line 96: This package depends on `@ckeditor/ckeditor5-dev-utils`, which has been installed twice now. Let's consider releasing ckeditor5-dev as stable, too, even if it means adding minor changes to the `build-tools` package.

package.json
```diff
@@ -203,10 +204,12 @@
     "clean-up-svg-icons": "node ./scripts/clean-up-svg-icons.js",
     "collect-svg-icons": "node scripts/collect-svg-icons.js",
     "check-dependencies": "ckeditor5-dev-dependency-checker",
-    "check-dependencies:versions-match": "node ./scripts/ci/check-dependencies-versions-match.js"
+    "check-dependencies:versions-match": "node ./scripts/ci/check-dependencies-versions-match.js",
+    "check-theme-lark-imports": "node ./scripts/check-theme-lark-imports.js",
+    "build:dist": "node ./scripts/build-ckeditor5.mjs"
```
pomek, line 209: Could we move the `build:dist` task below `build`?

scripts/build-package.mjs
```diff
@@ -0,0 +1,35 @@
...
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/* eslint-env node */
+
+import { createRequire } from 'module';
+import upath from 'upath';
+import { build } from '@ckeditor/ckeditor5-dev-build-tools';
+
+( async () => {
```
pomek, line 14: No need for IIFE and `async`. Top-level await is supported since Node 14.

scripts/build-package.mjs
```diff
@@ -0,0 +1,35 @@
...
+	const require = createRequire( import.meta.url );
+	const pkg = require( upath.resolve( process.cwd(), './package.json' ) );
+
+	await build( {
+		input: 'src/index.ts',
+		output: upath.resolve( process.cwd(), './dist/index.js' ),
```
pomek, line 20: TBH, I do not like that we depend on `process.cwd()`. I would see it as a CLI script - `bash build-package --path ...`.

Is it possible?

src/index.ts
pomek, line 1: We need a CI to check if `ckeditor5` really contains all exports from `packages/*`. We will forget to add a new package.

packages/ckeditor5-adapter-ckfinder/tsconfig.dist.json
pomek, line 1: Taking the occasion, couldn't we group the `tsconfig.*` files inside a directory?

scripts/build-ckeditor5.mjs
```diff
@@ -0,0 +1,85 @@
...
+
+/* eslint-env node */
+
+import { rm, copyFile } from 'fs/promises';
+import upath from 'upath';
+import chalk from 'chalk';
+import { build } from '@ckeditor/ckeditor5-dev-build-tools';
+
+function dist( path ) {
+	return upath.join( process.cwd(), 'dist', path );
```
pomek, line 16: Similar to `scripts/banner.mjs`. Could we avoid `process.cwd()` and instead, calculate a path to the root directory manually?

scripts/build-ckeditor5.mjs
```diff
@@ -0,0 +1,85 @@
...
+		output: dist( 'browser/index.js' ),
+		tsconfig,
+		banner,
+		sourceMap: true,
+		minify: true,
+		browser: true,
+		name: 'ckeditor5',
+		external: []
+	} );
+} )();
```
pomek, line 85: Also, no need for IIFE.

pomek (commented):
My two cents, guys. WDYT?

pomek:
> ci/circleci: release_prepare — Your tests failed on CircleCI

```
Errors:
1. /dist/browser/index.umd.js is written in CJS, but is interpreted as ESM. Consider using the .cjs extension, e.g. /dist/browser/index.umd.cjs
```

pomek:
Is content styles generator and building docs work as expected?

scripts/build-package.mjs
filipsobol: I used IIFE, because eslint is not happy about it top-level await

filipsobol: Top-level await support was added in eslint 8 with `ecmaVersion: 2022` or higher. We are currently using eslint 7 and `ecmaVersion: 2020` and upgrading would require a lot of work that we currently don't have the capacity for.

Source: https://eslint.org/blog/2021/06/whats-coming-in-eslint-8.0.0/#support-for-es2022-class-fields%2C-top-level-await%2C-and-regexp-matching-indices

src/index.ts
filipsobol: It's on our roadmap and will be added as a follow-up.

scripts/build-package.mjs
filipsobol: This script is called from each package. This means that in each package we would have to specify the path to it, for example

```bash
# packages/ckeditor5-adapter-ckfinder
node ../../scripts/build-package.mjs --path=packages/ckeditor5-adapter-ckfinder

# packages/ckeditor5-alignment
node ../../scripts/build-package.mjs --path=packages/ckeditor5-alignment

# packages/ckeditor5-autoformat
node ../../scripts/build-package.mjs --path=packages/ckeditor5-autoformat

# etc...
```

This path could be `.join`-ed with one of the `constants'.

IMO this would create unnecessary noise.

scripts/build-package.mjs
pszczesniak, line 1: Would be good to have some information that the build was ended successfully, WDYT? :thinking:

packages/ckeditor5-adapter-ckfinder/tsconfig.dist.json
filipsobol: This would add a lot of unnecessary noise to this already big PR, as we would need to align configuration of few tools.

package.json
filipsobol: This causes CI to fail, because there hasn't been a stable release of `@ckeditor/ckeditor5-premium-features` yet.

pomek: `@ckeditor/ckeditor5-premium-feature` or `ckeditor5-premium-feature`?

scripts/build-package.mjs
pomek: Please report this as a new issue. We should handle it soon.

scripts/build-package.mjs
pomek: While I do not like depending on `process.cwd()`, I am convinced to leave it as it is.

src/index.ts
pomek: Create an issue for it, please.

scripts/build-package.mjs
filipsobol: https://github.com/ckeditor/ckeditor5/issues/16258

src/index.ts
filipsobol: It's in our internal roadmap in Notion.

package.json
```diff
@@ -203,10 +206,11 @@
     "clean-up-svg-icons": "node ./scripts/clean-up-svg-icons.js",
     "collect-svg-icons": "node scripts/collect-svg-icons.js",
     "check-dependencies": "ckeditor5-dev-dependency-checker",
-    "check-dependencies:versions-match": "node ./scripts/ci/check-dependencies-versions-match.js"
+    "check-dependencies:versions-match": "node ./scripts/ci/check-dependencies-versions-match.js",
+    "check-theme-lark-imports": "node ./scripts/check-theme-lark-imports.js"
   },
   "lint-staged": {
-    "**/*.{js,ts}": [
+    "**/*.{js,ts,mjs}": [
```
filipsobol, line 213: Add `dist/**` and `packages/*/dist/**` to `eslintIgnore`

pszczesniak (approved):

packages/ckeditor5-engine/src/model/node.ts
```diff
@@ -17,9 +17,6 @@ import type Element from './element.js';
 
 import { CKEditorError, compareArrays, toMap } from '@ckeditor/ckeditor5-utils';
 
-// To check if component is loaded more than once.
-import '@ckeditor/ckeditor5-utils/src/version.js';
```
niegowski, line 21: Is it safe to remove this check for old installation methods? Do we have any other safety mechanism for version mismatch?

filipsobol: According to my tests it is safe to remove.

I tested it in Vite and Webpack with two different versions of the `engine` package. In both, I removed all these side-effect imports, and the `duplicated-module` error was still thrown. It worked whether I imported the code from the main package entry points (`import { Position } from '@ckeditor/ckeditor5-engine';`) or directly from the `src' folders (`import Position from '@ckeditor/ckeditor5-engine/src/model/position.js';`).

This is because any import from `@ckeditor/ckeditor5-utils` triggers this check, and we do it all over this package.

niegowski (approved):

PR #16118, opened 2024-03-28
Menu bar entry in the a11y dialog.

DawidKossowski:

Internal (core): Added menu bar entry in the a11y dialog.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-core/src/accessibility.ts
+++ b/packages/ckeditor5-core/src/accessibility.ts
@@ -52,6 +52,7 @@ export default class Accessibility {
 	 */
 	constructor( editor: Editor ) {
 		this._editor = editor;
+		const isMenuBarVisible = editor.config.get( 'menuBar.isVisible' );
 
 		const t = editor.locale.t;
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-core/src/accessibility.ts
```diff
@@ -61,40 +62,52 @@ export default class Accessibility {
...
+				label: t( 'Navigate through the toolbar or menu bar' ),
+				keystroke: [ [ 'arrowup' ], [ 'arrowright' ], [ 'arrowdown' ], [ 'arrowleft' ] ]
+			},
+			{
+				label: t( 'Execute the currently focused button' ),
+				keystroke: [ [ 'Enter' ], [ 'Space' ] ]
+			}
+		];
+
+		if ( isMenuBarVisible ) {
```
DawidKossowski, line 93: I could keep the previous structure and add the inline logic, like:
```
	...(
			isMenuBarVisible ?
				[ {
					label: t( 'Move focus to the menu bar, navigate between menu bars' ),
					keystroke: 'Alt+F9',
					mayRequireFn: true
				} ] :
				[]
		),
```

However, the current approach seems more clear.

PR #16125, opened 2024-03-28
Preserve repeated spaces when text is wrapped with a CSS `white-space` rule

### LINKED ISSUE #16124: Repeated spaces are not preserved when text is contained in an element that renders whitespace using a CSS `white-space` rule.

When text is wrapped in a block containing a CSS `white-space` rule with either `pre`, `pre-wrap` or `break-spaces`, repeated whitespace isn't collapsed but rendered instead. However, when text wrapped like this is inserted into CKEditor, the repeated spaces are lost.

Probably the most common scenario of this occurring is when copying content formatted with these `white-space` rules from other pages and pasting them into the editor.

As an example for the following HTML file:

```html
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="UTF-8">
	<title>Document</title>
	<style>
		div {
			white-space: pre-wrap;
		}
	</style>
</head>
<body>
	<div>    Foo    Bar    Baz    </div>
</body>
</html>
```

When I select the text in the rendered document and copy it to the clipboard with Chrome 122, the clipboard content looks like this (retrieved by running `xclip -selection clipboard -target text/html -out`):

```html
<span style="color: rgb(0, 0, 0); font-family: &quot;Times New Roman&quot;; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: pre-wrap; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">    Foo    Bar    Baz    </span>
```

So the copied structure from Chrome has a different shape than the content on the original page, and it contains the required inline CSS to render it correctly on the pasted target.

However, when pasted into CKEditor, the spaces are collapsed to a single one, and the leading and trailing whitespace is stripped.

It's also really easy to reproduce in one of CKEditor's manual tests, by setting the editor data to something like
```html
<span style="white-space: pre-wrap">Hello,    World!</span>
```

### ✔️ Expected result
I expect repeated spaces contained within a block with `white-space: pre`, `white-space: pre-wrap` or `white-space: break-spaces` to be preserved when being set or pasted into the editor.

### ❌ Actual result
Repeated spaces are collapsed.

### ❓ Possible solution
I submitted a pull request with a possible fix: https://github.com/ckeditor/ckeditor5/pull/16125

jessevanassen:

Fix (engine): Preserve repeated spaces in text that is contained within an element that renders (repeated) white space. Closes #16124

---

### Additional information

This pull request offers a possible solution for #16124.

It will preserve (repeated) whitespace when the text is contained within a block that preserves whitespace during rendering, which is done through the `white-space` CSS property. This is similar to what the code for the `<pre>` element is already doing.

Existing code within CKEditor is already taking care of swapping out the repeated spaces by `&nbsp;`-patterns, so they are also preserved when the data is extracted again.

### DIFF
--- a/packages/ckeditor5-engine/src/view/domconverter.ts
+++ b/packages/ckeditor5-engine/src/view/domconverter.ts
@@ -1604,8 +1604,7 @@ export default class DomConverter {
 	private _processDataFromViewText( node: ViewText | ViewTextProxy ): string {
 		let data = node.data;
 
-		// If any of node ancestors has a name which is in `preElements` array, then currently processed
-		// view text node is (will be) in preformatted element. We should not change whitespaces then.
+		// If the currently processed view text node is preformatted, we should not change whitespaces.
 		if ( this._isPreFormatted( node ) ) {
 			return data;
 		}
@@ -1661,7 +1660,7 @@ export default class DomConverter {
 
 	/**
 	 * Checks whether given text contains preformatted white space. This is the case if
-	 * * the text is contained within a `<pre>` element, or
+	 * * any of node ancestors has a name which is in `preElements` array, or
 	 * * the closest ancestor that has the `white-space` CSS property sets it to a value that preserves spaces
 	 *
 	 * @param node Node to check
--- a/packages/ckeditor5-engine/src/view/domconverter.ts
+++ b/packages/ckeditor5-engine/src/view/domconverter.ts
@@ -1508,7 +1508,7 @@ export default class DomConverter {
 			let data: string;
 			let nodeEndsWithSpace: boolean = false;
 
-			if ( _hasViewParentOfType( node, this.preElements ) ) {
+			if ( this._isPreFormatted( node ) ) {
 				data = getDataWithoutFiller( node.data );
 			} else {
 				// Change all consecutive whitespace characters (from the [ \n\t\r] set –
@@ -1606,7 +1606,7 @@ export default class DomConverter {
 
 		// If any of node ancestors has a name which is in `preElements` array, then currently processed
 		// view text node is (will be) in preformatted element. We should not change whitespaces then.
-		if ( node.getAncestors().some( parent => this.preElements.includes( ( parent as ViewElement ).name ) ) ) {
+		if ( this._isPreFormatted( node ) ) {
 			return data;
 		}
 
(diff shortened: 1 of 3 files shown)


Dumluregn (changes requested):
Hi, thank you for contributing to our repository.

1. To accept your code, we need you to sign the Contributor License Agreement (CLA) - see https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/contributing.html#contributor-license-agreement-cla.
2. Since the problem you're fixing isn't reproducible in the editor itself, please provide a sample environment showing that the fix works.
3. What if the `white-space` property is set in the style sheet instead of directly in data?
4. There are already two custom matchers in the `DomConverter`: https://github.com/jessevanassen/ckeditor5/blob/54ebaddc4f3a077e479fa636fd3675a51eb5400b/packages/ckeditor5-engine/src/view/domconverter.ts#L143-L148. Maybe it would be possible to implement your check in a similar manner? It could make checking the ancestors more difficult probably though, so it's not a must.

jessevanassen:
Hi @Dumluregn, thanks for coming back to me!

1. I've already signed the CLA
2. The sample environment is quite easy: just an HTML page containing the CSS rule applied globally, and some content with some white space padding around it. I put a sample in the description of #16124. \
Would you like me to add this test case to the `test/manual` folder of the repo as well?
3. This depends on the browser. The Chromium / Webkit family does it differently than the Firefox browsers. \
In a Chromium or Webkit browser, the HTML that is placed on the clipboard is derived from what is rendered. So if for instance `white-space: pre-wrap` is part of a style sheet, the HTML nodes on the clipboard will still get a `style="white-space: pre-wrap"` attribute. \
In Firefox on the other hand, it will basically copy over what is in the DOM, without adding additional style attributes.
4. I will take a look at that later and come back to you, thanks for the pointer!

jessevanassen:
I've looked at the Matchers, but to me it feels like they wouldn't be the right approach for this check.

If I understand them correctly, a `Matcher` is something that can be initialized with different kinds of patterns, and can then be used to "filter" elements.

However, for the check, I want to see if a specific element is preformatted, and take specific actions if it is (or isn't). I could try to implement it with a `Matcher`, the pattern that is a function that takes an `Element` can probably be used for this. I think turning it into a `Matcher` will make it less convenient to use though, without any benefits. The result of the `Matcher` (which element, rule, attributes or styles matched, etc.) will be irrelevant, so in the calling code I would only be checking _if_ I have a match, but the actual details won't be used. And the `Matcher` also requires more setup, and isn't immutable after being constructed (more patterns can be added later which isn't desirable).

Like it is now, a pure function that returns a boolean, feels better suited for the task. Would you agree, or do you have other insights?

packages/ckeditor5-engine/src/view/domconverter.ts
```diff
@@ -1593,7 +1593,7 @@ export default class DomConverter {
 
 		// If any of node ancestors has a name which is in `preElements` array, then currently processed
 		// view text node is (will be) in preformatted element. We should not change whitespaces then.
```
Dumluregn, line 1595: This comments need updating (mentioning the `white-space` property).

Dumluregn (commented):
I see it is indeed simple to reproduce the problem in any manual test, so I'll just repost it here:
> It's also really easy to reproduce in one of CKEditor's manual tests, by setting the editor data to something like
> `<span style="white-space: pre-wrap">Hello,    World!</span>`

Thank you for investigating the Matcher approach. I think it too little benefits to pursue it. 

Could you please merge the latest `master` branch and ensure it still works fine? I left just one comment and we'll be good to merge this PR.

packages/ckeditor5-engine/src/view/domconverter.ts
jessevanassen: I've simplified the comment by just mentioning if it is preformatted, because `_isPreFormatted` already documents the scenarios in which a node is considered preformatted.
I did notice that the JSDoc on `_isPreFormatted` mentioned `<pre>` elements directly and not the `preElements` array, so I've corrected it there as well.

PR #16140, opened 2024-04-02
Added the image resize balloon

Mati365:

Feature (image): Add custom image width input option in the image editing toolbar, as an alternative to drag-and-drop resizing

---

### Additional information

Related to https://github.com/cksource/ckeditor5-commercial/issues/6023
Related to https://github.com/ckeditor/ckeditor5/pull/16104
Screens:
obraz
obraz


oleq (changes requested):
1. You can submit an invalid value. I think we need to validate these for empty values just like links or media

https://github.com/ckeditor/ckeditor5/assets/1099479/0db2bbae-686a-4c6e-ad44-8e4ac6382e4f

2. When in px, there's a min value limit set to 50px. This is no good because we allow for resizing inline images too that are often used as emoji or just really small images (see the [docs](https://ckeditor.com/docs/ckeditor5/latest/features/images/images-overview.html)). They go way below 50px limit.

image

image

3. When in px, the upper limit of 2400 is also problematic. You can easily use a wider hero image in HiDPI than 2400 (or especially for print, export to PDF).

image


5. There's something odd going on with units and I spotted it in the [picture manual test](http://fake.ckeditor.com:8125/ckeditor5-image/tests/manual/picture.html). The input label is in % but the value is clearly in px (put 100 there and it will commit as 100%, though)

image

Mati365:
@oleq Good catch, will be fixed. I'll change a bit min / max values.

packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
```diff
@@ -39,7 +43,7 @@ export default class ImageResizeButtons extends Plugin {
 	 * @inheritDoc
 	 */
 	public static get requires() {
-		return [ ImageResizeEditing ] as const;
+		return [ ImageResizeEditing, ImageCustomResizeUI ] as const;
```
oleq, line 46: Is that requirement necessary? I mean, we could soft-detect whether the plugin is loaded and then add a button or not. This just feels like an unnecessary feature coupling.

packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
```diff
@@ -179,8 +193,12 @@ export default class ImageResizeButtons extends Plugin {
 
 			// Execute command when an item from the dropdown is selected.
 			this.listenTo( dropdownView, 'execute', evt => {
-				editor.execute( ( evt.source as any ).commandName, { width: ( evt.source as any ).commandValue } );
-				editor.editing.view.focus();
+				if ( 'onClick' in evt.source ) {
+					( evt.source as any ).onClick();
```
oleq, line 197: I hate these `ViewModels` and `as any` that it brings in so many places across the project :/ I wish we got rid of them. 

Maybe instead of duck typing we could assign a legitimate property to the view model like `#buttonType` or something and check that directly here? The outcome would be the same but this would look less like a magic.

oleq: Maybe this is a more elegant way of doing things (even though we still have to cast):

```diff
diff --git a/packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts b/packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
index 1775a9b136..7b25bb203a 100644
--- a/packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
+++ b/packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
@@ -193,10 +193,12 @@ export default class ImageResizeButtons extends Plugin {
 
 			// Execute command when an item from the dropdown is selected.
 			this.listenTo( dropdownView, 'execute', evt => {
-				if ( 'onClick' in evt.source ) {
-					( evt.source as any ).onClick();
+				const eventSource = evt.source as ImageResizeButtonModel;
+
+				if ( eventSource.type == 'custom' ) {
+					this.editor.plugins.get( 'ImageCustomResizeUI' )._showForm( this._resizeUnit );
 				} else {
-					editor.execute( ( evt.source as any ).commandName, { width: ( evt.source as any ).commandValue } );
+					editor.execute( eventSource.commandName, { width: eventSource.commandValue } );
 					editor.editing.view.focus();
 				}
 			} );
@@ -252,7 +254,6 @@ export default class ImageResizeButtons extends Plugin {
 		options: Array<ImageResizeOption>,
 		command: ResizeImageCommand
 	): Collection<ListDropdownItemDefinition> {
-		const customResizeUI = this.editor.plugins.get( 'ImageCustomResizeUI' );
 		const itemDefinitions = new Collection<ListDropdownItemDefinition>();
 
 		const optionsWithSerializedValues = options.map( option => {
@@ -282,14 +283,12 @@ export default class ImageResizeButtons extends Plugin {
 			if ( isCustomImageResizeOption( option ) ) {
 				definition = {
 					type: 'button',
-					model: new ViewModel( {
+					model: new ImageResizeButtonModel( {
 						label: this._getOptionLabelValue( option ),
 						role: 'menuitemradio',
 						withText: true,
 						icon: null,
-						onClick: () => {
-							customResizeUI._showForm( this._resizeUnit );
-						}
+						type: 'custom'
 					} )
 				};
 
@@ -299,13 +298,14 @@ export default class ImageResizeButtons extends Plugin {
 			} else {
 				definition = {
 					type: 'button',
-					model: new ViewModel( {
+					model: new ImageResizeButtonModel( {
 						commandName: 'resizeImage',
 						commandValue: option.valueWithUnits,
 						label: this._getOptionLabelValue( option ),
 						role: 'menuitemradio',
 						withText: true,
-						icon: null
+						icon: null,
+						type: 'fixed'
 					} )
 				};
 
@@ -349,3 +349,14 @@ function getIsOnCustomButtonCallback( allDropdownValues: Array<string | null> ):
 		dropdownValue => getIsOnButtonCallback( dropdownValue )( commandValue )
 	);
 }
+
+class ImageResizeButtonModel extends ViewModel {
+	declare public type: 'fixed' | 'custom';
+	declare public commandName: string;
+	declare public commandValue: string;
+	declare public icon?: string;
+	declare public label: string;
+	declare public onClick?: Function;
+	declare public role: string;
+	declare public withText: boolean;
+}
```

oleq (changes requested):
I also spotted some layout issue with narrow viewports

image

packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
Mati365: @oleq I'm not sure if it is needed. We already have `isCustomImageResizeOption` which detects if custom dropdown entry belongs to `ImageCustomResizeUI` plugin (and creates dropdown entry for such plugin) so plugins are coupled anyway. It would make sense if `ImageCustomResizeUI` had possibility to dynamically append dropdown entries to `ImageResizeButtons` plugin but on the other hand it'd be more complicated and probably noone would even use these two plugins separately because menu entries are configured via config.

Mati365:
@oleq All reported CR remarks fixed.

packages/ckeditor5-image/src/imageresize/resizeimagecommand.ts
```diff
@@ -77,4 +79,59 @@ export default class ResizeImageCommand extends Command {
...
+	 * Returns image width in specified units.
+	 *
+	 * 	* If image is not selected or command is disabled then `null` will be returned.
+	 * 	* If image is not fully loaded (and it is impossible to determine its natural size) then `null` will be returned.
+	 *  * If `targetUnit` percentage is passed then it will return width percentage of image related to its accessors.
+	 *
+	 * @param targetUnit Unit in which dimension will be returned.
+	 * @returns Parsed dimension with unit.
+	 */
+	public getSelectedImageWidthInUnits( targetUnit: string ): DimensionWithUnit | null {
```
oleq, line 93: This method is stateless and should be generic in `utils/` as `getImageWidthInUnits()` (for any passed image, selected or not). It has nothing to do with the command.

packages/ckeditor5-image/src/imageresize/resizeimagecommand.ts
```diff
@@ -77,4 +79,59 @@ export default class ResizeImageCommand extends Command {
...
+
+		const imageModelElement = imageUtils.getClosestSelectedImageElement( editor.model.document.selection )!;
+		const imageViewElement = editing.mapper.toViewElement( imageModelElement );
+		const imageDOMElement = editing.view.domConverter.mapViewToDom( imageViewElement! );
+
+		if ( !imageDOMElement ) {
+			return null;
+		}
+
+		const imageHolderWidth = imageDOMElement.getBoundingClientRect().width;
```
oleq, line 120: `new Rect( imageDOMElement ).width`

packages/ckeditor5-image/src/imageresize/resizeimagecommand.ts
```diff
@@ -77,4 +79,59 @@ export default class ResizeImageCommand extends Command {
...
+
+		if ( parsedWidth.unit === targetUnit ) {
+			return parsedWidth;
+		}
+
+		const imageModelElement = imageUtils.getClosestSelectedImageElement( editor.model.document.selection )!;
+		const imageViewElement = editing.mapper.toViewElement( imageModelElement );
+		const imageDOMElement = editing.view.domConverter.mapViewToDom( imageViewElement! );
+
+		if ( !imageDOMElement ) {
```
oleq, line 116: When would this happen? The method returns early if the command has no `value`.

packages/ckeditor5-widget/src/widgetresize/resizerstate.ts
```diff
@@ -189,9 +189,11 @@ export default class ResizeState extends ObservableMixin() {
 }
 
 /**
- * Calculates a relative width of a `domResizeHost` compared to its ancestor in percents.
+ * Calculates a relative width of a `domResizeHost` compared to its ancestor in pixels.
+ *
+ * @internal
  */
-function calculateHostPercentageWidth( domResizeHost: HTMLElement, resizeHostRect: Rect ) {
+export function calculateHostWidth( domResizeHost: HTMLElement ): number {
```
oleq, line 196: I don't think we'll get away with making this internal and exporting in the package. If exported, this will have to have proper API docs and unit tests.

Mati365:
@oleq Fixed

oleq (changes requested):

packages/ckeditor5-image/src/imageresize/resizeimagecommand.ts
Mati365: Condition removed

packages/ckeditor5-widget/src/widgetresize/resizerstate.ts
Mati365: @oleq API docs updated, unit tests added.

packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
Mati365: @oleq I made this dependency optional https://github.com/ckeditor/ckeditor5/pull/16140/commits/b24625433d2cb4bfa62a21e7312a6f316736e99b. However, I kept the custom button in the dropdown because it can be used in the future in third party custom integrations.

Mati365:
@oleq I fixed all CR remarks.

charlttsie:
## Inconsistencies when it comes to the minimal value that can be inserted
### Steps
1. Open e.g. http://localhost:8125/ckeditor5-image/tests/manual/imageresize.html
https://github.com/ckeditor/ckeditor5/assets/132071257/3f9db8db-4a3b-439a-b844-0e291f5d4463
2. Try to insert 0,1 as a custom image size
3. Resize an inline image to the smallest possible size using resize handler
4. Resize an inline image to the smallest possible size using resize handler

### Results
The minimal value that can be inserted is 0,2 - both for inline and block images:

https://github.com/ckeditor/ckeditor5/assets/132071257/6cb8f6ca-5743-45e4-8775-22576b3acd9b

However, an inline image can be made smaller using the resize handler:
inline-image

And at the same time, a block image can't be made that small and incorrect values are shown from a certain point:

https://github.com/ckeditor/ckeditor5/assets/132071257/7fad7ff8-7b2e-4e38-9c24-c387ce91ec84

charlttsie:
## Resize balloon shows a wrong maximum image size in pixels
### Steps
1. Open http://localhost:8125/ckeditor5-image/tests/manual/imageresizepx.html
2. Enlarge an image to the maximum possible size using resize balloon
2. Enlarge an image to the maximum possible size using resize handler

### Result
Resize balloon allows for setting the size to 1600 px, while the image has only 1548 px at its maximum size and this is what's shown after resize handler is used:

https://github.com/ckeditor/ckeditor5/assets/132071257/c383bda9-67ba-495d-8be4-70b717fa2dba

Mati365:
@charlttsie Please pull, some modifications regarding image resize range has been pushed.

charlttsie:
@Mati365 
https://github.com/ckeditor/ckeditor5/pull/16140#issuecomment-2051203348 looks good for inline images.
For block images, the limit looks good now:
block-1
However, the field can still show an incorrect value after using resize handler:
resize-block
This looks related to an issue that's also reproducible on master, where from a certain point when resizing, the actual image size doesn't change anymore yet the number is still changing:

https://github.com/ckeditor/ckeditor5/assets/132071257/8e5530a1-d8f0-4e0c-96be-d4a45a72a24d

So, I'm not sure if that's within the scope 🤔

Mati365:
@charlttsie It's out of scope, this is not related to this particular implementation (as it just works identically as "draggable" one). Anyway - thanks for report.

charlttsie:
Regarding https://github.com/ckeditor/ckeditor5/pull/16140#issuecomment-2051272727, when I take the same image from http://localhost:8125/ckeditor5-image/tests/manual/imageresizepx.html and resize it using resize handler, the size shown is 1548 px and it also shows inside the resize balloon field. However, when I want to submit it, it won't get accepted:

https://github.com/ckeditor/ckeditor5/assets/132071257/4be8e60a-7ef5-4453-8bc3-c1903a921d5b

charlttsie:
https://github.com/ckeditor/ckeditor5/pull/16140#issuecomment-2051616834 looks good 👍

charlttsie:
FYI: I reported the issue with the incorrect size when shrinking a block image here: https://github.com/ckeditor/ckeditor5/issues/16208

charlttsie:
We've finished testing with @kubaklatt and @mabryl and besides the already reported and fixed issues, the PR looks good. Thank you @Mati365 for the quick fixes 🎉

Mati365:
@oleq UI update test added.

PR #16190, opened 2024-04-08
Disable comments archive ui plugin when disabling commands.

mremiszewski:

Internal (comments): Disable comments archive buttons in source editing mode.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-source-editing/src/sourceediting.ts
+++ b/packages/ckeditor5-source-editing/src/sourceediting.ts
@@ -329,6 +329,13 @@ export default class SourceEditing extends Plugin {
 		for ( const command of editor.commands.commands() ) {
 			command.forceDisabled( COMMAND_FORCE_DISABLE_ID );
 		}
+
+		// Comments archive UI plugin will be disabled manually too.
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-source-editing/src/sourceediting.ts
```diff
@@ -329,6 +329,13 @@ export default class SourceEditing extends Plugin {
 		for ( const command of editor.commands.commands() ) {
 			command.forceDisabled( COMMAND_FORCE_DISABLE_ID );
 		}
+
+		// Comments archive UI plugin will be disabled manually too.
+		if ( !editor.plugins.has( 'CommentsArchiveUI' ) ) {
+			return;
+		}
+
+		( editor.plugins.get( 'CommentsArchiveUI' ) as Plugin ).forceDisabled( COMMAND_FORCE_DISABLE_ID );
```
scofalik, line 338: In this case, I would structure the code this way:

```
if ( something ) {
    disable() / enable();
}
```

Instead of returning early. Why? If there are more things that needs to be added here, it will need to be refactored anyway. Also, somebody may make a mistake by adding more stuff at the end (which wouldn't work if there's no comments archive).

PR #16197, opened 2024-04-11
Aria live announcements A11y improvements

### LINKED ISSUE #16053: Add aria-label to code block

Mati365:

Feature (code-block): Introduced screen reader announcements for entering or exiting code block elements in the document editor. Closes #16053.

Other (ui): Refactored the `AriaLiveAnnouncer` to use the `aria-relevant` attribute and make concurrent announcements queued by screen readers.

MINOR BREAKING CHANGE (ui): The region name argument of the `AriaLiveAnnouncer#announce()`  method has been dropped. Please check out the latest API documentation for more information. 

### Tickets to close

https://github.com/cksource/ckeditor5-commercial/issues/6031.
https://github.com/cksource/ckeditor5-commercial/issues/6033

Part of: 
https://github.com/cksource/ckeditor5-commercial/pull/6163

Announcer refactor:
https://github.com/ckeditor/ckeditor5/pull/16199

### DIFF
--- a/packages/ckeditor5-list/src/todolist/todolistediting.ts
+++ b/packages/ckeditor5-list/src/todolist/todolistediting.ts
@@ -382,9 +382,9 @@ export default class TodoListEditing extends Plugin {
 			const currentElementIsTodoList = isTodoListItemElement( focusParent );
 
 			if ( lastElementIsTodoList && !currentElementIsTodoList ) {
-				ui.ariaLiveAnnouncer.announce( t( 'Leaving todo list' ) );
+				ui.ariaLiveAnnouncer.announce( t( 'Leaving a to-do list' ) );
 			} else if ( !lastElementIsTodoList && currentElementIsTodoList ) {
-				ui.ariaLiveAnnouncer.announce( t( 'Entering todo list' ) );
+				ui.ariaLiveAnnouncer.announce( t( 'Entering a to-do list' ) );
 			}
 
 			lastFocusedCodeBlock = focusParent;
--- a/packages/ckeditor5-list/src/todolist/todolistediting.ts
+++ b/packages/ckeditor5-list/src/todolist/todolistediting.ts
@@ -555,5 +555,13 @@ function isLabelElement( viewElement: ViewElement | ViewDocumentFragment | null 
  * Returns true if the given element is a list item model element of a to-do list.
  */
 function isTodoListItemElement( element: Element | DocumentFragment | null ): boolean {
-	return !!element && element.is( 'element', 'paragraph' ) && element.getAttribute( 'listType' ) == 'todo';
+	if ( !element ) {
+		return false;
+	}
+
+	if ( !element.is( 'element', 'paragraph' ) && !element.is( 'element', 'listItem' ) ) {
+		return false;
+	}
+
+	return element.getAttribute( 'listType' ) == 'todo';
 }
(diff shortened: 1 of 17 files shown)


oleq (commented):
This PR addresses
* code block
* image upload
* list (todo-list)

What we're missing from https://github.com/cksource/ckeditor5-commercial/issues/6031 are:
*  AI Assistant does not properly indicates changes to the status of the content generation
*  AI commands and Templates search does not inform the user about the results of the search:
*  Case change feature does not make any sense for the screen reader user. “Title”, “title” and “TITLE” are all read the same. We should notify the user that the case changed.
*  Export to PDF and Export to Word toolbar buttons do not announce the current status of export.

packages/ckeditor5-list/src/todolist/todolistediting.ts
```diff
@@ -357,6 +361,35 @@ export default class TodoListEditing extends Plugin {
...
+
+			if ( lastElementIsTodoList && !currentElementIsTodoList ) {
+				ui.ariaLiveAnnouncer.announce( t( 'Leaving a to-do list' ) );
+			} else if ( !lastElementIsTodoList && currentElementIsTodoList ) {
+				ui.ariaLiveAnnouncer.announce( t( 'Entering a to-do list' ) );
+			}
+
+			lastFocusedCodeBlock = focusParent;
+		} );
+	}
```
oleq, line 392: This passes unit tests but fails in real life. Please visit the todo-documentlist-simple manual test to see that the announcements are not read out.

I think this is because depending on the configuration, the list are represented in a different way in the model https://github.com/ckeditor/ckeditor5/blob/f29e0a09a1e31101fc9c58501b0cbbaeb65fd9f8/packages/ckeditor5-list/src/list/listediting.ts#L150-L159.

The utility that discovers them has to be smarter. See

https://github.com/ckeditor/ckeditor5/blob/f29e0a09a1e31101fc9c58501b0cbbaeb65fd9f8/packages/ckeditor5-list/src/todolist/todolistediting.ts#L77-L79

oleq (changes requested):

packages/ckeditor5-list/src/todolist/todolistediting.ts
Mati365: Fixed

oleq (approved):

PR #16233, opened 2024-04-17
Docs: updating CSP guide.

godai78:

Docs: updating CSP guide.

---

### Additional information

### DIFF
--- a/docs/getting-started/setup/csp.md
+++ b/docs/getting-started/setup/csp.md
@@ -18,7 +18,7 @@ CKEditor&nbsp;5 is compatible with applications that use [<abbr title="Content S
 The recommended CSP configuration that allows the rich-text editor to run out–of–the–box with all standard features using the content like images or media from external hosts looks as follows:
 
 ```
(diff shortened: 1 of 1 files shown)


docs/getting-started/setup/csp.md
```diff
@@ -18,7 +18,7 @@ CKEditor&nbsp;5 is compatible with applications that use [<abbr title="Content S
 The recommended CSP configuration that allows the rich-text editor to run out–of–the–box with all standard features using the content like images or media from external hosts looks as follows:
 
 ```
-default-src 'none'; connect-src 'self'; script-src 'self'; img-src * data:; style-src 'self' 'unsafe-inline'; frame-src *
```
godai78, line 21: Here, not sure if we should remove all of this, or just the `unsafe-inline` part?  
  
Same below.

Witoso: Just `unsafe-inline` in this one.

`style-src 'self'` should stay in the latter.

PR #16241, opened 2024-04-18
Redoing API code docs

illia-stv:

Docs: Redoing API code docs. Closes #000.

---

### Additional information

_Merge with_ cksource/ckeditor5-commercial#6181


packages/ckeditor5-core/src/editor/editorconfig.ts
```diff
@@ -547,16 +570,18 @@ export interface EditorConfig {
 	plugins?: Array<PluginConstructor<Editor> | string>;
 
 	/**
+	 * ------TODO redirect to proper guide--------
```
illia-stv, line 568: @witek, @godai78 I would like to ask your opinion how to handle this part.

packages/ckeditor5-core/src/editor/editorconfig.ts
```diff
@@ -31,17 +31,16 @@ import type { MenuBarConfig } from '@ckeditor/ckeditor5-ui';
...
  * ```
- *
- * Check the {@glink getting-started/legacy/installation-methods/predefined-builds Configuration} guide for more information
- * about setting configuration options.
  */
 export interface EditorConfig {
 	context?: Context;
 
 	/**
+	 * ------TODO redirect to proper guide--------
```
illia-stv, line 39: @witek, @godai78 I would like to ask your opinion how to handle this part.

packages/ckeditor5-core/src/editor/editorconfig.ts
godai78: Not sure about it, we abandon "custom builds" so the closest here is a situation when we want to hack a Builder output file.  
  
Considering we should be using less and less of this command now, we can just leave it as is at this stage and address it later; we just need to keep a tab on it.

Witoso: IMHO we should describe what this method does, and AFAIK it's available on all editor types. Something like:  
`The list of additional plugins to load along those already available in the editor. This could be used when someone is using multiple editors on a page, and extends the base class.`

packages/ckeditor5-core/src/editor/editorconfig.ts
Witoso: IMHO we should describe what this method does, and AFAIK it's available on all editor types. Something like:  
\``he list of additional plugins to load along those already available in the editor. This could be used when someone is using multiple editors on a page, and extends the of the shared editor class`

packages/ckeditor5-core/src/editor/editorconfig.ts
Witoso: IMHO we should describe what this method does, and AFAIK it's available on all editor types. Something like:  
`The list of plugins which should not be loaded despite being available in editor This could be used when someone is using multiple editors on a page, and wants to dynamically decrease the plugins of the shared editor class.`

godai78:
Is there a companion PR for premium features, too?

illia-stv:
PR in Comercial: https://github.com/cksource/ckeditor5-commercial/pull/6181

packages/ckeditor5-core/src/editor/editorconfig.ts
```diff
@@ -36,11 +36,8 @@ export interface EditorConfig {
 	context?: Context;
 
 	/**
-	 * ------TODO redirect to proper guide--------
 	 * The list of additional plugins to load along those already available in the
-	 * {@glink getting-started/legacy/installation-methods/predefined-builds editor build}.
-	 * It extends the {@link #plugins `plugins`} configuration.
-	 * -------------------------------------------
+	 * editor. It extends the {@link #plugins `plugins`} configuration.
```
godai78, line 40: DO we actually use this notation for links? First time me seeing it, honestly.

illia-stv: Unfortunately I don't know, these notation for links was there before my changes.

PR #16243, opened 2024-04-18
The list properties toolbar is overriden by List plugin

### LINKED ISSUE #16192: The list properties toolbar is overriden by `List` plugin

Create an editor containing the plugins in the exact order: `[ ..., ListProperties, List, ... ]`.

To reproduce the issue, you can use the following manual test.

https://github.com/ckeditor/ckeditor5/blob/1b3913c87f52b23a10b0f81495ee8e24377a18c0/packages/ckeditor5-list/tests/manual/documentlist-properties.js#L49

### ✔️ Expected result

image

The `∨` symbol should be visible.

### ❌ Actual result

image

The `ListUI` overrides existing icons. The `∨` symbol is not visible.

### ❓ Possible solution

In the past (#7861), we agreed to override icon toolbars. While it makes sense when handling `List` => `ListProperties`, it does not make sense for `ListProperties` => `List`.

## 📃 Other details

* Browser: …
* OS: …

---

illia-stv:

Fix (List): Order of `List` and `ListProperties` plugins should not affect the appearance of the icon in the toolbar. Closes #16192.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/src/list/listui.ts
+++ b/packages/ckeditor5-list/src/list/listui.ts
@@ -28,8 +28,14 @@ export default class ListUI extends Plugin {
 	public init(): void {
 		const t = this.editor.t;
 
-		// Create two buttons and link them with numberedList and bulletedList commands.
-		createUIComponents( this.editor, 'numberedList', t( 'Numbered List' ), icons.numberedList );
-		createUIComponents( this.editor, 'bulletedList', t( 'Bulleted List' ), icons.bulletedList );
+		// Create button numberedList.
+		if ( !this.editor.ui.componentFactory.has( 'numberedList' ) ) {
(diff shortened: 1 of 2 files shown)


pszczesniak (changes requested):
Maybe would be good to have the check, if `componentFactory` has already an item, directly here, in this function `createUIComponents`: https://github.com/ckeditor/ckeditor5/blob/45a1f1e3fc463cf069310c20a3a337176028c1dc/packages/ckeditor5-list/src/list/utils.ts#L28 as an early return, WDYT?

This change would allow us to not need to had any side effects like this in the future.

pszczesniak (approved):

pszczesniak (commented):
Maybe would be good to update the commit message, WDYT about something like this:

Order of `List` and `ListProperties` plugins should not affect the appearance of the icon in the toolbar.

PR #16252, opened 2024-04-19
Docs: React multi-root guide

godai78:

Docs: React multi-root guide.

---

### Additional information

## Summary by CodeRabbit

- **Documentation**
    - Enhanced documentation for integrating CKEditor 5 into React apps, highlighting the multi-root editor hook.

### DIFF
--- a/docs/installation/integrations/react-multiroot.md
+++ b/docs/installation/integrations/react-multiroot.md
@@ -88,7 +88,7 @@ The `useMultiRootEditor` hook supports the following properties:
 * `disableWatchdog: Boolean` &ndash; If set to `true`, {@link features/watchdog the watchdog feature} will be disabled. It is set to `false` by default.
 * `watchdogConfig: WatchdogConfig` &ndash; {@link module:watchdog/watchdog~WatchdogConfig Configuration object} for the [watchdog feature](https://ckeditor.com/docs/ckeditor5/latest/features/watchdog.html).
 * `isLayoutReady: Boolean` &ndash; A property that delays the editor creation when set to `false`. It starts the initialization of the multi-root editor when sets to `true`. Useful when the CKEditor&nbsp;5 annotations or a presence list are used.
-* `disableTwoWayDataBinding: Boolean` &ndash; Allows disabling the two-way data binding mechanism between the editor state and `data` object in order to improve editor efficiency. The default value is `false`.
+* `disableTwoWayDataBinding: Boolean` &ndash; Allows disabling the two-way data binding mechanism between the editor state and `data` object to improve editor efficiency. The default value is `false`.
 * `onReady: Function` &ndash; It is called when the editor is ready with a {@link module:editor-multi-root/multirooteditor~MultiRootEditor `MultiRootEditor`} instance. This callback is also called after the reinitialization of the component if an error occurred.
 * `onChange: Function` &ndash; It is called when the editor data has changed. See the {@link module:engine/model/document~Document#event:change:data `editor.model.document#change:data`} event.
 * `onBlur: Function` &ndash; It is called when the editor was blurred. See the {@link module:engine/view/document~Document#event:blur `editor.editing.view.document#blur`} event.
@@ -132,7 +132,5 @@ By default, the two-way data binding is enabled. It means that every change done
 </info-box>
 
 ## Contributing and reporting issues
-
-<!-- Be sure to update this if needed -->
 
 The source code of rich text editor component for React is available on GitHub in [https://github.com/ckeditor/ckeditor5-react](https://github.com/ckeditor/ckeditor5-react).
--- a/docs/installation/integrations/react-multiroot.md
+++ b/docs/installation/integrations/react-multiroot.md
@@ -89,7 +89,7 @@ The `useMultiRootEditor` hook supports the following properties:
 * `watchdogConfig` &ndash; {@link module:watchdog/watchdog~WatchdogConfig Configuration object} for the [watchdog feature](https://ckeditor.com/docs/ckeditor5/latest/features/watchdog.html).
 * `isLayoutReady` &ndash; A property that delays the editor creation when set to `false`. It starts the initialization of the multi-root editor when sets to `true`. Useful when the CKEditor&nbsp;5 annotations or a presence list are used.
 * `disableTwoWayDataBinding` &ndash; Allows disabling the two-way data binding mechanism between the editor state and `data` object. The default value is false.
-* `onReady` &ndash; A function called when the editor is ready with an {@link module:editor-multi-root/multirooteditor~MultiRootEditor `MultiRootEditor`} instance. This callback is also called after the reinitialization of the component if an error occurred.
+* `onReady` &ndash; A function called when the editor is ready with a {@link module:editor-multi-root/multirooteditor~MultiRootEditor `MultiRootEditor`} instance. This callback is also called after the reinitialization of the component if an error occurred.
 * `onChange` &ndash; A function called when the editor data has changed. See the {@link module:engine/model/document~Document#event:change:data `editor.model.document#change:data`} event.
 * `onBlur` &ndash; A function called when the editor was blurred. See the {@link module:engine/view/document~Document#event:blur `editor.editing.view.document#blur`} event.
 * `onFocus` &ndash; A function called when the editor was focused. See the {@link module:engine/view/document~Document#event:focus `editor.editing.view.document#focus`} event.
--- a/docs/installation/integrations/react-multiroot.md
+++ b/docs/installation/integrations/react-multiroot.md
@@ -78,16 +78,16 @@ export default App;
 
 The `useMultiRootEditor` hook supports the following properties:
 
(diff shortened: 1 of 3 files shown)


godai78:
@DawidKossowski, @Witoso - let's decide whether we want it in a subfolder (like it is now) or side-by-side rather (like the Vue guides are). Before we merge it and add some more redirs :)

DawidKossowski:
When it comes to Vue integration, the sections are side-by-side because they concern different versions of Vue. In the case of React integration, we still refer to the same version but utilize different editor types and approaches. That's why I believe the subfolder is clearer.

docs/installation/integrations/react-multiroot.md
```diff
@@ -0,0 +1,139 @@
...
+
+<p>
+	<a href="https://www.npmjs.com/package/@ckeditor/ckeditor5-react" target="_blank" rel="noopener">
+		<img src="https://badge.fury.io/js/%40ckeditor%2Fckeditor5-react.svg" alt="npm version" loading="lazy">
+	</a>
+</p>
+
+CKEditor&nbsp;5 consists of {@link installation/getting-started/predefined-builds ready-to-use editor builds} and {@link framework/index CKEditor&nbsp;5 Framework} upon which the builds are based.
+
+The easiest way to use multi-root CKEditor&nbsp;5 in your React application is by using the {@link installation/getting-started/predefined-builds#multi-root-editor multi-root rich text editor build}. You can also use a customized editor built by using [CKEditor&nbsp;5 online builder](https://ckeditor.com/ckeditor-5/online-builder/) in any React application.
```
scofalik, line 20: We don't have multi-root in the online builder.

scofalik: This whole intro seems confusing / unnecessary. I take it that it is a copy-paste from another guide, but maybe let's think if it makes sense here. Why the user should read it? What info the user gets from it? Do we want to redirect the user or show they a better/different approach?

This is a guide for a specific, multi-root integration. I'd assume that the user knows what they are doing if they are here. It may make more sense to give here some hint that this is for multi-root editor that supports multiple editable areas. If they are looking for a classic integration, they should look at the regular React integration. This will help people who get here straight from Google, for example.

docs/installation/integrations/react-multiroot.md
```diff
@@ -0,0 +1,139 @@
...
+			// your config
+		}
+	};
+
+	const {
+		editor, toolbarElement, editableElements,
+		data, setData,
+		attributes, setAttributes
+	} = useMultiRootEditor( editorProps );
+
```
scofalik, line 64: Double empty line here.

docs/installation/integrations/react-multiroot.md
```diff
@@ -0,0 +1,139 @@
...
+const App = () => {
+	const editorProps = {
+		editor: MultiRootEditor,
+		data: {
+			intro: '<h1>React multi-root editor</h1>',
+			content: '<p>Hello from CKEditor&nbsp;5 multi-root!</p>'
+		},
+		config: {
+			// your config
```
scofalik, line 55:
```suggestion
			// Your editor config.
```

Make sure that the user does not confuse it with maybe some kind of React integration config, and they look for a correct thing.

docs/installation/integrations/react-multiroot.md
godai78: Wonder how will we do it with the New Installation Methods...

docs/installation/integrations/react-multiroot.md
```diff
@@ -0,0 +1,139 @@
...
+}
+
+export default App;
+```
+
+## Hook properties
+
+The `useMultiRootEditor` hook supports the following properties:
+
+* `editor` (required) &ndash; The {@link module:editor-multi-root/multirooteditor~MultiRootEditor `MultiRootEditor`} constructor to use.
```
scofalik, line 84: Not sure how we call it in the other places. Maybe we should call it build, not constructor? Just a thought. If we call it constructor somewhere else, then fine.

godai78: Same notice, we'll be dropping the idea of "builds" pretty soon. We tend to call them "presets" in the new approach, but it varies.

docs/installation/integrations/react-multiroot.md
godai78: In commit.

docs/installation/integrations/react-multiroot.md
godai78: In commit.

docs/installation/integrations/react-multiroot.md
```diff
@@ -0,0 +1,139 @@
...
+* `data` &ndash; The initial data for the created editor. See the {@link installation/getting-started/getting-and-setting-data Getting and setting data} guide.
+* `rootsAttributes` &ndash; The initial roots attributes for the created editor.
+* `config` &ndash; The editor configuration. See the {@link installation/getting-started/configuration Configuration} guide.
+* `disabled` &ndash; A Boolean value. The {@link module:editor-multi-root/multirooteditor~MultiRootEditor `MultiRootEditor`} is being switched to read-only mode if the property is set to `true`.
```
scofalik, line 87: Could we follow following convention for all parameters of this list?

```javascript
disabled: Boolean
config: EditorConfig
```

etc? WDYT?

scofalik: Instaed of "A Boolean value".

docs/installation/integrations/react-multiroot.md
```diff
@@ -0,0 +1,139 @@
...
+* `disableWatchdog` &ndash; A Boolean value. If set to `true`, {@link features/watchdog the watchdog feature} will be disabled. It is set to `false` by default.
+* `watchdogConfig` &ndash; {@link module:watchdog/watchdog~WatchdogConfig Configuration object} for the [watchdog feature](https://ckeditor.com/docs/ckeditor5/latest/features/watchdog.html).
+* `isLayoutReady` &ndash; A property that delays the editor creation when set to `false`. It starts the initialization of the multi-root editor when sets to `true`. Useful when the CKEditor&nbsp;5 annotations or a presence list are used.
+* `disableTwoWayDataBinding` &ndash; Allows disabling the two-way data binding mechanism between the editor state and `data` object. The default value is false.
+* `onReady` &ndash; A function called when the editor is ready with an {@link module:editor-multi-root/multirooteditor~MultiRootEditor `MultiRootEditor`} instance. This callback is also called after the reinitialization of the component if an error occurred.
```
scofalik, line 92:
```suggestion
* `onReady` &ndash; A function called when the editor is ready with a {@link module:editor-multi-root/multirooteditor~MultiRootEditor `MultiRootEditor`} instance. This callback is also called after the reinitialization of the component if an error occurred.
```

docs/installation/integrations/react-multiroot.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Context feature
+
+The `useMultiRootEditor` hook also supports the {@link features/context-and-collaboration-features context feature}, as described in the main {@link installation/integrations/react#context-feature React integration} guide.
+
+However, as the multi-root editor addresses most use cases of the context feature, consider if you need to employ it.
+
+## Two-way data binding
+
+By default, the two-way data binding is enabled. It means that every change done in the editor is automatically applied in the `data` object returned from `useMultiEditorHook`. Additionally, if you want to change or set data in the editor, you do not need to use the editor API, instead, the hook provides the `setData` method. It works the same way in case of attributes &ndash; the hook provides the `attributes` object and the `setAttributes` method to update them. It ensures that if you want to use or save the state of the editor, these objects are always up-to-date.
```
scofalik, line 126:
```suggestion
By default, the two-way data binding is enabled. It means that every change done in the editor is automatically applied in the `data` object returned from `useMultiEditorHook`. Additionally, if you want to change or set data in the editor, you can simply use `setData` method provided by the hook. It works the same way in case of attributes &ndash; the hook provides the `attributes` object and the `setAttributes` method to update them. It ensures that if you want to use or save the state of the editor, these objects are always up-to-date.
```

docs/installation/integrations/react-multiroot.md
DawidKossowski: I copied it from the React default integration docs, so I would stay with it for consistency's sake.

docs/installation/integrations/react-multiroot.md
```diff
@@ -0,0 +1,138 @@
...
+
+<info-box hint>
+	The multi-root editors in React is supported since version 6.2.0 of this package.
+
+	Unlike the {@link installation/integrations/react default integration}, in the multi-root editor we prepared the integration based on the hooks and new React mechanisms.
+</info-box>
+
+## Quick start
+
+This guide assumes you already have a React project. If you want to create a new one, you can use the [`create-react-app`](https://create-react-app.dev/) CLI. It allows you to create and customize your project with templates. For example, you can set up your project with TypeScript support.
```
Witoso, line 30: CRA is deprecated and discouraged right now, let's not treat it as a blocker but we will need to fix it.

Witoso (approved):

PR #16280, opened 2024-04-24
Add menu bar integration to multi root editor.

mremiszewski:

Feature (editor-multi-root): Add integration with menu bar.

---

### Additional information

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `MenuBarView` in the editor for enhanced navigation and focus handling.
	- Added keyboard shortcuts for focusing on the menu bar.
  
- **Enhancements**
	- Improved UI with updated CSS styling and layout adjustments for the menu and toolbar containers.

- **Tests**
	- Added and updated tests for new `MenuBarView` functionalities, including focus handling and UI integration.


DawidKossowski (commented):
Generally, it looks fine. However, I can see that a lot of logic is copied from the decoupled editor. I think we could, or even should, move the reusable parts, like `_initMenuBar`, to utils. Additionally, we will reduce the number of tests thanks to this.

packages/ckeditor5-editor-decoupled/src/decouplededitorui.ts
```diff
@@ -116,30 +116,7 @@ export default class DecoupledEditorUI extends EditorUI {
 	 * Initializes the editor menu bar.
 	 */
 	private _initMenuBar(): void {
```
DawidKossowski, line 118: This function is not necessary. Let's call `initMenuBar` directly.

packages/ckeditor5-editor-multi-root/src/multirooteditorui.ts
```diff
@@ -188,6 +190,13 @@ export default class MultiRootEditorUI extends EditorUI {
 		this.addToolbar( view.toolbar );
 	}
 
+	/**
+	 * Initializes the editor menu bar.
+	 */
+	private _initMenuBar(): void {
```
DawidKossowski, line 196: Same here.

packages/ckeditor5-ui/src/editorui/editorui.ts
```diff
@@ -649,3 +651,30 @@ function getToolbarDefinitionWeight( toolbarDef: FocusableToolbarDefinition ): n
 
 	return weight;
 }
+
+export function initMenuBar( editorUI: EditorUI & { view: EditorUIView & { menuBarView: MenuBarView } } ): void {
```
DawidKossowski, line 655: Please add docs.

DawidKossowski: @scofalik I was wondering if this is a good place for that function. I'd move it to `menubar/utils.js`. WDYT?

DawidKossowski (changes requested):

packages/ckeditor5-ui/src/editorui/editorui.ts
scofalik: At first glance, it feels to me that `menubar/utils.ts` is a better place as the function is about menu bar and editor UI is just a dependency here. Additionally, `editorui.ts` does not export any utils at this moment.

scofalik: But TBH, I don't see a reason why inline editor should not be able to have menu bar, and at that point we will probably move this into `EditorUI` class.

scofalik: I'd consider making it public-internal to avoid breaking changes. I don't believe there is a case where this util could be useful for integrators at this moment :thinking:.

scofalik:
Please use the util for `ClassicEditor` as well, AFAICS, the integration is same there as well.

packages/ckeditor5-editor-classic/src/classiceditorui.ts
```diff
@@ -11,11 +11,13 @@ import type { Editor, ElementApi } from 'ckeditor5/src/core.js';
 import {
 	EditorUI,
 	normalizeToolbarConfig,
-	normalizeMenuBarConfig,
 	DialogView,
 	type DialogViewMoveToEvent,
 	type Dialog,
-	type EditorUIReadyEvent
+	type EditorUIReadyEvent,
+	_initMenuBar,
```
scofalik, line 18: Don't mix `type` imports with normal imports, move it higher.

packages/ckeditor5-editor-classic/src/classiceditorui.ts
```diff
@@ -124,7 +118,11 @@ export default class ClassicEditorUI extends EditorUI {
 
 		this._initPlaceholder();
 		this._initToolbar();
-		this._initMenuBar();
+
+		if ( view.menuBarView ) {
+			_initMenuBar( this as EditorUI & { view: EditorUIView & { menuBarView: MenuBarView } } );
```
scofalik, line 123: Does this typing have to be this weird :grimacing:? Why these things are missing in `ClassicEditor`, can we do anything about it? It's ugly :D.

mremiszewski: `ClassicEditor` has `menuBarView?: MenuBarView` instead of `menuBarView: MenuBarView`, but I will try to improve this.

packages/ckeditor5-ui/src/menubar/utils.ts
```diff
@@ -1471,3 +1473,36 @@ function getIdFromGroupItem( item: string | MenuBarMenuDefinition ): string {
 function isMenuDefinition( definition: any ): definition is MenuBarMenuDefinition {
 	return typeof definition === 'object' && 'menuId' in definition;
 }
+
+/**
+ * Initializes menu bar in EditorUI.
```
scofalik, line 1478:
```suggestion
 * Initializes menu bar for given `editorUI`.
```

PR #16281, opened 2024-04-24
Add `onAfterDestroy` callback description to react docs

Mati365:

Docs: Add `onAfterDestroy` callback description to React integration docs.

### Why has been this callback introduced?

1. Editor uses async destroy method that does not work super fine with React lifecycle. 
2. Editor is most of the time destroyed _after_ unmounting of component or changing its ID property. 
3. This callback makes sure that all side effects (such as added editable HTML DOM nodes) are removed _after_ destroying all plugins attached to the editor. 
4. `onAfterDestroy` is being batched, so in other words rapid `id` change (10 in row) will trigger only ~1 call of this callback which prevents race conditions in `Strict Mode`. 

### Additional information

1. The `onAfterDestroy` callback has been added in this [commit](https://github.com/ckeditor/ckeditor5-react/commit/ce6cad06e7a847b48189ddfaeed57921ce314874#diff-17a78b4174b47153db11ac4738dfbb3608e1b36b02c7d70d86589722797e8515R181) in our React integration.
5. It's not released yet at this moment (20.05.2024).

### DIFF
--- a/docs/installation/integrations/react.md
+++ b/docs/installation/integrations/react.md
@@ -162,7 +162,7 @@ The `CKEditorContext` component supports the following properties:
 * `isLayoutReady` &ndash; A property that delays the context creation when set to `false`. It creates the context and the editor children once it is `true` or unset. Useful when the CKEditor&nbsp;5 annotations or a presence list are used.
 * `id` &ndash; The context ID. When this property changes, the component restarts the context with its editor and reinitializes it based on the current configuration.
 * `onReady` &ndash; A function called when the context is ready and all the editors inside were initialized with the `context` instance. This callback is also called after the reinitialization of the editor if an error has occurred.
-* `onAfterDestroy` &ndash; A function invoked after the successful destruction of an editor instance that was rendered by the component. This callback is also triggered after the editor has been reinitialized after an error. It is not guaranteed that the component will still be mounted when this function is called.
+* `onAfterDestroy` &ndash; A function called after the successful destruction of an editor instance rendered by the component. This callback is also triggered after the editor has been reinitialized after an error. The component is not guaranteed to be mounted when this function is called.
 * `onError` &ndash; A function called when the context has crashed during the initialization or during the runtime. It receives two arguments: the error instance and the error details.
 	Error details is an object that contains two properties:
 	* `{String} phase`: `'initialization'|'runtime'` &ndash; Informs when the error has occurred (during the editor or context initialization, or after the initialization).
@@ -231,6 +231,9 @@ class App extends Component {
 						if ( willEditorRestart ) {
 							this.editor.ui.view.toolbar.element.remove();
 						}
+					} }
+					onAfterDestroy={ editor => {
+						editor.ui.view.toolbar.element.remove();
 					} }
 					onChange={ ( event ) => console.log( event ) }
 					editor={ DecoupledEditor }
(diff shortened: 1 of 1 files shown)


Mati365:
Thank you @godai78. 

@pomek Can you merge it after release of React integration? 

fyi @scofalik

docs/installation/integrations/react.md
```diff
@@ -205,40 +206,40 @@ If you use the {@link framework/document-editor document (decoupled) editor}, yo
...
+						this.editor = editor;
+					} }
+					onError={ ( error, { willEditorRestart } ) => {
+						// If the editor is restarted, the toolbar element will be created once again.
+						// The `onReady` callback will be called again and the new toolbar will be added.
+						// This is why you need to remove the older toolbar.
+						if ( willEditorRestart ) {
+							this.editor.ui.view.toolbar.element.remove();
+						}
+					} }
```
pomek, line 234:
```suggestion
					onAfterDestroy={ editor => {
						editor.ui.view.toolbar.element.remove();
					} }
```

docs/installation/integrations/react.md
```diff
@@ -161,11 +161,12 @@ The `CKEditorContext` component supports the following properties:
...
 * `isLayoutReady` &ndash; A property that delays the context creation when set to `false`. It creates the context and the editor children once it is `true` or unset. Useful when the CKEditor&nbsp;5 annotations or a presence list are used.
 * `id` &ndash; The context ID. When this property changes, the component restarts the context with its editor and reinitializes it based on the current configuration.
-* `onReady` &ndash; A function called when the context is ready and all editors inside were initialized with the `context` instance. This callback is also called after the reinitialization of the component if an error has occurred.
+* `onReady` &ndash; A function called when the context is ready and all the editors inside were initialized with the `context` instance. This callback is also called after the reinitialization of the editor if an error has occurred.
+* `onAfterDestroy` &ndash; A function invoked after the successful destruction of an editor instance that was rendered by the component. This callback is also triggered after the editor has been reinitialized after an error. It is not guaranteed that the component will still be mounted when this function is called.
```
pomek, line 165:
```suggestion
* `onAfterDestroy` &ndash; A function called after the successful destruction of an editor instance rendered by the component. This callback is also triggered after the editor has been reinitialized after an error. The component is not guaranteed to be mounted when this function is called.
```

Minor updates:

* `s/invoked/called` - otherwise, we should update other callbacks too. 
* Simplifications.

pomek (commented):
Except for the mentioned things, please target the PR to the `#stable` branch if we aim to deploy these changes without waiting until the next release.

Mati365:
@pomek It can be deployed in normal flow I think. There is no need to deploy it instantly after merge because this callback is not critical to work with integration (especially because it was not present for years). Thanks for suggestion to docs anyway.

docs/installation/integrations/react.md
Mati365: I'm not sure if we should include this `onAfterDestroy` example with `editor.ui.view.toolbar.element.remove();`. It really depends on the way the users mounted `toolbar`. If it's somewhere outside of React box then it's fine but if it's in div that is managed by React then it will lead to crashes, probably. I'll leave there a comment that only external things should be removed here.

scofalik: For more context: this code sample is for Decoupled Editor. This editor is characteristic in the way that it is the integrator's responsibility to insert the toolbar where they want to.

scofalik:
LGTM, you can merge it. As far as I understand, neither the new callback nor the docs for the callback are going to be released in this release (which should happen very shortly)?

scofalik (approved):

pomek (approved):

pomek:
The required package has been released (https://github.com/ckeditor/ckeditor5-react/releases/tag/v7.0.0).

godai78:
We have ported these changes to new installation docs, which we plan on merging to `master` sometime soon, so this will be wvisible on nightly.

PR #16295, opened 2024-04-29
Name, role, value accessibility fixes

Mati365:

Fix (restricted-editing): Improved the accessibility of the restricted editing dropdown by setting the correct ARIA role on the list of menu items.

Fix (minimap): Addressed the issue of the minimap iframe being unnecessarily exposed to assistive technologies.   

---

### Additional information

Fixes https://github.com/cksource/ckeditor5-commercial/issues/6038
Commercial repo: https://github.com/cksource/ckeditor5-commercial/pull/6195

## Summary by CodeRabbit

- **New Features**
	- Enhanced accessibility in the editor by adding `aria-hidden` attribute to the minimap iframe.
	- Improved semantic structure in restricted editing mode by setting the dropdown button's role to 'menu'.

### DIFF
--- a/packages/ckeditor5-restricted-editing/src/restrictededitingmodeui.ts
+++ b/packages/ckeditor5-restricted-editing/src/restrictededitingmodeui.ts
@@ -52,6 +52,7 @@ export default class RestrictedEditingModeUI extends Plugin {
 			addListToDropdown( dropdownView, listItems );
 
 			dropdownView.buttonView.set( {
+				role: 'menu',
 				label: t( 'Navigate editable regions' ),
 				icon: lockIcon,
 				tooltip: true,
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-restricted-editing/src/restrictededitingmodeui.ts
```diff
@@ -52,6 +52,7 @@ export default class RestrictedEditingModeUI extends Plugin {
 			addListToDropdown( dropdownView, listItems );
 
 			dropdownView.buttonView.set( {
+				role: 'menu',
```
oleq, line 55: As discussed internally, for the `role=menu` to make sense, this should be set on the `ListView` instance instead. `addListToDropdown()` offers options configuration to set the role but can also be set manually.

oleq (changes requested):
I think `role=menu` should be propagated to other menus across the UI, no just the restricted editing menu. 

You can quickly find candidates by running `Array.from( document.querySelectorAll( '.ck-list' ) ).filter( element => element.role !== 'menu' ).map( el => el.textContent )` in a feature-rich manual test (I didn't refine them though).

Mati365:
@oleq I added `menu` role attribute to more plugins (except that ones that register dropdowns that are forms or more complex set of controls that list of buttons). Can you take a look again (both PRs)?

PR #16320, opened 2024-05-07
Add border to menu bar

mremiszewski:

Other (editor-classic): Add a separator between the menu bar and the toolbar.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-theme-lark/theme/ckeditor5-editor-classic/classiceditor.css
+++ b/packages/ckeditor5-theme-lark/theme/ckeditor5-editor-classic/classiceditor.css
@@ -20,7 +20,11 @@
 				border-bottom-width: 1px;
 			}
 
-			& .ck-menu-bar,
+			& .ck-menu-bar {
+				border: 0;
+				border-bottom: 1px solid var(--ck-color-base-border);
+			}
+
 			& .ck-toolbar {
 				border: 0;
 			}
--- a/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/toolbar/toolbar.css
+++ b/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/toolbar/toolbar.css
@@ -14,7 +14,7 @@
 	border: 1px solid var(--ck-color-toolbar-border);
 
 	& .ck.ck-toolbar__separator {
-		align-self: stretch;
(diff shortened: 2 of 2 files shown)


scofalik:
We should also try to make the separators look better.

Right now it is like this:



Separators + border makes it crowdy, it looks like a grid.

Proposal:



Separators are a bit smaller, there is a bit more breathing room.

I added `height` equal to `--ck-icon-size` and removed `align-self`:



@oleq -- WDYT?

oleq:
The change to separators looks fine but we need before/after screenshots for the following integrations to be sure it doesn't break anything:
* block toolbar, 
* balloon toolbar, 
* inline editor toolbar, 
* grouped items (`shouldGroupWhenFull`),
* widget toolbars

I think all of them can be easily obtained by building docs with all samples.

scofalik:
Adding screenshots of various builds, with various elements highlighted / focused to see how they look next to each other.

I also zipped them: [separators.zip](https://github.com/user-attachments/files/15537103/separators.zip).

AFAICS, image is the only widget with a toolbar that has separators. Let me know in case I missed something.

oleq:
> Adding screenshots of various builds, with various elements highlighted / focused to see how they look next to each other.

LGTM.

> AFAICS, image is the only widget with a toolbar that has separators. Let me know in case I missed something.

[Table toolbar](https://ckeditor.com/docs/ckeditor5/latest/api/module_table_tableconfig-TableConfig.html#member-tableToolbar) is also declarative. Any using `WidgetToolbarRepository` can also be like this.

BTW: Have you noticed the code snippet in `tableToolbar` API? 🙃

image

scofalik:
I guess they would look fine, if image toolbars look fine?

Reinmar:
Did you test a solution with those separators shorter only by 2px (1px top, 1px bottom)?

How much shorter are there in this PR? Because my initial feeling is "quite a bit" and I wonder if smaller change would not work better.

Though, nothing critically bad about the screenshots that you posted. I just wonder what other optionss do we have.

scofalik:
They are 6 pixels smaller, they are the height of the icon. I want to keep these separators scalable, so I based them on icon size, which is based on font size normal.

I can add some pixels using `calc()` though. But I think that smaller is better.

\-6px (screens above):



\-4px:



\-2px:



\-0px (current):

PR #16334, opened 2024-05-10
Add prototype submenu for image insert.

### LINKED ISSUE #16445: Add insert image integration for menu bar

Insert image component is not integrated with the menu bar as of now. The menu bar integration should behave in a similar way as the toolbar component.

Additionally, we should move the "Insert image via URL" form to a modal dialog.

mremiszewski:

Feature (image): Added menu bar integration for the insert image component. The `menuBar:insertImage` component is by default added to "Insert" menu and replaces current buttons related to image insertion. Closes #16445.

Fix (image): Buttons inside `insertImage` dropdown will no longer have an unnecessary tooltip.

Other (image): Following new components are now available: `insertImageViaUrl` toolbar button and `menuBar:insertImageViaUrl` menu bar button.

Other (image): "Insert image via URL" form has been move to a modal dialog instead of being available directly in the insert image dropdown.

MINOR BREAKING CHANGE (image): "Insert image via URL" form has been move to a modal dialog instead of being available directly in the insert image dropdown.


scofalik:
Quick functional review:

1.  Image insert options inside an "Image" submenu is okay.
2.  Add a proper icon to the menu bar "Image" item.
3.  We need to do something with "Image or file" item as it doesn't make sense / looks bad when there's also "Image" just above.
    1.  In toolbar, it's up to integrator which icon / dropdown option they will use. Nothing is set by default, so there's no obvious duplication.
    2.  In menu bar however, we propose default setup and we want it to be as complete as possible.
    3.  My proposal: change "Image or file" label to "File".
        1.  It should work fine with any file manager.
        2.  I think it will not be weird if you choose an image afterwards.
        3.  It will duplicate the functionality from "Image => Insert with file manager" but I am fine with this.
        4.  cc @Witoso @dagdzi 
4.  "Insert image via URL" (in menu bar) should not have a tooltip. No other menu item has a tooltip. I also think that the dropdown component should not have tooltips either, as the tooltips just duplicate the label information.
5.  "Insert image via URL" should look the same in the dropdown as in the menu bar. Convert the mini-form to the button that will open the modal dialog.
6.  Modal dialog size is incorrect (too narrow, with no paddings, I propose twice much size than now + some reasonable paddings).
7.  Labels in the menu bar dropdown should be: "Upload from computer", "Insert with file manager", "Insert via URL".
8.  I have doubts about the label changing from "Image" to "Replace image" but I'd like to get a second opinion @dagdzi @Witoso.

Ad 3:



Ad 4:



Ad 8:

Witoso:
> "Insert image via URL" should look the same in the dropdown as in the menu bar. Convert the mini-form to the button that will open the modal dialog.

Yep, good catch! I want this form gone :)  

> "Image or file" label to "File".

Which ui component is it? Shouldn't this be named `File manager`?

> I have doubts about the label changing from "Image" to "Replace image" but I'd like to get a second opinion [@dagdzi](https://github.com/dagdzi) [@Witoso](https://github.com/Witoso).

I agree, I think in main level in the menu bar we should keep nouns, and ideally don't change it to not confuse users.

scofalik:
This is file manager component, but it wouldn't follow a convention if we name it "File manager" in the menu. The menu is called "Insert" and each label in the menu is an entity that you insert, e.g. "Insert -> table", "Insert -> image", etc. "Insert -> file manager" does not follow that. Hence I proposed "File".

Witoso:
So maybe it doesn't make sense to keep it in Insert? File manager is also used for other things (CKBox), managing files, editing, upload is just one of the capabilities.

scofalik:
I think you should have a menu bar button that allows you to insert a file and is not hidden under "Insert -> Image -> Image with file manager". I am pretty positive.

We could also add "File manager" in "Tools -> File manager". But that would be a 3rd place where the same component is added :smile:. Also, I don't know if we can even open the file manager in a different mode than "insert something". I mean, you always have "Insert" button in the dialog. You don't open editor, and file manager from the editor, to organize your files, right? (Unless I miss something.)

Witoso:
Then let's go with the file.

packages/ckeditor5-image/src/imageinsert/imageinsertui.ts
```diff
@@ -111,13 +118,15 @@ export default class ImageInsertUI extends Plugin {
...
-		requiresForm
+		requiresForm,
+		menuBarButtonViewCreator
 	}: {
 		name: string;
 		observable: Observable & { isEnabled: boolean } | ( () => Observable & { isEnabled: boolean } );
 		buttonViewCreator: ( isOnlyOne: boolean ) => ButtonView;
 		formViewCreator: ( isOnlyOne: boolean ) => FocusableView;
 		requiresForm?: boolean;
+		menuBarButtonViewCreator: ( isOnlyOne: boolean ) => MenuBarMenuListItemButtonView;
```
scofalik, line 129: Move above the optional parameter for better clarity.

packages/ckeditor5-image/src/imageinsert/imageinsertui.ts
```diff
@@ -190,6 +200,46 @@ export default class ImageInsertUI extends Plugin {
 		return dropdownView;
 	}
 
+	/**
+	 * Creates the menu bar component.
+	 */
+	private _createMenuBarComponent( locale: Locale ): any {
```
scofalik, line 206: Let's not return `any`.

packages/ckeditor5-image/src/imageinsert/imageinsertviaurlui.ts
```diff
@@ -77,62 +85,30 @@ export default class ImageInsertViaUrlUI extends Plugin {
...
 
 	/**
 	 * Creates the toolbar button.
 	 */
-	private _createInsertUrlButton( isOnlyOne: boolean ): ButtonView {
-		const ButtonClass = isOnlyOne ? DropdownButtonView : ButtonView;
+	private _createInsertUrlButton<T extends typeof ButtonView | typeof MenuBarMenuListItemButtonView>(
+		ButtonClass: T
+	): InstanceType<T> {
+		// const ButtonClass = isOnlyOne ? DropdownButtonView : ButtonView;
```
scofalik, line 97: Remove.

packages/ckeditor5-image/src/imageinsert/imageinsertviaurlui.ts
```diff
@@ -77,62 +85,30 @@ export default class ImageInsertViaUrlUI extends Plugin {
...
 
 		const editor = this.editor;
-		const button = new ButtonClass( editor.locale );
+		const button = new ButtonClass( editor.locale ) as InstanceType<T>;
 		const t = editor.locale.t;
 
+		button.set( {
+			label: t( 'Set image url' ),
+			icon: icons.imageUpload
+		} );
```
scofalik, line 106: Clean this up. We use `.set()` below as well, which overwrites the icon. Also, the label does not seem correct (and we bind it below, but that should be kept only for toolbar).

packages/ckeditor5-image/src/imageinsert/imageinsertviaurlui.ts
```diff
@@ -147,10 +123,48 @@ export default class ImageInsertViaUrlUI extends Plugin {
...
+			const replaceImageSourceCommand: ReplaceImageSourceCommand = editor.commands.get( 'replaceImageSource' )!;
+
+			if ( replaceImageSourceCommand.isEnabled ) {
+				editor.execute( 'replaceImageSource', { source: form.imageURLInputValue } );
+			} else {
+				editor.execute( 'insertImage', { source: form.imageURLInputValue } );
+			}
+
+			dialog.hide();
+		}
```
scofalik, line 146: Could we extract this to a private method of this class?

packages/ckeditor5-image/src/imageinsert/imageinsertviaurlui.ts
```diff
@@ -147,10 +123,48 @@ export default class ImageInsertViaUrlUI extends Plugin {
...
+		const locale = editor.locale;
+		const t = locale.t;
+		const dialog = editor.plugins.get( 'Dialog' );
+
+		const form = this._createInsertUrlView();
+
+		function handleSave( form: ImageInsertUrlView ) {
+			const replaceImageSourceCommand: ReplaceImageSourceCommand = editor.commands.get( 'replaceImageSource' )!;
+
+			if ( replaceImageSourceCommand.isEnabled ) {
```
scofalik, line 139: I know you copy-pasted this part but it demands some comment because it is confusing.
```suggestion
			// If an image element is currently selected, we want to replace its source attribute (instead of inserting a new image).
			// We detect if an image is selected by checking `replaceImageSource` command state.
			if ( replaceImageSourceCommand.isEnabled ) {
```

packages/ckeditor5-image/src/imageinsert/imageinsertviaurlui.ts
```diff
@@ -40,6 +40,15 @@ export default class ImageInsertViaUrlUI extends Plugin {
 		return [ ImageInsertUI ] as const;
 	}
 
+	public init(): void {
+		this.editor.ui.componentFactory.add( 'menuBar:uploadUrl', locale => {
```
scofalik, line 44: I see you added this as a component but AFAICS, it is not used anywhere. So I think it should be used by `menuBarButtonViewCreator` right?

Additionally, there was a request from the team working on the builder to extract the toolbar button as a component as well, so it can be added directly to the toolbar. All image insert methods should have all corresponding components / integrations:

* toolbar button
* menu bar item
* integration with toolbar image insert dropdown
* integration with menu bar image insert menu

scofalik (changes requested):
Additionally go through all the files in the image package related to UI / inserting images and check if all docs are up-to-date.

E.g. `ImageInsertUI` has this description:

```
Adds the `'insertImage'` dropdown to the {@link module:ui/componentfactory~ComponentFactory UI component factory}
and also the `imageInsert` dropdown as an alias for backward compatibility.
```

Which doesn't even mention anything about menu bar.

packages/ckeditor5-image/src/imageinsert/imageinsertui.ts
```diff
@@ -36,6 +42,8 @@ import ImageUtils from '../imageutils.js';
  *
  * Adds the `'insertImage'` dropdown to the {@link module:ui/componentfactory~ComponentFactory UI component factory}
  * and also the `imageInsert` dropdown as an alias for backward compatibility.
+ *
+ * Adds the `'menuBar:insertImage'` sub-menu to the {@link module:ui/componentfactory~ComponentFactory UI component factory}.
```
scofalik, line 46: We also added new components in ckbox and ckfinder and AFAICS their API docs are not updated as well.

I also wanted you to go through existing guides but since core team is now reworking them AND we need to do other tasks related to docs (update demos, update all other guides not only images), then let's skip it for now.

scofalik:
Some missed things:

> I also think that the dropdown component should not have tooltips either, as the tooltips just duplicate the label information.

Please remove tooltips from the items in the toolbar dropdown as they are simply repeating the button label and are annoying.

Regarding labels and tooltips, as it seems that it got confusing:

1.  Menu bar:
    1.  Upload from computer
    2.  Insert with file manager
    3.  Insert via URL
2.  Toolbar dropdown (formview integration):
    1.  Upload/Replace from computer
    2.  Insert/Replace with file manager
    3.  Insert via URL / Update image URL
3.  Toolbar button (button integration):
    1.  Upload/Replace image from computer
    2.  Insert/Replace image with file manager
    3.  Insert image via URL / Update image URL

And regarding tooltips, again: only toolbar buttons (and all of them) should have tooltips.

packages/ckeditor5-image/src/imageinsert/imageinsertviaurlui.ts
```diff
@@ -46,28 +70,35 @@ export default class ImageInsertViaUrlUI extends Plugin {
...
+		};
+
 		this._imageInsertUI.registerIntegration( {
 			name: 'url',
 			observable: () => this.editor.commands.get( 'insertImage' )!,
-			requiresForm: true,
-			buttonViewCreator: isOnlyOne => this._createInsertUrlButton( isOnlyOne ),
-			formViewCreator: isOnlyOne => this._createInsertUrlView( isOnlyOne )
+			requiresForm: false,
+			buttonViewCreator,
```
scofalik, line 85: This is probably incorrect :thinking:. If insert via URL is the only available option, then it will be displayed in the toolbar with text, right? Which is wrong.

scofalik: Also, I just realized that `uploadUrl` is rather a bad name. First question: upload what? Second: we don't upload anything really.

The other components we have are called:

* `insertImage`
* `uploadImage`

`urlImage` sounds quite bad so I propose `insertImageUrl` (and ofc `menuBar:insertImageUrl`).

scofalik: Also please add info about this component to metadata file.

godai78:
Please be advised that this branch produces a ton of broken links while building locally. Make sure to update with the NIM docs changes from the current \`master\` base branch to not overwrite anything.

godai78 (approved):

scofalik:
I haven't got any errors locally and CI also was green. I also recently (yesterday or Wed?) merged with master. I just merged with master again and run docs and all seems fine.

PR #16358, opened 2024-05-15
Fix link dataset e2e crash

Mati365:

Internal (ui): Safety checks that prevent errors when tooltip data is accessed on a null element.

Closes https://github.com/cksource/ckeditor5-commercial/issues/6219

### DIFF
--- a/packages/ckeditor5-ui/src/tooltipmanager.ts
+++ b/packages/ckeditor5-ui/src/tooltipmanager.ts
@@ -467,13 +467,14 @@ export default class TooltipManager extends DomEmitterMixin() {
 	 * Hides the tooltip when the element is no longer visible in DOM or the tooltip text was removed.
 	 */
 	private _updateTooltipPosition() {
-		// Actually the #_currentElementWithTooltip member can be from time to time null. It can happen in this scenario:
+		// Actually, the #_currentElementWithTooltip member can occasionally be null. This can happen in the following scenario:
 		//
-		// 1. User opens ballon editor in content editable.
-		// 2. User clicks "Remove link" button which is attached to such ballon and link in editor.
-		// 3. Editor removed link ( and attached ballon ) from editable.
-		// 4. Unmounted ballon fires `update` (for example from form validation) event  because it's `destroy` method is not called.
-		// 5. We received `update` event and #_currentElementWithTooltip is null.
+		// 1. The user opens the balloon editor in a content editable.
+		// 2. The user clicks the "Remove link" button, which is attached to the balloon and the link in the editor.
+		// 3. The editor removes the link (and the attached balloon) from the editable.
+		// 4. The unmounted balloon fires an `update` event (for example, from form validation) because its `destroy` method is not called
+		//    and various listeners in ballon are still active. .
+		// 5. We receive the `update` event, and #_currentElementWithTooltip is null.
 		if ( !this._currentElementWithTooltip ) {
 			return;
 		}
--- a/packages/ckeditor5-ui/src/tooltipmanager.ts
+++ b/packages/ckeditor5-ui/src/tooltipmanager.ts
@@ -467,21 +467,7 @@ export default class TooltipManager extends DomEmitterMixin() {
 	 * Hides the tooltip when the element is no longer visible in DOM or the tooltip text was removed.
 	 */
 	private _updateTooltipPosition() {
-		// Actually, the #_currentElementWithTooltip member can occasionally be null. This can happen in the following scenario:
-		//
-		// 1. The user opens the balloon editor in a content editable.
-		// 2. The user clicks the "Remove link" button, which is attached to the balloon and the link in the editor.
-		// 3. The editor removes the link (and the attached balloon) from the editable.
-		// 4. The unmounted balloon fires an `update` event (for example, from form `blur` event).
-		// 5. We receive the `update` event, and #_currentElementWithTooltip is null.
-		// 6. Ballon fully unmounts and unbinds all `blur` listeners.
-		//
-		// Tooltip must be attached to balloon content in this scenario.
-		if ( !this._currentElementWithTooltip ) {
-			return;
-		}
-
-		const tooltipData = getTooltipData( this._currentElementWithTooltip );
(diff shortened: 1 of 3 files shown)


Reinmar:
Why is this change needed?
In what situations does it occur and why do we get to a state where we need that check?
Why did it work previously and when exactly was the issue introduced?

I can't see "what" but I miss all the "why"s.

Reinmar (changes requested):
More context needed.

Reinmar:
PS. One last nitpicking:



The commit convention is to use full sentence with a period at the end.

Also, "Fix links dataset" doesn't explain anything and isn't semantically relevant to the files in which the change was made. The change was made in the UI package. What do link have to do with it?

The commit message could be (hyphoteltically) something like: "Prevented a case when X happens after Y (e.g. triggered by Z)."

Mati365:
@Reinmar 

I discussed it with Dawid and Szymon internally, and I thought that it is clear that it's not trivial one issue and has to be fixed. QA confirmed that issue does not happen in release branches, so I delayed investigation (but as we can see it is). It is not a bug in actual production code, and it was introduced more likely in this sprint release. Most likely here: https://github.com/ckeditor/ckeditor5/commit/caea11e431fc56f911f4cf4ad4f73bc74d36a8b9#diff-2f067c74b61fa304de5d8501f56c29bd076ef2f0606d0b69778333bdba0bb6f3R219-R223. It fixes the position of balloon position when error occurs. That event seems to be async and seems to be fired after removal of whole form (maybe because there is blur effect during click of close button) and it's not working properly with tooltip logic. 

The main problem is that `update` event seems to be handed in these functions, besides that it is already unmounted.

https://github.com/ckeditor/ckeditor5/blob/f0463ddd0bb0696b78db8759114fd8dc2bd65870/packages/ckeditor5-ui/src/tooltipmanager.ts#L395

https://github.com/ckeditor/ckeditor5/blob/f0463ddd0bb0696b78db8759114fd8dc2bd65870/packages/ckeditor5-ui/src/tooltipmanager.ts#L443C10-L443C23

I suspect that it is related to async nature of such event.

Mati365:
@Reinmar

> Internal (ui): Safety checks that prevent errors when tooltip data is accessed on a null element. I can see the original issue https://github.com/cksource/ckeditor5-commercial/issues/6219 but it sounds like a bug in > the product. Not an internal change. Another thing that I miss is explanation what exact conditions need to be met for this to occur. This issue was reported in the commercial repo because there seemed to be some commercial features involved. However, when the fix lands in the OSS repo apparently the bug is actually in the OSS code. So, what's the bug in the OSS code? That deserves a ticket here.

It's not related to this PR. Please report that in issue and inform about that person who created that issue.

Mati365:
@Reinmar 

> The commit message could be (hyphoteltically) something like: "Prevented a case when X happens after Y (e.g. triggered by Z)."

During the rebasing process, the PR description is used as a commit message. From what I remember, it was discussed that all PRs are squashed before merge. What's the point of setting proper commit messages if they will be erased anyway?

Mati365:
@Reinmar I added small explanation how to reproduce this issue in null check in tooltip data getter. Besides that simple reorder of `unlink` and `hideUI` calls solves the issue, but I think it's better to keep this assertion because it's super easy to break it again in other integrations.

packages/ckeditor5-ui/src/tooltipmanager.ts
```diff
@@ -467,7 +467,21 @@ export default class TooltipManager extends DomEmitterMixin() {
...
 	private _updateTooltipPosition() {
-		const tooltipData = getTooltipData( this._currentElementWithTooltip! );
+		// Actually, the #_currentElementWithTooltip member can occasionally be null. This can happen in the following scenario:
+		//
+		// 1. The user opens the balloon editor in a content editable.
+		// 2. The user clicks the "Remove link" button, which is attached to the balloon and the link in the editor.
+		// 3. The editor removes the link (and the attached balloon) from the editable.
+		// 4. The unmounted balloon fires an `update` event (for example, from form `blur` event).
+		// 5. We receive the `update` event, and #_currentElementWithTooltip is null.
+		// 6. Ballon fully unmounts and unbinds all `blur` listeners.
```
DawidKossowski, line 477: I think those steps do not describe the real issue and are incorrect.

> 1. The user opens the balloon editor in a content editable.

It doesn't have to be a balloon editor, so this is an incorrect statement. Perhaps it was meant to refer to a balloon panel. In the next point, there is a reference to "Remove link", so it might be worth adding information that the example is a link balloon.

> 2. The user clicks the "Remove link" button, which is attached to the balloon and the link in the editor.

In reproducing the issue, it is crucial that the tooltip is attached to the button. This information is missing here, and it is found at the end of the comment.

> 3. The editor removes the link (and the attached balloon) from the editable.

This is also unclear - there is no information about what balloon content actually is. In the issue I tested, it specifically concerns a button in the balloon.


I'm attaching the video with the reproduction steps below:

https://github.com/ckeditor/ckeditor5/assets/20561161/83128dd5-2c03-4333-b45e-bf023cedb871

DawidKossowski:
@Mati365, Could you explain to me why this issue occurs only in the link balloon panel and we are not able to reproduce it anywhere else? I believe we have a very similar situation with the inline annotations (track changes / comments), but it works:

https://github.com/ckeditor/ckeditor5/assets/20561161/c7b19bf0-02c4-41f4-aefe-5d636efecd11

Mati365:
@DawidKossowski As I described in comment - balloon `unlink` action is removing underlying `link` editing node which has attached balloon popup to `a` tag inside editing view. It crashes because modification of editing view (by removing link) triggers validation in linkui input field which triggers `update` event (and it should because validation error messages affect ballon position). `update` event is later caught in tooltip coordination update handler https://github.com/ckeditor/ckeditor5/blob/b1415c611d566aeca30b4cd7d9304bd3f995d269/packages/ckeditor5-link/src/linkui.ts#L471 , that element was `a` anchor described at the start.

niegowski:
Fixed in https://github.com/ckeditor/ckeditor5/pull/16363.

PR #16362, opened 2024-05-15
Add typing for getting data with merge tags data

Dumluregn:

Internal (merge-tags): Add typing for getting data with merge tags data.

---


packages/ckeditor5-engine/src/controller/datacontroller.ts
```diff
@@ -190,6 +190,7 @@ export default class DataController extends /* #__PURE__ */ EmitterMixin() {
 		options: {
 			rootName?: string;
 			trim?: 'empty' | 'none';
+			mergeTagsData?: Record<string, string | ( () => string ) >;
```
niegowski, line 193: Same comment as for `Editor#getData()`.

packages/ckeditor5-core/src/editor/editor.ts
```diff
@@ -533,6 +533,7 @@ export default abstract class Editor extends /* #__PURE__ */ ObservableMixin() {
 	public getData( options?: {
 		rootName?: string;
 		trim?: 'empty' | 'none';
+		mergeTagsData?: Record<string, string | ( () => string )>;
```
niegowski, line 536: This is not needed here. For such cases, there is `[ key: string ]: unknown;` below.

Dumluregn:
Turns out this PR is not necessary.

PR #16363, opened 2024-05-15
The TooltipManager should verify if the tooltip is not already unpinned while trying to update its position.

niegowski:

Fix (ui): The `TooltipManager` should verify if the tooltip is not already unpinned while trying to update its position. 

Closes https://github.com/cksource/ckeditor5-commercial/issues/6219.

---

### Additional information

This is fixing the following flow:
- Unlink command executed => Model change => View change => View render
- View on render event => fire `layoutChanged` event
- EditorUI on `layoutChanged` event => fire UI `update` event:
1.
    - `LinkUI` on UI `update` event => no link => `_hideUI()` => call `EditingView` `focus()`
    - `TooltipManager` on `focus` (editable) event => no `getDescendantWithTooltip()` => call `_unpinTooltip()`
    - Note that it calls `stopListening( editor.ui, 'update' )`
2.
    - This is the same `update` event as above, the stop listening is affecting only future events, not the current one.
    - `TooltipManager` on UI `update` event => call `_updateTooltipPosition()` for null element

It is safe to check if the tooltip was not already unpinned (`_currentElementWithTooltip` is cleared).

This is fixing a bug introduced in v41.2.0.

### DIFF
--- a/packages/ckeditor5-ui/src/tooltipmanager.ts
+++ b/packages/ckeditor5-ui/src/tooltipmanager.ts
@@ -473,7 +473,7 @@ export default class TooltipManager extends DomEmitterMixin() {
 			return;
 		}
 
-		const tooltipData = getTooltipData( this._currentElementWithTooltip! );
+		const tooltipData = getTooltipData( this._currentElementWithTooltip );
 
 		// This could happen if the tooltip was attached somewhere in a contextual content toolbar and the toolbar
 		// disappeared (e.g. removed an image), or the tooltip text was removed.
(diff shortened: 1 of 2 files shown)


scofalik (approved):

niegowski:
Looks like in this [commit](https://github.com/ckeditor/ckeditor5/commit/039b302d9cbbda37e83a9dcaf4c6dead47e02194) we ignored the possibility that the target element is no longer there. The `isVisible()` helper checks for null.

Reinmar:
> This is fixing a bug introduced in v41.2.0.

We were discussing this outside GH and the hypothesis is that we didn't catch it until very recently, because some recent change made the tooltips are shown with a smaller delay now. But it would require further analysis to verify this.

PR #16370, opened 2024-05-16
Prevented error thrown when `ckeditor5-utils/src/env.ts` is used in an environment without window global object

### LINKED ISSUE #16368: window.matchMedia is not a function errors after ugprade to 41.4.1

1. Run CKEditor in a CLI Environment without access to `window` like Jest (in our Jest 24)

### ✔️ Expected result

### ❌ Actual result

```
 TypeError: window.matchMedia is not a function

      2 | import React from 'react';
      3 | import log from 'loglevel';
    > 4 | import AlignmentPlugin from '@ckeditor/ckeditor5-alignment/src/alignment';
        | ^
      5 | import BoldPlugin from '@ckeditor/ckeditor5-basic-styles/src/bold';
      6 | import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor';
      7 | import EssentialsPlugin from '@ckeditor/ckeditor5-essentials/src/essentials';

      at isMediaForcedColors (node_modules/@ckeditor/ckeditor5-utils/src/env.js:131:19)
      at Object.<anonymous> (node_modules/@ckeditor/ckeditor5-utils/src/env.js:34:26)
      at Object.<anonymous> (node_modules/@ckeditor/ckeditor5-utils/src/index.js:8:1)
      at Object.<anonymous> (node_modules/@ckeditor/ckeditor5-core/src/plugin.js:9:1)
      at Object.<anonymous> (node_modules/@ckeditor/ckeditor5-core/src/index.js:8:1)
      at Object.<anonymous> (node_modules/ckeditor5/src/core.js:8:1)
      at Object.<anonymous> (node_modules/@ckeditor/ckeditor5-alignment/src/alignment.js:8:1)
      at Object.<anonymous> (src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/CKEditor5.js:4:1)
      at Object.<anonymous> (src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/index.js:2:1)
      at Object.<anonymous> (src/Sulu/Bundle/AdminBundle/Resources/js/containers/index.js:2:1)
      at Object.<anonymous> (src/Sulu/Bundle/PageBundle/Resources/js/containers/TeaserSelection/tests/TeaserSelection.test.js:5:1)
```

### ❓ Possible solution

For server side renderers and cli best rendering it would be best if the `window.matchMedia` is only access when windowvariable is available.

## 📃 Other details

* Browser: none
* OS: Mac and Linux
* First affected CKEditor version: think 41.4.1 (CI did run yesterday but today not)

---

scofalik:

Fix (utils): Prevented error thrown when `ckeditor5-utils/src/env.ts` is used in an environment without `window` global object. Closes #16368.

### DIFF
--- a/packages/ckeditor5-utils/src/env.ts
+++ b/packages/ckeditor5-utils/src/env.ts
@@ -9,11 +9,13 @@
  * @module utils/env
  */
 
+import globalVar from './dom/global';
+
 /**
  * Safely returns `userAgent` from browser's navigator API in a lower case.
  * If navigator API is not available it will return an empty string.
  */
-export function getUserAgent( ): string {
+export function getUserAgent(): string {
 	// In some environments navigator API might not be available.
 	try {
 		return navigator.userAgent.toLowerCase();
@@ -109,7 +111,9 @@ const env: EnvType = {
 
 	isBlink: isBlink( userAgent ),
 
-	isMediaForcedColors: isMediaForcedColors(),
+	get isMediaForcedColors() {
+		return isMediaForcedColors();
+	},
 
 	get isMotionReduced() {
 		return isMotionReduced();
@@ -218,14 +222,18 @@ export function isRegExpUnicodePropertySupported(): boolean {
 
 /**
  * Checks if the user agent has enabled a forced colors mode (e.g. Windows High Contrast mode).
+ *
+ * Returns `false` in environments where `window` global object is not available.
  */
 export function isMediaForcedColors(): boolean {
-	return window.matchMedia( '(forced-colors: active)' ).matches;
+	return globalVar.window.matchMedia ? globalVar.window.matchMedia( '(forced-colors: active)' ).matches : false;
 }
 
 /**
  * Checks if user enabled "prefers reduced motion" setting in browser.
+ *
+ * Returns `false` in environments where `window` global object is not available.
  */
 export function isMotionReduced(): boolean {
-	return window.matchMedia( '(prefers-reduced-motion)' ).matches;
+	return globalVar.window.matchMedia ? globalVar.window.matchMedia( '(prefers-reduced-motion)' ).matches : false;
 }
--- a/packages/ckeditor5-utils/src/env.ts
+++ b/packages/ckeditor5-utils/src/env.ts
@@ -9,7 +9,7 @@
  * @module utils/env
  */
 
-import globalVar from './dom/global.js';
+import global from './dom/global.js';
 
 /**
  * Safely returns `userAgent` from browser's navigator API in a lower case.
@@ -226,7 +226,7 @@ export function isRegExpUnicodePropertySupported(): boolean {
  * Returns `false` in environments where `window` global object is not available.
  */
 export function isMediaForcedColors(): boolean {
-	return globalVar.window.matchMedia ? globalVar.window.matchMedia( '(forced-colors: active)' ).matches : false;
+	return global.window.matchMedia ? global.window.matchMedia( '(forced-colors: active)' ).matches : false;
 }
 
 /**
@@ -235,5 +235,5 @@ export function isMediaForcedColors(): boolean {
  * Returns `false` in environments where `window` global object is not available.
  */
 export function isMotionReduced(): boolean {
-	return globalVar.window.matchMedia ? globalVar.window.matchMedia( '(prefers-reduced-motion)' ).matches : false;
+	return global.window.matchMedia ? global.window.matchMedia( '(prefers-reduced-motion)' ).matches : false;
 }
--- a/packages/ckeditor5-utils/src/env.ts
+++ b/packages/ckeditor5-utils/src/env.ts
@@ -9,7 +9,7 @@
  * @module utils/env
  */
 
-import globalVar from './dom/global';
+import globalVar from './dom/global.js';
 
 /**
  * Safely returns `userAgent` from browser's navigator API in a lower case.
--- a/packages/ckeditor5-utils/tests/env.js
+++ b/packages/ckeditor5-utils/tests/env.js
@@ -4,7 +4,8 @@
  */
 
 import env, {
-	isMac, isWindows, isGecko, isSafari, isiOS, isAndroid, isRegExpUnicodePropertySupported, isBlink, getUserAgent, isMediaForcedColors
+	isMac, isWindows, isGecko, isSafari, isiOS, isAndroid, isRegExpUnicodePropertySupported, isBlink, getUserAgent,
+	isMediaForcedColors, isMotionReduced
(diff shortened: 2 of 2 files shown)


packages/ckeditor5-utils/src/env.ts
```diff
@@ -109,7 +111,9 @@ const env: EnvType = {
 
 	isBlink: isBlink( userAgent ),
 
-	isMediaForcedColors: isMediaForcedColors(),
+	get isMediaForcedColors() {
+		return isMediaForcedColors();
+	},
```
scofalik, line 116: I first wanted to get rid of these getters and make it so that the property is resolved immediately (just like all other `is*` properties above). However, the existing tests were failing and it appeared that we test whether the util reacts to the change of related settings.

Indeed, this setting can change (tested that), so I decided to keep these functions as getters.

packages/ckeditor5-utils/src/env.ts
```diff
@@ -218,14 +222,18 @@ export function isRegExpUnicodePropertySupported(): boolean {
...
 }
 
 /**
  * Checks if user enabled "prefers reduced motion" setting in browser.
+ *
+ * Returns `false` in environments where `window` global object is not available.
  */
 export function isMotionReduced(): boolean {
-	return window.matchMedia( '(prefers-reduced-motion)' ).matches;
+	return globalVar.window.matchMedia ? globalVar.window.matchMedia( '(prefers-reduced-motion)' ).matches : false;
```
scofalik, line 238: Because it was a getter, this function didn't crash in described scenario because the getter was never called. However, the getter would still crash if called in an environment without `window`. To make the solution a bit more future-proof, I decided to add a check for `globalVar.window`.

packages/ckeditor5-utils/tests/env.js
```diff
@@ -326,6 +351,44 @@ describe( 'Env', () => {
...
+		it( 'returns false if the document media query does not match prefers-reduced-motion', () => {
+			testUtils.sinon.stub( global.window, 'matchMedia' )
+				.withArgs( '(prefers-reduced-motion)' )
+				.returns( { matches: false } );
+
+			expect( isMotionReduced() ).to.be.false;
+		} );
+
+		it( 'returns false if window object is not available', () => {
+			// `global.window` is an empty object if `window` was not available in global space.
```
scofalik, line 384: If you have a better idea how to mock, please suggest.

packages/ckeditor5-utils/src/env.ts
```diff
@@ -9,11 +9,13 @@
  * @module utils/env
  */
 
+import globalVar from './dom/global.js';
```
DawidKossowski, line 12:
```suggestion
import global from './dom/global.js';
```

I think we always use this convention.

packages/ckeditor5-utils/src/env.ts
DawidKossowski: I believe we should mock all methods and fields in `globalVar.window` that we utilize in our codebase. Centralizing them would eliminate the need to repeat this logic in multiple places. However, perhaps this task could be deferred until later.

packages/ckeditor5-utils/src/env.ts
scofalik: Fixed.

scofalik:
> I believe we should mock all methods and fields in `globalVar.window` that we utilize in our codebase. Centralizing them would eliminate the need to repeat this logic in multiple places. However, perhaps this task could be deferred until later.

That would be a bit overkill and is not necessary right now. We are not able to get rid of DOM dependencies. You are basically asking about re-attaching whole UI and whole conversion from the DOM. That's an enormous, multi-level task (that could be titled "allow CKE5 to be run in Node").

I think it is enough to safeguard these functions that could be mistakenly used when they shouldn't.

packages/ckeditor5-utils/tests/env.js
```diff
@@ -326,6 +351,44 @@ describe( 'Env', () => {
...
+			const _window = global.window;
+			global.window = {};
+
+			expect( isMediaForcedColors() ).to.be.false;
+
+			global.window = _window;
+		} );
+	} );
+
+	describe( 'isMotionReduced()', () => {
```
Mati365, line 366: It is already tested above

scofalik: The above test checks `env.isMotionReduced`. I added additional tests for the util to match the other getter. I guess no harm in them. Also, I added `window = {}` test here.

DawidKossowski:
@scofalik, it seems like we didn't quite understand each other in that comment. Let's try again.

I don't want to replace the entire DOM dependencies, but only provide those mocks when `window` doesn't exist. So, I only meant that we have this logic:
```
try {
	globalVar = { window, document };
} catch ( e ) {
	globalVar = { window: {} as any, document: {} as any };
}
```

Instead of returning an empty object in that catch block, we could return something like this:
```
{
    window: {
         matchMedia: () => false,
         ...
    }
}
```

This would reduce the complexity of the logic you wrote to check if that function exists.

scofalik:
I understood but went too far with conclusion. Well, that could have worked but we have a lot of `window.` calls to many methods and properties. I guess some of them return instances of other DOM classes, and we go deeper and deeper. That's what I meant in my conclusion.

packages/ckeditor5-utils/src/env.ts
Reinmar: Doh... so maybe this should be a function? To indicate that it's a dynamic value?

packages/ckeditor5-utils/tests/env.js
Reinmar: The only thing I can think of is the use of before/after() so it's properly cleaned up on a fail, but it's a simple test so I don't see this as a must-have.

packages/ckeditor5-utils/src/env.ts
Reinmar: All the other properties of the `env` object are static (should not change with time) except the two recently introduced. So it'd make sense to think twice before we commit in any direction.

Reinmar:
Doh, GitHub lagged to refresh the thread for me and I started commenting in a thread:

[https://github.com/ckeditor/ckeditor5/pull/16370/files#r1603635378](https://github.com/ckeditor/ckeditor5/pull/16370/files#r1603635378)

I can see 3 options:

*   Keep those props as they are now if you don't agree with me that they should be functions.
*   Document them as private for now to change them in the upcoming version, without the need to declare this as a breaking change.
*   Change this ASAP (for v41.4.2) and treat this as a non-breaking-change since they were recently introduced.

PR #16402, opened 2024-05-22
Docs: updating error codes.

godai78:

Docs: updating error codes.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-upload/src/filerepository.ts
+++ b/packages/ckeditor5-upload/src/filerepository.ts
@@ -148,17 +148,10 @@ export default class FileRepository extends Plugin {
 			 * This warning shows up when {@link module:upload/filerepository~FileRepository} is being used
 			 * without {@link module:upload/filerepository~FileRepository#createUploadAdapter defining an upload adapter}.
 			 *
-			 * **If you see this warning when using one of the {@glink getting-started/legacy/installation-methods/predefined-builds
-			 * CKEditor 5 Builds}**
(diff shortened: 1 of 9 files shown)


packages/ckeditor5-upload/src/filerepository.ts
```diff
@@ -148,17 +148,10 @@ export default class FileRepository extends Plugin {
...
-			 * See the {@glink features/images/image-upload/image-upload comprehensive "Image upload overview"} to learn which upload
-			 * adapters are available in the builds and how to configure them.
-			 *
-			 * **If you see this warning when using a custom build** there is a chance that you enabled
-			 * a feature like {@link module:image/imageupload~ImageUpload},
-			 * or {@link module:image/imageupload/imageuploadui~ImageUploadUI} but you did not enable any upload adapter.
-			 * You can choose one of the existing upload adapters listed in the
+			 * If you see this warning there is a chance that you enabled a feature like
+			 * {@link module:image/imageupload~ImageUpload}, or {@link module:image/imageupload/imageuploadui~ImageUploadUI}
+			 * but you did not enable any upload adapter. You can choose one of the existing upload
```
Witoso, line 153: Why did we delete the predefined mention? I would go the same way we went with previous messages. It's still top 2 error which could fuel migration :D

Witoso (changes requested):

packages/ckeditor5-upload/src/filerepository.ts
godai78: OK

PR #16408, opened 2024-05-23
Block toolbar's handle tooltip misaligned

Mati365:

Fix (ui): Block toolbar's handle tooltip misaligned.

---

### Additional information

Setting balloon panel view class before pinning caused misalignment of tooltip. It looks like `BALLOON_CLASS` and `cssClass` were affecting width of tooltip _after_ pinning it (probably due to word break rules).


ISilviu:
Could you estimate when will this be included in a new build?

scofalik:
It will be included in the next release, which should happen in the second part of June.

ISilviu:
Hi, are we close to the next release? 🚀

Mati365:
@ISilviu It was released in `42.0.0` version.

PR #16415, opened 2024-05-24
The `EditorWatchdog`, `ContextWatchdog`, and `Context` classes are now exposed as static fields of the `Editor` class.

### LINKED ISSUE #13852: Expose Watchdog and Context on Editor class and update integrations.
Notes: 
* We exposed Watchdog on the editor class in order to remove that specific dependency on the package. See #13707.
* The Angular integration still has cke5-watchdog in its dependencies and imports it directly. 
* Most likely, the same changes would make now sense in the React integration too.
* There's also ckeditor/ckeditor5-angular#364, but to my knowledge seems unrelated.

niegowski:

Other (core): The `EditorWatchdog`, `ContextWatchdog`, and `Context` classes are now exposed as static fields of the `Editor` class. Closes #13852.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-core/package.json
+++ b/packages/ckeditor5-core/package.json
@@ -26,6 +26,7 @@
   "dependencies": {
     "@ckeditor/ckeditor5-engine": "41.4.2",
     "@ckeditor/ckeditor5-utils": "41.4.2",
+    "@ckeditor/ckeditor5-watchdog": "41.4.2",
     "lodash-es": "4.17.21"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


packages/ckeditor5-core/package.json
```diff
@@ -26,6 +26,7 @@
   "dependencies": {
     "@ckeditor/ckeditor5-engine": "41.4.2",
     "@ckeditor/ckeditor5-utils": "41.4.2",
+    "@ckeditor/ckeditor5-watchdog": "41.4.2",
```
filipsobol, line 29: Do we have to add watchdog to core DDL package? @pomek :thinking:

pomek: I think we should.

filipsobol: It's already there, my bad 🥲

filipsobol (approved):

Witoso:
Please create a follow up issue/task in Notion to update the docs.

arkflpc (approved):

PR #16431, opened 2024-05-28
Update the quick start, bring back editor lifecycle.

Witoso:

Docs: Update the quick start, bring back editor lifecycle.

---

### Additional information


godai78:
By the way, tehcnicla has nothing to do with bernicla. The latter is a species of a duck, knows as brant, while the former is more of a tired typo.

PR #16452, opened 2024-05-31
Find and replace always replaces the first match when track changes is enabled

Mati365:

Fix (find-and-replace): Do not replace only the first found word when using it with `Track Changes` plugin.

MINOR BREAKING CHANGE (find-and-replace): The `FindCommand#execute` now accepts a `findCallback` that can return an array of results or an object containing an array of results together with the `searchText` used to invalidate search results in commands like `find previous`, `find next`, and `replace search`.

---

### Additional information

Closes: https://github.com/cksource/ckeditor5-commercial/issues/6193
Commercial PR: https://github.com/cksource/ckeditor5-commercial/pull/6275

#### Debug notes

1. When search is performed, `FindCommand#execute` is called. 
2. It can be called with `callback` or `string` as a search [parameter](https://github.com/ckeditor/ckeditor5/blob/5ed6fb17ea3cb07571468213361035f4846ad483/packages/ckeditor5-find-and-replace/src/findcommand.ts#L55-L56). 
3. This bug is reproducible if `callback` is passed as search parameter.
4. `Track Changes` [intercepts](https://github.com/ckeditor/ckeditor5/blob/5ed6fb17ea3cb07571468213361035f4846ad483/packages/ckeditor5-find-and-replace/src/findcommand.ts#L85-L91) `find and replace` `find` command `execute` and forces using `callback` parameter (it was originally called with `string` search parameter).
5.  `FindCommand#execute` clears `FindAndReplaceState#searchText`.
6. If `string` search parameter is passed, `FindAndReplaceState#searchText` is being reverted after search (in `FindCommand#execute`). If `callback` search parameter is passed, nothing happens and  `FindAndReplaceState#searchText` stays empty.
7. It means that passing `callback` search parameter resets highlight index [here](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-find-and-replace/src/findandreplace.ts#L57-L94) and [here](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-find-and-replace/src/findandreplace.ts#L79), and it's causing an issue. `state.searchText` is empty (it was cleared in previous steps while using callback search), on the other hand, `data.searchText` is filled with current input value. 

#### Possible solutions

1. Return `searchText` along with `results` in `FindCallback`. 
1.1. It's more flexible if the command is executed with callback which defines the search phrase inside itself [1].
1.2. It's safer according to tests (none failed).
1.3. It introduces a breaking change in integrations that depends on arguments passed to `FindCommand#execute`.
1.4. Search UI seems not to be bound to `searchText` command value. It needs to be refactored, probably.
1.5. It's not super clear for the users why `searchText` has to be forwarded (despite adding comment). 
2. Get rid of `state.searchText`
2.1. Refactoring whole solution, it's kinda complex and many of the tests will be broken. 
2.2. Rename `searchText` to `queryID` or something like that. It'll be more obvious for the users. Less dangerous, but under the hood it'll work exactly like `1` solution.
3. Get rid of support `callback` find. It's an old feature, there are probably a lot of integrations that might depend on this behavior. 

[1]
```tsx
findCommand.execute( () => {
     const callback = typeof callbackOrText === 'string' ? plugin.findByTextCallback( 'Bunny' ) : 'Bunny';
     ...
     return {
        results: ...,
        searchText: 'Bunny'
     };
} );
```

#### Reproduction video

https://github.com/ckeditor/ckeditor5/assets/3949262/944f3f63-a1a9-4ecf-adb8-336f93f44b2c

### DIFF
--- a/packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
+++ b/packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
@@ -166,4 +166,21 @@ export default class FindAndReplaceState extends /* #__PURE__ */ ObservableMixin
 /**
  * The callback function used to find matches in the document.
  */
-export type FindCallback = ( ( { item, text }: { item: Item; text: string } ) => Array<ResultType> );
+export type FindCallback = ( { item, text }: { item: Item; text: string } ) => FindCallbackResult;
+
+/**
+ * Represents the result of a find callback.
+ *
+ * Try to avoid returning array of results. Return the `searchText` attribute in the result object instead.
+ * At this moment we have to keep this for backward compatibility.
+ *
+ * The `searchText` attribute in the result object is used to determine if the search text has been changed.
+ * If returned `searchText` is different than the last search text, the search results will be invalidated
+ * while searching for next item and the search will start from beginning instead of pointing second found item.
+ */
+type FindCallbackResult =
+	| Array<ResultType>
+	| {
+		results: Array<ResultType>;
+		searchText: string;
+	};
--- a/packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
+++ b/packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
@@ -166,21 +166,23 @@ export default class FindAndReplaceState extends /* #__PURE__ */ ObservableMixin
 /**
  * The callback function used to find matches in the document.
  */
-export type FindCallback = ( { item, text }: { item: Item; text: string } ) => FindCallbackResult;
+export type FindCallback = ( { item, text }: { item: Item; text: string } ) => FindCallbackResultObject | FindCallbackResult;
 
 /**
  * Represents the result of a find callback.
  *
- * Try to avoid returning array of results. Return the `searchText` attribute in the result object instead.
- * At this moment we have to keep this for backward compatibility.
- *
- * The `searchText` attribute in the result object is used to determine if the search text has been changed.
+ * The `searchText` attribute in the result object is used to determine if the search text has changed.
  * If returned `searchText` is different than the last search text, the search results will be invalidated
- * while searching for next item and the search will start from beginning instead of pointing second found item.
+ * while searching for next item and the search will start from the beginning of the document.
  */
-type FindCallbackResult =
-	| Array<ResultType>
-	| {
-		results: Array<ResultType>;
-		searchText: string;
-	};
+export type FindCallbackResultObject = {
+	results: Array<ResultType>;
+	searchText: string;
+};
+
+/**
+ * Represents the result of a find callback.
+ *
+ * @deprecated Use `FindCallbackResultObject` instead.
+ */
+export type FindCallbackResult = Array<ResultType>;
--- a/packages/ckeditor5-find-and-replace/src/findandreplaceutils.ts
+++ b/packages/ckeditor5-find-and-replace/src/findandreplaceutils.ts
@@ -47,7 +47,7 @@ export default class FindAndReplaceUtils extends Plugin {
 	public updateFindResultFromRange(
 		range: Range,
 		model: Model,
-		findCallback: ( { item, text }: { item: Item; text: string } ) => Array<ResultType>,
+		findCallback: ( { item, text }: { item: Item; text: string } ) => Array<ResultType> | { results: Array<ResultType> },
 		startResults: Collection<ResultType> | null
 	): Collection<ResultType> {
 		const results = startResults || new Collection();
@@ -67,13 +67,17 @@ export default class FindAndReplaceUtils extends Plugin {
 			[ ...range ].forEach( ( { type, item } ) => {
 				if ( type === 'elementStart' ) {
 					if ( model.schema.checkChild( item, '$text' ) ) {
-						const foundItems = findCallback( {
+						let foundItems = findCallback( {
 							item,
 							text: this.rangeToText( model.createRangeIn( item as Element ) )
 						} );
 
 						if ( !foundItems ) {
 							return;
+						}
+
+						if ( 'results' in foundItems ) {
+							foundItems = foundItems.results;
 						}
 
 						foundItems.forEach( foundItem => {
--- a/packages/ckeditor5-find-and-replace/src/findcommand.ts
+++ b/packages/ckeditor5-find-and-replace/src/findcommand.ts
@@ -60,15 +60,30 @@ export default class FindCommand extends Command {
 		const findAndReplaceUtils: FindAndReplaceUtils = editor.plugins.get( 'FindAndReplaceUtils' );
 
 		let findCallback: FindCallback | undefined;
+		let callbackSearchText: string = '';
 
 		// Allow to execute `find()` on a plugin with a keyword only.
 		if ( typeof callbackOrText === 'string' ) {
-			findCallback = findAndReplaceUtils.findByTextCallback( callbackOrText, { matchCase, wholeWords } );
-
-			this._state.searchText = callbackOrText;
+			findCallback = ( ...args ) => ( {
+				results: findAndReplaceUtils.findByTextCallback( callbackOrText, { matchCase, wholeWords } )( ...args ),
+				searchText: callbackOrText
+			} );
 		} else {
 			findCallback = callbackOrText;
 		}
+
+		// Wrap the callback to get the search text that will be assigned to the state.
+		const oldCallback = findCallback;
+
+		findCallback = ( ...args ) => {
+			const result = oldCallback( ...args );
+
+			if ( result && !Array.isArray( result ) ) {
+				callbackSearchText = result.searchText;
+			}
+
+			return result;
+		};
 
 		// Initial search is done on all nodes in all roots inside the content.
 		const results = model.document.getRootNames()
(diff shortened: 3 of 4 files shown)


Mati365:
@DawidKossowski / @oleq / @scofalik Can you take a look?

f1ames:
I started looking into this PR and the logic in general (well, wanted to understand what's going on before looking into changes itself). Couple of general remarks (not related to the changes itself):

1. The important piece of the puzzle here is what @Mati365 mentioned in:

> It means that passing callback search parameter resets highlight index [here](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-find-and-replace/src/findandreplace.ts#L57-L94) and [here](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-find-and-replace/src/findandreplace.ts#L79), and it's causing an issue. state.searchText is empty (it was cleared in previous steps while using callback search), on the other hand, data.searchText is filled with current input value.

Since `state.searchText` it's not there, pressing `Replace` dialog button, the logic there determines that search term is different than before and reruns find logic (line `79`):

https://github.com/ckeditor/ckeditor5/blob/a477c477d5158d0b24059c6e66b504d0b379641b/packages/ckeditor5-find-and-replace/src/findandreplace.ts#L77-L82

This won't happen when using dialog (because `Replace` button gets disabled when search term changes and user has to perform `Find` first), but can happen when using API directly.

2. TBH I find the logic of overwriting `executeCommand` a bit convoluted. In short, the code overwrites original `find` with own implementation, which calls original `find`, passing custom find callback, which is then executed multiple times in original `find` (for each root children). And every time original/custom find is called, we check `typeof callbackOrText === 'string' ? ...` there. And it's only to filter out deleted parts of the content from find results. I can describe this in more detail if there is time - I know it's not for now but maybe there will be a chance to look into this one day.

3. This is broken for every integration which passes `find` callback function so not only revision history. I see we don't have other usages like that, and probably haven't received any external requests about this. So maybe it's safe to assume that we can fix/extend how it works to allow future integrations to work correctly (what @Mati365 did here). Not sure if fixing this without touching integrations code would be reasonable/doable.

packages/ckeditor5-find-and-replace/src/findcommand.ts
```diff
@@ -60,16 +60,25 @@ export default class FindCommand extends Command {
...
 		} else {
 			findCallback = callbackOrText;
 		}
 
+		// Set the search text to the last search text if it was not provided.
+		findCallback = tapFunctionResult( findCallback, findCallbackResult => {
+			if ( findCallbackResult && !Array.isArray( findCallbackResult ) ) {
+				callbackSearchText = findCallbackResult.searchText;
+			}
+		} );
```
f1ames, line 80: I would avoid such generic wrapper here (especially that it's not used anywhere else), maybe there is a simple, more straightforward way to do this?

packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
```diff
@@ -166,4 +166,22 @@ export default class FindAndReplaceState extends /* #__PURE__ */ ObservableMixin
...
+	// The `searchText` attribute is used to determine if the search query has changed.
+	// If returned `searchText` is different than the last search query, the search results
+	// will be invalidated after clicking `find next` button and the search will be re-run.
+	//
+	// It's highly recommended to return the `searchText` attribute in the result object instead of
+	// returning array to ensure the search results are properly invalidated when the search query changes.
+	| {
+		results: Array<ResultType>;
+		searchText: string;
+	};
```
f1ames, line 187: I'm not sure if we have any convention, or how to do this exactly with TS, but would it be possible to deprecate `Array<ResultType>` with a proper `@deprecated` annotation? 🤔

f1ames: Also if you could move this comment as a part of official part of the docs so it shows up in the API docs. And I would avoid referring to the UI "after clicking `find next` button and", because this behavior is generic (e.g. when using API directly), so this might be confusing.

f1ames (changes requested):
I went through the code and I think the direction you @Mati365 proposed is valid 👍

The core issue here seems to be the fact that `find` command state behavior is a bit inconsistent (or can be perceived as such). If `callbackOrText` string (so a search text) is passed, command state would store this search text (as `state.searchText`), but if `callbackOrText` function is passed, there is no search text. External code which is based on the assumption of existence of `state.searchText`, like "find and replace" dialog may be buggy as in the initial issue.

The assumption that using `callbackOrText` function, requires passing a search text would fit here as a solution IMHO. I was thinking about different ways to do this (e.g. as `execute` param), but didn't come up with anything better. 

---

> 1.4. Search UI seems not to be bound to searchText command value. It needs to be refactored, probably.

@Mati365 can you expand on this?

packages/ckeditor5-find-and-replace/src/findandreplaceutils.ts
```diff
@@ -76,6 +76,10 @@ export default class FindAndReplaceUtils extends Plugin {
 							return;
 						}
 
+						if ( 'results' in foundItems ) {
+							foundItems = foundItems.results;
+						}
+
 						foundItems.forEach( foundItem => {
```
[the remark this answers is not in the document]
Mati365: We need to tell typescript the TS what is the type of `foundItems`, so we need `in` assertion and write it using ternary operator which is not super clear: 

```suggestion
						('results' in foundItems ? foundItems.results : foundItems).forEach( foundItem => {
```

That's why I decided to keep it as plain `if`.

packages/ckeditor5-find-and-replace/src/findcommand.ts
Mati365: Generic wrapper removed.

packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
Mati365: Docs changed.

Mati365:
@f1ames 

> I went through the code and I think the direction you @Mati365 proposed is valid 👍

Fixed CR remarks. Can you check again? 
 
> 1.4. Search UI seems not to be bound to searchText command value. It needs to be refactored, probably. 

From what I can see, here: https://github.com/ckeditor/ckeditor5/blob/ck/6193/packages/ckeditor5-find-and-replace/src/ui/findandreplaceformview.ts#L389-L401 we are using plain DOM attributes to accept search text which is later forwarded on submission of form to the state of the search. I'd refactor it to bind the search text value to the state directly and, depending on `change` events of such text value, clear the history of search results accordingly.

f1ames (changes requested):
Thanks for corrections and clarifications, looks good, just 2 additional suggestions and we should be good to go 👍

packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
```diff
@@ -166,4 +166,21 @@ export default class FindAndReplaceState extends /* #__PURE__ */ ObservableMixin
...
+ * The `searchText` attribute in the result object is used to determine if the search text has been changed.
+ * If returned `searchText` is different than the last search text, the search results will be invalidated
+ * while searching for next item and the search will start from beginning instead of pointing second found item.
+ */
+type FindCallbackResult =
+	| Array<ResultType>
+	| {
+		results: Array<ResultType>;
+		searchText: string;
+	};
```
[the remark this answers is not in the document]
Mati365: Looks good. I'll replace that, thx!

f1ames:
Found another issue while testing. Not related to the changes from this PR though - https://github.com/ckeditor/ckeditor5/issues/16648.

packages/ckeditor5-find-and-replace/src/findandreplacestate.ts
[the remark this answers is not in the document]
Mati365: Replaced

packages/ckeditor5-find-and-replace/src/findcommand.ts
```diff
@@ -60,16 +60,31 @@ export default class FindCommand extends Command {
...
 			findCallback = callbackOrText;
 		}
 
+		// Wrap the callback to get the search text that will be assigned to the state.
+		const oldCallback = findCallback;
+
+		findCallback = ( ...args ) => {
+			const result = oldCallback( ...args );
+
+			if ( result && !Array.isArray( result ) ) {
```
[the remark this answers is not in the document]
Mati365: It may be, I don't think that there'll be a huge difference besides semantics, although I replaced this check with your suggestion because it seems a little bit better in terms of readability.

Mati365:
@f1ames fixed CR remarks

f1ames (approved):
LGTM, good job @Mati365 👍

f1ames:
> Fix (find-and-replace): Do not replace only the first found word when using it with `Track Changes` plugin.
> 
> MINOR BREAKING CHANGE (find-and-replace): The `FindCommand#execute` now accepts a `findCallback` that can return an array of results or an object containing an array of results together with the `searchText` used to invalidate search results in commands like `find previous`, `find next`, and `replace search`.

Just a remark here, I don't see it as a breaking change - here, `findCallback` type was extended, nothing removed, so all previous implementations will work the same way (with the issue as before). So there is a full backwards compatibility in those changes.

Issue #3695, opened 2019-10-09
Refactor OT using `engine.treeModel.Range#getTransformedByMove`


scofalik:
After https://github.com/ckeditor/ckeditor5-engine/pull/392 is merged, we should refactor operational transformation algoirthms so they use the newly introduced method. Right now we have a bit of code duplication.

PR #16473, opened 2024-06-04
CKBox: Update configuration options passed down to CKBox instance by ckbox-plugin

illia-stv:

Feature (ckbox): Added more configuration options passed down to the CKBox. Closes [#3695](https://github.com/cksource/ckeditor5-internal/issues/3695)

---

### Additional information

### DIFF
--- a/packages/ckeditor5-ckbox/src/ckboxcommand.ts
+++ b/packages/ckeditor5-ckbox/src/ckboxcommand.ts
@@ -19,7 +19,6 @@ import type {
 	CKBoxAssetImageDefinition,
 	CKBoxAssetLinkAttributesDefinition,
 	CKBoxAssetLinkDefinition,
-	CKBoxDialog,
 	CKBoxRawAssetDefinition
 } from './ckboxconfig.js';
 
--- a/packages/ckeditor5-ckbox/src/ckboxcommand.ts
+++ b/packages/ckeditor5-ckbox/src/ckboxcommand.ts
@@ -128,31 +128,29 @@ export default class CKBoxCommand extends Command {
 	 * - tokenUrl The token endpoint URL.
 	 * - serviceOrigin The base URL of the API service.
 	 * - forceDemoLabel Whether to force "Powered by CKBox" link.
-	 * - dialog.onClose The callback function invoked after closing the CKBox dialog.
 	 * - assets.onChoose The callback function invoked after choosing the assets.
+	 * - dialog.onClose The callback function invoked after closing the CKBox dialog.
+	 * - dialog.width The dialog width in pixels.
+	 * - dialog.height The dialog height in pixels.
+	 * - categories.icons Allows setting custom icons for categories.
+	 * - view.openLastView Sets if the last view visited by the user will be reopened
+	 * on the next startup.
+	 * - view.startupFolderId Sets the ID of the folder that will be opened on startup.
+	 * - view.startupCategoryId Sets the ID of the category that will be opened on startup.
+	 * - view.hideMaximizeButton Sets whether to hide the ‘Maximize’ button.
+	 * - view.componentsHideTimeout Sets timeout after which upload components are hidden
+	 * after completed upload.
+	 * - view.dialogMinimizeTimeout Sets timeout after which upload dialog is minimized
+	 * after completed upload.
 	 */
 	private _prepareOptions() {
 		const editor = this.editor;
 		const ckboxConfig = editor.config.get( 'ckbox' )!;
 
 		const dialog = ckboxConfig.dialog!;
-		const { width, height } = dialog;
-
-		const icons = ckboxConfig.categories!.icons;
-
+		const categories = ckboxConfig.categories!;
 		const view = ckboxConfig.view!;
-		const {
-			openLastView,
-			startupFolderId,
-			startupCategoryId,
-			hideMaximizeButton
-		} = view;
-
 		const upload = ckboxConfig.upload!;
-		const {
-			componentsHideTimeout,
-			dialogMinimizeTimeout
-		} = upload;
 
 		return {
 			theme: ckboxConfig.theme,
@@ -166,21 +164,21 @@ export default class CKBoxCommand extends Command {
 			},
 			dialog: {
 				onClose: () => this.fire<CKBoxEvent<'close'>>( 'ckbox:close' ),
-				width,
-				height
+				width: dialog && dialog.width,
+				height: dialog && dialog.height
 			},
 			categories: {
-				icons
+				icons: categories && categories.icons
 			},
 			view: {
-				openLastView,
-				startupFolderId,
-				startupCategoryId,
-				hideMaximizeButton
+				openLastView: view && view.openLastView,
+				startupFolderId: view && view.startupFolderId,
+				startupCategoryId: view && view.startupCategoryId,
+				hideMaximizeButton: view && view.hideMaximizeButton
 			},
 			upload: {
-				componentsHideTimeout,
-				dialogMinimizeTimeout
+				componentsHideTimeout: upload && upload.componentsHideTimeout,
+				dialogMinimizeTimeout: upload && upload.dialogMinimizeTimeout
 			}
 		};
 	}
--- a/packages/ckeditor5-ckbox/src/ckboxcommand.ts
+++ b/packages/ckeditor5-ckbox/src/ckboxcommand.ts
@@ -19,6 +19,7 @@ import type {
 	CKBoxAssetImageDefinition,
 	CKBoxAssetLinkAttributesDefinition,
 	CKBoxAssetLinkDefinition,
+	CKBoxDialog,
 	CKBoxRawAssetDefinition
 } from './ckboxconfig.js';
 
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
...
+	 * ```ts
+	 * const ckboxConfig = {
+	 *		choosableFileExtensions: ['jpg', 'png']
+	 * };
+	 * ```
+	 */
+	choosableFileExtensions?: Array<string>;
 }
 
+export type CKBoxDialog = {
```
arkflpc, line 178: Why `export type` if the config uses `export interface`? Seems like an unnecessary inconsistency.

packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
...
+	 */
+	width: number;
+
+	/**
+	 * The dialog height in pixels.
+	 */
+	height: number;
+};
+
+export type CKBoxCategories = {
```
arkflpc, line 191: Why `export type` if the config uses `export interface`? Seems like an unnecessary inconsistency.

packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
...
+	 * 		}
+	 * }
+	 * ```
+	 */
+	icons: {
+		[ key: string ]: string;
+	};
+};
+
+export type CKBoxView = {
```
arkflpc, line 215: Why `export type` if the config uses `export interface`? Seems like an unnecessary inconsistency.

packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
...
+	startupCategoryId: string;
+
+	/**
+	 * Sets whether to hide the ‘Maximize’ button. By default, the button is shown and enabling
+	 * this option will hide it.
+	 */
+	hideMaximizeButton: boolean;
+};
+
+export type CKBoxUpload = {
```
arkflpc, line 245: Why `export type` if the config uses `export interface`? Seems like an unnecessary inconsistency.

packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
...
+	/**
+	 * Configures when dialog should be minimized and hidden.
+	 */
+	upload?: CKBoxUpload;
+
+	/**
+	 * Specifies the file extensions considered valid for user interaction. Whith this
+	 * option developers can restrict user interaction to only those assets whose file
+	 * extensions match those listed in the array. Assets whose file
+	 * extensions are not listed in the choosableFileExtensions array are
```
arkflpc, line 166:
```suggestion
	 * extensions are not listed in the `choosableFileExtensions` array are
```

packages/ckeditor5-ckbox/src/ckboxconfig.ts
arkflpc, line 178: I'd add a suffix `Config` to this type name. Like `ToolbarConfig` or `LanguageConfig` are named. In case, we have a class that actually represents a `CKBoxDialog`. And for consistency.

packages/ckeditor5-ckbox/src/ckboxconfig.ts
arkflpc, line 191: I'd add a suffix `Config` to this type name. Like `ToolbarConfig` or `LanguageConfig` are named. In case, we have a class that actually represents a `CKBoxDialog`. And for consistency.

packages/ckeditor5-ckbox/src/ckboxconfig.ts
arkflpc, line 215: I'd add a suffix `Config` to this type name. Like `ToolbarConfig` or `LanguageConfig` are named. In case, we have a class that actually represents a `CKBoxDialog`. And for consistency.

packages/ckeditor5-ckbox/src/ckboxconfig.ts
arkflpc, line 245: I'd add a suffix `Config` to this type name. Like `ToolbarConfig` or `LanguageConfig` are named. In case, we have a class that actually represents a `CKBoxDialog`. And for consistency.

packages/ckeditor5-ckbox/src/ckboxconfig.ts
arkflpc, line 215: Is it an all-or-nothing config? Shouldn't the members be optional?

packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
...
+	 */
+	icons: {
+		[ key: string ]: string;
+	};
+};
+
+export type CKBoxView = {
+
+	/**
+	 * If it is set to false the last view visited by the user will not be reopened on
```
arkflpc, line 218:
```suggestion
	 * If it is set to `false` the last view visited by the user will not be reopened on
```

packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
...
+	 * the next startup.
+	 */
+	openLastView: boolean;
+
+	/**
+	 * Sets the ID of the folder that will be opened on startup. This option can be paired
+	 * with setting view.openLastView to false to enforce CKBox to always open in a given
```
arkflpc, line 225:
```suggestion
	 * with setting `view.openLastView` to `false` to enforce CKBox to always open in a given
```

packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
...
+	 * folder at startup.
+	 */
+	startupFolderId: string;
+
+	/**
+	 * Sets the ID of the category that will be opened on startup. This option can be paired
+	 * with setting view.openLastView to `false` to enforce CKBox to always open in a given
```
arkflpc, line 232:
```suggestion
	 * with setting `view.openLastView` to `false` to enforce CKBox to always open in a given
```

packages/ckeditor5-ckbox/src/ckboxconfig.ts
arkflpc, line 245: Is it an all-or-nothing config? Shouldn't the members be optional?

packages/ckeditor5-ckbox/src/ckboxcommand.ts
```diff
@@ -127,24 +127,57 @@ export default class CKBoxCommand extends Command {
...
 	 */
 	private _prepareOptions() {
 		const editor = this.editor;
 		const ckboxConfig = editor.config.get( 'ckbox' )!;
 
+		const dialog = ckboxConfig.dialog!;
+		const categories = ckboxConfig.categories!;
+		const view = ckboxConfig.view!;
+		const upload = ckboxConfig.upload!;
+
```
arkflpc, line 153: Why not-null assertions (`!`)? Below, there are guards (like `dialog && dialog.with`) for each of those values.

packages/ckeditor5-ckbox/src/ckboxcommand.ts
```diff
@@ -127,24 +127,57 @@ export default class CKBoxCommand extends Command {
...
+			choosableFileExtensions: ckboxConfig.choosableFileExtensions,
 			assets: {
 				onChoose: ( assets: Array<CKBoxRawAssetDefinition> ) => this.fire<CKBoxEvent<'choose'>>( 'ckbox:choose', assets )
+			},
+			dialog: {
+				onClose: () => this.fire<CKBoxEvent<'close'>>( 'ckbox:close' ),
+				width: dialog && dialog.width,
+				height: dialog && dialog.height
+			},
+			categories: {
```
arkflpc, line 169: Do we want to pass an empty object if `categories` is not present in config?

packages/ckeditor5-ckbox/src/ckboxcommand.ts
```diff
@@ -127,24 +127,57 @@ export default class CKBoxCommand extends Command {
...
+				icons: categories && categories.icons
+			},
+			view: {
```
arkflpc, line 172: Do we want to pass an empty object if `view` is not present in config?

packages/ckeditor5-ckbox/src/ckboxcommand.ts
```diff
@@ -127,24 +127,57 @@ export default class CKBoxCommand extends Command {
...
+				openLastView: view && view.openLastView,
+				startupFolderId: view && view.startupFolderId,
+				startupCategoryId: view && view.startupCategoryId,
+				hideMaximizeButton: view && view.hideMaximizeButton
+			},
+			upload: {
```
arkflpc, line 178: Do we want to pass an empty object if `upload` is not present in config?

packages/ckeditor5-ckbox/tests/ckboxcommand.js
```diff
@@ -216,7 +216,27 @@ describe( 'CKBoxCommand', () => {
...
+						},
+						categories: {
+							icons: {
+								key: 'svg'
+							}
+						},
+						view: {
+							openLastView: true,
+							startupFolderId: 'id',
+							startupCategoryId: 'id',
```
arkflpc, line 233: If those values were different, the test would be more reliable.

arkflpc (changes requested):

packages/ckeditor5-ckbox/src/ckboxconfig.ts
arkflpc: I know that it does not make sense to pass an empty `categories` object in the config, but people sometimes do it? Shouldn't `icon` property be optional?

arkflpc:
Please change the suggested comment message. `Update` is too general for changelog. Something like "Added more/several options, ....".

packages/ckeditor5-ckbox/src/ckboxconfig.ts
```diff
@@ -137,8 +137,131 @@ export interface CKBoxConfig {
 	 * Defaults to {@link module:utils/locale~Locale#uiLanguage `Locale#uiLanguage`}
 	 */
 	language?: string;
+
+	/**
+	 * This option allows opening CKBox in dialog mode. It takes a boolean value or
+	 * a configuration object with the following attributes:
+	 */
+	dialog?: CKBoxDialog;
```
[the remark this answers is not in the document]
illia-stv: Indeed we pass `onClose` method to the CKBox, but we are using it for internal purposes (I mean closing CKBox dialog), so I guess this option should not be public, but private, in context of CKE5. Do you think we should allow users to override `onClose` method?

packages/ckeditor5-ckbox/src/ckboxconfig.ts
illia-stv: Sure, I just followed the way we do it in other places, like `export-pdf`, so here we declare an **interface**:
https://github.com/cksource/ckeditor5-commercial/blob/master/packages/ckeditor5-export-pdf/src/exportpdf.ts#L92
but option of the config is **type**:
https://github.com/cksource/ckeditor5-commercial/blob/master/packages/ckeditor5-export-pdf/src/exportpdf.ts#L219

But as I see in other places is done with `interfaces` indeed.

packages/ckeditor5-ckbox/src/ckboxcommand.ts
illia-stv: I don't know certainly why, but I afraid `ckboxConfig.dialog!.width!` is not valid. I tried it out and tests are crushed with that expression.

arkflpc (approved):

Issue #16544, opened 2024-06-11
A11y best practice: Landmarks should be unique


Joel-Levi:
Following [this pull request](https://github.com/ckeditor/ckeditor5/pull/16197)

There are now 2 aria live regions with role='region'
![image](https://github.com/ckeditor/ckeditor5/assets/43135922/85813537-ed5b-4b8f-8e19-097eff028d0a)

This violates the following best practice rule:

Rule: landmark-unique
Impact: moderate
Ensures landmarks are unique ([https://dequeuniversity.com/rules/axe/4.8/landmark-unique?application=axeAPI)](https://dequeuniversity.com/rules/axe/4.8/landmark-unique?application=axeAPI))
   
The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable

While it is not strictly required, it is considered a best practice, I found out about it because I run [chai a11y axe](https://www.npmjs.com/package/chai-a11y-axe)

* Browser: All
* OS: N/A
* First affected CKEditor version: v41.4.0
* Installed CKEditor plugins: ---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Mati365:
Hi @Joel-Levi! 

Thanks for report. Indeed, the `role=region` attribute should not be set there, so your tests were correct. We disused it, and we'll remove this attribute in this [PR](https://github.com/ckeditor/ckeditor5/pull/16550). 

Let's know in case of any further issues.

PR #16550, opened 2024-06-12
A11y best practice: Landmarks should be unique

Mati365:

Fix (ui): Remove unnecessary `role=region` from aria live views due to failing `landmarks should be unique` a11y axe rule.  Closes https://github.com/ckeditor/ckeditor5/issues/16544. 

### Further information

1. There is no need to set `role=region` on `aria-live` regions because it's ignored by readers if associated HTML element has no label attributes.
2. `role=region` is an alternative to `aria-live` and should not be used together. 
3. Checked using `NVDA` and it sounds fine.


DawidKossowski (approved):
LGTM.

Issue #16583, opened 2024-06-17
Add a scenario for inline container elements in `viewToPlainText()` function


Dumluregn:
## 📝 Provide a description of the improvement

Currently, the [`viewToPlainText()` ](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts#L25-L52) function (and more specifically in it's helper, [`newLinePadding()`](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts#L57-L91) ) always adds empty lines before and after the view container elements. It's not possible to have an inline container. It causes problems when copying such a content and then pasting it into a code block (as the `text/plain` data transfer type is used then) - the content is split into multiple paragraphs.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

Dumluregn:
Solved in https://github.com/ckeditor/ckeditor5/pull/16584.

PR #16584, opened 2024-06-17
Added handling for the inline view container elements in the `text/plain` data transfer

Dumluregn:

Feature (clipboard): Added handling for the elements with `dataPipeline:transparentRendering` property in the `text/plain` data transfer. Closes https://github.com/ckeditor/ckeditor5/issues/16583.

---

### Additional information

Theoretically, we could also solve it without adding a custom element property, but it's more complex: we need to convert the element to model one and then check in schema if it is an inline object or something similar.

### DIFF
--- a/packages/ckeditor5-clipboard/src/pasteplaintext.ts
+++ b/packages/ckeditor5-clipboard/src/pasteplaintext.ts
@@ -81,7 +81,7 @@ export default class PastePlainText extends Plugin {
 }
 
 /**
- * Returns true if specified `documentFragment` represents the unformatted inline content.
+ * Returns true if specified `documentFragment` represents an unformatted inline content.
  */
 function isUnformattedInlineContent( documentFragment: DocumentFragment, model: Model ): boolean {
 	const range = model.createRangeIn( documentFragment );
@@ -93,7 +93,7 @@ function isUnformattedInlineContent( documentFragment: DocumentFragment, model: 
 
 		const attributeKeys = Array.from( child.getAttributeKeys() );
 
-		if ( attributeKeys.length !== 0 && attributeKeys.find( key => model.schema.getAttributeProperties( key ).isFormatting ) ) {
+		if ( attributeKeys.find( key => model.schema.getAttributeProperties( key ).isFormatting ) ) {
 			return false;
 		}
 	}
--- a/packages/ckeditor5-clipboard/src/pasteplaintext.ts
+++ b/packages/ckeditor5-clipboard/src/pasteplaintext.ts
@@ -9,7 +9,7 @@
 
 import { Plugin } from '@ckeditor/ckeditor5-core';
 
-import type { DocumentFragment, Schema, ViewDocumentKeyDownEvent } from '@ckeditor/ckeditor5-engine';
+import type { DocumentFragment, Model } from '@ckeditor/ckeditor5-engine';
 
 import ClipboardObserver from './clipboardobserver.js';
 import ClipboardPipeline, { type ClipboardContentInsertionEvent } from './clipboardpipeline.js';
@@ -41,21 +41,14 @@ export default class PastePlainText extends Plugin {
 		const editor = this.editor;
 		const model = editor.model;
 		const view = editor.editing.view;
-		const viewDocument = view.document;
 		const selection = model.document.selection;
 
-		let shiftPressed = false;
-
 		view.addObserver( ClipboardObserver );
-
-		this.listenTo<ViewDocumentKeyDownEvent>( viewDocument, 'keydown', ( evt, data ) => {
-			shiftPressed = data.shiftKey;
-		} );
 
 		editor.plugins.get( ClipboardPipeline ).on<ClipboardContentInsertionEvent>( 'contentInsertion', ( evt, data ) => {
 			// Plain text can be determined based on the event flag (#7799) or auto-detection (#1006). If detected,
 			// preserve selection attributes on pasted items.
-			if ( !shiftPressed && !isPlainTextFragment( data.content, model.schema ) ) {
+			if ( !isUnformattedInlineContent( data.content, model ) ) {
 				return;
 			}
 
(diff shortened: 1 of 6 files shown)


scofalik (changes requested):
I don't like this magic property -- it's very easy to forget about setting such thing. If we go this way, we should probably add a method `view.Writer#createInlineContainerElement()` and then change our implementations if we have container inline elements anywhere (probably not?).

Instead, I see two different ways we can go about it:

1.  I think I earlier raised my doubt whether merge field should be a container element. It's more like an empty container, i.e. it is atomic, indivisible and you don't "edit inside it". AFAIK `EmptyElement` cannot be used from technical reasons, so I propose `RawElement`. If not that, then it is a bit problematic.
2.  Instead of introducing a new entity ("inline container element") and a new property for it, we can base on `dataPipeline:transparentRendering`. Such view element should be transparent when converting view to final data output. So, if it should be transparent, it should not add any padding (although its children can - I am not sure how exactly this method is used, whether it is fired recursively, I hope so).

I like both solutions, but I think that 2nd is also a good generic solution, so if we want to postpone discussion about `RawElement`, we can go with 2nd solution for now and it will not be incorrect even if we swap to `RawElement` anyway.

niegowski:
> I don't like this magic property -- it's very easy to forget about setting such thing. If we go this way, we should probably add a method `view.Writer#createInlineContainerElement()` and then change our implementations if we have container inline elements anywhere (probably not?).

I'd suggest using `DomConverter#blockElements` property to verify where we should separate with newlines.

scofalik:
That's a good suggestion too. We are talking about `<span>`s here, and they should not be marked as `blockElements`, so it will work too.

Although still I think that ignoring transparent elements makes sense, both improvements will fix our problem and make sense on their own.

niegowski:
> inline image is `` which I guess is block?

Those are `inlineObjectElements`.

Dumluregn:
I went with the `dataPipeline:transparentRendering` solution after all.

scofalik:
LGTM. @niegowski is it okay for you?

packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
```diff
@@ -86,6 +86,12 @@ function newLinePadding(
 		return '\n';
 	}
 
+	// Do not add padding around the elements that won't be rendered.
+	if ( ( element.is( 'element' ) && element.getCustomProperty( 'dataPipeline:transparentRendering' ) ) ||
+		( previous.is( 'element' ) && previous.getCustomProperty( 'dataPipeline:transparentRendering' ) ) ) {
```
niegowski, line 91:
```suggestion
	if ( 
		element.is( 'element' ) && element.getCustomProperty( 'dataPipeline:transparentRendering' ) ||
		previous.is( 'element' ) && previous.getCustomProperty( 'dataPipeline:transparentRendering' )
	) {
```

niegowski (approved):

scofalik:
Thanks @niegowski, we will wait with merge until merge fields feature is ready.

scofalik (approved):

packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
```diff
@@ -86,6 +86,14 @@ function newLinePadding(
 		return '\n';
 	}
 
+	// Do not add padding around the elements that won't be rendered.
+	if ( 
+		element.is( 'element' ) && element.getCustomProperty( 'dataPipeline:transparentRendering' ) ||
+		previous.is( 'element' ) && previous.getCustomProperty( 'dataPipeline:transparentRendering' )
+	) {
```
niegowski, line 93:
```suggestion
	if (
		element.is( 'element' ) && element.getCustomProperty( 'dataPipeline:transparentRendering' ) ||
		previous.is( 'element' ) && previous.getCustomProperty( 'dataPipeline:transparentRendering' )
	) {
```

Dumluregn:
PR was changed due to the decision to move the integration points to the paste plain text and code block features.

packages/ckeditor5-clipboard/src/pasteplaintext.ts
```diff
@@ -86,18 +81,22 @@ export default class PastePlainText extends Plugin {
 }
 
 /**
- * Returns true if specified `documentFragment` represents a plain text.
+ * Returns true if specified `documentFragment` represents the unformatted inline content.
```
scofalik, line 84:
```suggestion
 * Returns true if specified `documentFragment` represents an unformatted inline content.
```

packages/ckeditor5-clipboard/src/pasteplaintext.ts
```diff
@@ -86,18 +81,22 @@ export default class PastePlainText extends Plugin {
...
+		if ( !model.schema.isInline( child ) ) {
+			return false;
+		}
 
-	const child = documentFragment.getChild( 0 )!;
+		const attributeKeys = Array.from( child.getAttributeKeys() );
 
-	if ( schema.isObject( child ) ) {
-		return false;
+		if ( attributeKeys.length !== 0 && attributeKeys.find( key => model.schema.getAttributeProperties( key ).isFormatting ) ) {
```
scofalik, line 96: Do we need this part `attributeKeys.length !== 0 &&` in this `if`?

packages/ckeditor5-code-block/src/codeblockediting.ts
```diff
@@ -202,6 +203,31 @@ export default class CodeBlockEditing extends Plugin {
...
+				const selection = model.document.selection;
+
+				if ( !selection.anchor!.parent.is( 'element', 'codeBlock' ) ) {
+					return;
+				}
+
+				model.change( writer => {
+					const contentRange = writer.createRangeIn( data.content );
+
+					for ( const item of [ ...contentRange.getItems() ] ) {
```
scofalik, line 221: I think `for ( const item of contentRange.getItems() )` should be enough, no need to convert to array.

scofalik (changes requested):
Very minor issues.

packages/ckeditor5-code-block/src/codeblockediting.ts
Dumluregn: There's a removal in this loop. When it triggers, the iterator stops. If you have two merge fields in the pasted content, the second one will not be deleted.

PR #16591, opened 2024-06-18
Add menu bar to remaining editor types.

### LINKED ISSUE #16571: Make menu bar available in each editor type

For now, we have integrated the menu bar only in the classic, document, and multi-root editors. There is no reason why other editor types should not have the possibility to display the menu bar. Additionally, it will simplify the whole logic.

Task requirements:
1. Ensure that all editor types have the possibility to display the menu bar.
2. Refactor the code to simplify it by moving the logic responsible for registering the menu bar to `EditorUI`.

---

mremiszewski:

Feature (editor-balloon, editor-inline): Add menu bar support. Closes #16571.

---

### Additional information


DawidKossowski (changes requested):
@mremiszewski, In the case of the balloon editor, we cannot include the menu bar inside the balloon as it breaks the UX. Instead, we should handle it similarly to the multi-root editor. You should provide an independent element that integrators can render in any place. After making this change, moving the entire logic to `EditorUI` should be possible.

packages/ckeditor5-ui/src/editorui/editorui.ts
```diff
@@ -322,6 +324,34 @@ export default abstract class EditorUI extends /* #__PURE__ */ ObservableMixin()
...
+				cancel();
+			}
+		} );
+
+		this.editor.keystrokes.set( 'Alt+F9', ( data, cancel ) => {
+			if ( !menuBarViewElement.contains( this.editor.ui.focusTracker.focusedElement ) ) {
+				menuBarView!.focus();
+				cancel();
+			}
+		} );
```
Mati365, line 352: This handler has to be extended after merging this PR: https://github.com/ckeditor/ckeditor5/pull/16573, we added focus bordering flag assign on alt+f9.

DawidKossowski (commented):
Just one comment. Besides that LGTM.

packages/ckeditor5-ui/src/editorui/editorui.ts
Mati365: I see that you migrate it, thanks!

DawidKossowski (changes requested):
I just found that there are missing tests for `editor-balloon` in the context of the menu bar.

We also moved some logic from `ui/utils` to `ui/editorui`. Shouldn't we also move some unit tests (like focus handling tests)?

DawidKossowski (approved):

PR #16596, opened 2024-06-19
Formatting should not get lost on a text with a marker in a table cell during upcast

oleq:

Fix (engine): Formatting should not get lost on a text with a marker in a table cell during upcast. 

Reverted a now-obsolete fix for markers not getting auto-paragraphed together with text. 

---

### Additional information

See https://github.com/cksource/ckeditor5-commercial/issues/5287. See https://github.com/cksource/ckeditor5-commercial/issues/4061.

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
@@ -536,16 +536,7 @@ export function convertText() {
 				return;
 			}
 
-			// Wrap `$text` in paragraph and include any marker that is directly before `$text`. See #13053.
-			const nodeBefore = position.nodeBefore;
-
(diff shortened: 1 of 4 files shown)


DawidKossowski:
The broken unit tests were not dangerous since they tested the markers inside the table where autoparagraphing was applied. The positions had to be updated because we reverted the fix for recalculating the markers and placed them always into the element.

We agreed that reverting it is a better idea, and this bug is less annoying than it was previously. There is no risk in the context of comments since we have a postfixer to fix the positions. In the context of suggestions, there is a situation where the suggestion starts at the beginning of the table cell, and there is no possibility to write something before it. However, earlier this bug affected the whole cell.

DawidKossowski (approved):

PR #16615, opened 2024-06-26
Merge fields related improvements and adjustments

Dumluregn:

Other (ui): Added the `LabelWithHighlightView` and `ButtonLabelWithHighlightView` components to the UI library.

Other (ui): Added the `filterGroupAndItemNames()` helper to the UI library.

Other (ui): The `SearchTextView#reset()` method will also reset the scroll of its `filteredView` to the top.

Internal (code-block): Support for inline elements in code block. By default code block does not allow elements but some exceptions (like merge fields) are allowed.

Internal (engine): Fixed whitespace handling around `dataPipeline:transparentRendering` elements in downcast.

Internal (engine): Added `inlineObjectElements` option to `dev-utils` to correctly set model data with custom inline object elements.

---

### Additional information

Successor of #16096 after renaming the feature and branch.

To do: https://github.com/ckeditor/ckeditor5/pull/16096#discussion_r1595707864.


packages/ckeditor5-engine/src/view/domconverter.ts
```diff
@@ -182,7 +182,7 @@ export default class DomConverter {
 			'td', 'tfoot', 'th', 'thead', 'tr', 'ul'
 		];
 		this.inlineObjectElements = [
-			'object', 'iframe', 'input', 'button', 'textarea', 'select', 'option', 'video', 'embed', 'audio', 'img', 'canvas'
```
scofalik, line 185: Why this change was needed (what does it do?), and, also, it looks a bit weird, as we define DOM elements here, and `mergeField` is not a DOM element.

Also, as @niegowski commented, shouldn't it be injected from the feature? (I don't know this part of the code very well.)

packages/ckeditor5-ui/src/highlightedtext/buttonlabelwithhighlightview.ts
scofalik, line 1: I'd put this in `button` directory.

packages/ckeditor5-ui/src/highlightedtext/labelwithhighlightview.ts
scofalik, line 1: I'd put this in `label` directory.

scofalik (commented):
I am not sure about the `domconverter.ts` change because I don't know what that code does and why `inlineObjectElements` had to be expanded.

PR #16627, opened 2024-06-27
Docs: Highlight legacy docs more.

gorzelinski:

Docs: Highlight legacy docs more.

---

### Additional information

I also fixed some minor (formatting) issues.

### DIFF
--- a/docs/getting-started/advanced/content-styles.md
+++ b/docs/getting-started/advanced/content-styles.md
@@ -1,11 +1,11 @@
 ---
 category: advanced
-menu-title: Content styles
+menu-title: (Legacy) Content styles
 meta-title: Content styles | Legacy CKEditor 5 documentation
 order: 30
 ---
 
-# Content styles &ndash; Legacy guide
+# (Legacy) Content styles
 
 <info-box warning>
 	⚠️  We changed installation methods and this legacy guide is kept for users' convenience. If you are looking for current CKEditor 5 installation instructions, please refer to the newest version of the {@link getting-started/setup/css Editor and content styles} guide.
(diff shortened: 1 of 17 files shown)


gorzelinski:
I tested different structures, but "legacy" in brackets looked best for me.

godai78:
Not sure it makes sense to add (Legacy) to menu titles, when it resides in the Legacy folder. Guides titles are OK.

godai78 (approved):

Witoso:
> Not sure it makes sense to add (Legacy) to menu titles, when it resides in the Legacy folder. Guides titles are OK.

We agreed it's not best aesthetically but we need to catch the attention better.

Witoso (changes requested):
TBH I'm used to `[` and `]` in those scenarios. What was the reason to use parens?

gorzelinski:
I was saying the same thing, but as we discussed today, the folder name can be missable. It's a good point, I think.

gorzelinski:
Ok, there is a technical problem, actually. Square brackets in frontmatter (YAML) denote a sequence - they break our YAML. So, we have two options:

- Leave the menu titles as they are
- Put all those titles in apostrophes `'[Legacy] Blah, blah'` (they would differ from the rest of the docs)

cc @Witoso

Witoso:
Use the break `\[Legacy\]` I think it should work.

gorzelinski:
> Use the break \[Legacy\] I think it should work.

Oh, yeah, escaping the character is also a way.

gorzelinski:
> Oh, yeah, escaping the character is also a way.

...or not. It renders the backslash, unfortunately.

Screenshot 2024-06-27 at 14 34 30

Witoso:
Ok, let's go back to the parens, sorry!

gorzelinski:
No problem, I also thought there was a chance that it would work. I didn't push my local changes, so everything is as it was :)

Witoso (approved):

Issue #16311, opened 2024-05-06
Create dropdown menu component


DawidKossowski:
## 📝 Provide a description of the new feature

For now, we use menu components as a part of the menu bar. However, they are incompatible with the rest of the editor and cannot be used as independent components in the toolbar. We have at least two use cases where we need them: AI Commands and Merge tags. This task focuses on creating a new component that resembles menus in the menu bar.

The main requirements are as follows:

1. The main logic should be copied from the current menu bar components. Do not focus too much on refactoring and reusing some commons between the dropdown menu and menu bar components.
2. The components should land in the `ckeditor5-ui` package as a subfolder of the dropdown component.
3. Nice to have: Add the possibility to implement the search (like in the current AI Commands dropdown).

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

Mati365:
Follow-up ticket: https://github.com/ckeditor/ckeditor5/issues/16613

Mati365:
Assigned PRs:

**Public:**

1. https://github.com/ckeditor/ckeditor5/pull/16631
2. https://github.com/ckeditor/ckeditor5/pull/16314
3. https://github.com/ckeditor/ckeditor5/pull/16642

**Commercial:**

1. AI Integration: https://github.com/cksource/ckeditor5-commercial

Issue #5514, opened 2019-10-09
Dropdown panel should not not be cut off when the editor is inside in an overflow: hidden host


oleq:
ATM the dropdown recognizes the boundaries of the viewport and positions accordingly:

![image](https://user-images.githubusercontent.com/1099479/60428953-4b5c9a00-9bfa-11e9-9b10-825f78bbb5aa.png)

![image](https://user-images.githubusercontent.com/1099479/60428960-51527b00-9bfa-11e9-9716-e0b5637b227e.png)

But it does not recognize the `overflow: hidden` ancestors and gets cut off, which is definitely annoying. It makes it impossible to use dropdown in the toolbar in modal windows if the dropdown is to be somewhere later in the toolbar (closer to the right edge).

![image](https://user-images.githubusercontent.com/1099479/60428843-17817480-9bfa-11e9-84ac-88ee30130e0a.png)

datasoysa:
Issue seems a bit old and closed, but I still have the problem. Should we reopen this one or generate a new ticket I don't know but only thing i can tell that the approach by trying to understand the overflow:hidden ancestors won't work. As my condition it didn't work, It have a grid, grid-item with overflow-hidden and a flex-box stretches the content. Solution should be applied on a bit more generic way. And I'm sure that it will be more easier than understanding the ancestors. Create an overlay element on top of everything. for example a div element with position static at the end of body and locate your dropdowns inside that according to the reference elements .getBoundingClientRect().top and left and of course width. Than you'll never fall in to dropdown crops.

kvas-damian:
the same problem appears when CKEditor is rendered inside vertically scrollable container. For vertically scrollable container browsers, browser cut/hide content that overlaps horizontally (similar to `overflow-x: hidden` but you cannot override it).
Example: https://codepen.io/kvas-damian/pen/xxjEYXb

ikhidebright:
This is still an issue, its sad

nechutny:
This can be easily fixed with CSS anchors

```css
.ck.ck-dropdown {
		anchor-name: --ck-dropdown;

		.ck.ck-dropdown__panel {
			position: fixed;
			position-anchor: --ck-dropdown;
			top: anchor(bottom);
			left: anchor(left);
			z-index: 99999;
			transform: none;
			min-width: auto;
		}
	}
```

but currently it supports only Chrome based browsers. But still better then nothing.

LePtitJoNadeau:
This is still an issue...

[...]

PR #16631, opened 2024-06-28
Create dropdown menu component

Mati365:
Feature (ui): Create dropdown menu component. 

Part of https://github.com/ckeditor/ckeditor5/issues/16311.

---

### Other PRs

Focus change PR: https://github.com/ckeditor/ckeditor5/pull/16642
Search integration PR: https://github.com/ckeditor/ckeditor5/pull/16314
AI Integration PR: https://github.com/cksource/ckeditor5-commercial/pull/6213

### Screens

⚠️  Search function is introduced in another [PR](https://github.com/ckeditor/ckeditor5/pull/16314).

Menu
Scrolling
RTL

### Example usage

```ts
const view = new DropdownMenuRootListView( editor, [
  {
     menu: 'Menu 1',
     children: [
       // Initialize using definition object.
       {
          label: 'Menu Item 1.1',
          onExecute: () => { ... }
       },
       // Initialize using class instance.
       new DropdownMenuListItemButtonView( locale, 'Item' ),
       new ListSeparatorView( locale ),
       // Submenu initialization.
       {
         menu: 'Menu 1.1',
         children: [
           new DropdownMenuListItemButtonView( locale, 'Nested Item' ),
         ]
       }
     ] 
  } 
] );

// Somewhere later in the code.
view.factory.appendChild( {
  menu: 'Menu 2',
  children: [
    new DropdownMenuListItemButtonView( locale, 'Item 2' )
  ]
} );

// It is also possible to register custom menu using it's instance and modify it.
const customInstance = new DropdownMenuView( editor, 'Menu' );
customInstance.factory.appendChild( {
  menu: 'Sub menu',
  children: [
    new DropdownMenuListItemButtonView( locale, 'Item 3' ),
  ]
} );

// Walk over dropdown content. 
customInstance.walk( {
    Menu: () => { ... },
    Default: () => { ... }
} );

view.factory.appendChild( customInstance );
view.render(); 
```

###  Changes performed on menubar implementation

1. Generate `menus` property dynamically instead of storing it as array class member. It is possible to register already created `DropdownMenuView` instance that was created outside of menu. Such instance might have items added or deleted depending on external integration logic (for example, adding button menu entry). 

Example:

```ts
const customMenuInstance = new DropdownMenuView( editor, 'My custom menu' );
const view = new DropdownMenuRootListView( locale, [
  {
     menu: 'Menu 1',
     children: [
       // ... other menu items
       customMenuInstance
     ] 
  } 
] );

// ... later in the code ...
const newSubMenu =  new DropdownMenuView( editor, 'My submenu menu' );

newSubMenu.menuItems.addMany( [   
    new DropdownMenuListItemButtonView( locale, 'Your Searchable Item' ),
    new ListSeparatorView( locale ),
] );

// It's optional to use `appendMenuChildren` exported by root menu class. 
// New items can be added manually, without need to access root menu instance.
customMenuInstance.menuItems.add(
   new DropdownMenuListItemView( locale, menuInstance, newSubMenu)
);

// ... later in filter implementation ... 
// It should return 1 item that was added to `newSubMenu` because
// `DropdownMenuListItemButtonView` is automatically discovered while `tree` construction.
filterView.menuView.search( 'Your Searchable Item' );
```

2. Added `tree` getter that allows to traverse through whole menu that allows for lookup for specific menu entries in tests, filtering menu or creating another instance of dropdown menu with mapped definition. The one nice side effect of such approach is the ability to create `dump()` functionality, which allows to easily serialize whole menu to simple pseudo-HTML form, which is super easy to test in other integrations. 

```ts
const view = new DropdownMenuRootListView( editor, [
  {
     menu: 'Menu 1',
     children: [
       // initialize using definition object
       {
          label: 'Menu Item 1.1',
          onExecute: () => { ... }
       },
       // initialize using object
       new DropdownMenuListItemButtonView( locale, 'Item' ),
       new ListSeparatorView( locale ),
       // submenu initialization
       {
         menu: 'Menu 1.1',
         children: [
           new DropdownMenuListItemButtonView( locale, 'Nested Item' ),
         ]
       }
     ] 
  } 
] );

// Walk through menu and do some stuff on specific items.
view.walk( {
    Item: ( { node: { search } } ) => {
      if ( search.raw.includes( 'cats' ) ) {
         console.info( 'I see cat here...' ); 
      }
    }
} );

// Let's access tree and menu directly.
console.info( view.tree, view.menus );
```

3. Horizontal menu bar and vertical positioning dropdown button positioning functions have been removed.
4. DSL format of menu definition is simplified and `groups` menu attributes has been removed. Now it is possible to pass  `ListSeparatorView` manually.
5. Icons are optional, though it is possible to allocate space for missing if it is not present (by default it is disabled). 
6. Event delegation moved from root menu list to normal menu instances.
7. Menu panel views are now mounted to `ck-body-wrapper`.

### Drawbacks

1. ~`tree` / `menus` getter calls might be expensive for huge menus (> 500 entries). It's not real case and at this moment there are no actions required to be performed in this scope.~ It is fixed by adding optional lazy initialization of menus. 
2. ~Scrolling in large sub-menus is missing.~
3. It might look awful on mobile.

### Useful links and discussions

1. https://mdn.github.io/dom-examples/popover-api/nested-popovers/
2. https://blog.logrocket.com/use-css-anchor-positioning/
3. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover
4. https://github.com/ckeditor/ckeditor5/issues/5514#issuecomment-507230949


scofalik (changes requested):
Some further cleanup required:

*   `menu/definition` directory - please put all things into one file, they are all mostly used only by the factory util, with I thing just one exception. Remove "definition" factory and move the singular file into `menu/` directory.
*   Change the factory class name and file name to something like `DropdownMenuFactory` -- it does not create a definition, it creates a menu.
*   Drop everything related to "tree" utils, the whole directory. From what I checked, everything that is in there comes down to `.menus` getter. AFAICS, these utils are not used anywhere else (there is `tree` property too but it isn't used anywhere outside `.menus`, and there is `.walk()` which isn't used anywhere).
    *   To build `.menus` I am pretty sure we can do it with a simple tree-search algorithm that uses `while` and an `array` to store results. We won't need to use several utils, types and files.
    *   `preloadAllMenus()` seems not to be used anywhere besides tests (at least that what my IDE tells me).
    *   Same with `getCurrentlyVisibleMenusElements()`.

---

I wrote that maybe some of the utils may stay. I meant just `menus`, but I thought that `menus` include both menu views and button views. `tree` is not necessary. But we may need something like `items` (or `buttons`) -- all leaf items in the structure, flattened. So that we can operate on them when the editor state (e.g. selection) changes.

For `items` entries, we may need a "path" to them, i.e. an array of menus in which an item is. But I think that will be needed for the search only, and I purposely don't mention any search related things when discussing what should be dropped from this PR. As I mentioned multiple times, this PR should be only about the nested menus functionality.

Mati365:
@scofalik 

> menu/definition directory - please put all things into one file, they are all mostly used only by the factory util, with I thing just one exception. Remove "definition" factory and move the singular file into menu/ directory.

👍🏻 done

> Change the factory class name and file name to something like DropdownMenuFactory -- it does not create a definition, it creates a menu.

👍🏻 done

> Same with getCurrentlyVisibleMenusElements().

At this moment it's kinda crucial to get all DOM elements that belong to the menu. Without such helper it's impossible to render dropdown menu inside dropdown. Check this place in additional PR that implements AI support: https://github.com/cksource/ckeditor5-commercial/pull/6213/files#diff-c3e2cfc5f2b64e23fc4bb25cdb3e76e1615a85a1fd0fac616a71d2271cb1456cR165

> preloadAllMenus() seems not to be used anywhere besides tests (at least that what my IDE tells me).

As we discussed, it's used in search implementation right [here](https://github.com/ckeditor/ckeditor5/pull/16314/files#diff-41402c4b4c3d967f87213503412bb1476326838b31908c4cd7c12a1b0aa8ddcaR88). `walk` function cannot walk through menu that has collapsed menu entries.

> To build .menus I am pretty sure we can do it with a simple tree-search algorithm that uses while and an array to store results. We won't need to use several utils, types and files.

While I agree that this solution is good enough in terms of making tree flat, or simply iteration over such tree, it makes it difficult to perform search using such approach. We need to know when we enter to the node, when we leave, to construct proper search tree. Please check this [file](https://github.com/ckeditor/ckeditor5/pull/16314/files#diff-2f1df8d7d720b80b30793ff43ee07af6724c82d08f4e36d9655e13c52d33c443R183).

Joshuajrodrigues:
Damn this would be good, currently looking for exactly this

Mati365:
@Joshuajrodrigues It's already present on master and was deployed some time ago. In a little bit of different form (without search bar).

PR #16636, opened 2024-06-28
List upcast converters should not remove invalid nodes.

### LINKED ISSUE #16450: CKEditor5 rendering is stripping nested HTML content, providing data loss for user

1. Have CKEditor5 render the following payload: 

```
<div>Dear Andy Coffey,</div><div>&nbsp;</div><div><span style=\"color:#000000; display:inline!important; float:none; font-family:&quot;Times New Roman&quot;; font-size:medium; font-style:normal; font-weight:400; letter-spacing:normal; text-align:start; text-decoration-color:initial; text-decoration-style:initial; text-indent:0px; text-transform:none; word-spacing:0px\">Thereference case number for this solution is:<span>&nbsp;</span></span><span style=\"font-size:11.0pt\"><span style=\"line-height:107%\"><span style=\"font-family:&quot;Calibri&quot;,sans-serif\"><span style=\"color:black\">3000373510</span></span></span></span></div><div><span style=\"color:#000000; display:inline!important; float:none; font-family:&quot;Times New Roman&quot;; font-size:medium; font-style:normal; font-weight:400; letter-spacing:normal; text-align:start; text-decoration-color:initial; text-decoration-style:initial; text-indent:0px; text-transform:none; word-spacing:0px\">Byanalyzing the system service logs uploaded by you we propose the following solution:</span></div><div>&nbsp;</div><div> <div><b>Action Plan:</b></div><br> We found 1 issue in FFDC file (7Z71CTO1WW_J30A09H2_xcc_mini-log_20240509-144330.zip):<br> <ul> <li> One event occurred on CMOS Battery at 2024-05-09T10:26:44Z.</li> </ul> We suggest following actions:<br><br> For event FQXSPPW0031J, \"Numeric sensor CMOS Battery going low (lower non-critical) has asserted.\", 2024-05-09T10:26:44Z<br> <ul> Complete the following steps until the problem is solved:<br>1. Remove the CMOS battery for 20 seconds and then install it back.<br>2. Replace the system CMOS battery<br>3. If the issue persists, please choose \"NO\" in the subsequent emails.</ul><br> </div><div>&nbsp;</div><div><strong style=\"color:#000000; font-family:&quot;Times New Roman&quot;; font-size:medium; font-style:normal; letter-spacing:normal; text-align:start; text-decoration-color:initial; text-decoration-style:initial; text-indent:0px; text-transform:none; word-spacing:0px\"><em>Wehope that by completing the above steps, the issue will be resolved. We will send you a follow-up email in 3 days from now to check the status. Please reply to the follow-up email and we will process your case based on the response selected.</em></strong></div>
```

2. Notice above payload is invalid HTML here:
image
3. CKEditor5 renders the above payload like below:
image

### ✔️ Expected result

CKEditor5 rendering engine should not strip invalid `<ul>` content without `<li>` elements. If data absolutely has to be stripped, user should be notified.

### ❌ Actual result

CKEditor5 rendering engine removes the entire `<ul>` content. This presents a data loss for the user, and is a regression compared to CKEditor4, which did not do this.

### ❓ Possible solution

Do not strip `<ul>` content that does not have `<li>` child elements (same applies to additional nesting elements like tables too). CKEditor4 did not have this behavior.

## 📃 Other details

* Browser: …
* OS: …
* First affected CKEditor version: CKEditor5

---

niegowski:

Fix (list): Text nodes directly inside `<ul>` or `<ol>` should not be removed while loading editor data. Closes #16450.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/src/list/converters.ts
+++ b/packages/ckeditor5-list/src/list/converters.ts
@@ -121,29 +121,6 @@ export function listItemUpcastConverter(): GetCallback<UpcastElementEvent> {
 }
 
 /**
- * Returns the upcast converter for the `<ul>` and `<ol>` view elements that cleans the input view of garbage.
- * This is mostly to clean whitespaces from between the `<li>` view elements inside the view list element. However,
- * incorrect data can also be cleared if the view was incorrect.
- *
- * @internal
- */
-export function listUpcastCleanList(): GetCallback<UpcastElementEvent> {
-	return ( evt, data, conversionApi ) => {
-		if ( !conversionApi.consumable.test( data.viewItem, { name: true } ) ) {
-			return;
-		}
-
-		const viewWriter = new UpcastWriter( data.viewItem.document );
-
-		for ( const child of Array.from( data.viewItem.getChildren() ) ) {
-			if ( !isListItemView( child ) && !isListView( child ) ) {
-				viewWriter.remove( child );
-			}
-		}
-	};
-}
-
-/**
  * Returns a model document change:data event listener that triggers conversion of related items if needed.
  *
  * @internal
--- a/packages/ckeditor5-list/src/list/listediting.ts
+++ b/packages/ckeditor5-list/src/list/listediting.ts
@@ -45,7 +45,6 @@ import {
 	listItemDowncastConverter,
 	listItemDowncastRemoveConverter,
 	listItemUpcastConverter,
-	listUpcastCleanList,
 	reconvertItemsOnDataChange
 } from './converters.js';
 import {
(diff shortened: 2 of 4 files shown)


Witoso:
Lost items AFAICS wrap in `p`, so we can quickly avoid data loss situation, and just have data change. @Reinmar do you have any comments/historical context. IMHO works well enough to merge.

arkflpc (approved):

illia-stv (approved):

Reinmar (approved):
This change makes the content being preserved, but those incorrect pieces of HTML will split the surrounding lists (CKE4 would "auto-li" them).

Still, it's a simple change now while aligning perfectly to CKE4 might be much harder, so let's go for simplicity.

Issue #16646, opened 2024-07-01
v42 - Vite dev build fails using Yarn PnP


greggmoritz:
## 📝 Provide detailed reproduction steps (if any)

1. Upgrade to CKEditor5 v42 per instructions on https://ckeditor.com/docs/ckeditor5/latest/updating/nim-migration/predefined-builds.html (using `CKEditor` component from `@ckeditor/ckeditor5-react`)
2. Run vite dev build

### ✔️ Expected result

Vite dev builds/runs successfully

### ❌ Actual result

Vite dev build failed with 168 instances of the same error, but with different instances. Below is an example of just one instance of the error in the CLI output

```
[0]   VITE v5.3.0  ready in 1186 ms
[0]
[0]   ➜  Local:   http://localhost:3100/subdirectory
[0]   ➜  Network: use --host to expose
[0] ✘ [ERROR] Could not resolve "@ckeditor/ckeditor5-ui/dist/index.js"
[0]
[0]     ../.yarn/cache/@ckeditor-ckeditor5-alignment-npm-42.0.0-7db2aabb44-ad212925b0.zip/node_modules/@ckeditor/ckeditor5-alignment/dist/index.js:7:159:
[0]       7 │ ...nuListItemButtonView } from '@ckeditor/ckeditor5-ui/dist/index.js';
[0]         ╵                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[0]
[0]   The Yarn Plug'n'Play manifest forbids importing "@ckeditor/ckeditor5-ui" here because it's not listed as a dependency of this package:
[0]
[0]     ../.pnp.cjs:2831:31:
[0]       2831 │         "packageDependencies": [\
[0]            ╵                                ~~
[0]
[0]   You can mark the path "@ckeditor/ckeditor5-ui/dist/index.js" as external to exclude it from the bundle, which will remove this error and leave the unresolved path in the bundle.
[0]
[0] /path/to/yarncache/.yarn/unplugged/esbuild-npm-0.21.5-d85dfbc965/node_modules/esbuild/lib/main.js:1472
[0]   let error = new Error(text);
[0]               ^
[0]
[0] Error: Build failed with 1 error:
[...]

PR #16649, opened 2024-07-02
Add dependencies used in the new `dist` folder as production `dependencies`

filipsobol:

Fix: Add dependencies used in the new `dist` folder as production `dependencies` instead of `devDependencies`. Related to #16646.

---

### Additional information

Blocked by https://github.com/ckeditor/ckeditor5-dev/pull/964.


pszczesniak (commented):
As we discussed internally, proabably best solution is to have the `dependency-checker` task run while preparing the release. This is caused by this, that each package must be built and has to had it's own `dist` folder. As we know it's not obvious to force it locally, so the best place is to do it on CI while release process is firing.

When some package has not got a `dist` folder - results of `dependency-checker` will be wrong.

cc @pomek

filipsobol:
Will be continued in #16749.

PR #16661, opened 2024-07-03
Fixed the creation of fake markers in a sentence that contains an empty line

### LINKED ISSUE #16660: Creating a link crashes when the selected sentence contains an empty line

1. Open any demo with the Link plugin.
2. Add an empty line (Shift+Enter).
3. Select a sentence with the empty line (empty line must be above this sentence).
4. Click the link button from the toolbar.
5. Close the link balloon.

### ✔️ Expected result

The balloon closes without errors.

### ❌ Actual result

There is an error in the console: `TypeError: removeHighlightCallback is not a function` and the editor crashes.

https://github.com/ckeditor/ckeditor5/assets/20561161/b9cd9eaa-1ae5-4b45-b1d2-81ec79279044

### ❓ Possible solution

## 📃 Other details

* Browser: …
* OS: …

---

DawidKossowski:

Fix (link): Creating a link on sentences containing an empty line will no longer cause a crash. Closes #16660.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
+++ b/packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
@@ -18,7 +18,6 @@ import ViewContainerElement from '../../src/view/containerelement.js';
 import ViewUIElement from '../../src/view/uielement.js';
 import ViewText from '../../src/view/text.js';
 import ViewDocument from '../../src/view/document.js';
-import ViewEmptyElement from '../../src/view/emptyelement.js';
 
 import testUtils from '@ckeditor/ckeditor5-core/tests/_utils/utils.js';
 
@@ -4652,45 +4651,6 @@ describe( 'DowncastHelpers', () => {
 					} );
 
 					expect( viewToString( viewRoot ) ).to.equal( '<div><div>foo</div></div>' );
-				} );
-
-				it( 'should not crash when element has no addHighlight and removeHighlight properties', () => {
-					downcastHelpers.elementToElement( {
-						model: 'br',
-						view: () => {
-							const viewContainer = new ViewEmptyElement( viewDocument, 'br' );
-
-							return viewContainer;
-						}
-					} );
-
-					downcastHelpers.markerToElement( { model: 'marker', view: 'marker' } );
-
-					const modelElement = new ModelElement( 'br', null );
-
-					model.change( writer => {
-						writer.remove( markerRange );
-
-						writer.insert( modelElement, modelRootStart );
-					} );
-
-					model.change( writer => {
-						writer.addMarker( 'marker', { range: markerRange, usingOperation: false } );
-					} );
-
-					expect( viewToString( viewRoot ) ).to.equal(
-						'<div>' +
-						'<marker></marker>' +
-						'<br></br>' +
-						'<marker></marker>' +
-						'</div>'
-					);
-
-					model.change( writer => {
-						writer.removeMarker( 'marker' );
-					} );
-
-					expect( viewToString( viewRoot ) ).to.equal( '<div><marker></marker><br></br><marker></marker></div>' );
 				} );
 			} );
 		} );
--- a/packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
+++ b/packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
@@ -18,6 +18,7 @@ import ViewContainerElement from '../../src/view/containerelement.js';
 import ViewUIElement from '../../src/view/uielement.js';
 import ViewText from '../../src/view/text.js';
 import ViewDocument from '../../src/view/document.js';
+import ViewEmptyElement from '../../src/view/emptyelement.js';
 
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-engine/tests/conversion/downcasthelpers.js
```diff
@@ -4652,6 +4653,45 @@ describe( 'DowncastHelpers', () => {
...
+						'<br></br>' +
+						'<marker></marker>' +
+						'</div>'
+					);
+
+					model.change( writer => {
+						writer.removeMarker( 'marker' );
+					} );
+
+					expect( viewToString( viewRoot ) ).to.equal( '<div><marker></marker><br></br><marker></marker></div>' );
```
DawidKossowski, line 4693: This result looks incorrect, but I didn't find a better way than using the 'incorrect' helper callbacks to test this logic:
```				
if ( removeHighlightCallback ) {
	removeHighlightCallback( element, descriptor.id!, conversionApi.writer );
} 
```

@scofalik, maybe you will have a better idea.

DawidKossowski: Anyway, it exactly reproduces the issues that we found in the AI and Link plugins.

scofalik: As you know, I wasn't sure if we should add it, and I wasn't sure how you will test it.

AFAICS, this can crash in two scenarios:

1.  An element has `addHighlight` but does not have `removeHighlight`.
2.  An element is handled by "remove highlight helper" while it shouldn't.

Scenario 1. is just simply incorrectly written element implementation. If something has `addHighlight` it must have corresponding `removeHighlight`. This could happen either in our code (some engine internals or plugins, like widget) or this could happen in 3rd party code. In any case, we should not care about it -- there are million ways to crash editor if you implement something incorrectly.

For scenario 2. we should think why this happen. If there is a reasonable and correct-ish code that leads to this error we should implement it here.

With the \<br> case, first question is: should it have `addHighlight` and `removeHighlight`? It can be argued, but for now let's say that no. It's like that right now. \<br> has neither property, so that's not the scenario 1.

Then let's see why it crashed.

Looking at the "remove highlight helper" code:

```javascript
// Get all elements bound with given marker name.
const elements = conversionApi.mapper.markerNameToElements( data.markerName );

if ( !elements ) {
	return;
}

for ( const element of elements ) {
	conversionApi.mapper.unbindElementFromMarkerName( element, data.markerName );

	if ( element.is( 'attributeElement' ) ) {
		conversionApi.writer.unwrap( conversionApi.writer.createRangeOn( element ), viewHighlightElement );
	} else {
		// if element.is( 'containerElement' ).
		const removeHighlightCallback = element.getCustomProperty( 'removeHighlight' ) as RemoveHighlightCallback;

		removeHighlightCallback( element, descriptor.id!, conversionApi.writer );
	}
}
```

Note that, `elements` here are view elements that are the results of "add marker" conversion. 

I think that the `if` here is a bit optimistic. The non-obvious assumption here is that `elements` could have been added only by "add highlight helper" (`highlightText` or `highlightElement`). And in this case, it's either attribute element (created by `highlightText`); or `highlightElement` mapped the original element and we require it to have `'removeHighlight'` property (as per the very first requirement).

In our case though, an element was added to the mapper (was added to `elements`) from an "outside" converter and there's no handling for it here.

Note, that `highlightText`, `highlightElement` and `removeHighlight` functions are private and are only a part of a bigger conversion helper, so we should only consider them in the context of this conversion helper. We should not think that somebody will use only one of them, they are always in tandem. With this context, it made sense to be optimistic with `removeHighlight`. If the bigger conversion helper triggered for adding marker, it should also handle removing marker.

> In our case though, an element was added to the mapper (was added to `elements`) from an "outside" converter and there's no handling for it here.

The original problem happened because there were two conversion helpers for the same marker. So let's say we have conversion helpers A and B defined for a marker. Right now, I'd make opinion that if conversion helper A managed adding, then conversion helper A should fully and correctly managed removing. Helper B should not fire for either adding nor removing.

With this way of thinking, we don't need to add this `if` here at all.

If there is conversion other than `markerToHighlight`, then it should be implemented so that both converters are fired in excluding editor state (in other words, they are never fired together); or we must make sure that one fully overwrites the other one (i.e. adding A is fired before adding B; removing A is fired before removing B; adding A properly consumes and adding B does nothing; and removing A fully removes elements from mapper).

I am afraid that the only way to test this `if` is to write incorrect converters. And although it is easy to make this mistake, it tells me that we should not add this `if`. You can correct me if I am wrong, but I guess you have the same feeling.

DawidKossowski: Yes, I agree with you. I just wanted to add this `if` to avoid editor crashes, but as you said, we will never face this case with correct converters. Besides, this `if` doesn't fix the whole problem. So, I will remove it.

scofalik (approved):

Issue #16664, opened 2024-07-03
Selecting a color via the color picker in table styling tools throws an error


mabryl:
## 📝 Provide detailed reproduction steps (if any)

1. Select a table or table cell
2. Try to change its background color via the color picker

### ✔️ Expected result

The selected color should be applied, as is the case in v41.3.1 and earlier:

https://github.com/ckeditor/ckeditor5/assets/72079603/c87af206-8caf-4404-9842-4c319db3b8d8

### ❌ Actual result

Starting with v41.4.1 an error gets thrown:

https://github.com/ckeditor/ckeditor5/assets/72079603/7488ab2b-e845-4be2-948b-6e82319ea75d

Error description:

```
Uncaught TypeError: Cannot read properties of null (reading 'value')
    at Vg.getParsedColor (colorpickerview.js:297:66)
    at Eg.isValid (colorpickerview.js:199:31)
    at Qu.<anonymous> (colorpickerfragmentview.js:185:38)
    at Qu.fire (emittermixin.js:146:47)
    at oo.<anonymous> (buttonview.js:106:30)
    at En.n (template.js:802:26)
    at En.fire (emittermixin.js:146:47)
    at HTMLButtonElement.t (emittermixin.js:208:18)
```

## 📃 Other details

* Browser: cross-browser
* OS: cross-OS
* First affected CKEditor version: 41.4.1
* Installed CKEditor plugins: Table, Table Styles

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Mati365:
**Debug notes:**

1. Caused by this [commit](https://github.com/ckeditor/ckeditor5/commit/caea11e431fc56f911f4cf4ad4f73bc74d36a8b9#diff-1594461a97d09ec34385b4bd65b6d02f1426d81063af97cbdd197ae2941f8635R303).
2. I missed edge-case where `isValidation` is called after hiding of popup with form.

PR #16673, opened 2024-07-04
No longer crash after selecting color in table cell formatting form

Mati365:

Fix (ui): No longer crash while selecting color in table cell formatting form. Closes https://github.com/ckeditor/ckeditor5/issues/16664

---

### Additional information

1. Caused by this [commit](https://github.com/ckeditor/ckeditor5/commit/caea11e431fc56f911f4cf4ad4f73bc74d36a8b9#diff-1594461a97d09ec34385b4bd65b6d02f1426d81063af97cbdd197ae2941f8635R303).
2. I missed edge-case where `isValid` is called on color selector that has `hideInput` config flag set to `true`.
3. It means that `isValid` was called on a non-existing hex input DOM node, and it raised an exception. 
4. I added regression tests.

### DIFF
--- a/packages/ckeditor5-ui/tests/colorpicker/colorpickerview.js
+++ b/packages/ckeditor5-ui/tests/colorpicker/colorpickerview.js
@@ -397,6 +397,8 @@ describe( 'ColorPickerView', () => {
 			view.render();
 
 			expect( view.isValid() ).to.be.true;
+
+			view.destroy();
 		} );
 	} );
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-ui/tests/colorpicker/colorpickerview.js
```diff
@@ -372,6 +372,34 @@ describe( 'ColorPickerView', () => {
...
+
+			expect( view.isValid() ).to.be.false;
+		} );
+
+		it( 'should return true if the hex input is not shown', () => {
+			const view = new ColorPickerView( locale, { format: 'hex', hideInput: true } );
+
+			view.render();
+
+			expect( view.isValid() ).to.be.true;
```
niegowski, line 399: I'd expect the view to get `destroy()`-ed after use.

niegowski (changes requested):
As commented. I'd also add a test for the original issue in the table feature.

PR #16677, opened 2024-07-05
Docs: Update the Angular version table.

gorzelinski:

Docs: Update the Angular version table.

---

### Additional information

### DIFF
--- a/docs/getting-started/integrations/angular.md
+++ b/docs/getting-started/integrations/angular.md
@@ -83,7 +83,7 @@ Because of the breaking changes in the Angular library output format, the `@cked
 	<tr>
 	 <td><code>^1</code></td>
 	 <td><code>5.x&nbsp;-&nbsp;8.x</code></td>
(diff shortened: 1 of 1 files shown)


gorzelinski:
I tested different versions of all packages: our Angular component, CKEditor, and Angular itself. Our Angular component in version 7 seems to work with CKEditor 5 in ver. 37+. I wasn't able to marry version 8 with the CKEditor in ver. 37. There is a problem with types. That's probably because of changes in dependencies. Surprisingly, all setups seem to work with Angular as low as 13. Initially, the styling wasn't working, but I solved the problem by importing CSSes differently.

godai78 (approved):
LGTM

filipsobol:
>  Initially, the styling wasn't working, but I solved the problem by importing CSSes differently.

Do you think others might encounter the same problem? If so, it might be worth adding to the docs.

filipsobol (approved):

PR #16685, opened 2024-07-08
Block toolbar buttons do not scroll with editor content if block toolbar is added manually

### LINKED ISSUE #5460: BlockButton fails to reposition on scroll inside overflow: scroll element.
Followup https://github.com/ckeditor/ckeditor5-ui/issues/391.

See https://github.com/ckeditor/ckeditor5-ui/pull/392#issuecomment-392538832.

kapture 2018-05-28 at 16 14 43

Mati365:

Fix (ui): Block toolbar button no longer remains fixed in the same position while scrolling the editable content. Closes #5460.

---

### Additional information

Closes: ckeditor/ckeditor5-builder#676

1.  It closes this builder issue: ckeditor/ckeditor5-builder#676.
2.  I added manual tests in `ui` called `scrollable-blocktoolbar` 

#### Screens

**Before**

https://github.com/ckeditor/ckeditor5/assets/3949262/d9a6bc7d-8009-4c92-b171-26914bd56baf

**After**

https://github.com/ckeditor/ckeditor5/assets/3949262/1be91142-4bc4-4832-9404-5c7da09f9b5d


godai78:
Appears to be working correctly, now.

godai78 (approved):

Reinmar:
> It should also close this [issue](https://github.com/ckeditor/ckeditor5-builder/issues/676).

Should or does? 😄 If it does, please add it to "Closes" so we know that this PR should also be tested against that issue.

As for manual testing:

*   Am I right that this PR requires quite a lot of testing in probably not the most obvious HTML+CSS setups? Scrollable containers, scrollable page, vertical/horizontal scrolls, different ways of applying these styles. How and where did you test it? I can't see any new manual test but if there was one on which this issue is clearly reproducible, I guess we'd know about it. So, am I right that we're missing a manual test where such issues are reproducible (both of the closed ones)? If so, please create one.
*   In order to enable the QA team to look into this, providing the manual tests as well as potential directions to what to explore may help.

Reinmar:
One more thing for the QA team: Could you test this solution on a larger page with a lot of content and see if there's some impact on performance? Testing it in some sample in the docs might be a good idea, but demos on ckeditor.com might be even better.

Mati365:
@Reinmar 

> Should or does? 😄 If it does, please add it to "Closes" so we know that this PR should also be tested against that issue.

I updated description. I'm not sure if we should add it to `closes` commit message because that issue belongs to internal builder repo, so I placed it in `Additional Information`. 

> As for manual testing: [...]

Ok, it's only about manually adjust height of block toolbar container. I'll add manual test to cover that.

Reinmar:
And one more thing for clarity: This PR adds "clipping" of the block toolbar icon. So, when it reaches the edge of a scrollable container, we start to cut off the part that overflows its vertical boundaries. This is a very interesting approach. It wasn't necessarily in the scope of this issue, but seems fine.

@oleq, I think this deserves your attention :) Question I'd have to you is: do you see any significant risks (apart from performance) in this approach?

Reinmar:
OK, I found the first issue. A clipped button is not clickable:



You cannot click it now. It does not react to hover.

Meaning – it's kind of useless and confusing this way.

Which makes the clipping a bit pointless :( In my opinion, it'd be better, to be honest, to just hide the button once it cannot be clicked.

Mati365:
@Reinmar It was intended to be clickable, so this line seems broken somehow. Thanks for pointing out, I'll check that.

Mati365:
@Reinmar Indeed, this line was wrong. It was only about inversion of compare. I slightly renamed this method and made it more obvious. Should be working properly. 

One thing besides that - I forgot about A11y. I adjusted PR to change `isEnabled` state of button to mark button outside of viewport as `aria-hidden`. It'll make it a bit more screen reader friendly.

lkszzajac:
With this setup of editor elements:
```html
<div style="max-height: 300px; max-width: 300px; overflow: auto;">
	<div id="editor-scrollable">
	...
	</div>
</div>
```
the issue is still reproducible. I suspect that there might be setups where the "bounding" element is not the editor itself but something higher up in the element tree, is it possible to calculate the viewport in these cases?

Mati365:
@lkszzajac Nice catch! I managed to fix that in a recent [commit](https://github.com/ckeditor/ckeditor5/pull/16685/commits/3df30aba93125e24cc741373eec137e08fcfd5a3). Can you take a look?

illia-stv (approved):

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -111,6 +115,11 @@ export default class BlockToolbar extends Plugin {
 	 */
 	private _blockToolbarConfig: ReturnType<typeof normalizeToolbarConfig>;
 
+	/**
+	 * DOM Emitter.
+	 */
+	private _domEmitter: DomEmitter = new ( DomEmitterMixin() )();
```
niegowski, line 121: This is not used. Is this needed? Maybe we should use this emitter instead of `this.buttonView` as an emitter?

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -425,21 +439,59 @@ export default class BlockToolbar extends Plugin {
...
+		let pendingAnimationFrame = false;
+
+		// Reposition the button on scroll, but do it only once per animation frame to avoid performance issues.
+		const repositionOnScroll = () => {
+			if ( pendingAnimationFrame ) {
+				return;
+			}
+
+			pendingAnimationFrame = true;
+			window.requestAnimationFrame( () => {
```
niegowski, line 457:
```suggestion
			global.window.requestAnimationFrame( () => {
```

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -458,6 +510,83 @@ export default class BlockToolbar extends Plugin {
...
+	}
+
+	/**
+	 * Clips the button element to the viewport of the editable element.
+	 *
+	 * 	* If the button overflows the editable viewport, it is clipped to make it look like it's cut off by the editable scrollable region.
+	 * 	* If the button is fully hidden by the top of the editable, it is not clickable but still visible in the DOM.
+	 *
+	 * @param buttonView - The button view to clip.
+	 * @param editableElement - The editable element whose viewport is used for clipping.
```
niegowski, line 524:
```suggestion
	 * @param buttonView The button view to clip.
	 * @param editableElement The editable element whose viewport is used for clipping.
```

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -458,6 +510,83 @@ export default class BlockToolbar extends Plugin {
...
+	 */
+	private _clipButtonToViewport(
+		buttonView: ButtonView,
+		editableElement: HTMLElement
+	) {
+		const absoluteButtonRect = buttonView.element!.getBoundingClientRect();
```
niegowski, line 530: Maybe we could use our [`Rect`](https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-utils/src/dom/rect.ts#L24) wrapper.

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -458,6 +510,83 @@ export default class BlockToolbar extends Plugin {
...
+		const scrollableElement = closestScrollableParent( editableElement );
+
+		if ( !scrollableElement ) {
+			return;
+		}
+
+		// Get the viewport of the scrollable element. It's not guaranteed that the scrollable element is the same as the editable element.
+		// Sometimes the editable is inside a scrollable element, so we need to get the viewport of the scrollable parent element.
+		const scrollViewportRect = scrollableElement.getBoundingClientRect();
```
niegowski, line 539: Maybe we could use our [`Rect`](https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-utils/src/dom/rect.ts#L24) wrapper.

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -458,6 +510,83 @@ export default class BlockToolbar extends Plugin {
...
+
+		// Sets polygon clip path for the button element, if empty string is passed, the clip path is removed.
```
niegowski, line 541: This is not described as this function behaves.

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -497,3 +626,39 @@ export default class BlockToolbar extends Plugin {
 		return toPx( editableRect.width + offset );
 	}
 }
+
+/**
+ * Returns the closest scrollable parent element of the given element.
+ *
+ * @param element - The element for which to find the closest scrollable parent.
```
niegowski, line 633:
```suggestion
 * @param element The element for which to find the closest scrollable parent.
```

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -497,3 +626,39 @@ export default class BlockToolbar extends Plugin {
...
+		parent = parent.parentElement;
+	}
+
+	return parent;
+}
+
+/**
+ * Checks if an element is scrollable.
+ *
+ * @param element - The HTML element to check.
```
niegowski, line 659:
```suggestion
 * @param element The HTML element to check.
```

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -497,3 +626,39 @@ export default class BlockToolbar extends Plugin {
...
+	if ( isScrollable( element ) ) {
+		return element;
+	}
+
+	let parent = element.parentElement;
+	let depth = 0;
+
+	// Check up to 6 levels of parent elements to prevent performance issues.
+	while ( parent && !isScrollable( parent ) ) {
+		if ( depth++ >= 6 ) {
```
niegowski, line 646: Why such an arbitrary number? This could be hard to debug if it happens. If the performance is an issue we could store it and not search for it every time. Did you notice some performance issues without this limitation or is it an early optimization without real data that it is required?

niegowski (changes requested):
I'd suggest movin the manual tests to the UI package as it does not require rebuilding the build and also is not related to the build itself but the feature provided by the UI package.

Mati365:
@niegowski I applied your CR suggestions.

1. I moved manual tests to the UI 
2. Fixed incorrect comments
3. Removed `maxDepth` checks in the closest scrollable parent

After removal of `maxDepth`, I discovered edge case related to calculation of closest scrollable parent. It was returning `html` element as scrollable for editor containers placed within `position: absolute` element. While it's true in theory, it was buggy in our case. `position: absolute` container might have bigger height than the whole `html` element, which was causing incorrect clipping of the button. Manual and units tests should cover that. 

Can you check again?

packages/ckeditor5-ui/src/toolbar/block/blocktoolbar.ts
```diff
@@ -630,35 +621,46 @@ export default class BlockToolbar extends Plugin {
...
-function closestScrollableParent( element: HTMLElement ): HTMLElement | null {
-	if ( isScrollable( element ) ) {
-		return element;
-	}
+function closestScrollableElement( element: HTMLElement ): HTMLElement | null {
+	let currentElement: HTMLElement | null = element;
 
-	let parent = element.parentElement;
-	let depth = 0;
+	do {
```
niegowski, line 630: I wonder if some of this logic could be replaced with methods of the `Rect`, for example this one: https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-utils/src/dom/rect.ts#L254.

Mati365: I'll check that.

Mati365: Hm, it looks like it works ootb!

niegowski (approved):

Mati365:
I'll merge it after QA retest.

charlttsie:
We have re-tested the PR with @marcelmroz and @kubaklatt, and it looks good. The initial issue doesn't reproduce anymore and we haven't found any regressions either 👍

Mati365:
@charlttsie Cool! I'll merge it.

PR #16694, opened 2024-07-09
CKBox plugins should not wait for token fetch.

Mati365:

Other (ckbox): The plugin no longer slows down the editor startup because it fetches the token in the background instead of during the editor’s initialization. Closes https://github.com/ckeditor/ckeditor5/issues/16760

MINOR BREAKING CHANGE (ckbox): The `CKBoxUtils#getWorkspaceId` and `CKBoxUtils#getToken` methods now return a promise instead of a resolved value.

### DIFF
--- a/packages/ckeditor5-ckbox/tests/ckboxuploadadapter.js
+++ b/packages/ckeditor5-ckbox/tests/ckboxuploadadapter.js
@@ -996,7 +996,7 @@ describe( 'CKBoxUploadAdapter', () => {
 				for ( const { testName, workspaceId, tokenClaims } of testData ) {
 					it( testName, async () => {
 						TokenMock.initialToken = createToken( tokenClaims );
-						ckboxUtils._token.refreshToken();
+						( await ckboxUtils._token ).refreshToken();
 
 						sinonXHR.respondWith( 'GET', /\/categories/, [
 							200,
@@ -1035,9 +1035,9 @@ describe( 'CKBoxUploadAdapter', () => {
 			} );
 
 			describe( 'defaultUploadWorkspaceId is defined', () => {
-				it( 'should use the default workspace', () => {
+				it( 'should use the default workspace', async () => {
 					TokenMock.initialToken = createToken( { auth: { ckbox: { workspaces: [ 'workspace1', 'workspace2' ] } } } );
-					ckboxUtils._token.refreshToken();
+					( await ckboxUtils._token ).refreshToken();
 
 					sinonXHR.respondWith( 'GET', /\/categories/, [
 						200,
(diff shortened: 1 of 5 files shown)


illia-stv (approved):

packages/ckeditor5-ckbox/tests/ckboxuploadadapter.js
```diff
@@ -1137,9 +1137,9 @@ describe( 'CKBoxUploadAdapter', () => {
 						.then( () => {
 							throw new Error( 'Expected to be rejected.' );
 						}, err => {
-							expect( console.error.callCount ).to.equal( 1 );
+							expect( console.error.callCount ).to.equal( 2 );
```
niegowski, line 1140: Why has this changed?

packages/ckeditor5-ckbox/tests/ckboxuploadadapter.js
```diff
@@ -1137,9 +1137,9 @@ describe( 'CKBoxUploadAdapter', () => {
...
 							expect( console.error.firstCall.args[ 0 ] ).to.match( /^ckbox-access-default-workspace-error/ );
-							expect( err ).to.equal( 'Cannot access default workspace.' );
+							expect( err ).to.equal( 'Cannot determine a category for the uploaded file.' );
```
niegowski, line 1142: Why has this changed?

niegowski (commented):
I'd suggest adding a test for the original issue to make sure there is no regression in the future.

packages/ckeditor5-ckbox/tests/ckboxuploadadapter.js
Mati365: Hm, good question. I see that it should not be changed. but I think that `await this._token` might result in forwarding incorrect exception.

packages/ckeditor5-ckbox/tests/ckboxuploadadapter.js
Mati365: Reverted.

packages/ckeditor5-ckbox/tests/ckboxuploadadapter.js
Mati365: @niegowski I found an issue. The problem was that the exception should have been thrown before the `try {` block in `_getAvailableCategories`. After the changes, it was thrown inside the block, causing the exception to be caught in the wrong place.

Mati365:
> I'd suggest adding a test for the original issue to make sure there is no regression in the future.

Ok 👍🏻

niegowski (approved):

Mati365:
Asked QA for testing. @arkflpc Can you take a look?

arkflpc (approved):

charlttsie:
We have tested the PR with @marcelmroz and @kubaklatt and can confirm that the change noticeably reduces the editor's loading time when when fetching the token is delayed. At the same time, we haven't found any regressions, so the PR looks good to us 👍

Mati365:
@charlttsie Cool! Thanks!

PR #16696, opened 2024-07-09
Added './ckeditor5.css': './dist/ckeditor5.css' to the list of exports

djneely:
Including './ckeditor5.css': './dist/ckeditor5.css' to the list of exports for compatibility with some frameworks.

When integrating with Quasar (and other frameworks) there seemingly is a need to manually edit the node_mobuldes/ckeditor5/dist/package.json and add in the line "./ckeditor5.css": "./dist/ckeditor5.css" in order to expose the css for the editor to import into the component directly. Otherwise the editor's appearance is not displayed correctly.

### DIFF
--- a/scripts/release/utils/getckeditor5packagejson.js
+++ b/scripts/release/utils/getckeditor5packagejson.js
@@ -38,6 +38,7 @@ module.exports = function getCKEditor5PackageJson() {
 				'types': './dist/translations/*.d.ts',
 				'import': './dist/translations/*.js'
 			},
+			'./ckeditor5.css': './dist/ckeditor5.css',
 			'./*.css': './dist/*.css',
 			'./build/*': './build/*',
 			'./src/*': './src/*',


pomek:
cc @filipsobol

filipsobol:
I wonder why `'./*.css': './dist/*.css'`, which we already have, is not enough, since [Subpath patterns](https://nodejs.org/api/packages.html#subpath-patterns) are available since Node 12.

I'm against this change in the current state, because we're missing important information like:
1. Is this a problem with all versions of Quasar or just the old ones?
2. As far as I can tell from the Quasar docs, it can be installed with Vite, Webpack, "standalone" or as a Vite plugin. Does this issue affect all of these methods or just one/some of them?
4. If this problem is present in the latest version of Quasar, are the developers aware of this limitation and do they plan to fix it?
5. Since we have more CSS files and use this pattern for other directories (`./translations/*.js`, `./build/*`, `./src/*`), do we need to update them as well?

djneely:
@filipsobol please see below per your questions

We're currently working off of node 18.19.1

1) We're using the latest and greatest version of Quasar (2.16.5)
2) We're using vite on this current project
3) I'm assuming they're aware as this came from their discord where they supplied us this fix to get ckeditor to work
4) I'm not sure but the one line change added fixes it for the project
5) We've ran into this issue with a Vuetify project as well, and have recently been able to incorporate ckeditor5 into it successfully using this method. 

The line of code the errors out if we do not include this additional line in the package.json is: 
`import 'ckeditor5/ckeditor5.css';`

If there is an alternative means to include the css into the component such that this edit is not needed that would be great. We're curious too as to why the subpath pattern matching isn't working for this either.

filipsobol:
Thanks, this gives us some much needed context :pray: 

Could you create a minimal reproducible example that we can use to reproduce this issue and test other imports and fixes?

djneely:
Sure. Wrapping up a project so it may take me a day or two to get that over to you. Thanks for looking into this.

filipsobol:
This problem has been fixed in `42.0.2`.

djneely:
@filipsobol  the 42.0.2 release fixed this issue but is now back with the latest 43.0.0 release.

Witoso:
@djneely nothing between those two versions changed in the `exports` declaration. We also didn't change anything in the publication flow. Could you share more details or some reproducible sample?

djneely:
@Witoso I apologize for this. I was a bit hasty. It appears the issue is with the @ckeditor/ckeditor5-vue. I had upgraded both at the same time and saw the same issue as before. Again, apologies.

Witoso:
No worries, thank you for flagging!

PR #16707, opened 2024-07-10
Make code block outdent command work with elements

Dumluregn:

Feature (code-block): to do.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-code-block/src/outdentcodeblockcommand.ts
+++ b/packages/ckeditor5-code-block/src/outdentcodeblockcommand.ts
@@ -179,6 +179,19 @@ function getCodeLineTextNodeAtPosition( position: Position ): Text | null {
 		nodeAtPosition = position.nodeBefore;
 	}
 
+	// <codeBlock>foo <inlineElement></inlineElement>^</codeBlock>
+	// <codeBlock><inlineElement></inlineElement>^</codeBlock>
+	if ( nodeAtPosition && nodeAtPosition.is( 'element' ) && !nodeAtPosition.is( 'element', 'softBreak' ) ) {
+		while ( nodeAtPosition.previousSibling !== null && !nodeAtPosition.previousSibling.is( 'element', 'softBreak' ) ) {
+			nodeAtPosition = nodeAtPosition.previousSibling;
+		}
+
+		// <codeBlock><inlineElement></inlineElement>^</codeBlock>
+		if ( nodeAtPosition.is( 'element' ) ) {
+			return null;
+		}
+	}
+
 	// <codeBlock>^</codeBlock>
 	// <codeBlock>foo^<softBreak></softBreak>bar</codeBlock>
 	if ( !nodeAtPosition || nodeAtPosition.is( 'element', 'softBreak' ) ) {
--- a/packages/ckeditor5-code-block/src/outdentcodeblockcommand.ts
+++ b/packages/ckeditor5-code-block/src/outdentcodeblockcommand.ts
@@ -169,32 +169,44 @@ function getLastOutdentableSequenceRange( model: Model, position: Position, sequ
 	);
 }
 
+// For given position, find the closest position that is at the beginning of a line.
+// Beginning of a line is after a `softBreak` element or at the beginning of a code block.
+// Then, return a text node that is at the beginning of the line.
+// Even though code block doesn't allow other elements than `<softBreak>`, some features may overwrite this rule,
+// so they are taken into account.
+//
+// <codeBlock>foobar^</codeBlock>                         ->   <codeBlock>^[foobar]</codeBlock>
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-code-block/src/outdentcodeblockcommand.ts
```diff
@@ -179,6 +179,19 @@ function getCodeLineTextNodeAtPosition( position: Position ): Text | null {
 		nodeAtPosition = position.nodeBefore;
 	}
 
+	// <codeBlock>foo <inlineElement></inlineElement>^</codeBlock>
+	// <codeBlock><inlineElement></inlineElement>^</codeBlock>
+	if ( nodeAtPosition && nodeAtPosition.is( 'element' ) && !nodeAtPosition.is( 'element', 'softBreak' ) ) {
+		while ( nodeAtPosition.previousSibling !== null && !nodeAtPosition.previousSibling.is( 'element', 'softBreak' ) ) {
```
scofalik, line 185: Shouldn't this condition also check for text nodes? I mean, AFAIU, you want to jump left until you find:

*   start of code block - it means we were in the first line and there was no text before inline element(s),
*   soft break - it means that we are in 2+ line and there was no text before inline element(s),
*   text - it means there was some text before inline element(s) and we should take it.

Now it looks to me that you jump over text nodes until you find a soft break (if you find it).

So it seems to me this should work incorrectly for case like this:

`<codeBlock>    foo<inlineElement></inlineElement>^</codeBlock>`

If I am not mistaken, this will return `null` while it should return  `foo` text node, right?

packages/ckeditor5-code-block/tests/outdentcodeblockcommand.js
```diff
@@ -132,6 +132,30 @@ describe( 'OutdentCodeBlockCommand', () => {
 			expect( outdentCommand.isEnabled ).to.be.false;
 		} );
 
+		it( 'should not break with inline elements at the end of line', () => {
```
scofalik, line 135: I don't like these expects that we have here and the names for these tests. Make these tests analogical to the tests above. They should check if the command is enabled or not. And they should be named accordingly ("it should be false when ..."). Then, I think I am missing tests that show the positive scenario, i.e., there is a leading sequence in a line that includes inline element.

packages/ckeditor5-code-block/src/outdentcodeblockcommand.ts
```diff
@@ -179,6 +179,19 @@ function getCodeLineTextNodeAtPosition( position: Position ): Text | null {
 		nodeAtPosition = position.nodeBefore;
 	}
 
+	// <codeBlock>foo <inlineElement></inlineElement>^</codeBlock>
+	// <codeBlock><inlineElement></inlineElement>^</codeBlock>
```
scofalik, line 183: This is more complicated than it looks and I didn't understand it correctly at the first glance (and second glance). So I thought that it will be good to write a broader comment here. But then I thought that this whole function could be generalized and the first and the last `if` should not be needed.

So, let's generalize this function to: for given position, let's find the closest position that is at the beginning of a line, and then let's see if there is a text node after this position. I think this is what we want to do here regardless of the actual content (text, formatted text - I know it's not allowed, inline elements).

Beginning of the line is of course what you have in your while condition.

So I think your while loop (maybe with some tweaks) should also cover the first and the last `if` in this function. With a proper comment it may be even more understandable than the current comments. Like, there's no description for this function now and you need to check how it is used. So I think simply writing at the beginning something like I wrote above:

```
// For given position, find the closest position that is at the beginning of a line.
// Beginning of a line is after a `softBreak` element or at the beginning of a code block.
// Then, return a text node that is at the beginning of the line.
//
// <codeBlock>foobar^</codeBlock>                ->   <codeBlock>^[foobar]</codeBlock>
// <codeBlock>foo<softBreak />bar^</codeBlock>   ->   <codeBlock>foo<softBreak />^[bar]</codeBlock>
```

And a few more interesting examples. They probably won't fit in one line so you can go with:

```
// AAA
// AAA
//
// BBB
// BBB
//
// CCC
// CCC
```

scofalik: One more thing. I think it is important to write somewhere here that even though `codeBlock` disallows inline elements, we still consider them because some features may overwrite this disallow rule and put inline objects into code block anyway (e.g. merge fields).

scofalik: One more thing :smile:. How does this work for indent? I see you didn't make any changes in indent command but I am not sure if you checked scenarios where you have non-collapsed selection. In such cases you can still indent while having selection in the middle of a line so if there's similar algorithm for indent it could break too?

Also please check if multiline selections work well for indent and outdent.

PR #16744, opened 2024-07-17
Expose the source code from ckeditor5-demos and the builder

Reinmar:

Docs: Expose the source code from ckeditor5-demos and the builder.

---

### Additional information

This is to address the traffic to [https://github.com/ckeditor/ckeditor5/blob/master/docs/examples/builds/classic-editor.md](https://github.com/ckeditor/ckeditor5/blob/master/docs/examples/builds/classic-editor.md)

14% of the visitors of the [https://ckeditor.com/docs/ckeditor5/latest/examples/builds/classic-editor.html](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/classic-editor.html) page end up there.

### Work needed

If this change makes sense, it needs to be applied to the other 4 examples in that section (inline, balloon, balloon block, document). The other examples have their source code inline which is fine for now.

In the long-term, once we'll have more presets in the builder and ability to open a specific setup there, we'll be linking to the builder only


godai78:
We mentioned this today on planning, there's a suggestion to put it over the demo, as it's not on the immediate screen upon entering the example page.

gorzelinski:
> We mentioned this today on planning, there's a suggestion to put it over the demo, as it's not on the immediate screen upon entering the example page.

Exactly, I can make the changes.

Witoso:
AFAICS, it's already above the snippet.

Reinmar:
> AFAICS, it's already above the snippet.

Yes. The diff is unclear, but I moved it above indeed :)

godai78:
By the way, don't we want it on `stable` so that goes on production immediately?

godai78:
We can also turn off the GitHub top-right icon link in these guides, by the way.

gorzelinski (commented):
Everything is ok. But I wonder, were some editor demos always narrower than the content? Affected demos are:

* Inline editor
* Balloon editor
* Balloon block editor

Screenshot 2024-07-17 at 12 30 55

godai78:
> *   Balloon editor
> *   Balloon block editor  
>      

These two need to be narrower to accommodate the toolbar handle. Not sure about the inline, tho. Also, I believe they were that way already, not a recent change.

gorzelinski:
> > * Balloon editor
> > * Balloon block editor
> 
> These two need to be narrower to accommodate the toolbar handle. Not sure about the inline, tho. Also, I believe they were that way already, not a recent change.

I thought so. So, LGTM. I would also remove this top GitHub link if it confuses users.

Issue #16739, opened 2024-07-17
Image contextual toolbar does not disappear after clicking source editing.


blucayzer:
## 📝 Provide detailed reproduction steps (if any)

1. Have a CKEditor 5 Classic Editor.
2. Insert an image via URL
3. Click SourceEditing.

This bug can be replicated if the first HTML tag is a figure. If the first HTML tag isn't a figure, the image contextual toolbar will disappear.

### ✔️ Expected result
![image](https://github.com/user-attachments/assets/d409a1bd-c54b-4d30-9392-d315523ba131)

The Image contextual toolbar should disappear after clicking the Source Editing.

### ❌ Actual result
![image](https://github.com/user-attachments/assets/2261401b-ff91-4371-b877-0af572e4f1f5)

The Image contextual toolbar does not disappear after clicking the Source Editing.

## 📃 Other details

* Browser: Edge
* CKEditor Version: v39
* This bug can be seen in the latest version of CKEditor.
* Can be replicated in the CKEditor Builder as seen in the screenshot above
* Installed CKEditor plugins: Source Editing and Image

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Witoso:
Thanks for the report! Not reproducible in the docs, reproducible in the Builder and output (checked stackblitz).

Witoso:
**@blucayzer** do I understand correctly you reproduce it in v39?

blucayzer:
@Witoso, yes.

Mati365:
Looks like it's not related to view-source functionality but rather blur handling in toolbar. I reproduced it with menubar.

![obraz](https://github.com/user-attachments/assets/2de5eb6a-bd1c-4a0f-bba3-d52e4bfc876f)

Mati365:
I was able to reproduce this issue using the table toolbar, so it appears to be a global problem. It’s essential to ensure that the element with the toolbar is the first one in the document.

![obraz](https://github.com/user-attachments/assets/7eb3c0d3-e672-4c01-9490-b819e075861d)

PR #16750, opened 2024-07-17
Close toolbars on open or close source view

Mati365:

Fix (ui): Hide the pinned balloon panel view when the pin target is no longer visible. Closes https://github.com/ckeditor/ckeditor5/issues/16739. 

#### Before

https://github.com/user-attachments/assets/7c859ea6-8d94-49b8-a4c4-31ad788ae384

#### After

https://github.com/user-attachments/assets/98bc68b0-7be6-4fe4-bc8a-1b499a329228

#### Debug info

1. `SourceEditing` fires `change:isSourceEditingMode` after toggling source editing mode.
2. Some of the commands (e.g. `TableColumnResizeEditing`) tend to trigger `update ui` event while calling `_disableCommands` in `change:isSourceEditingMode` listener placed inside `SourceEditing` plugin.
3. These plugins tend to modify content or classes of editable, causing `update ui` to be fired, which lead to pin every tooltip again, even if pin target is hidden. 
4. I added pin target visibility watcher in balloon panel view and it looks ok.

### DIFF
--- a/packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
+++ b/packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
@@ -399,13 +399,22 @@ export default class BalloonPanelView extends View {
 		// Hide the panel if the target element is no longer visible.
 		if ( targetElement && !this._resizeObserver ) {
 			const checkVisibility = () => {
-				if ( !targetElement.checkVisibility( { checkVisibilityCSS: false } ) ) {
-					this.isVisible = false;
+				// If the target element is no longer visible, hide the panel.
+				if ( !targetElement.checkVisibility() ) {
+					this.hide();
 				}
 			};
 
+			// Element is being resize to 0x0 after being hidden, so we need to
+			// check size in order to determine if it's visible or not.
 			this._resizeObserver = new ResizeObserver( targetElement, checkVisibility );
-			checkVisibility();
+
+			// Check the visibility of the target element immediately, as the target element might be
+			// already hidden when the panel is being pinned. It's wrapped in requestAnimationFrame because the
+			// `_startPinning` method is called during `pin()` and the panel is not yet visible, and not every listener
+			// has been attached yet. This way, we wait until rest of the listeners are attached in caller methods
+			// and then check the visibility.
+			window.requestAnimationFrame( checkVisibility );
 		}
 	}
 
--- a/packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
+++ b/packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
@@ -260,7 +260,7 @@ export default class BalloonPanelView extends View {
 	 *
 	 * @param options Positioning options compatible with {@link module:utils/dom/position~getOptimalPosition}.
 	 * Default `positions` array is {@link module:ui/panel/balloon/balloonpanelview~BalloonPanelView.defaultPositions}.
-	 * @returns {Boolean} Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
+	 * @returns Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
 	 * provided in the options is invisible (e.g. element detached from DOM).
 	 */
 	public attachTo( options: Partial<PositionOptions> ): boolean {
@@ -388,7 +388,7 @@ export default class BalloonPanelView extends View {
 	 * Starts managing the pinned state of the panel. See {@link #pin}.
 	 *
 	 * @param options Positioning options compatible with {@link module:utils/dom/position~getOptimalPosition}.
-	 * @returns {Boolean} Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
+	 * @returns Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
 	 * provided in the options is invisible (e.g. element detached from DOM).
 	 */
 	private _startPinning( options: Partial<PositionOptions> ): boolean {
--- a/packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
+++ b/packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
@@ -405,8 +405,8 @@ export default class BalloonPanelView extends View {
 				}
 			};
 
-			// Element is being resize to 0x0 after being hidden, so we need to
-			// check size in order to determine if it's visible or not.
+			// Element is being resized to 0x0 after it's parent became hidden,
+			// so we need to check size in order to determine if it's visible or not.
 			this._resizeObserver = new ResizeObserver( targetElement, checkVisibility );
 
 			// Check the visibility of the target element immediately, as the target element might be
--- a/packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
+++ b/packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
@@ -260,8 +260,16 @@ export default class BalloonPanelView extends View {
 	 *
 	 * @param options Positioning options compatible with {@link module:utils/dom/position~getOptimalPosition}.
 	 * Default `positions` array is {@link module:ui/panel/balloon/balloonpanelview~BalloonPanelView.defaultPositions}.
+	 * @returns {Boolean} Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
+	 * provided in the options is invisible (e.g. element detached from DOM).
 	 */
-	public attachTo( options: Partial<PositionOptions> ): void {
+	public attachTo( options: Partial<PositionOptions> ): boolean {
+		const target = getDomElement( options.target );
+
+		if ( target && !isVisible( target ) ) {
+			return false;
+		}
+
 		this.show();
 
 		const defaultPositions = BalloonPanelView.defaultPositions;
@@ -299,6 +307,8 @@ export default class BalloonPanelView extends View {
 		this.left = left;
 		this.position = position;
 		this.withArrow = withArrow;
+
+		return true;
 	}
 
 	/**
@@ -338,6 +348,10 @@ export default class BalloonPanelView extends View {
 	public pin( options: Partial<PositionOptions> ): void {
 		this.unpin();
 
+		if ( !this._startPinning( options ) ) {
+			return;
+		}
+
 		this._pinWhenIsVisibleCallback = () => {
 			if ( this.isVisible ) {
 				this._startPinning( options );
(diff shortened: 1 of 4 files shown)


Mati365:
Asked for QA testing.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -387,6 +395,27 @@ export default class BalloonPanelView extends View {
 		this.listenTo( global.window, 'resize', () => {
 			this.attachTo( options );
 		} );
+
+		// Hide the panel if the target element is no longer visible.
+		if ( targetElement && !this._resizeObserver ) {
+			const checkVisibility = () => {
+				// If the target element is no longer visible, hide the panel.
+				if ( !targetElement.checkVisibility() ) {
+					this.hide();
```
niegowski, line 404: I don't know if this is the right method. Doesn't the `unpin()` fit better? I don't know how those methods are used from outside this class.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -395,6 +424,11 @@ export default class BalloonPanelView extends View {
 	private _stopPinning(): void {
 		this.stopListening( global.document, 'scroll' );
 		this.stopListening( global.window, 'resize' );
+
+		if ( this._resizeObserver ) {
+			this._resizeObserver.destroy();
+			this._resizeObserver = null;
+		}
```
niegowski, line 431: Maybe we should also destroy the `ResiseObserver` instance on `destroy()`.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -387,6 +395,27 @@ export default class BalloonPanelView extends View {
 		this.listenTo( global.window, 'resize', () => {
 			this.attachTo( options );
 		} );
+
+		// Hide the panel if the target element is no longer visible.
+		if ( targetElement && !this._resizeObserver ) {
+			const checkVisibility = () => {
+				// If the target element is no longer visible, hide the panel.
+				if ( !targetElement.checkVisibility() ) {
```
niegowski, line 403: AFAIK we have a helper for `isVisible()`. I'm not sure how this is different from this DOM method. Could you please verify it?

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
Mati365: I'll check that.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
Mati365: Good point, I'll add it to destroy method.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
Mati365: I changed it, basically it works exactly the same as `hide` but it seems to more convenient to use `unpin` method in `_startPinning`.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
Mati365: I checked it a little bit further, and it looks like `destroy` method does not unpin the tooltip properly. I adjusted destroy method and added tests.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
Mati365: It works exactly like this one above, I replaced it with your suggestion to make it more convenient.

Mati365:
@niegowski I applied CR suggestions.

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -387,6 +404,27 @@ export default class BalloonPanelView extends View {
...
+			// Element is being resized to 0x0 after it's parent became hidden,
+			// so we need to check size in order to determine if it's visible or not.
+			this._resizeObserver = new ResizeObserver( targetElement, checkVisibility );
+
+			// Check the visibility of the target element immediately, as the target element might be
+			// already hidden when the panel is being pinned. It's wrapped in `requestAnimationFrame` because the
+			// `_startPinning` method is called during `pin()` and the panel is not yet visible, and not every listener
+			// has been attached yet. This way, we wait until rest of the listeners are attached in caller methods
+			// and then check the visibility.
+			window.requestAnimationFrame( checkVisibility );
```
niegowski, line 426: I'm not sure about this approach. Can we listen to some event or observable property to know when it's ready for this check?

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -194,6 +203,14 @@ export default class BalloonPanelView extends View {
 		} );
 	}
 
+	/**
+	 * @inheritDoc
+	 */
+	public override destroy(): void {
+		this.hide();
```
niegowski, line 210: Why only `hide()`? Why not `unpin()`?

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
Mati365: @niegowski I'm not sure if we can do that without breaking integrations. `.pin()` used to return `void` so I assume that every integration assumes that balloon is pinned after calling this method (for example tooltip integration). The problem is that if we depend on some kind of property, or even sync event, there might be a chance that `isVisible` is switched to `false` before exiting `.pin()` method:

```ts
...
balloon.pin({
});

ballon.isVisible // is false
...
```

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
Mati365: @niegowski These two methods do the same, most of the time. I think that `hide` is a little bit safer. `unpin` does something like this:

```tsx
	public unpin(): void {
		if ( this._pinWhenIsVisibleCallback ) {
            ...
``` 

So `hide` might (in theory) keep `isVisible = true` when `_pinWhenIsVisibleCallback`  is `null`. Is this real case? Don't know.

charlttsie:
We have finished the tests with @kubaklatt and @marcelmroz and the feedback is positive - the initial issue is fixed and we haven't found any regressions 👍

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -387,6 +404,27 @@ export default class BalloonPanelView extends View {
...
+			const checkVisibility = () => {
+				// If the target element is no longer visible, hide the panel.
+				if ( !isVisible( targetElement ) ) {
+					this.unpin();
+				}
+			};
+
+			// Element is being resized to 0x0 after it's parent became hidden,
+			// so we need to check size in order to determine if it's visible or not.
+			this._resizeObserver = new ResizeObserver( targetElement, checkVisibility );
```
oleq, line 419: This code will likely be obsolete after this fix. Consider removing it https://github.com/ckeditor/ckeditor5/blob/ea0947ab56e3b6923b989cd80606688524fff474/packages/ckeditor5-ui/src/tooltipmanager.ts#L431-L437.

Mati365: @oleq I tried, but it's not that simple and makes tooltips buggy. It turns out that we cannot bind to `isVisible` of balloon panel view because it becomes hidden during updating of its position (and then it became is visible again).

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
oleq: I'm not a huge fan of this RAF either but I understand your concerns and saw errors caused by the synchronous `checkVisibility()` call here.

I iterated on this solution and came up with another one that makes `BPV#attachTo()` bulletproof to invisible `HTMLElements` and does not require the initial `checkVisible()` because this branch of code is never reached (if `attachTo()` failed). The `_pinWhenIsVisibleCallback` upon `change:isVisible` also is not activated. On top of it, I successfully removed a resize observer from the `TooltipManager`, too. 

Tests passed and I couldn't break the balloon system by testing manually. There are few tests missing for full CC, though.

You can see my approach here https://github.com/ckeditor/ckeditor5/compare/ck/16739...ck/16739-v2?expand=1.

Let me know what you think about it.

---

On a side note, if not for backward compatibility and unavoidable breaking changes, I'd warn if the pinning/attaching target is an invisible element because this whole logic of silently allowing divisible targets is a hack. You cannot position an element relative to another one that misses a position itself.

oleq (dismissed):

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -243,8 +260,16 @@ export default class BalloonPanelView extends View {
 	 *
 	 * @param options Positioning options compatible with {@link module:utils/dom/position~getOptimalPosition}.
 	 * Default `positions` array is {@link module:ui/panel/balloon/balloonpanelview~BalloonPanelView.defaultPositions}.
+	 * @returns {Boolean} Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
```
niegowski, line 263:
```suggestion
	 * @returns Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
```

packages/ckeditor5-ui/src/panel/balloon/balloonpanelview.ts
```diff
@@ -359,9 +388,13 @@ export default class BalloonPanelView extends View {
 	 * Starts managing the pinned state of the panel. See {@link #pin}.
 	 *
 	 * @param options Positioning options compatible with {@link module:utils/dom/position~getOptimalPosition}.
+	 * @returns {Boolean} Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
```
niegowski, line 391:
```suggestion
	 * @returns Whether the balloon was shown and successfully attached or not. Attaching can fail if the target
```

niegowski (approved):

charlttsie:
We have done a retest of the PR together with @Acrophost, @lkszzajac and @kubaklatt. It looks good to us and, as previously, haven't found any regressions 🎉

Issue #13693, opened 2023-03-16
Composition breaks when typing a Korean phonetic radical after typing `space` key on Android Chrome


byeokim:
## 📝 Provide detailed reproduction steps (if any)

1. Launch any CKEditor, e.g. [this demo](https://ckeditor.com/ckeditor-5/demo/mobile-support/), on Android Chrome
2. Put caret at the start of any empty line
3. Use GBoard keyboard
4. Add keyboard language: select Korean then 2-Bulsik
5. Change keyboard language to Korean
6. Type `ㄱ`
7. Type `ㅏ`
8. Type `space`
9. Type `ㄱ`
10. Type `ㅏ`

[screenrecording.webm](https://user-images.githubusercontent.com/29354589/225513821-4b0b753b-52c0-4274-be4b-133ffa9e2d03.webm)

### ✔️ Expected result

`가 가`

### ❌ Actual result

`가 ㄱㅏ`

### ❓ Possible solution

Not updating Text node when composing, i.e. after `compositionstart` event and before `compositionend` event, by adding the following code into `renderer.ts`:

https://github.com/ckeditor/ckeditor5/blob/1bbb4dfb66395f9de41d9d6353bd8e69293621e0/packages/ckeditor5-engine/src/view/renderer.ts#L714-L728

```diff
else if ( action === 'replace' ) {
+ if (!this.isComposing) {
    updateTextNode( actualDomChildren[ i ] as DomText, ( expectedDomChildren[ i ] as DomText ).data );
+ }
  i++;
```

On Android Chrome, when a character is typed and the typing triggers a `compositionstart` event, IME seems to put the character into the Text node sometime after `compositionupdate` event. Meanwhile CKEditor also effectively puts the character into the node by invoking `updateTextNode`. So there would be two same characters. `MutationObserver` detects the DOM mutation and invokes another `updateTextNode`. Because of that invocation one of the two characters is removed. This can be confirmed by uncommenting `CK_DEBUG_TYPING` `console.log`s. I suspect that which character is removed affects whether the current composition should break, i.e. IME seems to be keeping track of which character instance is involved in the current composition.

The suggested solution will prevent the character duplication. IME will put the character while CKEditor won't. Thus there would be no removal of the duplicated character. Neither be composition-breaking.

## 📃 Other details

* Browser: Chrome 91.0
* OS: Android 12
* First affected CKEditor version: 36.0.1
* Installed CKEditor plugins: Essentials, Paragraph

---

[...]

Issue #11569, opened 2022-04-05
Issue typing Korean after two space characters


slotterbackW:
## 📝 Provide detailed reproduction steps (if any)

When typing Korean in CKEditor, after typing two spaces vowels stop being combined.

1. Set keyboard language to 2-set Korean
2. Press "d", "k", "t" keys on keyboard
3. Notice vowels are combined into a single character
4. Press space twice
5. Type "d", "k", "t" again.
6. Notice vowels are not combined

![korean_ime_bug](https://user-images.githubusercontent.com/14916556/161845850-05d215b0-20c0-4dc3-bf54-43bed5fcf219.gif)

### ✔️ Expected result

The vowels should be combined again, and produce the same character which was produced when typing them before the spaces.

### ❌ Actual result

The vowels are not combined.

## 📃 Other details

* Browser: Chome
* OS: MacOS 11.6.2
* First affected CKEditor version: …
* Installed CKEditor plugins: Reproducible on classic editor and possibly others?

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

FilipTokarski:
Hi, thanks for the report. We are aware of issues with IME and we plan to address them in [migration to `beforeInput`](https://github.com/ckeditor/ckeditor5/issues/7462), which is currently very high on our priority list.

Acrophost:
The issue is no longer reproducible on branch `ck/11438-beforeinput-ime-research-vol1.1-android`

joniel:
Still problems...
even in ckeditor hompage too... T_T;

Witoso:
Related issue reported for Android Chrome: #13693.

niegowski:
The work for this issue and others related is completed in the https://github.com/ckeditor/ckeditor5/pull/16289, and we finalized the round of internal (successful) tests. I encourage everyone to test the PR if you have a chance. It should be merged in the following days, and will be a part of the next release.

PR #16755, opened 2024-07-18
Fix: composition breaks typing a Korean after space key on Android Chrome

Johnyworld:
This issue still exists after version `42.0.1`.

- https://github.com/ckeditor/ckeditor5/issues/11569
- https://github.com/ckeditor/ckeditor5/issues/13693


Witoso:
Those issues target milestone 76 which is not released yet. Released patches went through to the hotfix route.

Issue #6399, opened 2020-03-10
Introduce nested dropdowns


Mgsy:
## 📝 Provide a description of the new feature

Introduce some guide/API for creating nested dropdowns.

![image](https://user-images.githubusercontent.com/20988892/76304051-fb72be80-62c2-11ea-86a6-5ba2f52139e3.png)

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

regischen:
I am wondering can developer can implement this by themselves?

gumatias:
@Mgsy mind being more specific on what kind of guidance or API we are looking for? I might take a stab at it if I can.

rutvay-dv:
Hi @Mgsy is this introduced ?

Witoso:
We are investigating this in #16766.

PR #16766, opened 2024-07-19
Introduce nested menu component for dropdowns

scofalik:

Feature (ui): Introduced nested menu component for dropdowns. Closes: #6399

---

### Additional information


packages/ckeditor5-ui/src/editorui/editorui.ts
```diff
@@ -669,6 +671,22 @@ export default abstract class EditorUI extends /* #__PURE__ */ ObservableMixin()
...
+	private _bindBodyCollectionWithFocusTracker() {
+		const body = this.view.body;
+
+		for ( const view of body ) {
+			this.focusTracker.add( view.element! );
+		}
+
+		body.on<CollectionAddEvent<View>>( 'add', ( evt, view ) => {
+			this.focusTracker.add( view.element! );
+		}, { priority: 'low' } );
```
scofalik, line 684: Why `low` priority is needed?

packages/ckeditor5-ui/src/dropdown/utils.ts
```diff
@@ -129,11 +132,85 @@ export function createDropdown(
...
+ * After using this helper, the `dropdown` will fire {@link module:ui/dropdown/dropdownview~DropdownViewEvent `execute`} event when
+ * a nested menu button is pressed.
+ *
+ * The helper will make sure that the `dropdownMenuRootListView` is lazy loaded, i.e., the menu component structure will be initialized
+ * and rendered only after the `dropdown` is opened for the first time.
+ *
+ * @param dropdownView A dropdown instance to which the menu component will be added.
+ * @param dropdownMenuRootListView A menu component to be added to the dropdown.
+ */
+export function addMenuToDropdown( dropdownView: DropdownView, dropdownMenuRootListView: DropdownMenuRootListView ): void {
```
scofalik, line 187: Pass `definitions` and `body` instead of `dropdownMenuRootListView` - let's hide the complexity from the integrator.

scofalik: Let's set `dropdownView.menuView = dropdownMenuRootListview` to be consistent with other helpers.

packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/dropdown/menu/dropdownmenu.css
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2003-2024, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+.ck.ck-dropdown-menu__menu {
+	/* Enable font size inheritance, which allows fluid UI scaling. */
+	font-size: inherit;
```
oleq, line 8: Why do we need that? The scaling looks fine without it, for instance, if changed `--ck-font-size-base` to `18px`. 
image

oleq (changes requested):
I was thinking about the best approach to CSS class naming because the current one is not bad but not very precise either, and I came up with something like this. Let me know what you think about this.


JS component | CSS class in PR | Proposed CSS class
-- | -- | --
DropdownMenuListView | .ck-dropdown-menu | .ck-dropdown-menu-list (relates to .ck-list)
DropdownNestedMenuView | .ck-dropdown-menu__menu | .ck-dropdown-menu-list__nested-menu
DropdownMenuListItemView | .ck-dropdown-menu__menu__item | .ck-dropdown-menu-list__nested-menu__item
DropdownMenuPanelView | .ck-dropdown-menu__menu__panel (if it follows the ck-dropdown-menu__menu__ pattern, shouldn’t the JS component be DropdownMenuNestedMenuPanelView anyway?) | .ck-dropdown-menu__nested-menu__panel

---

Also: I think there's something wrong with the classes set on the panel, I inspected the following `class="..."` string: `ck ck-balloon-panel ck-balloon-panel_es ck-balloon-panel_visible ck-balloon-panel_with-arrow ck-ai-commands-dropdown ck-ai-assistant-ui ck-ai-assistant-ui_theme ck ck-reset ck-dropdown-menu__menu__panel ck-ai-commands-dropdown ck-ai-assistant-ui ck-ai-assistant-ui_theme ck-dropdown-menu__menu__panel_position_es`. Please note that `ck` is rendered twice. This is very odd and should be investigated.

packages/ckeditor5-ui/src/editorui/editorui.ts
scofalik: @oleq I actually don't know why I added this priority here. I tested both with AI Assistant and in a dropdowns manual test and it works fine on normal priority. Unit tests pass too. Speaking of, I missed a unit test for this functionality.

One thing that I thought about is that items are rendered on normal priority, so maybe I wanted to be sure they are rendered already. But as I said, this happens on normal priority in `ViewCollection`. Since `BodyCollection` extends `ViewCollection` and rendering callback is added in constructor, any other callback on normal priority added for `BodyCollection#event:add` will fire after rendering.

packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/dropdown/menu/dropdownmenu.css
scofalik: I have no idea. I am fine with removing it.

oleq (approved):

PR #16770, opened 2024-07-19
PoC of a single block editor.

### LINKED ISSUE #762: Allow enabling all types of editors on paragraph-like root
Developers should be able to initialize the editor on a paragraph-like roots. For example, someone may want to make content of just an `<h1>` element editable:

```html
<h1 id=editor>Title...</h1>
<script>
    InlineEditor.create( document.getElementById( 'editor' ) )
        .then( ... )
        .catch( ... );
</script>
```

The inline type of editors are most often to be used in such a cases and there's even a ticket for `InlineEditor` already (https://github.com/ckeditor/ckeditor5-editor-inline/issues/3 and https://github.com/ckeditor/ckeditor5-editor-inline/issues/5). But this requirement applies also to `BalloonEditor` and I may imagine that someone might want to enable `ClassicEditor` in such a mode. So, since most of the work is needed in the common part of the code base (core, engine), I'd say that the solution could be generic to all kinds of editors from day one.

In case of editors which upcast existing DOM elements (like inline and balloon editors), the information about the type of element (whether it's div-like or paragraph-like) can be taken from that element itself. But in case of the classic editor it will need to be specified in the config. The config option could work for all editors, overriding the upcasted element.

What's kinda tricky here is mapping DOM element name to schema item name so we can create a model root with the right name. However, I think that this should be enough, taken that the automatic behaviour will be overridable by configuration anyway:

```js
const paragraphLikeElements = [ 'p', 'h1', ... ];

getRootName( domElementName ) {
    if ( paragraphLikeElements.includes( domElementName ) { 
        return '$block';
    } else {
        return '$root';
    }
}
```

Defaulting to the generic `$block` item for all p/h1/etc elements should be enough.

---

Add :+1: to this post if you'd like this feature to be implemented.

niegowski:

Type: TODO. Closes #762.

---

### Additional information

### DIFF
--- a/tests/manual/single-block.html
+++ b/tests/manual/single-block.html
@@ -1,0 +1,3 @@
+<div id="editor">
+	This is an <b>editor</b>.
+</div>
--- a/tests/manual/single-block.js
+++ b/tests/manual/single-block.js
@@ -1,0 +1,197 @@
+/**
+ * @license Copyright (c) 2003-2024, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/* globals console, window, document */
+
+import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor.js';
+import AutoImage from '@ckeditor/ckeditor5-image/src/autoimage.js';
+import AutoLink from '@ckeditor/ckeditor5-link/src/autolink.js';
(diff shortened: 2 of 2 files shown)


Witoso:
This is only PoC for enabling such thing with a single plugin, but it won't solve the whole case listed in #762 (what about other root types like headings, etc.). We have shortlisted some time this year to take another look at the topic, but we don't have a specific ETA.

If you like, you can turn the proposed PoC to an external package that provides such plugin, and use it.

Issue #15744, opened 2024-01-24
Widget toolbar covers editor's sticky toolbar when scrolling


FilipTokarski:
## 📝 Provide detailed reproduction steps (if any)

1. [Go to demo](https://ckeditor.com/ckeditor-5/demo/feature-rich/)
2. Add a table
3. Focus table and scroll the content down

### ✔️ Expected result

Table toolbar detects editor's sticky toolbar and moves down to the bottom of the table.

### ❌ Actual result

Table toolbar does not detect sticky toolbar and covers it. It only moves to the bottom of the table when it reaches top border of the editor (or rather the top website navigation bar). This happens also for other widget toolbars, for example image.

https://github.com/ckeditor/ckeditor5/assets/34380544/9ecd627d-2dd3-451c-84b5-1ccc03a3ca49

### ❓ Possible solution

_If you have ideas, you can list them here. Otherwise, you can delete this section._

## 📃 Other details

* Browser: …
* OS: …
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

oleq:
We already have two integrations for `Dialog` positioning and `scrollToTheSelection()`

https://github.com/ckeditor/ckeditor5/blob/dd3b278ece67f714b7f440801b4b770fa2ac804a/packages/ckeditor5-editor-classic/src/classiceditorui.ts#L184-L251

Maybe there’s a similar way to hook into the `ContextualBalloon` system?

PR #16780, opened 2024-07-23
Widget toolbar no longer covers editor's sticky toolbar when scrolling

Mati365:

Fix (editor-classic): Widget toolbar no longer covers editor's sticky toolbar when scrolling. Closes https://github.com/ckeditor/ckeditor5/issues/15744.

---

### Before

https://github.com/user-attachments/assets/f805f96a-0cbf-4a7c-bec0-54ad2ff3c9b6

### After

https://github.com/user-attachments/assets/96017d0b-36ee-4da4-99b6-d961a4138307

### DIFF
--- a/packages/ckeditor5-editor-classic/src/classiceditorui.ts
+++ b/packages/ckeditor5-editor-classic/src/classiceditorui.ts
@@ -7,8 +7,6 @@
  * @module editor-classic/classiceditorui
  */
 
-import { isFunction } from 'lodash-es';
-
 import type { Editor, ElementApi } from 'ckeditor5/src/core.js';
 import {
 	EditorUI,
@@ -220,8 +218,8 @@ export default class ClassicEditorUI extends EditorUI {
 			// It's an issue because the contextual balloon can overlap top table cells when the table is larger than the viewport
 			// and it's placed at the top of the editor. It's better to overlap toolbar in that situation.
 			// Check this issue: https://github.com/ckeditor/ckeditor5/issues/15744
-			const target = isFunction( position.target ) ? position.target() : position.target;
-			const limiter = isFunction( position.limiter ) ? position.limiter() : position.limiter;
+			const target = typeof position.target === 'function' ? position.target() : position.target;
+			const limiter = typeof position.limiter === 'function' ? position.limiter() : position.limiter;
 
 			if ( target && limiter && new Rect( target ).height >= new Rect( limiter ).height - stickyPanelHeight ) {
 				return;
@@ -229,7 +227,7 @@ export default class ClassicEditorUI extends EditorUI {
 
 			// Ensure that viewport offset is present, it can be undefined according to the typing.
 			const viewportOffsetConfig = { ...position.viewportOffsetConfig };
-			const newTopViewportOffset = Math.max( stickyPanelHeight, viewportOffsetConfig.top || 0 );
+			const newTopViewportOffset = ( viewportOffsetConfig.top || 0 ) + stickyPanelHeight;
 
 			evt.return = {
 				...position,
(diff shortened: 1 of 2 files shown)


niegowski (changes requested):
This PR does not handle the floating widget toolbar. This is a position for the toolbar when the widget (for example a table) is higher than the whole editor viewport.

https://github.com/user-attachments/assets/bf6ea2db-32e0-4ee4-8a4d-fed617c6549e

I'm also not sure if decorating the `getPosition()` method of the `ContextualBalloon` is the best solution.

Mati365:
@niegowski 

> This PR does not handle the floating widget toolbar. This is a position for the toolbar when the widget (for example a table) is higher than the whole editor viewport.

I'll take a look.

> I'm also not sure if decorating the getPosition() method of the ContextualBalloon is the best solution.

Yep, It's not ideal but at this moment. We discussed it recently with @oleq [here](https://cksource.slack.com/archives/C07DN78D3GA/p1721737939154919), and it's good enough at this moment.

Mati365:
@niegowski I checked this issue a little bit more, and it looks like we have to disable this fix entirely for targets that are larger than limiters. The problem is that there is no enough space to show balloon without overlapping cells in the table. 

Example:
obraz

Current behavior (without fix): 
obraz

packages/ckeditor5-ui/src/panel/balloon/contextualballoon.ts
```diff
@@ -320,10 +323,35 @@ export default class ContextualBalloon extends Plugin {
...
-		this.view.pin( this._getBalloonPosition()! );
+		this.view.pin( this.getPosition()! );
 		this._fakePanelsView!.updatePosition();
 	}
 
+	/**
+	 * Returns position options of the last view in the stack.
+	 * This keeps the balloon in the same position when the view is changed.
+	 */
+	public getPosition(): Partial<PositionOptions> | undefined {
```
niegowski, line 334: Maybe we could rename this method as it is now public to something that would not confuse users. Maybe `getPositionOptions()` would better reflect what it does.

packages/ckeditor5-editor-classic/src/classiceditorui.ts
```diff
@@ -187,6 +192,61 @@ export default class ClassicEditorUI extends EditorUI {
...
+			const target = isFunction( position.target ) ? position.target() : position.target;
+			const limiter = isFunction( position.limiter ) ? position.limiter() : position.limiter;
+
+			if ( target && limiter && new Rect( target ).height >= new Rect( limiter ).height - stickyPanelHeight ) {
+				return;
+			}
+
+			// Ensure that viewport offset is present, it can be undefined according to the typing.
+			const viewportOffsetConfig = { ...position.viewportOffsetConfig };
+			const newTopViewportOffset = Math.max( stickyPanelHeight, viewportOffsetConfig.top || 0 );
```
niegowski, line 232: I believe this should be a sum of those 2 offsets. Otherwise, we would have such an effect on our docs:

Screenshot 2024-08-02 at 15 08 57

The sticky pannel is offset from the window top by the size of viewport offset so contextual balloon should be offset by the sum of those.

niegowski (changes requested):

packages/ckeditor5-editor-classic/src/classiceditorui.ts
Mati365: Yep, you are right. I'll check that.

packages/ckeditor5-editor-classic/src/classiceditorui.ts
```diff
@@ -187,6 +192,61 @@ export default class ClassicEditorUI extends EditorUI {
...
+
+			// Measure toolbar (and menu bar) height.
+			const stickyPanelHeight = new Rect( stickyPanel.element ).height;
+
+			// Handle edge case when the target element is larger than the limiter.
+			// It's an issue because the contextual balloon can overlap top table cells when the table is larger than the viewport
+			// and it's placed at the top of the editor. It's better to overlap toolbar in that situation.
+			// Check this issue: https://github.com/ckeditor/ckeditor5/issues/15744
+			const target = isFunction( position.target ) ? position.target() : position.target;
+			const limiter = isFunction( position.limiter ) ? position.limiter() : position.limiter;
```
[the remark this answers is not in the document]
Mati365: We can, I thought that it was convention because I picked it from other place.

niegowski (approved):

PR #16792, opened 2024-07-24
Allowed for configuring editor label via config.label property

### LINKED ISSUE #9731: Introduce CKEditor title/tooltip used by screen readers like in CKEditor 4

CKEditor 4 has title property which:   
\- you can set,   
\- is being displayed as a tooltip   
\- can be read by screen readers when editor gets focused (useful for visually impaired users)  
 

[https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-title](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-title)

Same could be introduced fro CKEditor 5.

---

### LINKED ISSUE #11863: Make editor's application labels unique

### LINKED ISSUE #15208: Make Editor title configurable again to fix accessibility issues

For accessibility reasons, it would be great if CKEditor's voice-label `<label>` element could be configurable so that it could reflect the name of the field that CKEditor is replacing.

Screen readers read the "voice label" `<label>` value that is generated by CKEditor, which is always "Rich Text Editor."

For example, this instance: 

Screen Shot 2023-10-18 at 3 09 20 PM

has this `<label>`: 
```
<label class="ck ck-label ck-voice-label" id="ck-editor__label_e94ba4e316f3b958963791e09f1ad2e78">Rich Text Editor</label>
```

For users who rely on screen readers, it would be much more helpful if the reader uttered "Comment," and not "Rich Text Editor." (See https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html, https://wcag.com/developers/3-3-2-labels-instructions.) 

This value was made configurable in CKEditor4 with the `config.title` property: https://dev.ckeditor.com/ticket/12204.

After digging through CKEditor5 docs, I have not found a similar feature. I realize that the architecture of CKEditor 5 is vastly different than 4, but my proposal would be that the editor title/voice label be a configurable value -- if at all possible. 

~~Additionally, the `aria-label` on `<div class="ck-content" ...>` should be removed. `aria-label` is only necessary if there is not an associated `<label>` tag. The effect of having both the `<label>` and the `aria-label` is that both are read by the screen reader -- which is confusing.~~

_What is the expected behavior of the proposed feature?_
A developer could add a title config value to the editor's `create()` method, and the generated voice label element would contain that title.

Something like:
```
ClassicEditor
	title: 'My Editor Title',
} )
...
```
would generate:

```
<label class="ck ck-label ck-voice-label" id="[some_id]">My Editor Title</label>
```

~~Additionally, the `aria-label` would be removed from the `ck-content` `<div>`:~~

```
[Edit: removed code example.]
```

**Edit 10/20/23:**
After some further research, I see that the `aria-label` is important for multi-root editors. So, I've removed that part of this request.

---

oleq:

Feature: Allowed to configure the accessible editable area label via `config.label` property. Closes #15208. Closes #11863. Closes #9731.

MINOR BREAKING CHANGE (ui): The default `aria-label` provided by `InlineEditableUIView` is now `'Rich Text Editor. Editing area: [root name]'` (previously `'Editor editing area: [root name]'`). Use the `options.label` constructor property to adjust the label.

---

### Additional information

* I left the comments editor untouched because the recent sidebar a11y PR handles the aria-label for the comments editor.

* [x] Q: What do we do about context config?
   * A: It appears that you can merge context and editor configs. See https://ckeditor.com/docs/ckeditor5/latest/api/module_core_context-Context.html#static-function-create.


packages/ckeditor5-editor-multi-root/src/multirooteditoruiview.ts
```diff
@@ -65,6 +67,7 @@ export default class MultiRootEditorUIView extends EditorUIView {
 		options: {
 			editableElements?: Record<string, HTMLElement>;
 			shouldToolbarGroupWhenFull?: boolean;
+			label?: string | Record<string, string>;
```
f1ames, line 70: Shouldn't it be just `label?: Record<string, string>;`? Or one can pass `string` and then all roots will have same label (and if so, is it correct behavior from accessibility perspective?).

packages/ckeditor5-editor-inline/tests/inlineeditor.js
```diff
@@ -284,6 +286,106 @@ describe( 'InlineEditor', () => {
...
+					label: 'Custom label'
+				} );
+
+				expect( editor.editing.view.getDomRoot().getAttribute( 'aria-label' ), 'Override value' ).to.equal(
+					'Custom label'
+				);
+
+				await editor.destroy();
+
+				expect( editorElement.getAttribute( 'aria-label' ), 'Restore value' ).to.equal( 'Pre-existing value' );
```
f1ames, line 357: More general question here, but regarding this behavior (restoring initial `aria-label` attribute value on editor destroy) - how this behaves if there is no `aria-label` set initially? Is it back to empty after editor destroy or `aria-label` is kept? (not sure how it should behave TBH, just food for though).

f1ames (changes requested):
Looking good 👍 Only some minor questions/suggestions from me.

> [x] Q: What do we do about context config?
> 
> * A: It appears that you can merge context and editor configs. See https://ckeditor.com/docs/ckeditor5/latest/api/module_core_context-Context.html#static-function-create.

Can you elaborate? From what I understand, you suggest that label should be set with editor config and not context config, right? Should we document that somehow that setting `label` on context is not recommended because it may result in multiple editor instances with the same label? 🤔

packages/ckeditor5-editor-balloon/src/ballooneditor.ts
```diff
@@ -77,7 +77,7 @@ export default class BalloonEditor extends /* #__PURE__ */ ElementApiMixin( Edit
 
 		this.model.document.createRoot();
 
-		const view = new BalloonEditorUIView( this.locale, this.editing.view, this.sourceElement );
+		const view = new BalloonEditorUIView( this.locale, this.editing.view, this.sourceElement, this.config.get( 'label' ) );
```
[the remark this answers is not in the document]
oleq: This is a good idea and I noticed this too, but that would be a breaking change in one of the key editor classes and I wanted to avoid that.

packages/ckeditor5-editor-multi-root/src/multirooteditoruiview.ts
oleq: Yes, you can pass string here and all roots will get the same name. Is it accessible? It's hard to tell. `aria-label` does not have to be unique and can be repeated on a website. 

Multiple roots can be used creatively, for instance, only one visible/reachable at a time so this approach could be valid. This is also useful if many new roots are created on the fly (e.g. editing a book, a chapter per root, but only one root is visible at a time). I even considered making it `label?: string | Record<string, string> | ( ( name: string ) => string )` for this purpose ("Rich text editor. Editing area: Chapter 11, foo bar baz") but I figured this would be a scope creep and an easy improvement in the future.

Anyway, I mirrored the [placeholder configuration](https://ckeditor.com/docs/ckeditor5/latest/api/module_core_editor_editorconfig-EditorConfig.html#member-placeholder) (also `string | Record<string, string>`) and implementation. 

https://github.com/ckeditor/ckeditor5/blob/54efa9938f46463d11f450b009b63140fe9bf946/packages/ckeditor5-editor-multi-root/src/multirooteditorui.ts#L203-L205

If you think we should escalate it and discuss it with more team members, I'm OK with that 🙂

packages/ckeditor5-editor-inline/tests/inlineeditor.js
oleq: It should work out just fine. See how the attributes are restored upon `(Editing)View#detachDomRoot()`:

https://github.com/ckeditor/ckeditor5/blob/54efa9938f46463d11f450b009b63140fe9bf946/packages/ckeditor5-engine/src/view/view.ts#L336-L343

First, we nuke all existing ones and then restore the cached values [from when the root was attached](https://github.com/ckeditor/ckeditor5/blob/54efa9938f46463d11f450b009b63140fe9bf946/packages/ckeditor5-engine/src/view/view.ts#L263-L291). This way we guarantee nothing from the editor's lifecycle will ever be able to stay (leak).

oleq:
> Can you elaborate? From what I understand, you suggest that label should be set with editor config and not context config, right? Should we document that somehow that setting label on context is not recommended because it may result in multiple editor instances with the same label? 🤔

I'm going to use an example to explain. What I meant is that you can make the following changes to the manual test

```diff
diff --git a/packages/ckeditor5-comments/tests/manual/annotations-context.js b/packages/ckeditor5-comments/tests/manual/annotations-context.js
index f8f3dbcd5d..a9fd00d803 100644
--- a/packages/ckeditor5-comments/tests/manual/annotations-context.js
+++ b/packages/ckeditor5-comments/tests/manual/annotations-context.js
@@ -100,7 +100,8 @@ import InlineAnnotations from '../../src/annotations/inlineannotations.js';
                            editorConfig: {
                                     extraPlugins: [ Bold, Italic, Underline, List, Autoformat ]
                            }
-                  }
+                  },
+                  label: 'Context-wide label'
          } );
 
          const editor1 = await createEditor( document.querySelector( '.editors' ), context, channelId, 'editor1' );
@@ -241,7 +242,8 @@ async function createEditor( container, context, channelId, name ) {
                                     editorConfig: {
                                              extraPlugins: [ Bold, Italic, Underline, List ]
                                     }
-                           }
+                           },
+                           label: `Editor-level label: ${ name }`
                   } );
 
          const toggleEditorButton = document.createElement( 'button' );
```

and because of  "label: `Editor-level label: ${ name }`" passed on the editor-level config, each editor will have a unique label. But if you remove this line, both will have "Context-wide label". So, as it turns out, there's no risk the context will force a specific label on all editors because this is still configurable.

packages/ckeditor5-ui/lang/contexts.json
```diff
@@ -1,6 +1,5 @@
 {
 	"Rich Text Editor": "Title of the CKEditor5 editor.",
-	"Editor editing area: %0": "Accessible label of the specific editing area belonging to a container with an ARIA application role.",
```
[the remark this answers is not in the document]
f1ames: Ok, thanks for explanation 👍

packages/ckeditor5-editor-balloon/src/ballooneditor.ts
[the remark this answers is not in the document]
f1ames: Ok, yes, makes sense to keep backwards compatibility in this case 👍

packages/ckeditor5-editor-multi-root/src/multirooteditoruiview.ts
f1ames: Thanks for detailed explanation 👍 

> `aria-label` does not have to be unique and can be repeated on a website.

I think the above solves it since this was my main concern here I guess. And the use case you mentioned is valid so it may be desired behavior in such cases 👍 

> I even considered making it `label?: string | Record<string, string> | ( ( name: string ) => string ) for this purpose ("Rich text editor. Editing area: Chapter 11, foo bar baz")`

Hah, I had similar idea but hold myself off from bringing it up too 👍 😄

packages/ckeditor5-editor-inline/tests/inlineeditor.js
f1ames: Ok, that's good 👍 Thanks for the code references and explanation 👍

oleq:
@lnmp4000 Check out https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_core_editor_editorconfig-EditorConfig.html#member-label and https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_editor-multi-root_multirooteditor-MultiRootEditor.html#function-createEditable.

Witoso:
This is scheduled for the next release, not published yet.

lnmp4000:
> @lnmp4000 Check out https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_core_editor_editorconfig-EditorConfig.html#member-label and https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_editor-multi-root_multirooteditor-MultiRootEditor.html#function-createEditable.

@oleq  Thanks for the link to that code. I probably have a fairly unique situation where I do not know what editable areas I will be attaching when I setup the mutliroot editor. So configuring labels within the `create` call is not possible.
Once the editor is configured, I later call `addRoot` to setup the editable areas. At this point I know the label, so is there a way of supplying config at this point?

e.g. at this point 
```
            editor.addRoot(id, {
                data: getIn(getFieldMeta.current().value, id),
                isUndoable: false
            });
```

oleq:
[`createEditable()` accepts an accessible label as an argument](https://ckeditor.com/docs/ckeditor5/latest/api/module_editor-multi-root_multirooteditor-MultiRootEditor.html#function-createEditable). Please see the documentation of the [`addRoot()` method](https://ckeditor.com/docs/ckeditor5/latest/api/module_editor-multi-root_multirooteditor-MultiRootEditor.html#function-addRoot) to learn when to use it.

lnmp4000:
> [`createEditable()` accepts an accessible label as an argument](https://ckeditor.com/docs/ckeditor5/latest/api/module_editor-multi-root_multirooteditor-MultiRootEditor.html#function-createEditable). Please see the documentation of the [`addRoot()` method](https://ckeditor.com/docs/ckeditor5/latest/api/module_editor-multi-root_multirooteditor-MultiRootEditor.html#function-addRoot) to learn when to use it.

@oleq  I'm not sure I have access to `createEditable` directly. I'm using `@ckeditor/ckeditor5-react`, which I think is complicating matters.

So the editor root is setup and the `editor` and `editableElements` (empty at this point) are loaded into a react context
```jsx
 const { editableElements, toolbarElement, editor } = useMultiRootEditor({
   // -- config here -- //
})
```

Later I use the `editor` from the context to add the contenteditable element to the root.

```jsx
  editor.addRoot(id, {
                data: getIn(getFieldMeta.current().value, id),
                isUndoable: false
            });
```
The `editableElements` updates which updates the context, and finally this is rendered inside my application along with the `toolbarElement`.

There is no direct call to `createEditable` I assume this is happening in the internals of the hook.

oleq:
Thank you @lnmp4000 for a detailed explanation. The core editor API is ready for your use-case but the react integration isn't. Could you create a corresponding issue in https://github.com/ckeditor/ckeditor5-react/issues, please? Thank you.

lnmp4000:
https://github.com/ckeditor/ckeditor5-react/issues/529

PR #16795, opened 2024-07-24
Formats the input source code by adding appropriate indentation to specific HTML elements.

huanghuafr:

Type: Message. Closes #000.

---

### Additional information


Reinmar:
We shipped Enhanced Source Code Editing (can be checked in https://ckeditor.com/ckeditor-5/demo/html-support/) which can format HTML much better than the standard feature. Thus, we don't want to improve complexity of the base Source Code Editing feature.

Issue #16819, opened 2024-07-30
`Maximum call stack size exceeded` error when trying to style a very long paragraph


mabryl:
## 📝 Provide detailed reproduction steps (if any)

1. Download the sample file and paste its contents into the editor - 
[long text ckeditor.txt](https://github.com/user-attachments/files/16425134/long.text.ckeditor.txt)

2. Select all of the content
3. Try to style it, e.g. by applying bold

### ✔️ Expected result

The chosen style should get applied and no errors should be thrown.

### ❌ Actual result

The `Maximum call stack size exceeded` error appears and the editor freezes:

https://github.com/user-attachments/assets/b4755547-4300-4321-b9fa-36b0574af21e

## 📃 Other details

* Browser: Chrome Version 126 for macOS, Safari Version 17.5 | **Firefox Version 128 for macOS is not affected by this issue**
* OS: cross-OS
* First affected CKEditor version: N/A
* Installed CKEditor plugins: N/A

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

PR #16821, opened 2024-07-30
Fix crashing differ on large paragraphs

Mati365:

Fix (engine): No longer crash editor when trying to style a very long paragraph. Closes https://github.com/ckeditor/ckeditor5/issues/16819

---

### Additional information

It looks like 200k arguments, passed to `.push` method of the Array, causes Google Chrome crashing.  I switched it to plain for loop and it works fine.

### DIFF
--- a/packages/ckeditor5-engine/src/model/differ.ts
+++ b/packages/ckeditor5-engine/src/model/differ.ts
@@ -1523,6 +1523,7 @@ function _generateDiffInstructionsFromChanges( oldChildrenLength: number, change
 			// Total maximum amount of arguments that can be passed to `Array.prototype.push` may be limited so we need to
 			// add them manually one by one to avoid this limit. However loop might be a bit slower than `push` method on
 			// smaller changesets so we need to decide which method to use based on the size of the change.
+			// See: https://github.com/ckeditor/ckeditor5/issues/16819
 			if ( change.howMany > 500 ) {
 				for ( let i = 0; i < change.howMany; i++ ) {
 					diff.push( 'a' );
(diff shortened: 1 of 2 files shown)


niegowski (commented):
Maybe a helper like our `spliceArray` could be a better solution?

https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-utils/src/splicearray.ts#L12-L33

Mati365:
@pszczesniak / @niegowski I'd recommend sticking with a for loop to maintain consistency with other parts of `_generateDiffInstructionsFromChanges`. I don’t believe there will be any noticeable performance gains from optimizing this specific region, as it’s highly likely that it’s already JIT compiled to machine code

niegowski (changes requested):
Please add an automatic test so we do not have a regression in the future.

Mati365:
@niegowski Added

packages/ckeditor5-engine/src/model/differ.ts
```diff
@@ -1520,7 +1520,17 @@ function _generateDiffInstructionsFromChanges( oldChildrenLength: number, change
 			// We removed `howMany` old nodes, update `oldChildrenHandled`.
 			oldChildrenHandled += change.howMany;
 		} else {
-			diff.push( ...'a'.repeat( change.howMany ).split( '' ) );
+			// Total maximum amount of arguments that can be passed to `Array.prototype.push` may be limited so we need to
+			// add them manually one by one to avoid this limit. However loop might be a bit slower than `push` method on
+			// smaller changesets so we need to decide which method to use based on the size of the change.
+			// See: https://github.com/ckeditor/ckeditor5/issues/16819
+			if ( change.howMany > 1500 ) {
```
pszczesniak, line 1527: Could you write (here and/or in the issue) why we choose 1500?

It's based on some tests or it's some kind of random value?

Mati365: Limitation of args stack depends on OS / Browser etc. I chose some kind of random value.

niegowski (approved):

PR #16859, opened 2024-08-05
Special characters menu item is not correctly marked when enabled

### LINKED ISSUE #16858: Special characters menu item is not correctly marked when enabled

The checkbox is missing.

image

scofalik:

Internal (special-characters): Toolbar and menu bar button should be marked as toggleable. Closes #16858.

---

### Additional information

I added `isToggleable` for both menu bar and toolbar buttons. I see it is done the same way for AI Assistant and Comments archive so I assume this is the pattern we follow.

### DIFF
--- a/packages/ckeditor5-special-characters/src/specialcharacters.ts
+++ b/packages/ckeditor5-special-characters/src/specialcharacters.ts
@@ -247,7 +247,7 @@ export default class SpecialCharacters extends Plugin {
 	}
 
 	/**
-	 * Creates a button for for menu bar that will show special characetrs dialog.
+	 * Creates a button for toolbar anda menu bar that will show special characters dialog.
 	 */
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-special-characters/src/specialcharacters.ts
```diff
@@ -247,7 +247,7 @@ export default class SpecialCharacters extends Plugin {
 	}
 
 	/**
-	 * Creates a button for for menu bar that will show special characetrs dialog.
+	 * Creates a button for toolbar anda menu bar that will show special characters dialog.
```
DawidKossowski, line 250:
```suggestion
	 * Creates a button for toolbar and menu bar that will show special characters dialog.
```

DawidKossowski (approved):

PR #16861, opened 2024-08-05
Added annotations keystrokes information to the Accessibility support guide

### LINKED ISSUE #16860: Add annotations keystrokes information to the Accessibility support guide

New keystrokes were added in the iteration that allow for entering and navigating the sidebar. They have been added to the a11y help dialog but not to the guide.

oleq:

Docs: Added annotations keystrokes information to the Accessibility support guide. Closes #16860.

### DIFF
--- a/docs/features/accessibility.md
+++ b/docs/features/accessibility.md
@@ -163,6 +163,47 @@ These keyboard shortcuts allow for quick access to content editing features.
 	</tbody>
 </table>
 
+#### Keystrokes for interacting with annotation threads (e.g. comments, track changes suggestions)
+
+<table>
+	<thead>
+		<tr>
+			<th>Action</th>
+			<th>PC</th>
+			<th>Mac</th>
+		</tr>
+	</thead>
+	<tbody>
+		<tr>
+			<td>Move focus to the thread when the selection is anchored in its marker</td>
+			<td><kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>E</kbd></td>
+			<td><kbd>⌘⇧E</kbd></td>
+		</tr>
+		<tr>
+			<td>Exit the annotation and move focus back to the edited content</td>
+			<td colspan="2"><kbd>Esc</kbd></td>
+		</tr>
+		<tr>
+			<td>Browse the focused annotation thread or thread comment</td>
+			<td colspan="2"><kbd>Enter</kbd></td>
+		</tr>
+		<tr>
+			<td>Move across internals of the annotation thread</td>
+			<td><kbd>⇥</kbd>, <kbd>Shift</kbd>+<kbd>⇥</kbd></td>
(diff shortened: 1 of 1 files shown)


docs/features/accessibility.md
```diff
@@ -163,6 +163,47 @@ These keyboard shortcuts allow for quick access to content editing features.
...
+			<td>Move across internals of the annotation thread</td>
+			<td><kbd>⇥</kbd>, <kbd>Shift</kbd>+<kbd>⇥</kbd></td>
+			<td><kbd>⇥</kbd>, <kbd>⇧⇥</kbd></td>
+		</tr>
+		<tr>
+			<td>Submit the reply while writing a comment</td>
+			<td><kbd>Ctrl</kbd>+<kbd>Enter</kbd></td>
+			<td><kbd>⌘Enter</kbd></td>
+		</tr>
+		<tr>
```
scofalik, line 200: Move after "Move focus to the thread when the selection is anchored in its marker" or after "Exit the annotation and move focus back to the edited content" so it has more logical order. Also, I wonder if we shouldn't add in the description somethine like "(when annotation is focused)". Or do you think it is obvious?

docs/features/accessibility.md
```diff
@@ -163,6 +163,47 @@ These keyboard shortcuts allow for quick access to content editing features.
...
+			<td>Exit the annotation and move focus back to the edited content</td>
+			<td colspan="2"><kbd>Esc</kbd></td>
+		</tr>
+		<tr>
+			<td>Browse the focused annotation thread or thread comment</td>
+			<td colspan="2"><kbd>Enter</kbd></td>
+		</tr>
+		<tr>
+			<td>Move across internals of the annotation thread</td>
```
scofalik, line 191: I miss also an info that you can go "up" with Esc. Probably after this entry.

scofalik (changes requested):
Some comments.

docs/features/accessibility.md
oleq: This is automatically generated by a script to match the a11y help dialog 1:1. If we start messing with it it may become hard to manage. We discussed ordering in the a11y help dialog in the past, this is probably the best way but it would take some effort.

docs/features/accessibility.md
oleq: See https://github.com/ckeditor/ckeditor5/pull/16861#discussion_r1703928823

scofalik (approved):

docs/features/accessibility.md
oleq: Also: I think the Esc key navigation is universal across the entire web system so it's not worth a dedicated entry here. Enter, on the other hand, is not the most obvious pattern that we used, that's why it's here. I'm not extremely attached to this strategy, though.

Issue #13983, opened 2023-04-26
Remove Format for block elements


filipsobol:
In the https://github.com/ckeditor/ckeditor5/issues/11580 issue, we added the Remove Format feature for inline styles.

We need to implement the same for block styles as well.

See the [scope of the previous ticket](https://github.com/ckeditor/ckeditor5/issues/11580#issuecomment-1495893382) and [implementation PR](https://github.com/ckeditor/ckeditor5/pull/13949) for more information.

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

niegowski:
To make it possible we need to augment the `RemoveFormatCommand` in 2 places:

Styles feature stores classes in GHS attributes so we can't use Schema attribute property to check it. Probably some extension point would be needed here:

https://github.com/ckeditor/ckeditor5/blob/67e060929bae2bacbc02bc8b82ceadaf1f96253f/packages/ckeditor5-remove-format/src/removeformatcommand.ts#L103-L111

While applying change it should be integrated with GHS to remove only classes registered in the Styles feature:

https://github.com/ckeditor/ckeditor5/blob/67e060929bae2bacbc02bc8b82ceadaf1f96253f/packages/ckeditor5-remove-format/src/removeformatcommand.ts#L54-L56

Witoso:
When we look into this, following the discussion of [https://github.com/ckeditor/ckeditor5/issues/11580#issuecomment-1495012553.](https://github.com/ckeditor/ckeditor5/issues/11580#issuecomment-1495012553.) 

The solution could be:

*   Remove classes and styles added via GHS (including Styles plugin)

Potential problems:

*   if someone enables all classes in GHS, but some of them don't relate to styling, they will be deleted anyway. We don't have the possibility to guess which is which. This could be later on followed up by some configuration option?

Witoso:
When we look into this, following the discussion of [https://github.com/ckeditor/ckeditor5/issues/11580#issuecomment-1495012553.](https://github.com/ckeditor/ckeditor5/issues/11580#issuecomment-1495012553.) 

The solution could be:

*   Remove classes and styles added via GHS (including Styles plugin)

Potential problems:

*   if someone enables all classes in GHS, but some of them don't relate to styling, they will be deleted anyway. We don't have the possibility to guess which is which. This could be later on followed up by some configuration option?

prithvi2k2:
Strict format needs to be implemented ASAP

Witoso:
[...]

Issue #15318, opened 2023-11-06
Remove format does not remove styles applied


aasarava:
## 📝 Provide detailed reproduction steps (if any)

1. Configure CKEditor 5 version 39+ to allow `class` and `style` attributes on `a` links.
2. When using the editor, create a link, switch to HTML/source and a `class` or `style` attribute to the `a` link.
3. Switch back to standard wysiwyg view, place cursor in the link, and attempt to use the Remove Format (Tx) button to remove the `class` and `style` attributes.

### ✔️ Expected result

When placing cursor within a link, the Remove Format (Tx) button becomes active. Selecting the button removes any `class` and `style` attributes applied to the `a` link element.

### ❌ Actual result

When placing cursor within a link, or highlight the entire link, Remove Format (Tx) button remains inactive.

### ❓ Possible solution

In Issue #11580, Remove Format was updated to allow removing inline styles -- however, the `a` element was explicitly excluded (see comment https://github.com/ckeditor/ckeditor5/issues/11580#issuecomment-1495893382.) Can we include the `a` element, as it is an inline element?

Sites are often configured to allow editors to use the Styles plugin to apply classes to links -- e.g. "more link", "external link", "button". CKEditor 4 allowed removing these styles with the Remove Format button. There is currently no easy way to remove these styles.

wimleers:
Affected `drupal.org` issue: https://www.drupal.org/project/drupal/issues/3321254

Witoso:
Thanks for the report!

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

PR #16867, opened 2024-08-06
Add remove format support from block elements

Mati365:

Feature (remove-format): Remove classes and styles added via GHS. Closes https://github.com/ckeditor/ckeditor5/issues/13983

---

### Additional information

⚠️  `Page Break` elements used together with GHS seems to be affected, and it's possible to remove their styling.

### DIFF
--- a/packages/ckeditor5-html-support/src/utils.ts
+++ b/packages/ckeditor5-html-support/src/utils.ts
@@ -221,3 +221,18 @@ export function toPascalCase( data: string ): string {
 export function getHtmlAttributeName( viewElementName: string ): string {
 	return `html${ toPascalCase( viewElementName ) }Attributes`;
 }
+
+/**
+ * Checks if the given attribute name is a GHS attribute name.
+ *
+ * @param name Attribute name to check.
+ * @returns `true` if the attribute name is a GHS attribute name, `false` otherwise.
+ * @example
+ * ```js
+ * isGHSAttributeName( 'htmlPAttributes' ); // true
+ * isGHSAttributeName( 'span' ); // false
+ * ```
+ */
+export function isGHSAttributeName<S extends string>( name: S ): boolean {
+	return name.startsWith( 'html' ) && name.endsWith( 'Attributes' );
+}
--- a/packages/ckeditor5-remove-format/src/removeformatcommand.ts
+++ b/packages/ckeditor5-remove-format/src/removeformatcommand.ts
@@ -8,6 +8,8 @@
  */
 
 import type { DocumentSelection, Item, Schema } from 'ckeditor5/src/engine.js';
+import { isGHSAttributeName, type GHSViewAttributes } from '@ckeditor/ckeditor5-html-support';
+
 import { Command } from 'ckeditor5/src/core.js';
 import { first } from 'ckeditor5/src/utils.js';
 
@@ -52,7 +54,18 @@ export default class RemoveFormatCommand extends Command {
 					const itemRange = writer.createRangeOn( item );
 
 					for ( const attributeName of this._getFormattingAttributes( item, schema ) ) {
-						writer.removeAttribute( attributeName, itemRange );
+						if ( isGHSAttributeName( attributeName ) ) {
+							const oldAttribute = item.getAttribute( attributeName ) as GHSViewAttributes;
+							const newAttribute: GHSViewAttributes = {
+								...oldAttribute,
+								classes: [],
+								styles: {}
+							};
+
+							writer.setAttribute( attributeName, newAttribute, itemRange );
+						} else {
+							writer.removeAttribute( attributeName, itemRange );
+						}
 					}
 				}
 			}
@@ -101,10 +114,22 @@ export default class RemoveFormatCommand extends Command {
 	 * @returns The names of formatting attributes found in a given item.
 	 */
 	private* _getFormattingAttributes( item: Item | DocumentSelection, schema: Schema ) {
-		for ( const [ attributeName ] of item.getAttributes() ) {
+		for ( const [ attributeName, attributeValue ] of item.getAttributes() ) {
 			const attributeProperties = schema.getAttributeProperties( attributeName );
+			const isFormatting = attributeProperties && attributeProperties.isFormatting;
 
-			if ( attributeProperties && attributeProperties.isFormatting ) {
+			if ( isGHSAttributeName( attributeName ) ) {
+				const {
+					styles = {},
+					classes = []
+				} = attributeValue as GHSViewAttributes;
+
+				if ( Object.keys( styles ).length || classes.length ) {
+					yield attributeName;
+				}
+			}
+
+			if ( isFormatting ) {
 				yield attributeName;
 			}
 		}
--- a/packages/ckeditor5-html-support/src/datafilter.ts
+++ b/packages/ckeditor5-html-support/src/datafilter.ts
@@ -50,6 +50,7 @@ import {
 
 import {
 	getHtmlAttributeName,
+	isGHSAttributeName,
 	type GHSViewAttributes
 } from './utils.js';
 
@@ -554,7 +555,7 @@ export default class DataFilter extends Plugin {
 				}
 
 				for ( const attr of change.attributes.keys() ) {
-					if ( !attr.startsWith( 'html' ) || !attr.endsWith( 'Attributes' ) ) {
+					if ( !isGHSAttributeName( attr ) ) {
 						continue;
 					}
 
--- a/packages/ckeditor5-html-support/src/index.ts
+++ b/packages/ckeditor5-html-support/src/index.ts
@@ -13,6 +13,7 @@ export { default as DataSchema, type DataSchemaBlockElementDefinition } from './
 export { default as HtmlComment } from './htmlcomment.js';
 export { default as FullPage } from './fullpage.js';
 export { default as HtmlPageDataProcessor } from './htmlpagedataprocessor.js';
+export { isGHSAttributeName, type GHSViewAttributes } from './utils.js';
 export type { GeneralHtmlSupportConfig } from './generalhtmlsupportconfig.js';
 export type { default as CodeBlockElementSupport } from './integrations/codeblock.js';
 export type { default as CustomElementSupport } from './integrations/customelement.js';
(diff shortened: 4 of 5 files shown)


Witoso:
> `Page Break` elements used together with GHS seems to be affected

What do you mean by affected? That with the remove format it's possible to remove their style/classes but they are inline elements?

Mati365:
@Witoso 

It looks like it's block element and:

obraz
become:
obraz

Because `page-break` class and `page-break-after:always;` style that are stored in GHS attribute are removed

Witoso:
I wonder if #15318 as well with this solution?

Mati365:
@Witoso I'm not sure, it looks like different issue. @niegowski what do you think?

Witoso:
Yep, sorry, I didn't notice `schema.isBlock( item )`. It won't work on links, but I'm thinking if we should cover it. That way, the remove format would be fully complete.

CKEditor 4 is a good reference to check. It leaves the page break intact and doesn't remove its styles.

Witoso:
This PR/prototype that was created shows that ideally we have some logic that allows us to mark some elements as formatting.

Problems that we may want to solve with this mini-project

*   allow users to remove styling attributes on blocks
*   allow users to remove styling attributes on links (inline blocks)?
*   give API for defining which attributes are styling.

Reference to CKEditor 4: https://github.com/ckeditor/ckeditor4/blob/3fd8eafd8d805024ca3280fd95899833bb16b163/plugins/removeformat/plugin.js#L143-L182

Mati365:
It turns out that there are many leftovers after consumption of model elements, and it needs to be fixed. 

For example:
obraz
obraz

Mati365:
@niegowski Can you take a look? At this moment we remove every class and style assigned via GHS. Should we focus on making it configurable? /cc @Witoso

Witoso:
> At this moment we remove every class and style assigned via GHS. Should we focus on making it configurable? /cc [@Witoso](https://github.com/Witoso) 

I think this would be an incremental value. We could create a follow-up ticket for this API.

packages/ckeditor5-html-support/src/utils.ts
```diff
@@ -221,3 +221,18 @@ export function toPascalCase( data: string ): string {
...
+/**
+ * Checks if the given attribute name is a GHS attribute name.
+ *
+ * @param name Attribute name to check.
+ * @returns `true` if the attribute name is a GHS attribute name, `false` otherwise.
+ * @example
+ * ```js
+ * isGHSAttributeName( 'htmlPAttributes' ); // true
+ * isGHSAttributeName( 'span' ); // false
+ * ```
```
niegowski, line 234: I can't find any other place in our code with the `example` tag used.

```suggestion
 * Checks if the given attribute name is a GHS attribute name.
 *
 * ```ts
 * isGHSAttributeName( 'htmlPAttributes' ); // true
 * isGHSAttributeName( 'span' ); // false
 * ```
 *
 * @param name Attribute name to check.
 * @returns `true` if the attribute name is a GHS attribute name, `false` otherwise.
```

packages/ckeditor5-remove-format/src/removeformatcommand.ts
```diff
@@ -8,6 +8,8 @@
  */
 
 import type { DocumentSelection, Item, Schema } from 'ckeditor5/src/engine.js';
+import { isGHSAttributeName, type GHSViewAttributes } from '@ckeditor/ckeditor5-html-support';
```
niegowski, line 11: Such cross-package import should not be possible. We used to have a linter rule, but it stopped working. This import will ARAIR break the DLL build.

packages/ckeditor5-remove-format/src/removeformatcommand.ts
```diff
@@ -52,7 +54,18 @@ export default class RemoveFormatCommand extends Command {
 					const itemRange = writer.createRangeOn( item );
 
 					for ( const attributeName of this._getFormattingAttributes( item, schema ) ) {
-						writer.removeAttribute( attributeName, itemRange );
+						if ( isGHSAttributeName( attributeName ) ) {
```
niegowski, line 57: Maybe we could verify an attribute if it's a GHS by the `Schema#setAttributeProperties()` with some GHS dedicated property? `AttributeProperties` already allows for non-engine-defined properties: 
https://github.com/ckeditor/ckeditor5/blob/0c3ab39c5e211d131c865ef44b8ef872656960c0/packages/ckeditor5-engine/src/model/schema.ts#L2067

Or maybe it should use some API exposed from the GHS plugin? See other comments.

packages/ckeditor5-remove-format/src/removeformatcommand.ts
```diff
@@ -52,7 +54,18 @@ export default class RemoveFormatCommand extends Command {
...
+							const oldAttribute = item.getAttribute( attributeName ) as GHSViewAttributes;
+							const newAttribute: GHSViewAttributes = {
+								...oldAttribute,
+								classes: [],
+								styles: {}
```
niegowski, line 62: I'm not sure if a direct modification of this model element attribute is a safe approach. What if we modify the internal format of how we store that data? Maybe we could use an approach similar to the style feature:

https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-style/src/stylecommand.ts#L201-L209

https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-html-support/src/generalhtmlsupport.ts#L131

packages/ckeditor5-remove-format/src/removeformatcommand.ts
```diff
@@ -101,9 +114,20 @@ export default class RemoveFormatCommand extends Command {
...
+			if ( item.is( 'element' ) && schema.isBlock( item ) && isGHSAttributeName( attributeName ) ) {
+				const {
+					styles = {},
+					classes = []
+				} = attributeValue as GHSViewAttributes;
+
+				if ( Object.keys( styles ).length || classes.length ) {
+					yield attributeName;
+				}
+			}
```
niegowski, line 129: Same as above I'd suggest using some GHS API for such checks. Example:

https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-style/src/styleutils.ts#L159-L164

packages/ckeditor5-table/src/converters/upcasttable.ts
```diff
@@ -102,6 +102,10 @@ export default function upcastTable() {
 
 			conversionApi.consumable.consume( viewTable, { name: true } );
 
+			if ( viewTable.hasClass( 'ck-table-resized' ) ) {
+				conversionApi.consumable.consume( data.viewItem, { classes: 'ck-table-resized' } );
```
niegowski, line 106:
```suggestion
				conversionApi.consumable.consume( viewTable, { classes: 'ck-table-resized' } );
```

niegowski (changes requested):

packages/ckeditor5-remove-format/src/removeformatcommand.ts
Mati365: Fixed

packages/ckeditor5-remove-format/src/removeformatcommand.ts
Mati365: @niegowski I checked these methods and I see few problems:

1. These methods assume that the element contains only one GHS attribute, but there are cases when there are multiple formatting GHS on a single element. The example is list item [1]
2. `getGhsAttributeNameForElement` is used in these methods, so if we map model element to view element, it turns out that for list items it's not working. For example, in the [1] case these paragraph elements are resolved later to `htmlSpanAttribute` GHS attribute name, so we can't remove list item styling despite it's visible.
3. I'm not sure if we are able to remove all classes or elements according to logic of `modifyGhsAttribute` function (that is used in  `removeModelHtmlClass`) and this particular `if`: https://github.com/ckeditor/ckeditor5/blob/ck/13983/packages/ckeditor5-html-support/src/utils.ts#L180-L182

[1]
obraz

packages/ckeditor5-remove-format/src/removeformatcommand.ts
Mati365: I added `getGhsAttributesForElement` method to GHS plugin. Let's continue in this thread https://github.com/ckeditor/ckeditor5/pull/16867#discussion_r1706475443.

packages/ckeditor5-html-support/src/generalhtmlsupport.ts
```diff
@@ -76,6 +77,25 @@ export default class GeneralHtmlSupport extends Plugin {
...
+	public getGhsAttributesForElement( item: Item | DocumentSelection ): Map<string, GHSViewAttributes> {
+		return Array
+			.from( item.getAttributes() )
+			.reduce<Map<string, GHSViewAttributes>>( ( acc, [ attributeName, attributeValue ] ) => {
+				if ( isGHSAttributeName( attributeName ) ) {
+					acc.set( attributeName, attributeValue as GHSViewAttributes );
+				}
+
+				return acc;
+			}, new Map() );
```
niegowski, line 96:
```suggestion
		return Array
			.from( item.getAttributes() )
			.reduce( ( acc, [ attributeName, attributeValue ] ) => {
				if ( isGHSAttributeName( attributeName ) ) {
					acc.set( attributeName, attributeValue as GHSViewAttributes );
				}

				return acc;
			}, new Map<string, GHSViewAttributes>() );
```

niegowski (changes requested):
I checked how it works on such payload:
```html
<p style="border-left: 3px solid dodgerblue; padding-left: 5px;">
	foo
</p>
<div style="border: 2px solid orange; padding: 0 10px;">
	<p>bar</p>
</div>
```
And unfortunately, there are multiple issues:
1. It replaces the `htmlPAttributes` with an empty `{}` instead of removing that attribute from the model.
2. Only some styles are properly removed from the view by downcast converters.
3. The `<div>` element attributes are completely ignored (as in such case is a container, not a block).

packages/ckeditor5-remove-format/src/removeformatcommand.ts
Mati365: It'll not work for elements containing multiple GHS attributes, as in the comment above: https://github.com/ckeditor/ckeditor5/pull/16867#discussion_r1706475443

Mati365:
@niegowski I fixed these issues, can you take a look?

niegowski:
This PR is replaced with #18603.

PR #16890, opened 2024-08-08
Token should retry after refresh failure to avoid eventual client disconnection from the editing session

oleq:

Fix (cloud-services): The `Token` refresh should retry after failure to avoid disconnection and data loss in real-time collaboration. 

---

### Additional information

Closes https://github.com/cksource/ckeditor5-commercial/issues/6561.

### DIFF
--- a/packages/ckeditor5-cloud-services/src/token/token.ts
+++ b/packages/ckeditor5-cloud-services/src/token/token.ts
@@ -112,7 +112,7 @@ export default class Token extends /* #__PURE__ */ ObservableMixin() {
 	 * This method will be invoked periodically based on the token expiry date after first call to keep the token up-to-date
 	 * (requires {@link module:cloud-services/token/token~TokenOptions auto refresh option} to be set).
 	 *
-	 * If the token refresh fails, the method will retry in 5 seconds intervals until success or until the token gets
+	 * If the token refresh fails, the method will retry in 5 seconds intervals until success or the token gets
 	 * {@link #destroy destroyed}.
 	 */
 	public refreshToken(): Promise<InitializedToken> {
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-cloud-services/src/token/token.ts
```diff
@@ -112,14 +114,29 @@ export default class Token extends /* #__PURE__ */ ObservableMixin() {
...
+				 * TODO
+				 *
+				 * @error token-refresh-failed
+				 */
+				console.warn( 'token-refresh-failed: TODO' );
+
+				// If the refresh failed, keep trying to refresh the token. Failing to do so will eventually
+				// lead to the disconnection from the RTC service and the editing session (and potential data loss
+				// if the user keeps editing).
+				this._registerFailedRefreshTokenTimeout();
```
oleq, line 129: I wonder if this should engage only when using the default refresh callback. A custom integrator's callback could come with its own refresh logic and this would be redundant or even buggy.

packages/ckeditor5-cloud-services/src/token/token.ts
```diff
@@ -112,14 +114,29 @@ export default class Token extends /* #__PURE__ */ ObservableMixin() {
...
 
 				return this as InitializedToken;
+			} )
+			.catch( err => {
+				/**
+				 * TODO
+				 *
+				 * @error token-refresh-failed
+				 */
+				console.warn( 'token-refresh-failed: TODO' );
```
oleq, line 124: Warn and retry or retry and re-throw the error?

packages/ckeditor5-cloud-services/src/token/token.ts
oleq: I figured out that it's unlikely for a custom refresh callback to come with a retry logic and integrators who fall into this category should also benefit from auto-retry logic.

packages/ckeditor5-cloud-services/src/token/token.ts
oleq: Decision: Both. Warn to let people know what's going on. Throw for people who manually call `refreshToken()` and they need a properly rejected promise.

DawidKossowski (approved):
All changes seem to be compliant with our f2f discussion. So, LGTM.

Issue #16806, opened 2024-07-26
Multi-root editor: cannot focus root which has only block widget inside


FilipTokarski:
## 📝 Provide detailed reproduction steps (if any)

1. Go to [multi-root editor demo](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/multi-root-editor.html)
2. Remove all content from one of the roots and add only a table
3. Focus other root and then try focusing the root with the table (but do not click on the table)

### ✔️ Expected result

Focus should stay in the root, similarly to a "standard" editor:

https://github.com/user-attachments/assets/e8a13c03-4516-4aab-b4a6-99bbb8c8324d

### ❌ Actual result

Focus jumps back to the other root, it's not possible to focus a root which has only block widget:

https://github.com/user-attachments/assets/45c1f190-1130-406b-a821-5f42f85c38a3

### ❓ Possible solution

_If you have ideas, you can list them here. Otherwise, you can delete this section._

## 📃 Other details

* Browser: …
* OS: …
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

pszczesniak:
It does not need to ba a table, it can be anything that selected got a toolbar/tooltip; for example an image:

https://github.com/user-attachments/assets/d59de37f-27ad-47ed-8a72-61c7643ab170

Mati365:
@FilipTokarski / @pszczesniak I checked it on Firefox, on latest docs, and I'm not able to reproduce this issue. Can you still reproduce this?

pszczesniak:
@Mati365 yes i can (Chrome).

pszczesniak:
On Firefox it works as expected.

Mati365:
Debug notes:

1. It's not recent regression. It's possible to reproduce it on 39 version. 
1. Increasing timeout in this function shows that clicking on editable does not focus nested block element:

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/view/observer/focusobserver.ts#L106-L110

![obraz](https://github.com/user-attachments/assets/b4cb0e71-7b7b-426e-b578-1d9fb3f12bb4)

so it's performing fallback to previously focused node

f1ames:
[...]

PR #16892, opened 2024-08-09
Fix focusing block widgets inside multi root editor

Mati365:

Fix (editor-multi-root): No longer lose selection on clicking editable containing only one block element. Closes https://github.com/ckeditor/ckeditor5/issues/16806

---

### Additional information

I adjusted multi-root demos with additional plugins to make debugging block items easier.
Debug log: https://github.com/ckeditor/ckeditor5/issues/16806#issuecomment-2275645126

#### Before

https://github.com/user-attachments/assets/b2ccd2c6-7149-4d2c-9e6a-f89f3c5be012

#### After

https://github.com/user-attachments/assets/d9f89d58-3cf0-4048-be5d-385f09314ff8

https://github.com/user-attachments/assets/d52fc851-2323-42e3-afb7-c1d51ee666e3

### DIFF
--- a/packages/ckeditor5-widget/src/widget.ts
+++ b/packages/ckeditor5-widget/src/widget.ts
@@ -18,11 +18,11 @@ import {
 	type Element,
 	type Node,
 	type ViewDocumentArrowKeyEvent,
-	type ViewDocumentFragment,
 	type ViewDocumentMouseDownEvent,
 	type ViewElement,
 	type Schema,
 	type Position,
+	type EditingView,
 	type ViewDocumentTabEvent,
 	type ViewDocumentKeyDownEvent
 } from '@ckeditor/ckeditor5-engine';
@@ -33,6 +33,7 @@ import {
 	env,
 	keyCodes,
 	getLocalizedArrowKeyCodeDirection,
+	getRangeFromMouseEvent,
 	type EventInfo,
 	type KeystrokeInfo
 } from '@ckeditor/ckeditor5-utils';
@@ -288,17 +289,25 @@ export default class Widget extends Plugin {
 			return;
 		}
 
-		// Do nothing for single or double click inside nested editable.
-		if ( isInsideNestedEditable( element ) ) {
-			return;
-		}
-
 		// If target is not a widget element - check if one of the ancestors is.
 		if ( !isWidget( element ) ) {
-			element = element.findAncestor( isWidget );
+			const editableOrWidgetElement = findClosestEditableOrWidgetAncestor( element );
 
-			if ( !element ) {
+			if ( !editableOrWidgetElement ) {
 				return;
+			}
+
+			if ( isWidget( editableOrWidgetElement ) ) {
+				element = editableOrWidgetElement;
+			} else {
+				// Pick view range from the point where the mouse was clicked.
+				const clickTargetFromPoint = getElementFromMouseEvent( view, domEventData );
+
+				if ( clickTargetFromPoint && isWidget( clickTargetFromPoint ) ) {
+					element = clickTargetFromPoint;
+				} else {
+					return;
+				}
 			}
 		}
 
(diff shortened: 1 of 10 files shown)


niegowski (changes requested):
Widgets have a dedicated handling of the selection. The DOM selection is handled differently while a widget is selected, it's anchored in a fake selection container. I feel like the fix should be located in this handler of the widget: https://github.com/ckeditor/ckeditor5/blob/dcb087fa6a5bb5568546a31f00e96176fe8af8bb/packages/ckeditor5-widget/src/widget.ts#L276-L320

Mati365:
Unfortunately it's still a bit buggy :/ It looks like, while cursor is properly placed in element, the table is receiving incorrectly focus. Probably related to `element.childCount` check.  

https://github.com/user-attachments/assets/df2586bf-85d1-4106-a0b0-be15ef36d1c0

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -630,9 +655,21 @@ function isInsideNestedEditable( element: ViewElement ) {
 		currentElement = currentElement.parent;
 	}
 
+	/* istanbul ignore next -- @preserve */
 	return false;
 }
```
Mati365, line 661: It looks like incorrect `istanbul` report, because `return false;` is reached during the tests.

packages/ckeditor5-engine/src/view/view.ts
```diff
@@ -810,6 +810,48 @@ export default class View extends /* #__PURE__ */ ObservableMixin() {
...
 }
 
+/**
+ * Get the view range from the given DOM event.
+ *
+ * @param view The view instance.
+ * @param domEvent The DOM event.
+ * @returns The view range or `null` if the range cannot be created.
+ */
+export function getPointViewRange(
```
niegowski, line 820: The `view.ts` does not look like a perfect spot for such a helper function especially since it is not used here. I'd suggest extracting this helper to the `utils` package as other DOM helpers, for example this one:

https://github.com/ckeditor/ckeditor5/blob/08270aafcc6e7e4fb4f4a4d2ebbc7451f031d303/packages/ckeditor5-utils/src/dom/scroll.ts#L60

Note that currently this helper expects a `View` instance as a parameter, but it is used only once so we could extract DOM-focused behavior to a helper util and use DOM to view mapping outside the helper itself.

niegowski (changes requested):
Partial review.

Mati365:
@niegowski I moved function to utils. Can you take a look?

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -612,25 +621,50 @@ export default class Widget extends Plugin {
...
+	}
+
+	const viewPosition = viewRange.start;
+
+	// Click after a widget tend to return position at the end of the editable element
+	// so use the node before it if range is at the end of a parent.
+	const viewNode = viewPosition.parent.is( 'editableElement' ) && viewPosition.isAtEnd && viewPosition.nodeBefore || viewPosition.parent;
 
-		currentElement = currentElement.parent;
+	// TODO There was a comment about broken drag&drop but without further explanation. What was happening exactly?
```
niegowski, line 662: Please explain what the original comment was about.

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -621,7 +621,10 @@ export default class Widget extends Plugin {
 }
 
 /**
- * TODO
+ * Finds the closest ancestor element that is either an editable element or a widget.
+ *
+ * @param element - The element from which to start searching.
```
niegowski, line 626:
```suggestion
 * @param element The element from which to start searching.
```

niegowski (approved):

kubaklatt:
It is not a regression, but I managed to find two more scenarios that cause identical problem on Safari - https://github.com/ckeditor/ckeditor5/issues/16978

charlttsie:
The issue still reproduces on Chrome when using tab to switch to a root with a block widget only - https://github.com/ckeditor/ckeditor5/issues/16981

PR #16909, opened 2024-08-12
Avoid creating arrays in getNodeStartOffset

sunesimonsen:
`NodeList.getNodeStartOffset` gets called quite a lot, so it is better not to create a new array slice every time.

### DIFF
--- a/packages/ckeditor5-engine/src/model/nodelist.ts
+++ b/packages/ckeditor5-engine/src/model/nodelist.ts
@@ -80,7 +80,16 @@ export default class NodeList implements Iterable<Node> {
 	public getNodeStartOffset( node: Node ): number | null {
 		const index = this.getNodeIndex( node );
 
-		return index === null ? null : this._nodes.slice( 0, index ).reduce( ( sum, node ) => sum + node.offsetSize, 0 );
+		if ( index === null ) {
+			return null;
+		}
+
+		let sum = 0;
+		for ( let i = 0; i < index; i++ ) {
+			sum += this._nodes[ i ].offsetSize;
+		}
+
+		return sum;
 	}
 
 	/**


scofalik:
Hi @sunesimonsen!

Can you provide some stats on how it impacts editor's performance?

sunesimonsen:
@scofalik I haven't performance tested the different, but with a 100 paragraph lorem ipsum document in our setup this method is called `12697` times when I insert a character, so I think it might be great to be as fast as possible 😅

sunesimonsen:
@scofalik was have a lot of performance problems with large document (documents with many nodes), so we are trying to see where we can shave off some milliseconds when setting and getting data from CKEditor. 

We already reduced how much we call `editor.getData()` to the absolute minimum, but setting the data is now the main problem taking +6 seconds for some of our customer articles (around 1MB).

I tried applying the change to our CKEditor dev setup and for a large document we use for testing and I'm seeing [this call](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/controller/datacontroller.ts#L471) go from  3727 ms to 1462 ms when setting the initial data.

scofalik:
That's sounds very promising! We planned to work on editor performance for big documents, it was planned for the second half of this year. We already have some own notes from earlier testings we did, we will add this one to our notes. Thanks for sharing this finding, we'll try to test it and include it in the editor sooner than later.

scofalik:
Thanks again for this suggestion. We plan to include it in next October CKEditor 5 release.

scofalik:
Introduced in #17144.

scofalik:
Hi @sunesimonsen, I wonder if you could share with us samples of the documents you struggle with. It could be an anonymized version of one of your documents (e.g. replace all letters with "a", remove all numbers, replace all `href` and `src` with `#`). We would like to add it our database of samples that we look at when testing performance fixes.

If you would like to share but do not feel comfortable with sharing it here, please reach us on support: https://ckeditor.com/contact/.

PR #16913, opened 2024-08-13
Avoid creating unnecessary arrays in stylesmap.getStylesNames

sunesimonsen:
When you have inline styles in your document `getStylesNames` gets called a lot, this PR optimises the code to avoid creating unnecessary arrays.

I tested the code in our internal editor setup with a 861kb document where `getStyleNames` gets called 112056 times. 

Before the change [this method call](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/controller/datacontroller.ts#L471) took 1463.974853515625 ms
after the change it takes 922.365966796875 ms.

### DIFF
--- a/packages/ckeditor5-engine/src/view/stylesmap.ts
+++ b/packages/ckeditor5-engine/src/view/stylesmap.ts
@@ -624,23 +624,22 @@ export class StylesProcessor {
 	 * @param styles Object holding normalized styles.
 	 */
 	public getStyleNames( styles: Styles ): Array<string> {
+		const styleNamesKeysSet = new Set<string>();
+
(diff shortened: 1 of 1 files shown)


scofalik:
Thanks for this suggestion! We plan to include it in next October CKEditor 5 release.

scofalik:
Introduced in #17144.

scofalik:
Hi @sunesimonsen, I wonder if you could share with us samples of the documents you struggle with. It could be an anonymized version of one of your documents (e.g. replace all letters with "a", remove all numbers, replace all `href` and `src` with `#`). We would like to add it our database of samples that we look at when testing performance fixes.

PR #16917, opened 2024-08-13
A HTML collection is already iterable, no need to turn it into an array

sunesimonsen:
I don't know how much of a differences this will make, but creating one less array is a good thing :-)


scofalik:
Thanks for this suggestion! We plan to include it in the October CKEditor 5 release.

scofalik:
Introduced in #17144.

PR #16948, opened 2024-08-20
Docs: Update the Laravel self-hosted guide.

gorzelinski:

Docs: Update the Laravel self-hosted guide.

---

### Additional information

### DIFF
--- a/docs/getting-started/integrations/laravel.md
+++ b/docs/getting-started/integrations/laravel.md
@@ -26,7 +26,9 @@ This guide assume you have a Laravel project. You can create a basic Laravel pro
 
 ## Integrating using ZIP
 
-There are two ways to acquire a ZIP archive with CKEditor&nbsp;5. You can use the mentioned [builder](https://ckeditor.com/builder/) to customize and download an editor suited to your needs. You can also use one of the generic ZIP archives [on the download page](https://ckeditor.com/ckeditor-5/download/#zip).
+<info-box>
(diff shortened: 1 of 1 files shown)


docs/getting-started/integrations/laravel.md
```diff
@@ -26,9 +26,7 @@ This guide assume you have a Laravel project. You can create a basic Laravel pro
 
 ## Integrating using ZIP
 
-<info-box>
-	Our new CKEditor&nbsp;5 Builder does not provide ZIP output yet &ndash; but it will in the future. In the meantime, you can use one of the generic ZIP packages provided [on the download page](https://ckeditor.com/ckeditor-5/download/#zip).
-</info-box>
+There are two ways to acquire a ZIP archive with CKEditor&nbsp;5. You can use the mentioned [builder](https://ckeditor.com/builder/) to customize and download an editor suited to your needs. You can also use one of the generic ZIP archives [on the download page](https://ckeditor.com/ckeditor-5/download/#zip).
```
gorzelinski, line 29: I wonder if we want to promote those generic ZIP archives.

Witoso: Technically, Builder doesn't offer ZIP (files of editor downloaded) it offers sample project but in NPM/CDN.

It will be added, I would keep old copy for now till builder has it. 

> I wonder if we want to promote those generic ZIP archives.

Same here, but we can test it on those more backend integrations like Laravel, etc.

gorzelinski: Oh, you're right! There is a ZIP output, but not the ZIP we want for this integration. The old copy makes more sense, then.

PR #16953, opened 2024-08-21
Test

Mati365:


Mati365:
Perf test must be done

niegowski (changes requested):
This makes code more difficult to understand. It's easy to forget to use those helpers or where they are defined. Their names are hard to understand what exactly is going to happen.

I'd suggest using `Object.create( null )` approach instead of `{}`. It could be cleaner code and a plain object without a prototype.

Mati365:
@niegowski Changed, looks like it works 🎉

Issue #16958, opened 2024-08-21
Contextual balloon does not work with non-empty selection


Tadaz:
## 📝 Provide detailed reproduction steps (if any)

I have just migrated to v.43 and I noticed that a couple of my custom plugins are not working properly. Both of them have the same issue that can be replicated in [your docs page](https://ckeditor.com/docs/ckeditor5/latest/framework/tutorials/abbreviation-plugin/abbreviation-plugin-level-2.html#demo).

As per the recording below you can see that trying to use your example "Abbreviation" is not possible.  When content is selected the contextual balloon does not appear and there is no error/warning about that.
![bug](https://github.com/user-attachments/assets/c9bd99f5-a4d8-419f-9953-71d76db8455b)

I have tried multiple browsers - the issue persists.

### ✔️ Expected result

The abbreviation contextual balloon should appear all the time.

### ❌ Actual result

The abbreviation contextual balloon appears only when the 0 coordinate is included in selection.

## 📃 Other details

* Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36
* OS: macOS Sonoma 14.4
* First affected CKEditor version: 43
* Installed CKEditor plugins: 
  * @ckeditor/ckeditor5-vue
  * ckeditor5

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Witoso:
Thanks for the report, we will begin investigation asap. Potential cause: #16739

Issue #16889, opened 2024-08-08
Issue with ContextualBalloon Not Appearing After Updating to CKEditor 43.0.0


w-beom:
## 📝 Provide detailed reproduction steps (if any)

After updating the CKEditor version to 43.0.0, an issue has occurred.

Previously, we were using a custom plugin that utilizes the ContextualBalloon. When clicking the button of the plugin, the ContextualBalloon would appear, allowing text input in the input field.

However, after the update, the ContextualBalloon does not appear when using the custom plugin with the SlashCommand. 

What should we do to resolve this?

### ✔️ Expected result

![image](https://github.com/user-attachments/assets/ae4389e5-9d81-4cc0-b447-9ec4995f6a3e)

### ❌ Actual result

![image](https://github.com/user-attachments/assets/db7a6d54-68b6-4d20-9192-4f4a0185c355)

![image](https://github.com/user-attachments/assets/c3e3243f-89e3-45a6-ae2b-efd186bea6e1)

## 📃 Other details

It works correctly in version 42.0.0. The commit related to the issue in version 43.0.0 is suspected to be this one: https://github.com/ckeditor/ckeditor5/issues/16739
---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Mati365:
Thanks for report! I'll take a look. It looks like something is trying to pin itself on the hidden element.

Witoso:
Hi @w-beom, please reach out to us via the support channel. We would like to see the code samples, etc. Thank you!

Witoso:
The best way to create a support ticket is [contact form](https://ckeditor.com/contact/).

Witoso:
We potentially identified the issue in #16959.

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #16959, opened 2024-08-22
Fix pinning the balloons to `#text` nodes

Mati365:

Fix (ui): Restored the ability to pin balloons to text nodes in the DOM tree. Closes https://github.com/ckeditor/ckeditor5/issues/16958 https://github.com/ckeditor/ckeditor5/issues/16889.

---

### Additional information

1. Introduced in https://github.com/ckeditor/ckeditor5/issues/16739 , we added visibility check before mounting of tooltips to avoid displaying them on hidden elements (which was causing another issue). 
2. Regression was caused by `isVisible` function. It was returning `false` for `$text` nodes, so it was impossible to pin balloons to such nodes.
3. I extracted docs example to standalone manual test. Apply this patch to your git repo to test it on your own [pin-balloons-demo.patch](https://github.com/user-attachments/files/16706397/pin-balloons-demo.txt)

#### Before

https://github.com/user-attachments/assets/b3019683-74ba-4697-a683-6ef9ea719563

#### After

https://github.com/user-attachments/assets/7fb27a4e-b564-4c36-a42a-fc77fb156050

### DIFF
--- a/packages/ckeditor5-utils/src/dom/isvisible.ts
+++ b/packages/ckeditor5-utils/src/dom/isvisible.ts
@@ -17,6 +17,18 @@
  * **Note**: This helper does not check whether the element is hidden by cropping, overflow, etc..
  * To check that, use {@link module:utils/dom/rect~Rect} instead.
  */
-export default function isVisible( element: HTMLElement | null | undefined ): boolean {
-	return !!( element && element.getClientRects && element.getClientRects().length );
+export default function isVisible( element: Text | HTMLElement | null | undefined ): boolean {
+	if ( !element ) {
+		return false;
+	}
+
+	if ( element instanceof Text ) {
+		return isVisible( element.parentElement );
+	}
+
+	if ( element.getClientRects ) {
+		return !!( element.getClientRects().length );
+	}
+
+	return false;
 }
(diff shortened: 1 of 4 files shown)


Witoso:
Does it potentially fix this scenario: #16889 as well?

Mati365:
@Witoso Most likely.

packages/ckeditor5-utils/src/dom/isvisible.ts
```diff
@@ -17,6 +17,18 @@
...
  * To check that, use {@link module:utils/dom/rect~Rect} instead.
  */
-export default function isVisible( element: HTMLElement | null | undefined ): boolean {
-	return !!( element && element.getClientRects && element.getClientRects().length );
+export default function isVisible( element: Text | HTMLElement | null | undefined ): boolean {
+	if ( !element ) {
+		return false;
+	}
+
+	if ( element instanceof Text ) {
```
niegowski, line 25:
```suggestion
	if ( element.nodeType == Node.TEXT_NODE ) {
```

niegowski (changes requested):
We use the `nodeType` property of a DOM node to verify node type.

packages/ckeditor5-utils/src/dom/isvisible.ts
```diff
@@ -3,6 +3,8 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
+import isText from './istext.js';
+
```
niegowski, line 7: Imports should be after the `@module` block.

niegowski (approved):

PR #16962, opened 2024-08-22
hgroup and summary should work with source editing

### LINKED ISSUE #16947: hgroup and summary do not accept heading elements

1. Enter source editing mode
2. Create an hgroup element and add a heading element
3. Create a details element with a child summary element and add a heading element
4. Toggle back to design mode
5. Toggle back to source mode

### ✔️ Expected result

Code is preserved as written.

### ❌ Actual result

hgroup and summary elements are removed

### ❓ Possible solution

submitting PR to fix this in schema definitions

## 📃 Other details

* Browser: Chrome, Firefox
* OS: Windows 11, Ubuntu

---

illia-stv:

Fix (html-support): `hgroup` and `summary` elements should work with source editing. Closes #16947 .

---

### Additional information

### DIFF
--- a/packages/ckeditor5-html-support/src/schemadefinitions.ts
+++ b/packages/ckeditor5-html-support/src/schemadefinitions.ts
@@ -349,7 +349,7 @@ export default {
 			model: 'htmlHgroup',
 			view: 'hgroup',
 			modelSchema: {
-				allowIn: '$container',
+				allowIn: [ '$root', '$container' ],
 				allowChildren: [
 					'paragraph',
 					'htmlP',
--- a/packages/ckeditor5-html-support/src/schemadefinitions.ts
+++ b/packages/ckeditor5-html-support/src/schemadefinitions.ts
@@ -279,8 +279,17 @@ export default {
 			model: 'htmlSummary',
 			view: 'summary',
 			modelSchema: {
-				allowChildren: '$text',
+				allowChildren: [
+					'htmlH1',
+					'htmlH2',
+					'htmlH3',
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-html-support/src/schemadefinitions.ts
```diff
@@ -279,8 +279,17 @@ export default {
...
+					'htmlH1',
+					'htmlH2',
+					'htmlH3',
+					'htmlH4',
+					'htmlH5',
+					'htmlH6',
+					'$text'
+				],
 				allowIn: 'htmlDetails',
+				isLimit: true,
```
niegowski, line 292: Why is this added?

jwylarsen: I added it because I found that, without it, CK would remove the summary element if it contained a heading. If it works without it, great.

packages/ckeditor5-html-support/src/schemadefinitions.ts
```diff
@@ -341,7 +349,10 @@ export default {
 			model: 'htmlHgroup',
 			view: 'hgroup',
 			modelSchema: {
+				allowIn: '$root',
```
jwylarsen, line 352: Why $root and not $container? hgroup can go anywhere

niegowski: I think it would be safe to use `$container` here.

jwylarsen (approved):

niegowski (approved):

PR #16980, opened 2024-08-27
Nested menus should not get their panels focused when the main button is clicked

### LINKED ISSUE #16857: Nested item gets focused in menu bar after clicking category button

https://github.com/user-attachments/assets/1f461809-3fbc-4fd6-bdab-5eade6a97a78

After clicking the "category" button like "Image", or "Text", the first item inside gets focused and stays focused. Earlier we decided to show focus only on keyboard interaction.

oleq:

Fix (ui): Nested menus should not get their panels focused when the main button is clicked. Closes #16857.


DawidKossowski (approved):
LGTM

Issue #16978, opened 2024-08-27
Multi-root editor on Safari: cannot focus root which has table of contents or media inside


kubaklatt:
## 📝 Provide detailed reproduction steps (if any)

1. Open a manual test that uses multi-root, for example [all-features-rtc-multi-root.html](http://localhost:8125/ckeditor5/external/ckeditor5-commercial/packages/ckeditor5-collaboration/tests/manual/all-features-rtc-multi-root.html)
2. Insert table of contents or media inside one of the roots
3. Focus on the other root and then try focusing the root with content inside (but do not click on the table of contents or the media)

### ✔️ Expected result

Focus should stay in the root, just like it looks on for example Google Chrome

https://github.com/user-attachments/assets/f812543d-99fa-4524-a1e9-bb461cb5ce64

### ❌ Actual result

Focus jumps back to the other root

https://github.com/user-attachments/assets/36482ca7-f863-47c3-902c-1b52dea29a9a

### ❓ Possible solution

_If you have ideas, you can list them here. Otherwise, you can delete this section._

## 📃 Other details

* Browser: Safari
* OS: any
* First affected CKEditor version: N/A
* Installed CKEditor plugins: N/A

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Mati365:
It looks like range calculation in Chrome and Safari is not standardized enough to make it work identically. It's not regression on our end, but we can bypass it I think. More explanation here:  https://github.com/ckeditor/ckeditor5/pull/16997

pomek:
Fixed in https://github.com/ckeditor/ckeditor5/pull/16997.

PR #16997, opened 2024-08-29
Fix inconsistent behave of `getRangeFromMouseEvent` on Safari

Mati365:

Internal (widget): It's now possible to focus root containing only table of contents or embed elements on Safari. Closes https://github.com/ckeditor/ckeditor5/issues/16978

---

### Additional information

It looks like `caretRangeFromPoint` behaves differently on Chrome and Safari when `click` event target occurs in an element that contains `contenteditable=false`.

On Chrome, `contenteditable=false` is being focused.
On Safari it is not, and it used to focus near `contenteditable=true` in the other root element.

#### Before

https://github.com/user-attachments/assets/36482ca7-f863-47c3-902c-1b52dea29a9a

#### After

https://github.com/user-attachments/assets/8d95c398-cec6-4a0f-8871-f5bda95be549

### DIFF
--- a/packages/ckeditor5-widget/src/widget.ts
+++ b/packages/ckeditor5-widget/src/widget.ts
@@ -24,7 +24,9 @@ import {
 	type Position,
 	type EditingView,
 	type ViewDocumentTabEvent,
-	type ViewDocumentKeyDownEvent
+	type ViewDocumentKeyDownEvent,
+	type ViewNode,
+	type ViewRange
 } from '@ckeditor/ckeditor5-engine';
 
 import { Delete, type ViewDocumentDeleteEvent } from '@ckeditor/ckeditor5-typing';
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-widget/src/widget.ts
```diff
@@ -649,13 +651,16 @@ function findClosestEditableOrWidgetAncestor( element: ViewElement ): ViewElemen
...
 	const domRange = getRangeFromMouseEvent( domEventData.domEvent );
-
-	if ( !domRange ) {
-		return null;
+	let viewRange: ViewRange | null = null;
+
+	if ( domRange ) {
+		viewRange = view.domConverter.domRangeToView( domRange );
+	} else {
+		// Fallback to create range on target element. It happens frequently on Safari browser.
```
niegowski, line 659:
```suggestion
		// Fallback to create range in target element. It happens frequently on Safari browser.
```

niegowski (approved):

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -649,13 +651,16 @@ function findClosestEditableOrWidgetAncestor( element: ViewElement ): ViewElemen
...
 function getElementFromMouseEvent( view: EditingView, domEventData: DomEventData<MouseEvent> ): ViewElement | null {
 	const domRange = getRangeFromMouseEvent( domEventData.domEvent );
-
-	if ( !domRange ) {
-		return null;
+	let viewRange: ViewRange | null = null;
+
+	if ( domRange ) {
+		viewRange = view.domConverter.domRangeToView( domRange );
+	} else {
```
niegowski, line 658: Maybe we should verify it the `target` is here as this is a view node so it could have no mapping from DOM:

```suggestion
	} else if ( domEventData.target ) {
```

The `target` in the event data is marked as it always returns an element but it could be a mistake.

packages/ckeditor5-widget/src/widget.ts
```diff
@@ -280,6 +282,11 @@ export default class Widget extends Plugin {
 		const viewDocument = view.document;
 		let element: ViewElement | null = domEventData.target;
 
+		// If target is null abort.
```
niegowski, line 285: I'd replace this comment with an explanation of why and not what.

niegowski (approved):

PR #17031, opened 2024-09-03
Docs: Update the Laravel CDN guide.

gorzelinski:

Docs: Update the Laravel CDN guide.

---

### Additional information

### DIFF
--- a/docs/getting-started/integrations-cdn/laravel.md
+++ b/docs/getting-started/integrations-cdn/laravel.md
@@ -1,12 +1,12 @@
 ---
 category: cloud
 meta-title: Compatibility with Laravel when using CDN | CKEditor 5 documentation
-meta-description: Integrate CKEditor 5 using Laravel with CDN.
+meta-description: Integrate CKEditor 5 with Laravel using CDN.
 order: 70
 menu-title: Laravel
 ---
 
-# Compatibility with Laravel
+# Compatibility with Laravel using CDN
 
 As a pure JavaScript/TypeScript application, CKEditor&nbsp;5 will work inside any environment that supports such components. While we do not offer official integrations for any non-JavaScript frameworks, you can include a custom configuration of CKEditor&nbsp;5 in a non-JS framework of your choice, for example, the PHP-based [Laravel](https://laravel.com/).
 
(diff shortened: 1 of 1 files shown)


gorzelinski:
I added a path for premium features, but I'm unsure if we want to promote premium features in non-JS integrations.

godai78 (approved):

docs/getting-started/integrations-cdn/laravel.md
```diff
@@ -49,80 +46,121 @@ Once the project has been prepared, create an `assets/vendor/ckeditor5.js` file
...
+			<p>Hello from CKEditor 5!</p>
+		</div>
+    </div>
+</body>
+</html>
+```
+
+Both previously attached scripts expose global variables named `CKEDITOR` and `CKEDITOR_PREMIUM_FEATURES`. You can use them to access the editor class and plugins. In our example, we use object destructuring (JavaScript feature) to access the editor class from the open-source global variable with a basic set of plugins. You can access premium plugins from the other variable the same way. Then, we pass the whole configuration to the `create()` method. Be aware that you need a proper {@link getting-started/licensing/license-key-and-activation license key} to use premium features.
 
 ```js
```
Witoso, line 84: From the flow of text, it's a bit confusing where this code block should appear, create some placeholder in the index.html?

Witoso: I would start showing it in the <script> from start I think.

docs/getting-started/integrations-cdn/laravel.md
```diff
@@ -26,17 +26,14 @@ This guide assume you have a Laravel project. You can create a basic Laravel pro
 
 ## Integrating from CDN
```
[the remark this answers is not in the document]
gorzelinski: I have the same doubts. I don't like both "integration" and "installation" regarding CDN. At the same time, it needs to be in the installation category. I gladly change it to "using."

docs/getting-started/integrations-cdn/laravel.md
gorzelinski: I wanted to show the JavaScript snippet, explain it a bit, and then show it in the template. Actually, it's a good idea.

PR #17033, opened 2024-09-03
Update explanation for the `ckeditor-duplicated-modules` error.

filipsobol:

Docs: Update explanation for the `ckeditor-duplicated-modules` error.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-utils/src/version.ts
+++ b/packages/ckeditor5-utils/src/version.ts
@@ -31,14 +31,14 @@ if ( globalThis.CKEDITOR_VERSION ) {
 	 * # Check dependency versions
 	 *
 	 * First, make sure that you use the latest version of all CKEditor&nbsp;5 dependencies.
-	 * Depending on the installation method you used, you should check the versions of the `ckeditor5`,
+	 * Depending on the installation method, you should check the versions of the `ckeditor5`,
 	 * `ckeditor5-premium-features`, or `@ckeditor/ckeditor5-<NAME>` packages. If you cannot update
-	 * to the latest version, make sure that all the CKEditor&nbsp;5 packages are
-	 * in the exact same version.
+	 * to the latest version, ensure that all the CKEditor&nbsp;5 packages are
+	 * in the same version.
 	 *
-	 * If you use third-party plugins, make sure to update them, too. If they are not compatible
+	 * If you use third-party plugins, make sure to update them, too. If they are incompatible
 	 * with the version of CKEditor&nbsp;5 you use, you may need to downgrade the CKEditor&nbsp;5 packages
-	 * (which we do not recommend). Ask the author of said plugin to upgrade the dependencies,
+	 * (which we do not recommend). Ask the plugin's author to upgrade the dependencies,
 	 * or fork their project and update it yourself.
 	 *
 	 * # Check imports
@@ -47,8 +47,8 @@ if ( globalThis.CKEDITOR_VERSION ) {
 	 *
 	 * **The {@glink updating/nim-migration/migration-to-new-installation-methods new installation methods}
 	 * are designed to prevent module duplication, so if you are not using them yet, you should consider
-	 * updating your project**. However, for backwards compatibility, a number of legacy installation methods
-	 * is still supported, and mixing them may result in module duplication.
+	 * updating your project**. However, several legacy installation methods are still supported for backward 
+	 * compatibility, and mixing them may result in module duplication.
 	 *
 	 * These are the most common import methods of the CKEditor&nbsp;5 packages.
 	 *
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-utils/src/version.ts
```diff
@@ -24,138 +24,162 @@ declare global {
...
-	 * import Highlight from '@ckeditor/ckeditor5-highlight/src/highlight';
+	 * This error is thrown when, due to a mistake in the way CKEditor&nbsp;5 was installed,
+	 * imported, or initialized, some of its modules were evaluated and executed twice.
+	 * Duplicate modules inevitably lead to runtime errors and increased bundle size.
+	 *
+	 * # Check dependency versions
+	 *
+	 * First, make sure that you use the latest version of all CKEditor&nbsp;5 dependencies.
+	 * Depending on the installation method you used, you should check the versions of the `ckeditor5`,
+	 * `ckeditor5-premium-features`, or `@ckeditor/ckeditor5-<NAME>` packages. If you cannot update
```
filipsobol, line 35: I used `@ckeditor/ckeditor5-<NAME>` instead of `@ckeditor/ckeditor5-*` because paths including `*/` would close the JavaScript comment. If I added an escape character before the `/` (`*\/`), then the import example was invalid.

packages/ckeditor5-utils/src/version.ts
```diff
@@ -24,138 +24,162 @@ declare global {
...
+	 * If you are using the {@glink updating/nim-migration/migration-to-new-installation-methods new installation methods},
+	 * you should only import code from the `ckeditor5` and `ckeditor5-premium-features` packages.
+	 * Do not import code from the `@ckeditor/ckeditor5-<NAME>` packages unless you follow
+	 * the {@glink getting-started/setup/optimizing-build-size Optimizing build size} guide and the imports from
+	 * the `@ckeditor/ckeditor5-<NAME>` packages end with `/dist/index.js`.
+	 *
+	 * If you are using a CDN, make sure that some files are not included twice in your project.
+	 *
+	 * Examples:
+	 *
```
godai78, line 92: This kinda looks as if importing Highlights twice was an OK action to do.

packages/ckeditor5-utils/src/version.ts
```diff
@@ -24,138 +24,162 @@ declare global {
...
-	 * import Essentials from '@ckeditor/ckeditor5-essentials/src/essentials';
-	 * import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
-	 * import Bold from '@ckeditor/ckeditor5-basic-styles/src/bold';
-	 * import Italic from '@ckeditor/ckeditor5-basic-styles/src/italic';
+	 * If you are missing some features from the
+	 * {@glink getting-started/legacy/installation-methods/predefined-builds#plugins-included-in-the-predefined-builds list of plugins},
+	 * you should switch to the {@glink updating/nim-migration/migration-to-new-installation-methods new installation methods}
+	 * which do not have this limitation.
+	 * If you cannot migrate to the new installation methods, you can try the
+	 * {@glink getting-started/legacy/installation-methods/predefined-builds#superbuild superbuild} instead,
```
godai78, line 125: Is it still available?

packages/ckeditor5-utils/src/version.ts
```diff
@@ -24,138 +24,162 @@ declare global {
...
+	 * This is a matrix showing which installation methods are compatible with each other:
+	 *
+	 * |                  | NIM | Optimized build | Predefined builds | Default imports | `src` | DLL builds |
+	 * |------------------|-----|-----------------|-------------------|-----------------|-------|------------|
+	 * | NIM              | ✅  | ✅              | ❌                | ❌              | ❌    | ❌         |
+	 * | Optimized builds | ✅  | ✅              | ❌                | ❌              | ❌    | ❌         |
+	 * | Predefined build | ❌  | ❌              | ✅                | ❌              | ❌    | ❌         |
+	 * | Default imports  | ❌  | ❌              | ❌                | ✅              | ✅    | ❌         |
+	 * | `src`            | ❌  | ❌              | ❌                | ✅              | ✅    | ❌         |
+	 * | DLL builds       | ❌  | ❌              | ❌                | ❌              | ❌    | ✅         |
```
gorzelinski, line 75: Do we have something like this in non-API docs? It looks useful.

gorzelinski (commented):
I added some suggestions.

packages/ckeditor5-utils/src/version.ts
filipsobol: We don't, and I'm not sure if we should. Because we only promote using NIM (and ways to optimize it), the average developers should not be aware of other methods unless they do something wrong and get this error.

I don't have a strong opinion on this, so feel free to use it in order places if you want.

packages/ckeditor5-utils/src/version.ts
filipsobol: Because all predefined builds were removed from CDN at the same time, the developers can migrate from one to another as long as they stay on the same CKEditor version.

packages/ckeditor5-utils/src/version.ts
filipsobol: I changed `Examples` to `Examples of valid and invalid import paths` to better explain this. Do you think it's enough?

godai78: Yeah, it's explicit now.

godai78 (approved):

packages/ckeditor5-utils/src/version.ts
gorzelinski: > the average developers should not be aware of other methods unless they do something wrong and get this error.

That's a good point. I agree 👍

gorzelinski (approved):

PR #17034, opened 2024-09-03
Reverted recent change to move `config.htmlEmbed.sanitizeHtml` to a top-level config property

scofalik:
Other (core,html-embed): Reverted recent change to move `config.htmlEmbed.sanitizeHtml` to a top-level config property (`config.sanitizeHtml`). Starting from v43.1.0 `config.htmlEmbed.sanitizeHtml` is no longer deprecated.

MINOR BREAKING CHANGE: In v43.0.0 we made a decision to move `config.htmlEmbed.sanitizeHtml` to a top-level property `config.sanitizeHtml`, so it can be used by multiple features (HTML Embed and Merge Fields). However, we realized that it was a wrong decision to expose such a sensitive property in a top-level config property and decided to revert that change. Starting with v43.1.0 you should again use `config.htmlEmbed.sanitizeHtml` and `config.mergeFields.sanitizeHtml`. The editor will throw an error if `config.sanitizeHtml` is used.


Dumluregn (dismissed):

packages/ckeditor5-core/src/editor/editor.ts
```diff
@@ -281,6 +280,15 @@ export default abstract class Editor extends /* #__PURE__ */ ObservableMixin() {
 	constructor( config: EditorConfig = {} ) {
 		super();
 
+		if ( 'sanitizeHtml' in config ) {
+			/**
+			 * Configuration property `config.sanitizeHtml()` was removed in CKEditor version 43.1.0 and is no longer supported.
```
Reinmar, line 285: I'd add a link to [https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-43.html](https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-43.html) and a short explanation there what happened and why.

Reinmar:
> MINOR BREAKING CHANGE: In v43.0.0 we made a decision to move `config.htmlEmbed.sanitizeHtml` to a top-level property `config.sanitizeHtml`, so it can be used by multiple features (HTML Embed and Merge Fields). However, we realized that it was a wrong decision to expose such a sensitive property in a top-level config property and decided to revert that change. Starting with v43.1.0 you should again use `config.htmlEmbed.sanitizeHtml` and `config.mergeFields.sanitizeHtml`. The editor will throw an error if `config.sanitizeHtml` is used.  
>  

Based on this I'd also add a section to the migration guide for better visibility and so we could link from the error code somewhere.

The migration guide could include a bit more context:

> In v43.0.0 we made a decision to move `config.htmlEmbed.sanitizeHtml` to a top-level property `config.sanitizeHtml`, so it can be used by multiple features (HTML embed and newly introduced Merge fields).
> 
> However, we realized that it was a wrong decision to expose such a sensitive property in a top-level config property. We are worried that integrators may incorrectly assume that this callback would allow them sanitize the entire output from CKEditor. Therefore, we decided to revert that change.
> 
> Starting with v43.1.0 you should again use `config.htmlEmbed.sanitizeHtml` and `config.mergeFields.sanitizeHtml`. The editor will throw an error if `config.sanitizeHtml` is used.
> 
> Note: CKEditor by default prevents execution of scripts in the editor content, while the content is being edited inside the editor. There are features (such as General HTML support or HTML embed) that can be configured to make CKEditor produce HTML output that contains executable scripts. Most importantly, though, CKEditor is a front-end component working in a browser and therefore, as an integrator, it is your responsibility to sanitize the content before it is displayed on your website or on other potentially vulnerable mediums.

Something like this.

godai78 (dismissed):
Only reviewed documentation, didnae touch the code.

Dumluregn (changes requested):

godai78 (approved):

Dumluregn (approved):

PR #17054, opened 2024-09-06
Schema definition for `bookmark`.

pszczesniak:

Internal (bookmark): Schema definition for `bookmark`.

---

### Additional information


packages/ckeditor5-bookmark/tests/bookmarkediting.js
```diff
@@ -0,0 +1,55 @@
...
+			} )
+			.then( newEditor => {
+				editor = newEditor;
+				model = editor.model;
+			} );
+	} );
+
+	afterEach( () => {
+		return editor.destroy().then( () => element.remove() );
+	} );
```
niegowski, line 37: I'd suggest using `async`/`await` as it's clearer.

niegowski (approved):

packages/ckeditor5-bookmark/tests/bookmarkediting.js
pszczesniak: Addressed in 2e3d86b.

PR #17055, opened 2024-09-06
Toolbar button and menu item

illia-stv:

Feature (bookmark): Register button in the toolbar and menu bar.

---

### Additional information


packages/ckeditor5-bookmark/src/bookmarkui.ts
```diff
@@ -0,0 +1,81 @@
...
+	 * @inheritDoc
+	 */
+	public init(): void {
+		const editor = this.editor;
+
+		editor.ui.componentFactory.add( 'bookmark', () => {
+			const buttonView = this._createButton( ButtonView );
+
+			buttonView.set( {
+				tooltip: true
```
pszczesniak, line 40: Lets add the `bookmarkIcon` to the button.

pszczesniak (changes requested):
Looking good :+1: but got one thing.

Could you remove all those commented code for now? I know that it will used later but for now it shades a bit everything.

Also those skipped test, please.

Lets focus on now, not the future, it will come :)

Maybe a good idea would to create a small task that will be responsible for connecting UI with editing API, when both parts will be done, WDYT?

packages/ckeditor5-ui/src/menubar/utils.ts
```diff
@@ -824,7 +825,8 @@ export const DefaultMenuBarItems: MenuBarConfigObject[ 'items' ] = [
 				items: [
 					'menuBar:link',
 					'menuBar:comment',
-					'menuBar:insertMergeField'
+					'menuBar:insertMergeField',
+					'menuBar:bookmark'
```
Witoso, line 829: Logically, this makes sense to be closer to link, as it has a similarity to it. 

Please fill in the standard GSheet with placements https://docs.google.com/spreadsheets/d/1bb3FzKZlxpT7elO_AFm_4O2eTYsmktNsWqwVVd1p5AM/edit?pli=1&gid=0#gid=0

packages/ckeditor5-bookmark/src/bookmarkui.ts
illia-stv: In that scenario the icon will not be visible in menu bar, so it will require to add icon in another place as well

pszczesniak: Correct. Bookmark icon should be also visible in menu bar.

illia-stv:
> Looking good 👍 but got one thing.
> 
> Could you remove all those commented code for now? I know that it will used later but for now it shades a bit everything.
> 
> Also those skipped test, please.
> 
> Lets focus on now, not the future, it will come :)
> 
> Maybe a good idea would to create a small task that will be responsible for connecting UI with editing API, when both parts will be done, WDYT?

Sure, issue already created🙂

packages/ckeditor5-bookmark/tests/bookmarkui.js
```diff
@@ -0,0 +1,85 @@
...
+		global.document.body.appendChild( element );
+
+		return ClassicTestEditor
+			.create( element, {
+				plugins: [ BookmarkUI ]
+			} )
+			.then( newEditor => {
+				editor = newEditor;
+			} );
+	} );
```
niegowski, line 31:
```suggestion
	beforeEach( async () => {
		element = global.document.createElement( 'div' );
		global.document.body.appendChild( element );

		editor = await ClassicTestEditor.create( element, {
			plugins: [ BookmarkUI ]
		} );
	} );
```

Issue #17052, opened 2024-09-06
Pasting an inline widget as plain text inserts additional empty lines before and after widget text


lkszzajac:
## 📝 Provide detailed reproduction steps (if any)

1. Open the inline widget implementation guide (https://ckeditor.com/docs/ckeditor5/latest/framework/tutorials/widgets/implementing-an-inline-widget.html)
2. Copy text containing the implemented placeholder inline widget from the demo editor
3. Paste it without formatting (`Cmd + Shift + V`)

### ✔️ Expected result

The inline widget gets pasted as its text representation inline

### ❌ Actual result

New empty lines are inserted before and after the inline widget pasted as plain text.

<img width="808" alt="image" src="https://github.com/user-attachments/assets/dfad579e-267e-4365-a1eb-cbe58ec62f72">

## 📃 Other details

* Browser: n/a
* OS: n/a
* First affected CKEditor version: 43.0.0
* Installed CKEditor plugins: n/a
* Seems to be caused by the `newLinePadding` in `viewtoplaintext.ts` in the Clipboard plugin https://github.com/ckeditor/ckeditor5/blob/f0422ec02eb218200472211c88d21ce9c41ff993/packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts#L57

PR #17068, opened 2024-09-10
Allow to disable new line padding on paste for specific elements

Mati365:

Feature (clipboard): Add `dataPipeline:noNewLinePaddingOnPaste` attribute that allows disabling prepending new lines around inline widget during clipboard paste pipeline. Closes https://github.com/ckeditor/ckeditor5/issues/17052

---

### DIFF
--- a/packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
+++ b/packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
@@ -87,13 +87,24 @@ function newLinePadding(
 	}
 
 	// Do not add padding around the elements that won't be rendered.
-	if (
-		element.is( 'element' ) && element.getCustomProperty( 'dataPipeline:transparentRendering' ) ||
-		previous.is( 'element' ) && previous.getCustomProperty( 'dataPipeline:transparentRendering' )
-	) {
+	if ( isNewLinePaddingDisabled( element ) || isNewLinePaddingDisabled( previous ) ) {
 		return '';
 	}
 
 	// Add empty lines between container elements.
 	return '\n\n';
+}
+
+/**
+ * Returns `true` if the new line padding should be disabled for the given element.
+ */
+function isNewLinePaddingDisabled( element: ViewElement ): boolean {
+	if ( !element.is( 'element' ) ) {
+		return false;
+	}
+
+	return !!(
+		element.getCustomProperty( 'dataPipeline:transparentRendering' ) ||
+		element.getCustomProperty( 'dataPipeline:noNewLinePaddingOnPaste' )
+	);
 }
--- a/packages/ckeditor5-clipboard/tests/utils/viewtoplaintext.js
+++ b/packages/ckeditor5-clipboard/tests/utils/viewtoplaintext.js
@@ -34,17 +34,19 @@ describe( 'viewToPlainText()', () => {
 		);
 	} );
 
-	it( 'should not put empty line before or after the element with `dataPipeline:transparentRendering` property', () => {
-		const viewString = 'Abc <container:h1>Header</container:h1> xyz';
-		const expectedText = 'Abc Header xyz';
+	for ( const property of [ 'transparentRendering', 'noNewLinePaddingOnPaste' ] ) {
+		it( `should not put empty line before or after the element with \`dataPipeline:${ property }\` property`, () => {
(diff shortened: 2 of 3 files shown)


Witoso:
Why this works on our merge fields and doesn't on custom written placeholders? Maybe we lack some abstraction?

Mati365:
@Witoso From what I can see merge tags use `dataPipeline:transparentRendering` to handle this, but in our case it's not working properly. It's not upcasting placeholders and keeps raw text after setting this attribute on `placeholder` element specified in example. 

@Dumluregn Do you know why this happens? Can we use `dataPipeline:transparentRendering` without breaking upcasting?

niegowski (changes requested):
The `newLinePadding()` helper is unfortunately based on the assumption that the view `ContainerElement` is a block element and this is not a valid assumption. Maybe we could use `DomConverter#blockElements` as a list of view elements which should be considered a block?

I'd suggest not adding another custom property.

Mati365:
@niegowski Isn't it breaking change? I'm not sure what should API look like, because we have plenty of downcasts which do not determine which view element is a block element and which not.

Mati365:
@niegowski Looks like it works 🎉

packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
```diff
@@ -19,10 +19,14 @@ const listElements = [ 'ol', 'ul' ];
...
  *
+ * @param converter The converter instance.
  * @param viewItem View item to convert.
  * @returns Plain text representation of `viewItem`.
  */
-export default function viewToPlainText( viewItem: ViewItem | ViewDocumentFragment ): string {
+export default function viewToPlainText(
+	converter: DomConverter,
+	viewItem: ViewItem | ViewDocumentFragment
+): string {
```
niegowski, line 29: This is a breaking change. Please mark it appropriately.

packages/ckeditor5-clipboard/tests/utils/viewtoplaintext.js
```diff
@@ -3,14 +3,23 @@
...
+import { DomConverter, StylesProcessor, ViewDocument } from '@ckeditor/ckeditor5-engine';
 import viewToPlainText from '../../src/utils/viewtoplaintext.js';
 
 import { parse as parseView } from '@ckeditor/ckeditor5-engine/src/dev-utils/view.js';
 
 describe( 'viewToPlainText()', () => {
+	let converter;
+
+	beforeEach( () => {
+		const viewDocument = new ViewDocument( new StylesProcessor() );
```
niegowski, line 15: `ViewDocument` should be destroyed in `afterEach`.

packages/ckeditor5-clipboard/src/clipboardpipeline.ts
```diff
@@ -323,7 +323,7 @@ export default class ClipboardPipeline extends Plugin {
 		this.listenTo<ViewDocumentClipboardOutputEvent>( viewDocument, 'clipboardOutput', ( evt, data ) => {
 			if ( !data.content.isEmpty ) {
 				data.dataTransfer.setData( 'text/html', this.editor.data.htmlProcessor.toData( data.content ) );
-				data.dataTransfer.setData( 'text/plain', viewToPlainText( data.content ) );
+				data.dataTransfer.setData( 'text/plain', viewToPlainText( editor.editing.view.domConverter, data.content ) );
```
niegowski, line 326: Since this is the clipboard pipeline, it operates more on raw data than on editing view. Also please see the above line, it uses DomConverter from data pipeline.
```suggestion
				data.dataTransfer.setData( 'text/plain', viewToPlainText( editor.data.htmlProcessor.domConverter, data.content ) );
```

niegowski (changes requested):

packages/ckeditor5-clipboard/src/utils/viewtoplaintext.ts
Mati365: Marked

niegowski (commented):
The code looks ok but the breaking change description must be updated to reflect the code change. The signature of that helper changed and since it's a public helper it could be used in some external integrations.

Witoso:
Please, let's plan this BC together for some specific release. Don't merge.

PR #17071, opened 2024-09-10
Fix MarkerOperation x MoveOperation OT for undo cases

### LINKED ISSUE #9296: Undo after splitting elements crashes the editor.
### Scenario 1: ###

1. Open any editor instance.
2. Type a few letters.
3. Set the caret between the letters and press `enter` 4 times.
4. Press `backspace` 4 times.
5. Press `Ctrl+Z` 5 times -> **The action is not executed properly**.
6. Press `Ctrl+Shift+Z` -> **The editor crashes, no action possible except of typing**.

This is the error thrown in the 6th point:
```Uncaught CKEditorError: split-operation-how-many-invalid
Read more: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-split-operation-how-many-invalid
    at l._validate (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:101751)
    at ft.on.priority (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:405830)
    at ft.fire (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:56829)
    at ft.<computed> [as applyOperation] (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:59799)
    at u._undo (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:462589)
    at Object.callback (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:463626)
    at ft._runPendingChanges (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:409575)
    at ft.enqueueChange (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:406706)
    at u.execute (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:463408)
    at u.<anonymous> (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:59748)
```

This is the error thrown while trying to do anything else with the editor:
```snippet.js:4 Uncaught CKEditorError: model-nodelist-offset-out-of-bounds {"offset":4,"nodeList":[{"data":"aaa"}]}
Read more: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-model-nodelist-offset-out-of-bounds
    at offsetToIndex (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:191156)
    at c.offsetToIndex (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:94211)
    at a (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:29321)
    at r.get textNode [as textNode] (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:23466)
    at W (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:390188)
    at H._validateSelectionRange (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:390027)
    at m.<anonymous> (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:135889)
    at m.fire (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:56829)
    at m._setRanges (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:193998)
    at m.setTo (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:193027)
```

### Scenario 2: ###

1. Open any editor instance.
2. Create content with a `header` and `paragraph`.
3. Set the caret at the beginning of the `header` and press `enter` 2 times, then `backspace` 2 times.
4. Set the caret between the letters of the header and press `enter` 2 times, then `backspace` 2 times.
5. Keep pressing `Ctrl-Z` until you get a console error and the editor crashes.

This is the error thrown in the 5th point:
```snippet.js:4 Uncaught CKEditorError: rename-operation-wrong-name
Read more: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-rename-operation-wrong-name
    at a._validate (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:174102)
    at ft.on.priority (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:405830)
    at ft.fire (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:56829)
    at ft.<computed> [as applyOperation] (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:59799)
    at l._undo (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:462589)
    at Object.callback (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:463078)
    at ft._runPendingChanges (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:409575)
    at ft.enqueueChange (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:406706)
    at l.execute (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:463052)
    at l.<anonymous> (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:59748)
```

This is the error thrown while trying to do anything else with the editor:
```snippet.js:4 Uncaught CKEditorError: model-nodelist-offset-out-of-bounds {"offset":4,"nodeList":[{"data":"AAA"}]}
Read more: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-model-nodelist-offset-out-of-bounds
    at offsetToIndex (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:191156)
    at c.offsetToIndex (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:94211)
    at a (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:29321)
    at r.get textNode [as textNode] (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:23466)
    at W (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:390188)
    at H._validateSelectionRange (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:390027)
    at m.<anonymous> (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:135889)
    at m.fire (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:56829)
    at m._setRanges (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:193998)
    at m.setTo (https://ckeditor.com/docs/ckeditor5/latest/snippets/build-classic-source/snippet.js:4:193027)
```

---

scofalik:

Fix (engine): Fixed incorrect marker handling in some scenarios involving undo and real-time collaboration, which earlier lead to `model-nodelist-offset-out-of-bounds` error.

Fix (engine): Fixed incorrect handling of merge changes during undo in some scenarios involving real-time collaboration, which earlier led to `model-nodelist-offset-out-of-bounds` error.

Fix (engine): Fixed conflict resolution error, which led to editor crash in some scenarios where two users removed bigger intersecting part of the content and used undo.

Fix (engine): Fixed incorrect undo behavior leading to an editor crash when a user pressed enter key multiple times, then pressed backspace that many times, then undone all the changes. Closes #9296.

### DIFF
--- a/packages/ckeditor5-engine/src/model/operation/transform.ts
+++ b/packages/ckeditor5-engine/src/model/operation/transform.ts
@@ -373,6 +373,9 @@ export function transformSets(
 		operationsB.splice( indexB, 1, ...newOpsB );
 	}
 
+	handlePartialMarkerOperations( operationsA );
+	handlePartialMarkerOperations( operationsB );
+
 	if ( options.padWithNoOps ) {
 		// If no-operations padding is enabled, count how many extra `a` and `b` operations were generated.
 		const brokenOperationsACount = operationsA.length - data.originalOperationsACount;
@@ -524,6 +527,9 @@ class ContextFactory {
 					const range = Range._createFromPositionAndShift( opB.sourcePosition, opB.howMany );
 
 					if ( opA.splitPosition.hasSameParentAs( opB.sourcePosition ) && range.containsPosition( opA.splitPosition ) ) {
+						// TODO: Potential bug -- we are saving offset value directly and it is not later updated during OT.
+						// TODO: This may cause a bug it here was an non-undone operation that may have impacted this offset.
+						// TODO: Similar error was with MarkerOperation relations, where full path was saved and never updated.
 						const howMany = range.end.offset - opA.splitPosition.offset;
 						const offset = opA.splitPosition.offset - range.start.offset;
 
@@ -564,22 +570,7 @@ class ContextFactory {
 				return;
 			}
 
-			if ( opB instanceof MoveOperation ) {
-				const movedRange = Range._createFromPositionAndShift( opB.sourcePosition, opB.howMany );
-
-				const affectedLeft = movedRange.containsPosition( markerRange.start ) ||
-					movedRange.start.isEqual( markerRange.start );
-
-				const affectedRight = movedRange.containsPosition( markerRange.end ) ||
-					movedRange.end.isEqual( markerRange.end );
-
-				if ( ( affectedLeft || affectedRight ) && !movedRange.containsRange( markerRange ) ) {
-					this._setRelation( opA, opB, {
-						side: affectedLeft ? 'left' : 'right',
-						path: affectedLeft ? markerRange.start.path.slice() : markerRange.end.path.slice()
-					} );
-				}
-			} else if ( opB instanceof MergeOperation ) {
+			if ( opB instanceof MergeOperation ) {
 				const wasInLeftElement = markerRange.start.isEqual( opB.targetPosition );
 				const wasStartBeforeMergedElement = markerRange.start.isEqual( opB.deletionPosition );
 				const wasEndBeforeMergedElement = markerRange.end.isEqual( opB.deletionPosition );
@@ -745,6 +736,62 @@ function updateBaseVersions( operations: ReadonlyArray<Operation>, baseVersion: 
 function padWithNoOps( operations: Array<Operation>, howMany: number ) {
 	for ( let i = 0; i < howMany; i++ ) {
 		operations.push( new NoOperation( 0 ) );
+	}
+}
+
+/**
+ * Transformed operations set may include marker operations which were broken into multiple marker operations during transformation.
+ * It represents marker range being broken into multiple pieces as the transformation was processed. Each partial marker operation is
+ * a piece of the original marker range.
+ *
+ * These partial marker operations ("marker range pieces") should be "glued" together if, after transformations, the ranges ended up
+ * next to each other.
+ *
+ * If the ranges did not end up next to each other, then partial marker operations should be discarded, as the marker range cannot
+ * be broken into two pieces.
+ *
+ * There is always one "reference" marker operation (the original operation) and there may be some partial marker operations. Partial
+ * marker operations have base version set to `-1`. If the `operations` set includes partial marker operations, then they are always
+ * after the original marker operation.
+ *
+ * See also `MarkerOperation` x `MoveOperation` transformation.
+ *
+ * @param operations
+ */
+function handlePartialMarkerOperations( operations: Array<Operation> ) {
+	const markerOps: Map<string, { op: MarkerOperation, ranges: Array<Range> }> = new Map();
+
+	for ( let i = 0; i < operations.length; i++ ) {
+		const op = operations[ i ];
+
+		if ( !( op instanceof MarkerOperation ) ) {
+			continue;
+		}
+
+		if ( op.baseVersion !== -1 ) {
+			markerOps.set( op.name, {
+				op,
+				ranges: op.newRange ? [ op.newRange ] : []
+			} );
+		} else {
+			if ( op.newRange ) {
+				// `markerOps.get( op.name )` must exist because original marker operation is always before partial marker operations.
+				// If the original marker operation was changed to `NoOperation`, then the partial marker operations would be changed
+				// to `NoOperation` as well, so this is not a case.
+				markerOps.get( op.name )!.ranges.push( op.newRange );
+			}
+
+			operations.splice( i, 1 );
+			i--;
+		}
+	}
+
+	for ( const { op, ranges } of markerOps.values() ) {
+		if ( ranges.length ) {
+			op.newRange = Range._createFromRanges( ranges );
+		} else {
+			op.newRange = null;
+		}
 	}
 }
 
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-engine/src/model/range.ts
```diff
@@ -975,21 +975,18 @@ export default class Range extends TypeCheckable implements Iterable<TreeWalkerV
...
-		if ( refIndex > 0 ) {
-			// eslint-disable-next-line no-constant-condition
-			for ( let i = refIndex - 1; true; i++ ) {
-				if ( ranges[ i ].end.isEqual( result.start ) ) {
-					( result as any ).start = Position._createAt( ranges[ i ].start );
-				} else {
-					// If ranges are not starting/ending at the same position there is no point in looking further.
-					break;
-				}
+		for ( let i = refIndex - 1; i >= 0; i-- ) {
```
scofalik, line 984: There was a bug here, we go "left" in the array, so we should decrement `i`.

scofalik:
I've wrapped unit tests from `transform/undo.js` into a separate `describe()` to show them better in unit tests logs but because of that the changes in that file are hard to digest. I added a unit test for each fix, four total. The unit tests are based on the scenarios from their related issues.

scofalik:
This PR includes four fixes. I already described the first one. Although that fix is complex and wide, I have very strong confidence that it is correct. There are three more fixes.

Fix (2): c5a614e

This fix makes it so that `MergeOperation` becomes `NoOperation` more often than before. Earlier, `!context.forceWeakRemove` had to be true (`forceWeakRemove` = undo mode). This, in essence, means that merge became `NoOperation` only outside undo. During undo we wanted to keep merge operation and keep transforming it, but this caused errors down the road. Now, the merge operation will become `NoOperation` also during undo.

The side effect of this change is that sometimes the content may not be recovered exactly as we expected -- i.e. two elements will be removed instead of one, or we would expect an "enter" between some text (so, we get `<p>Foobar</p>` instead of `<p>Foo</p><p>bar</p>`).

I have lower confidence towards this fix. It fixes one of the issues, and it also fixes an issue found by QA team when testing fix (1). Also, all unit tests were passing. But I assume it was done like this for a reason and we might not know about some scenario, where it made sense. But it may be only to provide better UX during undo (as described above).

Also, there is a "mirror case" in `MoveOperation` x `MergeOperation` transformation, and that case still uses `!context.forceWeakRemove` in its condition, which seems to lead to disparity and may be a problem. I want to investigate it a little more.

Fix (3): 4a9d71d

This fix is more straightforward as it appeared that earlier solution lacked transformation of one of the `MoveOperation` properties. I have pretty strong confidence in this kind of fixes. It is an omission fix. It seems that we forgot/missed that it should be transformed too. In contrary, in fix (2), we are changing rules when a transformation is applied. I know that maybe the difference is not tangible to someone who does not know this system, but, again, I feel quite confident with this change.

Before this change, the `MoveOperation` was incorrectly transformed, and it brought back incorrect nodes from graveyard. As a result, after undo, we would be ending up with something like this in model: `<p></p><p>XYZ</p>C` instead of, e.g. `<p>ABC</p><p>XYZ</p>`.

Fix (4): 471390b

This causes a special case in `MergeOperation` x `SplitOperation` to be applied less often. Normally, merge operation targets the end of an element, e.g. you have `<p>Foo</p><p>Bar</p>` and merge operation source is `[ 1, 0 ]` and target is `[ 0, 3 ]`.

During OT/undo when some nodes might have been removed, or splits may have happened, sometimes merge operation starts to target into inside of the element. This is a temporary situation for the merge operation and as the merge operation is further transformed, the target position should eventually end up pointing to an end of target element.

The special case was applied always when the merge operation was targeting into inside of an element instead of its end.

It appears that this assumption was rather not correct, because in the case I wanted to fix, it caused an error later during other transformations.

As you can see, I removed `mergeInside` from the `if`. Most importantly, in almost all unit tests (~10?) `mergeInside` was `true` only when the other condition (`context.abRelation == 'mergeTargetNotMoved'`) was `true` as well (it's something I'd expect). We had only one unit test, where `mergeInside` was `true` and the other was `false`.

That unit test was still green, even after removing `mergeInside`, which means that it was indifferent to whether a special transformation case was applied or not. That merge operation eventually ended up in the same state, regardless if at that concrete step, we took special case or not. I followed this transformation and I concluded that the special transformation case made the merge operation into "final" state and the merge operation was not affected by the rest of operations. I concluded that this was a bit lucky. After removing `mergeInside`, the merge operation did not receive special case and was affected by following operations. The way how merge operation interacted with other operations made sense to me and was more in line with what I expected. The final result was the same, but the merge operation correctly interacted (as it should) with other operations during transformation process. So, I believe, that in this one case that we earlier had, where `mergeInside` made difference, it was also correct to not apply the special case.

This gives me a bit more confidence towards this fix, but it is similar to fix (2) where we change rules _when_ we apply a fix. I still have more confidence than in fix (2) but less than for fix (1) or (3).

The unit test I described above is `pasting on non-collapsed selection undo and redo` from `transform/undo.js`.

packages/ckeditor5-engine/src/model/operation/transform.ts
```diff
@@ -748,6 +739,63 @@ function padWithNoOps( operations: Array<Operation>, howMany: number ) {
...
+ * be broken into two pieces.
+ *
+ * There is always one "reference" marker operation (the original operation) and there may be some partial marker operations. Partial
+ * marker operations have base version set to `-1`. If the `operations` set includes partial marker operations, then they are always
+ * after the original marker operation.
+ *
+ * See also `MarkerOperation` x `MoveOperation` transformation.
+ * See also https://github.com/ckeditor/ckeditor5/pull/17071.
+ *
+ * @param operations
```
niegowski, line 760:
```suggestion
```

niegowski (approved):

oleq (approved):

PR #17073, opened 2024-09-11
[Bookmark] Data upcast.

pszczesniak:

Internal (bookmark): Data upcast. Closes #000.

---

### DIFF
--- a/packages/ckeditor5-bookmark/src/bookmarkediting.ts
+++ b/packages/ckeditor5-bookmark/src/bookmarkediting.ts
@@ -10,7 +10,7 @@
 import { Plugin } from 'ckeditor5/src/core.js';
 import { toWidget } from 'ckeditor5/src/widget.js';
 import { IconView } from 'ckeditor5/src/ui.js';
-import type { ViewUIElement, DowncastWriter } from 'ckeditor5/src/engine.js';
+import { type ViewUIElement, type DowncastWriter, type ViewElement } from 'ckeditor5/src/engine.js';
 
 import bookmarkIcon from '../theme/icons/bookmark.svg';
 
@@ -53,6 +53,9 @@ export default class BookmarkEditing extends Plugin {
 		const { editor } = this;
 		const { conversion, t } = editor;
 
+		editor.data.htmlProcessor.domConverter.registerInlineObjectMatcher( element => this.upcastMatcher( element ) );
+		editor.editing.view.domConverter.registerInlineObjectMatcher( element => this.upcastMatcher( element ) );
+
 		conversion.for( 'dataDowncast' ).elementToElement( {
 			model: {
 				name: 'bookmark',
@@ -90,23 +93,7 @@ export default class BookmarkEditing extends Plugin {
 		} );
 
 		conversion.for( 'upcast' ).elementToElement( {
-			view: element => {
-				const isAnchorElement = element.name === 'a';
-
-				if ( !isAnchorElement ) {
-					return null;
-				}
-
-				const hasIdAttribute = element.hasAttribute( 'id' );
-				const hasHrefAttribute = element.hasAttribute( 'href' );
-				const isEmpty = element.isEmpty;
-
-				if ( !hasIdAttribute || hasHrefAttribute || !isEmpty ) {
-					return null;
-				}
-
-				return { name: true };
-			},
+			view: element => this.upcastMatcher( element ),
 			model: ( viewElement, { writer } ) => {
 				const bookmarkId = viewElement.getAttribute( 'id' );
 
@@ -135,5 +122,23 @@ export default class BookmarkEditing extends Plugin {
 
 			return domElement;
 		} );
+	}
+
+	private upcastMatcher( element: ViewElement ) {
+		const isAnchorElement = element.name === 'a';
+
+		if ( !isAnchorElement ) {
+			return null;
+		}
+
+		const hasIdAttribute = element.hasAttribute( 'id' );
+		const hasHrefAttribute = element.hasAttribute( 'href' );
+		const isEmpty = element.isEmpty;
+
+		if ( !hasIdAttribute || hasHrefAttribute || !isEmpty ) {
+			return null;
+		}
+
+		return { name: true, attributes: [ 'id' ] };
 	}
 }
--- a/packages/ckeditor5-bookmark/src/bookmarkediting.ts
+++ b/packages/ckeditor5-bookmark/src/bookmarkediting.ts
@@ -10,7 +10,7 @@
 import { Plugin } from 'ckeditor5/src/core.js';
 import { toWidget } from 'ckeditor5/src/widget.js';
 import { IconView } from 'ckeditor5/src/ui.js';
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-bookmark/tests/bookmarkediting.js
```diff
@@ -174,6 +224,19 @@ describe( 'BookmarkEditing', () => {
...
 
+		it( 'should properly downcast bookmark with text with space after', () => {
+			setModelData( model, '<paragraph><bookmark bookmarkId="foo"></bookmark> text</paragraph>' );
+
+			expect( getViewData( view, { withoutSelection: true } ) ).to.equal(
+				'<p>' +
+					'<a class="ck-bookmark ck-widget" contenteditable="false" id="foo">' +
+						'<span class="ck-bookmark__icon"></span>' +
+					'</a>' +
+					'text' +
```
niegowski, line 235: The space must be kept here also.

niegowski (changes requested):

packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -111,4 +123,22 @@ export default class BookmarkEditing extends Plugin {
 			return domElement;
 		} );
 	}
+
+	private upcastMatcher( element: ViewElement ) {
```
niegowski, line 127: This does not need to be exposed as a plugin public API. As it's using only parameters without side-effects we could use a plain helper function here (instead of method). That helper could be even not exported.

packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -53,6 +53,9 @@ export default class BookmarkEditing extends Plugin {
 		const { editor } = this;
 		const { conversion, t } = editor;
 
+		editor.data.htmlProcessor.domConverter.registerInlineObjectMatcher( element => this.upcastMatcher( element ) );
+		editor.editing.view.domConverter.registerInlineObjectMatcher( element => this.upcastMatcher( element ) );
```
niegowski, line 57: I'd add some code comments describing the reasons for this.

packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -10,7 +10,7 @@
 import { Plugin } from 'ckeditor5/src/core.js';
 import { toWidget } from 'ckeditor5/src/widget.js';
 import { IconView } from 'ckeditor5/src/ui.js';
-import type { ViewUIElement, DowncastWriter } from 'ckeditor5/src/engine.js';
+import { type ViewUIElement, type DowncastWriter, type ViewElement } from 'ckeditor5/src/engine.js';
```
niegowski, line 13:
```suggestion
import type { ViewUIElement, DowncastWriter, ViewElement } from 'ckeditor5/src/engine.js';
```

packages/ckeditor5-bookmark/tests/bookmarkediting.js
```diff
@@ -81,6 +83,17 @@ describe( 'BookmarkEditing', () => {
 			);
 		} );
 
+		it( 'should properly downcast bookmark with and space text before', () => {
+			setModelData( model, '<paragraph>text <bookmark bookmarkId="foo"></bookmark></paragraph>' );
+
+			expect( editor.getData() ).to.equal(
+				'<p>' +
+					'text' +
```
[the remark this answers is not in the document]
pszczesniak: As we debugged it together, the issue relies inside `SetModelData` function, which strips the spaces while setting the data.

For now the workaround is instead of:

```js
setModelData( model, '<paragraph>text <bookmark bookmarkId="foo"></bookmark></paragraph>' );
```

is to put the selectrion marker after the space

```js
setModelData( model, '<paragraph>text []<bookmark bookmarkId="foo"></bookmark></paragraph>' );
```

packages/ckeditor5-bookmark/tests/bookmarkediting.js
pszczesniak: https://github.com/ckeditor/ckeditor5/pull/17073#discussion_r1756213693

packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -88,6 +94,15 @@ export default class BookmarkEditing extends Plugin {
 				return toWidget( containerElement, writer, { label: labelCreator } );
 			}
 		} );
+
+		conversion.for( 'upcast' ).elementToElement( {
+			view: element => upcastMatcher( element ),
+			model: ( viewElement, { writer } ) => {
+				const bookmarkId = viewElement.getAttribute( 'id' );
```
illia-stv, line 101: what about `name`, or we will manage that in next iteration?

packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -112,3 +127,25 @@ export default class BookmarkEditing extends Plugin {
 		} );
 	}
 }
+
+/**
+ * A helper function to match an `anchor` element which must contain `id` attribute but without `href` attribute,
+ * also element must be empty when matcher is execute in data pipeline so it can be upcasted to a `bookmark`.
+ */
+function upcastMatcher( element: ViewElement, dataPipeline: boolean = true ) {
```
illia-stv, line 135: Why do we need `dataPipeline` argument, maybe it could goud idea do describe it in comment as @param?

packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -112,3 +127,25 @@ export default class BookmarkEditing extends Plugin {
...
+
+	if ( !isAnchorElement ) {
+		return null;
+	}
+
+	const hasIdAttribute = element.hasAttribute( 'id' );
+	const hasHrefAttribute = element.hasAttribute( 'href' );
+	const isEmpty = element.isEmpty;
+
+	if ( !hasIdAttribute || hasHrefAttribute || dataPipeline && !isEmpty ) {
```
illia-stv, line 146: So if I understand it correctly in editingDowncast `dataPipeline` is set on `false` (L60) so `anchor` element could be not empty?

packages/ckeditor5-bookmark/src/bookmarkediting.ts
pszczesniak: Not in this iteration.

packages/ckeditor5-bookmark/src/bookmarkediting.ts
pszczesniak: It's not empty cause in editing we will put there an `UIElement` width bookmark icon.

packages/ckeditor5-bookmark/src/bookmarkediting.ts
pszczesniak: Addressed in 3f7a20f.

illia-stv (approved):

PR #17077, opened 2024-09-11
[Bookmark] Balloon form (insert)

illia-stv:

Feature (bookmark): Implement balloon form for bookmarks.

---

### Additional information


pszczesniak (changes requested):
Partial review, few things for now to fix.

packages/ckeditor5-bookmark/src/ui/bookmarkformview.ts
```diff
@@ -0,0 +1,284 @@
...
+	 * @param validators  Form validators used by {@link #isValid}.
+	 */
+	constructor( locale: Locale, validators: Array<BookmarkFormValidatorCallback> ) {
+		super( locale );
+
+		const t = locale.t;
+
+		this._validators = validators;
+		this.idInputView = this._createIdInput();
+		this.insertButtonView = this._createButton( t( 'Insert' ), 'ck-button-insert', 'submit' );
```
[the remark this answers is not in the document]
illia-stv: According to the design it supposed to be "Insert" as I see. So is it should look like this?
image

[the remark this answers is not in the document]
pszczesniak: Correct, my bad. Please make the button text be `'Insert'` as you did previously.

packages/ckeditor5-bookmark/theme/bookmark.css
```diff
@@ -9,6 +9,10 @@
 	--ck-bookmark-icon-selected-fill-color: var(--ck-color-focus-border);
 	--ck-bookmark-icon-animation-duration: var(--ck-widget-handler-animation-duration);
 	--ck-bookmark-icon-animation-curve: var(--ck-widget-handler-animation-curve);
+	--ck-color-insert-button-background: var(--ck-color-base-action);
+	--ck-color-insert-button-hover-background: hsl(104, 53%, 40%);
```
pszczesniak, line 13: Based on comment in `ui/bookmarkformview.ts` in L93 those variables aren't needed.

packages/ckeditor5-bookmark/theme/bookmark.css
```diff
@@ -9,6 +9,10 @@
...
+	--ck-color-insert-button-active-background: hsl(104, 53%, 40%);
+	--ck-color-insert-button-disabled-background: hsl(104, 44%, 58%);
```
pszczesniak, line 15: Are those variables used somewhere?? :thinking:

packages/ckeditor5-bookmark/theme/bookmark.css
```diff
@@ -37,3 +41,17 @@
 		}
 	}
 }
+
+.ck-bookmark-form {
```
pszczesniak, line 45: Maybe would be good to have it in separate css file?

packages/ckeditor5-bookmark/src/ui/bookmarkview.ts
```diff
@@ -0,0 +1,95 @@
...
+		this.children = this.createCollection();
+
+		this.formView = new ( CssTransitionDisablerMixin( BookmarkFormView ) )( locale, validators );
+		this._formHeader = new FormHeaderView( locale, {
+			label: this.t!( 'Bookmark' )
+		} );
+
+		const classList = [ 'ck', 'ck-bookmark-view' ];
+
+		this.children.add( this._formHeader );
```
pszczesniak, line 65: Maybe we could we drop the private `_formHeader` property and just add it to the children collection directly?

```suggestion
		this.children.add( new FormHeaderView( locale, {
			label: this.t!( 'Bookmark' )
		} ) );
```

I've seen that this approach is mostly used in your codebase. For example: https://github.com/ckeditor/ckeditor5/blob/23a35bd30914544152c39ef6f6459e3eccedc2fe/packages/ckeditor5-table/src/tableproperties/ui/tablepropertiesview.ts#L268-L271

packages/ckeditor5-bookmark/src/ui/bookmarkview.ts
```diff
@@ -0,0 +1,95 @@
...
+	 * @param validators Form validators used by {@link #isValid}.
+	 */
+	constructor( locale: Locale, validators: Array<BookmarkFormValidatorCallback> ) {
+		super( locale );
+
+		this.children = this.createCollection();
+
+		this.formView = new ( CssTransitionDisablerMixin( BookmarkFormView ) )( locale, validators );
+		this._formHeader = new FormHeaderView( locale, {
+			label: this.t!( 'Bookmark' )
```
pszczesniak, line 60: This translations won't be ctched by translation util. It only searches for `t()`, forms like `this.editor.t` or `this.t` wont be founded, unless you are 100% sure that this translation will be served by different package.

We also need to remember that we can't duplicate keys in `contexts.json`.

I suggest to:

```
const { t } = this;

// ...

label: t( 'Bookmark' )
```

And extend the description in `contexts.json` for `Bookmark` key - similar like here :point\_right: https://github.com/ckeditor/ckeditor5/blob/23a35bd30914544152c39ef6f6459e3eccedc2fe/packages/ckeditor5-find-and-replace/lang/contexts.json#L2

pszczesniak (changes requested):
2nd part of review.

packages/ckeditor5-bookmark/theme/bookmark.css
illia-stv: Not yet, I thought it could be used in near future. But yeah maybe better to let it go.

packages/ckeditor5-bookmark/src/ui/bookmarkview.ts
illia-stv: I agree it make sense, but for testing purpose it is much more convenient to have reference for that view component, otherwise we need to provide `locale` and `t()` so it requires creating the instance of editor.

niegowski (changes requested):

packages/ckeditor5-bookmark/theme/bookmark.css
```diff
@@ -35,5 +35,11 @@
 		&.ck-widget_selected:hover {
 			outline: none;
 		}
+
+		& .ck-bookmark__icon {
+			& .ck-icon {
+				width: 1em;
```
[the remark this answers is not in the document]
illia-stv: We need to define font size of the icon, but it should also look good in the Headings.

With that style rule:
image

Without that style rule:
image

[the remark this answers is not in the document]
niegowski: I feel solution proposed by @pszczesniak is better since the em unit is related to the text height, not width.

packages/ckeditor5-bookmark/src/bookmarkui.ts
```diff
@@ -288,7 +288,7 @@ export default class BookmarkUI extends Plugin {
 	 *
 	 * @internal
```
niegowski, line 289: This is not needed for private members.

packages/ckeditor5-bookmark/src/bookmarkui.ts
```diff
@@ -27,6 +39,23 @@ export default class BookmarkUI extends Plugin {
...
+	 * @inheritDoc
+	 */
+	public static get requires() {
+		return [ ContextualBalloon ] as const;
+	}
+
+	/**
+	 * The form view displayed inside the balloon.
+	 */
+	public formView: BookmarkFormView & ViewWithCssTransitionDisabler | null = null;
```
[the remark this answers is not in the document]
niegowski: The destroy is still missing in the `BookmarkUI` plugin.

niegowski (changes requested):

packages/ckeditor5-bookmark/src/ui/bookmarkformview.ts
```diff
@@ -0,0 +1,332 @@
...
+		const t = this.t!;
+
+		children.add( new FormHeaderView( this.locale, {
+			label: t( 'Bookmark' )
+		} ) );
+		children.add( this._formView );
+
+		return children;
+	}
+
```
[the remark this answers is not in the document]
illia-stv: Oh sory, I was looking at wrong file, yes you right.

packages/ckeditor5-bookmark/theme/bookmark.css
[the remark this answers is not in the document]
pszczesniak: BTW, when we finish this iteration we need a review from @dwandeljanicka. Maybe some sizes would needs to be updated.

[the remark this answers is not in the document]
illia-stv: I agree, I am just worried about the value of `em`, do you think 0.75 is not too small? I feel like it harder to hit into the icon because of its small size.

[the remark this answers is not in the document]
pszczesniak: Please, suggest the size that would suite best from your point of view, TBH don't know, need to run manual test and check it out.

packages/ckeditor5-bookmark/lang/contexts.json
```diff
@@ -1,4 +1,8 @@
 {
-	"Bookmark": "The label of the bookmark toolbar button.",
+	"Bookmark": "The label of the bookmark toolbar button. Also, a bookmark form header.",
+	"Insert": "The button in the bookmark form which inserts bookmark before selection.",
+	"Add bookmark name": "The label of input where name of the bookmark should be provided.",
```
niegowski, line 4:
```suggestion
	"Add bookmark name": "The label of the input in the bookmark insert form.",
```

packages/ckeditor5-bookmark/lang/contexts.json
```diff
@@ -1,4 +1,8 @@
...
+	"Enter the bookmark name without spaces.": "The description of bookmark input.",
```
niegowski, line 5:
```suggestion
	"Enter the bookmark name without spaces.": "The description of bookmark input in the bookmark insert form.",
```

packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -13,8 +13,7 @@ import { IconView } from 'ckeditor5/src/ui.js';
 import type { ViewUIElement, DowncastWriter, ViewElement } from 'ckeditor5/src/engine.js';
 
 import UpdateBookmarkCommand from './updatebookmarkcommand.js';
-
-import bookmarkIcon from '../theme/icons/bookmark.svg';
+import bookmarkIcon from '../theme/icons/bookmark_inline.svg';
```
niegowski, line 16:
```suggestion

import bookmarkIcon from '../theme/icons/bookmark_inline.svg';
```

packages/ckeditor5-theme-lark/theme/index.css
```diff
@@ -55,6 +55,7 @@
 @import "./ckeditor5-editor-classic/classiceditor.css";
 
 /* Plugins */
+@import "./ckeditor5-bookmark/bookmark.css";
```
pszczesniak, line 58:
```suggestion
```

pszczesniak (approved):

PR #17095, opened 2024-09-13
[Bookmark] Update command.

pszczesniak:

Internal (bookmark): Update command. Closes #000.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-bookmark/src/bookmarkediting.ts
+++ b/packages/ckeditor5-bookmark/src/bookmarkediting.ts
@@ -12,6 +12,8 @@ import { toWidget } from 'ckeditor5/src/widget.js';
 import { IconView } from 'ckeditor5/src/ui.js';
 import type { ViewUIElement, DowncastWriter, ViewElement } from 'ckeditor5/src/engine.js';
 
+import EditBookmarkCommand from './editbookmarkcommand.js';
+
 import bookmarkIcon from '../theme/icons/bookmark.svg';
 
 /**
@@ -29,8 +31,12 @@ export default class BookmarkEditing extends Plugin {
 	 * @inheritDoc
 	 */
 	public init(): void {
+		const { editor } = this;
+
 		this._defineSchema();
 		this._defineConverters();
+
+		editor.commands.add( 'editBookmark', new EditBookmarkCommand( editor ) );
 	}
 
 	/**
--- a/packages/ckeditor5-bookmark/src/bookmarkediting.ts
+++ b/packages/ckeditor5-bookmark/src/bookmarkediting.ts
@@ -50,16 +50,6 @@ export default class BookmarkEditing extends Plugin {
 			allowAttributes: 'bookmarkId',
 			disallowAttributes: 'linkHref'
 		} );
-
-		// Disallow formatting attributes on `bookmark` element.
-		schema.addAttributeCheck( ( context, attributeName ) => {
-			const bookmark = context.last;
-			const isFormatting = schema.getAttributeProperties( attributeName ).isFormatting;
-
-			if ( isFormatting && bookmark && bookmark.name == 'bookmark' ) {
-				return false;
-			}
-		} );
 	}
 
 	/**
--- a/packages/ckeditor5-bookmark/src/bookmarkediting.ts
+++ b/packages/ckeditor5-bookmark/src/bookmarkediting.ts
@@ -12,7 +12,7 @@ import { toWidget } from 'ckeditor5/src/widget.js';
 import { IconView } from 'ckeditor5/src/ui.js';
 import type { ViewUIElement, DowncastWriter, ViewElement } from 'ckeditor5/src/engine.js';
 
-import EditBookmarkCommand from './editbookmarkcommand.js';
+import UpdateBookmarkCommand from './updatebookmarkcommand.js';
 
 import bookmarkIcon from '../theme/icons/bookmark.svg';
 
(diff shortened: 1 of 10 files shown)


packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -44,6 +50,16 @@ export default class BookmarkEditing extends Plugin {
 			allowAttributes: 'bookmarkId',
 			disallowAttributes: 'linkHref'
 		} );
+
+		// Disallow formatting attributes on `bookmark` element.
+		schema.addAttributeCheck( ( context, attributeName ) => {
```
illia-stv, line 55: Is this somehow related to update command, I feel like it more belongs to schema definition PR, but do we forget about it earlier or is it indeed a part of update command?

illia-stv:
I would create a separate folder dedicated for commands, because we will have a few more.

packages/ckeditor5-bookmark/src/editbookmarkcommand.ts
```diff
@@ -0,0 +1,77 @@
...
+
+	/**
+	 * @inheritDoc
+	 */
+	public override refresh(): void {
+		const model = this.editor.model;
+		const selection = model.document.selection;
+
+		this.isEnabled = !!getSelectedBookmark( selection );
+		this.value = this.isEnabled ? getSelectedBookmark( selection )!.getAttribute( 'bookmarkId' ) as string : undefined;
```
illia-stv, line 41:
```suggestion
		const selectedBookmark = getSelectedBookmark( selection );

		this.isEnabled = !!selectedBookmark;
		this.value = selectedBookmark ? selectedBookmark.getAttribute( 'bookmarkId' ) as string : undefined;
```

I would recommend creating a variable to store the value, so there’s no need to call the function twice. This would also allow us to get rid of the `!` as well.

illia-stv (approved):
Generally looks very nice, just added few suggestions✅

pszczesniak:
> I would create a separate folder dedicated for commands, because we will have a few more.

There will be two just commands: insert and edit, the delete command will be managed by `Delete` plugin, from `ckeditor5-typing` package.

packages/ckeditor5-bookmark/src/bookmarkediting.ts
pszczesniak: Yes, and now im wondering if better would be to put it into `insert command` PR - cause tests also would suit there better. Yes, will do it this way.

packages/ckeditor5-bookmark/tests/editbookmarkcommand.js
```diff
@@ -0,0 +1,110 @@
...
+	describe( '#isEnabled()', () => {
+		describe( 'should be `true`', () => {
+			it( 'when only bookmark element is selected', () => {
+				setModelData( model, '<paragraph>[<bookmark bookmarkId="foo"></bookmark>]</paragraph>' );
+
+				expect( command.isEnabled ).to.equal( true );
+			} );
+		} );
+
+		describe( 'should be `false`', () => {
```
niegowski, line 54: Maybe some additional tests of some other object selected could be useful here.

packages/ckeditor5-bookmark/src/bookmarkediting.ts
```diff
@@ -29,8 +31,12 @@ export default class BookmarkEditing extends Plugin {
 	 * @inheritDoc
 	 */
 	public init(): void {
+		const { editor } = this;
+
 		this._defineSchema();
 		this._defineConverters();
+
+		editor.commands.add( 'editBookmark', new EditBookmarkCommand( editor ) );
```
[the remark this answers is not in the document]
niegowski: Also, the editing tests should verify if this command is registered in the collection of the commands.

niegowski (changes requested):

packages/ckeditor5-bookmark/src/bookmarkediting.ts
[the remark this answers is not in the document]
pszczesniak: Yes, don't know why i named this `edit` instead of `update` - everywhere it's mention as an update.

packages/ckeditor5-bookmark/tests/editbookmarkcommand.js
pszczesniak: Added `Link` and `imageInline`.

packages/ckeditor5-bookmark/src/updatebookmarkcommand.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * @module bookmark/updatebookmarkcommand
+ */
+
+import { Command } from 'ckeditor5/src/core.js';
+import type { Selection, DocumentSelection } from 'ckeditor5/src/engine.js';
+
+/**
+ * The edit bookmark command.
```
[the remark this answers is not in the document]
pszczesniak: Addressed in f983c33.

packages/ckeditor5-bookmark/src/updatebookmarkcommand.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * The command is registered by {@link module:bookmark/bookmarkediting~BookmarkEditing} as `'updateBookmark'`.
+ *
+ * To update the bookmarkId of current selected bookmark element, execute the command passing the bookmark id as a parameter:
+ *
+ * ```ts
+ * editor.execute( 'updateBookmark', 'newId' );
```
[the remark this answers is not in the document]
pszczesniak: Addressed in f983c33.

niegowski (approved):

Issue #17069, opened 2024-09-10
Font color picker not working properly on first change


jessna96:
## 📝 Provide detailed reproduction steps (if any)

1. Go to https://ckeditor.com/docs/ckeditor5/latest/features/font.html#demo
2. Select any text in the editor (for example "message").
3. Click on the font color icon in the toolbar and select the "color picker".
4. Click in the right upper corner of the color gradient field.
5. Then select any other color in the gradient field.
6. Save the font color with the green check.

### ✔️ Expected result

The color selected the second time (step 5) is applied to the selected text.

### ❌ Actual result

The first color is applied to the selected text (the one selected in step 4).

## 📃 Other details

This only happens the very first time the font color picker is used after loading the page (going into editor mode of the ckEditor).
Further it only happens in Chrome and Edge. 

* Browser: Chrome
* OS: macOS Sonoma 14.6.1
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Witoso:
@ckeditor/qa-team would you mind verifying if this is a recent regression?

PR #17109, opened 2024-09-17
Fix font color picker not working properly on first change on Chrome

Mati365:

Fix (ui): Fixed an issue where the first selected color was applied instead of the second selected color when using the font color picker for the first time after loading the page. Closes https://github.com/ckeditor/ckeditor5/issues/17069

---

#### Before

https://github.com/user-attachments/assets/7b69a7dc-b0cd-4f3e-ba10-1298bea343a0

#### After

https://github.com/user-attachments/assets/7071c6bb-af7b-45bf-82f4-28fe53208c43

### DIFF
--- a/packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
+++ b/packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
@@ -199,13 +199,15 @@ export default class ColorPickerView extends View {
 	public focus(): void {
 		// In some browsers we need to move the focus to the input first.
 		// Otherwise, the color picker doesn't behave as expected.
+		// In Chrome, after selecting the color via slider the first time,
+		// the editor collapses the selection and doesn't apply the color change.
 		// In FF, after selecting the color via slider, it instantly moves back to the previous color.
 		// In all iOS browsers and desktop Safari, once the saturation slider is moved for the first time,
 		// editor collapses the selection and doesn't apply the color change.
 		// See: https://github.com/cksource/ckeditor5-internal/issues/3245, https://github.com/ckeditor/ckeditor5/issues/14119,
 		// https://github.com/cksource/ckeditor5-internal/issues/3268.
 		/* istanbul ignore next -- @preserve */
-		if ( !this._config.hideInput && ( env.isGecko || env.isiOS || env.isSafari ) ) {
+		if ( !this._config.hideInput && ( env.isGecko || env.isiOS || env.isSafari || env.isBlink ) ) {
 			const input: LabeledFieldView<InputTextView> = this.hexInputRow!.children.get( 1 )! as LabeledFieldView<InputTextView>;
 
 			input.focus();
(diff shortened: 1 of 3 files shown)


dil-bparihar:
Yes it is an issue. When this PR will be merged and deployed. @arkflpc

packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -186,9 +186,29 @@ export default class ColorPickerView extends View {
...
+
+			// At first, set the color internally in the component. It's converted to the configured output format.
+			this.set( 'color', color );
+
+			if ( colorSelectedEventDebounce !== null ) {
+				clearTimeout( colorSelectedEventDebounce );
+			}
+
+			// Let's wait until selection is being moved from web component to the editor.
+			colorSelectedEventDebounce = setTimeout( () => {
```
arkflpc, line 208: Consider the consequences (like memory leaks) of the case when the view is `destroy`ed before the timeout.

packages/ckeditor5-ui/src/colorpicker/colorpickerview.ts
```diff
@@ -186,9 +186,29 @@ export default class ColorPickerView extends View {
 			this.picker.shadowRoot!.appendChild( styleSheetForFocusedColorPicker );
 		}
 
+		// Workaround for unstable selection after firing 'color-changed' event in Webkit-based browsers.
+		// The selection is being collapsed after the event is fired, causing the font color plugin
+		// not to update the selected editor content. Setting the color immediately after the event
+		// is fired is a workaround for this issue. It's not a perfect solution, but it's the best we can do for now.
+		// It may be removed in the future if the issue is fixed in the browser.
```
gorzelinski, line 193: A little redundant
```suggestion
		// It may be removed if the issue is fixed in the browser.
```

gorzelinski (approved):

dil-bparihar:
@Mati365 when we are going to merge it ?

packages/ckeditor5-ui/tests/colorselector/colorselectorview.js
```diff
@@ -432,6 +433,11 @@ describe( 'ColorSelectorView', () => {
 		} );
 
 		it( 'should execute when color picker is focused and enter pressed', () => {
+			// Focusing input and then color picker seems to fail focus handling in test suite
+			// that uses headless chrome browser. This is a workaround for that, as this deactivates
+			// focusing input before the color picker.
```
gorzelinski, line 438:
```suggestion
			// Focusing input and then the color picker breaks focus handling in the test 
			// suite that uses a headless Chrome browser. It is a workaround for that, as 
			// this deactivates focusing input before the color picker.
```

mmotyczynska (approved):
This solution is much better than the previous one, and seems to work well 👍

PR #17110, opened 2024-09-17
Docs: Add the Next.js CDN guide.

gorzelinski:

Docs: Add the Next.js CDN guide.

---

### Additional information

### DIFF
--- a/docs/getting-started/integrations-cdn/next-js.md
+++ b/docs/getting-started/integrations-cdn/next-js.md
@@ -28,7 +28,7 @@ First, install the CKEditor 5 packages:
 * `ckeditor5` &ndash; package with open-source plugins and features.
 * `ckeditor5-premium-features` &ndash; package with premium plugins and features.
 
-Depending on your configuration and chosen plugins, you may need to install the first or both packages.
+Depending on your configuration and selected plugins, you may need to install only the first one or both.
 
 ```bash
 npm install ckeditor5 ckeditor5-premium-features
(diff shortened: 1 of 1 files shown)


docs/getting-started/integrations-cdn/next-js.md
```diff
@@ -17,8 +17,110 @@ CKEditor&nbsp;5 does not support server-side rendering yet, but you can integrat
...
+
+```bash
+npm install @ckeditor/ckeditor5-react
+```
+
+You will use the installed dependencies in a React component. Create a new component in the components directory, for example, `components/custom-editor.js`. Inside the component file, import all necessary dependencies. Then, create a functional component that returns the CKEditor&nbsp;5 React component. The example below shows how to use the component with both open-source and premium plugins.
+
+The App Router, by default, uses server components. It means you need to mark a component as client-side explicitly. You can achieve that by using the `'use client'` directive at the top of the file, above your imports. You do not need the directive if you use the Pages Router.
+
+In the below example, the `useCKEditorCloud` hook is used to load the editor code and plugins from CDN. To use premium plugins, set the `premium` property to `true` and provide your license key in the configuration. For more information about the `loadCKEditorCloud` helper, see the {@link getting-started/setup/loading-cdn-resources Loading CDN resources} guide.
```
godai78, line 47: DOn't we need to provide the licence call anyway, just without premium it's `GPL`? This might need to be clarified.

godai78 (approved):

docs/getting-started/integrations-cdn/next-js.md
```diff
@@ -17,8 +17,110 @@ CKEditor&nbsp;5 does not support server-side rendering yet, but you can integrat
...
+
+```bash
+npm install ckeditor5 ckeditor5-premium-features
+```
+
+Next.js is based on React, so install the [CKEditor 5 WYSIWYG editor component for React](https://www.npmjs.com/package/@ckeditor/ckeditor5-react), too:
+
+```bash
+npm install @ckeditor/ckeditor5-react
+```
```
Mati365, line 41: I'd note that these dependencies are only needed for typings and if Next.js in JavaScript mode is used then there is no need to install these packages.

docs/getting-started/integrations-cdn/next-js.md
```diff
@@ -17,8 +17,110 @@ CKEditor&nbsp;5 does not support server-side rendering yet, but you can integrat
...
+```jsx
+// components/custom-editor.js
+'use client' // only in App Router
+
+import React from 'react';
+import { CKEditor, useCKEditorCloud } from '@ckeditor/ckeditor5-react';
+
+const CustomEditor = () => {
+	const cloud = useCKEditorCloud( {
+		version: '43.1.0',
```
Mati365, line 58:
```suggestion
		version: '{@var ckeditor5-version}',
```

Mati365 (commented):
Looks cool. Small things noted

docs/getting-started/integrations-cdn/next-js.md
godai78: Nice catch!

Mati365 (approved):

godai78 (approved):

PR #17113, opened 2024-09-18
Delay selections in `should not enter infinite loop` test due to changes in latest Chrome

Mati365:

Tests (engine): Delay selections in `should not enter infinite loop` test due to changes in Chrome 129.

---

### Additional information

It looks like the newest Chrome version debounces `selectionchange` events, and I made a small delay to prevent this. In other words - firing multiple selection updates in small period of the time results in one `selectionchange` event instead of few. 

Fixes:
obraz

After bumping Chrome from `128` to `129` in test suite. 
edit: More likely Karma <--> Chrome driver issue, works fine in real browser.


pszczesniak (commented):
Maybe would be good to change the topic and commit message to mark in what version of `Chrome` it starts to fail, `latest` is too general.

pszczesniak (approved):

PR #17140, opened 2024-09-23
The dropdown menu component should not cause an editor blur if used in a BalloonToolbar

oleq:

Fix (ui): The dropdown menu component should not cause an editor blur if used in a `BalloonToolbar` while the user hovers a nested menu. 

Feature (utils): Made `FocusTracker` extendable with other `FocusTracker` instances to allow for logical focus tracking across separate DOM sub-trees.

---

### Additional information

Closes https://github.com/cksource/ckeditor5-commercial/issues/6633.


packages/ckeditor5-utils/src/focustracker.ts
```diff
@@ -136,3 +269,20 @@ export default class FocusTracker extends /* #__PURE__ */ DomEmitterMixin( /* #_
...
+export type ViewWithFocusTracker = View & { focusTracker: FocusTracker };
+
+/**
+ * Checks whether a view is an instance of {@link ~ViewWithFocusTracker}.
+ */
+export function isViewWithFocusTracker( view: any ): view is ViewWithFocusTracker {
+	return 'focusTracker' in view && view.focusTracker instanceof FocusTracker;
+}
+
+function isElement( value: any ): value is Element {
```
oleq, line 286: I didn't want to pollute this PR with unrelated changes, but it seems we have several similar functions scattered in the codebase. I'll create a follow-up and move this to `utils/src/dom` to share across the project.

packages/ckeditor5-utils/src/focustracker.ts
```diff
@@ -66,16 +76,92 @@ export default class FocusTracker extends /* #__PURE__ */ DomEmitterMixin( /* #_
...
 	 */
 	public get elements(): Array<Element> {
 		return Array.from( this._elements.values() );
 	}
 
 	/**
-	 * Starts tracking the specified element.
+	 * List of external focus trackers that contribute to the state of this focus tracker. See {@link #add} to learn more.
+	 */
+	public get externalFocusTrackers(): Array<FocusTracker> {
```
oleq, line 90: I decided I don't want to extend `#elements` with elements of external focus trackers to keep the previous API contract (avoid BCs). I came up with this property that will allow integrators to dig deeper if they want, though.

oleq: Until now `#focusedElement` always was one of `#elements.

But now `#focusedElement` may point to a DOM element outside of `#elements` if an external FT's element is focused. I didn't want to leave `#focusedElement === null` if `#isFocused === true` in that situation so the integrators may need to access `#externalFocusTrackers` to learn more.

PR #17166, opened 2024-09-26
Bookmark insert/update button text

illia-stv:

Feature (bookmark): Should update text of the form button whether bookmark is selected or not . Closes #000.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-bookmark/src/bookmarkui.ts
+++ b/packages/ckeditor5-bookmark/src/bookmarkui.ts
@@ -340,7 +340,7 @@ export default class BookmarkUI extends Plugin {
 	}
 
 	/**
-	 * If bookmark element is selected form button label is set on 'Update' otherwise its 'Insert'.
+	 * When bookmark element is selected, form button label should be set on 'Update' otherwise it is 'Insert'.
 	 */
 	private _updateFormButtonLabel() {
 		const locale = this.editor.locale;
--- a/packages/ckeditor5-bookmark/src/bookmarkui.ts
+++ b/packages/ckeditor5-bookmark/src/bookmarkui.ts
@@ -198,7 +198,7 @@ export default class BookmarkUI extends Plugin {
 		);
 
 		// Disable the "save" button if the command is disabled.
-		formView.insertButtonView.bind( 'isEnabled' ).toMany(
+		formView.buttonView.bind( 'isEnabled' ).toMany(
 			commands,
 			'isEnabled',
 			( ...areEnabled ) => areEnabled.some( isEnabled => isEnabled )
@@ -340,6 +340,17 @@ export default class BookmarkUI extends Plugin {
 	}
 
 	/**
+	 * If bookmark element is selected form button label is set on 'Update' otherwise its 'Insert'.
+	 */
+	private _updateFormButtonLabel() {
+		const locale = this.editor.locale;
+		const t = locale.t;
+		const element = !!this._getSelectedBookmarkElement();
+
+		this.formView!.buttonView.label = element ? t( 'Update' ) : t( 'Insert' );
+	}
+
+	/**
 	 * Adds the {@link #actionsView} to the {@link #_balloon}.
 	 *
 	 * @internal
@@ -412,7 +423,7 @@ export default class BookmarkUI extends Plugin {
 		if ( this._isFormInPanel ) {
 			// Blur the input element before removing it from DOM to prevent issues in some browsers.
 			// See https://github.com/ckeditor/ckeditor5/issues/1501.
-			this.formView!.insertButtonView.focus();
+			this.formView!.buttonView.focus();
 
 			// Reset the ID field to update the state of the submit button.
 			this.formView!.idInputView.fieldView.reset();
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-bookmark/src/bookmarkui.ts
```diff
@@ -339,6 +339,17 @@ export default class BookmarkUI extends Plugin {
 		} );
 	}
 
+	/**
+	 * When bookmark element is selected, form button label should be set on 'Update' otherwise it is 'Insert'.
+	 */
+	private _updateFormButtonLabel() {
+		const locale = this.editor.locale;
+		const t = locale.t;
```
pszczesniak, line 347:
```suggestion
		const t = this.editor.locale.t;
```

packages/ckeditor5-bookmark/src/ui/bookmarkformview.ts
```diff
@@ -54,7 +54,7 @@ export default class BookmarkFormView extends View {
 	/**
 	 * The Insert button view.
```
pszczesniak, line 55:
```suggestion
	 * The Submit button view.
```

packages/ckeditor5-bookmark/src/bookmarkui.ts
```diff
@@ -339,6 +339,17 @@ export default class BookmarkUI extends Plugin {
 		} );
 	}
 
+	/**
+	 * When bookmark element is selected, form button label should be set on 'Update' otherwise it is 'Insert'.
+	 */
+	private _updateFormButtonLabel() {
```
pszczesniak, line 345: In the methods where we already fire this function we already have `this._getSelectedBookmarkElement()`, maybe it would be better to pass it as a parameter (or just boolean) to not have to execute this function again?

For example:

```suggestion
	private _updateFormButtonLabel( isBookmarkSelected: boolean ) {
```

pszczesniak (changes requested):

packages/ckeditor5-bookmark/tests/bookmarkui.js
```diff
@@ -929,6 +929,37 @@ describe( 'BookmarkUI', () => {
...
+
+		describe( 'button label', () => {
+			it( 'should have "Insert" by default', () => {
+				bookmarkUIFeature._addFormView();
+				formView = bookmarkUIFeature.formView;
+
+				expect( formView.buttonView.label ).to.equal( 'Insert' );
+			} );
+
+			it( 'should have "Insert" label when bookamrk is not selected', () => {
```
mmotyczynska, line 941:
```suggestion
			it( 'should have "Insert" label when bookmamrk is not selected', () => {
```

packages/ckeditor5-bookmark/tests/bookmarkui.js
```diff
@@ -929,6 +929,37 @@ describe( 'BookmarkUI', () => {
...
+
+				bookmarkUIFeature._addFormView();
+				formView = bookmarkUIFeature.formView;
+
+				bookmarkUIFeature._showUI();
+
+				expect( formView.buttonView.label ).to.equal( 'Insert' );
+			} );
+
+			it( 'should have "Update" label when bookamrk is selected', () => {
```
mmotyczynska, line 952:
```suggestion
			it( 'should have "Update" label when bookmark is selected', () => {
```

packages/ckeditor5-bookmark/src/bookmarkui.ts
```diff
@@ -511,6 +522,8 @@ export default class BookmarkUI extends Plugin {
 		let prevSelectedBookmark = this._getSelectedBookmarkElement();
 		let prevSelectionParent = getSelectionParent();
 
+		this._updateFormButtonLabel();
```
mmotyczynska, line 525: Instead of calling this twice in `_startUpdatingUI()`, can we just update the button text once at the end of the `_showUI()`?

mmotyczynska (changes requested):

packages/ckeditor5-bookmark/src/bookmarkui.ts
```diff
@@ -339,6 +339,17 @@ export default class BookmarkUI extends Plugin {
 		} );
 	}
 
+	/**
+	 * When bookmark element is selected, form button label should be set on 'Update' otherwise it is 'Insert'.
```
mmotyczynska, line 343:
```suggestion
	 * When a bookmark element is selected, the form button label should be set to 'Update'; otherwise, it should be 'Insert'.
```

packages/ckeditor5-bookmark/src/bookmarkui.ts
illia-stv: I thought about it but found an edge case. In that case if balloon is not closed when you back to the form view the button is not updated, so we need to initialize properly the value of the button and updated properly.

[screen-capture (1).webm](https://github.com/user-attachments/assets/d53ac37b-e9c7-4d6c-9fe6-6c260731cbef)

mmotyczynska: Do we have test for this edge case?

illia-stv: No, but right on my way to add it

pszczesniak (approved):

packages/ckeditor5-bookmark/src/ui/bookmarkformview.ts
```diff
@@ -263,7 +263,7 @@ export default class BookmarkFormView extends View {
 	 * @param eventName An event name that the `ButtonView#execute` event will be delegated to.
```
mmotyczynska, line 263: This is no longer needed here, if you removed the function parameter.

mmotyczynska (changes requested):

mmotyczynska (approved):

PR #17167, opened 2024-09-26
[Bookmark] Enable drag and drop in manual tests.

pszczesniak:

Internal: Enable drag and drop in manual tests.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-bookmark/package.json
+++ b/packages/ckeditor5-bookmark/package.json
@@ -27,7 +27,6 @@
     "@ckeditor/ckeditor5-dev-utils": "^42.0.0",
     "@ckeditor/ckeditor5-easy-image": "43.1.1",
     "@ckeditor/ckeditor5-editor-classic": "43.1.1",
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-bookmark/package.json
```diff
@@ -38,7 +38,6 @@
     "@ckeditor/ckeditor5-select-all": "43.1.1",
     "@ckeditor/ckeditor5-table": "43.1.1",
     "@ckeditor/ckeditor5-theme-lark": "43.1.1",
-    "@ckeditor/ckeditor5-typing": "43.1.1",
```
mmotyczynska, line 41: If we add `Essentials` as dependency, do we still need `Enter` and `Undo` in dependencies?

pszczesniak: Good catch, wille remove, I just need to change plugins to `Essentials` in editing tests.

mmotyczynska (approved):

PR #17186, opened 2024-09-30
Bookmark id validation

illia-stv:

Internal: Bookmark id validation in command.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-bookmark/src/insertbookmarkcommand.ts
+++ b/packages/ckeditor5-bookmark/src/insertbookmarkcommand.ts
@@ -61,7 +61,7 @@ export default class InsertBookmarkCommand extends Command {
 			/**
 			 * Insert bookmark command can be executed only with a valid id.
 			 *
-			 * Spaces are not allowed in the bookmark id but id cannot be empty.
+			 * A valid bookmark ID must be a non-empty string and must not contain any spaces.
 			 *
 			 * @error insert-bookmark-command-executed-with-invalid-id
 			 */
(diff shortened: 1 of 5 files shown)


mmotyczynska:
I'm not sure if it's a good idea to log a warning **only** when the `name` contains spaces. I think we should either log a warning for **all invalid cases** (such as a missing/empty `name` or non-string `name`), or not warn at all and simply return early from command execution, as it was implemented before.

I'd prefer option 1 to maintain consistency and to inform the user when something is wrong with the `name`. WDYT @pszczesniak ?

Reference code:
```
		if ( !bookmarkId || typeof bookmarkId !== 'string' ) {
			return;
		}


		if ( !isBookmarkIdValid( bookmarkId ) ) {
			/**
			 * Insert bookmark command can be executed only with a valid id.
			 *
			 * Spaces are not allowed in the bookmark id.
			 *
			 * @error insert-bookmark-command-executed-with-invalid-id
			 */
			logWarning( 'insert-bookmark-command-executed-with-invalid-id' );


			return;
		}
```

pszczesniak (commented):
Yes @mmotyczynska , i agree, cause throwing an error on one case and not on other may confuse a user, when result will always be be the same - bookmark wont be added.

2nd thing to be remembered: gathering error codes to create `error-codes` page in docs. 

Every error will be reflected there, and when you add same error more than once, it will be there same amount of times. Example https://ckeditor.com/docs/ckeditor5/latest/support/error-codes.html#error-annotationsuis-missing-ui

mmotyczynska (changes requested):

mmotyczynska (changes requested):

mmotyczynska (approved):

PR #17193, opened 2024-10-01
Make sure existing picture element is handled correctly on 'sources' downcast.

### LINKED ISSUE #17192: Picture "sources" downcast converter duplicates "picture" element when "sources" model attribute updated

1. Go to CKBox manual test (e.g. http://localhost:8125/ckeditor5-ckbox/tests/manual/ckbox.html).
2. Insert any image from CKBox as a first editor element (it's important for image to have `picture` element).
3. Run `editor.model.change( writer => writer.setAttribute('sources', [ { srcset: './sample.png' } ], editor.model.document.getRoot().getChild(0)) );`.
4. Take a look on image element in CKE5 inspector in `View` tab.

### ✔️ Expected result

The view data has structure like `figure > picture > sources + img`.

### ❌ Actual result

The view data has structure like `figure > picture > sources > picture > sources + img`.

image

### ❓ Possible solution

The `sources` downcast converter does not check the structure of the updated element and always inserts new `picture` element. The good news is that on `editor.getData()` resulting HTML is correct without nested strucure.

## 📃 Other details

* Browser: Chrome
* OS: macOS
* First affected CKEditor version: Not sure, tested on latest one (`43.1.1`).
* Installed CKEditor plugins: See mentioned manual test.

---

f1ames:

Fix: Handle existing picture element correctly on 'sources' downcast. Closes #17192.

---

### Additional information

See https://github.com/ckeditor/ckeditor5/issues/17192.


packages/ckeditor5-image/src/image/converters.ts
```diff
@@ -249,9 +249,18 @@ export function downcastSourcesAttribute( imageUtils: ImageUtils ): ( dispatcher
...
-			// Insert the picture and move img into it.
-			viewWriter.insert( viewWriter.createPositionBefore( imgElement ), pictureElement );
-			viewWriter.move( viewWriter.createRangeOn( imgElement ), viewWriter.createPositionAt( pictureElement, 'end' ) );
+			const prevPictureElement = imgElement.parent!.is( 'element', 'picture' ) ? imgElement.parent : null;
+
+			if ( prevPictureElement ) {
+				// Insert a new picture, move img into it, remove previous picture element (ckeditor5#17192).
+				viewWriter.insert( viewWriter.createPositionBefore( prevPictureElement ), pictureElement );
+				viewWriter.move( viewWriter.createRangeOn( imgElement ), viewWriter.createPositionAt( pictureElement, 'end' ) );
+				viewWriter.remove( prevPictureElement );
```
niegowski, line 258: What would happen if other converters downcast other attributes to the picture element? Would those be lost? Maybe in such a case, we should remove source elements from the picture element and reinsert them into the already existing picture element.

niegowski: Maybe not emptying but removing source elements and inserting new ones (img element should also be reused)

f1ames: Fixed in https://github.com/ckeditor/ckeditor5/pull/17193/commits/893b1e307a0b2d9446f20d1f1aebd0d26f49a361 and even the code go slightly simpler 👍 I had to add a custom attribute handling in test because we don't have anything for picture from what I see.

packages/ckeditor5-image/src/image/converters.ts
```diff
@@ -249,8 +242,24 @@ export function downcastSourcesAttribute( imageUtils: ImageUtils ): ( dispatcher
...
+			// Reuse existing <picture> element (ckeditor5#17192) or create a new one.
+			const pictureElement = hasPictureElement ? imgElement.parent : viewWriter.createContainerElement( 'picture', null );
+
+			if ( !hasPictureElement ) {
+				viewWriter.insert( viewWriter.createPositionBefore( imgElement ), pictureElement );
+			}
+
+			viewWriter.remove( viewWriter.createRangeIn( pictureElement ) );
+
+			// Make sure <picture> does not break attribute elements, for instance <a> in linked images.
```
niegowski, line 256: This comment looks unrelated to the code. It's a more generic comment about collecting attribute elements and reapplying them.

f1ames: True, I moved it with the code but it doesn't seem to fit there. Fixed 👍

niegowski (approved):

Issue #17203, opened 2024-10-02
`npx webpack` with `ckeditor5-dev-utils` 44.0.0 fails to build CKEditor with `Error [ERR_REQUIRE_ESM]: require() of ES Module`


wyattoday:
## 📝 Provide detailed reproduction steps (if any)

1. Run `npm update --dd`
2. Then run `npx webpack`
3. Watch it fail with...

```
[webpack-cli] Failed to load 'C:\location-to\ckeditor5-custom\webpack.config.js' config
[webpack-cli] Error [ERR_REQUIRE_ESM]: require() of ES Module C:\location-to\ckeditor5-custom\node_modules\@ckeditor\ckeditor5-dev-utils\lib\index.js from C:\location-to\ckeditor5-custom\webpack.config.js not supported.
Instead change the require of index.js in C:\location-to\ckeditor5-custom\webpack.config.js to a dynamic import() which is available in all CommonJS modules.
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at Object.<anonymous> (C:\location-to\ckeditor5-custom\webpack.config.js:12:29) {
  code: 'ERR_REQUIRE_ESM'
}
```

The line in question that it's angry about (in our webpack.config.js):

```
const { bundler, styles } = require( '@ckeditor/ckeditor5-dev-utils' );
```

### ✔️ Expected result

A successful build.

### ❌ Actual result

A failure (see above).

### ❓ Possible solution

Rolling back `ckeditor5-dev-utils` to 43.0.0 (`npm install @ckeditor/ckeditor5-dev-utils@43.0.0`) fixes the bug. So it was introduced in the last few weeks between 43.0.0 and 44.0.0.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

pomek:
According to https://github.com/ckeditor/ckeditor5-dev/releases/tag/v44.0.0, all `ckeditor5-dev-*` packages are now ESM. Please update your configuration to use an `import` statement or an `import()` function.

wyattoday:
I see. The documentation hasn't been updated: https://ckeditor.com/docs/ckeditor5/latest/getting-started/legacy/advanced/alternative-setups/integrating-from-source-webpack.html

pomek:
The `@latest` npm tag for the `@ckeditor/ckeditor5-dev-utils` package (and others as well) now points to v43.

[...]

PR #17208, opened 2024-10-02
Aligned legacy webpack configurations to changes in ckeditor5-dev-*

pomek:

Internal: Aligned legacy webpack configurations to changes in ckeditor5-dev-*. See #17203.

---

### Additional information

### DIFF
--- a/docs/getting-started/advanced/integrating-from-source-webpack.md
+++ b/docs/getting-started/advanced/integrating-from-source-webpack.md
@@ -219,6 +219,8 @@ export default {
 If you use [Webpack Encore](https://github.com/symfony/webpack-encore), you can use the following configuration:
 
 ```js
+// Since v44, the `ckeditor5-dev-*` packages are ESM only.
+// Hence, ensure to install the latest compatible version of the packages.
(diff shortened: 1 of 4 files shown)


psmyrek (commented):
I checked the first guide - _Integrating from source using webpack_.

1. Missing `import { fileURLToPath } from 'url';` in `webpack.config.js`.
2. The package must be configured as a module to use imports instead of `require` calls (missing a note to add `"type": "module"` to `package.json` or `.mjs` extension).
3. Imports must have file extensions - `import ClassicEditor from './ckeditor.js';`.
4. Latest `ckeditor5-dev-*` packages are not compatible with latest CKEditor 5. There are a lot of warnings about dropped support for `@nest` thing from PostCSS.

pomek:
Mhm. Right. So, the only way to make the guides work is to point to the latest CJS version instead of ESM.

pomek:
Closing in favor of https://github.com/ckeditor/ckeditor5/issues/17203. I will create another PR soon. Thanks @psmyrek for the review.

PR #17216, opened 2024-10-03
[Bookmark] Update mobile look of balloon views in bookmarks.

pszczesniak:
Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Internal: Should bookmark balloons looks good on mobile devices.

---

### Additional information

_We've got a green light 🟢  from @dwandeljanicka _

### DIFF
--- a/packages/ckeditor5-bookmark/src/ui/bookmarkactionsview.ts
+++ b/packages/ckeditor5-bookmark/src/ui/bookmarkactionsview.ts
@@ -13,6 +13,7 @@ import { icons } from 'ckeditor5/src/core.js';
 
 // eslint-disable-next-line ckeditor5-rules/ckeditor-imports
 import '@ckeditor/ckeditor5-ui/theme/components/responsive-form/responsiveform.css';
+import '../../theme/bookmarkactions.css';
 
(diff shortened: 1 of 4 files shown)


mmotyczynska (commented):
I'd think about removing the tooltip from the button (or position it differently if it's possible) because it:
- covers two other buttons
- If you click on it (and it's quite easy to do so by accident), the balloon closes and you have to open it again.

But I don't know about the a11y in this case 🤔 
image

pszczesniak:
@mmotyczynska very same case is with link:



But not in 100% the same.

:thinking: 

Yes, will style it exactly like in link.

pszczesniak:
@mmotyczynska done.

mmotyczynska (approved):
Looks a bit better. At least we keep consistency in `link` and `bookmark` 👍

pszczesniak:
I've updated also the look of the submit button on mobiles, to make it align with look for example of "Find and replace" (especially the margins and paddings).

Find and replace (mobile):



Bookmark (mobile):

Issue #17182, opened 2024-09-30
Create a Basic plugin tutorial feedback


wwalc:
## 📝 Provide a description of requested docs changes

(1) The tutorial assumes Husky is installed, or at least with the latest node installed via nvm it did not work for me.

Follow https://ckeditor.com/docs/ckeditor5/43.1.1/framework/tutorials/creating-simple-plugin-timestamp.html without having Husky installed.

```
wiktor@Wiktors-MBP starter-files % node --version 
v20.17.0
wiktor@Wiktors-MBP starter-files % ls -al        
total 24
drwxr-xr-x  5 wiktor  staff  160 Sep 30 10:34 .
drwxr-xr-x  4 wiktor  staff  128 Sep 30 10:34 ..
-rw-r--r--  1 wiktor  staff  500 Sep 30 10:34 index.html
-rw-r--r--  1 wiktor  staff  610 Sep 30 10:34 main.js
-rw-r--r--  1 wiktor  staff  422 Sep 30 10:34 package.json
wiktor@Wiktors-MBP starter-files % npm install                                                       

> ckeditor5-tutorials-examples@0.1.0 prepare
> husky

sh: husky: command not found
npm error code 127
npm error path /Users/wiktor/workspace/ckeditor5-tutorials-examples
npm error command failed
npm error command sh -c husky
npm error A complete log of this run can be found in: /Users/wiktor/.npm/_logs/2024-09-30T08_37_04_944Z-debug-0.log
```

(2) The order of the code in sample project is different than in docs.

Docs:

![image](https://github.com/user-attachments/assets/3c1872d2-ee8e-410a-9d0c-e9e109863c78)

Real code in starter files:

![image](https://github.com/user-attachments/assets/ad3467b0-0604-42d6-9b73-4135cf9960d5)

This way it takes a little bit more time to understand what to add manually, especially that code that we introduced in the tutorial is not "marked" in any way (e.g. with code comments).

Witoso:
> (1) The tutorial assumes Husky is installed, or at least with the latest node installed via nvm it did not work for me.

@filipsobol I guess this is a recent change (last week?), added with the prevention to not leak mistakenly license keys. I wonder if it makes sense to have husky added as a dev dep. Also, maybe we should reconsider promoting `degit`?

> This way it takes a little bit more time to understand what to add manually, especially that code that we introduced in the tutorial is not "marked" in any way (e.g. with code comments).

[...]

PR #17232, opened 2024-10-07
Docs: Use `npx degit` instead of `git clone` in tutorials

filipsobol:

Docs: Use `npx degit` instead of `git clone` in tutorials. See #17182.

---

### Additional information

### DIFF
--- a/docs/tutorials/widgets/using-react-in-a-widget.md
+++ b/docs/tutorials/widgets/using-react-in-a-widget.md
@@ -31,19 +31,21 @@ There are a couple of things you should know before you start:
 
 ## Let's start
 
-The easiest way to set up your project is to grab the starter files from the [GitHub repository for this tutorial](https://github.com/ckeditor/ckeditor5-tutorials-examples/tree/main/react-widget/starter-files). We gathered all the necessary dependencies there, including some CKEditor 5 packages and other files needed to start the editor.
-
-The editor has already been created in the `main.js` file with some basic plugins. All you need to do is clone the repository, navigate to the starter-files directory, run the `npm install` command, and you can start coding right away.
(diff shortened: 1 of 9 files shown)


filipsobol:
The `npx` commands are often too long because of the long files paths.

Code snippet

This is an issue because the code snippet cannot be copied using the “Copy” button, as these commands are supposed to be run consecutively.

One way to fix this it to rename the repository from `ckeditor5-tutorials-examples` to `keditor5-tutorials` or even just `tutorials` and change the folder names. How do you feel about this? @Witoso @gorzelinski

Witoso:
IMHO, let's first merge asap, as there's another issue reported: #17239

Witoso (dismissed):

docs/tutorials/widgets/using-react-in-a-widget.md
```diff
@@ -31,19 +31,21 @@ There are a couple of things you should know before you start:
...
 npm install
 npm run dev
 ```
 
-You should see a "Hello world" application in your web browser, which might not be much but it is a good start:
+This will create a new directory called `react-widget` with the necessary files. The `npm install` command will install all the dependencies, and `npm run dev` will start the development server.
+
+The editor with some basic plugins is created in the `main.js` file.
+
+Open the URL displayed in your terminal. If everything went well, you should see a "Hello world" application in your web browser, which might not be much but it is a good start:
```
gorzelinski, line 48:
```suggestion
Open the URL displayed in your terminal. If everything goes well, you should see a "Hello world" application in your web browser, which might not be much, but it is a good start:
```

gorzelinski: "It ain't much but it's honest work"

Witoso (approved):

PR #17247, opened 2024-10-09
Made the page unscrollable while the modal is visible

### LINKED ISSUE #17093: Disable page scroll when modal is open

When dialog is opened in the model mode, it is still possible to scroll the page "behind it". I think this is a bad UX, and it may lead to an unintended scrolling of the page, especially when there are dialogs that also has scrolls.

We should either disable the scroll or remove it. This unfortunately leads to the page "jumping", so if possible, I think that the best would be to disable the scroll if it is visible. If that's not possible we can discuss if it's worse to make the website jump or to allow scrolling in background.

oleq:

Fix (ui): Made the page unscrollable while the modal is visible. Closes #17093.

---

## Some info

There are two main strategies for freezing the web page scroll available:
1. `overflow: hidden` on `<body>`: A very straightforward recipe. It fails in iOS, though. Still, I chose this one because I figured out we can sacrifice the mobile UX for the sake of a simple code and a quicker solution. Besides, Bootstrap.js uses this one and they don't seem to mind at all.
2. `position: fixed` on `<body>`: Initially, I went with this [one](https://css-tricks.com/prevent-page-scrolling-when-a-modal-is-open/) because it looked promising but soon I realized it has some drawbacks (also in non-mobile browsers). For instance, it causes page reflows when the body is narrower than the viewport or when the body was positioned with `absolute` in the first place. It requires some extra code (the article shows just basics, there's more to it once you look into it) and feels more complex. I decided it was not worth the hassle (for now).
3. A hybrid that uses `overflow: hidden` in most cases and `position: fixed` for iOS with additional hacks. It's a fairly popular [MIT-licensed npm module](https://www.npmjs.com/package/body-scroll-lock) but I found it overkill for what we need right now.

### DIFF
--- a/packages/ckeditor5-ui/tests/manual/dialog/dialog-body-scroll-lock.html
+++ b/packages/ckeditor5-ui/tests/manual/dialog/dialog-body-scroll-lock.html
@@ -1,0 +1,53 @@
+<form>
+	<fieldset>
+		<legend>Is document scrollable?</legend>
+		<div>
+			<input type="radio" id="yes" name="isScrollable" value="yes" />
+			<label for="yes">yes</label>
+		</div>
+		<div>
+			<input type="radio" id="no" name="isScrollable" value="no" checked />
+			<label for="no">no</label>
+		</div>
+	</fieldset>
+	<fieldset>
+		<legend>Has &lt;body&gt; position?</legend>
+		<div>
+			<input type="radio" id="yes" name="hasPosition" value="yes" />
+			<label for="yes">yes</label>
+		</div>
+		<div>
+			<input type="radio" id="no" name="hasPosition" value="no" checked />
+			<label for="no">no</label>
+		</div>
+	</fieldset>
+</form>
+
+<p>Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae, ultricies eget, tempor sit amet, ante. Donec eu libero sit amet quam egestas semper. Aenean ultricies mi vitae est. Mauris placerat eleifend leo.</p>
+
+<div id="editor">Editor content.</div>
+
+<p>Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae, ultricies eget, tempor sit amet, ante. Donec eu libero sit amet quam egestas semper. Aenean ultricies mi vitae est. Mauris placerat eleifend leo.</p>
+
+<style>
+	.ck.ck-editor {
+		margin-top: 300px;
+	}
+
+	body {
+		outline: 2px dashed blue;
+		outline-offset: -2px;
+	}
+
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-ui/tests/manual/dialog/dialog-body-scroll-lock.html
```diff
@@ -0,0 +1,53 @@
...
+	</fieldset>
+	<fieldset>
+		<legend>Has &lt;body&gt; position?</legend>
+		<div>
+			<input type="radio" id="yes" name="hasPosition" value="yes" />
+			<label for="yes">yes</label>
+		</div>
+		<div>
+			<input type="radio" id="no" name="hasPosition" value="no" checked />
+			<label for="no">no</label>
```
Dumluregn, line 21: Ids are duplicated so labels don't work well.

Dumluregn (commented):
I found two more potential issues with the current solution:

* Opening a modal blocks the "pull down" interactions on mobile, i.e. you can't refresh the page this way until you close a modal.
* If there's no padding on the `<body>` element reserved for the scrollbar, we get a small page reflow whenever we open or close the modal.

With that in mind, I'd still say this PR brings more value than problems so I'd go with the chosen solution if QA doesn't reveal something critical and business agrees to ignore iOS.

lkszzajac:
### Page is still scrollable on mobile or when using touch simulator in dev tools (see recording)

- On desktop touch simulator (devtools) scrolling is possible right away

https://github.com/user-attachments/assets/cc0a5fb2-20e1-4fac-b196-8ad8a9f783f4

- On iOS scroll lock is applied with a delay, and is lifted if the user zooms in/out via pinch-to-zoom (happens automatically when focusing on the text field in the manual test)
- On Android scrolling inside the dialog is not possible if the document itself is scrollable. Rest of the page is not scroll-locked.

https://github.com/user-attachments/assets/2e14896b-42a2-4199-9d6f-eb5be5693e60

packages/ckeditor5-ui/tests/manual/dialog/dialog-body-scroll-lock.html
oleq: Addressed in d03717ac.

oleq:
I added unit tests in [d3cb8e6](https://github.com/ckeditor/ckeditor5/pull/17247/commits/d3cb8e64bfe78e3d661dcf74db12be87554cb629).

During a summary call with @DawidKossowski and @f1ames we agreed with what @Dumluregn

> With that in mind, I'd still say this PR brings more value than problems so I'd go with the chosen solution if QA doesn't reveal something critical and business agrees to ignore iOS. 

This is the same approach used by Bootstrap.js or MUI and it should satisfy the majority of our usebase. If this becomes a serious problem for mobile (touch) device users, we'll revisit the solution but until then, we go with the simple `overflow: hidden`.

Dumluregn (approved):

Issue #16648, opened 2024-07-02
Find and replace highlight sometimes jumps to the first item instead of next after replacing


f1ames:
## 📝 Provide detailed reproduction steps (if any)

I'll start with videos since it illustrates this case nicely.

1st example works fine, and replace cycles nicely over words in order:

https://github.com/cksource/ckeditor5-commercial/assets/1061942/678d2099-5e1a-4a77-acef-43c844cb4ca0

2nd and 3rd shows that highlight sometimes jumps to the first item:

https://github.com/cksource/ckeditor5-commercial/assets/1061942/82279342-3e5f-446a-89bb-3b08bc81f928

https://github.com/cksource/ckeditor5-commercial/assets/1061942/0054df5f-27ed-4b23-9393-36c1f3c65059

### Steps

As in 3rd video:

1. Got to editor instance with Find and Replace enabled, e.g. https://ckeditor.com/docs/ckeditor5/42.0.0/features/find-and-replace.html.
2. Use complex document, e.g. content from https://ckeditor.com/docs/ckeditor5/42.0.0/examples/builds/classic-editor.html.
3. Open Find and Replace dialog.
4. Type 'You' as "Find" and press "Find" button.
5. Navigate to 5th result.
6. Fill replace text and press replace.
7. Press replace again.

### ✔️ Expected result

Replace should cycle over items in order.

### ❌ Actual result

Sometimes highlight jumps to first item.

### ❓ Possible solution

Not sure, I was suspecting complex structure and first item in a block/paragraph but was not able to confirm that with quick testing.

## 📃 Other details

* Browser: Chrome
* OS: MacOS
* First affected CKEditor version: Not sure
* Installed CKEditor plugins: Find and replace

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

rmoustgaard:
This is simple to replicate without tracked changes:
- Go to https://ckeditor.com/ckeditor-5/demo/feature-rich/
- Replace all the text with "This doesn't work. This doesn't work. This doesn't work."
- Open Find and Replace, and Find the word "work".
- Click 'Next result' to skip the first occurrence.
- Replace the second occurrence with another word.
- Result: Cursor jumps back to the first occurrence.

lkszzajac:
The issue is reproducible in blocks that contain 2+ occurrences of queried sequence. If there is only one occurrence, the selection is moved as expected.

- block with 1 occurrence:

[...]

PR #17292, opened 2024-10-21
No longer jump to the first item after replacing phrase

Mati365:

Fix (find-and-replace): Find and replace no longer randomly jumps to the first found item after the replace operation. Closes https://github.com/ckeditor/ckeditor5/issues/16648

---

### Additional information

After removal of marker, we used to reset search to the first changed item.  It's buggy because changed items refer to search results affected by the update algorithm, which used to change found elements, even if they were not replaced. It caused "random" jumps between items after replace. 

The example might be this list of results:

obraz

I replaced one word in a paragraph containing few search results with `2`, `3` and `4` indices. It updated item with `3` index, but it turns out that algorithm updated also items with `2` i `4` IDs. It's counterintuitive, and leaded to the incorrect assumption in the editing flow that we can reset to the first changed node when at least one yellow marker above the search result is removed. In this case, highlight item was pointed to `2` item instead of `4` item. 

I enforced restoring old highlight offset after operations, so it should be fine now.  

#### Before

https://github.com/user-attachments/assets/67bfde88-190d-4940-b5e3-08a76e954f7a

#### After

https://github.com/user-attachments/assets/d1c5cf12-0dfd-4b62-b8d2-9abb65e054ba

### DIFF
--- a/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
+++ b/packages/ckeditor5-find-and-replace/src/findandreplaceediting.ts
@@ -292,7 +292,7 @@ export default class FindAndReplaceEditing extends Plugin {
 			this.state!.highlightedResult = changedSearchResults[ 0 ];
 		} else {
 			// If there is already highlight item then refresh highlight offset after appending new items.
-			this.state!.refreshHighlightOffset();
+			this.state!.refreshHighlightOffset( model );
 		}
 	};
 }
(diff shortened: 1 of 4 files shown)


Mati365:
@charlttsie Can you take a look and perform a QA check on this change?

juliaflejterska:
## Find and replace jumps to incorrect item in multi-root editor
### Steps
1. Open the multi-root editor with multiple roots containing paragraphs with several replaceable items.
2. Use Find and Replace feature to replace the 2nd item in the 1st root.

### Result
After replacing the 2nd item in the 1st root, Find and replace jumps to the 2nd root and replaces the 2nd item:

https://github.com/user-attachments/assets/e24fc631-b3cd-4c69-a1e8-ba7cf5045086

charlttsie:
We've finished testing the PR with @juliaflejterska, and it looks good to us. The initial issue is fixed, and we haven't found any regressions other than the issue with the multi-root editor, which no longer occurs after the recent changes 👍

niegowski (approved):

PR #17299, opened 2024-10-22
Moved the document scroll lock for modal windows from <body> to <html>

### LINKED ISSUE #17298: Lock dialog (modal) page scroll on <html> instead of <body> to make it bulletproof
This is a follow-up of https://github.com/ckeditor/ckeditor5/issues/17093.

I discovered that the fix introduced in https://github.com/ckeditor/ckeditor5/pull/17247 does not work in the project's docs because `body { height: 100vh }` style that ensures `<body>` is never (vertically) scrollable. I propose moving the scroll lock from `<body>` to `<HTML>` to work around this issue.

oleq:

Internal (ui): Moved the document scroll lock for modal windows from `<body>` to `<html>` to make it bulletproof. Closes #17298.

### DIFF
--- a/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/dialog/dialog.css
+++ b/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/dialog/dialog.css
@@ -39,7 +39,7 @@
 	}
 }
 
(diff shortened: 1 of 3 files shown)


f1ames (approved):
LGTM 👍

Issue #17079, opened 2024-09-11
Pasting a large table from Excel causes whole page to scroll away from editor


jbalinski:
## 📝 Provide detailed reproduction steps (if any)

1. Copied a simple table with lots of rows (200) from Excel (Version 2406, Build 17726.20206)
2. Navigated to CKEditor, making sure to scroll down the page so that the top of the page isn't visible.  The editor is completely visible.
3. Pasted the content into CKEditor
4. My browser is scrolled to the top of the page.  CKEditor is no longer visible.

### ✔️ Expected result

Do not scroll to the top of the page when pasting in a table.

### ❌ Actual result

Browser scrolls to the top of the page.

### ❓ Possible solution

NA

## 📃 Other details

NOTE: This can be reproduced on the CKEditor 5 demo page here:
https://ckeditor.com/productivity-pack/paste-from-office-enhanced-demo/

* Browser: Edge 128, Chrome 128, Firefox 130
* OS: Windows 11
* First affected CKEditor version: Not sure, but occurs in 43.1.0
* Installed CKEditor plugins: See plugins used on demo page above.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Mati365:
Reproducible on tables pasted from LibreOffice Calc.

Mati365:
I proposed fix https://github.com/ckeditor/ckeditor5/pull/17314

PR #17314, opened 2024-10-24
No longer scroll to the top of the document if pasted element is larger than scrollable editable.

Mati365:

Fix (utils): No longer scroll to the top of the document if pasted element is larger than scrollable editable. Closes https://github.com/ckeditor/ckeditor5/issues/17079

---

### Additional information

It looks like `scrollWindowToShowRect` assumes that passed target elements entirely fits in the first scrollable parent _viewport rect_, which is not true for large tables (or any other widgets) pasted into editable. These pasted elements often are large enough to have a huge negative `top` position value that is calculated from the top of the viewport. It was causing `window.scrollTo` to fallback scroll to `0` top offset.

This PR enforces scrollable target rect to be the size of the first scrollable ancestor if it overflows it. 

#### Before

https://github.com/user-attachments/assets/b188b75c-0c25-43ce-8f37-fdff2efcdefc

#### After

https://github.com/user-attachments/assets/9a1022b3-a579-46ff-b523-5628622c5682

### DIFF
--- a/packages/ckeditor5-utils/src/dom/scroll.ts
+++ b/packages/ckeditor5-utils/src/dom/scroll.ts
@@ -113,7 +113,21 @@ export function scrollViewportToShowTarget<T extends boolean, U extends IfTrue<T
 
 		// Obtain the rect of the target after it has been scrolled within its ancestors.
 		// It's time to scroll the viewport.
-		const targetRect = getRectRelativeToWindow( target, currentWindow );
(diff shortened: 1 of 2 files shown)


pszczesniak (approved):
LGTM :+1:

Mati365:
Add to QA queue.

charlttsie:
@Mati365 We've finished testing the PR together with @juliaflejterska and can confirm it fixes the issue. We also haven't found any regressions, so looks good 👍

PR #17321, opened 2024-10-24
Optimize body wrapper lookup

bendemboski:
On pages with lots of content and a large number of editors being created at once, the `querySelector` in `attachToDom()` shows up as a non-trivial in performance profiling. The `ck-body-wrapper` element is always added as a direct child of the `body`, so we can speed up the lookup a lot by just iterating over the children of the `body` rather than running `querySelector()`.

### DIFF
--- a/packages/ckeditor5-ui/src/editorui/bodycollection.ts
+++ b/packages/ckeditor5-ui/src/editorui/bodycollection.ts
@@ -44,6 +44,12 @@ export default class BodyCollection extends ViewCollection {
 	private _bodyCollectionContainer?: HTMLElement;
 
 	/**
+	 * The wrapper element that holds all of the
+	 * {@link #_bodyCollectionContainer} elements.
+	 */
+	private static _bodyWrapper?: HTMLElement;
+
+	/**
 	 * Creates a new instance of the {@link module:ui/editorui/bodycollection~BodyCollection}.
 	 *
 	 * @param locale The {@link module:core/editor/editor~Editor editor's locale} instance.
@@ -82,14 +88,12 @@ export default class BodyCollection extends ViewCollection {
 			children: this
 		} ).render() as HTMLElement;
 
-		let wrapper = document.querySelector( '.ck-body-wrapper' );
-
-		if ( !wrapper ) {
-			wrapper = createElement( document, 'div', { class: 'ck-body-wrapper' } );
-			document.body.appendChild( wrapper );
+		if ( !BodyCollection._bodyWrapper ) {
+			BodyCollection._bodyWrapper = createElement( document, 'div', { class: 'ck-body-wrapper' } );
+			document.body.appendChild( BodyCollection._bodyWrapper );
 		}
 
-		wrapper.appendChild( this._bodyCollectionContainer );
+		BodyCollection._bodyWrapper.appendChild( this._bodyCollectionContainer );
 	}
 
 	/**
@@ -103,10 +107,9 @@ export default class BodyCollection extends ViewCollection {
 			this._bodyCollectionContainer.remove();
 		}
 
-		const wrapper = document.querySelector( '.ck-body-wrapper' );
-
-		if ( wrapper && wrapper.childElementCount == 0 ) {
-			wrapper.remove();
+		if ( BodyCollection._bodyWrapper && BodyCollection._bodyWrapper.childElementCount == 0 ) {
+			BodyCollection._bodyWrapper.remove();
+			delete BodyCollection._bodyWrapper;
 		}
 	}
 }
--- a/packages/ckeditor5-ui/src/editorui/bodycollection.ts
+++ b/packages/ckeditor5-ui/src/editorui/bodycollection.ts
@@ -16,20 +16,51 @@ import type View from '../view.js';
 import { createElement, type Locale } from '@ckeditor/ckeditor5-utils';
 
 /**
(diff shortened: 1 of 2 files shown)


scofalik:
Hi @bendemboski. May I ask you what kind of a difference you see when compared before and after? A number or a screenshot from profiler? How many editor instances you create?

Thanks for the suggestion!

bendemboski:
Hi @scofalik.

Here is a screen shot of the profiler:

image

In this profile we are initializing 675 editors on a page where `new XMLSerializer().serializeToString(document).length === 15_223_133` and `document.querySelectorAll('*').length === 104_218`, so lots of editors on a very large page, but you can see that this `querySelector()` is the biggest performance culprit, accounting for 8.7% of the processing (more with the profiling overhead removed).

packages/ckeditor5-ui/src/editorui/bodycollection.ts
```diff
@@ -82,7 +82,7 @@ export default class BodyCollection extends ViewCollection {
 			children: this
 		} ).render() as HTMLElement;
 
-		let wrapper = document.querySelector( '.ck-body-wrapper' );
+		let wrapper = Array.from( document.body.children ).find( el => el.classList.contains( 'ck-body-wrapper' ) );
```
oleq, line 85: We are aware that some integrators move the `.ck-body` to other places in DOM to achieve some specific side-effects, for instance, make all editor balloons and floatables clipped by some arbitrary ancestor with overflow, etc. (they wrap the editor instance and `.ck-body` in a div with `overflow: hidden`).

Knowing this, we should not assume the `.ck-body` wrapper is the child of `<body>`. Doing that may cause some regressions for these integrators (and customers). What we can do though, is keep the static reference to the wrapper somewhere in `BodyCollection` and make it a kind of singleton (since it already acts as one, at least in this regard). This will be as fast as possible, anyway.

bendemboski: @oleq is the latest commit I pushed (rebased) along the lines of what you're thinking?

packages/ckeditor5-ui/src/editorui/bodycollection.ts
```diff
@@ -82,14 +88,12 @@ export default class BodyCollection extends ViewCollection {
 			children: this
 		} ).render() as HTMLElement;
 
-		let wrapper = document.querySelector( '.ck-body-wrapper' );
-
-		if ( !wrapper ) {
-			wrapper = createElement( document, 'div', { class: 'ck-body-wrapper' } );
-			document.body.appendChild( wrapper );
+		if ( !BodyCollection._bodyWrapper || !BodyCollection._bodyWrapper.isConnected ) {
```
bendemboski, line 91: I originally added the `|| !BodyCollection._bodyWrapper.isConnected` to fix the tests before I figured out to add [this line](https://github.com/ckeditor/ckeditor5/pull/17321/files#diff-f600513e49b56cb5c923a1518409d6c2ce166751f21e603ec7b3435e5540575cR29). So this isn't strictly needed, but seems like a good safety mechanism so that if somebody somehow removes the body wrapper from the DOM, `BodyCollection`s aren't forever broken, installing themselves in a detached element.

oleq: This is what I had in mind but I'd keep it as simple as possible. 

With the `isConnected` condition that you proposed, if somebody removed the wrapper from DOM, its reference would be lost when a new editor instance gets created (also another wrapper would be created). If the old wrapper somehow gets back into DOM (e.g. the same way it got lost in the first place), it will remain an orphan forever even if you destroy all editor instances. And there will be two or more wrappers in DOM at a time, which itself is quite unpredictable.

In such a scenario, I'd rather see the UI of all editors broken: at least it would be obvious that something is wrong. If the integrator decides to manage `.ck-body` on their own, it's their job to ensure it sits in DOM when new editor instances are created.

I know you could make this logic even more bulletproof, e.g. store references to all (past) wrappers, iterate over them and so on. But this will be nothing more than a workaround for a problem that should not exist in the first place: there should be one `.ck-body` wrapper at a time that sits in DOM all the time.

---

P.S. We'll need [tests for the changes to accept this PR](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/contributing.html#tests). Thank you!

oleq (changes requested):

packages/ckeditor5-ui/src/editorui/bodycollection.ts
bendemboski: Okay, so remove the `|| !BodyCollection._bodyWrapper.isConnected` -- no problem.

As far as testing, I'm not exactly sure what kind of test to write since the end-result behavior (the presence and singleton-ness of the body wrapper) is already tested. I could add an `expect( wrappers[ 0 ] ).to.equal( BodyCollection._bodyWrapper );` [here](https://github.com/ckeditor/ckeditor5/blob/f70eedd6879dc0a0e6a1183381d494f0b7c0ab57/packages/ckeditor5-ui/tests/editorui/bodycollection.js#L57) and/or [here](https://github.com/ckeditor/ckeditor5/blob/f70eedd6879dc0a0e6a1183381d494f0b7c0ab57/packages/ckeditor5-ui/tests/editorui/bodycollection.js#L112). Does that make sense? Or do you think we need a new test, and if so, what is it testing?

oleq (approved):

Issue #16967, opened 2024-08-23
Uploading image gets deleted when being dragged and dropped


Sanderovich:
## 📝 Provide detailed reproduction steps (if any)

1. Set network throttling to 3G in the developers console. A real scenario would be user having a slow internet connection.
2. Upload an image.
3. Drag and drop the uploading image inside the editor.

Demonstartion of the issue in the [CKEditor 5 demo pages](https://ckeditor.com/ckeditor-5/demo/feature-rich/):

[Screencast from 23-08-24 14:44:17.webm](https://github.com/user-attachments/assets/7725edd7-b90c-402b-b71c-1e2f33f4351a)

### ✔️ Expected result

Image drops and continues uploading.

### ❌ Actual result

Image is deleted.

## 📃 Other details

* Browser: Firefox, Chromium. I was not able to test on Safari

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

kubaklatt:
I can confirm that the reported issue is reproducible and valid.

PR #17327, opened 2024-10-25
Add ability to copy & paste images that are in uploading state.

Mati365:

Fix (image): Copying and pasting images in the uploading state is now possible. Closes https://github.com/ckeditor/ckeditor5/issues/16967

---

### Additional information

It looks like we forgot to dump images that contain `uploadId` to clipboard data. It resulted in blank `img` being exported to the HTML:

obraz

I added a converter that lookups in file repository and checks if there is matching `uploadId` entry containing `data`. If so, it stores `base64` of loaded image in `src` of `img` element, which is already pretty well handled later on. 

Easily reproducible on http://localhost:8125/ckeditor5-image/tests/manual/imageupload.html

#### Before

https://github.com/user-attachments/assets/309b30b9-9b32-48d1-a05e-42289c7f6dc2

#### After

https://github.com/user-attachments/assets/cf1f2317-d8d0-4bbe-9f52-7576eb501a88


arkflpc:
After upload is done, will the copied image have `src` updated to the target or stick to the data-url?

arkflpc (dismissed):
LGTM
@niegowski , wdyt?

juliaflejterska:
@Mati365 We've finished testing the PR with @kubaklatt - it fixes the issue and looks good to us 👍

niegowski (changes requested):
If you trigger `editor.getData()` while an image is not fully uploaded, the editor data would contain the `data:` URL. For the user, it would look like the image is successfully uploaded but the URL would not point to the real asset on the server. 
The proposed solution restarts the upload after the drop, but we should just keep the `uploadId` so the loader can match the given element and continue uploading.

Mati365:
> but we should just keep the uploadId

@niegowski we cannot, as it'll be no longer possible to paste such image to another instance of the editor. The second thing is that uploadId will be no longer valid after successful upload of original image.

Mati365:
@juliaflejterska / @charlttsie We changed a bit the behavior of this fix, so large `data:base64` payload is no longer present in `editor.getData` and it's no longer possible to copy & paste such element from editor A, editor B (it'll work exactly like it is on production right now). Can you check it again?

charlttsie:
Together with @juliaflejterska, we’ve retested the fix and image upload works the same as before the recent changes - copying and pasting images that are in the uploading state works well. We also haven’t found any regressions, so it looks good to us :tada: However, as we didn’t have a lot of time for the retests, let’s pay special attention to this area during the testing phase :slightly_smiling_face:

niegowski:
> copying and pasting images that are in the uploading state works well.

This part surprised me as we know it should not work (especially after the upload is completed).

I'd suggest writing down all scenarios and analyzing what happens and what is expected.

* while an upload is in progress
  * copy and paste
  * cut and paste
  * drag and drop
* start action while an upload is in progress and finish action after an image upload is complete
  * copy and paste
  * cut and paste
  * drag and drop 

The above should also include scenarios with drop/paste outside the editor.

As we checked during the review, not all scenarios behave as expected.

niegowski (changes requested):
As discussed F2F we need to write down all scenarios and make sure those all work as expected.

charlttsie:
### Undoing the pasting of an uploading image paste results in a crash
**Steps**
1. Open http://localhost:8125/ckeditor5-image/tests/manual/imageupload.html
2. Upload an image
3. Copy and paste the image
4. Undo

**Result**
The console shows the following error and there's no way to use the editor:
```
filerepository.ts:522 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'isFulfilled')
    at FileLoader.abort (filerepository.ts:522:1)
    at Document.<anonymous> (imageuploadediting.ts:279:1)
    at Document.fire (emittermixin.ts:240:1)
    at Document._handleChangeBlock (document.ts:370:1)
    at Model._runPendingChanges (model.ts:1112:1)
    at Model.enqueueChange (model.ts:349:1)
    at imageuploadediting.ts:453:1
```

https://github.com/user-attachments/assets/3c7e17c6-8b92-4148-a71b-09a410d2a0c3

juliaflejterska:
### Copying an image while the upload is in progress and pasting after upload results in inconsistent behavior

#### Steps
1. Open [http://localhost:8125/ckeditor5-image/tests/manual/imageupload.html](http://localhost:8125/ckeditor5-image/tests/manual/imageupload.html)
2. Upload an image.
3. Copy the image while it is still in the uploading state.
4. Finish the upload.
5. Paste the image two times.

#### Result
The first paste successfully inserts the image. The second (and any subsequent) paste results in a blank image element.

https://github.com/user-attachments/assets/d697b521-b739-48eb-8ac8-bd45609d26d9

charlttsie:
We've finished the retest with @juliaflejterska and have found the two issues reported above: 
- https://github.com/ckeditor/ckeditor5/pull/17327#issuecomment-2507941984
- https://github.com/ckeditor/ckeditor5/pull/17327#issuecomment-2508080974

The remaining scenarios behave as expected, also in the multi-root editor, and when pasting into a different editor.

Mati365:
@charlttsie Thanks. I pushed some modifications and hope these issues should be no longer present. Can you check again?  

Commits:
1. **Undoing the pasting of an uploading image paste results in a crash** - https://github.com/ckeditor/ckeditor5/pull/17327/commits/269db0db2b26b54eb3f77e34b95d60248b8f35cf
2. **Copying an image while the upload is in progress and pasting after upload results in inconsistent behavior** - https://github.com/ckeditor/ckeditor5/pull/17327/commits/c28e4801e033c390d8ac17e3cc80e0e3df98a0ac

charlttsie:
@Mati365 According to our tests, both issues are fixed. The rest of the scenarios behave as before and we have not found anything new, so the PR looks good to us 🎉

packages/ckeditor5-image/src/imageupload/imageuploadediting.ts
```diff
@@ -217,14 +256,39 @@ export default class ImageUploadEditing extends Plugin {
...
 						if ( !loader ) {
+							// If the loader does not exist, it means that the image was already uploaded
+							// and the loader promise was removed from the registry. In that scenario we need
+							// to restore response object from the internal map.
+							if ( !isInsertedInGraveyard && this._uploadedImages.has( uploadId ) ) {
+								// Fire `uploadComplete` to set proper attributes on the image element.
+								this.fire<ImageUploadCompleteEvent>( 'uploadComplete', {
+									data: this._uploadedImages.get( uploadId )!,
+									imageElement: imageElement as Element
+								} );
```
niegowski, line 267: Maybe we should wrap this in a non-undoable batch as in another place when we fire this event?
```
model.enqueueChange( { isUndoable: false }, writer => {
```

Mati365: I'll check that

Mati365: @niegowski I checked that, works correctly. https://github.com/ckeditor/ckeditor5/pull/17327/commits/1844484b270dbd2c78ba069f65ab54583b559d26

arkflpc (approved):

niegowski (approved):

Issue #17230, opened 2024-10-07
Improved linking experience


Witoso:
> [!NOTE]  
> This initiative is connected to the [Bookmarks work](https://github.com/ckeditor/ckeditor5/issues/1944).

### Goal

Our main goal is to allow users to easily link to the bookmarks created in the document. But we are aware that the linking interface is one of the most used, and it would be good to redesign it thinking also about other (future) use cases.

### Description

Linking experience is the core of the editor. It's important to remember linking to bookmarks is just one use case in the whole context of linking.

When we think about linking, there are a couple of actions that could the user do in modern systems:

- I can add/edit a link.
    - By hand,
    - By selecting from a list,
    - By searching some repository of links.
- I can add/edit the text of the link.
- I can configure the link (what we call decorators).

We decided to spend some time, iterate on the current link UI, and cover most of the use cases above.

### Scope

We are still defining the final scope, for now, the main directions of development for us are:

- Refreshing the UI, adding the possibility to add/edit text.
- Moving decorators to a separate subpage.
    - This gives a space for more advanced inputs in the future (select, text input).
- Providing a Bookmarks list. We are considering a possibility for integrators to define their own lists.

We don’t plan to provide the search of links feature yet. This would have a large scope to cover, and making accessible search is non-trivial. But we definitely consider this in the future.

We are also aware that there are some integrators override/extend the link UI. We will try to figure out what to do with it, and implement minimal breaking changes if possible.

### Progress

1. See comment for [the latest decision regarding the UI](https://github.com/ckeditor/ckeditor5/issues/17230#issuecomment-2476196426).

#### Linking
https://github.com/user-attachments/assets/6bf99d59-a201-499d-a92f-d363fdf82436

#### Bookmarks
https://github.com/user-attachments/assets/f0399957-fbaa-4815-b7ed-ad43ebb34025

#### Decorators
https://github.com/user-attachments/assets/a0bf4f99-3ad7-4804-a953-1507ffba70f9

Witoso:
To those that are following, we slightly changed our approach as we noticed some problems during our prototyping/designing.

The new Link UI was suppose to feature an advanced subpage where users can apply settings like "open in new tab" or "downloadable". Previously, these options were located directly below the link area. However, to enhance the UI and introduce new features like bookmarks, I had to move these decorators to a separate page for more space.

[...]

PR #17347, opened 2024-10-28
Hide the settings button if there are no manual decorators.

filipsobol:

Other (link): Hide the settings button if there are no manual decorators. See #17230


filipsobol:
Failed CI job is caused by issues with CircleCI itself. It can be ignored.

packages/ckeditor5-link/src/linkui.ts
```diff
@@ -253,7 +253,7 @@ export default class LinkUI extends Plugin {
 		formView.saveButtonView.bind( 'isEnabled' ).to( linkCommand, 'isEnabled' );
 
 		// Enable the "Advanced" button only when there are manual decorators.
```
pszczesniak, line 255:
```suggestion
		// Show the "Advanced" button only when there are manual decorators.
```

pszczesniak (approved):
I left just a suggestion to change the comment - other than that :+1:

PR #17359, opened 2024-10-30
Fix incorrect `inheritDoc` comment above editor name.

Mati365:

Internal: Fix incorrect `inheritDoc` comment above editor name.


godai78 (approved):
LGTM

PR #17417, opened 2024-11-07
fix: switched to relative `font-size`

### LINKED ISSUE #17416: Absolute `font-size` doesn't scale on text resizing
The "powered by" text gets defined with an absolute `font-size` or in other words by a pixel value. This should be a relative value like all others as it doesn't scale (upwards) nicely on users setting bigger font-sizes in their environments for accessibility reasons.

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-ui/theme/globals/_poweredby.css#L38

mfranzke:
### Fix(ui): switched to relative `font-size` ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Type: Message. Closes #17416

### DIFF
--- a/packages/ckeditor5-ui/theme/globals/_poweredby.css
+++ b/packages/ckeditor5-ui/theme/globals/_poweredby.css
@@ -35,7 +35,7 @@
 		}
 
 		& .ck-powered-by__label {
-			font-size: 7.5px;
+			font-size: 0.57692308em;
 			letter-spacing: -.2px;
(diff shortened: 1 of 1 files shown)


pszczesniak:
Thank You for your contribution!

We've updated all `px` values to be relative based on value of --ck-font-size-base CSS variable. Also the icon size.

PR #17444, opened 2024-11-12
Use binary search in `insertToPriorityArray`

filipsobol:

Fix (utils): Use binary search in `insertToPriorityArray()` for better performance when handling big tables.

---

Reduces the time needed to render a table in the `tableHuge` test by ~67%.

### DIFF
--- a/packages/ckeditor5-utils/src/inserttopriorityarray.ts
+++ b/packages/ckeditor5-utils/src/inserttopriorityarray.ts
@@ -35,6 +35,7 @@ export interface ObjectWithPriority {
 export default function insertToPriorityArray<T extends ObjectWithPriority>( objects: Array<T>, objectToInsert: T ): void {
 	const priority = priorities.get( objectToInsert.priority );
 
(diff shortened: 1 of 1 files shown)


scofalik (dismissed):

scofalik (approved):

scofalik:
I think this change is safe enough to merge it into upcoming release @Witoso

Witoso:
Good for me :+1: another good reason to bump the version!

niegowski:
Does it affect the order of event listeners on the same priority? Until now, if there were multiple listeners on the same priority, then the order of registration was used. Does this PR change this?

filipsobol:
> Does it affect the order of event listeners on the same priority? Until now, if there were multiple listeners on the same priority, then the order of registration was used. Does this PR change this?

No. You can test the previous and current implementation on the following playground by running the `node index.js` command in the console: https://stackblitz.com/edit/stackblitz-starters-86xsef

PR #17466, opened 2024-11-14
Handle refresh token when editor destroyed during token fetching

### LINKED ISSUE #17462: Token refresh request is being sent after editor is destroyed

1. Create an editor that sends a request for an auth token to Cloud Services
2. (optional) In the user-defined token endpoint, modify the lifespan of the token to some low value, like 20s
3. While using the editor with a valid token, close the endpoint token service to fail the token refresh request.
4. Immediately after the first failure, destroy the editor using `editor.destroy()`

### ✔️ Expected result

Editor is destroyed, no token requests are being made

### ❌ Actual result

- token refresh requests are being made every 5s
- this seems to happen at random
- opening other secondary editors (Revision History, Comment/Suggestion input etc) may influence the behavior

## 📃 Other details

* Browser: n/a
* OS: n/a
* First affected CKEditor version: n/a
* Installed CKEditor plugins: n/a

---

f1ames:

Fix (cloud-services): Handle refresh token when editor destroyed during token fetching. Closes #17462.

---

### Additional information

The issue was caused by the race condition when token fetching request is initialized, editor is destroyed in the meantime and then the request rejects, reinitiating the token refresh logic.

I added `Token._isDestroyed` property to abort all further refresh action when token is already destroyed.

### DIFF
--- a/packages/ckeditor5-cloud-services/src/token/token.ts
+++ b/packages/ckeditor5-cloud-services/src/token/token.ts
@@ -49,6 +49,11 @@ export default class Token extends /* #__PURE__ */ ObservableMixin() {
 	private _tokenRefreshTimeout?: ReturnType<typeof setTimeout>;
 
 	/**
+	 * Flag indicating whether the token has been destroyed.
+	 */
+	private _isDestroyed = false;
+
+	/**
 	 * Creates `Token` instance.
 	 * Method `init` should be called after using the constructor or use `create` method instead.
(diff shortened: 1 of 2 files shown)


f1ames:
@lkszzajac would you be able to find few minutes to test it the same way as before so we have consistent, kind of, check how it behaves?

scofalik (approved):

PR #17469, opened 2024-11-14
Basic support for "Displayed text" input.

pszczesniak:

Feature (link): Add basic support for "Displayed text" input.

---

### Additional information


Mati365 (approved):
I was using it locally for a while during development of other changes and it worked fine.

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -138,10 +138,16 @@ export default class LinkCommand extends Command {
 	 * @fires execute
 	 * @param href Link destination.
 	 * @param manualDecoratorIds The information about manual decorator attributes to be applied or removed upon execution.
+	 * @param displayedText Text of the link.
```
pszczesniak, line 141: We need to extend the API docs for the `displayedText` argument.

Mati365 (approved):

packages/ckeditor5-link/src/linkcommand.ts
pszczesniak: Addressed in 224c646.

PR #17476, opened 2024-11-15
Badges improvements.

### LINKED ISSUE #17416: Absolute `font-size` doesn't scale on text resizing

pszczesniak:

Fix(ui): Badges should now scale dynamically based on value of `--ck-font-size-base` CSS variable. Closes #17416.

---

### Additional information

Also while base font size will be changed - CKEditor logo will be also scaled:

<table><tbody><tr><td>Before</td><td><figure class="image"></figure></td></tr><tr><td>After</td><td><figure class="image"></figure></td></tr></tbody></table>

### DIFF
--- a/packages/ckeditor5-ui/src/editorui/poweredby.ts
+++ b/packages/ckeditor5-ui/src/editorui/poweredby.ts
@@ -16,9 +16,6 @@ import IconView from '../icon/iconview.js';
 
 import poweredByIcon from '../../theme/icons/project-logo.svg';
 
-const ICON_WIDTH = 53;
-const ICON_HEIGHT = 10;
-
 const DEFAULT_LABEL = 'Powered by';
 
 type PoweredByConfig = Required<UiConfig>[ 'poweredBy' ];
@@ -107,21 +104,6 @@ class PoweredByView extends View<HTMLDivElement> {
 		iconView.set( {
 			content: poweredByIcon,
 			isColorInherited: false
-		} );
-
-		// Get the base font size to calculate the icon size in `em` units.
-		const baseFontSize = parseInt( getComputedStyle( document.documentElement ).getPropertyValue( '--ck-font-size-base' ), 10 );
-		// Calculate the icon size in `em` units.
-		const iconWidthInEm = `${ ICON_WIDTH / baseFontSize }em`;
-		const iconHeightInEm = `${ ICON_HEIGHT / baseFontSize }em`;
-
-		iconView.extendTemplate( {
-			attributes: {
-				style: {
-					width: iconWidthInEm,
-					height: iconHeightInEm
-				}
-			}
 		} );
 
 		this.setTemplate( {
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-ui/src/editorui/poweredby.ts
```diff
@@ -109,11 +109,17 @@ class PoweredByView extends View<HTMLDivElement> {
 			isColorInherited: false
 		} );
 
+		// Get the base font size to calculate the icon size in `em` units.
+		const baseFontSize = parseInt( getComputedStyle( document.documentElement ).getPropertyValue( '--ck-font-size-base' ), 10 );
+		// Calculate the icon size in `em` units.
+		const iconWidthInEm = `${ ICON_WIDTH / baseFontSize }em`;
+		const iconHeightInEm = `${ ICON_HEIGHT / baseFontSize }em`;
```
niegowski, line 116: Maybe we could include this in the stylesheet?

packages/ckeditor5-ui/theme/globals/_evaluationbadge.css
```diff
@@ -4,8 +4,9 @@
  */
 
 :root {
-	--ck-evaluation-badge-font-size: 7.5px;
-	--ck-evaluation-badge-line-height: 7.5px;
+	--ck-evaluation-badge-font-size: 0.576923em;
+	--ck-evaluation-badge-line-height: 0.576923em;
+	--ck-evaluation-badge-letter-spacing: -0.015em;
```
niegowski, line 9: Why not relative to `--ck-font-size-base`?

packages/ckeditor5-ui/theme/globals/_evaluationbadge.css
```diff
@@ -4,8 +4,9 @@
...
 	--ck-evaluation-badge-padding-vertical: 2px;
```
gorzelinski, line 10: Maybe we could order those CSS variables similarly to the CSS properties? Like padding at the top and so on.

packages/ckeditor5-ui/theme/globals/_evaluationbadge.css
pszczesniak: We cannot do it like this:

```css
calc(7.5px / var(--ck-font-size-base))
```

Because in CSS calc() division the right side must be a `<number>` therefore unit based values cannot be used in division like this. (`--ck-font-size-base` is `13px`)

packages/ckeditor5-ui/src/editorui/poweredby.ts
pszczesniak: You mean already calculated value in `em` units?

pszczesniak: Addressed in 65028dc.

gorzelinski (approved):

Issue #17171, opened 2024-09-27
iOS: When you add a space after the link, formatting as link is still being added


kubaklatt:
## 📝 Provide detailed reproduction steps (if any)

1. Visit for example https://ckeditor.com/ckeditor-5/demo/mobile-support/
2. Select text and add a link to it
3. Move the selection right after the link, add a space, and try to continue writing the text

### ✔️ Expected result

The link is no longer added to the text that comes after a space

### ❌ Actual result

Formatting as the link is still added after the space

https://github.com/user-attachments/assets/1593167e-60e2-4458-b370-79aed9980d1f

## 📃 Other details

* Browser: Safari
* OS: iOS

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

niegowski:
I suspect it happens because the 2-step caret movement handles the `mousedown` event to clear link attributes from the selection and it probably does not fire on mobile.

https://github.com/ckeditor/ckeditor5/blob/8e662a2060cb596c920656b0e92b7e2b659aa551/packages/ckeditor5-typing/src/twostepcaretmovement.ts#L477

Witoso:
I cannot reproduce this, @kubaklatt could you share the iOS/Safari version.

Mati365:
@kubaklatt Can you test if this still happens in this [PR](https://github.com/ckeditor/ckeditor5/pull/17481)?

Witoso:
Ok, "remove the selection" step needs to always move after the link, it cannot be moved to other piece of text and back to the link.

gorzelinski:
I tried to reproduce the problem, but I couldn't. It seems to be working fine now.

PR #17481, opened 2024-11-18
Fix not working two step caret movement on iOS devices.

Mati365:

Fix (typing): Fix not working two-step caret movement on iOS devices. Closes https://github.com/ckeditor/ckeditor5/issues/17171

---

### Additional information

It looks like `mousedown` events behaves differently on iOS and Android. On iOS, it's called **after** selection change, but on Android it's called immediately after `touchstart` which happens **before** selection change.  I added `touchstart` event observer, to track clicks before selection change occurs.


arkflpc (approved):

niegowski (approved):

charlttsie:
@Mati365 I've tested the PR, and from what I can see, it resolves the issue. I haven't found any regressions either, so it looks good to me 👍

Issue #16395, opened 2024-05-21
Improve tree-shaking of third-party dependencies


filipsobol:
This is a continuation of the task #16292, which you can refer to for more context and results of the initial work.

In it, we were able to improve tree shaking and reduce the number of side effects in `ckeditor5` and `ckeditor5-premium-features` by about 80% by doing about 20% of the work (80/20 rule strikes again). This task focuses on the remaining work needed to make CKEditor bundles as small and performant as possible.

### `ckeditor5`

The following code results in a bundle size of 197 Kib for Rollup/Vite and 365 KiB in esbuild:

```js
import 'ckeditor5';
```

This is the treemap chart generated from the esbuild bundle:

![Treemap chart showing how much each packages added to the bundle](https://github.com/ckeditor/ckeditor5/assets/4145208/8be4856b-b759-4742-944b-e0a00a129f5a)

It shows that 280 out of 364 KiB come from CKEditor code, with almost 99% coming from `ckeditor5-engine` and `ckeditor5-utils`, which are probably needed in all CKEditor projects anyway. 

The remaining 85 KiB come from external dependencies:

* `lodash-es`,
* `color-convert`,
* `color-name`,
* `marked`,
* `turndown`,
* `turndown-plugin-gfm`.

The first three packages are also likely to be used in all CKEditor projects, so there's no point in optimizing them.

However, the same cannot be said for `marked` and `turndown`, as these dependencies are only needed when using the `ckeditor5-markdown-gfm` plugin. These two packages have side effects, so esbuild is not able to remove them from the bundle (Rollup can tree-shake `turndown`). **We should look for ways to remove them from the bundle, either by contributing to them, forking them, or replacing them with alternatives that provide better tree-shaking.**

### `ckeditor5-premium-features`

The following code results in a bundle size of 64 KiB for Rollup/Vite and 74 KiB in esbuild (with all `ckeditor5` packages marked as `external`):

```js
import 'ckeditor5-premium-features';
```

This is the treemap chart generated from the esbuild bundle:

![Treemap chart showing how much each packages added to the bundle](https://github.com/ckeditor/ckeditor5/assets/4145208/f5514981-b0a5-406d-9eeb-95239a016151)

It shows that 64 out of 74 KiB come from external dependencies and only 10 KiB from CKEditor itself (which can be ignored). Of those 64 KiB, 36 KiB are from `socket.io` or its dependencies and 21 KiB are from `protobufjs`.

[...]

PR #17496, opened 2024-11-20
Replace `lodash-es` with `es-toolkit`

filipsobol:

Other: Replace `lodash-es` with `es-toolkit`

---

You can find more information about why we make this change here: https://github.com/ckeditor/ckeditor5/issues/16395#issuecomment-2399644734


packages/ckeditor5-ui/src/template.ts
```diff
@@ -23,7 +23,8 @@ import {
 	type ObservableChangeEvent
 } from '@ckeditor/ckeditor5-utils';
 
-import { isObject, cloneDeepWith } from 'lodash-es';
+import { isObject } from 'es-toolkit/compat';
+import { cloneDeepWith } from 'lodash-es';
```
arkflpc, line 27: Do we intend to keep lodash?

arkflpc (approved):

Witoso:
TBH, this is the first time I heard we keep lodash... I thought it's 1:1 replacement, and having two libs makes less sense for me.

R- from me, as it's increasing dependencies count w/o discussing this openly. And what are potential workarounds for this.

filipsobol:
We've decided to put this on hold until `extend` and `cloneDeepWith` are available in `es-toolkit`.

I'll also contribute to the `isElement` implementation because it cannot be replaced with simple `instanceof HTMLElement` in our case (as I did in the PR).

packages/ckeditor5-image/tests/imagestyle/imagestyleediting.js
```diff
@@ -176,7 +176,7 @@ describe( 'ImageStyleEditing', () => {
 				plugins: [ ImageBlockEditing, ImageStyleEditing ]
 			} );
 
-			sinon.assert.calledOnceWithMatch( normalizationSpy, {
+			expect( normalizationSpy.firstCall.args[ 0 ] ).to.deep.equal( {
```
filipsobol, line 179: This change is related to https://github.com/toss/es-toolkit/issues/953 because `sinon` internally tries to use `toString` which is not available. IMO, current `es-toolkit` behavior makes more sense than in `lodash-es`.

packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
```diff
@@ -301,7 +300,7 @@ export default class ImageResizeButtons extends Plugin {
 					} )
 				};
 
-				const allDropdownValues = map( optionsWithSerializedValues, 'valueWithUnits' );
+				const allDropdownValues = Object.values( optionsWithSerializedValues ).map( option => option.valueWithUnits );
```
filipsobol, line 303: This is the same, so let's just use native APIs.

PR #17509, opened 2024-11-21
Prevent blurry of the inline bookmark icon while zooming in.

pszczesniak:

Internal: Prevent blurry of the inline bookmark icon while zooming in the browser.

---

Rendered with zoom set to 100% and then zoom in to 300%.

Before:
Screenshot 2024-11-21 at 10 18 10

After:
Screenshot 2024-11-21 at 10 17 39

### Additional information

⚠️ `release` branch set as the target ⚠️


pszczesniak:
This PR will be closed because there is more generic solution :point\_right: #17511

Issue #8152, opened 2020-09-24
[Safari] UI Icons are not scalling on page zoom.


niegowski:
## 📝 Provide detailed reproduction steps (if any)

1.  Zoom in/out any editor (<kbd>Cmd</kbd>+<kbd>+</kbd>)

### ✔️ Expected result

![](https://user-images.githubusercontent.com/1232187/94162551-200d3d00-fe87-11ea-8118-a0400cf9a4b8.png)

### ❌ Actual result

![](https://user-images.githubusercontent.com/1232187/94162583-27cce180-fe87-11ea-84c6-e0eedd14d923.png)

## 📃 Other details

*   Browser: Safari 14
*   OS: macOS 10.14.6
*   CKEditor version: very old, even on v15

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Reinmar:
cc @oleq Didn't it work in the past? :O

oleq:
[Known issue](https://bugs.webkit.org/show_bug.cgi?id=199236).

niegowski:
There is a workaround - it scales properly if you set `font-size: initial` on the SVG

oleq:
> There is a workaround - it scales properly if you set `font-size: initial` on the SVG

Which we can't because we use an arbitrary font size for icons: https://github.com/ckeditor/ckeditor5/blob/b5e8ca02b2d6d7ff8f710bd9a9472b46296ef161/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/icon/icon.css#L15

Reinmar:
Decision: We need to choose between resolving this issue on Safari@macOS and scaling icons automatically with the font size. We choose the latter.

oleq:
FYI: SVG icons follow the `--ck-font-size-base` custom property.

PR #17511, opened 2024-11-21
Stop blurry in the UI icons while zooming in the browser.

### LINKED ISSUE #17510: UI icons get bit blurry while scaling up the page zoom.

1. Go to https://ckeditor.com/docs/ckeditor5/latest/examples/builds/classic-editor.html
2. Zoom the page to 250%
3. Take a look at the toolbar.

### ✔️ Expected result

Icons are sharp when zooming in.

Image

### ❌ Actual result

Image

### ❓ Possible solution

Remove `will-change: transform` property https://github.com/ckeditor/ckeditor5/blob/e1e85cb4ac7d99db2b343b208d5d643a5dc47627/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/icon/icon.css#L22

## 📃 Other details

* Browser: …
* OS: …

---

pszczesniak:

Fix(theme-lark):Should prevent of blurry the UI icons while zooming in the page. Closes #17510.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/icon/icon.css
+++ b/packages/ckeditor5-theme-lark/theme/ckeditor5-ui/components/icon/icon.css
@@ -18,9 +18,6 @@
 	/* Inherit cursor style (#5). */
 	cursor: inherit;
 
-	/* This will prevent blurry icons on Firefox. See #340. */
-	will-change: transform;
-
 	& * {
 		/* Inherit cursor style (#5). */
 		cursor: inherit;


charlttsie:
We've tested the PR with @juliaflejterska and the issue looks fixed - it doesn't seem to occur on Chrome, Edge and Firefox anymore. We also haven't found any regressions the fix would introduce 🎉 

However, we were not able to reproduce the issue on `master` and in online docs on mobile devices and desktop Safari. On mobile, the issue doesn't seem to occur at all, while on Safari, we've noticed that a different thing happens when zooming in and the icons are not scaling. We've found out that there was an issue for that reported in the past: https://github.com/ckeditor/ckeditor5/issues/8152 where the conclusion was that it should stay the way it is 🙂

Mati365 (approved):

gorzelinski (approved):

PR #17515, opened 2024-11-21
Search for package dependencies in snippet adapter only in known directories where package can exist

psmyrek:

Internal: Search for package dependencies in snippet adapter only in known directories where package can exist.

---

### Additional information

Closes cksource/ckeditor5-internal#3881.


pomek (approved):
LGTM.

PR #17517, opened 2024-11-21
Improve mapper performance

filipsobol:

Other (engine): Improve performance of the `getModelLength` method.

---

While this looks like a simple change, it introduces significant performance improvements by avoiding recursion and calling `this._viewToModelLengthCallbacks.get()` on nodes without name or when the map is empty (most cases).

| Test name                    	| Improvement in % 	| Improvement in ms 	|
|------------------------------	|-----------------:	|------------------:	|
| `lists`                      	|            -4.9% 	|            -100ms 	|
| `formatting-long-paragraphs` 	|           -15.0% 	|            -730ms 	|
| `inline-styles`              	|           -22.2% 	|           -2310ms 	|
| `wiki`                       	|            -4.0% 	|            -140ms 	|

We could gain even more performance by removing the public `registerViewToModelLength` method and the `this._viewToModelLengthCallbacks` map, since it's only used in legacy lists, which we are deprecating. However, it's likely that it's used by the integrators and may not be possible. This would improve the performance (on top of proposed changes) of `formatting-long-paragraphs` by another 180ms (-4.4%)  and `inline-styles` by 300ms (-3.8%).

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/mapper.ts
+++ b/packages/ckeditor5-engine/src/conversion/mapper.ts
@@ -56,7 +56,7 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
 	 * A map containing callbacks between view element names and functions evaluating length of view elements
 	 * in model.
 	 */
-	private _viewToModelLengthCallbacks: Record<string, ( element: ViewElement ) => number> = {};
+	private _viewToModelLengthCallbacks = new Map<string, ( element: ViewElement ) => number>();
 
 	/**
 	 * Model marker name to view elements mapping.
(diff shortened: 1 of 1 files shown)


scofalik:
Heads up - I was going to rewrite this piece of code in a different way with my Mapper improvements. @niegowski -- do not spent time reviewing it as it may change.

scofalik:
A bit of summary with what happened here recently...

As mentioned above, I rewrote the part of the code where `Mapper#findPositionIn()` uses `getModelLength()`. I rewrote it so that `getModelLength()` is not used at all, as it makes unnecessary steps the way it was used in `findPositionIn()`.

This unfortunately didn't bring any improvements, actually, it looks like it brought small decrease in the performance. I guess that additional side-effects were higher than the gain, and the rest was just statistical error. I've seen around +3% on times on different data sets.

So, I also tested the fix from this PR with my Mapper changes. It looks like the difference is again a statistical error (-1.5% to +2.5%). I believe this is because `getModelLength()` is much much less often called with my Mapper implementation. But still, I believe these changes are positive, so I am for merging them into `master` after all.

PR #17523, opened 2024-11-22
Created `EmojiMentionIntegration` plugin

### LINKED ISSUE #17379: Create a `EmojiMentionIntegration` plugin

The goal is to allow for inserting emoji after typing `:`.

Image

In other words, let's use the Mention plugin as a base.

We can initially use the `emoji-picker-element-data` package as input data. We may change it at some point.

### Definition of Done

* _The name is temporary, and it's subject to change._
	* Confirm the name with @Witoso.
* After typing a colon, we should display a dropdown that allows for inserting emoji.
	* I would require a single character before reviewing the database.
* We need a manual test that allows the feature to be tested.
* We need automated tests for the new code.
* Feature branch and a pull request should target the `#epic/ck/17361` branch.

przemyslaw-zan:

Internal: Created `EmojiMentionIntegration` plugin. Closes #17379.

---

### Additional information


packages/ckeditor5-emoji/src/emojilibraryintegration.ts
```diff
@@ -0,0 +1,79 @@
...
+
+	/**
+	 * Returns feed function for mention config.
+	 *
+	 * @public
+	 */
+	public getQueryEmojiFn( queryLimit: number ): ( searchQuery: string ) => Promise<Array<MentionFeedObjectItem>> {
+		const emojiDatabase = new Database( {
+			dataSource: this.localDataUrl
+		} );
```
pomek, line 58: Can't this be stored as a class property? Do we need to create a new instance whenever we need to ask?

packages/ckeditor5-emoji/src/emojilibraryintegration.ts
```diff
@@ -0,0 +1,79 @@
...
+	public init(): void {
+		this.localDataUrl = URL.createObjectURL(
+			new Blob( [ JSON.stringify( emojiDataRaw ) ] )
+		);
+	}
+
+	/**
+	 * Returns feed function for mention config.
+	 *
+	 * @public
```
pomek, line 53:
```suggestion
```
No need.

packages/ckeditor5-emoji/src/emojilibraryintegration.ts
```diff
@@ -0,0 +1,79 @@
...
+		return true;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public init(): void {
+		this.localDataUrl = URL.createObjectURL(
+			new Blob( [ JSON.stringify( emojiDataRaw ) ] )
+		);
```
pomek, line 47: This should be destroyed in a plugin destructor (`destroy()`).

https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static

packages/ckeditor5-emoji/src/emojimentionintegration.ts
```diff
@@ -0,0 +1,130 @@
...
+		const config = editor.config.get( 'mention.feeds' )! as Array<MentionFeed>;
+		const emojiLibraryIntegration = editor.plugins.get( EmojiLibraryIntegration );
+
+		config.push( {
+			marker: ':',
+			minimumCharacters: 1,
+			dropdownLimit: this.emojiDropdownLimit,
+			itemRenderer: this._customItemRenderer,
+			feed: emojiLibraryIntegration.getQueryEmojiFn( this.emojiDropdownLimit )
+		} );
```
pomek, line 82: We should not mutate objects. Use `const arr = [ ...values, {...} ]` instead.

packages/ckeditor5-emoji/src/emoji.ts
```diff
@@ -37,10 +40,22 @@ export default class Emoji extends Plugin {
 		return true;
 	}
 
+	/**
+	 * @inheritDoc
+	 */
+	constructor( editor: Editor ) {
+		super( editor );
+
+		// TODO: is this needed?
```
pomek, line 49: To resolve.

Witoso:
Product feedback


1.  One letter is not finding emoji
    

    *   either configure the database to return on one letter
    *   or configure mention plugin minimum characters, to not show the dropdown on min 2.
3. I think I would tweak the number down on the list to 5 max + See more...
4. On See more... the dialog with all emoji will open, but make sure that your implementation will delete the chars up to `:` before that.

Nice work!

packages/ckeditor5-emoji/src/emojilibraryintegration.ts
przemyslaw-zan: This is a function factory, therefore it is only created once.

Witoso:
Also, not sure if possible, in this moment:



To add some italic _No results found_ above the See more... But it may be hard to achieve with Mention. It's for sure nice-to-have.

filipsobol:
We've decided to add `moduleResolution: bundler` to `tsconfig.release.json` too, as it seems to resolve the issue.

If this makes the CI green, then it should be fine.

packages/ckeditor5-emoji/src/insertemojicommand.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+/**
+ * @module emoji/insertemojicommand
+ */
+
+import type { Range } from 'ckeditor5/src/engine.js';
+import { Command } from 'ckeditor5/src/core.js';
+
+export default class InsertEmojiCommand extends Command {
```
martnpaneq, line 13: I think we should not create this `InsertEmojiCommand`, because it does not do what it says. It just inserts plain text, and it could be missleading for the end user. Besides in this code it is already used in a wrong way to insert empty space.

packages/ckeditor5-emoji/src/emojilibraryintegration.ts
```diff
@@ -0,0 +1,88 @@
...
+
+						return {
+							id: formatEmojiId( id ),
+							text: emoji.unicode
+						};
+					} );
+				} );
+
+			return [
+				...processedQuery.filter( ( item, index ) => index < queryLimit - 1 ),
```
martnpaneq, line 83: * to consider using slice, because it could be more performant.

martnpaneq (approved):

Issue #17378, opened 2024-11-04
Polish the emoji plugin UI


pomek:
## Provide a description of the task

Before we start processing this issue, we need to identify areas for enhancement:

* In #17377, we created a plugin that offers dropdown. Unfortunately, we miss an icon (toolbar item) for it.
* Category size. Currently, it looks like emojis are bigger than categories.
	![Image](https://github.com/user-attachments/assets/5ce671b6-14f8-4b84-aba9-cf2bccf9e992)
* Labels and placeholders.
		E.g., ![Image](https://github.com/user-attachments/assets/a22b53c9-e2df-41b3-b3fd-e318af0d210e)
* The empty state view. Is it ok?
		![Image](https://github.com/user-attachments/assets/4330cd36-3acc-4353-8b41-975a8c4d08b1)

pomek:
Below is a list of things that we need to fix:

- **The emojis are not centered.** Each one is aligned to the left edge.
	![Image](https://github.com/user-attachments/assets/1a4d6a1d-75ce-45bb-8032-f58dfcd4fe0e)
	![Image](https://github.com/user-attachments/assets/adad751b-0611-40cb-b05a-69772b9ea837)
- Rows and columns. **The category row is misaligned to the other columns**, making the layout slightly disorganized.
	![Image](https://github.com/user-attachments/assets/75fcc997-a21e-40c6-81d4-f00a298f90ec)
- **Can we make the category emojis the same size as those below?** It looks odd that they are smaller.
	![Image](https://github.com/user-attachments/assets/7720db14-c4d1-4eea-ab32-0dc3fed68427)
- **The category buttons are not square** and are unusually large.
	![Image](https://github.com/user-attachments/assets/4a2cd9c5-d137-4abd-8fad-82834582306a)
	![Image](https://github.com/user-attachments/assets/60d52ed7-8d9e-4e6e-b108-372e0bb8cc6a)
- Skin-tone button. **The padding between the input field and the skin-tone button** (the purple one) is too large.
	![Image](https://github.com/user-attachments/assets/2a41af4a-fbaf-4bfa-bbe0-926d0556c01a)
- Balloon height. **Currently, the balloon is quite long**, even with scrolling. We suggest making it slightly shorter - perhaps 8-9 rows of emojis instead of 11 should be sufficient? On the screen there is an example with 8 rows.
	![Image](https://github.com/user-attachments/assets/8053891e-bc8a-43b7-a0f7-e6a808d89f44)
[...]

PR #17537, opened 2024-11-26
Created `EmojiPicker` plugin

### LINKED ISSUE #17377: Create a `EmojiUi` plugin

After clicking upon a UI component, we want to display a dialog that renders the emoji picker. After selecting one, it should be inserted into the content.

### Definition of Done

* An icon can be implemented later (#17378). At the beginning we can use a text label as a UI component.
* We need a manual test that allows the feature to be tested.
* We need automated tests for the new code.
* Feature branch and a pull request should target the `#epic/ck/17361` branch.

przemyslaw-zan:

Internal: Created `EmojiPicker` plugin. Closes #17377.

---

### Additional information


przemyslaw-zan:
`Content Security Policy` is still broken in manual tests, hence failed CI. I was unable to silence the error. The plugin works though.

packages/ckeditor5-emoji/src/emojipicker.ts
```diff
@@ -0,0 +1,193 @@
...
+	 */
+	public init(): void {
+		this._balloon = this.editor.plugins.get( ContextualBalloon );
+		this._emojiDialog = this._createEmojiDialog();
+
+		this.editor.ui.componentFactory.add( 'emoji', () => {
+			const button = new ButtonView();
+
+			button.set( {
+				label: this.editor.t( 'Emoji' ),
```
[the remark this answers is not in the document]
martnpaneq: And also what about the styles of the input. They should match the styles of the other inputs.

packages/ckeditor5-emoji/src/emojipicker.ts
```diff
@@ -0,0 +1,193 @@
...
+class EmojiDialog extends View {
+	public emojiElement: HTMLElement;
+
+	constructor( editor: Editor ) {
+		super( editor.locale );
+
+		this.emojiElement = new Picker( {
+			dataSource: editor.plugins.get( EmojiLibraryIntegration ).localDataUrl
+		} );
+		this.emojiElement.style.height = '400px';
```
martnpaneq, line 169: I'm not sure if hardcoding '400px' is the correct practise in CKEditor 5. We should see how it is done in other plugins.

packages/ckeditor5-emoji/src/emojipicker.ts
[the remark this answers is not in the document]
Witoso: I reached out to @dagdzi @dwandeljanicka to have some design check.

packages/ckeditor5-emoji/src/emojipicker.ts
```diff
@@ -0,0 +1,191 @@
...
+};
+
+/**
+ * The emoji picker plugin.
+ *
+ * Introduces the `'emoji'` dropdown.
+ */
+export default class EmojiPicker extends Plugin {
+	private _balloon!: ContextualBalloon;
+	declare private _emojiDialog: EmojiDialog;
```
Witoso, line 35: Dialog is something different (opening a separate "window"). I would call it `EmojiBalloon` or `EmojiView`?

packages/ckeditor5-emoji/src/emojipicker.ts
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public init(): void {
+		this._balloon = this.editor.plugins.get( ContextualBalloon );
+		this._emojiDialog = this._createEmojiDialog();
+
+		this.editor.ui.componentFactory.add( 'emoji', () => {
```
pomek, line 65: We need a dedicated integration with the menu bar, right? Example: https://github.com/ckeditor/ckeditor5/blob/db95303a82cb0aa649bc2d16920b71f531c1fc66/packages/ckeditor5-basic-styles/src/bold/boldui.ts#L51

pomek: It would also require a change in the menu bar configuration itself.

packages/ckeditor5-emoji/src/emojipicker.ts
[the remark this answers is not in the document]
pomek: Icon is a separate issue to solve (#17378).

packages/ckeditor5-emoji/tests/manual/emoji.js
```diff
@@ -6,17 +6,27 @@
...
-		toolbar: [ 'undo', 'redo' ]
+		plugins: [
+			Emoji,
+			Clipboard,
+			Enter,
+			Mention,
+			Paragraph,
+			Typing,
+			Undo
+		],
```
pomek, line 28:
```suggestion
		plugins: [
			Emoji,
			Mention,
			Paragraph
		],
```

We can reduce an array of plugins by loading the `Essentials` plugin.

https://github.com/ckeditor/ckeditor5/blob/db95303a82cb0aa649bc2d16920b71f531c1fc66/packages/ckeditor5-essentials/src/essentials.ts#L41

This would reduce entries in pkg.json too.

packages/ckeditor5-emoji/tests/manual/emoji.md
```diff
@@ -1,5 +1,3 @@
-## foo
+## Emoji plugin
 
-### bar
-
-1. baz
+This test allows for testing of both `EmojiMention` and `EmojiPicker` plugins.
```
pomek, line 3: We need to explain what to do to test these.

* For `EmojiMention`, a user needs to type `:` and then two letters. Also, opening a picker directly from the suggestions dropdown is possible.
* For `EmojiPicker,` a user can open a view that allows them to pick an emoji. Also, it would be good to check for integration with the menu bar feature.

packages/ckeditor5-emoji/src/emojipicker.ts
pomek: Could we use CSS variables here? This way, an integrator can align the height without hacking the plugin code.

przemyslaw-zan:
☝️ Fixed

packages/ckeditor5-emoji/src/emojimention.ts
```diff
@@ -60,11 +61,11 @@ export default class EmojiMentionIntegration extends Plugin {
 	constructor( editor: Editor ) {
 		super( editor );
 
-		this.emojiDropdownLimit = editor.config.get( 'emoji.dropdownLimit' ) || DEFAULT_DROPDOWN_LIMIT;
-		this.mentionMarker = editor.config.get( 'emoji.marker' ) || DEFAULT_MENTION_MARKER;
+		this._emojiDropdownLimit = editor.config.get( 'emoji.dropdownLimit' ) || DEFAULT_DROPDOWN_LIMIT;
+		this._mentionMarker = editor.config.get( 'emoji.marker' ) || DEFAULT_MENTION_MARKER;
```
pomek, line 65: By default, `editor.config.get( 'emoji' )` returns `undefined`. Even if the Emoji plugin is loaded. Hence, I propose using `editor.config.define( 'emoji', { /* ... */ } )` to set up initial values.

packages/ckeditor5-emoji/src/emojimention.ts
```diff
@@ -141,17 +142,24 @@ export default class EmojiMentionIntegration extends Plugin {
 			}
 
 			let textToInsert = eventData.mention.text;
+			let postInsertionCallback;
```
pomek, line 145:
```suggestion
			let shouldShowEmojiView = false;
```

Then, later,

```ts
if ( eventData.mention.id === getShowAllEmojiId() ) {
	shouldShowEmojiView  = true;
}

// ...

if ( shouldShowEmojiView ) {
	this.editor.plugins.get( EmojiPicker ).showUI( eventData.mention.text );
}
```

pomek (commented):
I would fix two things:

* the previous utils are now a part of the plugin. I would keep them private (internals) in the plugin. So, there is no method but a function in the same file. This way, we do not have to focus on documenting what they do because API docs will not render them.
* sort methods by the visibility level - public ones should be defined before protected/private.

Other than that, LGTM.

packages/ckeditor5-emoji/src/emojimention.ts
```diff
@@ -0,0 +1,212 @@
...
+
+const EMOJI_PREFIX = 'emoji';
+const SHOW_ALL_EMOJI = '__SHOW_ALL_EMOJI__';
+
+/**
+ * The emoji mention plugin.
+ *
+ * Introduces the autocomplete of emojis while typing.
+ */
+export default class EmojiMention extends Plugin {
```
pomek, line 29: See the order of methods in https://github.com/ckeditor/ckeditor5/blob/71c7bf3d62962b0899706dad787764319c4d5193/packages/ckeditor5-autosave/src/autosave.ts#L55.

The static getters are defined before constructor.

packages/ckeditor5-emoji/src/emojipicker.ts
```diff
@@ -0,0 +1,204 @@
...
+
+		return {
+			target
+		};
+	}
+
+	/**
+	 * Creates a (toolbar or menu bar) button for the emoji picker feature.
+	 */
+	private _createButton<T extends typeof ButtonView | typeof MenuBarMenuListItemButtonView>( ButtonClass: T ): InstanceType<T> {
```
pomek, line 111: This could be an internal function that receives _this_ as a plugin instance. No need to expose this function as a method in class.

packages/ckeditor5-emoji/src/emojimention.ts
```diff
@@ -33,6 +33,27 @@ export default class EmojiMention extends Plugin {
...
+	/**
+		 * @inheritDoc
+		 */
+	public static get pluginName() {
+		return 'EmojiMention' as const;
+	}
+
+	/**
+		 * @inheritDoc
+		 */
```
pomek, line 52:
```suggestion
	/**
	 * @inheritDoc
	 */
	public static get pluginName() {
		return 'EmojiMention' as const;
	}

	/**
	 * @inheritDoc
	 */
```

pomek (approved):
LGTM.

PR #17548, opened 2024-11-27
Use is method instead of instanceof

filipsobol:

Other (engine): Use `.is()` instead of `instanceof` where possible for better performance.

---

This introduces minor performance improvements of 1% to 5% across all performance tests.


scofalik (approved):

scofalik:
I propose squashing commits with reverts.

filipsobol:
The PR was squashed + merged.

PR #17557, opened 2024-11-28
Use iterator instead of generator in ListWalker for better performance.

filipsobol:

Other (list): Use iterator instead of generator in ListWalker for better performance.

---

The generator introduces some unnecessary overhead. This PR makes the `lists` performance demo 9% faster.

### DIFF
--- a/packages/ckeditor5-list/src/list/utils/listwalker.ts
+++ b/packages/ckeditor5-list/src/list/utils/listwalker.ts
@@ -224,15 +224,21 @@ export class SiblingListBlocksIterator implements IterableIterator<ListIteratorV
 		let previousNodeInList: ListElement | null = null;
 
 		if ( this.previous ) {
-			nodeIndent = this.node.getAttribute( 'listIndent' ) as number;
+			nodeIndent = this.node.getAttribute( 'listIndent' );
 			const previousNodeIndent = this.previousNodeIndent!;
 
+			// Let's find previous node for the same indent.
+			// We're going to need that when we get back to previous indent.
 			if ( nodeIndent > previousNodeIndent ) {
 				this.previousNodesByIndent[ previousNodeIndent ] = this.previous;
-			} else if ( nodeIndent < previousNodeIndent ) {
+			}
+			// Restore the one for given indent.
+			else if ( nodeIndent < previousNodeIndent ) {
 				previousNodeInList = this.previousNodesByIndent[ nodeIndent ];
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-list/src/list/utils/listwalker.ts
```diff
@@ -191,47 +191,69 @@ export default class ListWalker {
...
-): IterableIterator<ListIteratorValue> {
-	const isForward = direction == 'forward';
-	const previousNodesByIndent: Array<ListElement> = []; // Last seen nodes of lower indented lists.
-	let previous = null;
+export class SiblingListBlocksIterator implements IterableIterator<ListIteratorValue> {
+	private node: Node | null;
+	private isForward: boolean;
+	private previousNodesByIndent: Array<ListElement> = [];
+	private previous: ListElement | null = null;
+	private previousNodeIndent: number | null = null;
```
niegowski, line 200:
```suggestion
	private _node: Node | null;
	private _isForward: boolean;
	private _previousNodesByIndent: Array<ListElement> = [];
	private _previous: ListElement | null = null;
	private _previousNodeIndent: number | null = null;
```

filipsobol: I forget about this every time 😮‍💨

Fixed

packages/ckeditor5-list/src/list/utils/listwalker.ts
```diff
@@ -191,47 +191,69 @@ export default class ListWalker {
...
 
-		yield { node, previous, previousNodeInList };
+		const value = {
+			node: this._node,
+			previous: this._previous,
+			previousNodeInList
+		};
+
+		this._previous = this._node as ListElement;
+		this._previousNodeIndent = nodeIndent;
```
scofalik, line 253: It's not easy to compare these changes so forgive me if I am making a mistake somewhere here, but.

After first iteration `nodeIndent` is `null`, so `_previousNodeIndent` will be also `null`. In second iteration, in line 228, `previousNodeIndent` will be set to `null`. I think this is different than before.

scofalik: Other than it looks to me that the iteration logic is the same.

filipsobol: You're right, fixed.

scofalik (approved):

PR #17562, opened 2024-11-29
Do not show back button if balloon was not opened using another balloon.

Mati365:

Internal (bookmark, link, image): Do not show back button if balloon was not opened using another balloon.
Fix (image): No longer crash editor when using image resize buttons in toolbar. 

### Additional information

Take a look at back button arrows.

#### Before

https://github.com/user-attachments/assets/f5964832-3b2f-4fce-9078-3c61ee0355b8

#### After

https://github.com/user-attachments/assets/850dca9c-f012-4348-94d8-edd6b59caba8

### DIFF
--- a/packages/ckeditor5-bookmark/src/bookmarkui.ts
+++ b/packages/ckeditor5-bookmark/src/bookmarkui.ts
@@ -281,9 +281,7 @@ export default class BookmarkUI extends Plugin {
 			button.bind( 'isEnabled' ).to( updateBookmarkCommand );
 
 			this.listenTo( button, 'execute', () => {
-				this._showFormView( {
-					showBackButton: true
-				} );
+				this._showFormView();
 			} );
 
 			return button;
@@ -328,9 +326,7 @@ export default class BookmarkUI extends Plugin {
 		} );
 
 		// Execute the command.
-		this.listenTo( view, 'execute', () => this._showFormView( {
-			showBackButton: updateCommand.isEnabled
-		} ) );
+		this.listenTo( view, 'execute', () => this._showFormView() );
 
 		view.bind( 'isEnabled' ).toMany(
 			[ insertCommand, updateCommand ],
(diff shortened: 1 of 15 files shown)


Witoso:
This should be implemented for all balloons

*   text alternative
*   custom image resize,

as someone can add them to the main toolbar, not balloon.

To test: behavior in the block toolbar / menu bar. (no back)

Mati365:
Hm, I implemented that this way, but I thought that it's impossible to set custom size using menubar / or toolbar for such features.

Mati365:
@Witoso I checked both places, and I was unable to open image resize / image text alternative widget balloons without having image selected. It's crucial because selection of image triggers showing the default widget toolbar, which is being hidden after stacking another balloon on top of it (and after pressing back in these stacked balloons, it's shown again).

packages/ckeditor5-bookmark/src/bookmarkui.ts
```diff
@@ -281,7 +281,7 @@ export default class BookmarkUI extends Plugin {
 			button.bind( 'isEnabled' ).to( updateBookmarkCommand );
 
 			this.listenTo( button, 'execute', () => {
-				this._showFormView();
+				this._showFormView( true );
```
pszczesniak, line 284: Im wondering about how to make it more descriptive. Cause this is not telling too much.

Maybe creta a descriptive variable, like:

```
const showBackButton = true;

this._showFormView( showBackButton );
```

Or maybe use object as a param of this/those functions?

```
_showFormView( { showBackButton: boolean } )
```

:thinking:

packages/ckeditor5-link/src/linkui.ts
```diff
@@ -718,11 +718,13 @@ export default class LinkUI extends Plugin {
 	/**
 	 * Adds the {@link #formView} to the {@link #_balloon}.
 	 */
-	private _addFormView(): void {
+	private _addFormView( showBackButton: boolean = true ): void {
```
pszczesniak, line 721: Default value :+1:

pszczesniak (commented):
One thing to consider.

packages/ckeditor5-bookmark/src/bookmarkui.ts
Mati365: Good point, imho, object approach is much more intuitive. I'll ref it this way.

Mati365: @pszczesniak done https://github.com/ckeditor/ckeditor5/pull/17562/commits/7b2541224a5060e753aecd276f2c96a1f3b67d8d

niegowski (commented):
There is a strange behavior when you put caret in a link (link toolbar appears) and you hit Cmd+K. There is no back button but it should be there.

Mati365:
@niegowski I simplified logic, it should work fine after modification.

niegowski (approved):

PR #17564, opened 2024-11-29
Skip already visited list elements during reconversion and post fixing

### LINKED ISSUE #17625: Improve the performance of lists
When investigating performance issues, we noticed that a lot of time in the `lists` performance test is spent in `findAndAddListHeadToMap` used in post-fixers. A lot of that work can be avoided, by skipping the list elements that already were visited.

filipsobol:

Other (list): Skip already visited list elements during reconversion and post fixing for better performance. Fixes #17625.

---

This change makes the `lists` performance demo ~18% faster.

### DIFF
--- a/packages/ckeditor5-list/src/list/converters.ts
+++ b/packages/ckeditor5-list/src/list/converters.ts
@@ -138,32 +138,33 @@ export function reconvertItemsOnDataChange(
 		const changes = model.document.differ.getChanges();
 		const itemsToRefresh = [];
 		const itemToListHead = new Map<ListElement, ListElement>();
+		const visited = new Set<Element>();
 		const changedItems = new Set<Node>();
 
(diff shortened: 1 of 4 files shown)


scofalik:
AFAICS, this change seems correct.

However, after your change, I think that `listItemToHead` could simply become a `Set`. We only use the values, and the only place where key is used is here:

```javascript
if ( itemToListHead.has( listHead ) ) {
	return;
}
```

But it seems to me it would be enough to add make sure that `previousNode` is always added to `visited` too.

IDK if this will have any impact on performance, though. But the fact it was a `Map` made it a bit more difficult to understand what is going on when I was debugging this part of the code.

WDYT?

filipsobol:
Done :heavy_check_mark: There's very little to no performance difference.

scofalik (approved):

PR #17566, opened 2024-11-29
The insertText event and related beforeinput events should not be canceled

### LINKED ISSUE #13428: Mac Text Replacement Shortcuts Don't Work In Safari

1. On MacOS create a text replacement shortcut (System Preferences > Keyboard > Text). In this example I created one where the text is "omw" and the replacement is "On my way!".
Screen Shot 2023-02-07 at 5 15 58 PM

2. Open the CKEditor Classic editor demo page in Safari or Firefox: https://ckeditor.com/docs/ckeditor5/latest/examples/builds/classic-editor.html
3. Type "omw"

### ✔️ Expected result

The text is highlighted and the replacement text appears:

Screen Shot 2023-02-07 at 5 14 11 PM

Upon pressing a key the replacement text is inserted instead of the original text

Screen Shot 2023-02-07 at 5 15 03 PM

### ❌ Actual result

No replacement text appears:

Screen Shot 2023-02-07 at 5 14 56 PM

## 📃 Other details

* Browser: Safari (Only)
* OS: macOS Monterrey 12.6.1
* First affected CKEditor version: We can reproduce the issue on the latest version, though I'm guessing this might be related to the beforeInput migration.
* Installed CKEditor plugins: Classic editor.

---

### LINKED ISSUE #15436: No typing echo from VoiceOver for Mac when writing

1. Turn on VoiceOver, press VO-V and set Typing Echo to "Characters and Words" (the default)
2. In Safari for Mac, open a page with CKEditor
3. Move focus to the editing area and type content

### ✔️ Expected result

VoiceOver repeats each letter as it's typed, words are spoken once they're finished.

### ❌ Actual result

VoiceOver says nothing while typing, not letters and not words.

### ❓ Possible solution

Not a solution but a crude workaround: VoiceOver will read text that's already in the editor, one can type then select all the text in the editor to hear it.

## 📃 Other details

* Browser: Safari 17.1
* OS: macOS 13.6.2 (Ventura)
* First affected CKEditor version: Unknown, have reproduced the problem locally in 38.1.0, in [CKEditor5 Predefined builds](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/multi-root-editor.html), but not in [CKEditor 4 demo](https://ckeditor.com/ckeditor-4/demo/).

I can also reproduce the problem using Safari Technology Preview Release 183 and with Safari 16.5 on macOS 12.6 but haven't reproduced it with any browser that isn't Safari for Mac. I cannot reproduce the problem using VoiceOver with Safari on iOS 17.1 (with virtual or Bluetooth keyboard), VoiceOver with Firefox or Chrome for Mac, or NVDA with Firefox on Windows 10.

niegowski:

Fix (typing): Typing in track changes mode does not reverse typed letters in Safari.

Fix (typing): The Mac VoiceOver echoes typed characters when typing. Closes #15436. 

Fix (typing): Mac text replacement shortcuts replace text in Safari. Closes #13428.

Other (typing): Improved typing performance in large documents.

MINOR BREAKING CHANGE: The `beforeinput` DOM events are not default-prevented in plain text typing scenarios. The engine waits for DOM mutations and applies changes to the model after that.

---

### Additional information

Closes https://github.com/cksource/ckeditor5-commercial/issues/6835.

Before this PR we default-prevented all typing related `beforeinput` events (including enter and delete handling). With an exception to the composition events which are not cancelable. In this PR we stopped default-preventing the plain typing events with some exceptions:
* the delete/enter related events are still stopped
* when the event target range is partly covering an inline filler we prevent default action and adjust the target range
* when the target range is the fake selection, we prevent modifying the content of the fake selection container
* when table cells are selected (from outside) we prevent it

In cases when we prevent browser default behavior, this is handled as previously so beforeinput event is immediately applied to the model and downcast conversion + dom converter are applying changes to the DOM.
In cases where we let browser apply changes to the DOM, we just queue the data from the before input and apply those changes to the model as soon as we detect DOM mutations made by the browser.


Reinmar:
Note: Add to the commit message a separate item about the performance improvements in typing in huge content.

mmotyczynska (approved):

arkflpc (approved):

PR #17582, opened 2024-12-03
Improve performance of `getSelectedBlocks`

### LINKED ISSUE #17629: Improve performance of `getSelectedBlocks`
One of the features of the `getSelectedBlocks` method is that it doesn't return children of the block elements. However, we currently still iterate over them, which adds unnecessary overhead.

We should update the implementation so that we jump over these child elements, straight to the end of the block element.

filipsobol:

Other (engine): Improve performance of `getSelectedBlocks` when selection contains only one block element. Closes #17629.

---

This greatly improves how long it takes to make a selection over a big table.

---

Use `Squash and merge`, as it's a fairly simply change with too many commits.

### DIFF
--- a/packages/ckeditor5-engine/src/model/selection.ts
+++ b/packages/ckeditor5-engine/src/model/selection.ts
@@ -17,7 +17,7 @@ import type DocumentSelection from './documentselection.js';
 import type Element from './element.js';
 import type Item from './item.js';
 
-import { CKEditorError, EmitterMixin, first, isIterable } from '@ckeditor/ckeditor5-utils';
+import { CKEditorError, EmitterMixin, isIterable } from '@ckeditor/ckeditor5-utils';
 
 /**
  * Selection is a set of {@link module:engine/model/range~Range ranges}. It has a direction specified by its
@@ -667,17 +667,11 @@ export default class Selection extends /* #__PURE__ */ EmitterMixin( TypeCheckab
 				yield startBlock as any;
 			}
 
-			const firstElement = first( range.getWalker() );
-			const lastElement = first( range.getWalker( { direction: 'backward' } ) );
+			const containedElement = range.getContainedElement();
 
-			if (
-				firstElement &&
-				lastElement &&
-				firstElement.item === lastElement.item &&
-				lastElement.type == 'elementEnd' &&
-				lastElement.item.root.document!.model.schema.isBlock( lastElement.item )
-			) {
-				yield lastElement.item as any;
+			if ( containedElement ) {
+				// Fast path for selection that contains only one element (e.g. big table)
+				yield containedElement;
 			} else {
 				for ( const value of range.getWalker() ) {
 					const block = value.item;
--- a/packages/ckeditor5-engine/src/model/selection.ts
+++ b/packages/ckeditor5-engine/src/model/selection.ts
@@ -667,20 +667,24 @@ export default class Selection extends /* #__PURE__ */ EmitterMixin( TypeCheckab
 				yield startBlock as any;
 			}
 
-			const containedElement = range.getContainedElement();
+			const treewalker = range.getWalker();
 
-			if (
-				containedElement &&
-				containedElement.root.document!.model.schema.isBlock( containedElement )
-			) {
-				// Fast path for selection that contains only one **block** element (e.g. big table).
-				yield containedElement;
-			} else {
-				for ( const value of range.getWalker() ) {
-					const block = value.item;
+			for ( const value of treewalker ) {
+				const block = value.item;
 
-					if ( value.type == 'elementEnd' && isUnvisitedTopBlock( block as any, visited, range ) ) {
-						yield block as Element;
+				if ( value.type == 'elementEnd' && isUnvisitedTopBlock( block as any, visited, range ) ) {
+					yield block as Element;
+				} else if ( block.is( 'model:element' ) && block.root.document!.model.schema.isBlock( block ) ) {
+					if ( value.type == 'elementEnd' ) {
+						treewalker.jumpTo( value.nextPosition );
+					} else {
+						const position = treewalker.position.clone();
+
+						position.offset = block.maxOffset;
+
+						if ( range.containsPosition( position ) ) {
+							treewalker.jumpTo( position );
+						}
 					}
 				}
 			}
--- a/packages/ckeditor5-engine/src/model/selection.ts
+++ b/packages/ckeditor5-engine/src/model/selection.ts
@@ -674,18 +674,14 @@ export default class Selection extends /* #__PURE__ */ EmitterMixin( TypeCheckab
 
 				if ( value.type == 'elementEnd' && isUnvisitedTopBlock( block as any, visited, range ) ) {
 					yield block as Element;
-				} else if ( block.is( 'model:element' ) && block.root.document!.model.schema.isBlock( block ) ) {
-					if ( value.type == 'elementEnd' ) {
-						treewalker.jumpTo( value.nextPosition );
-					} else {
-						const position = treewalker.position.clone();
-
-						position.offset = block.maxOffset;
-
-						if ( range.containsPosition( position ) ) {
-							treewalker.jumpTo( position );
-						}
-					}
+				}
+				// If element is block, we can skip its children and jump to the end of it.
+				else if (
+					value.type == 'elementStart' &&
+					block.is( 'model:element' ) &&
+					block.root.document!.model.schema.isBlock( block )
+				) {
+					treewalker.jumpTo( Position._createAt( block, 'end' ) );
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


scofalik:
Looks good and is a simple and safe fix.

However, since it only targets a specific scenario (one element in selection), I am afraid we may need to revisit it.

The problem with current implementation is that it only looks at `elementEnd` while it seems it could look at `elementStart` as well. If the treewalker traverses over `elementStart` and this is a top-most unvisited block element, we should "jump" over it, as we never return blocks inside blocks:

> ```
>  * **Note:** `getSelectedBlocks()` returns blocks that are nested in other non-block elements
>  * but will not return blocks nested in other blocks.
> ```

Of course, we should respect this special case that is mentioned later:

> ```
> * **Special case**: Selection ignores first and/or last blocks if nothing (from user perspective) is selected in them.
> ```

Maybe it's even possible to limit the number of `isUnvisitedTopBlock()` calls with some smart observations, but what I proposed seems like a safe change as well. One difficulty is that, AFAIR, it is not possible currently to change `TreeWalker#position`. There is `.skip()` but it simply traverses the tree, so that's exactly what we want to avoid. We would need something like `.jumpToPosition( position )` and this should basically reset the treewalker and set it to that position. Alternatively, we could create a new treewalker and overwrite the current one.

packages/ckeditor5-engine/src/model/selection.ts
```diff
@@ -667,11 +667,25 @@ export default class Selection extends /* #__PURE__ */ EmitterMixin( TypeCheckab
...
 					yield block as Element;
+				} else if ( block.is( 'model:element' ) && block.root.document!.model.schema.isBlock( block ) ) {
+					if ( value.type == 'elementEnd' ) {
+						treewalker.jumpTo( value.nextPosition );
+					} else {
+						const position = treewalker.position.clone();
+
+						position.offset = block.maxOffset;
+
+						if ( range.containsPosition( position ) ) {
```
scofalik, line 685: And if the position is not in range, then what?

I think it should be safe to `break;` here. Even if we find something inside, it will not be recorded because this will be the top block. Even more -- it shouldn't be recorded (at least according to the method's description), so if it is, it looks like a bug, right?

After the loop we check the final position anyway. It would be good to test it with some more interesting cases like:

```
<block>..[..</block>
<block>
	..
	<inlineElement>..]..</inlineElement>
	..
</block>
```

It should work fine, though, from the quick glance at `getParentBlock`.

Note, that we currently don't have inline elements that can include selection, text or other elements.

packages/ckeditor5-engine/src/model/selection.ts
```diff
@@ -667,11 +667,25 @@ export default class Selection extends /* #__PURE__ */ EmitterMixin( TypeCheckab
...
 
 				if ( value.type == 'elementEnd' && isUnvisitedTopBlock( block as any, visited, range ) ) {
 					yield block as Element;
+				} else if ( block.is( 'model:element' ) && block.root.document!.model.schema.isBlock( block ) ) {
+					if ( value.type == 'elementEnd' ) {
+						treewalker.jumpTo( value.nextPosition );
+					} else {
+						const position = treewalker.position.clone();
+
+						position.offset = block.maxOffset;
```
scofalik, line 683:
```suggestion
						const position = Position._createAt( block, 'end' );
```

This should work fine.

scofalik (changes requested):

packages/ckeditor5-engine/src/model/treewalker.ts
```diff
@@ -183,6 +183,28 @@ export default class TreeWalker implements Iterable<TreeWalkerValue> {
 		}
 	}
 
+	/**
+	 * Moves treewalker {@link #position} to provided `position`. Treewalker will
+	 * continue traversing from that position.
```
scofalik, line 188:
```suggestion
	 * continue traversing from that position.
	 *
	 * Note: in contrary to {@link ~TreeWalker#skip}, this method does not iterate over the nodes along the way.
	 * It simply sets the current tree walker position to a new one.
	 * From the performance standpoint, it is better to use {@link ~TreeWalker#jumpTo} rather than {@link ~TreeWalker#skip}.
```

packages/ckeditor5-engine/src/model/treewalker.ts
```diff
@@ -183,6 +183,28 @@ export default class TreeWalker implements Iterable<TreeWalkerValue> {
...
+	 *
+	 * If the provided position is before the start boundary, the position will be
+	 * set to the start boundary. If the provided position is after the end boundary,
+	 * the position will be set to the end boundary.
+	 * This is done to prevent the treewalker from traversing outside the boundaries.
+	 *
+	 * @param position Position to jump to.
+	 */
+	public jumpTo( position: Position ): void {
```
scofalik, line 197: I am worried that this method does not clone the position.

I know we should avoid cloning due to possible performance problems. However, it may lead to nasty bugs after using this function, when:

*   tree walker will keep on iterating and will change the position, that you may want to use at some later point,
*   or you will change that position offset and affect the tree walker.

So far, we don't have a guideline whether you (as integrator/developer) should always pass a "new" (cloned) position to API, or whether the functions will do it for you. We can add this note do API docs too, to help avoid creating unnecessary positions.

At the beginning of the performance initiative, I checked how how big impact has creating so many position clones on the overall performance. I was surprised how many positions we clone (a few hundreds thousands for some of the test samples). But I was also surprised that cutting it in half* didn't really have much impact. And in the profiler, I couldn't see that `Position` constructor takes a lot of time. Not sure now, though, as we optimized a lot of stuff, and the percentage gains may be more significant.

\* - this change turned to be bugged anyway, but still, I could measure the performance change.

packages/ckeditor5-engine/src/model/treewalker.ts
```diff
@@ -183,6 +183,28 @@ export default class TreeWalker implements Iterable<TreeWalkerValue> {
...
+		if ( this._boundaryStartParent && position.isBefore( this.boundaries!.start ) ) {
+			position = this.boundaries!.start;
+		} else if ( this._boundaryEndParent && position.isAfter( this.boundaries!.end ) ) {
+			position = this.boundaries!.end;
+		}
+
+		this._position = position;
```
scofalik, line 204:
```suggestion
        this._position = position.clone();
```

packages/ckeditor5-engine/src/model/treewalker.ts
```diff
@@ -183,6 +183,28 @@ export default class TreeWalker implements Iterable<TreeWalkerValue> {
 		}
 	}
 
+	/**
+	 * Moves treewalker {@link #position} to provided `position`. Treewalker will
```
scofalik, line 187:
```suggestion
	 * Moves tree walker {@link #position} to provided `position`. Tree walker will
```

I know I proposed this but I made my mind and now I think it should be with a space, since classname is `TreeWalker` not `Treewalker`.

scofalik (changes requested):
Please also apply mentioned changes in the view tree walker.

packages/ckeditor5-engine/src/view/treewalker.ts
```diff
@@ -159,6 +159,27 @@ export default class TreeWalker implements IterableIterator<TreeWalkerValue> {
...
+	 * @param position Position to jump to.
+	 */
+	public jumpTo( position: Position ): void {
+		if ( this._boundaryStartParent && position.isBefore( this.boundaries!.start ) ) {
+			position = this.boundaries!.start;
+		} else if ( this._boundaryEndParent && position.isAfter( this.boundaries!.end ) ) {
+			position = this.boundaries!.end;
+		}
+
+		this._position = position;
```
scofalik, line 180:
```suggestion
		this._position = position.clone();
```

packages/ckeditor5-engine/src/view/treewalker.ts
```diff
@@ -159,6 +159,27 @@ export default class TreeWalker implements IterableIterator<TreeWalkerValue> {
 		}
 	}
 
+	/**
+	 * Moves treewalker {@link #position} to provided `position`. Treewalker will
+	 * continue traversing from that position.
+	 *
+	 * If the provided position is before the start boundary, the position will be
```
scofalik, line 166:
```suggestion
	 * continue traversing from that position.
	 *
	 * Note: in contrary to {@link ~TreeWalker#skip}, this method does not iterate over the nodes along the way.
	 * It simply sets the current tree walker position to a new one.
	 * From the performance standpoint, it is better to use {@link ~TreeWalker#jumpTo} rather than {@link ~TreeWalker#skip}.
	 *
	 * If the provided position is before the start boundary, the position will be
```

packages/ckeditor5-engine/src/view/treewalker.ts
```diff
@@ -159,6 +159,27 @@ export default class TreeWalker implements IterableIterator<TreeWalkerValue> {
 		}
 	}
 
+	/**
+	 * Moves treewalker {@link #position} to provided `position`. Treewalker will
```
scofalik, line 163:
```suggestion
	 * Moves tree walker {@link #position} to provided `position`. Tree walker will
```

scofalik (approved):

scofalik (approved):

PR #17586, opened 2024-12-03
Dynamic caching in `Mapper` for better performance

### LINKED ISSUE #17623: Introduce caching mechanism in `Mapper` to improve performance

Model-to-view position mapping is heavily used, especially during editor initialization, when the whole model is downcasted after being loaded, and also in `editor.getData()`.

Currently, `Mapper` recalculates the mapping for each single request. As a result, we do calculations for the same inputs multiple times, additionally, we don't use earlier, partially calculated results (e.g. if we calculated position for model offset 8, if we ask for model offset 9, we calculate again from the beginning).

Introducing some caching mechanisms will greatly improve performance.

scofalik:

Other (engine): Introduced dynamic caching in `Mapper` in order to improve view-to-model mapping performance, and as a result improve editor load and data save performance. Closes #17623.

Other (engine): New parameter `data` was added for `change:children` event fired by `ViewElement` and `ViewDocumentFragment` when their children change. `data` is an object with `index` property, which says at which index the change happened. Related to #17623.

Other (engine): `Mapper#registerViewToModelLength()` is now deprecated and will be removed in one of upcoming releases. Note: if this method is used, the caching mechanism for `Mapper` will be turned off which may degrade performance when handing big documents. Note: this method is used by the deprecated legacy lists feature. See #17623.

MINOR BREAKING CHANGE (engine): `Mapper#registerViewToModelLength()` is now deprecated and will be removed in one of the upcoming releases. This method is useful only in obscure and complex converters, where model element, or a group of model elements, are represented very differently in the view. We believe that every feature that uses custom view-to-model length callback can be rewritten in a way that this mechanism is no longer necessary. Note: if this method is used, the caching mechanism for `Mapper` will be turned off which may degrade performance when handing big documents. Note: this method is used by the deprecated legacy lists feature.

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/mapper.ts
+++ b/packages/ckeditor5-engine/src/conversion/mapper.ts
@@ -20,7 +20,7 @@ import type ViewElement from '../view/element.js';
 import type ViewText from '../view/text.js';
 import type ModelElement from '../model/element.js';
 import type ModelDocumentFragment from '../model/documentfragment.js';
-import type ViewNode from '../view/node.js';
+import type { default as ViewNode, ViewNodeChangeChildrenEvent, ViewNodeChangeEvent } from '../view/node.js';
 
 /**
  * Maps elements, positions and markers between the {@link module:engine/view/document~Document view} and
@@ -990,13 +990,13 @@ export class MapperCache extends /* #__PURE__ */ EmitterMixin() {
 		//
 		// Possible performance improvement. This event bubbles, so if there are multiple tracked (mapped) elements that are ancestors
 		// then this will be unnecessarily fired for each ancestor. This could be rewritten to listen only to roots and document fragments.
-		this.listenTo( viewContainer, 'change:children', ( evt, viewNode, data ) => {
-			this._invalidateCacheOnChildrenChange( viewNode as ViewElement | ViewDocumentFragment, ( data as any ).index as number );
+		this.listenTo<ViewNodeChangeChildrenEvent>( viewContainer, 'change:children', ( evt, viewNode, data ) => {
+			this._invalidateCacheOnChildrenChange( viewNode, data.index );
 		} );
 
-		this.listenTo( viewContainer, 'change:text', ( evt, viewNode ) => {
+		this.listenTo<ViewNodeChangeEvent>( viewContainer, 'change:text', ( evt, viewNode ) => {
 			// Text node has changed. Clear all the cache starting from before this text node.
-			this._clearCacheStartingBefore( viewNode as ViewText );
+			this._clearCacheStartingBefore( viewNode );
 		} );
 
 		return initialCacheItem;
--- a/packages/ckeditor5-engine/src/conversion/mapper.ts
+++ b/packages/ckeditor5-engine/src/conversion/mapper.ts
@@ -181,7 +181,7 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
 			const wasFound = this._viewToModelMapping.delete( viewElement );
 
 			if ( wasFound ) {
-				// Stop tracking after the element is no longer mapped.
+				// Stop tracking after the element is no longer mapped. We want to track all mapped elements and only mapped elements.
 				this._cache.stopTracking( viewElement );
 			}
 
@@ -211,7 +211,7 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
 			const wasFound = this._viewToModelMapping.delete( viewElement );
 
 			if ( wasFound ) {
-				// Stop tracking after the element is no longer mapped.
+				// Stop tracking after the element is no longer mapped. We want to track all mapped elements and only mapped elements.
 				this._cache.stopTracking( viewElement );
 			}
 		}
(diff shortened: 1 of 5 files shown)


scofalik:
This PR introduces caching in `Mapper` which greatly improves performance. Before, we only kept bindings between model elements and view elements, and whenever we wanted to map model position to view, we started in the mapped view element and traversed it until we reach the target model offset. This caused a lot of unnecessary traversals. In this PR, after a node is traversed for the first time, its view position and model offset is cached. Next time `Mapper` is asked about a model offset, it can use the cache to quickly retrieve the view position.

I included a lot of explanations in the API docs and inline comments, so go through the PR to learn more.

My main assumptions were:

*   Keep the whole logic in one area.
*   Don't force developers to learn/remember about a new mechanism (it should work automagically).

I decided to implement it fully inside `Mapper`.

Other options I considered were:

*   Make it inside view, i.e. `ViewNode`, `ViewElement`, etc. In a similar fashion as with recent model offset evaluation performance improvement. I didn't like it because:
    *   View in theory should not care that model exists.
    *   View is used in upcast, where it is not mapped, and nothing related to mapping/model is necessary there.
    *   Some view elements woud need to know that they are mapped (because we store the cache only for mapped elements).
    *   Even though data would be stored in view, creating this data would be inside `Mapper`, or view would somehow need to know what are views model lengths.
    *   Advantage would be that maybe the implementation would be easier to understand. But the way it is now, it is easier to turn off and maybe even to debug, as all is in one place.
*   Make it inside `DowncastWriter`.
    *   Again, changes would need to be introduced both in `Writer` and `Mapper`.
    *   Changes would need to happen in all `Writer` methods, which seemed more complex and more prone to errors to me, especially at the beginning.

I decided that making it directly in `Mapper` makes most sense.

---

Possible improvement that occured to me at later point, and I didn't try it. Maybe it would be easier to understand and simpler to implement if each traversed node kept total `modelOffset` so far. Maybe for each mapped element we could keep such flat array of descendants instead of a bit abstract structures that we have so far. It seems that many of the mechanics would be the same, just the structure of `MappingCache` and `CacheItem` would be different, possibly simpler.

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -675,6 +742,405 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
...
+		return initialCacheItem;
+	}
+
+	/**
+	 * Stops tracking given `viewContainer`.
+	 *
+	 * It removes the cached data and stops listening to {@link module:engine/view/node~ViewNodeChangeEvent `change` event} on the
+	 * `viewContainer`.
+	 *
+	 * @param viewContainer
```
niegowski, line 991: Missing description.

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -675,6 +742,405 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
...
+	 * Starts tracking given `viewContainer`, which must be mapped to a model element or model document fragment.
+	 *
+	 * Note, that this method is automatically called by {@link ~MapperCache#get `MapperCache#get()`} and there is no need to call it
+	 * manually.
+	 *
+	 * This method initializes the cache for `viewContainer` and adds callbacks for
+	 * {@link module:engine/view/node~ViewNodeChangeEvent `change` event} fired by `viewContainer`. `MapperCache` listens to `change` event
+	 * on the tracked elements to invalidate the stored cache.
+	 *
+	 * @param viewContainer
```
niegowski, line 956: Missing description.

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -675,6 +742,405 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
...
+	 * {@link ~CacheItem cache item (view position and related model offset)} to the requested model offset one.
+	 *
+	 * It can be the exact, requested mapping, or it can be mapping that is the closest starting point to look for the requested mapping.
+	 *
+	 * `viewContainer` must be a view element or document fragment that is mapped by the {@link ~Mapper Mapper}.
+	 *
+	 * If `viewContainer` is not yet tracked by the `MapperCache`, it will be automatically tracked after calling this method.
+	 *
+	 * @param viewContainer
+	 * @param modelOffset
```
niegowski, line 924: Missing description.

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -675,6 +742,405 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
...
+	 */
+	private _invalidateCacheOnChildrenChange( viewParent: ViewElement | ViewDocumentFragment, index: number ) {
+		if ( index == 0 ) {
+			// Change at the beginning of the parent.
+			if ( this._cachedMapping.has( viewParent ) ) {
+				// If this is a tracked element, clear all cache.
+				this._clearCacheAll( viewParent );
+			} else {
+				// If this is not a tracked element, remove cache starting from before this element.
+				this._clearCacheBefore( viewParent as ViewElement );
```
niegowski, line 1013:
```suggestion
				this._clearCacheStartingBefore( viewParent as ViewElement );
```

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -675,6 +742,405 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
...
+			if ( !this._cachedMapping.has( viewParent ) ) {
+				// If the parent is a non-tracked element, try clearing the cache starting before it.
+				// This situation may happen e.g. if structure like `<p><strong><em>Foo</em></strong>...` was stepped over in
+				// `Mapper#_findPositionIn()` and the children are not cached yet, but the `<strong>` element is. If something changes
+				// inside this structure, make sure to invalidate all the cache after `<strong>`.
+				return this._clearCacheBefore( viewParent );
+			} else {
+				// If the parent is a tracked element, then it means there's no cache to clear (nothing after the element has been cached).
+				return;
+			}
```
niegowski, line 1074:
```suggestion
				this._clearCacheBefore( viewParent );
			} else {
				// If the parent is a tracked element, then it means there's no cache to clear (nothing after the element has been cached).
			}

			return;
```

scofalik:
Ready for re-review.

scofalik:
Okay I didn't notice TS failed in other places -\_-.

scofalik:
I reverted some of the changes as they were breaking changes. This is the best I could do with running into more issues.

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -675,6 +775,431 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
...
+			viewPosition = new ViewPosition( parent, offset );
+		}
+
+		return viewPosition;
+	}
+
+	/**
+	 * Starts tracking given `viewContainer`, which must be mapped to a model element or model document fragment.
+	 *
+	 * Note, that this method is automatically called by {@link ~MapperCache#get `MapperCache#get()`} and there is no need to call it
```
niegowski, line 1016:
```suggestion
	 * Note, that this method is automatically called by {@link ~MapperCache#getClosest `MapperCache#getClosest()`} and there is no need to call it
```

niegowski (dismissed):

scofalik:
Looks like there's a memory leak in the mapper cache. Consecutive `editor.getData()` calls make the memory snapshot grow with time. I cannot reproduce the same on master.

scofalik:
I fixed the error in API docs and also fixed the memory leak problem. We cannot be sure if `stopTracking()` will be called for tracked view elements, hence `WeakMap`s are used, but since these are `WeakMap`s, we can't iterate over them to call `stopTracking()`. (E.g. assuming that `Mapper#clearBindings()` would call something like `MapperCache#clear()`.)

Another way to go would be to use `Map`s instead of `WeakMap`s and try to make sure that `stopTracking()` and `clear()` are called when needed, but if possible, I think that using `WeakMap`s is better, cause it just works.

Unfortunately, `this.listenTo()` retained view elements that were no longer necessary and should be garbage collected. Hence, I changed it to `.on()` and now memory is correctly freed.

The problem is that I needed to assign the class methods to properties, so that we can call `.off()`. If you have a better idea how to do it in a more elegant way, please propose.

On the side note, I am not sure if we need `.stopTracking()` at all. As I wrote, view elements are not retained by `MapperCache`, (after `.on()` is used), so we don't need to actually call `.stopTracking()` for memory management purposes. But I am afraid that it may be necessary to prevent some errors? Like, we stop tracking after view element is unbound in mapper, so it is no longer necessary. In theory something might "happen" inside it (rather unlikely after a model element is removed). As we see, we don't even have any pointers to these elements anyway. So maybe `.stopTracking()` is not actually needed and we don't need these property-functions to call `.off()`.

scofalik:
On the other node, I can see that memory grows anyway, around 0.2 MB after every `editor.getData()` call. Looks like at least some of that is related to tables/slots. I am afraid that it may grow out of hand after, let's say, an hour of working over a big document. I will create an issue for that.

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -844,6 +844,37 @@ export class MapperCache extends /* #__PURE__ */ EmitterMixin() {
...
+		super();
+
+		this._invalidateOnChildrenChangeCallback = ( evt, viewNode, data ) => {
+			this._clearCacheInsideParent( viewNode as ViewElement | ViewDocumentFragment, data!.index );
+		};
+
+		this._invalidateOnTextChangeCallback = ( evt, viewNode ) => {
+			// Text node has changed. Clear all the cache starting from before this text node.
+			this._clearCacheStartingBefore( viewNode );
+		};
```
niegowski, line 875: Those could be assigned above where we declare those properties.

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -1013,7 +1044,7 @@ export class MapperCache extends /* #__PURE__ */ EmitterMixin() {
 	/**
 	 * Starts tracking given `viewContainer`, which must be mapped to a model element or model document fragment.
 	 *
-	 * Note, that this method is automatically called by {@link ~MapperCache#get `MapperCache#get()`} and there is no need to call it
+	 * Note, that this method is automatically called by {@link ~MapperCache#get `MapperCache#getClosest()`} and there is no need to call it
```
niegowski, line 1047:
```suggestion
	 * Note, that this method is automatically called by {@link ~MapperCache#getClosest `MapperCache#getClosest()`} and there is no need to call it
```

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -675,6 +775,456 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
...
+	}
+
+	/**
+	 * Stops tracking given `viewContainer`.
+	 *
+	 * It removes the cached data and stops listening to {@link module:engine/view/node~ViewNodeChangeEvent `change` event} on the
+	 * `viewContainer`.
+	 */
+	public stopTracking( viewContainer: ViewElement | ViewDocumentFragment ): void {
+		this.stopListening( viewContainer );
```
niegowski, line 1082: This probably won't do anything since `this` knows nothing about those listeners. Shouldn't it be using those callback properties?

scofalik: Of course it does nothing :facepalm:. After rewriting some of the code I forgot to use `.off` here :facepalm:.

niegowski (dismissed):

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -893,16 +894,20 @@ export class MapperCache extends /* #__PURE__ */ EmitterMixin() {
 			// `Mapper#getModelLength()` are implemented so that parents are cached before their children.
 			//
 			// So, don't create new cache if one already exists. Instead, only save `_nodeToCacheListIndex` value for the related view node.
-			if ( viewOffset > 0 ) {
```
scofalik: When it comes to removing this `if`, I figured out that the way `save()` is used, we should never pass `viewOffset == 0` here. Running all unit tests confirmed this. There were specific unit tests for mapper caching which passed `viewOffset = 0` here, but I've changed them as it was an incorrect use of the method.

niegowski (approved):

scofalik:
Force pushed to sign commits. No changes.

Issue #17533, opened 2024-11-26
ckeditor5-react does not work with optimized imports


stonebk:
## 📝 Provide detailed reproduction steps (if any)

Using [optimized imports](https://ckeditor.com/docs/ckeditor5/latest/getting-started/setup/optimizing-build-size.html) with `@ckeditor/ckeditor5-react` results in incompatible type errors in TypeScript.

I modified the [Headless editor](https://github.com/ckeditor/ckeditor5-demos/tree/master/headless) demo to include an optimized import to `@ckeditor/ckeditor5-core` and I get build errors. 

```diff
/**
 * @license Copyright (c) 2003-2024, CKSource Holding sp. z o.o. All rights reserved.
 * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
 */

import React, { useState, useEffect, StrictMode } from 'react';
import { createRoot } from 'react-dom/client';
import { CKEditor } from '@ckeditor/ckeditor5-react';
import HeadlessEditor from './headless.js';
+import { EditorConfig } from '@ckeditor/ckeditor5-core/dist/index.js';

function App() {
	const [editor, setEditor] = useState(null);

+	const config: EditorConfig = {
+		placeholder: 'Type your text here!',
+	};

	return (
		<div className="App">
			<EditorToolbar editor={editor} />

			<CKEditor
+				config={config}
				editor={HeadlessEditor}
```

### ✔️ Expected result

No build errors.

### ❌ Actual result

![Image](https://github.com/user-attachments/assets/8c166cb6-48ae-4fea-94d3-488cc95f9b83)

### ❓ Possible solution

Switching the import from `'@ckeditor/ckeditor5-core/dist/index.js'` to `'@ckeditor/ckeditor5-core/src/index.js'` seems to resolve the build time error, but this required all imports to now point to `src` instead of `dist`, and this resulted in runtime errors:

![Image](https://github.com/user-attachments/assets/fcfd4a7b-a059-436d-83e2-637420dafc35)

This is actually the point I started from. I was previously using optimized imports before trying to upgrade with new installation methods when I ran into #17504. Upgrading without using the new installation method resulted in the same runtime error as above. I noticed that I was importing from `src` instead of `dist` so I tried updating all the imports and instead ran into the TS build error.

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

amitjangid12:
[...]

PR #17597, opened 2024-12-05
Fix: Unify TypeScript declaration files

### LINKED ISSUE #17575: Cannot combine imports from `ckeditor5` with deep imports
Consider the following code:

```ts
import { EditorUIView, type Locale, type EditingView } from 'ckeditor5';
import EditableUIView from '@ckeditor/ckeditor5-ui/dist/editableui/editableuiview.js';

export default class MyEditorUIView extends EditorUIView {
  readonly editable: EditableUIView;

  constructor(
    locale: Locale,
    editingView: EditingView,
    editableElement: HTMLElement,
  ) {
    super(locale);

    this.editable = new EditableUIView(locale, editingView, editableElement);
  }
}
```

As-is (running against CKEditor 43.3.1), the code produces this type error:

```
src/base-editor/editoruiview.ts(5,12): error TS2416: Property 'editable' in type 'MyEditorUIView' is not assignable to the same property in base type 'EditorUIView'.
  Type 'import("/project/node_modules/@ckeditor/ckeditor5-ui/dist/editableui/editableuiview").default' is not assignable to type 'import("/project/node_modules/@ckeditor/ckeditor5-ui/src/editableui/editableuiview").default'.
    Property '_editingView' is protected but type 'EditableUIView' is not a class derived from 'EditableUIView'.
```

This is because `ckeditor5` resolves to [ckeditor5/dist/index.d.ts](https://unpkg.com/browse/ckeditor5@43.3.1/dist/index.d.ts), which re-exports the contents of `@ckeditor/ckeditor5-ui`, which resolves to `@ckeditor/ckeditor5-ui/src/index.d.ts` (note that we've "hopped over" from `dist/` directories to `src/` directories). Since `@ckeditor/ckeditor5-ui/dist/index.d.ts` is a copy of `@ckeditor/ckeditor5-ui/src/index.d.ts` (rather than re-exporting it), they export different symbols that share the same name `EditorUIView`, resulting in the error above.

The type error can be fixed by changing line 2 to `import EditableUIView from '@ckeditor/ckeditor5-ui/src/editableui/editableuiview';` (replacing `dist` with `src`). However, this causes module duplication because while the _types_ in `ckeditor5` re-export `@ckeditor/ckeditor5-ui/src`, the [Javascript](https://unpkg.com/browse/ckeditor5@43.3.1/dist/ckeditor5.js) re-exports `@ckeditor/ckeditor5-ui/dist/index.js`. So now the types agree, but the bundled application will include both `@ckeditor/ckeditor5-ui/dist/index.js` (from the import via `ckeditor`) and `@ckeditor/ckeditor5-ui/src/editableui/editableuiview.js` (via the deep import), and now at runtime there will be two copies of the `EditableUIView` class, inflating the bundle and potentially messing up `instanceof` checks and the like.

I just used `@ckeditor/ckeditor5-ui` as an example -- the same issue exists with all the `@ckeditor/ckeditor5-*` packages.

The fundamental problem here is that `dist` is not isolated from `src`, and anytime the module graph "hops over" from `dist` to `src`, there will likely be a conflict that results in either type errors or module duplication.

This also runs deeper than just `ckeditor5/dist/index.d.ts` -- there are many `.d.ts` files in `dist` directories of `@ckeditor/ckeditor5-*` packages that import from the module root of other `@ckeditor/ckeditor5-*` packages, which resolves to `src`, for example the import from `@ckeditor/ckeditor5-utils` [here](https://unpkg.com/browse/@ckeditor/ckeditor5-ui@43.3.1/dist/editorui/editoruiview.d.ts), and there are many many other examples. So it's simply not safe to consume files from `dist/` folders because they might pull in files from `src/` folders that conflict with the corresponding `dist/` file imported elsewhere in the module graph.

### ❓ Possible solution

The most straightforward solution would probably be to change the `@ckeditor/ckeditor5-*` packages' `main` and `types` entries in `package.json` to point to `dist` rather than `src`, but this would be a change to the module layout that might raise compatibility issues.

The other solution would be to ensure that modules in `dist/` always import from `@ckeditor/ckeditor5-*/dist/*` rather than from the `@ckeditor/ckeditor5-*` module roots.

## 📃 Other notes

This makes importing from the `ckeditor5` module root, as well as consuming modules anywhere in `dist/`, un-tenable for my application at the moment. A workaround is to import from `ckeditor5/src/index.js` rather than `ckeditor5`, so I'm only consuming from `src/`, but based on the release notes, my impression is that CKEditor is trying to move consumers from `src/` to `dist/`. Unfortunately, that's not possible for me right now.

---

filipsobol:

Fix: Unify TypeScript declaration files. Fixes #17575 and #17533.


scripts/nim/build-ckeditor5.mjs
```diff
@@ -42,7 +42,9 @@ function dist( path ) {
 		 * We don't want to repeat this in other steps.
 		 */
 		clean: true,
-		declarations: true,
+
+		// Temporarily disabled to avoid problems caused by duplicate declarations.
+		// declarations: true,
```
pomek, line 47:
```suggestion

		// Temporarily disabled to avoid problems caused by duplicate declarations.
		declarations: false,
```

Or, remove the entire line. We do not need a commented code if we can restore it from git history.

charlttsie:
We've completed the QA tests with @MarBourb in setups using different installation methods and it looks good to us 👍

filipsobol:
I've tested the internal release against both linked issues and the fix seems to work fine.

PR #17607, opened 2024-12-08
Partial update of link text with attributes preservation.

niegowski:

Feature (link): Formatting is no longer lost after changing a URL or a displayed-text of the selected link.

### DIFF
--- a/packages/ckeditor5-link/src/linkcommand.ts
+++ b/packages/ckeditor5-link/src/linkcommand.ts
@@ -10,7 +10,7 @@
 import { Command } from 'ckeditor5/src/core.js';
 import { findAttributeRange } from 'ckeditor5/src/typing.js';
 import { Collection, diff, first, toMap } from 'ckeditor5/src/utils.js';
-import { LivePosition, type Range } from 'ckeditor5/src/engine.js';
+import { LivePosition, type Range, type Item } from 'ckeditor5/src/engine.js';
 
 import AutomaticDecorators from './utils/automaticdecorators.js';
 import { extractTextFromLinkRange, isLinkableElement } from './utils.js';
@@ -175,11 +175,11 @@ export default class LinkCommand extends Command {
 		}
 
 		model.change( writer => {
-			const updateLinkAttributes = ( range: Range ): void => {
-				writer.setAttribute( 'linkHref', href, range );
+			const updateLinkAttributes = ( itemOrRange: Item | Range ): void => {
+				writer.setAttribute( 'linkHref', href, itemOrRange );
 
-				truthyManualDecorators.forEach( item => writer.setAttribute( item, true, range ) );
-				falsyManualDecorators.forEach( item => writer.removeAttribute( item, range ) );
+				truthyManualDecorators.forEach( item => writer.setAttribute( item, true, itemOrRange ) );
+				falsyManualDecorators.forEach( item => writer.removeAttribute( item, itemOrRange ) );
 			};
 
 			const updateLinkTextIfNeeded = ( range: Range, linkHref?: string ): Range | undefined => {
@@ -217,9 +217,17 @@ export default class LinkCommand extends Command {
 							.from( attributes )
 							.filter( ( [ key ] ) => model.schema.getAttributeProperties( key ).isFormatting );
 
+						// Create a new text node.
+						const newTextNode = writer.createText( expected, formattingAttributes );
+
+						// Set link attributes before inserting to document to avoid Differ attributes edge case.
+						updateLinkAttributes( newTextNode );
+
 						// Replace text with formatting.
-						model.insertContent( writer.createText( expected, formattingAttributes ), subRange );
-						insertsLength += expected.length; // Sum of all previous inserts.
+						model.insertContent( newTextNode, subRange );
+
+						// Sum of all previous inserts.
+						insertsLength += expected.length;
 					}
 
 					return writer.createRange( range.start, range.start.getShiftedBy( newText.length ) );
--- a/packages/ckeditor5-link/src/linkcommand.ts
+++ b/packages/ckeditor5-link/src/linkcommand.ts
@@ -382,7 +382,7 @@ export default class LinkCommand extends Command {
  *
  * @internal
  * @param oldText The original text to compare.
- * @param newText The new text to compare against.
+ * @param newText  The new text to compare against.
  * @returns Array of change objects containing offset and actual/expected content.
  *
  * @example
--- a/packages/ckeditor5-link/src/linkcommand.ts
+++ b/packages/ckeditor5-link/src/linkcommand.ts
@@ -380,9 +380,10 @@ export default class LinkCommand extends Command {
  * Uses the diff utility to find the differences and groups them into chunks containing information
  * about the offset and actual/expected content.
  *
- * @param oldText The original text to compare
- * @param newText The new text to compare against
- * @returns Array of change objects containing offset and actual/expected content
+ * @internal
+ * @param oldText The original text to compare.
+ * @param newText The new text to compare against.
+ * @returns Array of change objects containing offset and actual/expected content.
  *
  * @example
  * findChanges( 'hello world', 'hi there' );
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -406,7 +406,7 @@ export default class LinkCommand extends Command {
  * 	}
  * ]
  */
-function findChanges( oldText: string, newText: string ) {
+export function findChanges( oldText: string, newText: string ): Array<{ offset: number; actual: string; expected: string }> {
```
niegowski, line 409: I'd mark this function as internal.

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -458,7 +458,7 @@ function findChanges( oldText: string, newText: string ) {
  * @param linkRange Range of the entire link element
  * @returns Text node for content updates
  */
-function getLinkPartTextNode( range: Range, linkRange: Range ) {
+export function getLinkPartTextNode( range: Range, linkRange: Range ): Item | null {
```
niegowski, line 461: I'd mark this function as internal. As I understand it is exported only for test purposes.

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
 }
+
+/**
+ * Compares two strings and returns an array of changes needed to transform one into another.
+ * Uses the diff utility to find the differences and groups them into chunks containing information
+ * about the offset and actual/expected content.
+ *
+ * @param oldText The original text to compare
+ * @param newText The new text to compare against
+ * @returns Array of change objects containing offset and actual/expected content
```
niegowski, line 385:
```suggestion
 * @param oldText The original text to compare.
 * @param newText The new text to compare against.
 * @returns Array of change objects containing offset and actual/expected content.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+ * 	},
+ * 	{
+ * 		"offset": 3,
+ * 		"actual": "ld",
+ * 		"expected": "e"
+ * 	}
+ * ]
+ */
+export function findChanges( oldText: string, newText: string ): Array<{ offset: number; actual: string; expected: string }> {
+	// Get array of operations (insert/delete/equal) needed to transform oldText into newText
```
niegowski, line 410:
```suggestion
	// Get array of operations (insert/delete/equal) needed to transform oldText into newText.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+	// Example: diff('abc', 'abxc') returns ['equal', 'equal', 'insert', 'equal']
+	const changes = diff( oldText, newText );
+
+	// Track position in both strings based on operation type
```
niegowski, line 414:
```suggestion
	// Track position in both strings based on operation type.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+	const counter = { equal: 0, insert: 0, delete: 0 };
+	const result = [];
+
+	// Accumulate consecutive changes into slices before creating change objects
```
niegowski, line 418:
```suggestion
	// Accumulate consecutive changes into slices before creating change objects.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+	let actualSlice = '';
+	let expectedSlice = '';
+
+	// Adding null as sentinel value to handle final accumulated changes
```
niegowski, line 422:
```suggestion
	// Adding null as sentinel value to handle final accumulated changes.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+	for ( const action of [ ...changes, null ] ) {
+		if ( action == 'insert' ) {
+			// Example: for 'abxc', at insert position, adds 'x' to expectedSlice
```
niegowski, line 425:
```suggestion
			// Example: for 'abc' -> 'abxc', at insert position, adds 'x' to expectedSlice.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+			expectedSlice += newText[ counter.equal + counter.insert ];
+		}
+		else if ( action == 'delete' ) {
+			// Example: for 'abc' -> 'ac', at delete position, adds 'b' to actualSlice
```
niegowski, line 429:
```suggestion
			// Example: for 'abc' -> 'ac', at delete position, adds 'b' to actualSlice.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+			actualSlice += oldText[ counter.equal + counter.delete ];
+		}
+		else if ( actualSlice.length || expectedSlice.length ) {
+			// On 'equal' or end: bundle accumulated changes into a single change object
```
niegowski, line 433:
```suggestion
			// On 'equal' or end: bundle accumulated changes into a single change object.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+				offset: counter.equal,
+				actual: actualSlice,
+				expected: expectedSlice
+			} );
+
+			actualSlice = '';
+			expectedSlice = '';
+		}
+
+		// Increment appropriate counter for the current operation
```
niegowski, line 445:
```suggestion
		// Increment appropriate counter for the current operation.
```

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,105 @@ export default class LinkCommand extends Command {
...
+
+	return result;
+}
+
+/**
+ * Returns text node withing the link range that should be updated.
+ *
+ * @param range Selection range
+ * @param linkRange Range of the entire link element
+ * @returns Text node for content updates
```
niegowski, line 459:
```suggestion
 * @param range Partial link range.
 * @param linkRange Range of the entire link.
 * @returns Text node.
```

niegowski (commented):
Partial review.

Mati365:
@niegowski done

packages/ckeditor5-link/src/linkcommand.ts
```diff
@@ -343,3 +374,107 @@ export default class LinkCommand extends Command {
...
 }
+
+/**
+ * Compares two strings and returns an array of changes needed to transform one into another.
+ * Uses the diff utility to find the differences and groups them into chunks containing information
+ * about the offset and actual/expected content.
+ *
+ * @internal
+ * @param oldText The original text to compare.
+ * @param newText  The new text to compare against.
```
niegowski, line 385:
```suggestion
 * @param newText The new text to compare against.
```

packages/ckeditor5-link/tests/linkcommand.js
```diff
@@ -628,6 +633,62 @@ describe( 'LinkCommand', () => {
 
 					expect( command.value ).to.equal( 'url' );
 				} );
+
+				describe( 'keep formatting attributes', () => {
```
niegowski, line 637: I'd add some tests with the selection collapsed in a link and different formatting attributes mixed to make sure the link range is found properly and updated. Especially with the same URL but changed text.

packages/ckeditor5-link/tests/linkcommand.js
```diff
@@ -982,6 +1043,156 @@ describe( 'LinkCommand', () => {
 			expect( command.automaticDecorators ).to.be.an.instanceOf( AutomaticDecorators );
 		} );
 	} );
+
+	describe( 'findChanges()', () => {
```
niegowski, line 1047: It would be better to test the command execution in more complex scenarios to cover all cases instead of exporting this function and testing separately, especially since we do not check if it is used by the link command.

niegowski: Same for the other exported helper.

packages/ckeditor5-link/tests/linkui.js
```diff
@@ -2436,11 +2436,15 @@ describe( 'LinkUI', () => {
 				) ).to.be.true;
 
 				expect( getModelData( editor.model ) ).to.equal(
-					'<paragraph>fo<$text linkHref="http://ckeditor.com">http://ckeditor.com</$text>[]ar</paragraph>'
+					'<paragraph>fo' +
+						'<$text linkHref="http://ckeditor.com">http://</$text>' +
+						'<$text bold="true" linkHref="http://ckeditor.com">ckeditor.com</$text>' +
+						'<$text bold="true">[]</$text>ar' +
```
niegowski, line 2442: The bold attribute is set on the selection after updating a link. Can we check how it behaves on the current master? I'm not sure if the selection should not have overridden gravity here. But looks like a case for a follow-up.

Mati365: On the master, there is no `'<$text bold="true">[]</$text>ar'`. I thought it's feature, as the bold attribute will be inherited directly start typing after applying command.

packages/ckeditor5-link/tests/linkcommand.js
Mati365: Removed

packages/ckeditor5-link/tests/linkcommand.js
```diff
@@ -628,6 +637,152 @@ describe( 'LinkCommand', () => {
...
+					} );
+
+					it( 'should handle single character change at the beginning of a formatted link', () => {
+						setData( model, '<$text linkHref="url" bold="true">[L]ink text</$text>' );
+
+						command.execute( 'url', {}, 'LL' );
+
+						expect( getData( model ) ).to.equal(
+							'[<$text bold="true" linkHref="url">LL]ink text</$text>'
+						);
```
niegowski, line 703:
```suggestion
						setData( model, '<$text linkHref="url" bold="true">Lin[]k text</$text>' );

						command.execute( 'url', {}, 'LLink text' );

						expect( getData( model ) ).to.equal(
							'<$text bold="true" linkHref="url">LLink text</$text>[]'
						);
```

PR #17615, opened 2024-12-09
Fix (ui): Add aria-hidden to icons inside ButtonView to prevent accessing them individually.

### LINKED ISSUE #17554: IconViews used in buttons should be marked as decorative since they lack accessible labels

Even though toolbar buttons have their own labels, the icons used inside don't, and accessing them individually, e.g. using the "virtual cursor" feature offered by assistive technologies presents an accessibility issue. 

### ❓ Possible solution

Icons that are used as a decoration should come with an `aria-hidden` attribute.

### Possible questions

Should all `IconView` get `aria-hidden` or just those used in buttons/menus/dropdowns/lists? Frankly, I don't remember any place using `IconView` for anything other than decoration but who knows (3rd-party). It might be a BC so it needs consideration.

---

l-lejman:

Fix: Disable accessing icons individually inside toolbar buttons. Closes #17554.

### DIFF
--- a/packages/ckeditor5-ui/src/button/buttonview.ts
+++ b/packages/ckeditor5-ui/src/button/buttonview.ts
@@ -236,7 +236,8 @@ export default class ButtonView extends View<HTMLButtonElement> implements Butto
 		this.iconView = new IconView();
 		this.iconView.extendTemplate( {
 			attributes: {
-				class: 'ck-button__icon'
+				class: 'ck-button__icon',
+				'aria-hidden': true
 			}
 		} );
 


packages/ckeditor5-ui/tests/icon/iconview.js
```diff
@@ -189,6 +189,13 @@ describe( 'IconView', () => {
 				expect( view.element.children[ 1 ].style.fill ).to.equal( 'red' );
 			} );
 		} );
+
+		describe( 'aria-hidden', () => {
+			it( 'should be added on icon creation', () => {
```
DawidKossowski, line 194: Could you move this test to the existing `describe` block for the constructor? Please align it with the style of the existing tests - for example, you can name it something like `it( 'sets #viewBox' ` to maintain consistency.

DawidKossowski, line 194: Could you move this test to the existing `describe` block for the constructor and align it with the style of the existing tests? Since we already have a unit test for checking the template, you can extend that test instead.

f1ames (approved):
LGTM 👍 

@l-lejman tested it with VoiceOver (before and after) and it's no longer possible to move to `svg` itself with virtual cursor.
Also there are no places in our codebase where `IconView` is used as standalone element/component. It wouldn't make much sense for interactive elements to use it like that - and due to that I would not consider this as BC.

DawidKossowski (approved):

PR #17622, opened 2024-12-11
Not initializing emoji plugin when merge fields use `:` as first char in prefix.

### LINKED ISSUE #17591: Verify if the : character does not conflict with merge fields

As the title suggests, we want to avoid collisions. Currently, we are only verifying the feature configuration mentioned. We should also check if merge fields will be fine.

When detecting a conflict, the emoji mention integration should not be loaded (initialized).

martnpaneq:

Internal: Not initializing emoji plugin when merge fields use `:` as first char in prefix. Closes #17591

---

### Additional information


pomek (approved):
LGTM.

Issue #17424, opened 2024-11-07
Recognize `{lower,upper}-alpha` list style types as aliases for `{lower,upper}-latin` in UI


gebsh:
## 📝 Provide a description of the improvement

_How the feature works now and what you'd like to change_?

Currently, when a list with `list-style-type` set to `lower-alpha` or `upper-alpha` is set as the editor's content, the numbered list dropdown doesn't highlight the list type.

![CKEditor 5 containing a single list item with list type set to 'lower-latin'. The numbered list dropdown is expanded and the corresponding list type option is highlighted as active.](https://github.com/user-attachments/assets/c76f86f7-5c9b-4ca3-b47f-e4690ce4c507)

![CKEditor 5 containing a single list item with list type set to 'lower-alpha'. The numbered list dropdown is expanded, and no list type option is highlighted as active.](https://github.com/user-attachments/assets/4bb3e91d-13df-4c77-8365-3f40aaa22150)

It'd be great if the editor treated both variants interchangeably.

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

PR #17627, opened 2024-12-11
Add support for `latin-alpha` / `upper-alpha` list type highlighting in list style properties buttons.

Mati365:

Feature (list): Add support for `lower-alpha` / `upper-alpha` list type highlighting in list style properties buttons. Closes https://github.com/ckeditor/ckeditor5/issues/17424
Fix (list): The list style buttons should show proper list type after clicking list for the first time.

MINOR BREAKING CHANGE (list): `lower-alpha` and `upper-alpha` list styles are now upcasted to `lower-latin` and `upper-latin` styles.

---

#### Aliasing

According to https://www.w3.org/TR/1998/PR-CSS2-19980324/lists.html#list-props:

```
lower-latin or lower-alpha
    Lower case ascii letters (a, b, c, ... z). 
upper-latin or upper-alpha
    Upper case ascii letters (A, B, C, ... Z). 
```

So it's valid to highlight the same list style property button.

##### Before

https://github.com/user-attachments/assets/387bf1c1-98f4-4a13-a424-46618dad1027

##### After

https://github.com/user-attachments/assets/d7ce6b8c-8974-49c2-a947-18d0fd6e69cb

### DIFF
--- a/packages/ckeditor5-list/src/listproperties/listpropertiesui.ts
+++ b/packages/ckeditor5-list/src/listproperties/listpropertiesui.ts
@@ -299,9 +299,7 @@ function getStyleButtonCreator( {
 
 		button.set( { label, icon, tooltip } );
 
-		listStyleCommand.on( 'change:value', () => {
-			button.isOn = listStyleCommand.value === type;
-		} );
+		button.bind( 'isOn' ).to( listStyleCommand, 'value', value => value === type );
 
 		button.on( 'execute', () => {
 			// If the content the selection is anchored to is a list, let's change its style.
--- a/packages/ckeditor5-list/src/legacylistproperties/legacylistpropertiesediting.ts
+++ b/packages/ckeditor5-list/src/legacylistproperties/legacylistpropertiesediting.ts
@@ -29,6 +29,7 @@ import LegacyListReversedCommand from './legacylistreversedcommand.js';
 import LegacyListStartCommand from './legacyliststartcommand.js';
 import { getSiblingListItem, getSiblingNodes } from '../legacylist/legacyutils.js';
 import type { ListPropertiesConfig } from '../listconfig.js';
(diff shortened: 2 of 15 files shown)


packages/ckeditor5-list/src/listproperties/listpropertiesui.ts
```diff
@@ -299,9 +300,12 @@ function getStyleButtonCreator( {
...
 
-		listStyleCommand.on( 'change:value', () => {
-			button.isOn = listStyleCommand.value === type;
-		} );
+		const refreshIsOn = () => {
+			button.isOn = getAllListStyleAliases( listStyleCommand.value ).includes( type );
+		};
+
+		refreshIsOn();
+		listStyleCommand.on( 'change:value', refreshIsOn );
```
pszczesniak, line 308: Maybe would be good to bind it using editors API?

Something like this:

```
button.bind( 'isOn' ).to( listStyleCommand, 'value', value => getAllListStyleAliases( value! ).includes( type ) );
```

WDYT?

Mati365: hm... good idea, I'll check that

Mati365: works as expected, thx. Fixed https://github.com/ckeditor/ckeditor5/pull/17627/commits/94d254372c81a57aba63606b78a35cbe4fdc5aee

pszczesniak (dismissed):
LGTM :+1:

Mati365:
@Witoso It modifies only the state of UI button and focus of opened toolbar balloon. I'm not sure if it's needed.

pszczesniak (approved):
:+1:

Nitpick: to consider - renaming `normalizeListStyle` to `getNormalizedListStyle`.

Mati365:
I'm squashing and merging, any further issues will be fixed in later PRs.

PR #17633, opened 2024-12-12
Fix (image): Add form tag and submit button to insert via url dialog window to allow closing it by enter key

### LINKED ISSUE #16902: Inconsistent behavior for URL fields

1. Open https://ckeditor.com/docs/ckeditor5/latest/features/images/images-overview.html
2. Select any text and press <kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>K</kbd> to open link balloon toolbar.
3. Type any URL (e.g. `https://ckeditor.com`) and press <kbd>Enter</kbd> while the URL input is still focused. The link should be added
4. Open the Image dropdown and select the "Insert via URL" option.
5. Type any URL (e.g. `https://placehold.co/200x300`) and press <kbd>Enter</kbd> while the URL input is still focused.

### ✔️ Expected result

The image should be inserted into the editor – just like it happened with the link.

### ❌ Actual result

Nothing happens, the <kbd>Enter</kbd> keypress is ignored.

### ❓ Possible solution

The <kbd>Enter</kbd> key behavior from the Link balloon is connected with the [implicit form submission behavior](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission): if the text `input` in the form is focused when the user presses <kbd>Enter</kbd>, the form is submitted.

The `input` in the Link balloon is encapsulated inside a `form` and thanks to that, pressing <kbd>Enter</kbd> inserts the link. However, the `input` in the Image dialog is not encapsulated inside a `form` leading to a worse user experience (the user needs to either click the "Insert" button or – if they use only the keyboard for navigation – tab to the "Insert" button and press <kbd>Enter</kbd>/<kbd>Space</kbd>). Adding the `form` element to the Image dialog should fix the issue.

## 📃 Other details

* Browser: N/A (de facto all browsers support the implicit form submission)
* OS: N/A
* First affected CKEditor version: 42.0.0
* Installed CKEditor plugins: `image`

---

l-lejman:

Fix (image): Insert image via url dialog can be closed by "enter" button when input is focused. Closes #16902.

---

I tried to add better unit tests for this case but submitting form by "enter" button works different in browser than in unit test environment. I left comment to prevent confusion why unit test has no "enter" click.

### DIFF
--- a/packages/ckeditor5-image/tests/imageinsert/imageinsertviaurlui.js
+++ b/packages/ckeditor5-image/tests/imageinsert/imageinsertviaurlui.js
@@ -120,6 +120,11 @@ describe( 'ImageInsertViaUrlUI', () => {
 			expect( dialog.view.actionsView.children.get( 1 ).label ).to.equal( 'Accept' );
 		} );
 
+		it( 'has submittable form', () => {
+			expect( dialog.view.element.querySelector( 'form.ck-dialog' ) ).to.exist;
+			expect( acceptButton.type ).to.equal( 'submit' );
+		} );
+
 		it( 'should bind #isImageSelected', () => {
 			expect( urlView.isImageSelected ).to.be.false;
 
@@ -186,6 +191,7 @@ describe( 'ImageInsertViaUrlUI', () => {
 
 		testSubmit( 'accept button', () => acceptButton.fire( 'execute' ) );
 
+		// Can not be checked by simulating enter click because of unit test limitations
 		testSubmit( 'form submit (enter key)', () => {
 			const form = dialog.view.contentView.children.get( 0 );
 
--- a/packages/ckeditor5-ui/src/dialog/dialogview.ts
+++ b/packages/ckeditor5-ui/src/dialog/dialogview.ts
@@ -260,7 +260,7 @@ export default class DialogView extends /* #__PURE__ */ DraggableViewMixin( View
 			},
 			children: [
 				{
-					tag: 'div',
+					tag: 'form',
 					attributes: {
 						tabindex: '-1',
 						class: [
--- a/packages/ckeditor5-image/src/imageinsert/imageinsertviaurlui.ts
+++ b/packages/ckeditor5-image/src/imageinsert/imageinsertviaurlui.ts
@@ -196,6 +196,7 @@ export default class ImageInsertViaUrlUI extends Plugin {
 				{
 					label: t( 'Accept' ),
 					class: 'ck-button-action',
+					type: 'submit',
 					withText: true,
 					onExecute: () => this._handleSave()
 				}
--- a/packages/ckeditor5-ui/src/dialog/dialogactionsview.ts
+++ b/packages/ckeditor5-ui/src/dialog/dialogactionsview.ts
@@ -149,7 +149,7 @@ export default class DialogActionsView extends View {
  */
 export type DialogActionButtonDefinition =
 	Pick<Button, 'label'> &
-	Partial<Pick<Button, 'withText' | 'class' | 'icon'>> &
+	Partial<Pick<Button, 'withText' | 'class' | 'icon' | 'type'>> &
 	{
 		onExecute: Function;
 		onCreate?: ( button: ButtonView ) => void;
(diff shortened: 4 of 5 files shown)


packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -260,7 +260,7 @@ export default class DialogView extends /* #__PURE__ */ DraggableViewMixin( View
 			},
 			children: [
 				{
-					tag: 'div',
+					tag: 'form',
```
f1ames, line 263: `DialogView` is a "generic" component reused by CKEditor 5 in multiple places by multiple plugins. This change would affect all of the dialogs so this is not the right place for such change if we only want to adjust the behavior of a specific plugins dialog.

packages/ckeditor5-image/tests/imageinsert/imageinsertviaurlui.js
```diff
@@ -186,6 +191,7 @@ describe( 'ImageInsertViaUrlUI', () => {
 
 		testSubmit( 'accept button', () => acceptButton.fire( 'execute' ) );
 
+		// Can not be checked by simulating enter click because of unit test limitations
```
f1ames, line 194: It's a good approach to describe "not so obvious" code, however this could use some more details to be clear. You could describe that we fire form `submit` event to simulate what browser does on `enter` press instead of trying to trigger `enter` key event on a button directly.

f1ames (changes requested):
The approach for this fix went into a good direction, to have a form in place so `enter` could be handled 👍  However, the changes you applied affects `DialogView` component globally (see review comments) which is not what we want IMHO.

The right approach would be to modify to `ImageInsertViaUrlUI` to have the form element rendered into the dialog and then proper submit handling. There are other plugins (like `MediaEmbed`) which already have this so you may take a look how it's done there.

packages/ckeditor5-image/tests/imageinsert/imageinsertviaurlui.js
```diff
@@ -186,6 +191,10 @@ describe( 'ImageInsertViaUrlUI', () => {
 
 		testSubmit( 'accept button', () => acceptButton.fire( 'execute' ) );
 
+		/**
+	 	* Browsers handle clicking Enter for form submitting by its own but unit test can not, it is known limitation.
+		* We fire a form submit event to simulate the behavior that occurs when pressing Enter in the browser.
+	 	*/
```
f1ames, line 197:
```suggestion
		/**
	 	* Browsers handle pressing Enter on forms natively by submitting it. We fire a form submit event to simulate that behavior.
	 	*/
```

f1ames: I think it could be a bit shorter as above. And a "nitpick", `enter` is pressed not clicked (usually at least 😄).

f1ames (changes requested):
First thing, always make sure that CI is green when passing a PR for a review. Here, there is one job failing (cc is not 100% after recent changes).

Second thing, as for:
> The approach for this fix went into a good direction, to have a form in place so `enter` could be handled 👍 However, the changes you applied affects `DialogView` component globally (see review comments) which is not what we want IMHO.
> 
> The right approach would be to modify to `ImageInsertViaUrlUI` to have the form element rendered into the dialog and then proper submit handling. There are other plugins (like `MediaEmbed`) which already have this so you may take a look how it's done there.

Well, you solved it by adding a flag, but it's unnecessary and makes the generic dialog more complex. As each plugin has a full control over it's views and can render anything there, there is no reason for lifting this logic higher in components hierarchy.
As suggested before, we already have plugins doing this. You may take a look on MediaEmbed plugin and its UI ([here is the view definition](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-media-embed/src/ui/mediaformview.ts)) and see how "Image insert via URL" view can be adjusted ([here](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-image/src/imageinsert/ui/imageinserturlview.ts)).

f1ames (approved):
LGTM 👍

PR #17651, opened 2024-12-16
Internal: Update the logic of the check.

gorzelinski:

Internal: Update the logic of the license check.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-core/src/editor/editor.ts
+++ b/packages/ckeditor5-core/src/editor/editor.ts
@@ -1110,8 +1110,8 @@ type LicenseErrorReason =
 	'featureNotAllowed' |
 	'evaluationLimit' |
 	'trialLimit' |
-	'usageLimit' |
 	'developmentLimit' |
+	'usageLimit' |
 	'distributionChannel';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-core/src/editor/editor.ts
```diff
@@ -542,6 +542,13 @@ export default abstract class Editor extends /* #__PURE__ */ ObservableMixin() {
 				} );
 			}
 
+			if ( licensePayload.licenseType === 'development' ) {
+				console.info(
+					'You are using the development version of CKEditor 5. ' +
+					'Make sure you will not use it in the production environment.'
+				);
+			}
```
gorzelinski, line 550: I thought about extracting logging into a separate if block for all types. But, after the change those type differ, so it's probably better this way.

packages/ckeditor5-core/tests/editor/licensecheck.js
```diff
@@ -466,7 +466,7 @@ describe( 'Editor - license check', () => {
 				dateNow.restore();
 			} );
 
-			it( 'should block editor after 10 minutes (development license)', () => {
+			it( 'also should not block editor after 10 minutes (development license)', () => {
```
arkflpc, line 469:
```suggestion
			it( 'should not block editor after 10 minutes (development license)', () => {
```

arkflpc (changes requested):

arkflpc (approved):

PR #17672, opened 2024-12-19
Internal: Use `import.meta.resolve` instead of `require.resolve`.

### LINKED ISSUE #17671: Fix nightly docs
Nightly docs no longer build after the changes introduced in #17597.

filipsobol:

Internal: Use `import.meta.resolve` instead of `require.resolve`. Fixes #17671.

---

We've added the `exports` field to the `package.json` files with `types` and `import` fields, but no `require` because our packages are ESM-only. This is the reason for the `require.resolve` call fails.

### DIFF
--- a/scripts/docs/utils.mjs
+++ b/scripts/docs/utils.mjs
@@ -5,14 +5,11 @@
 
 /* eslint-env node */
 
-import module from 'module';
 import path from 'path';
 import fs from 'fs/promises';
 import { glob } from 'glob';
 import { loaders } from '@ckeditor/ckeditor5-dev-utils';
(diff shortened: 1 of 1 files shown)


pomek:
According to:

* https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/development-environment.html#requirements
	image
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve:
	image

The proposed solution cannot be used.

Witoso:
If this blocks us for good, we can bump it as it is our internal dependency. (But I understand it needs to be tested across all things)

pomek (approved):

PR #17675, opened 2024-12-19
Fix: Add `default` to `exports` entries in `package.json` files.

### LINKED ISSUE #17671: Fix nightly docs

filipsobol:

Internal: Add `default` to `exports` entries in `package.json` files. Fixes #17671.

---

In a few places in our tools, we use `require` instead of `import` to get data from packages. Instead of fixing all of them and potentially introducing breaking changes, it's easier to add a fallback `default` entry to the `exports` field.


pomek (approved):
LGTM.

pomek (approved):

Issue #16561, opened 2024-06-13
`View#detachDomRoot()` might operate on non-existing DOM element


pomek:
## 📝 Provide detailed reproduction steps (if any)

There is no easy way to reproduce the issue without a React application. Due to its nature and double rendering in [StrictMode](https://react.dev/reference/react/StrictMode), the editor might try to operate on a DOM element that is no longer available.

https://github.com/ckeditor/ckeditor5/blob/ec3648676a5bf466e0f619f3101de453e419bd0b/packages/ckeditor5-engine/src/view/view.ts#L318-L337

L319, `domRoot` might be `undefined` when a DOM element is no longer available.

When the application crashes, it shows the following trace:

```
TypeError: Cannot read properties of undefined (reading 'attributes')
    at Zc.detachDomRoot (VM3382 ckeditor5.js:5:161689)
    at wC.destroy (VM3382 ckeditor5.js:5:768794)
    at vC.destroy (VM3382 ckeditor5.js:5:771313)
```

### ✔️ Expected result

No error. The application should not crash.

### ❌ Actual result

As reported above.

## 📃 Other details

* Related PR: https://github.com/ckeditor/ckeditor5-react/pull/480.
* The following diff resolves all my issues.
	```
	diff --git a/packages/ckeditor5-engine/src/view/view.ts b/packages/ckeditor5-engine/src/view/view.ts
	index 6c1e533122..11477e433e 100644
	--- a/packages/ckeditor5-engine/src/view/view.ts
	+++ b/packages/ckeditor5-engine/src/view/view.ts
	@@ -316,7 +316,11 @@ export default class View extends /* #__PURE__ */ ObservableMixin() {
	         * @param name Name of the root to detach.
	         */
	        public detachDomRoot( name: string ): void {
	-               const domRoot = this.domRoots.get( name )!;
	+               const domRoot = this.domRoots.get( name );
	+
	+               if ( !domRoot ) {
	+                       return;
	+               }
	
	                // Remove all root attributes so the DOM element is "bare".
	                Array.from( domRoot.attributes ).forEach( ( { name } ) => domRoot.removeAttribute( name ) );
	```
* This is not the first time that suppressing TypeSccript possible null values results in unexpected errors in the applications consuming the editor.

Reinmar:
The error mentioned in the original issue is thrown when calling `destroy()` twice on a single editor instance.

We should secure this method from being called twice. However, we need to warn the integrator that they called it twice. 

Idea:

[...]

PR #17684, opened 2024-12-23
No longer crash during `destroy()` of `MultiRootEditor` or `InlineEditor` when the editable was detached manually before destroying.

Mati365:

Fix (editor-inline, editor-multi-root):  No longer crash the editor during `destroy()` when the editable was manually detached before destroying. Closes https://github.com/ckeditor/ckeditor5/issues/16561

---

### Additional information

Some editor types check if elements exist before destroying, while others don't. Here's what works correctly:

https://github.com/ckeditor/ckeditor5/blob/ck/16561/packages/ckeditor5-editor-classic/src/classiceditorui.ts#L141-L143
https://github.com/ckeditor/ckeditor5/blob/ck/16561/packages/ckeditor5-editor-balloon/src/ballooneditorui.ts#L102-L104
https://github.com/ckeditor/ckeditor5/blob/ck/16561/packages/ckeditor5-editor-decoupled/src/decouplededitorui.ts#L96-L98

These editors don't have the checks yet:

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-editor-inline/src/inlineeditorui.ts#L117-L119
https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-editor-multi-root/src/multirooteditorui.ts#L162-L164

This PR adds missing checks to fix crashes in React integration shown here:

1. https://github.com/ckeditor/ckeditor5-react/blob/474c8aa9c222b61ed86dad329e05beb92ede29a8/src/useMultiRootEditor.tsx#L64-L88
2. https://github.com/ckeditor/ckeditor5/issues/16561

tl;dr React integration removes editables from DOM just before destroying the editor in Strict Mode. This was causing the multiroot editor to crash.

### DIFF
--- a/packages/ckeditor5-editor-balloon/tests/ballooneditorui.js
+++ b/packages/ckeditor5-editor-balloon/tests/ballooneditorui.js
@@ -189,6 +189,13 @@ describe( 'BalloonEditorUI', () => {
 
 			sinon.assert.callOrder( parentDestroySpy, viewDestroySpy );
 		} );
+
+		it( 'should not crash if called twice', async () => {
+			const newEditor = await VirtualBalloonTestEditor.create( '' );
+
+			await newEditor.destroy();
+			await newEditor.destroy();
+		} );
 	} );
 
 	describe( 'element()', () => {
--- a/packages/ckeditor5-editor-classic/tests/classiceditorui.js
+++ b/packages/ckeditor5-editor-classic/tests/classiceditorui.js
@@ -694,6 +694,13 @@ describe( 'ClassicEditorUI', () => {
 
 			sinon.assert.callOrder( parentDestroySpy, viewDestroySpy );
 		} );
+
+		it( 'should not crash if called twice', async () => {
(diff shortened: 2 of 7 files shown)


pomek:
AFAIR, there was also a case where somebody called `#destroy()` twice, and it was also thrown. Could we check if before and after your changes it is valid? If so, I'd add one more test.

As for me, LGTM, but somebody from the Core team should accept it before merging. I am not too familiar with the editor's internals.

Mati365:
Sure thing, I'll take a look.

Mati365:
@pomek  

> AFAIR, there was also a case where somebody called #destroy() twice, and it was also thrown. Could we check if before and after your changes it is valid? If so, I'd add one more test.

Confirm, the exception was thrown in inline editor and multi root editor. So it looks like this one fix solved two independent React issues (that one with multiroots and the other in the builder).

niegowski (approved):

PR #17687, opened 2024-12-23
Added a search input to emoji picker.

### LINKED ISSUE #17659: Implement `SearchTextView` instead of `InputTextView`

To align the input search to CKEditor 5 features:

Image

We should replace the `InputTextView` class with `SearchTextView`.

The class/view we want to implement:

https://github.com/ckeditor/ckeditor5/blob/aa1ea784172b9b30a9cda351bd753116c6d937b2/packages/ckeditor5-ui/src/search/text/searchtextview.ts#L43-L45

Also, thanks to that, we will improve the view the mechanism does not find emojis:

Image

Btw could we drop the footer section in such a case?

martnpaneq:

Internal: Added a search input to emoji picker. Closes #17659.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-emoji/src/ui/emojisearchview.ts
+++ b/packages/ckeditor5-emoji/src/ui/emojisearchview.ts
@@ -7,7 +7,7 @@
  * @module emoji/ui/emojisearchview
  */
 
-import { View, createLabeledInputText, LabeledFieldView, type InputView } from 'ckeditor5/src/ui.js';
+import { View, createLabeledInputText, SearchTextQueryView, type InputView } from 'ckeditor5/src/ui.js';
 import { type Locale } from 'ckeditor5/src/utils.js';
 
 export default class EmojiSearchView extends View {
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-emoji/src/ui/emojisearchview.ts
```diff
@@ -29,7 +29,7 @@ export default class EmojiSearchView extends View {
 		this.setTemplate( {
 			tag: 'div',
 			attributes: {
-				class: [ 'ck', 'ck-emoji-input' ]
+				class: [ 'ck', 'ck-emoji-input', 'ck-search' ]
```
martnpaneq, line 32: In order for the styles to look correct, the `ck-search` class is needed. So I'm not sure if exposing the search input is a good idea. The component should be self-sustaining, without needing any additional modifications like adding classes, to work.

pomek (approved):
LGTM.

PR #17690, opened 2024-12-23
Prevent bookmark popup flickering when clicking paragraphs with trailing bookmarks.

Mati365:

Fix (bookmark): Prevent bookmark popup flickering when clicking paragraphs with trailing bookmarks.

---

### Additional information

When a widget (e.g., a bookmark anchor) is clicked, two consecutive `update` events may fire:

1. First, when the widget gets focused (this is fine)
3. Second, when the selection changes shortly after (beware, dragons live here)

While the widget usually stays selected, sometimes the selection may move to a different element (e.g., to the start of the parent paragraph). This can cause the toolbar to flicker as it's quickly shown and hidden.

This issue particularly affects widgets like anchors that contain a single SVG element. 

Making the element `inline-block` seems to reduce flickering. 

#### Before

https://github.com/user-attachments/assets/c931542c-fe3a-439a-9cf8-2070355d72a9

#### After

https://github.com/user-attachments/assets/7200e9b3-e3c6-4222-be14-fed75f83254e

### DIFF
--- a/packages/ckeditor5-theme-lark/theme/ckeditor5-bookmark/bookmark.css
+++ b/packages/ckeditor5-theme-lark/theme/ckeditor5-bookmark/bookmark.css
@@ -12,6 +12,7 @@
 
 .ck-bookmark {
 	&.ck-widget {
+		display: inline-block;
 		outline: none;
 
 		& .ck-bookmark__icon .ck-icon__fill {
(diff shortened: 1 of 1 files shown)


niegowski (changes requested):
I suggest fixing the bookmark stylesheet instead. Adding `display: inline-block;` to the `.ck-bookmark.ck-widget` selector should fix the problem.

pszczesniak:
Fix that @niegowski suggested doesn't eliminate it completely, but it limits it more.

It also fixes issue shown on the video (clicking at the end of paragraphs):

https://github.com/user-attachments/assets/5155b4da-fb0f-496f-926c-8fffc79ff414

niegowski (approved):

niegowski:
Please update the PR title.

PR #17698, opened 2024-12-30
Not displaying unsupported emoji in the emoji picker and mention plugin.

### LINKED ISSUE #17666: Spike: find a solution to display supported emoji only

Right now, it is possible to render an emoji tile that is unsupported on a particular OS. We would like to render only working emojis.

A proposal: let's use some emoji font, e.g., https://fonts.google.com/noto/specimen/Noto+Color+Emoji. I believe there will be more than this, so let's find alternatives.

If using a dedicated font is not an option, perhaps we could remove unsupported emojis from the picker?

martnpaneq:

Internal: Not displaying unsupported emoji in the emoji picker and mention plugin. Closes #17666.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-emoji/src/emojipicker.ts
+++ b/packages/ckeditor5-emoji/src/emojipicker.ts
@@ -32,6 +32,7 @@ import {
 import EmojiToneView, { type SkinToneId } from './ui/emojitoneview.js';
 
 const VISUAL_SELECTION_MARKER_NAME = 'emoji-picker';
+const BASELINE_EMOJI_WIDTH = 24;
 
 /**
  * The emoji picker plugin.
@@ -168,7 +169,6 @@ export default class EmojiPicker extends Plugin {
 	} ): Promise<EmojiGroup> {
 		const databaseGroup = await this._emojiDatabase.getEmojiByGroup( databaseId );
 		const container = this._createEmojiWidthTestingContainer();
-		const baselineEmojiWidth = 24;
 
 		const items = databaseGroup
 			.filter( item => {
@@ -178,7 +178,7 @@ export default class EmojiPicker extends Plugin {
 				// against are the ones that are 2x the size, because those are truly broken (person with red hair = person with
 				// floating red wig, black cat = cat with black square, polar bear = bear with snowflake, etc.)
 				// So here we set the threshold at 1.8 times the size of the baseline emoji.
-				return ( emojiWidth / 1.8 < baselineEmojiWidth ) && ( emojiWidth >= baselineEmojiWidth );
+				return ( emojiWidth / 1.8 < BASELINE_EMOJI_WIDTH ) && ( emojiWidth >= BASELINE_EMOJI_WIDTH );
 			} )
 			.map( item => {
 				const name = item.annotation;
@@ -458,7 +458,7 @@ export default class EmojiPicker extends Plugin {
 		container.style.position = 'absolute';
 		container.style.left = '-9999px';
 		container.style.whiteSpace = 'nowrap';
-		container.style.fontSize = '24px';
+		container.style.fontSize = BASELINE_EMOJI_WIDTH + 'px';
 		document.body.appendChild( container );
 
 		return container;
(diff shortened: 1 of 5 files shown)


martnpaneq:
This solution bases on the solution used in [emoji-picker-element](https://github.com/nolanlawson/emoji-picker-element/blob/ae637c74fdc9307566f21a630aa1d84c69e7d2e7/src/picker/utils/checkZwjSupport.js).

martnpaneq:
Testing this would be hard, because we would have to mock an unsupported emoji. I'm not adding any tests for now - if anyone thinks I should add them, let me know.

psmyrek (commented):
It looks like most of the emojis with flags have been filtered out on Windows, even though they were displayed correctly.

Before:

before

After:

after

The strange [C][Q] emoji should be filtered but it is still available.

packages/ckeditor5-emoji/src/emojipicker.ts
```diff
@@ -167,11 +167,20 @@ export default class EmojiPicker extends Plugin {
 		databaseId: number; title: string; exampleEmoji: string;
 	} ): Promise<EmojiGroup> {
 		const databaseGroup = await this._emojiDatabase.getEmojiByGroup( databaseId );
-
-		return {
-			title,
-			exampleEmoji,
-			items: databaseGroup.map( item => {
+		const container = this._createEmojiWidthTestingContainer();
+		const baselineEmojiWidth = this._getNodeWidth( container, '🙂' );
```
psmyrek, line 171: I think that we do not need to calculate the baseline emoji width, but just store the container width as a baseline value:

```suggestion
		const baselineEmojiWidth = 24;
```

After that, it seems that all emojis are displayed correctly. Not supported emojis are also filtered out.

martnpaneq: Thanks, that looks good!

psmyrek (approved):
LGTM 🎉 

I didn't find any unsupported emoji. All other emojis are displayed correctly.

obraz

obraz

pomek (changes requested):
R- from my side, due to missing options (approaches to meet the requirements and goals), cons, pros, and conclusions (what, why, etc.) under the issue.

packages/ckeditor5-emoji/src/emojipicker.ts
```diff
@@ -432,6 +449,33 @@ export default class EmojiPicker extends Plugin {
 			} );
 		}
 	}
+
+	/**
+	 * Creates a div for emoji width testing purposes.
+	 */
+	private _createEmojiWidthTestingContainer(): HTMLDivElement {
+		const container = document.createElement( 'div' );
```
pomek, line 457: To avoid processing, I would mark this element with a proper [aria attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden).

martnpaneq: Added in [abad86a](https://github.com/ckeditor/ckeditor5/pull/17698/commits/abad86a60f60e17dce635513e10d4378999c019a)

martnpaneq:
> R- from my side, due to missing options (approaches to meet the requirements and goals), cons, pros, and conclusions (what, why, etc.) under the issue.

Conclusions added to the spike: https://github.com/ckeditor/ckeditor5/issues/17666#issuecomment-2575798888

scripts/ci/generate-circleci-configuration.mjs
```diff
@@ -140,6 +140,7 @@ const persistToWorkspace = fileName => ( {
 			machine: true,
 			steps: [
 				...bootstrapCommands(),
+				'install_newest_emoji',
```
pomek, line 143: I am afraid the same change will be needed in the Commercial repository.

pomek: But we can cover it in a follow-up issue.

pomek (approved):

PR #17701, opened 2024-12-31
Make link toolbar button enabled when text above and link are selected.

### LINKED ISSUE #16965: Link is not editable when text above and link are selected.

1. Add the elements (in described order) text, link
2. Select all editor content

Reproduction as mentioned above:
[Screencast from 23-08-24 10:51:20.webm](https://github.com/user-attachments/assets/48eb46b1-fe90-4da4-a38f-806e0df8877a)

More elaborate demo of the issue:
[Screencast from 23-08-24 10:14:13.webm](https://github.com/user-attachments/assets/8304aaf9-b39d-4acc-b90e-8fcdc2854409)
[Screencast from 23-08-24 10:15:55.webm](https://github.com/user-attachments/assets/50938312-d737-4b77-9e1d-f54951a6f3d8)

### ✔️ Expected result

I would expect that the link control in the toolbar would be highlighted and I can edit the link. Similiar behaviour happens when the editor has the content in described order link text. When the text is below the link and I select the whole content the link control in the toolbar is highlighted and I can edit the link.

### ❌ Actual result

The link control in the toolbar is not highlighted and I cannot edit the link.

### ❓ Possible solution

It seems to be an issue in the `LinkCommand` `refresh`-method.

## 📃 Other details

* Browser: Firefox and Chromium. I was not able to test for Safari.

---

Mati365:

Fix (link):  The `Link` toolbar button should now be enabled when the selection starts at a non-link element and ends at the end of a link. Closes #16965

---

#### Before

https://github.com/user-attachments/assets/e85d4e2c-dfc1-4a00-9050-c8e429d964b7

#### After

https://github.com/user-attachments/assets/e7363195-5b4b-4b85-b0f4-e4612cb18aa1

### DIFF
--- a/packages/ckeditor5-engine/src/model/documentselection.ts
+++ b/packages/ckeditor5-engine/src/model/documentselection.ts
@@ -1136,20 +1136,47 @@ class LiveSelection extends Selection {
 
 		if ( !this.isCollapsed ) {
 			// 1. If selection is a range...
-			const range = new TreeWalker( {
-				boundaries: this.getFirstRange(),
-				ignoreElementEnd: true,
-				shallow: true
-			} );
+			const firstRange = this.getFirstRange();
 
 			// ...look for a first character node in that range and take attributes from it.
-			for ( const value of range ) {
+			for ( const value of firstRange ) {
 				// If the item is an object, we don't want to get attributes from its children...
 				if ( value.item.is( 'element' ) && schema.isObject( value.item ) ) {
 					// ...but collect attributes from inline object.
-					attrs = mergeAttributes( attrs, getTextAttributes( value.item, schema ) );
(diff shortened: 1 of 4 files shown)


Mati365:
I discovered that it's related to general attributes handling mechanism, as it's reproducible on highlighted nodes. 

This one is highlighted:
obraz

... but this one is not:
obraz

Not sure if making exception only for links is a good idea :thinking:

Mati365:
It's related to this function: https://github.com/ckeditor/ckeditor5/blob/ck/epic/17230-linking-experience-16965/packages/ckeditor5-engine/src/model/documentselection.ts#L1127-L1142 , it stops lookup after finding first `element`.

PR #17707, opened 2025-01-02
Add `sourceEditorId` property to clipboard input events.

### LINKED ISSUE #15935: Add custom property for the content copied from the editor in the clipboard pipeline

In CKEditor 4, `dataTransfer` object has a custom property (https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_clipboard_dataTransfer.html#property-sourceEditor) that indicates whether the data has been copied from the editor.

It would be nice to introduce a way for detecting whether copied content is coming from CKEditor 5.

---

Mati365:

Feature (clipboard): Add ability to detect paste events originating from editor. Closes #15935 

---

### Additional information

The main use case is we need to track if the user has pasted content from outside the editor so we can decide if the content needs to be modified before it’s added into CKEditor. See more https://github.com/ckeditor/ckeditor5/issues/15935#issuecomment-1968599167

### DIFF
--- a/packages/ckeditor5-clipboard/tests/clipboardpipeline.js
+++ b/packages/ckeditor5-clipboard/tests/clipboardpipeline.js
@@ -507,7 +507,7 @@ describe( 'ClipboardPipeline feature', () => {
 				sinon.assert.calledWith( spy, editor.id );
 			} );
 
-			it( 'should contain editor ID when pasting content copied from the same editor', () => {
+			it( 'should contain an editor ID when pasting content copied from the same editor', () => {
 				const spy = sinon.spy();
 
 				setModelData( editor.model, '<paragraph>f[oo]bar</paragraph>' );
--- a/packages/ckeditor5-clipboard/tests/clipboardpipeline.js
+++ b/packages/ckeditor5-clipboard/tests/clipboardpipeline.js
@@ -480,7 +480,7 @@ describe( 'ClipboardPipeline feature', () => {
 				sinon.assert.calledWith( inputTransformationSpy, null );
 			} );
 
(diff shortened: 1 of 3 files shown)


Mati365:
Solution needs to be changed, as it'll be better to make it more unified with drag & drop which uses mimetypes.

Mati365:
@niegowski I applied suggestion, mime type is now used to determine source editor.

packages/ckeditor5-clipboard/tests/clipboardpipeline.js
```diff
@@ -462,10 +462,105 @@ describe( 'ClipboardPipeline feature', () => {
...
+
+				viewDocument.fire( 'paste', {
+					dataTransfer: dataTransferMock,
+					preventDefault: () => {},
+					stopPropagation: () => {}
+				} );
+
+				sinon.assert.calledWith( contentInsertionSpy, null );
+			} );
+		} );
```
mmotyczynska, line 553: The test description states `should be propagated to contentInsertion event`, but it actually checks something else. You can either create two separate tests (similar to what you did earlier for `inputTransformation`) to cover both scenarios: one for outside the editor (no propagation) and one for within the editor (propagation occurs), or adjust the test itself to ensure it aligns with the current description.

mmotyczynska (changes requested):

packages/ckeditor5-clipboard/tests/clipboardpipeline.js
Mati365: Fixed

mmotyczynska (approved):

PR #17724, opened 2025-01-08
Treat type dependencies as production dependencies

### LINKED ISSUE #17213: Type of `Editor.ui` not resolvable with Yarn (strict) PnP mode

1. Setup a project with TypeScript + Yarn PnP (default strict mode)
2. Create a plugin file along the lines of
```ts
import {Plugin} from 'ckeditor5';

export default class FooUI extends Plugin {
  init() {
    this.editor.ui.componentFactory.add('addFoo', (locale) => {
      ...
  }
}
```
3. Run `tsc` to check types

### ✔️ Expected result

The type of `this.editor.ui` to be resolved correctly, so that consequently the type of `locale` in the `componentFactory.add` call can be inferred.

### ❌ Actual result

The type of `this.editor.ui.componentFactory` is marked as `any` yielding an error `Parameter 'locale' implicitly has an 'any' type.ts(7006)`.

### ❓ Possible solution

This issue only appears with Yarn PnP mode, as the import `import type { EditorUI } from '@ckeditor/ckeditor5-ui';` within the `ckeditor5-core` cannot be resolved, since the package is not listed as dependency in `ckeditor5-core`. To make this resolvable, the `@ckeditor/ckeditor5-ui` should be listed as dependency, peerDependency or optional peerDependency in the `ckeditor5-core`, with the last one probably being the best option.

Manual workarounds: 
- import the type `EditorUI` and manually cast with `(this.editor.ui as EditorUI)`
- use `pnpMode: loose` or non-pnp mode with Yarn
- patch the (optional) peer dependency within your own project by adding the following to `.yarnrc.yml`:
```yaml
packageExtensions:
  # See https://github.com/ckeditor/ckeditor5/issues/17213
  "@ckeditor/ckeditor5-core@*":
    peerDependencies:
      "@ckeditor/ckeditor5-ui": "*"
    peerDependenciesMeta:
      "@ckeditor/ckeditor5-ui":
        optional: true
```

## 📃 Other details

* Browser: N/A
* OS: N/A
* First affected CKEditor version: 42.0.0 (possibly earlier); *edit:* still observed on 43.2.0

---

filipsobol:

Fix: Treat type dependencies as production dependencies. Fixes #17213.


packages/ckeditor5-ui/package.json
```diff
@@ -13,6 +13,7 @@
   "main": "src/index.ts",
   "dependencies": {
     "@ckeditor/ckeditor5-core": "44.1.0",
+    "@ckeditor/ckeditor5-editor-multi-root": "44.1.0",
```
arkflpc, line 16: I would get rid of this dependency in this PR.

packages/ckeditor5-basic-styles/package.json
```diff
@@ -16,6 +16,7 @@
     "@ckeditor/ckeditor5-core": "44.1.0",
     "@ckeditor/ckeditor5-typing": "44.1.0",
     "@ckeditor/ckeditor5-ui": "44.1.0",
+    "@ckeditor/ckeditor5-utils": "44.1.0",
```
niegowski, line 19: This is used only in tests. I'm not sure if those should not be dev dependencies.

niegowski (changes requested):
Looks like the new lint rule is to greedy and marks imports in test files as production one but that code is not published to npm.

packages/ckeditor5-basic-styles/package.json
filipsobol: It's also used in typings for translations in the dist folder, so it's a production dependency:

```ts
import type { Translations } from '@ckeditor/ckeditor5-utils';
declare const translations: Translations;
export default translations;
```

niegowski (approved):

niegowski (dismissed):

niegowski (approved):

filipsobol:
Continued in #17790.

PR #17728, opened 2025-01-09
Add `aria-checked` attribute to align toolbars in table properties.

### LINKED ISSUE #17722: Buttons in table and cell properties panels are missing aria-pressed attribute

When pressing buttons in table and cell properties panel, for example:

Image
Image

buttons should use `aria-pressed` attribute.

---

Mati365:

Feature (table): Improve aria attributes in table and cell align toolbars. Closes #17722

---

### Additional information

I added role `role=radiogroup` and `role=radio` to table properties radio buttons like this one:

image

It means that `pressed` state of button will be indicated in `aria-checked` attribute.

### DIFF
--- a/packages/ckeditor5-table/src/utils/ui/table-properties.ts
+++ b/packages/ckeditor5-table/src/utils/ui/table-properties.ts
@@ -161,8 +161,6 @@ export function getBorderStyleDefinitions(
  */
 export function fillToolbar<TView extends View, TPropertyName extends keyof TView>(
 	options: {
-		isToggleable?: boolean;
-		role?: string;
 		view: TView;
 		icons: Record<string, string>;
 		toolbar: ToolbarView;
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-table/src/utils/ui/table-properties.ts
```diff
@@ -161,6 +161,8 @@ export function getBorderStyleDefinitions(
  */
 export function fillToolbar<TView extends View, TPropertyName extends keyof TView>(
 	options: {
+		isToggleable?: boolean;
+		role?: string;
```
arkflpc, line 165: Can we name this option, so it's clearer that it's about aria role?

It is a good opportunity to update the API docs of this function.

Nitpick: is it feasible to have string literal union instead of `string` as a type?

Mati365: I dropped these options entirely. It looks like `fillToolbar` creates only radio buttons (as there is single `propertyName`). https://github.com/ckeditor/ckeditor5/pull/17728/commits/51ee7d595ff089f826264e9b5db8bae2beb5f836

PR #17729, opened 2025-01-09
Close SourceEditing view on plugin disabling

l-lejman:

Other (source-editing): The SourceEditing plugin now closes its view upon being disabled, preventing users from remaining stuck in this mode.

---

### Additional information

A change is needed to ensure that SourceEditing is properly closed when CommentsOnly mode is enabled or when SourceEditing is disabled by API calls. These two plugins should not operate simultaneously.

### DIFF
--- a/packages/ckeditor5-source-editing/src/sourceediting.ts
+++ b/packages/ckeditor5-source-editing/src/sourceediting.ts
@@ -10,7 +10,6 @@
 /* global console */
 
 import { type Editor, Plugin, PendingActions } from 'ckeditor5/src/core.js';
-import { CommentsOnly } from '@ckeditor/ckeditor5-comments';
 import { ButtonView, MenuBarMenuListItemButtonView, type Dialog } from 'ckeditor5/src/ui.js';
 import { CKEditorError, createElement, ElementReplacer } from 'ckeditor5/src/utils.js';
 import { formatHtml } from './utils/formathtml.js';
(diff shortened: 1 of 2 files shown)


scofalik:
Sorry for the confusion, as I suggested that maybe source editing should close if the plugin is disabled, but I didn't fully understand how this plugin operates now and what we want to achieve.

Right now, the plugin has some issues, both from:

*   product perspective (e.g. if read-only mode kicks in when you are in source mode, you are stuck in it, also it should be possible to see the source in read-only mode, just not change it), and
*   from code perspective (e.g. the plugin does not have any commands and the logic is written in _unfortunate_ way, e.g. closing source mode is coupled with saving the data).

We won't solve all the problems without a bigger refactor, so let's stick to something that makes sense and is quick.

Right now, when you switch to read-only following things happen in source editing plugin:

1.  Toolbar and menu bar buttons for source mode are disabled. You can't switch to/from the source mode.
2.  Textarea is disabled.

Let's go with exactly same behavior for when the plugin is disabled (e.g. for comments-only mode, note, that comments-only mode will be set to `true` or `false` at the beginning of the editor initialization and likely won't change after).

So, if the plugin gets disabled (comments-only is activated), we will disable the buttons and textareas.

And I think that for this behavior we don't need any change in the plugin, correct?

f1ames:
Thanks @scofalik for looking into this 👍  I agree that Source Editing behavior could use some polishing, anyways...

> Let's go with exactly same behavior for when the plugin is disabled (e.g. for comments-only mode, note, that comments-only mode will be set to `true` or `false` at the beginning of the editor initialization and likely won't change after).
> 
> So, if the plugin gets disabled (comments-only is activated), we will disable the buttons and textareas.

Makes sense for me.

>  (e.g. for comments-only mode, note, that comments-only mode will be set to `true` or `false` at the beginning of the editor initialization and likely won't change after).

The opposite use case - switching to comments-only via API was main reason for this change AFAIR (so user don't get stuck in source mode). So if we go with such assumption as you describe, we can simplify and stick to readonly only.

> And I think that for this behavior we don't need any change in the plugin, correct?

Yes.

@l-lejman please take a look if you agree or have any comments/questions. If not we can proceed as @scofalik suggested.

PR #17741, opened 2025-01-10
Overlapping of lists next to floating tables.

### LINKED ISSUE #8412: Table can overlap with List

pszczesniak:

Type: Message. Closes #8412.

---

### Additional information

This is not 100% super solid solution but some kind of a workaround for this problem. 

Description:

In `Tables` the change were made from inline styles that were responsible for floating to CSS classes. That helped to add `margin-inline-end` with value of `2em` to preserve the space when lists were next to the floating table. Also by this class there is a chance to set on the lists `overflow: hidden` and `padding-inline-start` (starting from the second nesting ONLY) when they are preceded by an element with CSS class mention earlier.

The only potential risk is the `overflow: hidden`, but regular user wont' be able to break anything - without specific CSS that could manipulate the content (like `position: absolute` and manipulating the `top`, `left`, `right` or `bottom`).  

---

Before:
Screenshot 2025-01-10 at 14 58 48

After:
Screenshot 2025-01-10 at 14 57 33

---

⚠️ It's not working with `Export to Word`.

### DIFF
--- a/packages/ckeditor5-list/theme/list.css
+++ b/packages/ckeditor5-list/theme/list.css
@@ -40,7 +40,8 @@
 }
 
 /* When an element (e.g. table) has `style` tag contains `float` `left`/`right` and has a list as a sibling,
-add margin to the end of the element. */
+add margin to the end of the element.
+See https://github.com/ckeditor/ckeditor5/issues/8412 */
 [dir="ltr"] [style*="float:left"]:has( ~ul ),
 [dir="ltr"] [style*="float:left"]:has( ~ol ),
 [dir="rtl"] [style*="float:right"]:has( ~ul ),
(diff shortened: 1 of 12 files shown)


packages/ckeditor5-list/theme/list.css
```diff
@@ -38,3 +38,18 @@
...
+See https://github.com/ckeditor/ckeditor5/issues/8412 */
+.ck-content .table-style-align-right ~ ul ul,
+.ck-content .table-style-align-right ~ ul ol,
+.ck-content .table-style-align-right ~ ol ol,
+.ck-content .table-style-align-right ~ ol ul,
+.ck-content .table-style-align-left ~ ul ul,
+.ck-content .table-style-align-left ~ ul ol,
+.ck-content .table-style-align-left ~ ol ol,
+.ck-content .table-style-align-left ~ ol ul {
+	overflow: hidden;
```
Reinmar, line 53: This is risky. All overflow:hidden caused issues in the past.

Reinmar:
One quick note: Apart from the issues with overflow:hidden, this solution won't handle scenarios like img+h1+ol :(

PR #17747, opened 2025-01-10
Added skin tone config option

### LINKED ISSUE #17736: Define a default skin tone in via config

By default, all emojis use the _Default skin tone_ option as a variant.

While the feature preserves changes when a user updates the variant, it would be good to allow users to define their skin tone via the configuration option.

Hence, let's introduce the `emoji.skinTone` option with supported values: 

```ts
type EmojiSkinTone = 'default' | 'dark' | 'medium-dark' | 'medium' | 'medium-light' | 'light';
```

Perhaps it should be exported from the package to make it visible in API docs.

### Definition of Done

* When the `emoji.skinTone` option is specified, its value is used instead of _default_ when showing the emojis.
* The' default' value is used instead if the option is missing in a configuration.
* When a user modifies the skin tone, the value is preserved (like it's now).
* Add the new option in the feature guide.

---

przemyslaw-zan:

Internal: Added skin tone config option for Emoji plugin. Closes #17736.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-emoji/docs/features/emoji.md
+++ b/packages/ckeditor5-emoji/docs/features/emoji.md
@@ -90,6 +90,20 @@ ClassicEditor
 	.catch( /* ... */ );
 ```
 
+* `skinTone` &ndash; Initial skin tone for the emoji picker.
+
+```js
+ClassicEditor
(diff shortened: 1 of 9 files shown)


pomek (approved):
LGTM.

PR #17753, opened 2025-01-13
Added validator for SVGs

przemyslaw-zan:

Internal: Added CI validator for editor SVG icons.

---

### Additional information


.circleci/template.yml
```diff
@@ -148,6 +148,11 @@ jobs:
           when: always
           name: Check if all packages are exported in the "ckeditor5" package
           command: yarn run check-exports
+      - run:
+          when: always
+          name: Check if all editor icon SVGs are properly defined
+          working_directory: external/ckeditor5-commercial
```
pomek, line 154:
```suggestion
```

pomek:
As we discussed internally, could we extend an existing SVG validator ([`svgo`](https://www.npmjs.com/package/svgo)) by a custom plugin to avoid doing the same thing in different ways?

scripts/utils/svgo-viewBox-plugin.cjs
pomek, line 1: We do not accept camel cases in file names in this project. Pleas, use `viewbox` instead.

scripts/svgo.config.cjs
```diff
@@ -28,6 +30,7 @@ module.exports = {
 		},
 		{ name: 'removeTitle' },
 		{ name: 'removeComments' },
-		{ name: 'removeMetadata' }
+		{ name: 'removeMetadata' },
+		svgoViewBoxPlugin
```
pomek, line 34: It works as expected.

image

However, it turns out that the commercial repository has not executed this validator.

Fortunately, icons will land in a single repository soon so that we can deal with it.

przemyslaw-zan: I'm not sure how to reproduce this issue, could you provide reproduction steps?

pomek (approved):

PR #17758, opened 2025-01-16
Redesign internals of the Emoji feature

### LINKED ISSUE #17378: Polish the emoji plugin UI

### LINKED ISSUE #17662: Do we need `emoji-picker-element` at all?

Currently, the package offers a database (a mechanism to search for emojis) but also registers side effects.

The question I'd like to discuss is to replace the `emoji-picker-element` package with some emoji collection (JSON). That's all we need. We can write a custom query mechanism.

### LINKED ISSUE #17730: Use our CDN to fetch the emoji database
### Definition of Ready

* https://github.com/cksource/ckeditor5-internal/issues/3903 must be resolved.

### Definition of Done

* The emoji plugin uses our CDN instead of jsdelivr.net (it's a default in the [`emoji-picker-element`](https://www.npmjs.com/package/emoji-picker-element) package).
* A user can specify the version of the database. It's `16` or `15`, with the default being `16`.
* It applies to both plugins: autocomplete and dropdown.
* Mention the new option in the feature guide.

### LINKED ISSUE #17732: Emoji misses a default category indicator after playing with the search

In the editor with the Emoji plugin loaded:

1. Open the emoji dropdown.
2. Type whatever.
3. Press `X` to clear the input value.

### ✔️ Expected result

Image

### ❌ Actual result

Pay attention to the bottom border of the active category.

Image

pomek:

Internal (emoji): Improved the design of the emoji feature. Closes #17378.

Internal (emoji): Replaced the `emoji-picker-element` package with a custom implementation of the database feature. Closes #17730.

Internal (emoji): The emoji database is now downloaded from the CKEditor 5 CDN. Closes #17662.

Internal (emoji): After clearing the input value, the _active category_ indicator is now displayed correctly in the emoji picker. Closes #17732.

---

### Additional information


pomek:
TODO: Understand what it means.

```
❌ Importing CSS files that do not exist:
- "/home/circleci/project/packages/ckeditor5-theme-lark/theme/index.css" imports "./ckeditor5-emoji/emojiinfo.css"
- "/home/circleci/project/packages/ckeditor5-theme-lark/dist/index-content.css" imports "./ckeditor5-emoji/emojiinfo.css"
- "/home/circleci/project/packages/ckeditor5-theme-lark/dist/index-editor.css" imports "./ckeditor5-emoji/emojiinfo.css"
- "/home/circleci/project/packages/ckeditor5-theme-lark/dist/index.css" imports "./ckeditor5-emoji/emojiinfo.css"
```

I know that it tries to load a non-existing file. However, I'd like to know if each created CSS file should be included in `theme-lark`.

packages/ckeditor5-emoji/LICENSE.md
```diff
@@ -18,7 +18,7 @@ Where not otherwise indicated, all CKEditor content is authored by CKSource engi
 
 The following libraries are included in CKEditor&nbsp;5 emoji feature under the [Apache License, version 2.0](http://www.apache.org/licenses/LICENSE-2.0):
 
-* emoji-picker-element - Copyright (c) 2020 Nolan Lawson.
+* Fuse.js - Copyright (c) 2017 Kirollos Risk.
```
pomek, line 21:
```suggestion
* emoji-picker-element-data - Copyright (c) 2020 Nolan Lawson.
* Fuse.js - Copyright (c) 2017 Kirollos Risk.
```

packages/ckeditor5-emoji/lang/contexts.json
```diff
@@ -1,6 +1,6 @@
 {
 	"Emoji": "Label of the button that inserts an emoji.",
-	"Show all emoji...": "Button appearing at the end of the list of results of emoji search.",
+	"Show all emojis...": "Button appearing at the end of the list of results of emoji search.",
```
pomek, line 3: @Witoso, please, confirm this one.

packages/ckeditor5-emoji/src/ui/emojipickerview.ts
```diff
@@ -38,49 +44,49 @@ export default class EmojiPickerView extends View<HTMLDivElement> {
...
+	 * An instance of the `EmojiGridView`.
 	 */
-	public infoView: EmojiInfoView;
+	public readonly resultsView: SearchInfoView;
 
 	/**
 	 * @inheritDoc
 	 */
-	constructor( locale: Locale, dropdownPanelContent: DropdownPanelContent ) {
+	constructor( locale: Locale, dropdownPanelContent: EmojiDropdownPanelContent ) {
```
oleq, line 77: This class should take over lots of responsibilities from the controller (plugin). See my comments in the `EmojiPicker` plugin to learn more.

packages/ckeditor5-emoji/src/ui/emojitoneview.ts
```diff
@@ -7,70 +7,69 @@
...
 		);
 
-		this._mainDropdownButton.withText = true;
-		this._mainDropdownButton.label = this._getSkinTone( this.selectedSkinTone ).example;
-		this._mainDropdownButton.tooltip = 'Select skin tone';
+		this.mainDropdownButton.withText = true;
+		this.mainDropdownButton.label = this._getSkinTone( this.skinTone ).icon;
+		this.mainDropdownButton.tooltip = 'Select skin tone';
 
 		addToolbarToDropdown(
```
oleq, line 70: Semantically a toolbar gives you plenty of options that are independent. A list gives you many options but they are related to a specific domain. 

Is the skin tone a toolbar or a list? I'd say the latter. We're selecting one of possible skin options. There's a `addListToDropdown()` helper that will help you out.

BTW: This also affects accessibility.

oleq (changes requested):
The feature looks nice and complete, good job 🙂

I left a bunch of comments focusing on MVC architecture and missing layer/responsibilities separation. Long story short, when you write a MVC code (or any code for what's worth), make sure 
* you reduce coupling between layers,
* each module has one responsibility only ([The Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) from SOLID)

If you want to find out whether the separation is done well, ask yourself a question "How much effort would it take to replace module X in my system?". For instance, to replace your current `EmojiPickerView` with another view, you'd have to re-write the whole `EmojiPicker` plugin because it is so involved in picker view creation and state management. `EmojiPickerView` cannot be tested without the plugin (and the whole editor) because the composition of sub-views and state management goes through the plugin. This is unnecessary. If you identify and limit the points where the modules meet, testing, maintaining, and development become much easier down the road. 

----

Aside from that, I noticed a potential UX issue. When the user starts searching and they type a single character, they get the feedback that no emojis were found where the grid would normally be. This is not the best communication because actually they don't see the results due to our internal logic that expects more than a single character before the results are shown. I know the label of the input includes a hint but still, a search message along the lines of "keep on typing to see the results" would be better. Now it may give the user an impression they did something wrong while they didn't. 

https://github.com/user-attachments/assets/bd5d18b4-f7a2-4761-bf01-9ec8502e2bcd

f1ames (changes requested):
I spent some time going through the UI and how everything is connected, so this review was UI-focused mostly, and quite high-level.

## UI

### Public API

We don’t have really well-defined convention here around public plugin APIs I guess, but usually it’s a good approach to expose main "actions" of the plugin so those can be triggered programmatically. Here, it could be showing emoji picker and maybe inserting emoji.

The 1st one can be done with `editor.plugins.get( 'EmojiPicker' ).showUI()`. I think the more usual approach is to expose a command (UI command as we call it) to do the UI action (in this case, showing emoji picker balloon).

About inserting emoji, I'm not really sure, this is just plain `insertContent` and I'm not sure what value such API could bring 🤷 

### Views in plugin code

Looking at the [_createDropdownPanelContent in emojipicker.ts](https://github.com/ckeditor/ckeditor5/pull/17758/files#diff-9a52a3e57b5d9de64c79be07ae30f68ec75924a03d2735b4617c8b8cbcb40427R259) I'm not sure it should a part of this class. I would expect that `EmojiPickerView` will create all of its subviews and manage them. I know it’s most probably done because there is common stuff in `EmojiPicker` class which is then shared between view parts and might be hard to untangle, but it seems like it's mixing the plugin-view responsibilities.

## Others

1. I noticed in manual test that one emoji renders incorrectly. When inserted it looks like emoji, though I'm not sure the intended one:
image

2. There is some logic to handle emojis db loading failure - empty array is returned and console warn shown. Still, there are further errors when trying to open emoji picker as the code there assumes emojis array will not be empty:
image

packages/ckeditor5-emoji/src/ui/emojigridview.ts
```diff
@@ -92,13 +137,113 @@ export default class EmojiGridView extends View<HTMLDivElement> {
...
+				const emoji = item.skins[ this.skinTone ] || item.skins.default;
+
+				return this._createTile( emoji, item.annotation );
+			} )
+			// Insert new elements.
+			.forEach( item => {
+				this.tiles.add( item );
+				this.focusTracker.add( item );
+			} );
+	}
```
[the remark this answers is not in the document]
pomek: To cover in https://github.com/ckeditor/ckeditor5/issues/17776.

packages/ckeditor5-emoji/src/ui/emojitoneview.ts
pomek: Fixed in ba94394d82e19e52a7fa5da1fe6b54813eef616c.

psmyrek (approved):
LGTM 🎉

martnpaneq (approved):

PR #17766, opened 2025-01-17
Add metadata for `ImageUtils` and `ImageEditing` plugins

Dumluregn:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Internal (image): Add metadata for `ImageUtils` and `ImageEditing` plugins. Closes #17765.

---

PR #17770, opened 2025-01-17
Docs: updating API links in metadata.

godai78:

Docs: updating API links in metadata.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-image/ckeditor5-metadata.json
+++ b/packages/ckeditor5-image/ckeditor5-metadata.json
@@ -274,11 +274,11 @@
 			"path": "src/imageutils.js"
 		},
 		{
-			"name": "Image editing",
-			"className": "ImageEditing",
+			"name": "Picture editing",
+			"className": "PictureEditing",
 			"description": "Provides basic common image engine features.",
 			"docs": "features/images/images-overview.html",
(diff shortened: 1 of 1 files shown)


pomek:
Looks like this PR modifies results merged in #17766. The question is - why?

godai78:
The link is wrong, there is no 'ImageEditing' on 'master', just 'PictureEditing'.

pomek:
Are you sure?

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-image/src/image/imageediting.ts

image

Dumluregn:
The problem is that `ImageEditing` is not directly in `src` dir (but in `src/image`) and the https://ckeditor.com/docs/ckeditor5/latest/framework/architecture/plugins.html building script generates incorrect API docs link. I'm thinking about the best solution.

pomek:
Ok, now I understand what we are talking about :D

Dumluregn:
I used incorrect path in https://github.com/ckeditor/ckeditor5/pull/17766/files#diff-032f9962d70f8d2f9eb026529f405eb720fa23fa9897344e79c954f24d303911R281. Checking if that's not enough to fix the problem.

PR #17796, opened 2025-01-22
Do not initialize Emoji plugins if emoji database is not available

### LINKED ISSUE #17772: Do not initialize Emoji plugins if its database is not available

Shortly, when the `EmojiDatabase` could not fetch the resource from a server, `Emoji*` plugins should not be initialized because the editor has nothing to show.

Right now, it ends with an ugly error when opening a picker.

```
emojigridview.ts:216 Uncaught TypeError: Cannot read properties of undefined (reading 'map')
    at EmojiGridView._updateGrid (emojigridview.ts:216:1)
    at EmojiGridView.filter (emojigridview.ts:168:1)
    at EmojiSearchView.search (emojisearchview.ts:68:1)
    at EmojiPicker.showUI (emojipicker.ts:143:1)
    at ButtonView.<anonymous> (emojipicker.ts:168:1)
    at ButtonView.fire (emittermixin.ts:240:1)
```

`EmojiMention` does not display anything from the database but suggests using a picker that will not show anything more.

Image

### Definition of Ready

* #17758 needs to be merged.

### Technical details

* https://github.com/ckeditor/ckeditor5/blob/f7577fc6358b5bed9c2467e25b2165c17b7cc8ae/packages/ckeditor5-emoji/src/emojidatabase.ts#L80
* https://github.com/ckeditor/ckeditor5/blob/f7577fc6358b5bed9c2467e25b2165c17b7cc8ae/packages/ckeditor5-emoji/src/emojipicker.ts#L96
* https://github.com/ckeditor/ckeditor5/blob/f7577fc6358b5bed9c2467e25b2165c17b7cc8ae/packages/ckeditor5-emoji/src/emojimention.ts#L180

psmyrek:

Internal: Do not initialize Emoji plugins if emoji database is not available. Closes #17772.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-emoji/src/emojidatabase.ts
+++ b/packages/ckeditor5-emoji/src/emojidatabase.ts
@@ -38,12 +38,12 @@ export default class EmojiDatabase extends Plugin {
 	/**
 	 * Emoji database.
 	 */
-	declare private _emojiDatabase: Array<EmojiEntry>;
+	private _emojiDatabase: Array<EmojiEntry>;
 
 	/**
 	 * An instance of the [Fuse.js](https://www.fusejs.io/) library.
 	 */
-	declare private _fuseSearch: Fuse<EmojiEntry> | null;
+	private _fuseSearch: Fuse<EmojiEntry> | null;
 
 	/**
 	 * @inheritDoc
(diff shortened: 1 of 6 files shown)


pomek (approved):
LGTM.

PR #17799, opened 2025-01-22
Emoji mention includes a hint when a query is too short

### LINKED ISSUE #17782: Improve mention dropdown when an input is too short

As the emoji search requires at least two characters to filter the database and show results, I would suggest a slight improvement based on https://github.com/ckeditor/ckeditor5/pull/17758#pullrequestreview-2558915403.

When the provided query is too short, let's display a hint _"keep on typing to see the results"_

Image

The element should not be clickable. When a user "selects" this option, we do nothing.

Once the query is valid (in terms of the database), we switch to our current behavior.

### Definition of Ready

* #17758 needs to be merged.

### Technical details

* https://github.com/ckeditor/ckeditor5/blob/9576023aa978c906f5e9ce9c48201bd72f273baa/packages/ckeditor5-emoji/src/emojimention.ts#L198-L199

pomek:

Internal: Emoji mention includes a hint when a query is too short. Closes #17782.


psmyrek (approved):
LGTM 🎉

PR #17820, opened 2025-01-27
Improved handling the read only mode in the Emoji feature

pomek:

Internal: Connected the menu bar item and the toolbar icon with the `emoji` command, so its state impacts the UI state.

Internal: Use the `insertText` command for text insertion to disable inserting emoji when an editor switches to read-only mode.

Closes https://github.com/cksource/ckeditor5-commercial/issues/6981.

---

### Additional information


psmyrek (approved):
LGTM 🎉

PR #17872, opened 2025-02-03
Replace `lodash-es` with `es-toolkit/compat`

filipsobol:

Other: Replace `lodash-es` with `es-toolkit`. See #16395.

---

You can find more information about why we made this change here: https://github.com/ckeditor/ckeditor5/issues/16395#issuecomment-2399644734


packages/ckeditor5-image/tests/imagestyle/imagestyleediting.js
```diff
@@ -176,7 +176,7 @@ describe( 'ImageStyleEditing', () => {
 				plugins: [ ImageBlockEditing, ImageStyleEditing ]
 			} );
 
-			sinon.assert.calledOnceWithMatch( normalizationSpy, {
+			expect( normalizationSpy.firstCall.args[ 0 ] ).to.deep.equal( {
```
filipsobol, line 179: This change is related to https://github.com/toss/es-toolkit/issues/953 because `sinon` internally tries to use `toString` which is not available. IMO, current `es-toolkit` behavior makes more sense than in `lodash-es`.

packages/ckeditor5-image/src/imageresize/imageresizebuttons.ts
```diff
@@ -308,7 +307,7 @@ export default class ImageResizeButtons extends Plugin {
 					} )
 				};
 
-				const allDropdownValues = map( optionsWithSerializedValues, 'valueWithUnits' );
+				const allDropdownValues = Object.values( optionsWithSerializedValues ).map( option => option.valueWithUnits );
```
filipsobol, line 310: This is the same, so let's just use native APIs.

arkflpc (dismissed):

arkflpc (approved):

PR #17874, opened 2025-02-03
Internal: Update Emoji metadata

filipsobol:

Internal: Update Emoji metadata.

### DIFF
--- a/packages/ckeditor5-emoji/ckeditor5-metadata.json
+++ b/packages/ckeditor5-emoji/ckeditor5-metadata.json
@@ -5,14 +5,21 @@
 			"className": "Emoji",
 			"description": "Allows for inserting emojis via a dropdown or by typing `:` followed by emoji name.",
 			"path": "src/emoji.js",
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-emoji/ckeditor5-metadata.json
```diff
@@ -5,14 +5,21 @@
 			"className": "Emoji",
 			"description": "Allows for inserting emojis via a dropdown or by typing `:` followed by emoji name.",
 			"path": "src/emoji.js",
-			"docs": "features/emoji.html"
+			"docs": "features/emoji.html",
+			"requires": [
+				"EmojiMention",
+				"EmojiPicker"
+			]
```
pomek, line 12:
```suggestion
			"requires": [
				"Mention"
			]
```

Metadata must include soft dependency requirements only, as the hard one (via class constructor) is available in run time (due to passing a reference instead of a string). Hence, I would go with the proposed change.

pomek (approved):
LGTM.

PR #17883, opened 2025-02-05
Internal: Add range exceptions to `check-dependencies-versions-match` script.

filipsobol:

Internal: Add range exceptions to `check-dependencies-versions-match` script.

---

Related to cksource/ckeditor5-commercial#6939

### DIFF
--- a/scripts/ci/check-dependencies-versions-match.mjs
+++ b/scripts/ci/check-dependencies-versions-match.mjs
@@ -184,12 +184,9 @@ function getNewestVersion( packageName, newVersion = '0.0.0', currentMaxVersion 
 		return newVersion;
 	}
 
-	if ( semver.valid( newVersion ) ) {
-		return semver.gt( newVersion, currentMaxVersion ) ? newVersion : currentMaxVersion;
-	}
-
-	const versions = getVersionsList( packageName );
-	const newMaxVersion = semver.maxSatisfying( versions, newVersion );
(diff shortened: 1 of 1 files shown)


scripts/ci/check-dependencies-versions-match.mjs
```diff
@@ -162,14 +180,19 @@ function getExpectedDepsVersions( packageJsons, isCkeditor5Package ) {
...
 
-		return semver.gt( newMaxVersion, currentMaxVersion ) ? newMaxVersion : currentMaxVersion;
+	if ( semver.valid( newVersion ) ) {
+		return semver.gt( newVersion, currentMaxVersion ) ? newVersion : currentMaxVersion;
 	}
 
-	return semver.gt( newVersion, currentMaxVersion ) ? newVersion : currentMaxVersion;
+	const versions = getVersionsList( packageName );
+	const newMaxVersion = semver.maxSatisfying( versions, newVersion );
+	const range = versionExceptions[ packageName ] || '';
```
psmyrek, line 193: It seems that in this line `versionExceptions[ packageName ]` is always `undefined`, because otherwise it would go inside the `if ( versionExceptions[ packageName ] )` condition above.

filipsobol: Good catch, I changed the logic and forgot to update this part.

psmyrek (approved):
LGTM 🎉

PR #17884, opened 2025-02-05
Added support in release scripts for publishing the next pre-release version

psmyrek:

Internal: Added support in release scripts for publishing the next pre-release version.

---

### Additional information

### DIFF
--- a/scripts/release/utils/getreleasedescription.mjs
+++ b/scripts/release/utils/getreleasedescription.mjs
@@ -20,6 +20,12 @@ export default async function getReleaseDescription( cliArguments ) {
 		];
 	}
 
+	if ( cliArguments.nightlyNext ) {
+		return [
+			await releaseTools.getNextPreRelease( '0.0.0-next' ), null
+		];
+	}
+
 	if ( cliArguments.nightly ) {
 		return [
 			await releaseTools.getNextNightly(), null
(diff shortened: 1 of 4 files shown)


scripts/release/utils/getreleasedescription.mjs
```diff
@@ -20,6 +20,12 @@ export default async function getReleaseDescription( cliArguments ) {
 		];
 	}
 
+	if ( cliArguments.nightlyNext ) {
+		return [
+			await releaseTools.getNextPreRelease( '0.0.0-next' ), null
```
pomek, line 25: I wonder if the `@next` dist tag should follow our nightly pattern or bump a major instead.

Random thoughts:

* `@next` usually means _next version_.
	* Perhaps some developer wants to use it if it does not break anything and helps resolve any issue. 
* We want to deploy it nightly.
* Does the `v45.0.0-next.0` block `v45.0.0-alpha.0`? `next > alpha`, so publishing alpha does not make sense.
* `0.0.0-*` pattern looks like a release you should not use for anything except for testing.

Could we agree to use `@nightly-next` and `0.0.0-nightly-next`?

psmyrek: +1 for using `@nightly-next` and `0.0.0-nightly-next`.

pomek: We agreed to follow our nightly format: `0.0.0-nightly-next-{DATE}`.

psmyrek: Addressed in https://github.com/ckeditor/ckeditor5/pull/17884/commits/de1c2cfbcf89731357977794d22bb4b229933039.

pomek (approved):
LGTM.

Let's merge the PR once we release `ckeditor5-dev`.

PR #17902, opened 2025-02-10
Warn about cdn use in emoji

filipsobol:

Internal (emoji): Add the `definitionsUrl` configuration option.

Internal (emoji): Add the `emoji-repository-redundant-version` warning.

Internal (emoji): Add the `emoji-repository-cdn-use` warning.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-emoji/src/emojiconfig.ts
+++ b/packages/ckeditor5-emoji/src/emojiconfig.ts
@@ -64,6 +64,25 @@ export interface EmojiConfig {
 	skinTone?: SkinToneId;
 
 	/**
+	 * The URL from which the emoji definitions should be loaded.
+	 *
+	 * ```ts
+	 *	ClassicEditor
+	 *		.create( editorElement, {
+	 *			plugins: [ Emoji, ... ],
+	 *			emoji: {
+	 *				definitionsUrl: ''
+	 *			}
+	 *		} )
+	 *		.then( ... )
+	 *		.catch( ... );
+	 * ```
+	 *
+	 * @default 'https://cdn.ckeditor.com/ckeditor5/data/emoji/\{version\}/en.json'
+	 */
+	definitionsUrl?: string;
+
+	/**
 	 * The emoji database version.
 	 *
 	 * ```ts
@@ -77,6 +96,9 @@ export interface EmojiConfig {
 	 *		.then( ... )
 	 *		.catch( ... );
 	 * ```
+	 *
+	 * If the {@link module:emoji/emojiconfig~EmojiConfig#definitionsUrl `emoji.definitionsUrl`}
+	 * option is provided, this option should be omitted.
 	 *
 	 * @default 16
 	 */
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-emoji/src/emojiconfig.ts
```diff
@@ -78,6 +97,9 @@ export interface EmojiConfig {
 	 *		.catch( ... );
 	 * ```
 	 *
+	 * If the {@link module:emoji/emojiconfig~EmojiConfig#definitionsUrl `emoji.definitionsUrl`}
+	 * option is provided, this option should be omitted.
```
psmyrek, line 101:
```suggestion
	 * option is provided, `version` is ignored as the defined URL takes precedence over the `version`.
```

packages/ckeditor5-emoji/src/emojiconfig.ts
```diff
@@ -63,6 +63,25 @@ export interface EmojiConfig {
...
+	 *			plugins: [ Emoji, ... ],
+	 *			emoji: {
+	 *				definitionsUrl: ''
+	 *			}
+	 *		} )
+	 *		.then( ... )
+	 *		.catch( ... );
+	 * ```
+	 *
+	 * @default 'https://cdn.ckeditor.com/ckeditor5/data/emoji/\{version\}/en.json'
```
psmyrek, line 81: I'm not sure about default values for `definitionsUrl` and `version`: we do not set those values in the editor config. If they are not provided by the integrator, they remain `undefined`.

packages/ckeditor5-emoji/src/emojirepository.ts
```diff
@@ -219,24 +220,97 @@ export default class EmojiRepository extends Plugin {
...
+	 * Returns the URL from which the emoji repository is downloaded. If the URL is not provided
+	 * in the configuration, the default URL is used with the version from the configuration.
+	 *
+	 * If both the URL and version are provided, a warning is logged.
+	 */
+	private _getUrl(): URL {
+		const { definitionsUrl, version } = this.editor.config.get( 'emoji' )!;
+
+		if ( !definitionsUrl || definitionsUrl === 'cdn' ) {
+			// Url was not provided or is set to 'cdn', so we use the default CDN URL.
```
psmyrek, line 233: To keep the naming consistent.

```suggestion
			// URL was not provided or is set to 'cdn', so we use the default CDN URL.
```

packages/ckeditor5-emoji/src/emojirepository.ts
```diff
@@ -11,13 +11,15 @@ import Fuse from 'fuse.js';
 import { groupBy } from 'lodash-es';
 
 import { type Editor, Plugin } from 'ckeditor5/src/core.js';
-import { logWarning, version } from 'ckeditor5/src/utils.js';
+import { logWarning, version as EditorVersion } from 'ckeditor5/src/utils.js';
```
psmyrek, line 14: Why `EditorVersion` in uppercase?

psmyrek (commented):
It works 🎉 

--- 

Manual test currently allows configuring supported options in the test itself. Do we want to add the ability to provide custom URL as well?

packages/ckeditor5-emoji/src/emojiconfig.ts
filipsobol: It's still the default here: https://github.com/ckeditor/ckeditor5/pull/17902/files#diff-bcad77073801aff462771b1e2b907a1f3006cea6057f28fa11504ea853cca0c9R235

gorzelinski (approved):

PR #17920, opened 2025-02-12
Warn about missing TableLayout in email integration

mmotyczynska:

Internal: Editor should warn about missing `TableLayout` plugin in email integration.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-email/src/integrations/table.ts
+++ b/packages/ckeditor5-email/src/integrations/table.ts
@@ -69,7 +69,7 @@ export default class TableEmailIntegration extends Plugin {
 		if ( !plugins.has( 'TableLayout' ) ) {
 			/**
 			 * The `TableLayout` plugin is required to use tables in the email integration.
-			 * Tables may not be exported or displayed correctly in some email clients without it.
+			 * Tables in the editor may not look similar to how they would look in email clients.
 			 *
 			 * @error email-integration-missing-table-layout-plugin
 			 */
--- a/packages/ckeditor5-email/src/integrations/table.ts
+++ b/packages/ckeditor5-email/src/integrations/table.ts
@@ -41,18 +41,22 @@ export default class TableEmailIntegration extends Plugin {
 	 * @inheritDoc
 	 */
 	public afterInit(): void {
-		this._checkPlainTableOutputPlugin();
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-email/src/integrations/table.ts
```diff
@@ -41,18 +41,22 @@ export default class TableEmailIntegration extends Plugin {
...
-		this._checkPlainTableOutputPlugin();
+		this._checkRequiredTablePlugins();
 		this._checkTableConfig();
 	}
 
 	/**
 	 * Checks if the `PlainTableOutput` plugin is available when the `Table` plugin is enabled.
 	 */
-	private _checkPlainTableOutputPlugin(): void {
+	private _checkRequiredTablePlugins(): void {
```
pszczesniak, line 51: I think that description should also be changed to reflect what function does now.

packages/ckeditor5-email/src/integrations/table.ts
```diff
@@ -61,6 +65,16 @@ export default class TableEmailIntegration extends Plugin {
 			 */
 			utils._logSuppressibleWarning( 'email-integration-missing-plain-table-output-plugin' );
 		}
+
+		if ( !plugins.has( 'TableLayout' ) ) {
+			/**
+			 * The `TableLayout` plugin is required to use tables in the email integration.
+			 * Without it, tables will not be exported correctly and may not be displayed correctly in some email clients.
```
pszczesniak, line 72: Im wondering about it cause description is not reflects what this plugin is needed for.

Maybe something like this? But im not fully convinced :thinking: :

```
The `TableLayout` plugin is required to use tables in `layout` mode in the email integration. Without it, tables in the editor won't look similar like they will be presented in the email clients, that may be confusing for editor's users.
```

pszczesniak (changes requested):

packages/ckeditor5-email/src/integrations/table.ts
gorzelinski: I tweaked the message a bit. It may be a bit better understand 🤔

Mati365 (approved):

Mati365 (approved):

pszczesniak (approved):

PR #17924, opened 2025-02-13
Table layout UI

mmotyczynska:

Internal: Introduced Table layout UI.

---

### Additional information

Please only check the UI part (e.g. command is only added because we still don't have the editing part, I left some TODO's for that).
Updating icon is extracted to a separate issue.


packages/ckeditor5-table/tests/tablecaption/toggletablecaptioncommand.js
```diff
@@ -57,6 +57,14 @@ describe( 'ToggleTableCaptionCommand', () => {
 			);
 			expect( command.isEnabled ).to.be.true;
 		} );
+
+		it( 'should be false if in the table that does not allow captions', () => {
```
gorzelinski, line 61:
```suggestion
		it( 'should be false if it is in a table that does not allow captions', () => {
```

packages/ckeditor5-table/tests/tablelayout/tablelayoutui.js
```diff
@@ -0,0 +1,212 @@
...
+	describe( 'insertTableLayout dropdown', () => {
+		let insertTableLayout;
+
+		beforeEach( () => {
+			insertTableLayout = editor.ui.componentFactory.create( 'insertTableLayout' );
+			insertTableLayout.render();
+
+			document.body.appendChild( insertTableLayout.element );
+
+			insertTableLayout.isOpen = true; // Dropdown is lazy loaded, so make sure it's open (#6193).
```
gorzelinski, line 58: Don't we want to change the issue number to a URL?

mmotyczynska: I copied that from other tests 😉 But you are right, why not make it a url? 👍

Mati365 (approved):
Looks fine :+1:

packages/ckeditor5-table/src/commands/inserttablelayoutcommand.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license Copyright (c) 2003-2025, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-licensing-options
+ */
+
+/**
+ * @module table/commands/inserttablecommand
```
niegowski, line 7: This should match the filename.

mmotyczynska: This file will be replaced with the one from editing part.

PR #17931, opened 2025-02-14
`TableLayout` plugin.

pszczesniak:

Feature(table): Introduce the `TableLayout` plugin.

---

### Additional information


packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,212 @@
...
+			const viewTable = data.viewItem;
+
+			// When element was already consumed then skip it.
+			if ( !conversionApi.consumable.test( viewTable, { name: true } ) ) {
+				return;
+			}
+
+			const { rows, headingRows, headingColumns } = scanTable( viewTable );
+
+			// // Only set attributes if values is greater then 0.
```
mmotyczynska, line 135:
```suggestion
			// Only set attributes if values are greater then 0.
```

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,212 @@
...
+					modelRange
+				);
+			}
+		}, { priority: 'low' } );
+	};
+}
+
+/**
+ * Checks if the table is a content table.
+ * Returns `true` if on of below conditions is met:
```
mmotyczynska, line 200:
```suggestion
 * Returns `true` if any of the following conditions are met:
```

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,212 @@
...
+ * - the `<table>` is wrapped with `< figure>`,
```
mmotyczynska, line 201:
```suggestion
 * - the `<table>` is wrapped with `<figure>`,
```

mmotyczynska (changes requested):
Partial review.

Please check the table header, as it does not seem to work correctly in content tables. When switching a row to a header row for the first time, the model updates, but the switch reverts to the off state. Then, after a second click, the switch finally turns on, but as a result, a second row is also converted into a header row.

https://github.com/user-attachments/assets/32a29660-e4bd-44df-a04e-c0738187baa8

packages/ckeditor5-table/src/commands/setheadercolumncommand.ts
```diff
@@ -92,4 +98,17 @@ export default class SetHeaderColumnCommand extends Command {
...
+	 * Checks if the table is a content table.
+	 */
+	private _isTableTypeContent( selectedCells: Array<Element> ): boolean {
+		const table = selectedCells[ 0 ].findAncestor( 'table' )!;
+
+		if ( table.getAttribute( 'tableType' ) == 'layout' ) {
+			return false;
+		}
+
+		return true;
```
Mati365, line 112: Can be simplified to 

```suggestion
		return table.getAttribute( 'tableType' ) !== 'layout';
```

packages/ckeditor5-table/src/tablelayout.ts
```diff
@@ -20,6 +24,13 @@ export default class TableLayout extends Plugin {
 		return 'TableLayout' as const;
 	}
 
+	/**
+	 * @inheritDoc
+	 */
+	public static get requires() {
+		return [ Table, PlainTableOutput, TableLayoutEditing ] as const;
```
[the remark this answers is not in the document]
pszczesniak: Good catch :+1:

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,212 @@
...
+
+	/**
+	 * Defines the converters for the table layout feature.
+	 */
+	private _defineConverters() {
+		const { editor } = this;
+		const { conversion } = editor;
+
+		conversion.for( 'upcast' ).add( upcastTableOverride() );
+		conversion.for( 'upcast' ).add( upcastTableToSetTableType() );
```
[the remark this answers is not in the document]
niegowski: Or even maybe we could use a single helper to register both converters.

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,212 @@
...
+ * This conversion helper overrides the default table converter to meet table layout conditions.
+ *
+ * @returns Conversion helper.
+ */
+function upcastTableOverride() {
+	return ( dispatcher: UpcastDispatcher ): void => {
+		dispatcher.on<UpcastElementEvent>( 'element:table', (
+			evt: EventInfo,
+			data: UpcastConversionData<ViewElement>,
+			conversionApi: UpcastConversionApi
```
niegowski, line 124:
```suggestion
			data: UpcastConversionData<ViewElement>,
			conversionApi: UpcastConversionApi
```
Are those types required or are infered from the `<UpcastElementEvent>`?

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,212 @@
...
+		) => {
+			const viewTable = data.viewItem;
+
+			// When element was already consumed then skip it.
+			if ( !conversionApi.consumable.test( viewTable, { name: true } ) ) {
+				return;
+			}
+
+			const { rows, headingRows, headingColumns } = scanTable( viewTable );
```
niegowski, line 133: Instead of redoing what the table converter is doing, I'd check if this should be handled as a layout table and handle it here. Otherwise, I'd just allow the original converter to handle the table.

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,212 @@
...
+				return;
+			}
+
+			conversionApi.consumable.consume( viewTable, { name: true } );
+
+			// Upcast table rows in proper order (heading rows first).
+			rows.forEach( row => conversionApi.convertItem( row, conversionApi.writer.createPositionAt( table, 'end' ) ) );
+
+			// Convert everything else.
+			conversionApi.convertChildren( viewTable, conversionApi.writer.createPositionAt( table, 'end' ) );
```
niegowski, line 160: Please check how table columns are upcasted so they are at the end of table as model expects.

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * Sets only the table type attribute.
+ */
+function upcastTableToSetTableType() {
+	return ( dispatcher: UpcastDispatcher ): void => {
+		dispatcher.on<UpcastElementEvent>( 'element:table', (
+			evt: EventInfo,
+			data: UpcastConversionData<ViewElement>,
+			conversionApi: UpcastConversionApi
```
niegowski, line 183: Those types probably are not needed as they should be inferred.

packages/ckeditor5-table/src/converters/upcasttable.ts
```diff
@@ -199,8 +199,10 @@ function getViewTableFromFigure( figureView: ViewElement ) {
  * headingRows    - The number of rows that go as table headers.
  * headingColumns - The maximum number of row headings.
  * rows           - Sorted `<tr>` elements as they should go into the model - ie. if `<thead>` is inserted after `<tbody>` in the view.
+ *
+ * @internal
  */
-function scanTable( viewTable: ViewElement ) {
+export function scanTable( viewTable: ViewElement ): { headingRows: number; headingColumns: number; rows: Array<ViewElement> } {
```
niegowski, line 205: This is not needed.

packages/ckeditor5-table/src/utils/common.ts
```diff
@@ -102,3 +102,14 @@ export function getSelectionAffectedTable( selection: DocumentSelection ): Eleme
...
+/**
+ * Based on selected table cells, checks if the table is a content table or not.
+ *
+ * @param selectedCells Selected table cells.
+ */
+export function isTableTypeContent( selectedCells: Array<Element> ): boolean {
+	const table = selectedCells[ 0 ].findAncestor( 'table' )!;
+
+	return table.getAttribute( 'tableType' ) !== 'layout';
+}
```
niegowski, line 115: This looks a bit overcomplicated for a single-purpose helper. Please use other common helpers or inline attribute checks where needed.

packages/ckeditor5-table/src/commands/setheaderrowcommand.ts
```diff
@@ -43,10 +43,16 @@ export default class SetHeaderRowCommand extends Command {
...
 		const tableUtils: TableUtils = this.editor.plugins.get( 'TableUtils' );
 		const model = this.editor.model;
+
 		const selectedCells = tableUtils.getSelectionAffectedTableCells( model.document.selection );
 		const isInTable = selectedCells.length > 0;
+		let isContentTable = true;
+
+		if ( isInTable ) {
+			isContentTable = isTableTypeContent( selectedCells );
+		}
```
niegowski, line 53: We should check if schema allows `headingRows` attribute instead of coupling those features.

packages/ckeditor5-table/src/commands/setheadercolumncommand.ts
```diff
@@ -44,13 +45,18 @@ export default class SetHeaderColumnCommand extends Command {
...
 		const tableUtils: TableUtils = this.editor.plugins.get( 'TableUtils' );
+		const model = this.editor.model;
 
 		const selectedCells = tableUtils.getSelectionAffectedTableCells( model.document.selection );
 		const isInTable = selectedCells.length > 0;
+		let isContentTable = true;
+
+		if ( isInTable ) {
+			isContentTable = isTableTypeContent( selectedCells );
+		}
```
niegowski, line 57: Same as in rows.

packages/ckeditor5-table/src/utils/structure.ts
```diff
@@ -59,8 +59,17 @@ export function cropTableToDimensions(
...
-	// Create empty table with empty rows equal to crop height.
+	// Initialize the cropped table element.
 	const croppedTable = writer.createElement( 'table' );
+
+	// Copy table type attribute if present.
+	const sourceTableType = sourceTable.getAttribute( 'tableType' );
+
+	if ( sourceTableType ) {
+		writer.setAttribute( 'tableType', sourceTableType, croppedTable );
+	}
```
niegowski, line 70: This looks like a tight coupling. What should happen with `tableWidth` attribute?

niegowski (changes requested):

packages/ckeditor5-table/src/utils/structure.ts
Mati365: It handles the case when user selects part of `layout` table and then paste it directly to the editing. Without picking attributes it'll not work. `tableWidth` - dunno, while it makes to sense to copy it too.

packages/ckeditor5-table/tests/tablelayout/tablelayoutediting.js
```diff
@@ -0,0 +1,975 @@
...
+						'<tbody>' +
+							'<tr><td>Foo</td></tr>' +
+						'</tbody>' +
+					'</table>'
+				);
+			} );
+
+			it( 'should convert to content table if paste from external', () => {
+				const dataTransferMock = createDataTransfer( {
+					'text/html': '<figure><table><tbody><tr><td>Foo</td></tr></tbody></table></figure>'
```
niegowski, line 812:
```suggestion
					'text/html': '<table><tbody><tr><td>Foo</td></tr></tbody></table>'
```

Mati365: @niegowski I restored clipboard handler in this https://github.com/ckeditor/ckeditor5/pull/17931/commits/f383923035c5b64c551e6314c24d29bb06b4a2ed. Let me know if it fixes the issue.

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -204,13 +249,9 @@ function isTableTypeContent( viewTable: ViewElement ): boolean {
...
  * It is used to disallow attributes or children that is managed by `Schema`.
  */
-function layoutTableCheck() {
-	return ( context: SchemaContext ) => {
-		if ( context.endsWith( 'table' ) && context.last.getAttribute( 'tableType' ) == 'layout' ) {
-			return false;
-		}
-	};
+function layoutTableCheck( context: SchemaContext ) {
+	return !( context.endsWith( 'table' ) && context.last.getAttribute( 'tableType' ) == 'layout' );
```
niegowski, line 256: I'm not sure if returning `true` would not allow on every other element. Please check.

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -0,0 +1,257 @@
...
+			const { mapper, writer } = conversionApi;
+
+			if ( !conversionApi.consumable.test( item, evt.name ) ) {
+				return;
+			}
+
+			const table = mapper.toViewElement( item );
+
+			writer.addClass( `${ attributeNewValue }-table`, table );
+			writer.setAttribute( 'role', 'presentation', table );
```
niegowski, line 228: This should be set only for layout tables.

pszczesniak:
Added postfxer that takes care of adding `tableLayout` type to tables created using `insertTable` command.

niegowski (changes requested):

niegowski (approved):

PR #17849, opened 2025-01-30
Email integration package

Mati365:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Feature (email): Initialize email integration package. 

---

### Additional information

// TODO

PR #17939, opened 2025-02-17
Add basic clipboard support for table layout editing.

Mati365:

Internal (table): Fix missing `tableType` attribute in `cropTableToDimensions` output. 

---

### Additional information

Parent PR: https://github.com/ckeditor/ckeditor5/pull/17931
Grandparent PR: https://github.com/ckeditor/ckeditor5/pull/17849


Witoso:
>T ables pasted from a different editor instance are always converted to content tables, regardless of their original type.

Why is that? I think it should behave same as in one editor. Only `external` tables (not from editor), should behave like content.

Issue #17940, opened 2025-02-17
Add methods for disabling and enabling menu bar


Dumluregn:
## 📝 Provide a description of the new feature

Now you'd have to do this separately for each menu bar menu.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

Issue #17915, opened 2025-02-12
Disabled menu bar menu can still be open using keyboard


Dumluregn:
## 📝 Provide detailed reproduction steps (if any)

1. Open any demo with menu bar, e.g. https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/full-featured-editor.html
2. Disable a menu from menu bar typing `editor.ui.view.menuBarView.menus[ 0 ].isEnabled = false;` in the console.
3. Focus the editor.
4. Press `Opt+F9` to focus menu bar.
5. Press "arrow down" when the disabled menu is focused.

### ✔️ Expected result

Nothing happens, menu should not be open.

### ❌ Actual result

Menu opens normally.

![Image](https://github.com/user-attachments/assets/e5ab1316-a9c5-44f3-b691-ce22e4a63c34)

### ❓ Possible solution

_If you have ideas, you can list them here. Otherwise, you can delete this section._

## 📃 Other details

* Browser: …
* OS: …
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

PR #17941, opened 2025-02-17
Add integration for fullscreen feature with revision history

Dumluregn:

Internal (fullscreen): Add integration with revision history.

Fix (ui): Do not open disabled menu bar menu on arrow down press. Closes https://github.com/ckeditor/ckeditor5/issues/17915.

Feature (ui): Add `MenuBarView#disable()` and `MenuBarView#enable()` methods. They disable/enable all top-level menus in menu bar. Closes https://github.com/ckeditor/ckeditor5/issues/17940.

Internal (ui): Add a `toolbar` property to the `EditorUIView` class for coherent access for all editor types.

Internal (ui): Make `EditorUI#_initMenuBar()` method public and rename it accordingly to allow for runtime toolbar creation.

---

### Additional information

* Requires `ck/fullscreen-revision-history` in the related repository (one of the used types needs to be exported).
* Disabling menu bar was supposed to be done through a flag. However looking at API I proposed to use `disable()` and `enable()` methods instead as it will be simpler and more coherent with the rest of the component code.
* Following the idea to have as little code in separate editor handlers (https://github.com/ckeditor/ckeditor5/pull/17897#discussion_r1952523571), most of the RH integration was put in the `AbstractEditorHandler` class. It should be treated as a go-to class, not the one that only throws errors for unsupported editor types.
* I added a `RevisionHistoryMock` class in unit tests providing as much integration as I found useful. It's arbitrary and I'm open for suggestions.
* `EditorUIView#toolbar` was introduced to make the editor types UIs more coherent and better accessible. Without it here we'd have to use `as` keyword to narrow the editor types used in `AbstractEditorHandler`.
* As discussed, for now the CC for revision history integration implementation details is skipped. We're waiting for some decisions regarding how to approach the problem of testing premium features integration in the public repo.

### DIFF
--- a/packages/ckeditor5-editor-classic/src/classiceditoruiview.ts
+++ b/packages/ckeditor5-editor-classic/src/classiceditoruiview.ts
@@ -27,7 +27,7 @@ export default class ClassicEditorUIView extends BoxedEditorUIView {
 	/**
 	 * Toolbar view instance.
 	 */
-	public readonly toolbar: ToolbarView;
+	public override readonly toolbar: ToolbarView;
 
 	/**
 	 * Editable UI view.
--- a/packages/ckeditor5-editor-decoupled/src/decouplededitoruiview.ts
+++ b/packages/ckeditor5-editor-decoupled/src/decouplededitoruiview.ts
@@ -25,7 +25,7 @@ export default class DecoupledEditorUIView extends EditorUIView {
 	/**
 	 * The main toolbar of the decoupled editor UI.
 	 */
-	public readonly toolbar: ToolbarView;
+	public override readonly toolbar: ToolbarView;
 
 	/**
 	 * The editable of the decoupled editor UI.
--- a/packages/ckeditor5-editor-inline/src/inlineeditoruiview.ts
+++ b/packages/ckeditor5-editor-inline/src/inlineeditoruiview.ts
@@ -34,7 +34,7 @@ export default class InlineEditorUIView extends EditorUIView {
 	/**
 	 * A floating toolbar view instance.
 	 */
-	public readonly toolbar: ToolbarView;
+	public override readonly toolbar: ToolbarView;
 
 	/**
 	 * The offset from the top edge of the web browser's viewport which makes the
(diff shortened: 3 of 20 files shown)


f1ames (changes requested):
I tested it and it works fine 👍 

Couple of issues to check:
1. CI is failing. I restarted it but still failing, can you take a look?
2. Please check the review comments.
3. I found one edge case with RH and fullscreen mode (see video 1 below).
4. Disabled menu bar focus behaves a bit inconsistently (see video 2 below).

## RH + fullscreen command

When you open RH viewer and then switch to fullscreen mode with a command, the fullscreen mode editor is kind of in RH state (toolbar buttons disabled, editor read-only), but the UI is not there.

https://github.com/user-attachments/assets/1aa988bb-75f5-4168-94fd-0c912cca276d

It's kind of an edge case (and we don't have UI for switching to fullscreen from RH directly), but still I imagine it might be something that an integrator could try to do.

As a bonus (noticed after watching the recording), there is RH balloon left visible when switched to fullscreen (last 10 seconds of the video).

## Disabled menu bar focus indicator

When switching to RH in fullscreen mode for the first time, it seems the menu bar items do not have a focus indicator (clicking on them does not show it). However, when I click and press the left/right arrow, the focus indicator appears. After that, it shows on hover and does not disappear when it should.

https://github.com/user-attachments/assets/a301dd21-9963-432a-b791-504c3f604ede

Dumluregn:
> When switching to RH in fullscreen mode for the first time, it seems the menu bar items do not have a focus indicator (clicking on them does not show it). However, when I click and press the left/right arrow, the focus indicator appears. After that, it shows on hover and does not disappear when it should.

As discussed, this behavior is the same for non-disabled menu bar - so it should probably be fixed globally (if at all): https://github.com/ckeditor/ckeditor5/issues/18019.

> When you open RH viewer and then switch to fullscreen mode with a command, the fullscreen mode editor is kind of in RH state (toolbar buttons disabled, editor read-only), but the UI is not there.

Unfortunately there's no access to revision viewer instance at the moment when we enter fullscreen and we don't really want to expose it yet, so we can't easily move the RH UI to the fullscreen. Instead let's just hide RH (there are changes in the related PR necessary).

f1ames (changes requested):
Looks good 👍

Just one tiny detail to fix which I missed earlier 🙊 🙏 

As for the CI, @Dumluregn checked it and it's an issue with our job not using the same branch from commercial repo. The other one is related to icon which will be changed anyways.

f1ames (approved):
👍

PR #17943, opened 2025-02-17
Fix icons typings

filipsobol:

Internal: Fix icon typings.

---

No issue because everything is already described in the script.


pomek (approved):
LGTM.

PR #17948, opened 2025-02-18
Split build tools into smaller modules

psmyrek:

Internal: Split build utils for NIM into smaller re-usable modules.

---

### Additional information


scripts/nim/utils.mjs
```diff
@@ -0,0 +1,87 @@
...
+
+		/**
+		 * Because this build runs first, it cleans up the old output folder
+		 * and generates TypeScript declarations and translation files.
+		 * We don't want to repeat this in other steps.
+		 */
+		clean: true,
+		translations: 'packages/**/*.po'
+	} );
+}
```
filipsobol, line 34: How about adding a `overrides` parameter that is destructured at the end of every function? This way, each function will come with defaults which can be easily overriden.

```suggestion
export function initializeCKEditor5NpmBuild( overrides = {} ) {
	return build( {
		output: dist( 'ckeditor5.js' ),
		tsconfig,
		banner,
		sourceMap: true,
		external: [],

		/**
		 * Because this build runs first, it cleans up the old output folder
		 * and generates TypeScript declarations and translation files.
		 * We don't want to repeat this in other steps.
		 */
		clean: true,
		translations: 'packages/**/*.po',
		...overrides
	} );
}
```

psmyrek: Good idea, addressed in https://github.com/ckeditor/ckeditor5/pull/17948/commits/281bfc720399489c8facfe0c4b400b658bc8baed.

PR #17956, opened 2025-02-19
Docs: adding feature information.

### LINKED ISSUE #17921: Document in Restricted editing that it's meant to work only with inline changes in the "unrestricted areas"
The Restricted editing feature, in the "restricted" mode supports only inline-type of changes. You can type, you can delete content, you can format text. But enter key (splitting paragraphs), inserting tables or inserting block images is not supported. This isn't mentioned anywhere in that guide.

The other thing that's missing is that the `allowedCommands` option should not be configured with commands inserting or affecting block content. Such as `insertTable` or `enter` commands. This needs to be mentioned in:

* https://ckeditor.com/docs/ckeditor5/latest/features/restricted-editing.html#configuration
* https://ckeditor.com/docs/ckeditor5/latest/api/module_restricted-editing_restrictededitingconfig-RestrictedEditingConfig.html

godai78:

Docs: adding feature information. Closes #17921

---

### Additional information

### DIFF
--- a/packages/ckeditor5-restricted-editing/docs/features/restricted-editing.md
+++ b/packages/ckeditor5-restricted-editing/docs/features/restricted-editing.md
@@ -106,6 +106,10 @@ ClassicEditor
 
 **Note**: Typing and deleting text is always possible in restricted editing regions. For more information, check out the {@link module:restricted-editing/restrictededitingconfig~RestrictedEditingConfig `config.restrictedEditing`} documentation.
 
+<info-box warning>
+	Only inline content inserting or editing commands are allowed in this setting. Block content commands such as `insertTable` or `enter` cannot be allowed via this setting, as they are not supported in the restricted editing mode. 
+</info-box>
+
 ### Enabling commands in the restricted editing mode
 
 The restricted editing mode allows modifying the editor content only in designated regions. Outside these regions, most of the editor commands are turned off by default. If you wish to enable some commands outside the restricted editing regions, you can use the {@link module:restricted-editing/restrictededitingmodeediting~RestrictedEditingModeEditing#enableCommand `RestrictedEditingModeEditing.enableCommand()`} method. You must execute this method in the {@link module:core/plugin~PluginInterface#afterInit `afterInit()`} callback of an editor plugin.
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-restricted-editing/docs/features/restricted-editing.md
```diff
@@ -106,6 +106,10 @@ ClassicEditor
 
 **Note**: Typing and deleting text is always possible in restricted editing regions. For more information, check out the {@link module:restricted-editing/restrictededitingconfig~RestrictedEditingConfig `config.restrictedEditing`} documentation.
 
+<info-box warning>
+	Only inline content inserting or editing commands are allowed in this setting. It means, block content commands such as `insertTable` or `enter` cannot be allowed via this setting, as they are not supported in the restricted editing mode.
```
gorzelinski, line 110:
```suggestion
	Only inline content inserting or editing commands are allowed in this setting. Block content commands such as `insertTable` or `enter` cannot be allowed via this setting, as they are not supported in the restricted editing mode. 
```

packages/ckeditor5-restricted-editing/src/restrictededitingconfig.ts
```diff
@@ -44,7 +44,10 @@ export interface RestrictedEditingConfig {
 	 * };
 	 * ```
 	 *
-	 * To make a command always enabled (also outside non-restricted areas) use
+	 * Only inline content inserting or editing commands are allowed in this setting. It means, block content commands such as
```
gorzelinski, line 47:
```suggestion
	 * Only inline content inserting or editing commands are allowed in this setting. Block content commands such as
```

packages/ckeditor5-restricted-editing/docs/features/restricted-editing.md
Witoso: This part should be moved below the `Enabling commands in the restricted editing mode` header as it's directly relates to this part.

godai78: @Witoso I moved it, but I'm not sure this should actually be moved. Where it was, it corresponded to commands enabled in restricted areas. Where it is now, it relates to global regions. Please build and have another look at it as it is now, as I'm not sure we should move it.

Witoso: it's good!

Witoso: Ok, I was perphaps wron, re-reading..

Witoso: You were right, we need to move it back, above this heading, sorry!

Witoso (changes requested):
Move infobox back up to the Configuration.

Witoso (approved):

PR #17970, opened 2025-02-21
Add fullscreen mode integration with comments and suggestions

Dumluregn:

Internal (fullscreen): Add integration with comments and suggestions sidebar.

---

### Additional information

Merge https://github.com/ckeditor/ckeditor5/pull/17897 first.

### DIFF
--- a/packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
+++ b/packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
@@ -7,7 +7,7 @@
  * @module fullscreen/handlers/abstracteditorhandler
  */
 
-import { CKEditorError, createElement } from 'ckeditor5/src/utils.js';
+import { createElement } from 'ckeditor5/src/utils.js';
 import { type Editor } from 'ckeditor5/src/core.js';
 
 /**
@@ -52,23 +52,6 @@ export default class AbstractEditorHandler {
 	}
 
 	/**
-	 * Returns the moved elements to their original places.
-	 */
-	public returnMovedElements(): void {
-		this._movedElements.forEach( ( placeholder, moved ) => {
-			placeholder.replaceWith( moved );
-			placeholder.remove();
-		} );
-
-		this._movedElements.clear();
-
-		if ( this._container ) {
-			this._container.remove();
-			this._container = null;
-		}
-	}
-
-	/**
 	 * Returns the fullscreen mode container element.
 	 */
 	public getContainer(): HTMLElement {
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -122,4 +140,41 @@ export default class AbstractEditorHandler {
...
+		// Switch to the wide sidebar.
+		const sidebarPlugin = this._editor.plugins.get( 'Sidebar' ) as Sidebar;
+
+		if ( !sidebarPlugin.container ) {
+			sidebarPlugin.setContainer(
+				this.getContainer().querySelector( '[data-ck-fullscreen="right-sidebar"]' ) as HTMLElement
+			);
+		}
+
+		annotationsUIs.switchTo( 'wideSidebar' );
```
DawidKossowski, line 161: I'm not sure about this. It will display all annotations, ignoring any filters. Imagine that someone has a filter that hides very important annotations that should not be visible to others.

My proposal:

1. We check if someone has a filter defined for `wideSidebar`. If so, we retrieve and apply it in fullscreen.
2. If no filter is defined for `wideSidebar`, we read the filters for the active display mode and apply them.
3. If no filters are defined for the active display mode, we proceed without applying a filter.

DawidKossowski (changes requested):

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -122,4 +140,70 @@ export default class AbstractEditorHandler {
 			this._container = null;
 		}
 	}
+
+	/**
+	 * Stores the current state of the annotations UIs to restore it when leaving fullscreen mode.
+	 * Code coverage is provided in the commercial package repository as integration unit tests.
```
DawidKossowski, line 146: Could you move the info about code coverage outside the API docs, just above the Istanbul ignore?

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -122,4 +140,70 @@ export default class AbstractEditorHandler {
...
+		}
+
+		const annotationsFilters = new Map<string, ( annotation: Annotation ) => boolean>();
+
+		for ( const [ uiName, data ] of [ ...this.annotationsUIsData! ] ) {
+			// Default filter is `() => true`. Only store filters that are different.
+			if ( data.filter.toString() !== '() => true' ) {
+				annotationsFilters.set( uiName, data.filter );
+			}
+		}
```
DawidKossowski, line 170: I think we should have an API in the `Annotations` plugin for that. Anyway, for now, we can ignore it. However, I would create a follow-up task to rethink and improve it.

Dumluregn: https://github.com/cksource/ckeditor5-commercial/issues/7153

Dumluregn:
I added a small commit at the end integrating annotations sidebar with RH (hiding sidebar when RH is open and restoring it after RH is closed).

PR #17975, opened 2025-02-24
Presence list and document outline for fullscreen mode

l-lejman:

---

### Additional information


packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -122,4 +126,55 @@ export default class AbstractEditorHandler {
...
+		const presenceListWrapper = `
+			<div class="ck ck-fullscreen__left-sidebar-item">
+				<div class="ck ck-fullscreen__left-sidebar-header">Connected users</div>
+				<div class="ck ck-fullscreen__presence-list" data-ck-fullscreen="presence-list"></div>
+			</div>
+		`;
+
+		const fragment = document.createRange().createContextualFragment( presenceListWrapper );
+
+		document.querySelector( '[data-ck-fullscreen="left-sidebar-sticky"]' )!.appendChild( fragment );
```
Dumluregn, line 144: Maybe for consistency across this one file we'll keep the convention of creating an element and setting its `innerHTML` here and later in DO?

packages/ckeditor5-fullscreen/src/handlers/classiceditor.ts
```diff
@@ -35,6 +35,8 @@ export default class ClassicEditorHandler extends AbstractEditorHandler {
 
 			this.moveToFullscreen( editorUI.getEditableElement()!, 'editor' );
 			this.moveToFullscreen( editorUIView.toolbar.element!, 'toolbar' );
+			this.generatePresenceListElement();
+			this.generateDocumentOutlineElement();
```
Dumluregn, line 39: Let's try to keep in the specialized classes only the code that differs between them and the rest in the abstract editor, after calling the `defaultEnable()` and before executing the configured callback. It will make it easier to spot the differences (and perhaps unify the editor types one day to get rid of them).

packages/ckeditor5-fullscreen/tests/handlers/abstracteditor.js
```diff
@@ -3,32 +3,62 @@
...
 
+/* globals document */
+
 import global from '@ckeditor/ckeditor5-utils/src/dom/global.js';
 import { ClassicEditor } from '@ckeditor/ckeditor5-editor-classic';
 import { Paragraph } from '@ckeditor/ckeditor5-paragraph';
 import { Essentials } from '@ckeditor/ckeditor5-essentials';
+import { PresenceListUI, PresenceList } from '@ckeditor/ckeditor5-real-time-collaboration';
+import { DocumentOutline, DocumentOutlineUI } from '@ckeditor/ckeditor5-document-outline';
+import mockCloudServices, { CloudServicesMock } from '@ckeditor/ckeditor5-real-time-collaboration/tests/_utils/mockcloudservices.js';
```
Dumluregn, line 14: There's a problem with using the commercial features in the open source repository. We decided to use `/* istanbul ignore next -- @preserve */` to prevent CC to fall below 100%, but provide the coverage on the premium packages side. So e.g. in document outline package there should be a test with fullscreen loaded that checks if everything is ok.

packages/ckeditor5-fullscreen/tests/handlers/abstracteditor.js
```diff
@@ -3,32 +3,62 @@
...
+			},
+			collaboration: {
+				channelId: 'test'
+			},
+			presenceList: {
+				container: presenceList
+			},
+			documentOutline: {
+				container: documentOutline
+			}
```
Dumluregn, line 53: For simplicity, I'd keep the config here minimalistic and create more complex editors on demand in tests that need them. But here as I wrote before we don't want to test the commercial features anyway.

packages/ckeditor5-fullscreen/tests/manual/fullscreen.ts
```diff
@@ -52,6 +52,7 @@ import CloudServices from '@ckeditor/ckeditor5-cloud-services/src/cloudservices.
 import Style from '@ckeditor/ckeditor5-style/src/style.js';
 import GeneralHtmlSupport from '@ckeditor/ckeditor5-html-support/src/generalhtmlsupport.js';
 import Bookmark from '@ckeditor/ckeditor5-bookmark/src/bookmark.js';
+import PresenceList from '@ckeditor/ckeditor5-real-time-collaboration/src/presencelist.js';
```
Dumluregn, line 55: This manual test throws for me right now as there is no `.presence` element in HTML, but let's only use open-source features in the manual test (I'm not sure if it even worked on CI) and premium features - using `all-features-rtc` manual test from commercial repo.

packages/ckeditor5-fullscreen/theme/fullscreen.css
```diff
@@ -80,3 +85,69 @@ Fullscreen layout:
 .ck-fullscreen__main-container .ck-fullscreen__editor .ck-source-editing-area {
 	min-width: calc(210mm + 2px);
 }
+
+.ck-fullscreen__left-sidebar {
+	font-family: Helvetica, Arial, sans-serif;
+	--ck-user-avatar-size: 32px;
```
Dumluregn, line 91: Correct me if I'm wrong, but in presence list outside the fullscreen avatars have 40px and in annotations - 28px. I'm not sure if we really want to introduce a third size? In designs it's also 28px as far as I can see - https://www.figma.com/design/0fCM0UevmXtoL8DbVPt3wT/Full-screen?node-id=147-20312&m=dev.

packages/ckeditor5-fullscreen/theme/fullscreen.css
```diff
@@ -44,25 +44,30 @@ Fullscreen layout:
 }
 
 .ck-fullscreen__main-container .ck-fullscreen__editor-wrapper {
+	--ck-fullscreen-editor-top-margin: 28px;
```
Dumluregn, line 47: Good idea with extracting it to the variable, we should probably do it with some other properties at some point.

packages/ckeditor5-fullscreen/theme/fullscreen.css
```diff
@@ -44,25 +44,30 @@ Fullscreen layout:
...
 }
 
-.ck-fullscreen__main-container .ck-fullscreen__sidebar {
+.ck-fullscreen__sidebar {
 	width: 270px;
-	margin-top: 28px;
+	margin-top: var(--ck-fullscreen-editor-top-margin);
 	margin-left: 10px;
-	margin-right: 10px;
 	background-color: hsl(34.29, 77.78%, 91.18%);
```
Dumluregn, line 58: It's time to remove this color :D

packages/ckeditor5-fullscreen/theme/fullscreen.css
```diff
@@ -44,25 +44,30 @@ Fullscreen layout:
...
+	--ck-fullscreen-editor-top-margin: 28px;
 	display: flex;
 	justify-content: center;
 	overflow: auto;
 	max-height: 100%;
 }
 
-.ck-fullscreen__main-container .ck-fullscreen__sidebar {
+.ck-fullscreen__sidebar {
 	width: 270px;
```
Dumluregn, line 55: According to the designs width should be 300px.

Dumluregn (changes requested):
I think it would make sense that left scroll should only be displayed to the height of document outline, not the whole sidebar.

With default styling, the H1 headings in document outline are larger than the "Document outline" caption, which seems odd - let's maybe discuss it on some meeting with Dagmara, I'll add it to the logbook.

There's an error in the console if you go to fullscreen, go back and click in the editable area:


https://github.com/user-attachments/assets/6fbd0c14-bd5f-41da-916a-0216710b3616

packages/ckeditor5-fullscreen/theme/fullscreen.css
l-lejman: in collapsed view it should be 32px because of white borders which looks like avatars are smaller than counter,
it was mentioned on some call and then i asked Dagmara if my changes looks fine 
but i will remove 32px from not collapsed view

packages/ckeditor5-fullscreen/theme/fullscreen.css
Dumluregn, line 1: In this file we have a snippet with fullscreen container layout, please update it according to the changes from the PR.

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -196,6 +204,74 @@ export default class AbstractEditorHandler {
 		}
 	}
 
+	/* istanbul ignore next -- @preserve */
+	public generatePresenceListElement(): void {
```
Dumluregn, line 208: Missing API docs for some of the new methods. They should also be private.

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -196,6 +204,74 @@ export default class AbstractEditorHandler {
...
+		if ( !this._editor.plugins.has( 'PresenceListUI' ) ) {
+			return;
+		}
+
+		const presneceListElement = createElement( document, 'div', {
```
Dumluregn, line 213: `presneceListElement` => `presenceListElement`

Dumluregn (changes requested):

Dumluregn (approved):

Issue #17978, opened 2025-02-24
Hide document outline and annotations when entering source editing mode


Dumluregn:
## 📝 Provide a description of the new feature

When you enable classic source editing mode, the UI elements related to the editable are not too useful. Let's hide them for cleaner UX.

This will also improve how source editing looks in fullscreen mode.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

PR #17979, opened 2025-02-24
Hide document outline and annotations when entering source editing mode

Dumluregn:

Feature (source-editing): Hide document outline and annotations when entering source editing mode. Closes https://github.com/ckeditor/ckeditor5/issues/17978.

Internal (fullscreen): Display source editing with the same container width as editable.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-source-editing/src/sourceediting.ts
+++ b/packages/ckeditor5-source-editing/src/sourceediting.ts
@@ -292,6 +292,32 @@ export default class SourceEditing extends Plugin {
 	}
 
 	/**
+	 * Restores all hidden editing roots and sets the source data in them.
+	 */
+	private _hideSourceEditing(): void {
+		const editor = this.editor;
+		const editingView = editor.editing.view;
+
+		this.updateEditorData();
+
+		editingView.change( writer => {
+			for ( const [ rootName ] of this._replacedRoots ) {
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-source-editing/src/sourceediting.ts
```diff
@@ -307,9 +311,39 @@ export default class SourceEditing extends Plugin {
...
+			( document.querySelector( '.ck-document-outline' )! as HTMLElement ).style.visibility = '';
+		}
+	}
+
+	/**
+	 * Hides the annotations when entering the source editing mode and shows back them after leaving it.
+	 */
+	private _refreshAnnotationsVisibility() {
+		if ( this.editor.plugins.has( 'Annotations' ) ) {
+			( this.editor.plugins.get( 'Annotations' ) as Annotations ).refreshVisibility();
```
DawidKossowski, line 343: Do we need the casting here?

DawidKossowski (approved):

packages/ckeditor5-source-editing/src/sourceediting.ts
Dumluregn: It's funny, actually. If I remove the casting but leave the unused `Annotations` import, it still works. But without import, TS is no longer able to connect the dots. I think it's clearer to leave casting here than unused import.

PR #17982, opened 2025-02-25
fix: improve line break handling during paste

nks0005:
fix: improve line break handling during paste

- Resolve issue where line breaks were lost when pasting in text/plain environments
- Convert consecutive line breaks to <p></p> and single line breaks to <br> to preserve formatting

Closes #000


Reinmar:
Hi! Thanks for sharing.

We do not plan to change the current behavior, so we cannot accept your PR. Furthermore, for the PR to be merged, it'd need to also adjust the existing tests.

PR #17999, opened 2025-02-26
Fixed emoji panel not being visible

### LINKED ISSUE #17964: Emoji cannot be used in Comments Archive

1. Open a manual test containing the Emoji and Comments plugins (the Emoji plugin needs to be passed to `config.comments.editorConfig.extraPlugins`).
2. Create a comment thread.
3. Archive the thread.
4. Open the Comments Archive dialog.
5. Type `:smile` in the comment input.

### ✔️ Expected result

The Emoji balloon is visible.

### ❌ Actual result

The Emoji balloon is displayed behind the Comments Archive dialog.

Image

### ❓ Possible solution

A global solution would require a lot of work and is probably impossible at this moment. However, we can pass an additional CSS class when opening the Emoji/Mention balloon, setting a higher `z-index` value. This approach makes sense because when the Emoji/Mention balloon appears, it indicates that someone intends to insert it, so it should always have a higher priority than other panels or dialogs.

## 📃 Other details

* Browser: …
* OS: …

---

przemyslaw-zan:

Other (emoji): Fixed emoji panel not being visible while using it in comments archive. Closes #17964.

---

### Additional information


packages/ckeditor5-emoji/theme/emojipicker.css
```diff
@@ -30,3 +30,8 @@
 	margin-right: -1px;
 	outline: solid 1px hsla(0, 0%, 100%, .5);
 }
+
+/* See: https://github.com/ckeditor/ckeditor5/issues/17964 */
+div.ck.ck-balloon-panel:has( .ck.ck-emoji ) {
+	z-index: calc( var( --ck-z-dialog ) + 1 );
```
psmyrek, line 36: Comments archive uses `--ck-z-panel` for the `z-index`. Could you check if after replacing the variable it still works fine?

przemyslaw-zan: It does not work. See:

https://github.com/ckeditor/ckeditor5/blob/26bf48f62284cffdb1079083c509d3a7d091b386/packages/ckeditor5-ui/theme/globals/_zindex.css#L6-L10

psmyrek (commented):
I think we should apply the `z-index` via toggling the class in a template.

Something like:

```
attributes: {
    class: [
        bind.if( 'isEmojiPanelOpen', 'ck-emoji__panel_open' )
    ]
}
```

and then define the `z-index` in `.ck-emoji__panel_open` class in CSS file:

```
.ck-emoji__panel_open {
    z-index: calc( var( --ck-z-dialog ) + 1 );
}
```

packages/ckeditor5-emoji/src/emojipicker.ts
```diff
@@ -138,6 +138,8 @@ export default class EmojiPicker extends Plugin {
 			this.emojiPickerView = this._createEmojiPickerView();
 		}
 
+		this.emojiPickerView.isEmojiPanelOpen = true;
```
pomek, line 141: I am wondering if we could use the binding engine to update the field automatically.

https://ckeditor.com/docs/ckeditor5/latest/framework/deep-dive/observables.html

If it goes well, we could bind the CSS class without defining an additional class property.

pomek:
Answering to question regarding observables. It works as expected.

```diff
diff --git a/packages/ckeditor5-emoji/src/emojipicker.ts b/packages/ckeditor5-emoji/src/emojipicker.ts
index 9b89fe5881..ec99250cf1 100644
--- a/packages/ckeditor5-emoji/src/emojipicker.ts
+++ b/packages/ckeditor5-emoji/src/emojipicker.ts
@@ -138,8 +138,6 @@ export default class EmojiPicker extends Plugin {
                        this.emojiPickerView = this._createEmojiPickerView();
                }

-               this.emojiPickerView.isEmojiPanelOpen = true;
-
                if ( searchValue ) {
                        this.emojiPickerView.searchView.setInputValue( searchValue );
                }
@@ -213,6 +211,10 @@ export default class EmojiPicker extends Plugin {
                        cancel();
                } );

+               emojiPickerView.bind( 'isEmojiPanelOpen' ).to( this.balloonPlugin, 'visibleView', visibleView => {
+                       return visibleView === emojiPickerView;
+               } );
+
                // Close the dialog when clicking outside of it.
                clickOutsideHandler( {
                        emitter: emojiPickerView,
@@ -228,8 +230,6 @@ export default class EmojiPicker extends Plugin {
         * Hides the balloon with the emoji picker.
         */
        private _hideUI(): void {
-               this.emojiPickerView!.isEmojiPanelOpen = false;
-
                this.balloonPlugin.remove( this.emojiPickerView! );
                this.emojiPickerView!.searchView.setInputValue( '' );
                this.editor.editing.view.focus();
diff --git a/packages/ckeditor5-emoji/src/ui/emojipickerview.ts b/packages/ckeditor5-emoji/src/ui/emojipickerview.ts
index 61b8e00c2e..0bfbec19fc 100644
--- a/packages/ckeditor5-emoji/src/ui/emojipickerview.ts
+++ b/packages/ckeditor5-emoji/src/ui/emojipickerview.ts
@@ -34,8 +34,10 @@ import type { EmojiCategory, SkinTone } from '../emojirepository.js';
 export default class EmojiPickerView extends View<HTMLDivElement> {
        /**
         * Tracks whether the panel is currently open.
+        *
+        * @observable
         */
-       public isEmojiPanelOpen: boolean;
+       declare public isEmojiPanelOpen: boolean;

        /**
         * A collection of the focusable children of the view.
@@ -96,7 +98,7 @@ export default class EmojiPickerView extends View<HTMLDivElement> {
        ) {
                super( locale );

-               this.isEmojiPanelOpen = false;
+               this.set( 'isEmojiPanelOpen', false );
                const categoryName = emojiCategories[ 0 ].title;

                this.gridView = new EmojiGridView( locale, {
diff --git a/packages/ckeditor5-mention/src/mentionui.ts b/packages/ckeditor5-mention/src/mentionui.ts
index d57cfc6989..8fdc26ae0b 100644
--- a/packages/ckeditor5-mention/src/mentionui.ts
+++ b/packages/ckeditor5-mention/src/mentionui.ts
@@ -82,7 +82,7 @@ export default class MentionUI extends Plugin {
        /**
         * The contextual balloon plugin instance.
         */
-       private _balloon: ContextualBalloon | undefined;
+       private _balloon: ContextualBalloon;

        private _items = new Collection<{ item: MentionFeedObjectItem; marker: string }>();

@@ -120,6 +120,7 @@ export default class MentionUI extends Plugin {
        constructor( editor: Editor ) {
                super( editor );

+               this._balloon = editor.plugins.get( ContextualBalloon );
                this._mentionsView = this._createMentionView();
                this._mentionsConfigurations = new Map();
                this._requestFeedDebounced = debounce( this._requestFeed, 100 );
@@ -136,8 +137,6 @@ export default class MentionUI extends Plugin {
                const commitKeys = editor.config.get( 'mention.commitKeys' ) || defaultCommitKeyCodes;
                const handledKeyCodes = defaultHandledKeyCodes.concat( commitKeys );

-               this._balloon = editor.plugins.get( ContextualBalloon );
-
                // Key listener that handles navigation in mention view.
                editor.editing.view.document.on<ViewDocumentKeyDownEvent>( 'keydown', ( evt, data ) => {
                        if ( isHandledKey( data.keyCode ) && this._isUIVisible ) {
@@ -270,6 +269,10 @@ export default class MentionUI extends Plugin {
                        return listItemView;
                } );

+               mentionsView.bind( 'isMentionsBalloonOpen' ).to( this._balloon, 'visibleView', visibleView => {
+                       return visibleView === mentionsView;
+               } );
+
                mentionsView.on( 'execute', ( evt, data ) => {
                        const editor = this.editor;
                        const model = editor.model;
@@ -463,8 +466,6 @@ export default class MentionUI extends Plugin {
         * Shows the mentions balloon. If the panel is already visible, it will reposition it.
         */
        private _showOrUpdateUI( markerMarker: Marker ): void {
-               this._mentionsView.isMentionsPanelOpen = true;
-
                if ( this._isUIVisible ) {
                        // @if CK_DEBUG_MENTION // console.log( '%c[UI]%c Updating position.', 'color: green', 'color: black' );

@@ -488,8 +489,6 @@ export default class MentionUI extends Plugin {
         * Hides the mentions balloon and removes the 'mention' marker from the markers collection.
         */
        private _hideUIAndRemoveMarker(): void {
-               this._mentionsView.isMentionsPanelOpen = false;
-
                // Remove the mention view from balloon before removing marker - it is used by balloon position target().
                if ( this._balloon!.hasView( this._mentionsView ) ) {
                        // @if CK_DEBUG_MENTION // console.log( '%c[UI]%c Hiding the UI.', 'color: green', 'color: black' );
diff --git a/packages/ckeditor5-mention/src/ui/mentionsview.ts b/packages/ckeditor5-mention/src/ui/mentionsview.ts
index 94a23966dd..2ec8d01a1e 100644
--- a/packages/ckeditor5-mention/src/ui/mentionsview.ts
+++ b/packages/ckeditor5-mention/src/ui/mentionsview.ts
@@ -18,7 +18,12 @@ import '../../theme/mentionui.css';
  * The mention ui view.
  */
 export default class MentionsView extends ListView {
-       public isMentionsPanelOpen: boolean;
+       /**
+        * Tracks whether the balloon is currently open.
+        *
+        * @observable
+        */
+       declare public isMentionsBalloonOpen: boolean;

        public selected: MentionListItemView | undefined;

@@ -30,13 +35,13 @@ export default class MentionsView extends ListView {
        constructor( locale: Locale ) {
                super( locale );

-               this.isMentionsPanelOpen = false;
+               this.set( 'isMentionsBalloonOpen', false );

                this.extendTemplate( {
                        attributes: {
                                class: [
                                        'ck-mentions',
-                                       this.bindTemplate.if( 'isMentionsPanelOpen', 'ck-mentions__panel_open' )
+                                       this.bindTemplate.if( 'isMentionsBalloonOpen', 'ck-mentions__panel_open' )
                                ],

                                tabindex: '-1'
```

packages/ckeditor5-mention/src/ui/mentionsview.ts
```diff
@@ -28,10 +30,13 @@ export default class MentionsView extends ListView {
...
 		super( locale );
 
+		this.isMentionsPanelOpen = false;
+
 		this.extendTemplate( {
 			attributes: {
 				class: [
-					'ck-mentions'
+					'ck-mentions',
+					this.bindTemplate.if( 'isMentionsPanelOpen', 'ck-mentions__panel_open' )
```
pomek, line 39: This line isn't covered via tests.

packages/ckeditor5-emoji/tests/ui/emojipickerview.js
```diff
@@ -69,6 +69,10 @@ describe( 'EmojiPickerView', () => {
 	describe( 'constructor()', () => {
 		const some = ( arg, callback ) => [ ...arg ].some( callback );
 
+		it( 'should create class with isEmojiPanelOpen set to false', () => {
+			expect( emojiPickerView.isEmojiPanelOpen ).to.equal( false );
+		} );
```
pomek, line 74: This test does not check anything.

packages/ckeditor5-emoji/src/ui/emojipickerview.ts
```diff
@@ -160,7 +166,12 @@ export default class EmojiPickerView extends View<HTMLDivElement> {
 			],
 			attributes: {
 				tabindex: '-1',
-				class: [ 'ck', 'ck-emoji', 'ck-search' ]
+				class: [
+					'ck',
+					'ck-emoji',
+					'ck-search',
+					this.bindTemplate.if( 'isEmojiPanelOpen', 'ck-emoji__panel_open' )
```
pomek, line 173: This line isn't covered via tests.

packages/ckeditor5-emoji/src/emojipicker.ts
pomek: > If it goes well, we could bind the CSS class without defining an additional class property.

Binds work as expected, but a view instance cannot access the `ContextualBalloon` plugin. Hence, we need two changes:

1. The view registers an onservable and updates its structure based on its value.
2. Bind the observable with `ContextualBalloon` in the controller (a UI plugin).

See: https://github.com/ckeditor/ckeditor5/pull/17999#issuecomment-2696350260 - I made PoC.

packages/ckeditor5-emoji/theme/emojipicker.css
```diff
@@ -30,3 +30,8 @@
 	margin-right: -1px;
 	outline: solid 1px hsla(0, 0%, 100%, .5);
 }
+
+/* See: https://github.com/ckeditor/ckeditor5/issues/17964 */
+div.ck.ck-balloon-panel:has( .ck.ck-emoji ) {
```
oleq: Instead of targeting emoji feature in CSS, I'd add a CSS class to the balloon, for instance:

```diff
diff --git a/packages/ckeditor5-emoji/src/emojipicker.ts b/packages/ckeditor5-emoji/src/emojipicker.ts
index 68fbdb54be..cb7652c1ca 100644
--- a/packages/ckeditor5-emoji/src/emojipicker.ts
+++ b/packages/ckeditor5-emoji/src/emojipicker.ts
@@ -147,7 +147,8 @@ export default class EmojiPicker extends Plugin {
                   if ( !this.balloonPlugin.hasView( this.emojiPickerView ) ) {
                            this.balloonPlugin.add( {
                                     view: this.emojiPickerView,
-                                    position: this._getBalloonPositionData()
+                                    position: this._getBalloonPositionData(),
+                                    balloonClassName: 'ck-emoji-picker-balloon'
                            } );
                   }
```

and then use `div.ck.ck-balloon-panel.ck-emoji-picker-balloon`. This way, you can add a unit test that says "it should have ... class to make sure z-index does not conflict with the dialog system" that only checks the presence of the CSS class. This will create a trace of the business logic in tests and reduce the chances this gets forgotten or lost (not by much, but a little bit).

oleq (changes requested):
Oh god, I hate these issues 😀 I suggested a minor tweak to the fix since I don't see anything better we could do without a full-scale z-index management system which we don't have time and resources to develop right now (there are technical obstacles too, long story...).

Disclaimer: I haven't tested this fix live so make sure this goes through QA.

packages/ckeditor5-emoji/theme/emojipicker.css
oleq: We could do the same to the mention plugin: name the balloon that hosts the mentions view, and leave a trace in tests saying why the name (CSS class) is important.

psmyrek (approved):
LGTM 🎉

pomek (approved):

oleq (approved):

PR #18028, opened 2025-02-28
Dialog plugin position adjustments in fullscreen mode

l-lejman:

Internal: Dialog plugin position adjustments in fullscreen mode.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
+++ b/packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
@@ -11,6 +11,7 @@ import { PresenceListUI } from '@ckeditor/ckeditor5-real-time-collaboration';
 import { DocumentOutlineUI } from '@ckeditor/ckeditor5-document-outline';
 import { createElement } from 'ckeditor5/src/utils.js';
 import { type Editor } from 'ckeditor5/src/core.js';
+import { type Dialog } from 'ckeditor5/src/ui.js';
 
 /**
  * The abstract editor type handler. It should be extended by the particular editor type handler.
@@ -176,5 +177,15 @@ export default class AbstractEditorHandler {
 		documentOutlineUI.view._documentOutlineContainer = document.querySelector( '[data-ck-fullscreen="left-sidebar"]' ) as HTMLElement;
 
 		this.moveToFullscreen( documentOutlineUI.view.element!, 'document-outline' );
+	}
+
+	public updateDialogPosition(): void {
+		const dialog = this._editor.plugins.get( 'Dialog' ) as Dialog;
+		dialog.on( 'change:isOpen', ( ) => {
+			const dialogView = dialog.view!;
+			dialogView.position = null;
+
+			dialogView.moveTo( 0, 0 );
+		}, { priority: 'highest' } );
 	}
 }
--- a/packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
+++ b/packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
@@ -9,9 +9,9 @@
 
 import { PresenceListUI } from '@ckeditor/ckeditor5-real-time-collaboration';
 import { DocumentOutlineUI } from '@ckeditor/ckeditor5-document-outline';
-import { createElement } from 'ckeditor5/src/utils.js';
+import { type EventInfo, createElement, Rect } from 'ckeditor5/src/utils.js';
 import { type Editor } from 'ckeditor5/src/core.js';
-import { type Dialog } from 'ckeditor5/src/ui.js';
+import { DialogViewPosition, type Dialog } from 'ckeditor5/src/ui.js';
 
 /**
  * The abstract editor type handler. It should be extended by the particular editor type handler.
@@ -126,6 +126,8 @@ export default class AbstractEditorHandler {
 			this._container.remove();
 			this._container = null;
 		}
+
+		this.unregisterFullscreenDialogPositionAdjustements();
 	}
 
 	public generatePresenceListElement(): void {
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -177,4 +180,94 @@ export default class AbstractEditorHandler {
...
+		}
+
+		const dialog = this._editor.plugins.get( 'Dialog' ) as Dialog;
+		const dialogView = dialog.view;
+
+		if ( dialogView && dialogView.position === null ) {
+			dialogView.position = DialogViewPosition.EDITOR_TOP_SIDE;
+		}
+
+		dialogView?.updatePosition();
```
Dumluregn, line 215: Unfortunately we're not supposed to use `?.` operator. I can't find the rationale right now so you have to trust me :D but I'll share it once I get it.

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -177,4 +180,94 @@ export default class AbstractEditorHandler {
...
+			}
+		}
+	}
+
+	/**
+	 * Adjusts the dialog position to utilize the empty space on the right side of the editable.
+	 * The new dialog position should be on the right side of the fullscreen view with a 30px margin.
+	 * Only dialogs with the position set to "editor-top-side" should have their position changed.
+	 */
+	public setNewDialogPosition(): void {
```
Dumluregn, line 243: I guess it can be private?

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -177,4 +180,94 @@ export default class AbstractEditorHandler {
...
+		if ( dialogView && dialogView.position === null ) {
+			dialogView.position = DialogViewPosition.EDITOR_TOP_SIDE;
+		}
+
+		dialogView?.updatePosition();
+
+		dialog.off( 'change:isOpen', this.updateDialogPositionCallback );
+	}
+
+	public updateDialogPositionCallback = this.updateDialogPosition.bind( this );
```
Dumluregn, line 220: We shouldn't leave undocumented methods (especially public) and it's really hard to write anything smart here, so I'd probably use `this.updateDialogPosition.bind( this )` directly even though it will be duplication. If you prefer to keep it please add one sentence of docs.

packages/ckeditor5-fullscreen/src/handlers/classiceditor.ts
```diff
@@ -37,6 +37,7 @@ export default class ClassicEditorHandler extends AbstractEditorHandler {
 			this.moveToFullscreen( editorUIView.toolbar.element!, 'toolbar' );
 			this.generatePresenceListElement();
 			this.generateDocumentOutlineElement();
+			this.registerFullscreenDialogPositionAdjustements();
```
Dumluregn, line 40: Let's move it to the abstract handler (IIRC the two above calls should also be there).

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -177,4 +180,94 @@ export default class AbstractEditorHandler {
...
+			dialogView.moveTo(
+				fullscreenViewContainerRect.left + fullscreenViewContainerRect.width - dialogRect.width - DIALOG_OFFSET,
+				editorContainerRect.top + DIALOG_OFFSET
+			);
+		}
+	}
+
+	private _getVisibleContainerRect( container: HTMLElement ): Rect | null {
+		return new Rect( container )?.getVisible();
+	}
```
Dumluregn, line 272: This helper is only used in one method so maybe we'd move it to that functions scope to keep the interface cleaner. Or leave it here but add docs.

Dumluregn (changes requested):
Mostly some small stuff. Please merge the latest version of [epic/1235-fullscreen-mode-presence-list](https://github.com/ckeditor/ckeditor5/tree/epic/1235-fullscreen-mode-presence-list) as well 🙏🏻

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
l-lejman: I need the same reference to the callback function for .on() and .off() to work correctly.
Since .bind(this) returns a new reference each time, the function must be stored in a variable to allow proper event detachment. I will add comment.

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -393,4 +397,95 @@ export default class AbstractEditorHandler {
...
+		if ( !dialogView || dialogView.position !== DialogViewPosition.EDITOR_TOP_SIDE ) {
+			return;
+		}
+
+		const fullscreenViewContainerRect = new Rect( this._container! ).getVisible();
+		const editorContainerRect = new Rect( document.querySelector( '.ck-fullscreen__editable' ) as HTMLElement ).getVisible();
+		const dialogRect = new Rect( dialogView.element!.querySelector( '.ck-dialog' ) as HTMLElement ).getVisible();
+
+		if ( fullscreenViewContainerRect && editorContainerRect && dialogRect ) {
+			const DIALOG_OFFSET = 28;
```
Dumluregn, line 482: We used to define constants at the top level before the plugin class declaration, it's easier to spot them this way.

packages/ckeditor5-fullscreen/src/handlers/abstracteditor.ts
```diff
@@ -393,4 +397,95 @@ export default class AbstractEditorHandler {
...
+			dialogView.position = null;
+
+			dialogView.moveTo(
+				fullscreenViewContainerRect.left + fullscreenViewContainerRect.width - dialogRect.width - DIALOG_OFFSET,
```
Dumluregn, line 486: We need to take the scrollbar width into account. 
image

Dumluregn (changes requested):

Dumluregn (approved):

PR #18030, opened 2025-03-03
Introduced the `emoji.useCustomFont` option

### LINKED ISSUE #17988: Add additional logging during the `EmojiRepository` plugin initialization

We should add additional logging during the `EmojiRepository` plugin initialization since something might go wrong during the process related to emojis/fonts etc. Without this plugin being ready, the toolbar item and command are not registered and integrators will see the message that the toolbar item is unavailable without any further clue. 

While integrating samples on CI that used Ubuntu images we encountered an issue with (as it occurred later) missing font and there were no logs that would help us to determine that was the case.

As we discussed with @pomek, it would be beneficial to add warnings at least if:
- there are no items
- there are no fonts
- the resource was not loaded

This will streamline debugging such kind of issues for integrators.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

---

pomek:

Other (emoji): Introduced the `emoji.useCustomFont` option to disable the filtering mechanism. Closes #17988.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-emoji/docs/features/emoji.md
+++ b/packages/ckeditor5-emoji/docs/features/emoji.md
@@ -130,6 +130,20 @@ The feature can be configured via the {@link module:emoji/emojiconfig~EmojiConfi
 		.catch( /* ... */ );
 	```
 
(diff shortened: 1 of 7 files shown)


Witoso (approved):

psmyrek (approved):
LGTM 🎉 

Minor issue: loading the Noto font in manual test changes also the look of the whole editor from serif font to Helvetica font:

obraz

obraz

PR #18035, opened 2025-03-03
Add fullscreen integration with editor created in the iframe

Dumluregn:

Internal (fullscreen): Add fullscreen integration with editor created in the iframe.

Internal (fullscreen): Allow to configure a custom container where fullscreen should be initialized.

---

### Additional information


f1ames (changes requested):
The code itself looks good 👍 

---

However, I have some doubts about the manual test. The editor styling is missing and the iframe is quite small so it's difficult to test it or figure out what's going on there):

https://github.com/user-attachments/assets/ed04d471-1603-4293-8c81-45d136e9ec20

Two suggestions:
1. Make the iframe bigger. And maybe add some margins to its body so one can see how fs editor expands?
2. Try to add styles to the editor in the iframe.

**Ad.2**
As we talked F2F it's a bit tricky (I didn't investigate too much) to add styles. The quick workaround could be using styles loaded from CDN:

```
<head>
	<meta http-equiv="Content-Security-Policy"
		content="default-src 'none'; connect-src 'self' https://cksource.com http://*.cke-cs.com; script-src 'self' https://cksource.com; img-src * data:; style-src * 'self' 'unsafe-inline'; frame-src * data:">
</head>

// ...

<iframe id="iframe" style="display:none;" srcdoc='
	<link rel="stylesheet" type="text/css" href="https://cdn.ckeditor.com/ckeditor5/44.2.1/ckeditor5.css">
	<div class="editor-container">
	<div id="editor">

// ...
```

And this works fine:
image

However, it's a bit cumbersome to maintain and update so I'm not 100% sure. But still better than no styles I guess.

f1ames:
> As we talked F2F it's a bit tricky (I didn't investigate too much) to add styles. The quick workaround could be using styles loaded from CDN...

This turned out to be not a very good idea (sorry 😓) since styles from CDN are missing fullscreen styling. So the fullscreen mode in iframe was looking odd.

https://github.com/user-attachments/assets/ec1e2920-c92c-4834-8814-c19de8f54c76

However, I found another solution, just to copy CKE5 styles from the main document to iframe document - https://github.com/ckeditor/ckeditor5/pull/18035/commits/99ab389a0afa52244f0ca8e703bbe33eb3073ad2. And now it works as expected, all new CSS rules will be used in the iframe and so it can be tested properly:

https://github.com/user-attachments/assets/b447d0c6-cdc9-4d29-a658-62145fde4895

f1ames (approved):
LGTM 👍

PR #18048, opened 2025-03-04
Add keystroke for fullscreen mode

l-lejman:

Internal: Add keystroke for fullscreen mode.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-fullscreen/src/fullscreenediting.ts
+++ b/packages/ckeditor5-fullscreen/src/fullscreenediting.ts
@@ -11,6 +11,8 @@ import { Plugin, type Editor } from 'ckeditor5/src/core.js';
 
 import FullscreenCommand from './fullscreencommand.js';
 
+const FULLSCREEN = 'fullscreen';
+
 /**
  * A plugin that registers the fullscreen mode command.
  */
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-fullscreen/src/fullscreenediting.ts
```diff
@@ -43,5 +45,21 @@ export default class FullscreenEditing extends Plugin {
...
+		const t = this.editor.locale.t;
+
+		// Set the Ctrl+Shift+F keystroke.
+		this.editor.keystrokes.set( 'Ctrl+Shift+F', FULLSCREEN );
+
+		// Add the information about the keystroke to the accessibility database.
+		this.editor.accessibility.addKeystrokeInfos( {
+			keystrokes: [
+				{
+					label: t( 'Fullscreen mode' ),
```
Dumluregn, line 58: There's a convention to name these entries to sound like human-friendly commands, i.e. start with a verb:

image

So this should be more like "Toggle fullscreen mode".

packages/ckeditor5-fullscreen/src/fullscreenediting.ts
```diff
@@ -43,5 +45,21 @@ export default class FullscreenEditing extends Plugin {
 	 */
 	public init(): void {
 		this.editor.commands.add( 'fullscreen', new FullscreenCommand( this.editor ) );
+
+		const t = this.editor.locale.t;
+
+		// Set the Ctrl+Shift+F keystroke.
+		this.editor.keystrokes.set( 'Ctrl+Shift+F', FULLSCREEN );
```
Dumluregn, line 52: I see we use such convention in some places (e.g. for `bold` plugin) but it's only useful if such const can be used in multiple places. Here we can use `fullscreen` as a command name directly. Or if we want to have it as a const, then also use it in line `47` where we add the command to the editor.

Dumluregn (changes requested):
After using the keystroke, it seems like browser steal the focus and the second stroke doesn't toggle editor fullscreen but is handled by the browser. We should add an `editor.editing.view.focus();` call either here to the keystroke handler, or directly in the command (I think it makes sense - but then remember to remove this call from button press in UI plugin).

Dumluregn (approved):

PR #18076, opened 2025-03-06
Fixed `template-generic.svg` icon to work correctly as a template dropdown default icon.

martnpaneq:

Fix (icons): Fixed `template-generic.svg` icon to work correctly as a template dropdown default icon.
---

### Additional information


pomek:
I updated the proposed commit message. It's an internal change. A bug was introduced in the current iteration, so from the users' perspective, we do not fix anything.

pomek (changes requested):
PR is empty. It wasn't expected, I think.

martnpaneq:
@pomek Your right. Looks like our script for cleaning up icons `yarn run clean-up-svg-icons` meses up this particular icon. you can see that I used a correct icon in the first commit: [6b23cdc](https://github.com/ckeditor/ckeditor5/pull/18076/commits/6b23cdc274b3b6d713ed4a916765a8a71c6f8e76), but after running icons cleanup, it got replaced with a broken icon. 

I'm going to investigate further.

PR #18082, opened 2025-03-07
Docs: Fix snippets

filipsobol:

Docs: Fix snippets.

---

### Additional information

Changes introduced in this PR:

* Always load CKBox from CDN, as it's used in most snippets.
* Disable CKBox plugins in few snippets that are incompatible with it.
* Remove snippets that only included CKBox script tag.
* Fix paths to editor styles in snippets using `Export to PDF` and `Export to Word` features.
* Restore a few “snippet builds” that included some content or styles.
* Fix Bootstrap example.

### DIFF
--- a/docs/_snippets/features/keyboard-support.js
+++ b/docs/_snippets/features/keyboard-support.js
@@ -215,8 +215,8 @@ ClassicEditor
 		exportPdf: {
 			stylesheets: [
 				'../../assets/pagination-fonts.css',
-				'EDITOR_STYLES',
-				'../../snippets/features/pagination/snippet.css',
+				'../../assets/ckeditor5/ckeditor5.css',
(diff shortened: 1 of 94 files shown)


docs/_snippets/features/keyboard-support.js
```diff
@@ -215,8 +215,8 @@ ClassicEditor
 		exportPdf: {
 			stylesheets: [
 				'../../assets/pagination-fonts.css',
-				'EDITOR_STYLES',
-				'../../snippets/features/pagination/snippet.css',
+				'../../assets/ckeditor5/ckeditor5.css',
```
pomek, line 218: I wouldn't say I like the relative paths, but I don't want to spend more on it than needed, so let's leave it and improve in the future (if it becomes a problematic case).

pomek (approved):
I've verified all reported cases, which behave like the production docs. So, I would say it's fixed.

PR #18109, opened 2025-03-12
Introduce `TableTypeCommand` command.

pszczesniak:

Internal: Introduce `TableTypeCommand` command. See https://github.com/ckeditor/ckeditor5/issues/18131.

---

### Additional information


packages/ckeditor5-table/src/tablelayout/commands/settabletype.ts
```diff
@@ -0,0 +1,76 @@
...
+ * The command is registered by {@link module:table/tablelayout/tablelayoutediting~TableLayoutEditing}
+ * as the `'setTableType'` editor command.
+ *
+ * To set the table type at the current selection, execute the command and specify the table type:
+ *
+ * ```ts
+ * editor.execute( 'setTableType', 'layout' );
+ * ```
+ */
+export default class SetTableType extends Command {
```
niegowski, line 29:
```suggestion
export default class TableTypeCommand extends Command {
```

packages/ckeditor5-table/src/tablelayout/commands/settabletype.ts
```diff
@@ -0,0 +1,76 @@
...
+			const tableChildren = table.getChildren();
+
+			for ( const child of tableChildren ) {
+				if ( model.schema.checkChild( newTable, child ) ) {
+					writer.append( writer.cloneElement( child as Element ), newTable );
+				}
+			}
+
+			writer.insert( newTable, table, 'after' );
+			writer.remove( table );
```
niegowski, line 73: Why this way? It would be simpler to change the attribute and remove not allowed children according to the schema. This way it could be more complicated for OT or RH. Also TC could misinterpret this.

pszczesniak: Yes, that makes sense.

Mati365 (approved):

packages/ckeditor5-table/src/tablelayout/commands/tabletypecommand.ts
```diff
@@ -0,0 +1,80 @@
...
+			writer.setAttribute( 'tableType', tableType, table );
+
+			const allTableAttributes = table.getAttributes();
+
+			// Check if all attributes are allowed for the new table type.
+			for ( const [ attributeName ] of allTableAttributes ) {
+				if ( attributeName !== 'tableType' && !model.schema.checkAttribute( table, attributeName ) ) {
+					writer.removeAttribute( attributeName, table );
+				}
+			}
```
niegowski, line 68: Maybe we could use the Schema dedicated method for this? https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-Schema.html#function-removeDisallowedAttributes

packages/ckeditor5-table/src/tablelayout/commands/tabletypecommand.ts
```diff
@@ -0,0 +1,80 @@
...
+
+			const tableChildren = table.getChildren();
+
+			// Check if all children are allowed for the new table type.
+			for ( const child of tableChildren ) {
+				if ( !model.schema.checkChild( table, child ) ) {
+					writer.remove( child );
+				}
+			}
```
niegowski, line 77: Should we add a post-fixer to ensure that some other feature won't break the model structure? I wonder if TC suggestion could alter the structure in an unexpected way.

packages/ckeditor5-table/src/tablelayout/commands/tabletypecommand.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @module table/tablelayout/commands/tabletypecommand
+ */
+
+import { Command } from 'ckeditor5/src/core.js';
+
+import { getSelectionAffectedTable } from '../../utils/common.js';
+
+export type TableType = 'layout' | 'content';
```
pszczesniak, line 14: Let's import this type from :point\_right: [https://github.com/ckeditor/ckeditor5/pull/18121#discussion_r1998622816](https://github.com/ckeditor/ckeditor5/pull/18121#discussion_r1998622816) (when [this PR](https://github.com/ckeditor/ckeditor5/pull/18121) will merged).

pszczesniak: Done in 130e16a.

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -161,6 +164,26 @@ export default class TableLayoutEditing extends Plugin {
...
+							// Check if all children are allowed for the new table type.
+							for ( const child of tableChildren ) {
+								if ( !editor.model.schema.checkChild( item, child ) ) {
+									writer.remove( child );
+									hasChanged = true;
+								}
+							}
+						}
+					}
+				}
```
niegowski, line 186: What about some captions added to a layout table? Or a heading cols/rows set on it?

niegowski: Ok, let's leave it for a potential follow-up if such a case appears.

niegowski (approved):

PR #18121, opened 2025-03-13
Layout table heuristics configuration.

pszczesniak:

Internal: Introducing the layout table heuristics configuration. See https://github.com/ckeditor/ckeditor5/issues/18131.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-table/src/tableconfig.ts
+++ b/packages/ckeditor5-table/src/tableconfig.ts
@@ -208,6 +208,17 @@ export interface TableConfig {
 	 * Read more about configuring the table feature in {@link module:table/tableconfig~TableConfig}.
 	 */
 	tableCellProperties?: TableCellPropertiesConfig;
+
+	/**
+	 * Configuration of the table layout feature.
+	 *
+	 * ```ts
+	 * const tableConfig = {
+	 * 	tableLayout: ... // Table layout feature config.
+	 * };
+	 * ```
+	 */
+	tableLayout?: TableLayoutConfig;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-table/src/tableconfig.ts
```diff
@@ -372,3 +383,35 @@ export interface TableCellPropertiesOptions {
...
+/**
+ * The configuration of the table layout feature.
+ */
+export interface TableLayoutConfig {
+
+	/**
+	 * The preferred type of the external table to be inserted.
+	 *
+	 * By using this setting you can change the table type detection heuristics.
+	 * The highest priority has recognition by CSS classnames: `content-table` and `layout-table`.
```
niegowski, line 396: I'd change it to less technical (no priorities).

packages/ckeditor5-table/src/tableconfig.ts
```diff
@@ -372,3 +383,35 @@ export interface TableCellPropertiesOptions {
...
+	 * If the classnames are not present, the default heuristic is used.
```
niegowski, line 397:
```suggestion
	 * If the class names are not present, the default heuristic is used.
```

packages/ckeditor5-table/src/tableconfig.ts
```diff
@@ -372,3 +383,35 @@ export interface TableCellPropertiesOptions {
...
+	 * 		}
+	 * 	} )
+	 * 	.then( ... )
+	 * 	.catch( ... );
+	 * ```
+	 */
+	preferredExternalTableType: TableType;
+}
+
+export type TableType = 'content' | 'layout';
```
niegowski, line 417: The same is declared in the command, maybe we should have a single type for it exported from for example editing?

niegowski (changes requested):

packages/ckeditor5-table/src/tableconfig.ts
pszczesniak: I would keep it in the configuration file, as I have found that these types of files are most commonly used in such cases.

pszczesniak: Added a [comment](https://github.com/ckeditor/ckeditor5/pull/18109/files#r1998882906) to fix it in the TableTypeCommand PR.

niegowski (approved):

PR #18133, opened 2025-03-14
Pass path to a dynamically imported file as `file://` protocol URL to make it work in Windows

psmyrek:

Internal: Fixed path calculation to a dynamically imported file in snippets adapter to work independently of the OS. Previously, it did not work in Windows, because there is a specific requirement that absolute path loaded in `import()` must be valid `file://` protocol URL.

---

### Additional information


przemyslaw-zan (approved):
The fix works.

pomek (approved):
LGTM. The proposed solution works fine on Unix as well.

PR #18155, opened 2025-03-18
Docs: Prevent source leaking.

filipsobol:

Docs: Prevent source leaking.

### DIFF
--- a/scripts/docs/snippetadapter.mjs
+++ b/scripts/docs/snippetadapter.mjs
@@ -136,11 +136,28 @@ async function buildSnippets( snippets, paths, constants, imports ) {
 			 * bundled. This will cause issues, because the `tests` directory is not available in the CDN build.
 			 */
 			{
-				name: 'external',
+				name: 'externalize-ckeditor5',
 				setup( build ) {
 					build.onResolve( { filter: /.*/ }, args => ( {
 						external: externals.some( name => name.endsWith( '/' ) ? args.path.startsWith( name ) : args.path === name )
 					} ) );
+				}
+			},
+
+			/**
+			 * Code from "ckeditor5" and "ckeditor5-premium-features" is externalized to avoid bundling it with the snippet
+			 * code. However, if the source file is imported directly, it can be bundled and (in case of commercial features)
+			 * not obfuscated. To prevent this, we throw an error if such an import is detected.
+			 */
+			{
+				name: 'throw-on-source-import',
+				setup( build ) {
+					build.onLoad( { filter: /\.ts$/ }, args => {
+						throw new Error(
+							`The source file "${ args.path }" was likely imported by mistake. ` +
+							'Update the import to use "ckeditor5" or "ckeditor5-premium-features" instead.'
+						);
+					} );
 				}
 			}
 		]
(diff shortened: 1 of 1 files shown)


pomek (commented):
While importing a TS file is not allowed, I figured out that importing a built file (JS) still works.

* Execute `yarn build` in `ckeditor5-utils`.
* Add the following import to any snippet: `import version from '@ckeditor/ckeditor5-utils/src/version.js';`.
* Execute: `yarn run docs --skip-api --skip-validation --dev --skip-ckeditor5`

We expect to crash. 

```
Error: Build failed with 1 error:
node_modules/esbuild/lib/main.js:1277:27: ERROR: [plugin: throw-on-source-import] The source file "/home/pomek/Projects/ckeditor/ckeditor5/packages/ckeditor5-utils/src/version.ts" was likely imported by mistake. Update the import to use "ckeditor5" or "ckeditor5-premium-features" instead.
```

Instead, it works.

```
$ node --max-old-space-size=8192 scripts/docs/build-docs.mjs --skip-api --skip-validation --dev --skip-ckeditor5 --guides features/emoji
Clean build directory.
Emptying Hexo source directory.
Executing the snippet adapter for "ckeditor5".
Started building snippets.
Finished building snippets.
Running webpack for "app.js"...
[Webpack] 1 asset
412 modules
webpack 5.98.0 compiled successfully in 2231 ms
Deprecation Warning [legacy-js-api]: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.

More info: https://sass-lang.com/d/legacy-js-api

Symlink for CKEditor 5 ./44.3.0 - created.
Building documentation complete in 0m 8.150167698s.
Done in 8.65s.
```

filipsobol:
I updated the RegExp and how the error is returned from the plugin.

scripts/docs/snippetadapter.mjs
```diff
@@ -136,12 +136,31 @@ async function buildSnippets( snippets, paths, constants, imports ) {
...
+
+			/**
+			 * Code from "ckeditor5" and "ckeditor5-premium-features" is externalized to avoid bundling it with the snippet
+			 * code. However, if the source file is imported directly, it can be bundled and (in case of commercial features)
+			 * not obfuscated. To prevent this, we throw an error if such an import is detected.
+			 */
+			{
+				name: 'throw-on-source-import',
+				setup( build ) {
+					build.onLoad( { filter: /packages\/(.*)\/src/ }, args => ( {
```
pomek, line 155: @psmyrek, could you check if adding the following snippet

```js
import version from '@ckeditor/ckeditor5-utils/src/version.js';
```

to any snippet, then building them throws an error on Windows?

While I believe all paths are absolute at this stage, I am unsure if they use an OS separator.

pomek (dismissed):
LGTM.

Importing JS and TS files ends with the error.

Before merging, let's wait for the final confirmation on a Windows environment.

scripts/docs/snippetadapter.mjs
psmyrek, line 155: It seems that paths have OS-specific separators, because currently this filter does not match anything on Windows.

```suggestion
					build.onLoad( { filter: /packages[\\/](.*)[\\/]src/ }, args => ( {
```

pomek:
```suggestion
					build.onLoad( { filter: /packages[\\/].*[\\/]src/ }, args => ( {
```

We do not have to capture anything. I am unsure if it speeds up (:D).

filipsobol: Fixed

psmyrek (approved):
LGTM 🎉

PR #18165, opened 2025-03-19
Improve source editing integration with document outline in fullscreen mode

Dumluregn:

Internal (fullscreen): Improve source editing integration with document outline in fullscreen mode.

---

### Additional information

1. Source editing integration was not perfect, it didn't take into account a situation when there are multiple document outline containers on the page.
2. Fullscreen has to detect when source mode is open and toggle "Document outline" header visibility.

### DIFF
--- a/packages/ckeditor5-source-editing/src/sourceediting.ts
+++ b/packages/ckeditor5-source-editing/src/sourceediting.ts
@@ -15,6 +15,7 @@ import { ButtonView, MenuBarMenuListItemButtonView, type Dialog } from 'ckeditor
 import { CKEditorError, createElement, ElementReplacer } from 'ckeditor5/src/utils.js';
 import { formatHtml } from './utils/formathtml.js';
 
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-source-editing/src/sourceediting.ts
```diff
@@ -322,17 +322,17 @@ export default class SourceEditing extends Plugin {
 	 * Hides the document outline if it is configured.
 	 */
 	private _hideDocumentOutline() {
-		if ( document.querySelector( '.ck-document-outline' ) ) {
-			( document.querySelector( '.ck-document-outline' )! as HTMLElement ).style.visibility = 'hidden';
+		if ( this.editor.plugins.has( 'DocumentOutline' ) ) {
+			( this.editor.plugins.get( 'DocumentOutlineUI' ) as any ).view.element.style.display = 'none';
```
DawidKossowski, line 326: Could we cast it to `DocumentOutlineUI` instead of `any`?

DawidKossowski (approved):

PR #18173, opened 2025-03-19
Fixed broken svg icons.

przemyslaw-zan:

Internal: Fixed broken SVG icons.

---

### Additional information


pomek:
* Please. improve the PR title. We know which icon is broken (look at the original issue).
* The issue mentions a single icon only. This PR fixes three. Is there any reason for that?
* It's an internal change. Icons grouped into a single package are part of the iteration 84 release.

przemyslaw-zan:
The PR fixes three icons instead of one, because the issue says:

> Additional info
>
> We should probably check all the icons in the project.

pomek (approved):
LGTM.

Issue #16923, opened 2024-08-14
Table dimensions are read as "0 by 0" by Windows Narrator / `role="textbox"` considered invalid in `td` and `th`


lkszzajac:
## 📝 Provide detailed reproduction steps (if any)

1. Open any editor and remove its initial content (e.g. https://ckeditor.com/ckeditor-5/demo/feature-rich/)
2. Create a table (e.g. 3x3, can be empty)
3. Enable Windows Narrator
4. Focus the table widget

### ✔️ Expected result

Narrator reads the widget as a table and announces its dimensions

### ❌ Actual result

Narrator reads the widget as a table and reports its dimensions as 0x0

https://github.com/user-attachments/assets/fb1ce81a-e0de-45df-9f26-0a3d6630dc1c

## 📃 Other details

* Browser: Edge
* OS: Windows 11
* First affected CKEditor version: …
* Installed CKEditor plugins: …
* Related issues in other screen readers:
    * https://github.com/ckeditor/ckeditor5/issues/11867
    * https://github.com/ckeditor/ckeditor5/issues/12224
    * https://github.com/ckeditor/ckeditor5/issues/13479
* Issue does not reproduce in other screen readers because no other reader reports on table dimension by default on focus. VoiceOver can read the table dimensions only after focusing a cell and going "up" in focus by one level

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

lkszzajac:
We've checked the behavior in these setups:

Windows/Edge:
- Narrator - announces the table with "Enter table, 0x0"
- JAWS - announces the table with "Edit, land, type of text"
- NVDA - announces the table with "Figure, table, mutliline, blank"

macOS/Chrome:
- VoiceOver - announces the table with "table" without any dimensions.

macOS/Firefox/Safari:
- VoiceOver - announces the table with dimensions correctly, does not announce the cells as editable

Comandeer:
[...]

PR #18201, opened 2025-03-24
Remove `role=textbox` from `td` / `th` cell editables.

Mati365:

Fix (table): Removed `role: textbox` from `td` / `th` in editables. Windows Narrator no longer reads table dimensions as 0 by 0.  Closes https://github.com/ckeditor/ckeditor5/issues/16923

---

### Additional information

Testing summary: https://github.com/ckeditor/ckeditor5/issues/16923#issuecomment-2733427130

> As removing role=textbox doesn't degrade the experience in most screen readers, but rather improves it (with table dimensions appearing in NVDA, Narrator, JAWS) we decided to implement this solution.

### DIFF
--- a/packages/ckeditor5-widget/src/utils.ts
+++ b/packages/ckeditor5-widget/src/utils.ts
@@ -243,6 +243,7 @@ export function getLabel( element: ViewElement ): string {
  * * adds the `ck-editor__editable` and `ck-editor__nested-editable` CSS classes,
  * * adds the `ck-editor__nested-editable_focused` CSS class when the editable is focused and removes it when it is blurred.
  * * implements the {@link ~setHighlightHandling view highlight on widget's editable}.
+ * * sets the `role` attribute to `textbox` for accessibility purposes.
  *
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-widget/src/utils.ts
```diff
@@ -275,18 +276,24 @@ export function getLabel( element: ViewElement ): string {
...
  * @param options.label Editable's label used by assistive technologies (e.g. screen readers).
+ * @param options.withAriaRole Whether to add the role="textbox" attribute on the editable. Defaults to `true`.
  * @returns Returns the same element that was provided in the `editable` parameter
  */
 export function toWidgetEditable(
 	editable: ViewEditableElement,
 	writer: DowncastWriter,
 	options: {
 		label?: string;
+		withAriaRole?: boolean;
```
niegowski, line 287: I'd suggest:
```suggestion
		ariaRole?: string = 'textbox';
```

niegowski: Or maybe as other aria role does not make sense we could name it simply `ariaRole` without "with"?

niegowski (approved):

Issue #18132, opened 2025-03-14
Layout tables.


pszczesniak:
### Goal:

Allow integrators to have tables used for layouting for good email experience.

### Description

Our community often reported that our tables are too opinionated in styles, and loaded HTML templates don’t look correct. Meaning the content looks different in the editor than in the actual renderd outputed data.

The objective is to introduce a new type of table, namely layout tables, as opposed to content tables.

### Scope

- HTML markup (output data)
  - Such a table should not be wrapped in `<figure>` tag,
  - Accessibility `role="presentation"` on table.

- UI Interaction
  - By default layout table is added via editor's toolbar/menu,
  - Toggling between table types available via toolbar, table/content toolbar.

- Styling
  - Look of layout tables in the editing view should be as close as possible to what can be seen after render the outputted data.

- Configuration
  - Ability to set the preferred external table type

---
This initiative is connected to the https://github.com/cksource/ckeditor5-commercial/issues/7212

Issue #18131, opened 2025-03-14
Table Toggling Feature


Mati365:
# 🔄 Table Toggling Feature

## ✨ Overview

The **Table Toggling** feature enables seamless switching between **Content Tables** and **Layout Tables** within CKEditor. This allows for flexible table usage while maintaining a user-friendly UI.

---

## 🛠️ Functionality

- **Toggling Behavior**
  - Users can switch between **Content Table** and **Layout Table**.
  - Some data loss is expected (e.g., captions, `<th>` elements).
  - Table dimensions (columns, width, etc.) remain unchanged.
  - Integrators should NOT be warned about potential data loss.

- **UI Implementation**
  - The `tableProperties` button is transformed into a **split button**.
  - Clicking the arrow expands a dropdown menu with:
    - **Content Table**
    - **Layout Table**
  - Similar approach to the file upload icon.
  - Implemented via `TableLayout`, which overrides the default component.

---

## 🎯 Key Considerations

- **Predicting Use Cases:**  
  - Introducing toggling as an option **now** is preferable to avoid future complexity.  
- **Technical Approach:**  
  - Follows the same logic as other toolbar split buttons.  
- **Warnings & Constraints:**  
  - Inform integrators about data loss risks.  
  - Ensure a smooth transition without breaking layouts.

---

## 🏗️ Next Steps

- Implement UI changes in `tableProperties`.
- Integrate `TableLayout` override.
- Test edge cases for switching behavior.

📌 **Question:** Should we allow customization of the transition behavior (e.g., preserving some attributes)? 🤔

---

Parent issue https://github.com/ckeditor/ckeditor5/issues/18132

Mati365:
Possible follow-up for extending support in TC: https://github.com/cksource/ckeditor5-commercial/issues/7236

Issue #18126, opened 2025-03-14
Improve Drag & Drop Behavior for Tables


Mati365:
# 🔄 Improve Drag & Drop Behavior for Tables  

## 🎯 Goal  

Change the global drag-and-drop behavior for tables to ensure correct placement and handling, particularly in relation to templates.  

## 📝 Details  

- **Dragging tables (via selection handle):**  
  - When dragging a table, the entire table should be placed on drop.  
  - `ClipboardPipeline` detects whether it's a `drop` or `paste`.  
  - `TrackChanges` integration needs to be verified.  
  - Ensure that markers are preserved.  
  - Example: Moving a nested table to an adjacent row should place the entire table there.  

- **Copy-Paste Behavior:**  
  - Should **continue merging** the data as it currently does.  

### 🔍 Current Issue  

- Right now, tables get merged when inserting templates.  
- Drag-and-drop has a flag to control this behavior.  
- Template insertion could use a separate flag (quirky but feasible).  

### ✅ Expected Behavior  

- **Paste:** Inline the table within existing content (no nesting).  
- **Template Insert:** Nest the table instead of merging.  
- **Drag and Drop:** Nest the table in the target location.  

## ❓ Open Questions  

- Should we modify this behavior for other cases?  
- ✅ **Green light** for behavior changes related to template handling.

PR #18058, opened 2025-03-05
Email integration package

Mati365:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Feature (email): Initialize email integration package. 

#### Additional Information

Epic branch, do not merge. 
Commercial PR: https://github.com/cksource/ckeditor5-commercial/pull/7064

Original PR: https://github.com/ckeditor/ckeditor5/pull/17849

PR #18202, opened 2025-03-24
Email integration package (#18058)

Mati365:
Feature (table): Introduced Layout Tables to enable constructing grids with tables, for example for email editing. These tables are designed for layout purposes, and include `role="presentation"` for accessibility. Users can insert layout tables via the editor toolbar and switch between content and layout tables. The editing view now closely matches the rendered output. Closes https://github.com/ckeditor/ckeditor5/issues/18132

Feature (table): Added the ability to toggle between Content Tables and Layout Tables. Users can switch table types using a split button in the table properties UI. While captions and `<th>` elements may be lost, table structure remains intact. Closes https://github.com/ckeditor/ckeditor5/issues/18131

Feature (table): Dragging and dropping a table into another table no longer merges them. Instead, the dropped table is placed as a whole inside the target cell. Pasting tables remains unchanged. Closes https://github.com/ckeditor/ckeditor5/issues/18126

Feature (html-support): Introducing the ability to render `<style>` elements from the `<head>` section of editor data content using `FullPage` plugin. See https://github.com/ckeditor/ckeditor5/issues/13482.

Type: Message. Closes #000.

---

### Additional information


godai78 (approved):
LGTM

Issue #18250, opened 2025-03-27
[Firefox] Focus is lost when entering fullscreen using the Cmd+Shift+F keystroke


oleq:
## 📝 Provide detailed reproduction steps (if any)

0. Open a sample with a fullscreen feature
1. Focus the editor
2. Press Cmd+Shift+F

### ✔️ Expected result

The user can type right away.

### ❌ Actual result

The focus is lost. Typing is impossible. Another keystroke press is intercepted by the web browser for window management purposes.

It works on Chrome, so this is an FF-specific thing.

https://github.com/user-attachments/assets/ad295882-e745-4305-ba73-e551de5e37bb

## 📃 Other details

* Browser: FF
* OS: MacOS

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

martnpaneq:
This bug is also present in Safari.

https://github.com/user-attachments/assets/74f24410-eb39-4d83-a91a-3204da38c740

Dumluregn:
Done in https://github.com/ckeditor/ckeditor5/pull/18273.

Issue #18247, opened 2025-03-27
[Firefox] Alt+F10 keystroke to focus the toolbar does not work in fullscreen


oleq:
## 📝 Provide detailed reproduction steps (if any)

1. Go to fullscreen docs.
2. Focus the editor.
3. Use Alt+F10 (focus moves to the toolbar)
4. Enter fullscreen
5. Use Alt+F10 

### ✔️ Expected result

Focus should also move to the toolbar.

### ❌ Actual result

It does not. It works in Chrome but not in Firefox. Alt+F9 works (focuses menu bar).

## 📃 Other details

* Browser: 136.0.3 (aarch64)
* OS: MacOS

oleq:
Actually, it's even weirder than that. Touching the toolbar in the screen mode brings back the keystroke support

https://github.com/user-attachments/assets/2fdcf6db-b745-4a75-8838-04bcc042cc32

Dumluregn:
Done in https://github.com/ckeditor/ckeditor5/pull/18273.

lkszzajac:
The issue still persists:
- `Alt + F10` works only if FS is entered via `CMD + Shift + F` shortcut
- does not work at all after exiting the FS
- does not work if FS is entered via the toolbar button

Recording:

https://github.com/user-attachments/assets/6a10b330-c181-494e-bb29-3e4da4d7d2bf

scofalik:
Fixed on `release`.

PR #18273, opened 2025-03-28
Fix keystrokes behavior in fullscreen on non-Blink browsers

Dumluregn:

Internal (fullscreen): Fix Alt+F10 keystroke after entering fullscreen on non-Blink browsers. Closes https://github.com/ckeditor/ckeditor5/issues/18247.

Internal (fullscreen): Fix focus after pressing Cmd+Shift+F keystroke on non-Blink browsers. Closes https://github.com/ckeditor/ckeditor5/issues/18250.

---

### Additional information

Why issues occur:
* When switching to fullscreen using keystroke, editable is moved. 
* On Chrome, it causes the editor to fire `blur` event. 
* Then `editor.editing.view.document.isFocused` is changed to `false`. 
* After executing the command, we focus the view again using `editor.editing.view.focus()`.

* On non-Blink browsers, the `blur` event is not fired.
* `editor.editing.view.document.isFocused` is never set to `false`.
* `editor.editing.view.focus()` is executed, but since `editor.editing.view.document.isFocused`, nothing happens. The `document.activeElement` is still body.

So, there are a couple of ways to fix it:
1) Fire `blur` event.
2) Set `editor.editing.view.document.isFocused` to `false` explicitly.
3) Change `document.activeElement` manually.

I chose 2) because it seemed to be the most precise fix.


oleq (approved):
I think the fix is correct. AFAIR, different browsers fire (or don't...) different events when a focused element is removed/hidden in DOM. So our `DomEventObserver`/`FocusObserver` doesn't get a chance to update the state of the document. 

Run this sandbox in Chrome, then in Firefox to see the difference https://codesandbox.io/p/sandbox/4jtsdz.

For the reference, see the discussion in https://github.com/jsdom/jsdom/issues/2931.

PR #18295, opened 2025-03-31
Remove `overflow: hidden` from table while using TableColumnResize.

### LINKED ISSUE #18268: [Layout tables] Widget elements don't fit into layout table cells

1. Open e.g. http://fake.ckeditor.com:8080/ckeditor5/44.3.0/features/email-editing/email.html#demo
2. Resize the image at the top of the table and insert some other widgets

### ✔️ Expected result

Widgets and their borders are fully visible and their borders are not cut

### ❌ Actual result

Some borders extend beyond the layout table, its cells, or overlap with the table borders, and some parts of the elements are cut (e.g. the resize handles or the size indicators):

https://github.com/user-attachments/assets/c0325b39-a0e7-4437-8dc8-17748ee97494

It's particularly visible on Safari, where, for example, the upper border of the image is not visible at all and the resize handles are cut the most:

Image

## 📃 Other details

* Browser: Any
* OS: MacOS

---

pszczesniak:

Fix (table): Widgets UI elements should be visible when are inside tables. Closes #18268.

---

### Additional information

⚠️ `release` branch is the target.

ℹ️  To prevent cut off any element that lies inside a layout table (or content table) we decided to calculate the table resizer on demand instead of using `overflow: hidden` hack with huge value of top and bottom to make sure that resizer element will be streched to the top and bottom of the table.

### DIFF
--- a/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.ts
+++ b/packages/ckeditor5-table/src/tablecolumnresize/tablecolumnresizeediting.ts
@@ -13,6 +13,7 @@ import {
 	global,
 	DomEmitterMixin,
 	Rect,
+	toUnit,
 	type EventInfo,
 	type DomEmitter,
 	type ObservableChangeEvent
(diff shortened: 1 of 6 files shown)


niegowski (approved):

gorzelinski (commented):
LGTM

Issue #18246, opened 2025-03-27
Fullscreen should restore the window/container scroll position while closing


oleq:
## 📝 Provide detailed reproduction steps (if any)

1. Go to the Fullscreen feature docs
2. Scroll the web page
3. Enter fullscreen in one of the editors
4. Close fullscreen

### ✔️ Expected result

The scroll position of the webpage should be the same it was at 3.

### ❌ Actual result

It resets to the top. It may get very annoying for the users using editors further down the webpage (see the one in the " Customized layout - pageless editor demo" section).

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Dumluregn:
Related: https://github.com/ckeditor/ckeditor5/issues/18206

Dumluregn:
Fixeed on release in https://github.com/ckeditor/ckeditor5/pull/18306.

PR #18301, opened 2025-04-01
Make fullscreen scroll to the selection upon entering and leaving

Dumluregn:

Internal (fullscreen): Fullscreen will scroll to the selection upon entering and leaving. Closes https://github.com/ckeditor/ckeditor5/issues/18246.

---

### Additional information

* Fullscreen will scroll to the selection upon entering and leaving using toolbar button, menu bar button or keystrokes.
* It will try to align the editor at the top of the page.
* It uses the `config.ui.viewportOffset` if used to take into account the navbars etc (see our docs where it's set to 100px).
* `ancestorOffset` is used to scroll the view enough to display at least the editor toolbar (in classic editor at least, see manual test).
* Small CSS fixes: for table margin in fullscreen mode and smaller padding for pageless demo.

### DIFF
--- a/packages/ckeditor5-fullscreen/docs/_snippets/features/fullscreen-default.html
+++ b/packages/ckeditor5-fullscreen/docs/_snippets/features/fullscreen-default.html
@@ -255,6 +255,10 @@
 		word-break: break-word;
 	}
 
+	.editor-content.ck.ck-content table {
+		margin: 0;
+	}
+
 	.editor-content.ck.ck-content,
 	.editor-content.ck.ck-content.ck-focused {
 		border: 1px solid hsl(0, 0%, 88%);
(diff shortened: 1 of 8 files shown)


scofalik:
I don't think this works as intended and it may still lead to confusion. The browser tab scroll is not saved and restored, which means that after leaving fullscreen I may end up in a bit different place. Can't we simply store `window.scrollY` and then restore it? (And then possibly scroll the editor content.)

See the video below. I have small editor with some extra content around. After leaving fullscreen I have my page scrolled more towards top (the editor is at the top browser tab edge). I would expect to see exactly what I've seen when I turned on fullscreen (unless the website structure/sizes didn't change in the meantime, of course).

https://github.com/user-attachments/assets/8a7da387-3844-431a-9a1a-454cbb94402d

Also, I am not sure if viewport offset works correctly (see the end of video) but maybe that's just incorrect configuration.

scofalik:
Additionally, I am not 100% convinced that we should scroll to selection on entering/leaving fullscreen, instead we should try to figure out what was visible (so we should base on the editor content scroll too). That may be more complicated so we can leave it as a followup.

scofalik:
> Can't we simply store `window.scrollY` and then restore it? (And then possibly scroll the editor content.)

Or rather we should take/apply it depending on where the editor was put (default vs custom container).

Dumluregn:
Closing in favor of https://github.com/ckeditor/ckeditor5/pull/18306.

Issue #18259, opened 2025-03-27
[Fullscreen] Preserve the scroll position in the editing when entering the fullscreen mode


pszczesniak:
## 📝 Provide a description of the improvement

It would be nice to preserve the scroll position in the editing when entering the fullscreen mode. Currently it's always scrolled to top.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

Dumluregn:
Fixeed on release in https://github.com/ckeditor/ckeditor5/pull/18306.

PR #18306, opened 2025-04-02
Improve fullscreen scrolling

Dumluregn:

Internal (fullscreen): Fullscreen will restore the window and container scroll position while closing. Closes https://github.com/ckeditor/ckeditor5/issues/18246.

Internal (fullscreen): Fullscreen will preserve the scroll to the current selection in the editable when toggling fullscreen. Closes https://github.com/ckeditor/ckeditor5/issues/18259.

---

### Additional information

* On entering fullscreen mode, it will save the scroll position of all (scrollable) ancestors and restore them after leaving.
* Fullscreen will scroll the editor contents to the selection upon entering and leaving. 
* Small CSS fixes: for table margin in fullscreen mode and smaller padding for pageless demo.

### DIFF
--- a/packages/ckeditor5-fullscreen/src/handlers/abstracteditorhandler.ts
+++ b/packages/ckeditor5-fullscreen/src/handlers/abstracteditorhandler.ts
@@ -677,9 +677,16 @@ export default class AbstractEditorHandler {
 		}
 
 		while ( element ) {
-			const overflow = element.style.overflowY || global.window.getComputedStyle( element ).overflowY;
+			const overflowY = element.style.overflowY || global.window.getComputedStyle( element ).overflowY;
+			const overflowX = element.style.overflowX || global.window.getComputedStyle( element ).overflowX;
 
-			if ( overflow === 'auto' || overflow === 'scroll' ) {
+			// Out of 5 possible keyword values: visible, hidden, clip, scroll and auto - only the last two allow for scrolling.
+			if (
+				overflowY === 'auto' ||
+				overflowY === 'scroll' ||
+				overflowX === 'auto' ||
+				overflowX === 'scroll'
+			) {
 				this._savedAncestorsScrollPositions.set( element, {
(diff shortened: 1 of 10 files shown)


packages/ckeditor5-fullscreen/src/handlers/abstracteditorhandler.ts
```diff
@@ -650,4 +665,33 @@ export default class AbstractEditorHandler {
...
+		let element = domElement.parentElement;
+
+		if ( !element ) {
+			return;
+		}
+
+		while ( element ) {
+			const overflow = element.style.overflowY || global.window.getComputedStyle( element ).overflowY;
+
+			if ( overflow === 'auto' || overflow === 'scroll' ) {
```
scofalik, line 682: Why do we check only for these overflow values? Is it because otherwise element can't be scrolled (and/or the `scrollTop` and `scrollLeft` values are always `0`) or is it because it produced quirks on non-scrollable / overflow:hidden elements if we tried to scroll them?

Although in general I agree with this code, I wonder if we are not accidentally missing something by this kind of "filtering".

BTW, shouldn't we also check `overflowX`? It can be set independently from `overflowY`, so it seems we should check both in case we have applications with horizontal scrolls, and the parent was scrolled horizontally only.

Dumluregn: I based on https://github.com/ckeditor/ckeditor5/blob/6eca87a05212e01c7067426d41dc21a3a73543af/packages/ckeditor5-utils/src/dom/findclosestscrollableancestor.ts#L18-L39

with an extension that we're also interested in window (<html> element) scroll, where the difference is it can be scrollable even without overflow property set.

Out of 5 possible keyword values: `visible`, `hidden`, `clip`, `scroll` and `auto` - only the last two allow for scrolling.

As for the `overflowX` part, I agree it's possible - I'll enhance the checks.

PR #18317, opened 2025-04-03
Fix Alt+F10 behavior with fullscreen on non-Blink browsers

Dumluregn:

Internal (fullscreen): Fix Alt+F10 behavior with fullscreen on non-Blink browsers. Closes https://github.com/ckeditor/ckeditor5/issues/18247.

---

### Additional information

Continuation of https://github.com/ckeditor/ckeditor5/pull/18273. 

Turns out, we also need to manually reset the toolbar `focusedElement` - it isn't set to `null` when toggling fullscreen, so if you do it using a button or shortcut but with toolbar focused, then toolbar can't be "refocused" later by Alt+F10.

### DIFF
--- a/packages/ckeditor5-fullscreen/src/fullscreenediting.ts
+++ b/packages/ckeditor5-fullscreen/src/fullscreenediting.ts
@@ -63,6 +63,7 @@ export default class FullscreenEditing extends Plugin {
 				this.editor.ui.view.toolbar!.focusTracker.focusedElement = null;
 			}
 
+			// The order of scroll and focus is not important here.
 			this.editor.editing.view.scrollToTheSelection();
 			this.editor.editing.view.focus();
 
--- a/packages/ckeditor5-fullscreen/src/fullscreenediting.ts
+++ b/packages/ckeditor5-fullscreen/src/fullscreenediting.ts
@@ -55,15 +55,16 @@ export default class FullscreenEditing extends Plugin {
 		this.editor.keystrokes.set( 'Ctrl+Shift+F', ( evt, cancel ) => {
 			this.editor.execute( 'toggleFullscreen' );
 
-			// On non-Chromium browsers, the editor view is not blurred properly after moving the editable,
-			// even though the `document.activeElement` is changed. Hence we need to blur the view manually.
+			// On non-Chromium browsers, the editor view and toolbar are not blurred properly after moving the editable,
+			// even though the `document.activeElement` is changed. Hence we need to blur them manually.
 			// Fixes https://github.com/ckeditor/ckeditor5/issues/18250 and https://github.com/ckeditor/ckeditor5/issues/18247.
 			if ( !env.isBlink ) {
 				this.editor.editing.view.document.isFocused = false;
+				this.editor.ui.view.toolbar!.focusTracker.focusedElement = null;
 			}
 
-			this.editor.editing.view.focus();
 			this.editor.editing.view.scrollToTheSelection();
+			this.editor.editing.view.focus();
 
 			cancel();
 		} );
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-fullscreen/src/fullscreenediting.ts
```diff
@@ -55,15 +55,16 @@ export default class FullscreenEditing extends Plugin {
...
+			// even though the `document.activeElement` is changed. Hence we need to blur them manually.
 			// Fixes https://github.com/ckeditor/ckeditor5/issues/18250 and https://github.com/ckeditor/ckeditor5/issues/18247.
 			if ( !env.isBlink ) {
 				this.editor.editing.view.document.isFocused = false;
+				this.editor.ui.view.toolbar!.focusTracker.focusedElement = null;
 			}
 
-			this.editor.editing.view.focus();
 			this.editor.editing.view.scrollToTheSelection();
+			this.editor.editing.view.focus();
```
scofalik, line 67: If the order here is important maybe it makes sense to mention why in a comment?

packages/ckeditor5-fullscreen/src/fullscreenui.ts
```diff
@@ -82,8 +83,16 @@ export default class FullscreenUI extends Plugin {
...
+
+			// On non-Chromium browsers, toolbar is not blurred properly after moving the editable,
+			// even though the `document.activeElement` is changed. Hence we need to blur the view manually.
+			// Fixes https://github.com/ckeditor/ckeditor5/issues/18250 and https://github.com/ckeditor/ckeditor5/issues/18247.
+			if ( !env.isBlink ) {
+				this.editor.ui.view.toolbar!.focusTracker.focusedElement = null;
+			}
+
 			editor.editing.view.scrollToTheSelection();
+			editor.editing.view.focus();
```
scofalik, line 95: AFAIU, we only blur toolbar (not view document) because we clicked a button in the toolbar so we know that toolbar exactly is focused?

Also, I wonder, couldn't this be a part of the command execution?

Finally, did you think about other scenarios where something may have need to be blurred? Menu bar comes to my mind 🤔.

Dumluregn: > AFAIU, we only blur toolbar (not view document) because we clicked a button in the toolbar so we know that toolbar exactly is focused?

Yes

> Also, I wonder, couldn't this be a part of the command execution?

Hm, I wanted to keep this separate as it's UI-related so the command works when there is no view... but I just realised it totally doesn't make sense for this particular command. I'll move it.

> Finally, did you think about other scenarios where something may have need to be blurred? Menu bar comes to my mind 🤔.

Yes, I tested it with menu bar but it works completely different than toolbar. It allows to focus the editor while still being open (using tab) so I think we shouldn't touch it.
I found one more case for missing blur: open some dropdown (e.g. heading) and toggle fullscreen using the shortcut. Dropdown will stay open even though the focus is properly moved to the editable. Unfortunately I don't have an easy fix for this, it seems we don't really have a simple way to access this open dropdown view and blur it. But I propose to treat it as a followup/backlog issue, it's not easy to reproduce (remember it's still only Firefox and Safari) and nothing blows up.

packages/ckeditor5-fullscreen/src/fullscreenediting.ts
Dumluregn: Actually it doesn't, I changed it in the meantime during testing but it can be both ways - which I think is still worth a comment.

packages/ckeditor5-fullscreen/src/fullscreenui.ts
Dumluregn: > Hm, I wanted to keep this separate as it's UI-related so the command works when there is no view... but I just realised it totally doesn't make sense for this particular command. I'll move it.

I still have second thoughts about it. We don't really include things like focus management in commands. I'd rather keep it in the UI layer. Sorry for changing my mind.

scofalik: On my second thought I think that maybe not even in command but maybe in *Handlers. But let's leave this for possible future change.

> We don't really include things like focus management in commands.

You are generally correct, because a command usually should change something in the document and this is not connected in anyway with UI but with document. However in this case, the whole feature is about managing the UI, so that's a bit different case.

PR #18346, opened 2025-04-07
Speed-up rendering of dropdowns with large amount of items.

### LINKED ISSUE #18094: Performance regression of long dropdown lists

1. Configure the heading feature to include 2000 items
```js
heading: {
	options: Array( 2000 ).fill( null ).map( ( x, i ) => (
		{ model: 'heading' + i, view: 'h2', title: 'Paragraph ' + i, class: 'ck-heading_paragraph' }
	) )
}
```
3. In editor open headings list from toolbar
### ✔️ Expected result

The dropdown should open asap.

### ❌ Actual result

It takes almost 2s and in editor version 37.0.1 it took less than 400ms.

## 📃 Other details

* Browser: …
* OS: …

---

Mati365:

Fix (ui): Optimized performance of dropdowns when opening with many items. Closes #18094  

---

### Additional information

Previously, after adding each item, the entire list of other items was checked and their flags were updated accordingly. This resulted in behavior close to O(n²) complexity. In the new solution, only the added (or removed) items are checked, and the full item list is updated only once when a new toggleable is added. This change reduced the processing time from **5.6s** to **1.6s**.

#### Before

obraz

#### After

obraz

### DIFF
--- a/packages/ckeditor5-ui/src/dropdown/utils.ts
+++ b/packages/ckeditor5-ui/src/dropdown/utils.ts
@@ -712,7 +712,7 @@ function bindViewCollectionItemsToDefinitions(
  *
  * Buttons in dropdowns have reserved space for a check icon when they are toggleable.
  * When at least one button in the list is toggleable, all other buttons (even non-toggleable ones)
- * will have space allocated on their left side to align properly with toggleable buttons.
+ * will have space on their left side to align with toggleable buttons.
  *
  * This function handles a special case where a new toggleable button is added (or removed) to a list
  * where previous buttons weren't toggleable. In that case, those previous buttons will
@@ -754,7 +754,7 @@ function bindDropdownToggleableButtonsAlignment( listItems: ViewCollection ) {
 	};
 
 	// Helper function that checks if a view item is a toggleable button.
-	// Returns the button if it's toggleable, otherwise returns null.
+	// Returns the button if it's toggleable - otherwise, returns null.
 	const pickListItemToggleableButtonIfPresent = ( item: View<HTMLElement> ) => {
 		const listItemButtonView = pickListItemButtonIfPresent( item );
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-ui/src/dropdown/utils.ts
```diff
@@ -725,6 +707,117 @@ function bindViewCollectionItemsToDefinitions(
...
+ * |    Another Button |
+ * +-------------------+
+ * | ✓ Toggle Button   |
+ * +-------------------+
+ * ```
+ *
+ * @param listItems Collection of list items to observe for toggleable buttons.
+ */
+function bindDropdownToggleableButtonsAlignment( listItems: ViewCollection ) {
+	// Keep track of how many toggleable buttons are in the list
```
Mati365, line 743:
```suggestion
	// Keep track of how many toggleable buttons are in the list.
```

stefanseeger:
Hello,
thank you for fixing this!
I do not really know the internals here. For me the 'hasCheckSpace' would not be something that I would expect with all the buttons. For me this sounds like something css coild fix without a flag.

I might be totally wrong, but just wanted to share my thoughts.

Mati365:
@stefanseeger Hi! Thanks — unfortunately, CSS alone isn't enough to solve this issue. We display a :heavy_check_mark: icon before text content of dropdown buttons, but the tricky part is that not all dropdown items are toggleable. In other words, if a menu initially contains only non-toggleable items, and we later add a toggleable one, we need to allocate some placeholder space for the non-toggleable items to maintain alignment (as it's described in comment).

stefanseeger:
@Mati365 yes, I understand. 

I expected it to be not that easy. Just was hoping for it :) 

Thanks for your time.

stefanseeger:
I'm on my phone so I cannot try it. 

But if you would use a css grid. Where the checkbox would be in the first column and the label in the second. And the checkbox column has a `min-width: 0`. Then it should move all labels right when atleast one has a checkbox. And otherwise no spacing.

But I can't test it. So might not work at all.

Have a nice weekend anyways.

gorzelinski (approved):

Mati365:
Thx, merging. Potential follow-ups will be done in separate PRs.

PR #18349, opened 2025-04-08
Add missing consume for `image_resize` class on image elements.

### LINKED ISSUE #18287: Image resize is not correctly consumed with GHS

1. Create an image (block or inline)
2. Resize the image.
3. Get data and Set data.

https://github.com/user-attachments/assets/ba5abbdc-676d-4bd0-b020-9279f866f2bb

https://github.com/user-attachments/assets/2aba2c71-c3e0-41c1-9f57-a5fe54fbc6b6

### ✔️ Expected result

Attributes correctly consumed to the model.

### ❌ Actual result

Attributes consumed by GHS in case of a resize.

### ❓ Possible solution

## 📃 Other details

* Browser: …
* OS: …

---

Mati365:

Fix (image): Consume `image_resize` class and `aspect-ratio` style during the upcast of the images. Closes #18287 

---

### Additional information

Previously, the `image_resize` class was incorrectly consumed by the General HTML Support (GHS) plugin. This class is meant to be handled by the `ImageResize` plugin only, as it is automatically applied by it. This PR ensures that GHS no longer consumes this attribute.

### DIFF
--- a/packages/ckeditor5-image/src/imageresize/imageresizeediting.ts
+++ b/packages/ckeditor5-image/src/imageresize/imageresizeediting.ts
@@ -210,5 +210,12 @@ export default class ImageResizeEditing extends Plugin {
 					}
 				}
 			} );
+
+		editor.conversion.for( 'upcast' )
+			.add( dispatcher => {
+				dispatcher.on( `element:${ imageType === 'imageBlock' ? 'figure' : 'img' }`, ( evt, data, conversionApi ) => {
+					conversionApi.consumable.consume( data.viewItem, { classes: [ 'image_resized' ] } );
+				} );
+			} );
 	}
 }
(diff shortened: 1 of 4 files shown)


Witoso:
Notice `aspect-ratio` is not consumed as well in the second video.

Witoso:
`aspect-ratio` consumption should not be connected to the resize, if you load image, and setData/getData it's also not consumed.

Mati365:
@Witoso Isn't it too generic assumption? I enforced it to maximize backward compatibility, as `aspect-ratio` might be part of "plain" GHS styling.

Witoso:
I would need @mmotyczynska to check but this looks as an improper working of the image consumption. Even if GHS styles something, we normalize it to the feature.

Mati365:
I moved `aspect-ratio` consume to `ImageSizeAttributes`. fyi @niegowski

mmotyczynska:
> `aspect-ratio` consumption should not be connected to the resize

True.

arkflpc (approved):

Issue #18180, opened 2025-03-20
[Windows] Mixed local and online images from MSWord paste cause some images to disappear


arkflpc:
## 📝 Provide detailed reproduction steps (if any)

1. Copy a section from a Word document that contains both locally stored images (`file://`) and online images.
2. Paste the content into CKEditor 5’s demo page or an editor using the paste-from-office plugin.
3. Observe that only the online images load, while the local file-based images are missing.

[Image Paste Test.docx](https://github.com/user-attachments/files/19367046/Image.Paste.Test.docx)

### ✔️ Expected result

All images (both local and online) should be successfully embedded and displayed in the editor after pasting.

### ❌ Actual result

Only online images load correctly, while local file-based images are missing after pasting the entire document. However, if only the local file-based images are copied separately, they load correctly.

### ❓ Possible solution

The issue is caused by how the [`replaceImagesFileSourceWithInlineRepresentation` function](https://github.com/ckeditor/ckeditor5/blob/27495f120fbac3bf9adaa0f737da4c3f0a116672/packages/ckeditor5-paste-from-office/src/filters/image.ts#L311) processes images. It assumes that all images in a pasted Word document are either local files or online images, without considering a mix of both. The function needs to properly track image sources and handle file:// references separately from online images.

## 📃 Other details

* Browser: …
* OS: Windows
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

PR #18363, opened 2025-04-10
Improve image matching in RTF content pasted from MS Word.

Mati365:

Fix (paste-from-office): Mixed local and online images from MSWord paste no longer cause some images to disappear. Closes https://github.com/ckeditor/ckeditor5/issues/18180

---

#### Root cause
Clipboard content from Excel included a mix of images: some embedded as base64 in text/html, others referenced via file:// paths. Our previous logic assumed that if a file:// image is detected, the actual image data could be extracted from the RTF part of the clipboard. This worked when all images were file-referenced. However, when formats were mixed (base64 + file references), the fallback to RTF failed.

#### Fix
Improved the image extraction heuristic to correctly handle mixed image sources (base64 + file paths). It now parses all available formats and prioritizes actual image data when available.

#### Limitations
On some Windows systems, Chrome may block access to RTF clipboard content due to security policies. In those cases, only the file:// path is available, with no embedded image data. This results in images not being pasted correctly. This issue does not occur in Firefox and is less likely in documents with minimal styling.


Mati365:
I ordered QA tests.

kubaklatt:
From what I can see on Windows the inline images from [the original issue](https://github.com/ckeditor/ckeditor5/issues/18180) don't load properly on Chrome and Edge. On Firefox they load properly:

image

Mati365:
Thx, @kubaklatt it's known limitation of clipboard behavior on Chrome. RTF tends to be blocked on Chrome on Windows for documents with custom styling. It's a limitation that we cannot bypass, as the information about image content is lost.

arkflpc (approved):

PR #17809, opened 2025-01-24
Fix: Nested editable on Safari.

gorzelinski:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Fix: Nested editable on Safari. Closes https://github.com/ckeditor/ckeditor5/issues/17383.

---

### Additional information

_For example – encountered issues, assumptions you had to make, other affected tickets, etc._

Issue #17383, opened 2024-11-04
Deleting an empty block at the beginning of an editable block widget removes itself and parent widget


lkszzajac:
## 📝 Provide detailed reproduction steps (if any)

1. Open demo for block widget implementation example in docs: https://ckeditor.com/docs/ckeditor5/latest/framework/tutorials/widgets/implementing-a-block-widget.html#demo on Safari.
2. Delete Title box contents.
3. Click Backspace when Title box is empty.

### ✔️ Expected result

- Title box is not removed

### ❌ Actual result

- Title box is removed
- Entire block widget is removed as well
- Recording:

https://github.com/user-attachments/assets/4f4a11ea-bcbe-4cb3-9431-411f17f2b716

## 📃 Other details

* Browser: Safari
* OS: macOS
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

niegowski:
The problem is that we get the delete `beforeinput` event target range crossing the limit element (it starts before the widget and ends in the first editable). Then it's correctly mapped to the view range but after that [this range is fixed](https://github.com/ckeditor/ckeditor5/blob/6fde22b9391c0d8fef6054502c9d11c9d7dfca37/packages/ckeditor5-engine/src/controller/editingcontroller.ts#L257-L277) because range is not supposed to cross the limit element boundaries. This could be fixed by hinting the selection fixing with the desired direction of the fix so it would not expand the selection, it should collapse it in such case.

niegowski:
We could implement a temporary workaround, sth like this:

```typescript
this.listenTo<ViewDocumentKeyDownEvent>( viewDocument, 'keydown', ( evt, data ) => {
	if ( viewDocument.isComposing || data.keystroke != keyCodes.backspace || !editor.model.document.selection.isCollapsed ) {
		return;
	}

	const ancestorLimit = editor.model.schema.getLimitElement( editor.model.document.selection );
	const limitStartPosition = editor.model.createPositionAt( ancestorLimit, 0 );

	if ( limitStartPosition.isTouching( editor.model.document.selection.getFirstPosition()! ) ) {
		data.preventDefault();
	}
} );
```

The same bug can be reproduced while pressing backspace in an empty image caption.

gorzelinski:
I dived into the problem a bit and found out interesting cases.
The proposed solution works, but there is an exception on Safari.

## Chrome

[...]

PR #18366, opened 2025-04-10
No longer keep todo list items placed on start of the editable when press backspace.

### LINKED ISSUE #18356: Removing the first checkbox from the to-do list is not possible / Removal of first list item or element

1. Open [feature-rich editor demo](https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/full-featured-editor.html)
2. Enter the to-do list and try to remove the first checkbox

### ✔️ Expected result

The checkbox is removed

### ❌ Actual result

There is no option to remove the first checkbox

https://github.com/user-attachments/assets/7dd99482-9cb3-4cdc-9756-a674814648f6

## 📃 Other details

* Browser: Any
* OS: Any
* First affected CKEditor version: 44.3.0

---

Mati365:

Fix (typing): The list items and headings at the beginning of the editable area can be changed to a plain paragraph on backspace keypress. Closes #18356.

---

### Additional information

Caused by https://github.com/ckeditor/ckeditor5/issues/17383 and https://github.com/ckeditor/ckeditor5/pull/17809

### DIFF
--- a/packages/ckeditor5-typing/src/delete.ts
+++ b/packages/ckeditor5-typing/src/delete.ts
@@ -95,11 +95,29 @@ export default class Delete extends Plugin {
 			}
 
 			const ancestorLimit = editor.model.schema.getLimitElement( modelDocument.selection );
+			const selectionPosition = modelDocument.selection.getFirstPosition()!;
 			const limitStartPosition = editor.model.createPositionAt( ancestorLimit, 0 );
 
-			if ( limitStartPosition.isTouching( modelDocument.selection.getFirstPosition()! ) ) {
-				data.preventDefault();
+			// If the selection is not at the beginning of the nested editable, do nothing.
+			if ( !limitStartPosition.isTouching( selectionPosition ) ) {
+				return;
 			}
+
+			// Workaround for Safari where pressing Backspace at the beginning of a nested editable
+			// can move the selection to the parent element and delete the entire element.
+			// This issue primarily affects empty paragraphs without attributes, while elements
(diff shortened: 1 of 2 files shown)


gorzelinski (dismissed):
I tested it on different browsers with key combinations - and it seems to work fine.

Witoso:
Please check it with all list items, not just todo.

Mati365:
@Witoso Been that, did that.

packages/ckeditor5-typing/src/delete.ts
```diff
@@ -95,11 +95,29 @@ export default class Delete extends Plugin {
...
+			// See: https://github.com/ckeditor/ckeditor5/issues/18356
+			if (
+				selectionPosition.parent !== ancestorLimit &&
+				selectionPosition.parent.is( 'element' ) &&
+				[ ...selectionPosition.parent.getAttributeKeys() ].length
+			) {
+				return;
 			}
+
+			data.preventDefault();
```
niegowski, line 120: Maybe a less synthetic solution would be to prevent-default the keydown event (as we do here). This would stop the invalid `beforeinput` event. Still, instead of detecting whether a parent has attributes, we could fire a `ViewDocumentDeleteEvent` (as listened above) that would replace the one generated with invalid target ranges. As we fire the `delete` event from the `DeleteObserver`, I'd suggest moving this keydown listener to `DeleteObserver`.

niegowski (changes requested):
As commented above.

arkflpc (approved):

PR #18373, opened 2025-04-11
Fix formatting when processing single-line pre tags.

### LINKED ISSUE #18360: Single line `<pre></pre>` causes loss of indentation on subsequent lines in source mode
The code formatter used when viewing Source mode does not correctly indent lines if a `<pre>` tag is encountered that both opens and closes on a single line. Lines before the `<pre>` tag will be indented correctly, lines after the `<pre>` tag will lose indentation inside of the tag (though the lines with the tags themselves are still indented correctly).

1. Use the demo at https://ckeditor.com/ckeditor-5/demo/html-support/
2. Create 3 paragraphs of text. Select the middle paragraph and click "Insert code block" (forth button from right) to make it a `<pre>` tag. In the editor it should look like this:
Image
3. Click "Edit source" (second button from left).

### ✔️ Expected result

The code should be formatted as follows in source mode:
```
    <p>
        paragraph before
    </p>
    <pre><code class="language-plaintext">single line of code</code></pre>
    <p>
        paragraph before
    </p>
```

### ❌ Actual result

The first paragraph is indented correctly, but the second paragraph's contents loose their indentation within the `<p>` tags.

```
    <p>
        paragraph before
    </p>
    <pre><code class="language-plaintext">single line of code</code></pre>
    <p>
paragraph before
    </p>
```

Many subsequent tags are also missing their indentation when this occurs.

Below is an annotated screenshot of the incorrect markup.

Image

This only occurs if there is a single line of code sample. If there are multiple lines, it works fine.

### ❓ Possible solution

The [isPreformattedBlockLine()](https://github.com/ckeditor/ckeditor5/blob/93678a20d107ef3f203c856e28342cc5c289a1bb/packages/ckeditor5-source-editing/src/utils/formathtml.ts#L156) method in `formathtml.ts` has the following code block:

```
	if ( new RegExp( '<pre( .*?)?>' ).test( line ) ) {
		return 'first';
	} else if ( new RegExp( '</pre>' ).test( line ) ) {
		return 'last';
	} else if ( isPreviousLinePreFormatted === 'first' || isPreviousLinePreFormatted === 'middle' ) {
		return 'middle';
	} else {
		return false;
	}
```

What's happening here is that it is failing to identify the situation where both the opening `<pre>` and closing `</pre>` are _the same line_. Instead, it's returning `first`. What then happens is the rest of the formatter code thinks it's permanently inside a `<pre>` tag, because it will never find the closing `</pre>` tag (since it was on the same line as the opening tag).

## 📃 Other details

* Browser: Firefox/Chrome
* OS: Ubuntu 24.04

---

quicksketch:

Fix: Formatting when processing single-line pre tags. Closes #18360.


quicksketch:
Sure, I'll give it a shot. I'll run them locally, of course, but could you enable test running in this PR?

quicksketch:
@Witoso I can't seem to get either `master` or `stable` to build past TypeScript validation with the error `Property 'showRevisionViewerCallback' does not exist on type 'RevisionHistoryConfig'` (and other TypeScript errors related to revision handling). I'm going to pause this work.

packages/ckeditor5-source-editing/src/utils/formathtml.ts
```diff
@@ -154,15 +154,23 @@ function indentLine( line: string, indentCount: number, indentChar: string = '
  * @param isPreviousLinePreFormatted Information on whether the previous line was preformatted (and how).
  */
 function isPreformattedBlockLine( line: string, isPreviousLinePreFormatted: 'first' | 'last' | 'middle' | false ) {
+	let preLineType = false;
```
quicksketch, line 157:
```suggestion
	let preLineType: 'first' | 'last' | 'middle' | false = false;
```

Witoso:
We are investigating, thanks!

Reinmar:
I executed tests on a fresh clone of ckeditor5. No issues with building manual tests too.

@quicksketch, maybe you could rebase your branch? We might've had some dependency to the closed-source repo for a short while, but all seems to be fine at the moment. You shouldn't have any issues with TypeScript, tests or whatever.

In case you'll have any more issues, please provide more details on where that happens exactly (what CLI commands).


```
➜  ckeditor5 git:(master) yarn test --files=source-editing
yarn run v1.22.22
$ node --max_old_space_size=4096 node_modules/@ckeditor/ckeditor5-dev-tests/bin/testautomated.js --files=source-editing
(node:87300) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
⚠ You're running tests in dev mode - some error protections are loose. Use the `--production` flag to use strictest verification methods.
karma-webpack does not support custom optimization configurations for webpack, the optimizations passed will be ignored.

START:
Webpack bundling...
asset commons.js 13.7 MiB [emitted] (name: commons) (id hint: commons)
asset runtime.js 13.6 KiB [emitted] (name: runtime)
asset entry-point.879525208.js 472 bytes [emitted] (name: entry-point.879525208)
Entrypoint entry-point.879525208 13.7 MiB = runtime.js 13.6 KiB commons.js 13.7 MiB entry-point.879525208.js 472 bytes
webpack 5.99.7 compiled successfully in 1884 ms
30 04 2025 10:09:39.136:INFO [karma-server]: Karma v6.4.4 server started at http://localhost:9876/
30 04 2025 10:09:39.136:INFO [launcher]: Launching browsers CHROME_LOCAL with concurrency unlimited
30 04 2025 10:09:39.144:INFO [launcher]: Starting browser Chrome
30 04 2025 10:09:44.450:INFO [Chrome 135.0.0.0 (Mac OS 10.15.7)]: Connected on socket jqQPX8bwqnmWrm2KAAAB with id 79404617
  SourceEditing
    ✔ should be named
    ✔ should have `isOfficialPlugin` static flag set to `true`
.....
......
      ✔ should not inject extra white spaces at the beginning of preformatted lines in <pre>
      ✔ should not inject extra white spaces at the beginning of preformatted lines in <pre> (deep structure)
      ✔ should keep all attributes unchanged
      ✔ should not crash when a pathological <iframe> content appears in source

Finished in 0.993 secs / 0.123 secs @ 10:09:45 GMT+0200 (Central European Summer Time)

SUMMARY:
✔ 73 tests completed
✨  Done in 11.53s.
➜  ckeditor5 git:(master)
```

niegowski:
This PR was extracted and fixed in https://github.com/ckeditor/ckeditor5/pull/18480. Thank you @quicksketch

quicksketch:
Perfect, thank you @niegowski!

PR #18462, opened 2025-04-30
Docs: Introduced the Spring Boot integration guides.

### LINKED ISSUE #18461: Prepare guides for integration with Spring Boot

We need to prepare guides for the integration of CKEditor 5 with Spring Boot in the Getting started -> Installation section.

lkszzajac:

Docs: Introduced the Spring Boot integration guides. Closes #18461.


godai78:
Is `master` the right target? As this is docs only, we usually push it straight to `stable` unless it needs to wait for the release for some reason?

docs/getting-started/integrations-cdn/spring-boot.md
```diff
@@ -0,0 +1,218 @@
...
+		Bold,
+		Italic,
+		Font,
+		Paragraph
+	} = CKEDITOR;
+	const { FormatPainter } = CKEDITOR_PREMIUM_FEATURES;
+
+	ClassicEditor
+		.create( document.querySelector( '#editor' ), {
+			licenseKey: '<YOUR_LICENSE_KEY>',
```
godai78, line 120: We may want to adress the GPL licence, too.

godai78 (approved):

docs/getting-started/integrations-cdn/spring-boot.md
Witoso: It's CDN 🙃

Witoso:
+1 to target stable.

docs/getting-started/integrations-cdn/spring-boot.md
godai78: Then we may want to rephrase the paragraph above about premium features instead, as they are not the case point, and now it looks like you only need a license when you use them.

godai78:
I'll rebase it by hand on local.

godai78 (changes requested):
As mentioned in the comment about licensing.

lkszzajac:
> Also any reason for the ordering?
Not really, I was basing this on the Laravel guide and I increased the value by 5 only 😅 I adjusted it to reflect the guide creation timeline better.

godai78 (approved):

godai78:
What's with the CLA? I believe you had it fixed before?

godai78:
I'll try crunching it again, and we'll see the effect.

godai78 (approved):

Mgsy:
As I can see, guides are missing the _New_ label in the sidebar:

image

docs/getting-started/integrations-cdn/spring-boot.md
```diff
@@ -0,0 +1,218 @@
...
+│   │       ├── templates
+│   │       │   ├── index.html
+│   │       │   └── ...
+│   │       └── application.properties
+│   └── test
+├── pom.xml
+└── ...
+```
+
+## Adding CKEditor5 container, scripts and styles
```
Mgsy, line 67:
```suggestion
## Adding CKEditor&nbsp;5 container, scripts and styles
```

docs/getting-started/integrations/spring-boot.md
```diff
@@ -0,0 +1,165 @@
...
+		<artifactId>spring-boot-starter-thymeleaf</artifactId>
+	</dependency>
+</dependencies>
+```
+
+## Integrating using ZIP
+
+<info-box info>
+	Our new CKEditor&nbsp;5 Builder does not provide ZIP output yet &ndash; but it will in the future. In the meantime, you can use one of the generic ZIP packages provided [on the download page](https://ckeditor.com/ckeditor-5/download/#zip).
+</info-box>
```
Mgsy, line 43: Actually, it's not true anymore. You can download a zip from the builder, so we should mention that option here.

image

godai78: We may simply remove this infobox, Builder is linked on top. I'll fix these.

Mgsy (approved):

PR #18464, opened 2025-04-30
Introduce `doNotFixSelection` for `deleteContent`

### LINKED ISSUE #18448: Add option to force `model.deleteContent` to remove exactly the given range

Current, `model.deleteContent()` tries to fix the range-to-remove to provide better UX. If the range-to-delete is as follows:

```<p>[Foo</p><p>]Bar</p>```

Then `deleteContent()` will act like the range ends in the first paragraph:

```<p>[Foo]</p><p>Bar</p>```

Which is a special handling for "triple click" behavior, where a lot of times user would not realize what selection they made and that they are about to merge the following paragraph (with "Bar").

However, in some cases, you want to remove exactly the range that you passed. One such case is track changes integration, where e.g. the insertion suggestion marker may be like in the first example, and upon discard you want to revert the whole insertion, so you don't want to shrink the range-to-delete. Right now, in track changes mode, if you write something, press enter and discard the suggestion, you will be incorrectly left with an empty paragraph.

scofalik:

Other (engine): Introduced `doNotFixSelection` option for `deleteContent()` util which can be used to force making deletion exactly on the provided selection. Closes #18448.

### DIFF
--- a/packages/ckeditor5-engine/src/model/utils/deletecontent.ts
+++ b/packages/ckeditor5-engine/src/model/utils/deletecontent.ts
@@ -61,6 +61,18 @@ import type Writer from '../writer.js';
  * **Note:** If there is no valid position for the selection, the paragraph will always be created:
  *
  * `[<imageBlock src="foo.jpg"></imageBlock>]` -> `<paragraph>[]</paragraph>`.
+ *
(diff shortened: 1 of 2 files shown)


scofalik:
I am not sure about the flag name, it could be more specific too. I am also not sure about the commit message 😄. But I assume that anyone who's interested will read the API docs where details are explained.

niegowski (approved):

PR #18478, opened 2025-05-06
Improve behavior of empty sidebars in fullscreen mode

### LINKED ISSUE #18474: Smarter handling of empty sidebars in fullscreen mode

Right now, both sidebars have a fixed width regardless their content. That ensures the layout is always the same, e.g. editable is in the screen center. However in some cases it is troublesome, e.g. for narrower displays the right sidebar could be hidden if it doesn't contain any UI elements.

This could be partially solved by setting the width on particular sidebar items rather than the containers, but it fails when you only have one empty sidebar (the editable is no longer in the middle then), so it requires a bit more sophisticated logic.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

---

Dumluregn:

Feature (fullscreen): Empty sidebars will no longer lock empty space around the editable in fullscreen mode. Closes #18474.

---

### Additional information

So, the main changes are:
* No longer prevent sidebars from shrinking (removed `flex-shrink: 0`).
* To keep their fixed width, added `min-width` on the first child of each sidebar. So if they are empty, it won't be applied.
* Sticky part of the left sidebar will only be created if there is a document outline or presence list.

Some styles were just moved around for better organization, unfortunately it creates a bit of fuzz in the PR, sorry.


DawidKossowski (commented):
Just a one typo.

DawidKossowski (approved):

Issue #18485, opened 2025-05-07
Annotations in Fullscreen mode disappear if an AnnotationUI is activated


lkszzajac:
## 📝 Provide detailed reproduction steps (if any)

1. Open an editor with Track Changes/Comments and Fullscreen plugins
2. Add a few annotations (both comments and suggestions)
3. Open Fullscreen mode
4. Activate an AnnotationUI (e.g. annotation filter: `editor.plugins.get( 'AnnotationsUIs' ).activate( 'wideSidebar', annotation => annotation.type.includes( 'suggestion' ) );`

### ✔️ Expected result

Annotation UI is activated with annotations filtered according to the provided callback function.

### ❌ Actual result

All annotations disappear from the sidebar.

## 📃 Other details

* Browser: n/a
* OS: n/a
* First affected CKEditor version: 45.0.0
* Installed CKEditor plugins: Track Changes, Comments, Fullscreen

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Dumluregn:
Done in https://github.com/ckeditor/ckeditor5/pull/18489.

PR #18489, opened 2025-05-08
Fix setting the filters on annotations while in fullscreen mode

Dumluregn:

Fix (fullscreen): Allow for setting the custom filters on annotations while in fullscreen mode. Closes https://github.com/ckeditor/ckeditor5/issues/18485.

---

### Additional information

Diff looks worse than it actually is, most of the common logic was moved to `switchToWideSidebar()` function.

There are two scenarios to cover: wide sidebar is already in use (so it will be configured in Sidebar plugin and the element already exists) or not (Sidebar doesn't have a container set and the sidebar element is not created yet). Depending on the case, we need to perform the same actions in different order - it's explained in the inline comments.

There is a separate PR with tests in the comments package.

### DIFF
--- a/packages/ckeditor5-fullscreen/src/handlers/abstracteditorhandler.ts
+++ b/packages/ckeditor5-fullscreen/src/handlers/abstracteditorhandler.ts
@@ -506,7 +506,7 @@ export default class AbstractEditorHandler {
 	}
 
 	/**
-	 * Stores the current state of the annotations UIs to restore it when leaving fullscreen mode.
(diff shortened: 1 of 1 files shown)


oleq (approved):
LGTM

On a side-note (unrelated to this PR), I think there could be a memory leak in this abstract editor handler. We don't `this._annotationsUIsData = null;` when the editor gets destroyed and that could lead to non-GCed data referenced in the command

Screenshot 2025-05-09 at 13 59 00

I didn't check it with the memory inspector, though. It could be a false alarm.

PR #7297, opened 2020-05-26
Added tests to ensure that upcasting works with non-html DPs (markdown)

fredck:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Tests (code-block): Added tests to ensure that the code-block plugin works with markdown. Follow up to #5977.

---

### Additional information

This PR transformed over time. It originally solved #5977 but it seems to be not necessary any more since v20.0.0.

All summed, this PR is reduced to adding a test only to ensure no #5977 regression.

Original PR: ckeditor/ckeditor5-code-block#6.

PR #18501, opened 2025-05-09
Fix manual tests language.ui value

l-lejman:

Tests: Fix manual tests language.ui value.

---

### Additional information

Closes https://github.com/cksource/ckeditor5-commercial/issues/7297.


packages/ckeditor5-utils/src/config.ts
```diff
@@ -193,9 +193,9 @@ export default class Config<Cfg> {
 
 		// Iterate over parts to check if currently stored configuration has proper structure.
 		for ( const part of parts ) {
-			// If there is no object for specified part then create one.
-			if ( !isPlainObject( target[ part ] ) ) {
-				target[ part ] = Object.create( null );
+			// If target[part] is not an object or array, initialize as empty object.
+			if ( typeof target[ part ] !== 'object' || target[ part ] === null ) {
+				target[ part ] = {};
```
niegowski, line 198: Why such a change? This was changed recently to avoid prototype pollution.

packages/ckeditor5-utils/src/config.ts
```diff
@@ -206,7 +206,7 @@ export default class Config<Cfg> {
 		if ( isPlainObject( value ) ) {
 			// We take care of proper config structure.
 			if ( !isPlainObject( target[ name ] ) ) {
-				target[ name ] = Object.create( null );
+				target[ name ] = {};
```
niegowski, line 209: Why such a change? This was changed recently to avoid prototype pollution.

niegowski (changes requested):

packages/ckeditor5-core/src/context.ts
```diff
@@ -160,11 +162,36 @@ export default class Context {
...
-		} );
+		if ( !translations && global.window.CKEDITOR_TRANSLATIONS ) {
+			/**
+			 * Function _translate from translation-service.ts gets translations from
+			 * global.window.CKEDITOR_TRANSLATIONS when translations injected into Locale are empty.
+			 * Value of global.window.CKEDITOR_TRANSLATIONS is provided by CKEditorTranslationsPlugin from
+			 * ckeditor5-dev-translations package (for example in manual tests) when --language flag is used.
+			 * Function _translate is called often and has no access to the editing editor config, so here is a better place to
+			 * check if translations will be taken from dev-translations, and if yes – update config.language.ui value.
+			 */
```
Mati365, line 173: imho, this comment could be clearer. Two things:

1. It sounds like `global.window.CKEDITOR_TRANSLATIONS` is _always_ set by `CKEditorTranslationsPlugin`, but that’s not true. It can also come from translation files loaded from a CDN, so we shouldn’t assume it always comes from the plugin.
2. Without reading the issue ([#7297](https://github.com/cksource/ckeditor5-commercial/issues/7297)), it’s hard to tell what the problem is and why this function is needed. The comment should explain it in a more straightforward way.

Mati365 (changes requested):

packages/ckeditor5-utils/src/config.ts
Mati365: I agree with @niegowski comment. This part of the code is within the mitigation area for prototype pollution, which is also flagged by code scanners. The change is fully intentional. Bypassing security measures just to apply a bugfix isn’t a good practice. It’s better to fix this on the side of the library that relies on this behavior.

l-lejman:
i added pr to wiris to resolve mathtype manual error https://github.com/wiris/html-integrations/pull/1107

PR #18513, opened 2025-05-13
[WiP]: Add `data-list-item-id` attribute on `<li>` elements in data downcast

### LINKED ISSUE #18407: `listItemId` model attribute is not represented in the editor output data

**Introduction**

We represent list items in the document model by adding a `listItemId` to "regular" model elements like paragraphs, headings or tables. We also have a bunch of other list-related attributes.

`listItemId` is used to group all elements that are grouped under the same list item. For example, a list item could have two paragraphs inside:

```html
...
<paragraph listItemId="abc123" ...>Foo</paragraph>
<paragraph listItemId="abc123" ...>Bar</paragraph>
...
```

These ids currently are **not** represented in the editor output data:

```html
...
<ul>
    <li>
        <p>Foo</p>
        <p>Bar</p>
    </li>
</ul>
...
```

Note, that if you have just one paragraph without any styling or markers, the `<p>` is omitted:

```html
...
<ul>
    <li>Foo</li>
    ...
</ul>
...
```

The ids are generated on-the-fly each time the data is upcasted. This leads to a situation, where the same HTML data would generate different model states. It contradicts our rule and a general guidance, that the model and output data should stay stable between multiple data loads and saves if the data itself didn't change.

I checked it, and it *looks like* it does not generate problems related to compression mechanism.

However, there are still some problems to consider:

1. Diffing algorithms will report these "unstable IDs" as changes. We can't filter them though, as sometimes it makes sense that the `listItemId` changed.
    a. Consider also external systems where our data could be post-processed?
2. It creates a constant headache for future. When designing complex systems and algorithms like we have with CKE5, we should try to do everything to stick to some fundamental rules, so we can rely on them and build stable software.

---

**Possible solutions**

There are two ways we can solve this problem that I see:

1. Output `listItemId` into the final data.
2. Get rid of `listItemId`, instead use something like `listNewItem=true|false` which would inform whether given element is part of the same list item as the element above (forced true for first item of the list of course).

I admit, I prefer the second option. The IDs seem to me like an unnecessary complication. It seems that all we need to know is whether we should "start a new list item" or not. It does not pollute HTML output. It is possibly a bit easier to process for external mechanisms (like e.g. LLMs). It possibly leads to less code in the editor (e.g. we don't need to check for IDs duplicates). It's simply more elegant.

One consideration is if all behaviors can stay the same. As far as I imagine, only actual move operations in the model may lead to different result. These are very rare, though. From top of my head, only wrapping (i.e. blockquotes) use them. Even copy/paste and drag-and-drop use downcast->upcast to handle the data.

I think it is safe and better, however, it would lead to a bigger refactor, which means it is more time consuming.

A simpler and faster solution is to simply output IDs to the final HTML. However, there are some considerations here as well.

**How this relates to "unique block IDs", headings, and structure editing? What attribute to use and on which element?**

First, we need to consider two similar functionalities:

* headings IDs (existing functionality),
* unique elements IDs (upcoming functionality),

Headings already have their own `headingId` attribute which is downcasted to `id=""` attribute in HTML. We want these IDs to be used as anchors for links, we also use them ourselves in document outline and table of contents features. Every heading ID must be unique

We don't have yet a mechanism that would set a unique ID for every element in the model, but this is something that is inevitably coming to the editor. Similarly to heading IDs, every ID must be unique. Actually, this would be expanding the heading ID functionality on all elements (but probably implemented as `model.Element` property, not a regular attribute).

It's up to be decided, whether unique block elements IDs will be set in HTML as `id=""` or a custom attribute. In my opinion, we could use `id=""` for all elements, if we want to allow linking to any element, but there are some things to consider here as well, so I think this is an open question.

However, how these two subjects relate to `listItemId`s?

They are two separate things. Note, that `listItemId` does not identify a certain element in the model. Neither is this value unique among model elements, as multiple elements in the same list item have the same value. Finally, it is not exclusive with unique element ID -- elements inside lists should have both, and one cannot replace the other.

When it comes to representing `listItemId` in output data, I believe we should use something like `data-list-item-id=""`.

Another question is where to set the attribute. We can choose between `<li>` that groups the elements, and the elements themselves. Putting it on `<li>` would be more beneficial for three reasons:

* it simply makes more sense, it is the ID for list item, it should be set on the list item element,
* it makes less mess in HTML (it's repeated only once),
* it does not force us to output `<p>` in `<li>` if not needed (right now if there is only one paragraph in list item and it does not have styles/markers, we skip it in the output).

In relation to headings IDs / unique elements IDs, there is an idea, that if `id=""` is set on `<li>` element, it could be upcasted to `listItemId` then. It's interesting but I'd rather be explicit, and not go into solutions like that. These "smart" things always leads to confusion, hidden knowledge, more things to explain when introducing people to the features, etc. Finally, we aren't even sure right now if unique element IDs will be represented as `id=""` in output HTML; or if we won't actually need `id=""` on `<li>` for a different purpose.

**Mess in HTML.**

Some people may dislike additional "unnecessary" IDs. We may consider give them a way to output "clean HTML". It will be a bit more problematic if we introduce "unique blocks IDs", as it will add a lot of ids in the output HTML.

I'd consider introducing an option for `getData( { omitIds: true } )` or something more semantic like `getData( { cleanData: true } )` to be able to return HTML for "display" purposes only. We would need to underline in docs that this data should not be used to initialize editor, only to display the data (e.g. as an article on the website). `cleanData` could also drop marker related tags, like `<comment-start />`.

**Loading data without list items IDs.**

Note, that we will still need to handle HTML without IDs for handling external and legacy data (both during editor initialization and while pasting). Adding IDs to list items during upcast will need to stay.

The question one may have is: if we need support to load HTML without IDs anyway, then why `getData( { cleanData: true } )` should be used for "display purposes only"? Like, we need to handle loading it anyway, right?

The answer here is: we should have only one "normalized" "correct" document model format. We should expect that legacy data will gradually get converted into that format. Yes, we will handle HTML without IDs, but only in two scenarios: pasting and loading data once. This makes maintenance easier, as well as any future reasoning. We can be sure that after initial load, the editor will save "correct" data, and also any in-editor feature will have "correct" data available. Any data post-processed from the editor data will be correct too. Strict rules and clear processes is the only way to maintain sanity in such a big project.

In a perfect world, I'd separate the action of "creating a new document" from "loading the document", as these are two different actions, for which we may have different requirements/limitations. For a given document, you create it only once, and you can create it from any data (also non-supported HTML). OTOH, you load given document multiple times, but then you can be sure that it has correct format and you can store additional metadata with it.

**Fixing corrupted data (duplicated IDs).**

In general, we have multiple places in the codebase where we need to generate unique IDs (e.g. comments IDs). We have an utility function for this and, well *we hope and believe it works well*.

However, you can end up with duplicated IDs in copy/paste scenarios or after incorrectly post-processing the data. So, we will need a mechanism for that.

---

**Summary**

* The `listItemId` model attribute is not represented in output data, which leads to some problems.
* It seems that it would be better to change list feature logic not to use/compare IDs, instead we should have an attribute like `listItemNew=true|false` which would tell you if given element is also a part of the previous list item. This means possibly bigger refactor.
* Alternative is to output `listItemId` to HTML data. It's simpler but we pollute HTML and we would need to make some decisions.
* `listItemId` is a completely separate topic from headings IDs or (upcoming) unique element IDs.

Dumluregn:

Feature (list): Add `data-list-item-id` attribute on `<li>` elements in data downcast. Closes #18407.

---

### Additional information


packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -78,7 +78,7 @@ export function listItemUpcastConverter(): GetCallback<UpcastElementEvent> {
 			return;
 		}
 
-		const listItemId = ListItemUid.next();
+		const listItemId = data.viewItem.getAttribute( 'data-list-item-id' ) || ListItemUid.next();
```
niegowski, line 81: Please ensure this attribute is consumed so it does not leak to GHS.

scofalik (changes requested):
I don't see anything else other than what Kuba already mentioned.

packages/ckeditor5-list/docs/features/lists.md
```diff
@@ -133,6 +133,38 @@ ClassicEditor
 	The {@link module:list/listproperties~ListProperties} feature overrides UI button implementations from the {@link module:list/list/listui~ListUI}.
 </info-box>
 
+### Editor output configuration
+
+Starting with version 45.1.1, list items (`<li>` elements) in the editor data output contain an additional attribute called `data-list-item-id`. It's necessary to ensure better data comparisons within some editor features, like revision history.
```
scofalik:
```suggestion
Starting with version 45.1.1, list items (`<li>` elements) in the editor data output contain an additional attribute `data-list-item-id`. The attribute is necessary to ensure that the lists feature work correctly with other editor features and mechanisms.
```

packages/ckeditor5-list/docs/features/lists.md
```diff
@@ -133,6 +133,38 @@ ClassicEditor
...
+
+```html
+<ul>
+	<li data-list-item-id="e72808ee4144975064acb5d66e5cfba13">Hello</li>
+	<li data-list-item-id="e5d719ab356409767e9d4358485476358">There!</li>
+</ul>
+```
+
+If data loaded to the editor does not contain the `data-list-item-id` attribute on list items, it will be automatically added when data is saved.
```
scofalik:
```suggestion
If data loaded in the editor does not contain the `data-list-item-id` attributes on list items, they will be automatically added when data is saved.
```

packages/ckeditor5-list/docs/features/lists.md
```diff
@@ -133,6 +133,38 @@ ClassicEditor
...
+
+If you need to save the editor contents without this attribute, you can set the `skipListItemIds` flag to `true` when calling `editor.getData()` method:
```
scofalik:
```suggestion
If you wish to save the editor contents without this attribute, you can set the `skipListItemIds` flag to `true` when calling `editor.getData()` method. **However, please note, that data without IDs should be used only for presentation purposes (e.g. displaying the document as HTML), and should not be used to initialize the editor.**
```

packages/ckeditor5-list/docs/features/lists.md
```diff
@@ -133,6 +133,38 @@ ClassicEditor
...
+```html
+<ul>
+	<li>Hello</li>
+	<li>There!</li>
+</ul>
+```
+
+<info-box>
+	It is not recommended to use data without `data-list-item-id` attribute in the long run as some features may work incorrectly.
+</info-box>
```
scofalik: I wouldn't even suggest that it is okay to use this `data-list-item-id` as "editor data".

I proposed a different message in comment above. We could have it in a box as well (warning box?), but I'd like to have this warning box directly after the paragraph that mentions `skipListItemIds` so there's a bigger chance that people will read it.

Dumluregn: I think putting it in bold as you proposed will be enough.

Issue #18475, opened 2025-05-06
Switch to ESLint 9


pomek:
## Provide a description of the task

_What steps should be taken to fulfill the task?_

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …
* Installed CKEditor plugins: …

PR #18520, opened 2025-05-14
Internal: Bump required Node version to `>=20.12.0`.

filipsobol:

Internal: Bump required Node version to `>=20.12.0`.

---

We already use `util.styleText` which requires Node 20.12.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -167,7 +167,7 @@
     "webpack": "^5.94.0"
   },
   "engines": {
-    "node": ">=20.0.0"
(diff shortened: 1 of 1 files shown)


pomek:
While working on #18475 (ESLint migration), we must bump the `engines.node` version to match the range: `^18.18.0 || ^20.9.0 || >=21.1.0`.

This PR overlaps with #18499, so I would like to drop it in favor of #18512.

PR #18536, opened 2025-05-19
Update optimizing build size doc

filipsobol:

Docs: Update `Optimizing build size` document.

---

Following the recent tree-shaking optimizations, importing the editor and its features from individual packages offers marginal benefits, even in performance-critical projects. For this reason, I propose deleting this section and only showing how to optimize translations and styles, as this provides more meaningful results.

I updated the “The result” section to show only gzipped numbers so that the numbers are less confusing when scanning the document.

### DIFF
--- a/docs/getting-started/setup/optimizing-build-size.md
+++ b/docs/getting-started/setup/optimizing-build-size.md
@@ -14,40 +14,11 @@ The build size optimization is only possible when using the npm build and module
 
 ## How to optimize the build size
 
-To optimize the build size, you only need to make a few changes to the way you import the editor, styles, and optionally the translations, compared to what was shown in the {@link getting-started/integrations-cdn/quick-start Quick start} guide. You do not need to change anything in the editor configuration or in the way you use the editor.
-
-### Code imports
-
-The first step in optimizing build size is to import only the editor features you need, as adding more plugins to the editor configuration will increase build size.
-
-The next step is to change the way you import the editor features. Currently, you are probably importing them from the following packages:
-
-```js
-import { /* ... */ } from 'ckeditor5';
-import { /* ... */ } from 'ckeditor5-premium-features';
-```
-
-These two packages export all the editor features, and most are tree-shakeable. However, some code may be added to the build even if it is not used. To ensure that the unused code is not imported, you can import the editor features directly from the packages that contain them.
-
-For example, if you are using the classic editor type with the bold, italic, and table features, you can change the imports like this:
-
-```diff
-- import { ClassicEditor, Bold, Italic and Table } from 'ckeditor5';
-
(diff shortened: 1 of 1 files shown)


docs/getting-started/setup/optimizing-build-size.md
```diff
@@ -85,7 +56,7 @@ import '@ckeditor/ckeditor5-table/dist/index.css';
 // ...
 ```
 
-Regarding plugin styles, the rule of thumb is to import the styles from all the individual packages from which you import the editor features. For example, if you import the bold feature from `@ckeditor/ckeditor5-basic-styles/dist/index.js`, you should also import `@ckeditor/ckeditor5-basic-styles/dist/index.css`.
+The rule of thumb is to import styles from all the individual packages from which you import the editor features. In your IDE you likely can <kbd>Ctrl</kbd>+<kbd>Left click</kbd> the imported feature name to see in which pages they live. For example, if you click the `Bold` feature, it will lead you to the `@ckeditor/ckeditor5-basic-styles` package in `node_modules`. This means that should also import the styles from the `@ckeditor/ckeditor5-basic-styles/dist/index.css` file. Make sure to not to miss any of the plugins or import the same styles multiple times.
```
godai78, line 59: In which packages? Cause "pages" are a bit unclear.

godai78 (dismissed):
LGTM

gorzelinski (approved):

docs/getting-started/setup/optimizing-build-size.md
filipsobol: Yeah, it should be `package`

Witoso (approved):

godai78 (dismissed):

docs/getting-started/setup/optimizing-build-size.md
```diff
@@ -14,40 +14,15 @@ The build size optimization is only possible when using the npm build and module
...
-
-+ import { ClassicEditor } from '@ckeditor/ckeditor5-editor-classic/dist/index.js';
-+ import { Bold, Italic } from '@ckeditor/ckeditor5-basic-styles/dist/index.js';
-+ import { Table } from '@ckeditor/ckeditor5-table/dist/index.js';
-```
-
-<info-box warning>
-	Note the `/dist/index.js` part of the import paths. This is important to ensure that the editor functions are imported from the correct files.
+<info-box>
+	Before version 45.0.0, we recommended importing the editor and its features using the `@ckeditor/ckeditor5-PACKAGE_NAME/dist/index.js` import paths. However, this is no longer necessary. As of version 45.0.0, the code has been improved to allow for better tree-shaking. Now, you can import all the code directly from the `ckeditor5` and `ckeditor5-premium-features` packages.
```
godai78, line 20: We may consider linking the update guide here, but I don't find it necessary.

godai78 (approved):

PR #18576, opened 2025-05-27
Added upcast converters for legacy HTML table attributes.

### LINKED ISSUE #18575: Table and table cell properties should handle legacy visual html attributes

The following attributes are not upcasted and normalized to common model attributes:
* `<table>`
  * `width`
  * `height`
  * `bgcolor`
* `<td>`/`<th>`
  * `width`
  * `height`
  * `bgcolor`

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

---

niegowski:

Other (table): The legacy HTML table attributes are now normalized to editor features (`width`, `height`, `bgcolor`). Closes #18575.

Other (engine): The `atributeToAtribute` and `elementToAtribute` upcast helpers should not consume anything if conversion callback returns `undefined` value. See #18575.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
+++ b/packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
@@ -332,7 +332,6 @@ export default class UpcastHelpers extends ConversionHelpers<UpcastDispatcher> {
 			key: string;
 			value: unknown |
 				( ( viewElement: ViewElement, conversionApi: UpcastConversionApi, data: UpcastConversionData<ViewElement> ) => unknown );
-			name?: string;
 		};
 		converterPriority?: PriorityString;
 	} ): this {
@@ -1072,7 +1071,7 @@ function prepareToAttributeConverter(
 			config.model.value( data.viewItem, conversionApi, data ) : config.model.value;
 
 		// Do not convert if attribute building function returned falsy value.
-		if ( modelValue === null ) {
+		if ( modelValue === null || modelValue === undefined ) {
 			return;
 		}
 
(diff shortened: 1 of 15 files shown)


packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
```diff
@@ -255,14 +267,17 @@ function enableAlignmentProperty( schema: Schema, conversion: Conversion, defaul
 		// Support for the `align` attribute as the backward compatibility while pasting from other sources.
 		.attributeToAttribute( {
 			view: {
+				name: 'table',
```
niegowski, line 270: This was provided on the model side, but that notation was not supported, and as a result, this converter was triggered on any element with the given attribute.

packages/ckeditor5-engine/src/conversion/upcasthelpers.ts
```diff
@@ -1072,7 +1071,7 @@ function prepareToAttributeConverter(
 			config.model.value( data.viewItem, conversionApi, data ) : config.model.value;
 
 		// Do not convert if attribute building function returned falsy value.
-		if ( modelValue === null ) {
+		if ( modelValue === null || modelValue === undefined ) {
```
niegowski, line 1074: Without the check for `undefined`, the following code attempted to apply a model attribute with an `undefined` value (with no effect, as the model writer ignores changes from `undefined` to `undefined`). It consumed it without applying anything to the model. The issue is that we often use conversion helpers simply by putting `return` without any value, and TypeScript is not helping here either.

mmotyczynska (approved):

PR #18594, opened 2025-05-28
Increase specificity of `figure.table` fixes due to print mode issues.

Mati365:

Fix (table): Improve calculation of pagination page-breaks on documents with long tables. Closes https://github.com/ckeditor/ckeditor5/issues/18600

---

### Additional information

Fully reproducible using instructions from https://github.com/cksource/ckeditor5-commercial/issues/5494

Issue https://github.com/cksource/ckeditor5-commercial/issues/7850
Commercial PR https://github.com/cksource/ckeditor5-commercial/pull/7857
See more https://github.com/cksource/ckeditor5-commercial/pull/7857

### DIFF
--- a/packages/ckeditor5-table/theme/table.css
+++ b/packages/ckeditor5-table/theme/table.css
@@ -81,8 +81,6 @@
 		 * 	* The table is inside a figure element.
 		 * 	* The table has `display: table` style set.
 		 * 	* The block element is placed directly after the table.
-		 *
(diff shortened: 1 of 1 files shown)


pszczesniak (approved):
:+1:

Just thinking - should it be `Internal` 🤔  - maybe a `Fix` is a better option; to mention about that fix in a changelog. If yes, we should create an issue in CKE5 repo.

PR #18596, opened 2025-05-28
Updated eslint config

przemyslaw-zan:

Internal: Updated eslint config.

---

### Additional information


pomek:
To deny what I wrote in: https://github.com/cksource/ckeditor5-internal/issues/4018#issuecomment-2916928426, I am thinking about restricting imports in the whole package, instead of sources/tests only. So, even snippets cannot use Commercial features when their sources are in the open-source package.

Does it make sense?

pomek:
This way we could unify the environment. Developers have to follow a single rule only: no commercial imports.

pomek:
> A problem: How do we determine the commercial packages? There are two cases:

We could define a list of packages that are not features, but integrations, like Inspector, React, etc.

psmyrek:
> So, LGTM. Did I miss anything?

In some cases the JavaScript version of a rule works as expected for TypeScript sources. But since the `no-restricted-imports` is marked as an extension rule explicitely by the `@typescript-eslint` team, I guess it makes sense to follow their advice described in https://typescript-eslint.io/rules/#extension-rules.

pomek (approved):
LGTM.

PR #18645, opened 2025-06-06
Migrate `markdown-gfm` to `remark` and `rehype`

### LINKED ISSUE #18684: Migrate `markdown-gfm` package to `remark` and `rehype`
Currently, the `markdown-gfm` package uses `marked` and `turndown` to turn HTML to Markdown and vice versa. The issue is that these packages are maintained independently and sometimes yield different results. Using `remark` and `rehype` packages from the same and much richer ecosystem could solve that.

We also observed that some plugins, while stable, are not actively developed (e.g. `turndown-plugin-gfm`), which adds additional motivation to update the ecosystem.

filipsobol:

Other (markdown-gfm): Migrate to remark / rehype packages. Closes #18684.

MINOR BREAKING CHANGE (markdown-gfm): Migrate from `marked` and `turndown` to `remark` and `rehype`.

MINOR BREAKING CHANGE (markdown-gfm): Enable autolinking in Markdown (works only when loading Markdown content into the editor).

---

#### Things to prioritize when testing this PR

The following areas should be prioritized when testing this PR, as they previously had (or have now) custom handling:

* Lists – both those loaded into the editor and those created within the editor (especially to-do lists).
* Links – regular links and plain text links like `www.example.com` that are converted into links when loaded into the editor.
* Code blocks
* HTML embedded within the Markdown
* The `keepHtml` function (which appears to lack documentation aside from the [API reference](https://ckeditor.com/docs/ckeditor5/latest/api/module_markdown-gfm_gfmdataprocessor-GFMDataProcessor.html#function-keepHtml))
* Loading Markdown generated by the older version of the integration into the new one

---

#### A few thoughts on the migration

The `remark` and `rehype` combo works more predictably – likely because they come from the same ecosystem – compared to `marked` and `turndown`, which are developed independently. This is also reflected in the tests, which needed to be updated to pass.

One benefit of the older setup was its smaller size. Migrating from `marked` and `turndown` to `remark` and `rehype` **increases the build size by 40 KiB gzipped** (140.51 KiB uncompressed).

The `unifiedjs` ecosystem (which includes `remark` and `rehype`) is also much more modular, while `turndown` and `marked` don't have any external dependencies. **This increases the total number of dependencies from 10 to 97.** The number of direct dependencies in markdown's `package.json` increased from 5 to 14. However, the following plugins are foundational and are already dependencies of the larger ones:

* `hast`
* `hast-util-from-dom`
* `hast-util-to-html`
* `hast-util-to-mdast`
* `unist-util-visit`

#### Challenges

There were two major challenges when implementing this change.

##### Autolinking

The first was **autolinking**. The original implementation had autolinking (automatically turning text links like `http://example.com` into `<http://example.com>`) disabled. While this wasn't compliant with the GFM spec, there may have been reasons for it. However, we didn’t treat text links like regular text either, because we explicitly disabled escaping of them. This meant they were neither a link nor plain text.

As a result, Markdown returned by the editor could include unescaped text links, which would later be turned into links by other Markdown parsers. We decided to enable autolinking so text links are turned into regular links when loaded in the editor. Text links typed directly into the editor will not be turned into links, but instead escaped like regular text –providing a better WYSIWYG experience.

```markdown
// Markdown => HTML
[www.example.com](www.example.com)              => <a href="www.example.com">www.example.com</a>
www.example.com                                 => <a href="www.example.com">www.example.com</a>

// HTML => Markdown
<a href="www.example.com">www.example.com</a>   => [www.example.com](www.example.com)
www.example.com                                 => www\.example.com
```

##### Bundle size

Another issue we anticipated was bundle size. Typically, Markdown and HTML parsing happens on the server, so most packages are built with that environment in mind. However, unlike the browser, server environments don't have built-in DOM parsers and instead use packages like `parse5`. While `parse5` is excellent, it's completely unnecessary in a browser environment, which already has a built-in DOM parser.

The `rehype-dom-parse` and `rehype-dom-stringify` packages are DOM-based alternatives to `rehype-parse` and `rehype-stringify`, and they resolved most of the related issues. One issue they didn’t solve was parsing HTML **inside** Markdown. Unfortunately, there doesn’t appear to be a first-party solution for this, which necessitates writing a small custom plugin.

You can find more information here: https://github.com/orgs/rehypejs/discussions/202. While this discussion hadn’t received an answer at the time of writing, it provides useful context for the issue, so I’m leaving the link for “future us”.


filipsobol:
~~PR is still WIP, I just need CI to run the tests~~

EDIT: PR is ready for review.

packages/ckeditor5-markdown-gfm/src/markdown2html/markdown2html.ts
```diff
@@ -7,51 +7,56 @@
...
-		marked.use( {
-			tokenizer: {
-				// Disable the autolink rule in the lexer.
-				autolink: () => null as any,
-				url: () => null as any
-			},
-			renderer: {
-				checkbox( ...args: Array<any> ) {
-					// Remove bogus space after <input type="checkbox"> because it would be preserved
-					// by DomConverter as it's next to an inline object.
```
filipsobol, line 37: To double check

packages/ckeditor5-markdown-gfm/tests/_utils/utils.js
```diff
@@ -27,17 +27,17 @@ export function testDataProcessor( markdown, viewString, normalizedMarkdown, opt
...
 		options.setup( dataProcessor );
 	}
-	const viewFragment = dataProcessor.toView( markdown );
 
+	const viewFragment = dataProcessor.toView( markdown );
 	const html = cleanHtml( _stringifyView( viewFragment ) );
 
 	// Check if view has correct data.
-	expect( html ).to.equal( viewString );
+	expect( JSON.stringify( html ) ).to.equal( JSON.stringify( viewString ) );
```
filipsobol, line 35: This improves the readability when the output is different from the expected value.

packages/ckeditor5-markdown-gfm/tests/gfmdataprocessor/code.js
```diff
@@ -23,18 +23,6 @@ describe( 'MarkdownGfmDataProcessor', () => {
...
 
-		it( 'should process spaces inside inline code', () => {
-			testDataProcessor(
-				'regular text and` inline code`',
-
-				'<p>regular text and<code> inline code</code></p>',
-
-				// When converting back it will be normalized and spaces
-				// at the beginning of inline code will be removed.
-				'regular text and `inline code`'
```
filipsobol, line 34: I don't see a reason why this should be “normalized” like this. It seems invalid.

filipsobol:
> I assume that `LICENSE.md` file must be updated also.

Done :heavy_check_mark:

packages/ckeditor5-markdown-gfm/LICENSE.md
pomek, line 1: Could you update the root LICENSE.md file too? See: https://github.com/ckeditor/ckeditor5/blob/master/LICENSE.md.

pszczesniak (dismissed):
LGTM 👍

charlttsie:
### Markdown output doesn't contain nested to-do list items
**Steps**
1. Open [markdown.html](http://localhost:8125/ckeditor5-markdown-gfm/tests/manual/markdown.html)
2. Insert a to-do list with nested items

**Result**
Nested items are missing from the output:
Screenshot 2025-06-23 at 13 44 16

filipsobol:
> Markdown output doesn't contain nested to-do list items

Fixed

filipsobol:
> Found one case that does not work the same way as previously: adding a to-do list with code block.

Nice catch @psmyrek. This is an upstream bug that I reported [here](https://github.com/micromark/micromark-extension-gfm-task-list-item/issues/4). However, this doesn't work in the current implementation either, so I don't think it's a blocker. I added a [test](https://github.com/ckeditor/ckeditor5/pull/18645/commits/42ad7bef47916f0d0bbed53f52f4e85951c13f9c) with proper context in the comment and mentioned this case in a [list of known issues](https://github.com/ckeditor/ckeditor5/pull/18645/commits/d5db33fc99ee523995949977da21e242ab9a9203).

charlttsie:
### New lines in tables are missing from the output
**Steps**
1. Open [markdown.html](http://localhost:8125/ckeditor5-markdown-gfm/tests/manual/markdown.html)
2. Insert a table
3. Add some text inside a table cell
4. Press enter and add some text in a new line

**Result**
Newline is missing from the output:
Screenshot 2025-06-24 at 09 33 00

filipsobol:
> New lines in tables are missing from the output

Good catch. This didn't work in the old implementation too, so I don't think it's a blocker. We will discuss whether to fix this now or as a follow-up.

filipsobol:
PR is ready for technical re-review.

charlttsie:
@filipsobol We've finished testing the changes with @juliaflejterska and apart from reported issues the changes look good to us 👌

psmyrek (approved):

pszczesniak (approved):

pomek (approved):

psmyrek (dismissed):

pomek (approved):

PR #18690, opened 2025-06-12
Mapper should cache all view items it visits while it maps model position to view position

### LINKED ISSUE #18678: Mapper cache fails to invalidate cache leading to downcast errors

In some scenarios, `MapperCache` fails to remove cache when changes happen in the view. As a result, outdated cache is still stored and used, which leads to errors in model-to-view position mappings. This was reproduced in scenarios using lists (multi level lists, precisely) as a lot "happens" in view when list structure changes.

Below is the scenario that reproduces this error:

1. Open the editor with MultiLevelList and ListProperties plugins enabled, for example:
https://ckeditor.com/docs/ckeditor5/latest/examples/builds-custom/full-featured-editor.html
2. Add the multi-level list with 3 levels of nesting, like in the example below:

```
<ol class="multi-level-list legal-list" style="list-style-type:none;"><li><span class="multi-level-list__marker">1. </span>test<ol class="multi-level-list legal-list" style="list-style-type:none;"><li><span class="multi-level-list__marker">1.1. </span>test<ol class="multi-level-list legal-list" style="list-style-type:none;"><li><span class="multi-level-list__marker">1.1.1. </span>test</li></ol></li><li><span class="multi-level-list__marker">1.2. </span>test</li></ol></li><li><span class="multi-level-list__marker">2. </span>test</li></ol>
```

3. Place the cursor at the end of the last list item.
4.  (4.1) Press "Enter" to escape the list (4.2) or Change the last list item into a heading

### ✔️ Expected result

4.1. The list is escaped correctly and its structure remains unchanged.
or
4.2. The last list item is converted into a heading and the rest of the list structure remains unchanged.

### ❌ Actual result

The list structure breaks, leading to errors in console.

### Additional information

As debugged, this problem is related to mapper cache. Turning cache off makes the issue go away.

scofalik:

Fix (list-multi-level): Fixed editor crash which happened when editing deeply nested multi-level lists. Closes #18678.

Fix (engine): Fixed incorrect cache invalidation in `Mapper`, which may lead to crashes when editing heavily formatted content or when using complex features like multi-level lists. Closes #18678.

Internal (engine): `Mapper` should cache all view items it visits while it maps model position to view position.


scofalik:
I checked performance tests as mapper is doing now some more actions, but the additional overhead was from 0 to 4%, depending on the case.

PR #18720, opened 2025-06-16
List item bold upcast and downcast converters

### LINKED ISSUE #18714: Upcast and downcast converters for Bold in list items

- Use new format attribute: `listItemBold`.
- Downcast to class (not inline style) and use `::marker`.
- Place integration in list feature.
- The integration should be in a separate plugin (but if it’s not loaded do not load integration).
- There should be a glue-plugin that requires all integrations.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

mmotyczynska:

Internal (list): Added upcast and downcast converters for `listItemBold` attribute. Closes #18714.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-list/theme/listformatting.css
+++ b/packages/ckeditor5-list/theme/listformatting.css
@@ -12,3 +12,7 @@
 		font-style: initial;
 	}
 }
+
+.ck-content .ck-bold {
+	font-weight: bold;
+}
--- a/packages/ckeditor5-list/src/augmentation.ts
+++ b/packages/ckeditor5-list/src/augmentation.ts
@@ -28,6 +28,7 @@ import type {
 
 	ListFormatting,
 	ListItemFontFamilyIntegration,
+	ListItemBoldIntegration,
 
 	LegacyList,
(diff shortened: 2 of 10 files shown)


packages/ckeditor5-list/theme/listformatting.css
```diff
@@ -12,3 +12,7 @@
 		font-style: initial;
 	}
 }
+
+.ck-content .ck-bold {
+	font-weight: bold;
+}
```
niegowski, line 18: Maybe for bold (and italic) we could use `:marker` as we use classes for those, and it might not require style resetting inside a list.

niegowski: Ok, Safari does not support italic (for sure) and in theory bold, so let's do it as we planned.

niegowski: But after checking it, it supports those styles. My main concern is whether we would be able to properly reset styles for blocks inside a block quote, which we style as italic in our content styles.

niegowski (approved):

PR #18728, opened 2025-06-17
Fixed MapperCache cache invalidation when view item is inserted at a beginning of a non-tracked element

### LINKED ISSUE #18727: MapperCache incorrectly invalidates cache when item is inserted at the beginning of a non-tracked element

This is an internal issue in the `MapperCache`. It turned out that in a particular scenario, the cache is not invalidated properly. The scenario is:

```
<p>ab<span>^cd<em><b>e</b>f</em></span>gh</p>
```

If a view item is placed at `^`, `MapperCache` should invalidate all cache starting from `<span>`. Instead, cache starting after `<span>` is invalidated. This error is only reproducible if an item is placed at the beginning of a non-tracked element, as in such case we use the parent as the "anchor" for cache invalidation, and it was done incorrectly.

The problem could be reproduced in following scenario using comments feature:

1. Add some text to empty editor.
2. Apply font family on the whole text.
3. Apply comment on the whole text.
4. Apply additional comment somewhere in the middle of the text.
5. Click before the inside comment and then into the inside comment.

This resulted in editor crash, although the steps probably could be simplified, the error could be also reproduced with other features, as it is a general mapper issue.

scofalik:

Fix (engine): Fixed editor crash that happened in a specific scenario, when editing heavily formatted text, text with multiple comments, or text with comments and formatting. Closes #18727.

Internal (engine): Renames in `MapperCache` methods so names correctly reflect what the methods do.

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/mapper.ts
+++ b/packages/ckeditor5-engine/src/conversion/mapper.ts
@@ -870,7 +870,7 @@ export default class Mapper extends /* #__PURE__ */ EmitterMixin() {
  * * it is an internal tool, used by `Mapper`, transparent to the outside (no additional effort when developing a plugin or a converter),
  * * it stores all the necessary data internally, which makes it easier to disable or debug,
  * * it is optimized for initial downcast process (long insertions), which is crucial for editor init and data save,
- * * it does not save all possible positions for memory considerations, although it is a possible improvement, which may have increase
+ * * it does not save all possible positions mapping for memory considerations, although it is a possible improvement, which may increase
  *   performance, as well as simplify some parts of the `MapperCache` logic.
  *
  * @internal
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -1120,23 +1121,16 @@ export class MapperCache extends /* #__PURE__ */ EmitterMixin() {
 				this._clearCacheAll( viewParent );
 			} else {
 				// If this is not a tracked element, remove cache starting from before this element.
-				this._clearCacheStartingBefore( viewParent as ViewElement );
+				// Since it is not a tracked element, it has to have a parent.
+				this._clearCacheInsideParent( ( viewParent as ViewElement ).parent!, ( viewParent as ViewElement ).index! );
```
scofalik, line 1125: This PR includes some renames and changes in comments/API docs, but this is the place where the actual change/fix happens.

Instead of using `this._clearCacheStartingBefore( viewNode )` we use `this._clearCacheInsideParent( viewNode.parent, viewNode.index )`.

It turned out that `_clearCacheStartingBefore()` does not really do what it says. Instead, this method guarantees only that all cache after the node is cleared, so it's more like `_clearCacheAfter()`, and hence it was renamed too.

It actually does something in the middle, because it takes the cached mapping after the node, and clears from there including that mapping. For simple structures this is the same as "clear cache before", however for more complex structures, like from the example:

```
<p>ab<span>^cd<em><b>e</b>f</em></span>gh</p>
```

It didn't clear starting from before `<span>`. Instead it clears starting from "f" --- more or less.

Hence, it was incorrectly used here. Instead of using the method now renamed to `clearCacheAfter` we invoke `_clearCacheInsideParent` which correctly clears cache from given view position.

packages/ckeditor5-engine/src/conversion/mapper.ts
```diff
@@ -1155,53 +1150,54 @@ export class MapperCache extends /* #__PURE__ */ EmitterMixin() {
...
-			// In this case, there's nothing to do.
+			// In this case, there's nothing to do. We assume that there are no "holes" in caching in direct children of tracked element
+			// (that is if some children is cached, then its previous sibling is cached too, and we would not end up inside this `if`).
+			//
+			// TODO: Most probably this `if` could be removed altogether, after recent changes in Mapper.
+			// TODO: Now we cache all items one after another, so there aren't any "holes" anywhere, not only on top-level.
 			//
 			if ( !this._cachedMapping.has( viewParent ) ) {
-				this._clearCacheStartingBefore( viewParent );
+				this._clearCacheInsideParent( viewParent.parent!, viewParent.index! );
```
scofalik, line 1182: This is another place where the same fix was introduced, although given the latest changes in `Mapper` it seems that this code may not be reachable outside of artificial unit tests crafted for `MapperCache`. As in "TODO" above, it may be a dead code.

scofalik:
This fix leads to another +3-4% of performance overhead for certain scenarios (inline styles and tables).

Issue #11347, opened 2022-02-28
[Document list] List items are not removed correctly with backspace


FilipTokarski:
Part of #11377.

## 📝 Provide detailed reproduction steps (if any)

1.  Switch to `ck/10812-document-list-editing`
2.  Create a few list items
3.  In an empty list item press `backspace`

### ✔️ Expected result

Item gets removed, cursor goes up to the previous list item.

### ❌ Actual result

Cursor does not go to the previous list item. All other list items get spaced out, which looks bad. Only after second `backspace` they get back to normal.

### Screencast

https://user-images.githubusercontent.com/34380544/155957333-e7e6f6f8-15a3-40ad-ab3b-51ca38b44eaa.mp4

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

Reinmar:
That's how it works on Google Docs:

https://user-images.githubusercontent.com/156149/156347235-d3385a63-03f8-4046-a56d-25d3bc179ed9.mp4

Note, that I have paragraphs ("Normal text") style in Google Docs to have bigger space after them:

<img width="1038" alt="image" src="https://user-images.githubusercontent.com/156149/156347401-21b42592-a3fd-4e9f-87f1-f9870df879a8.png">

If you'd test it with the default styles that Google Docs come with (no spacing around paragraphs), it would look a bit differently:


https://user-images.githubusercontent.com/156149/156347822-bc8d6074-7375-4517-b38f-c5ad56c99f82.mp4

To sum up:

* The document changes as expected.
* We could change the styling a bit to make it less confusing.

Reinmar:
Goals:

*   Try to limit the visual noise when merging two paragraphs into one list item
*   But don't overdo it – let's not overcomplicate the stylesheets

oleq:
```css
.ck-content li > p {
    margin: 0;
}
```

for starters and let's see where do we go from here.

Reinmar:
Related ticket: #16866.

Reinmar:
Related (not functionally, but due to combined confusion that these issues create together): #16884.

PR #15079, opened 2023-09-27
Remove list items correctly with backspace.

pszczesniak:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/git-commit-message-convention.html))

Fix (list): Document list items should now be removed correctly using backspace. Closes #11347.

---

### Additional information

_For example – encountered issues, assumptions you had to make, other affected tickets, etc._

PR #18801, opened 2025-07-03
Improved rendering paragraph within the list and table cell element

### LINKED ISSUE #17440: Improve table cell alignment behavior while using Alignment feature

Currently, if we insert a table and write something in a cell, then apply text alignment using the Alignment feature from the main toolbar (not table cell properties), the text inside the cell gets wrapped in the paragraph, and then the alignment is applied.

That behavior might be confusing for users, as the paragraph adds margin, so the cell appearance changes after applying the alignment, which is not the case while using table cell properties. We can discuss whether these behaviors should be unified.

**Alignment using table cell properties**

Image

**Alignment using the alignment feature**

Image

---

pomek:
### 🚀 Summary

Fix (list, table): Improved rendering of the `<p>` element when it is an only child of a list item (`<li>`) or a table cell (`<td>` or `<th>`). Closes #17440.

Fix: (list, table): Improve the paragraphs spacing in lists and tables. Closes: #11347. 

MAJOR BREAKING CHANGE (list): Removed vertical spacing in list items by resetting margins for `<p>` elements that are the child of a `<li>` element.

MAJOR BREAKING CHANGE (table): Removed vertical spacing in table cells by collapsing margins of `<p>` elements that are the only child of a `<td>` or `<th>` element.

---

### 📌 Related issues

* Closes cksource/ckeditor5-commercial#7990.

---

### 💡 Additional information

This PR aligns the margin-reset rules for paragraphs inside <strong>table cells</strong> (<code inline="">td</code>, <code inline="">th</code>) and <strong>list items</strong>. The change prevents unexpected vertical spacing when editors insert or modify content at the beginning or end of a cell or list item

#### Motivation / Context

* Fixes unintended content shift when users press <strong>Enter</strong> at the top of a table cell or creates a multi-block list item.

* Ensures symmetrical behaviour between the first and last paragraphs in the table cells,

* Handles the presence of CKEditor’s internal elements (e.g. <code inline="">.ck-table-column-resizer</code>) by switching from <code inline="">:only-child</code> to <code inline="">:only-of-type</code>, so the rule is applied as long as the paragraph is the sole <strong>paragraph</strong>, even if other non-paragraph nodes exist.</p>

* We thought of `margin-block-start/end` usage, but it's not supported well in emails :/

### Table cells (td, th) - line height

**Before**

image

**After**

 image

### Lists (li) - line height

**Before**

 image

**After**

image

### Table cells (td, th) - editing

**Pay attention to the top/bottom of the cell**

**Before**

https://github.com/user-attachments/assets/a0d1c6ae-18a6-4681-9e84-5b4e6b4f106e

**After**

https://github.com/user-attachments/assets/fedc6e55-aa8b-4c7c-92a3-66a9ea9ce2ea

### Lists (li) - editing 

**Before**

https://github.com/user-attachments/assets/9676643d-3eee-493a-98ec-fc96ae38fe20

**After**

https://github.com/user-attachments/assets/0b5fa244-e756-4ee3-98b6-2e3c5880b44e

### DIFF
--- a/packages/ckeditor5-list/theme/listformatting.css
+++ b/packages/ckeditor5-list/theme/listformatting.css
@@ -10,24 +10,22 @@
 }
 
 .ck-content li {
-	/* Our opiniotanted styling in the lists, it prevents
-	 * the shift of content when a list is turned into a multi-block one
-	 * (enter + backspace scenario).
+	/* Opinionated list content styling: prevents content shift
+	 * when a list becomes multi-block (Enter + Backspace scenario).
 	 * See: https://github.com/ckeditor/ckeditor5/pull/18801
 	 */
 	& > p:first-of-type {
-		margin-block-start: 0;
+		margin-top: 0;
 	}
 
-	/* Rule which prevents the margins to appear when a
-	 * bogus paragraph gets a line height or text alignment.
-	 * We use only-of-type as the cell may have a div with column resize.
-	 * to provide more consistent experience with the first one.
+	/* Prevents margins from appearing when a bogus paragraph
+	 * receives line height or text alignment.
+	 * Ensures a consistent experience with the first paragraph.
 	 * See: https://github.com/ckeditor/ckeditor5/pull/18801
 	 */
 	& > p:only-child {
-		margin-block-start: 0;
-		margin-block-end: 0;
+		margin-top: 0;
+		margin-bottom: 0;
 	}
 
 	&.ck-list-marker-bold::marker {
--- a/packages/ckeditor5-table/theme/table.css
+++ b/packages/ckeditor5-table/theme/table.css
@@ -54,31 +54,25 @@
 
 				& > td,
 				& > th {
-					/* Our opiniotanted styling in the tables, it prevents
-					 * the shift of content when the enter is pressed at the top.
+					/* Opinionated table content styling: prevents content from shifting
+					 * when Enter is pressed in the first cell.
 					 * See: https://github.com/ckeditor/ckeditor5/pull/18801
 					 */
 					& > p:first-of-type {
-						margin-block-start: 0;
+						margin-top: 0;
 					}
 
-					/* Symmetrical to the above rule which styles the last paragraph
-					 * to provide more consistent experience with the first one.
+					/* Mirrors the rule above for the last paragraph to keep the
+					 * experience consistent with the first paragraph.
+					 *
+					 * Together, these rules prevent margins from appearing when a
+					 * bogus paragraph becomes a real paragraph after it receives line
+					 * height, text alignment, or other block style.
+					 *
 					 * See: https://github.com/ckeditor/ckeditor5/pull/18801
 					 */
 					& > p:last-of-type {
-						margin-block-end: 0;
-					}
-
-					/* Rule which prevents the margins to appear when a
-					 * bogus paragraph gets a line height or text alingment.
-					 * We use only-of-type as the cell may have a div with column resize.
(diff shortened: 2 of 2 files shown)


Witoso:
Notes from architecture guild (@oleq, @mmotyczynska, @niegowski, @pszczesniak)
* overall ok, and everyone was eager to do it,
* but we need to think how this affects https://github.com/ckeditor/ckeditor5/pull/15079, and if the rule is not general for lists and tables: _Cut the margin of the first item, cut the margin of the first item_, which initially in one paragraph cuts both.

pomek:
I do not understand too much. I need more guidance on that, or please, take it from me and finish it on your own.

Witoso:
Yep, I will take it for now.

packages/ckeditor5-table/theme/table.css
```diff
@@ -54,9 +54,35 @@
...
+					 * bogus paragraph gets a line height or text alingment.
+					 * We use only-of-type as the cell may have a div with column resize.
+					 * to provide more consistent experience with the first one.
+					 * See: https://github.com/ckeditor/ckeditor5/pull/18801
+					 */
+					& > p:only-of-type {
+						margin-block-start: 0;
+						margin-block-end: 0;
+					}
+
```
pszczesniak:
```suggestion
```

It's not needed cause `& > p:first-of-type` and `& > p:last-of-type` will cover the case when only single `p` will be there.

packages/ckeditor5-list/theme/listformatting.css
```diff
@@ -10,6 +10,26 @@
...
+	/* Rule which prevents the margins to appear when a
+	 * bogus paragraph gets a line height or text alignment.
+	 * We use only-of-type as the cell may have a div with column resize.
+	 * to provide more consistent experience with the first one.
+	 * See: https://github.com/ckeditor/ckeditor5/pull/18801
+	 */
+	& > p:only-child {
+		margin-block-start: 0;
+		margin-block-end: 0;
+	}
```
pszczesniak: Similar like in the other comment, we can change it to:

```suggestion
	& > p:last-of-type {
		margin-block-end: 0;
	}
```

And comment should also be updated to match the selector we used.

Witoso: we will loose this space then in multiblock lists:

image

pszczesniak: Yes, correct, i forget about multiblock lists.

packages/ckeditor5-table/theme/table.css
pszczesniak: ~~Hmm... when there will be more than one paragraph - should the last one have margin bottom?~~

pszczesniak: ~~If yes, all is fine~~

After thinking it once again - yes, it can be removed.

pszczesniak (approved):

packages/ckeditor5-list/theme/listformatting.css
```diff
@@ -10,6 +10,24 @@
...
+	& > p:first-of-type {
+		margin-top: 0;
+	}
+
+	/* Prevents margins from appearing when a bogus paragraph
+	 * receives line height or text alignment.
+	 * Ensures a consistent experience with the first paragraph.
+	 * See: https://github.com/ckeditor/ckeditor5/pull/18801
+	 */
+	& > p:only-child {
```
Witoso, line 26:
```suggestion
	& > p:only-of-type {
```

PR #18811, opened 2025-07-04
Display errors when building the editor for docs

### LINKED ISSUE #18810: Display errors when building the editor for docs
Currently, when building the `ckeditor5` and `ckeditor5-premium-features` packages for docs, the actual error that caused the build to fail is hidden and only a generic `error Command failed with exit code 1` message is displayed.

We should show the actual error message to make it easier to find the root cause of the issue.

filipsobol:
### 🚀 Summary

Show the actual error that caused the docs build to fail instead of only showing a generic `error Command failed with exit code 1` message.

---

### 📌 Related issues

* Closes #18810

---

### 💡 Additional information

**Before:**
Zrzut ekranu z 2025-07-04 14-56-09

**After:**
Zrzut ekranu z 2025-07-04 14-55-47

### DIFF
--- a/scripts/docs/build-ckeditor-packages.mjs
+++ b/scripts/docs/build-ckeditor-packages.mjs
@@ -13,7 +13,8 @@ import { CKEDITOR5_ROOT_PATH, CKEDITOR5_COMMERCIAL_PATH } from '../constants.mjs
 import generateCKEditor5DocsBuild from './generate-ckeditor5-docs-build.mjs';
 
 buildCKEditorPackages()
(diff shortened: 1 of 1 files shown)


pomek:
What about an error in case of building the Premium features package? Will it be displayed, too?

filipsobol:
> What about an error in case of building the Premium features package? Will it be displayed, too?

This function builds both the core and commercial packages, so errors from both are now displayed properly.

pomek (approved):
If so, then R+.

Issue #18822, opened 2025-07-08
Emoji loading causing significant loading overhead?


eliandoran:
## 📝 Ask a question

I've noticed this under Electron, it appears that first time loading is slower than it used to.
While doing some profiling I've noticed that `_loadAndCacheEmoji` takes a total of 540ms.

<img width="1208" height="350" alt="Image" src="https://github.com/user-attachments/assets/0ef2b158-f0e2-4c5d-a47d-7ee241d466e1" />

On a closer look, it appears that most of the time is spent recalculating layout:

<img width="536" height="53" alt="Image" src="https://github.com/user-attachments/assets/a3cf0fee-997e-482b-aef7-b9d4d9988e35" />

The stack trace is:

getNodeWidth | @ | ckeditor5.js:269:1553
-- | -- | --
  isEmojiZwjSupported | @ | ckeditor5.js:269:1422
  isEmojiSupported | @ | ckeditor5.js:269:1209
  (anonymous) | @ | ckeditor5.js:269:5139
  _normalizeEmoji | @ | ckeditor5.js:269:5127
  _loadAndCacheEmoji | @ | ckeditor5.js:269:4789

eliandoran:
[Trace-20250708T195034.json](https://github.com/user-attachments/files/21126398/Trace-20250708T195034.json)

eliandoran:
From what I see in the minified code, the calculation of emoji size appears to be quite time-consuming since it creates an element on the DOM for a large number of emojis, forcing a reflow every time.

```
Get node width  😶‍🌫️
ckeditor5.js:35438 node: 0.364013671875 ms
ckeditor5.js:35432 Get node width  😮‍💨
ckeditor5.js:35438 node: 0.14306640625 ms
ckeditor5.js:35432 Get node width  🙂‍↔️
ckeditor5.js:35438 node: 0.273193359375 ms
ckeditor5.js:35432 Get node width  🙂‍↕️
ckeditor5.js:35438 node: 0.205810546875 ms
ckeditor5.js:35432 Get node width  😵‍💫
ckeditor5.js:35438 node: 0.1318359375 ms
ckeditor5.js:35432 Get node width  ❤️‍🔥
ckeditor5.js:35438 node: 0.138916015625 ms
```

Witoso:
Hi!

> first time loading

Do you see lag in the UI loading? It's not something I can reproduce on any website.

> than it used to.

Please specify what you are comparing.

[...]

PR #18848, opened 2025-07-14
Improve performance of `isEmojiZwjSupported` method

filipsobol:
### 🚀 Summary

Improve the initialization performance of the Emoji plugin by optimizing the `isEmojiSupported` method.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/18822

---

### 💡 Additional information

This PR introduces a two-step validation for checking supported emojis:
1. **Canvas-based detection** – Fast but less reliable. We first attempt to detect emoji support using the canvas API for performance.
2. **DOM-based fallback** – Slower but more accurate. If the canvas check fails, we fall back to using a DOM-based method, which offers broader and more reliable emoji support.

This hybrid strategy improves overall detection speed without sacrificing compatibility.


packages/ckeditor5-emoji/src/emojiutils.ts
```diff
@@ -44,6 +44,9 @@ const EMOJI_SUPPORT_LEVEL: Record<string, number> = {
 
 const BASELINE_EMOJI_WIDTH = 24;
 
+const CANVAS_FOR_TESTING_EMOJI_WIDTH = /** #__PURE__ */ ( () => document.createElement( 'canvas' ).getContext( '2d' )! )();
+const CONTAINER_STYLES_MAP = /* #__PURE__ */ new Map<HTMLDivElement, CSSStyleDeclaration>();
```
pomek: Consider moving both of the constants to a plugin class as private members. This way, accessing the `document` global will crash only at run-time instead of the import phase when running SSR.

The initialization can be done in the `Plugin#init()` method. Do we need to clean anything? Perhaps, on `Plugin#destroy()` we could leave a blank canvas to avoid leaking _what it does_.

pomek (commented):
Consider:

* creating the `<span>` element and container only once in `init()`
* updating the text content each time `getNodeWidth()` is called

The element could be removed from the DOM in the `destroy()` method.

It may not impact the performance as high as we expect, but it would be aligned with another change I propose (related to Canvas).

https://github.com/ckeditor/ckeditor5/blob/a7b8f25f3049438d11253984636ce2afc18b88fe/packages/ckeditor5-emoji/src/emojiutils.ts#L128-L139

Reinmar:
@filipsobol, since in the PR we claim that improves the performance – could we have sobe before/after numbers?

pomek:
I made several checks by profiling the `EmojiRepository#_normalizeEmoji()` method.

https://github.com/ckeditor/ckeditor5/blob/166422b7a55f6f07cecc51ec0c00bdc98e976327/packages/ckeditor5-emoji/src/emojirepository.ts#L337-L361

The proposed solution speeds up by about ~25% (total execution time).

When it comes to DOM recalculating:

- `#master` 
	image
- `#ck/18822-improve-emoji-performance` ()
	image

Below is an analysis of two traces (`before` and `after` code changes), focusing on the performance of the `_normalizeEmoji` function.


| Metric                              | Before Changes                      | After Changes                       |
|-------------------------------------|-------------------------------------|-------------------------------------|
| Number of `_normalizeEmoji` calls   | ≥ 15                                | **4** calls                         |
| Associated `Layout` triggers        | ≥ 9 instances                       | **0 direct `Layout` events**       |
| Total CPU time (`_normalizeEmoji`)  | noticeable (tens of ms)             | **~3.0 ms** total                   |
| Side effects (`Recalc Style`, `Paint`) | present                          | **no direct correlation observed** |

- After the update, `_normalizeEmoji` **no longer triggers layout or style recalculations** directly.
- The trace shows that execution time has significantly decreased, with no layout-related activity tied to this function.

At least, on my machine.

Data:

- [Trace-20250715T104058--before.json](https://github.com/user-attachments/files/21229127/Trace-20250715T104058--before.json)
- [Trace-20250715T104324--after.json](https://github.com/user-attachments/files/21229129/Trace-20250715T104324--after.json)

packages/ckeditor5-emoji/src/emojiutils.ts
filipsobol: Done :heavy_check_mark:

packages/ckeditor5-emoji/src/emojiutils.ts
```diff
@@ -48,6 +48,11 @@ const BASELINE_EMOJI_WIDTH = 24;
  * The Emoji utilities plugin.
  */
 export class EmojiUtils extends Plugin {
+	/**
+	 * Used for testing whether the given emoji is supported by the environment.
+	 */
+	private _CANVAS_FOR_TESTING_EMOJI_WIDTH: CanvasRenderingContext2D | null = null;
```
pomek:
```suggestion
	/**
	 * Used for testing whether the environment supports the given emoji.
	 */
	private _emojiCanvas: CanvasRenderingContext2D | null = null;
```

filipsobol:
> This pull request updates an existing method (`isEmojiZwjSupported()`) and adds a new one (`getNodeWidthUsingCanvas()`).
> 
> Both of them miss the corresponding tests.

`isEmojiZwjSupported` is already covered by tests, and both the `getNodeWidth` and `getNodeWidthUsingCanvas` are just its implementation details. Adding tests for them would essentially come down to duplicating existing tests for `isEmojiZwjSupported` which already calls these two methods under the hood.

I would make the `getNodeWidthUsingCanvas` a private method if `getNodeWidth` wasn't already public.

pomek:
> are just its implementation details

Both of the methods are available for integrators. I see that `getNodeWidth()` misses its tests, so it would be good to add them as well.

I propose rewriting `isEmojiZwjSupported()` tests to be unit-like, and writing integration-like tests for `getNodeWidthUsingCanvas()` and `getNodeWidth()`. This way, we avoid duplicating the code and cover all use cases.

filipsobol:
LMK if this is what you expect.

.changelog/20250714130235_ck_18822_improve_emoji_performance.md
```diff
@@ -0,0 +1,7 @@
+---
+type: Other
+scope: ckeditor5-emoji
+closes: https://github.com/ckeditor/ckeditor5/issues/18822
+---
+
+Improve the initialization performance of the Emoji plugin by optimizing the `isEmojiSupported` method.
```
pomek:
```suggestion
Improve the initialization performance of the Emoji feature by optimizing the `EmojiUtils#isEmojiSupported()` method.
```

pomek (dismissed):
LGTM.

PR #18859, opened 2025-07-15
Set concurrency in release scripts to guarantee at least 1 CPU core and 1 GB of free RAM per task

filipsobol:
### 🚀 Summary

Set concurrency in release scripts to guarantee at least 1 CPU core and 1 GB of free RAM per task.

### 📌 Related issues

* See https://github.com/cksource/ckeditor5-commercial/issues/8048.

---

This PR addresses an out-of-memory (OOM) issue by capping the maximum build concurrency.

I benchmarked the impact of different concurrency levels on build performance. The results were:

* **1 → 4 concurrent builds**: Significant improvement, reducing build time from ~450 s to ~270 s.
* **4 → 8 concurrent builds**: Modest improvement, dropping to ~240 s.
* **8 → 16 concurrent builds**: No measurable performance gain.

~~Given the diminishing returns beyond 4 concurrent builds and to avoid more complex solutions (for example, based on CPU cores **and** available memory), this PR sets the maximum concurrency to 4 as a balanced default.~~

I verified the fix by running the full CI pipeline three times for this PR and cksource/ckeditor5-commercial#8052. The builds completed successfully (the failing CI status in the second PR is unrelated and occurred in post-build steps).

### DIFF
--- a/scripts/release/utils/parsearguments.mjs
+++ b/scripts/release/utils/parsearguments.mjs
@@ -4,7 +4,7 @@
  */
 
 import minimist from 'minimist';
-import os from 'os';
+import { cpus, freemem } from 'os';
 import replaceKebabCaseWithCamelCase from '../../utils/replacekebabcasewithcamelcase.mjs';
 
 /**
@@ -41,7 +41,16 @@ export default function parseArguments( cliArguments ) {
 			nightly: false,
 			'nightly-alpha': false,
 			'nightly-next': false,
-			concurrency: os.cpus().length / 2,
+			// Set concurrency to guarantee at least 1 CPU core and 1 GB of free RAM per task.
+			// No less than 1, no more than 8.
+			concurrency: Math.max(
+				1,
+				Math.min(
+					cpus().length,
+					Math.floor( freemem() / ( 1024 ** 3 ) ),
+					8
+				)
+			),
 			'compile-only': false,
 			packages: null,
 			branch: 'release',


scripts/release/utils/parsearguments.mjs
```diff
@@ -41,7 +41,7 @@ export default function parseArguments( cliArguments ) {
 			nightly: false,
 			'nightly-alpha': false,
 			'nightly-next': false,
-			concurrency: os.cpus().length / 2,
+			concurrency: Math.min( os.cpus().length, 4 ),
```
pomek: The problem occurs mostly on CI. Is there any reason why we limit it locally?

pomek: We can hide the math behind a CI flag:

https://github.com/ckeditor/ckeditor5/blob/9827a22a5a60fe70bc3089e8ec46e822733f6e74/scripts/release/utils/parsearguments.mjs#L83-L85

pomek: My results:

| `--concurrency` | `real` Time |
|-----------------|-------------|
| 2               | 3m23.623s   |
| 4               | 2m31.454s   |
| 8               | 1m59.192s   |
| 16              | 2m0.753s    |
| 32              | 2m3.181s    |

filipsobol: I updated the implementation to guarantee at least 1 CPU core and 1 GB of free RAM per task, but no less than 1 task, and no more than 8 tasks.

pomek (approved):
LGTM.

PR #12972, opened 2022-12-05
Use global not as variable name to fix problem with Jest

alexander-schranz:
### Suggested merge commit message ([convention](https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html))

Type: Message. Closes https://github.com/ckeditor/ckeditor5/issues/12971.

---

### Additional information

Currenlty jest tests failing when upgrading ckeditor from 34 to 35.

PR #18886, opened 2025-07-23
Fixed duplicate global identifier

### LINKED ISSUE #18856: Uncaught SyntaxError: Identifier 'global' has already been declared
Hello,

When running CKEditor V46 in an Expo Web application I've got this error : `Uncaught SyntaxError: Identifier 'global' has already been declared`.

There is an old PR #12972 which has fixed it but this recent PR #18689 created by @martnpaneq has broken code and ignored original comment from this file :
https://github.com/ckeditor/ckeditor5/blob/7bc3680386d86ee351a39b2ff9e91deb50aa8adb/packages/ckeditor5-utils/src/dom/global.ts#L35

Can I create a PR to revert this file to the previous commit?

przemyslaw-zan:
### 🚀 Summary

Fixed the `Identifier 'global' has already been declared.` error being thrown in some environments due to global variable name in `ckeditor5-utils` package.

---

### 📌 Related issues

* Closes #18856

---

### 💡 Additional information

### DIFF
--- a/.changelog/20250723121141_ck_18856.md
+++ b/.changelog/20250723121141_ck_18856.md
@@ -1,0 +1,45 @@
+---
+# Required: Type of change.
+# Allowed values:
+# - Feature
+# - Fix
+# - Other
+# - Major breaking change
+# - Minor breaking change
+#
+# For guidance on breaking changes, see:
+# https://ckeditor.com/docs/ckeditor5/latest/updating/versioning-policy.html#major-and-minor-breaking-changes
+type: Fix
+
+# Optional: Affected package(s), using short names.
+# Can be skipped when processing a non-mono-repository.
+# Example: ckeditor5-core
+scope:
+  - ckeditor5-utils
+
+# Optional: Issues this change closes.
+# Format:
+# - {issue-number}
+# - {repo-owner}/{repo-name}#{issue-number}
+# - Full GitHub URL
+closes:
+  - 18856
+
+# Optional: Related issues.
+# Format:
+# - {issue-number}
+# - {repo-owner}/{repo-name}#{issue-number}
+# - Full GitHub URL
+see:
+  - 
+
+# Optional: Community contributors.
+# Format:
+# - {github-username}
+communityCredits:
+  - 
+
+# Before committing, consider removing all comments to reduce file size and enhance readability.
+---
+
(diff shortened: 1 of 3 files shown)


pomek:
Could you confirm whether the tests in Commercial are still passing without issue due to the change?

przemyslaw-zan:
☝️ Private PR has passed CI

pomek:
I performed several checks with AI to confirm whether the proposed solution resolves the issue.

---

# Summary: CKEditor 5 `global` Identifier Issue in Expo Web

## ✅ Problem

Using CKEditor 5 in an **Expo Web** app throws:

```
Uncaught SyntaxError: Identifier 'global' has already been declared
```

## 🛠️ Cause

Some bundlers (e.g., Metro or Webpack in Expo/React Native) treat `global` as a reserved or global identifier, which causes naming collisions when CKEditor 5 internally declares a variable named `global`.

## 🔧 Original Fix

[PR #12972](https://github.com/ckeditor/ckeditor5/pull/12972) renamed the internal variable to `globalVar`, while still exporting it as `global` for public API compatibility:

```ts
let globalVar = { window, document };
export { globalVar as global };
```

This avoids redefining the `global` identifier in the module scope.

## 🚨 Recent Regression

[PR #18689](https://github.com/ckeditor/ckeditor5/pull/18689) unintentionally broke this fix, possibly by ignoring or removing the original safeguard comment.

## ✅ Safe Workaround in Your App

When importing, **alias the export** to avoid naming conflicts:

```ts
import { global as globalVar } from '@ckeditor/ckeditor5-utils';

console.log(globalVar.window); // ✅ Safe access
```

This avoids introducing a new `global` identifier in your local scope.

## 📌 Conclusion

- Do **not** rename the internal variable back to `global`.
- Always **export as `global`**, but **use `globalVar` internally** to prevent naming collisions.
- Consumers can **safely import** it as `globalVar` using aliasing.

```ts
import { global as globalVar } from '@ckeditor/ckeditor5-utils';
```

pomek:
@rmarquois, would you like to take a look to confirm if we follow the right direction?

pomek:
@przemyslaw-zan, could you add a unit test that confirms the `utils/dom/global` module **DOES NOT** export a member called `global`?

rmarquois:
> @rmarquois, would you like to take a look to confirm if we follow the right direction?

Looks good for me. Thanks for the PR :)

PR #18893, opened 2025-07-24
Allow using custom button labels in a dropdown component [WIP]

f1ames:
### 🚀 Summary

Extended dropdown component utils to allow creating a dropdown list with custom `ButtonLabelView` (passed from outside code, not auto-created).

---

### 📌 Related issues

None.

---

### 💡 Additional information

Required for https://github.com/cksource/ckeditor5-commercial/pull/8070.

### DIFF
--- a/packages/ckeditor5-ui/src/dropdown/utils.ts
+++ b/packages/ckeditor5-ui/src/dropdown/utils.ts
@@ -678,19 +678,14 @@ function bindViewCollectionItemsToDefinitions(
 			groupView.items.delegate( 'execute' ).to( dropdownView );
 
 			return groupView;
-		} else if ( def.type === 'button' || def.type === 'switchbutton' || def.type === 'labelledbutton' ) {
+		} else if ( def.type === 'button' || def.type === 'switchbutton' ) {
 			const isToggleable = def.model.role === 'menuitemcheckbox' || def.model.role === 'menuitemradio';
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-ui/src/dropdown/utils.ts
```diff
@@ -677,7 +678,7 @@ function bindViewCollectionItemsToDefinitions(
 			groupView.items.delegate( 'execute' ).to( dropdownView );
 
 			return groupView;
-		} else if ( def.type === 'button' || def.type === 'switchbutton' ) {
+		} else if ( def.type === 'button' || def.type === 'switchbutton' || def.type === 'labelledbutton' ) {
```
oleq: Can't we just make `def.label` a general property for when `def.type === 'button'` (in short: add it to `ListDropdownButtonDefinition`) and avoid adding additional type here? It's not that different, to deserve a new type IMO.

oleq (changes requested):

packages/ckeditor5-ui/src/dropdown/utils.ts
oleq: Addressed in https://github.com/ckeditor/ckeditor5/pull/18893/commits/ce8ec83df34ce02f6b3676eb01d017152359d8a3.

PR #18923, opened 2025-08-01
Enhanced error tracking.

Mati365:
### 🚀 Summary

This PR adds a plugin that allows to dynamically track actions performed in the editor.


arkflpc (dismissed):

Reinmar:
I can see the `onBeforeAction` and `onAfterAction` callbacks, but I don't really understand why are there two callbacks.

The PR also doesn't include documentation how to actually use the ActionsRecorder in real systems. Where should integrators plug their monitoring? How to serialize the entries to send them to your backend? What about the `before` and `after` props that they contain? What's the meaning of those?

There's also a manual test but it does not answer any of my questions. It's not really showing (e.g. by simulating this) what to do with this data.

In other words: We need to see a real integration scenario. Ideal, in form of a longer description with in the ActionsRecorder API docs and a manual test.

packages/ckeditor5-engine/src/view/documentfragment.ts
```diff
@@ -125,6 +125,22 @@ export class ViewDocumentFragment extends /* #__PURE__ */ EmitterMixin( ViewType
 		yield* this._customProperties.entries();
 	}
 
+	/**
+	 * Converts `ViewDocumentFragment` instance to plain object and returns it.
+	 * Takes care of converting all of this document fragment's children.
+	 *
+	 * @returns `ViewDocumentFragment` instance converted to plain object.
+	 */
+	public toJSON(): unknown {
```
arkflpc, line 134: Nitpick: returning `{children: [ ... ]}` would be more future-proof than plain array.

packages/ckeditor5-engine/src/view/rooteditableelement.ts
```diff
@@ -49,6 +49,15 @@ export class ViewRootEditableElement extends ViewEditableElement {
 		this._setCustomProperty( rootNameSymbol, rootName );
 	}
 
+	/**
+	 * Converts `ViewRootEditableElement` instance to string and returns it.
+	 *
+	 * @returns `ViewRootEditableElement` instance converted to string.
+	 */
+	public override toJSON(): unknown {
+		return this.rootName;
```
arkflpc, line 58: Nitpick: returning `{rootName: '...'}` would be more future-proof.

packages/ckeditor5-watchdog/src/actionsrecorderconfig.ts
```diff
@@ -0,0 +1,238 @@
...
+	 */
+	params?: Array<any>;
+
+	/**
+	 * The result returned by the executed action. Those depend on the actual action.
+	 */
+	result?: any;
+
+	/**
+	 * The error if the action throw any.
```
arkflpc:
```suggestion
	 * The error if the action throws one.
```

packages/ckeditor5-watchdog/src/actionsrecorderconfig.ts
```diff
@@ -0,0 +1,238 @@
...
+	 */
+	after?: ActionsRecorderEntryEditorSnapshot;
+
+	/**
+	 * Params provided for the executed action. Those depend on the actual action.
+	 */
+	params?: Array<any>;
+
+	/**
+	 * The result returned by the executed action. Those depend on the actual action.
```
arkflpc:
```suggestion
	 * The result returned by the executed action. It depends on the actual action.
```

packages/ckeditor5-watchdog/src/actionsrecorderconfig.ts
```diff
@@ -0,0 +1,238 @@
...
+	 */
+	before: ActionsRecorderEntryEditorSnapshot;
+
+	/**
+	 * The editor state after the action was executed.
+	 */
+	after?: ActionsRecorderEntryEditorSnapshot;
+
+	/**
+	 * Params provided for the executed action. Those depend on the actual action.
```
arkflpc:
```suggestion
	 * Params provided for the executed action. They depend on the actual action.
```

packages/ckeditor5-watchdog/src/actionsrecorder.ts
```diff
@@ -0,0 +1,780 @@
...
+
+import { isPlainObject } from 'es-toolkit/compat';
+
+/**
+ * A plugin that records user actions and editor state changes for debugging purposes. It tracks commands execution, model operations,
+ * UI interactions, and document events. It just collects data locally, and does not send it anywhere, integrator is responsible
+ * for gathering data from this plugin for further processing.
+ *
+ * By default, plugin stores latest 1000 action entries. Integrator can register an `onError` callback to collect those entries
+ * along with the caught exception, augment those data with application data such as page-id or session-id, depending on the application,
```
arkflpc: The collected entries contain exceptions. IMO, "along with the caught exception" is confusing (especially in singular form, when talking about many entries - they have individual `error` fields). The rest of the description (starting with "augment") should be a separate statement(s) and we need to make it clear that we're done with describing the callback and we're back to integrator duties.

packages/ckeditor5-watchdog/tests/actionsrecorder.js
```diff
@@ -0,0 +1,424 @@
...
+
+			const records = plugin.getEntries();
+			const commandRecord = records.find( record => record.action === 'commands.foo:execute' );
+
+			expect( commandRecord ).to.exist;
+			expect( commandRecord.error ).to.exist;
+		} );
+	} );
+
+	describe( 'record management', () => {
```
arkflpc, line 232: Missing test(s) for `onMaxEntries` callback (include scenario for my comment in `src/actionsrecorder.ts`).

packages/ckeditor5-watchdog/src/actionsrecorder.ts
```diff
@@ -0,0 +1,780 @@
...
+		};
+
+		// Apply filter if configured, only add to entries if filter passes.
+		if ( !this._filterCallback || this._filterCallback( callFrame, this._entries ) ) {
+			// Add the call frame to the entries.
+			this._entries.push( callFrame );
+
+			// Enforce max entries limit.
+			if ( this._entries.length >= this._maxEntries ) {
+				this._maxEntriesCallback();
```
arkflpc: Calling `_maxEntriesCallback()` in `_enterFrame` instead of `_leaveFrame` (or, alternatively, adding it to `_entries` here) means that the last entry will have no `after`, `result` nor `error` fields properly set. If the integrator follows the examples in the docs for `onMaxEntries`, they'll call `flushEntries()` and this record would be already lost when `after`, `result` or `error` is available.

arkflpc: Actually, the problem is more in adding records to `_entries` than calling `_maxEntriesCallback`. Consider the scenario when the integrator sets `maxEntries` to infinity and checks them in `setInterval` - they could read and clear incomplete frames.

packages/ckeditor5-engine/src/view/documentfragment.ts
niegowski: This is symmetrical to the model equivalent.

packages/ckeditor5-engine/src/view/rooteditableelement.ts
niegowski: This is symmetrical to the model equivalent. Additionally, when dumping view position, view range, and view selection, we report the root, and dumping the whole editable would incur huge overhead every time.

packages/ckeditor5-watchdog/src/actionsrecorder.ts
niegowski: Addressed in 4a912b5.

packages/ckeditor5-watchdog/tests/actionsrecorder.js
niegowski: Addressed in 4a912b5.

packages/ckeditor5-watchdog/src/actionsrecorder.ts
niegowski: Addressed in d757c29.

PR #18925, opened 2025-08-01
Root package migration to packages directory

### LINKED ISSUE #18890: Introduce the `ckeditor5` package in the repository

This issue aims to make a standalone `ckeditor5` package by extracting it from the root repository to its own directory: `./packages/ckeditor5/`.

It reorganizes source and test files, updates configuration and build scripts, and refines the release and documentation workflows to align with the new package structure.

This change is necessary because some package managers do not recognize the root project directory as a valid package, which can result in missing dependencies in `node_modules/`. Moving `ckeditor5` into a dedicated package ensures consistent installation and resolution.

### Definition of Done

- Before the final merge, #18891 must be resolved. It does not block working on the issue, though.
- Create a new directory: `./packages/ckeditor5/`
  - Move the `src/` and `tests/` to the new directory.
  - The new directory exports a package called `ckeditor5`.
  - Based on [`getCKEditor5PackageJson()`](https://github.com/ckeditor/ckeditor5/blob/master/scripts/release/utils/getckeditor5packagejson.mjs) and the root `package.json`, define a dedicated JSON for the `ckeditor5` package.
  - Ignore the `exports` field. It must be added while preparing the packages to release.
  - Duplicate the [`COPYING.GPL`](https://github.com/ckeditor/ckeditor5/blob/master/COPYING.GPL) file to make sure it lands on npm.
- Rename the root `package.json` ([ref](https://github.com/ckeditor/ckeditor5/blob/master/package.json#L2)): `@ckeditor/ckeditor5`. It should be `private: true`. 
- Move `ckeditor5` related `tsconfig.*.json` files to the dedicated directory.
- Move dedicated npm scripts from the root `package.json` to `ckeditor5`.
- Remove `ckeditor5` from ESLint config. https://github.com/ckeditor/ckeditor5/blob/ede1a8e569d1acf2ddf82f0b2cde696d87afdfef/eslint.config.mjs#L202

**New installation methods:**

- Update internal checks to verify the `ckeditor5` package instead of the root directory.

**DLLs:**

- Ensure the DLL builder works:
  - The main DLL build lands in `./packages/ckeditor5/build/`.
  - CKEditor 5 features land in `./packages/*/build/` and they **DO NOT** include anything related to the framework, e.g., the `Plugin` class isn't defined, but it's imported.
  - Ensure Commercial DLL does not include CKEditor 5 sources.
  - Ensure Commercial features do not include CKEditor 5 and Commercial content.

**Documentation:**

- Ignore the `./packages/ckeditor5/` files when building API docs.
- Ensure docs assets are copied from the proper directories. https://github.com/ckeditor/ckeditor5/blob/ede1a8e569d1acf2ddf82f0b2cde696d87afdfef/scripts/docs/build-docs.mjs#L65-L82

**Release process:**

- Update the release flow to avoid creating the `ckeditor5` package from scratch. https://github.com/ckeditor/ckeditor5/blob/ede1a8e569d1acf2ddf82f0b2cde696d87afdfef/scripts/release/preparepackages.mjs#L155-L270
- Update the changelog generator to avoid including the `@ckeditor/ckeditor5` package in the _Released packages_ summary section.
- The `ckeditor5` package needs to define its `exports` https://github.com/ckeditor/ckeditor5/blob/ede1a8e569d1acf2ddf82f0b2cde696d87afdfef/scripts/release/utils/getckeditor5packagejson.mjs#L29-L39

przemyslaw-zan:
### 🚀 Summary

This PR moves the root `ckeditor5` package to `packages` directory.

---

### 📌 Related issues

* Closes #18890

---

### 💡 Additional information

https://github.com/ckeditor/ckeditor5-dev/pull/1188 has to be merged and released, and dependencies in this PR updated before the merge.


przemyslaw-zan:
DLL, NIM and docs build successfully. Generated `release` directory has the same structure as before. CI is green. I went through DoD and everything seems in order. Few comments:

> Update the release flow to avoid creating the ckeditor5 package from scratch. 

That step was not removed, just adjusted, as the `ckeditor5` package has to be compiled first for the rest of the process to pass.

> Update the changelog generator to avoid including the @ckeditor/ckeditor5 package in the Released packages summary section.

I've generated the changelog and the new package was not listed, so it seems that no changes are needed.

packages/ckeditor5-adapter-ckfinder/webpack.config.mjs
```diff
@@ -10,7 +10,7 @@ import { builds } from '@ckeditor/ckeditor5-dev-utils';
 export default builds.getDllPluginWebpackConfig( webpack, {
 	themePath: fileURLToPath( import.meta.resolve( '@ckeditor/ckeditor5-theme-lark' ) ),
 	packagePath: import.meta.dirname,
-	manifestPath: fileURLToPath( import.meta.resolve( 'ckeditor5/build/ckeditor5-dll.manifest.json' ) ),
+	manifestPath: fileURLToPath( import.meta.resolve( '@ckeditor/ckeditor5/packages/ckeditor5/build/ckeditor5-dll.manifest.json' ) ),
```
pomek: This is wrong. We should use `resolve( 'ckeditor5/build/...' )`.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "ckeditor5",
+  "name": "@ckeditor/ckeditor5",
```
filipsobol: Why don't we use a name like `ckeditor5-repository` or a similar one to clearly describe what this name represents? The `@ckeditor/ckeditor5` name is too ambiguous.

package.json
```diff
@@ -205,10 +113,8 @@
     "translations:synchronize": "node ./scripts/translations/synchronize.mjs",
```
martnpaneq, line 201: The `yarn test` command is not working for me. I get some error connected with Vitest. I have only `ckeditor5-dev` in external. 

What is interesting, running tests for a single package works fine, e.g. `yarn -f emoji`.

<details>
<summary>Error screenshot</summary>
> 
image

</details>

martnpaneq, line 201: There are some new errors when running `yarn manual`:

<details>

<summary>New errors:</summary>
image

</details>

pomek: `yarn test` without parameters tries to run all tests, including `ckeditor5-dev`. It never worked as expected if you have non-karma tests in `external/`.

martnpaneq: I just checked, and `yarn test` is not throwing on `master`, with `ckeditor5-dev` in `external`.
It would be good to at least, to figure out why the old version didn't throw any errors.

I feel like each package/repository should run the tests itself with a correct test runner and config, instead of running all of the tests globally.

packages/ckeditor5-adapter-ckfinder/webpack.config.mjs
przemyslaw-zan: What about `import.meta.resolve( 'ckeditor5-root/tsconfig.dll.json' )`? That import refers to the new private root package, is this okay to use? I could just do `../..`, But it will be a problem when location of the webpack config changes.

package.json
przemyslaw-zan: I went for `ckeditor5-root`

package.json
martnpaneq: The old version did throw. It started throwing for me after running `yarn reinstall`. 

But the point with each repo running it's own tests is still a good idea. Probably for a followup.

martnpaneq: There are no new errors after a fresh `yarn reinstall`.

packages/ckeditor5/CHANGELOG.md
```diff
@@ -0,0 +1,944 @@
+Changelog
```
martnpaneq, line 1: Are we sure we want to keep the changelog in the `ckeditor5` package, and not the root?

przemyslaw-zan: Addressed in https://github.com/ckeditor/ckeditor5/pull/18925/commits/6bb200cd2ebb6d2b3f3a8f07ac8c3adbe762944d - changelog will now only be in the root of the project, and will be copied during package preparation to the release directory

martnpaneq (approved):

PR #18947, opened 2025-08-10
Add AI chat toggle button to menu bar.

mmotyczynska:
### 🚀 Summary

Internal: Add AI chat toggle button to menu bar.

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

### DIFF
--- a/packages/ckeditor5-ui/src/menubar/utils.ts
+++ b/packages/ckeditor5-ui/src/menubar/utils.ts
@@ -716,7 +716,7 @@ export const MenuBarMenuViewPanelPositioningFunctions: Record<string, Positionin
  * 				items: [
  * 					'menuBar:aiAssistant',
  * 					'menuBar:aiCommands',
- * 					'menuBar:aiChat'
+ * 					'menuBar:toggleAi'
  * 				]
  * 			},
  * 			{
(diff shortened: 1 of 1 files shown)


scofalik:
First of all, I remade this branch/PR. The old one was based on old `master` and it showed 50 commits, while there was just one small change in one commit. I tried to rebase it on `release` but I had tons of conflicts, it was easier for me to recreate this.

Secondly, I wonder if we actually want this. A lot of integrators will not want this button because it will mess up with their integration. I think we could skip it, or make it configurable. But... then I thought - wait - the button will not be added, if there's no plugin (`AIToggleButton`)... But then I reminded myself that we decided to automatically add this plugin (in `AIEditorIntegration`). So... maybe we should revert on this decision and just have `AIToggleButton` as an optional plugin, and no "main editor" plugin for now? cc @DawidKossowski

Witoso:
> But then I reminded myself that we decided to automatically add this plugin (in AIEditorIntegration)

But adding a button still requires toolbar setup, right?

scofalik:
To toolbar, yes. But menu bar has default setting, as soon as the component is available (plugin loaded) it will be there.

Witoso:
Ah gotcha, I vote opt in with a plugin then, I think. Looking at how much power AI integration needs in the demo, I think toggleable behavior should be opt-it as there are too many decisions to make and toggling is only a part of it.

Buut I'm going to contradict myself: if overlay is the default (is it?), toggling is necessary, and I wish to have it.

scofalik (approved):

Issue #18744, opened 2025-06-20
view.TreeWalker may try to create positions "after root" in some specific scenarios


scofalik:
## 📝 Provide detailed reproduction steps (if any)

We were reported following editor crash that happened in some scenarios after the user clicked inside the editor.

![Image](https://github.com/user-attachments/assets/86488fc7-bca3-462d-83ca-d77eb0358949)

Looking at this stack trace, the error happens during `view.Range#getTrimmed()`, most probably fired for a selection range of selection created from DOM when handing selection change event:

```js
const newViewSelection = this.domConverter.domSelectionToView( domSelection );
```

Regardless of the steps to reproduce this error, it seems that maybe in some cases `view.Range#getTrimmed()` can "step out of range" and lead to `TreeWalker` creating an incorrect view position (after an item which has no parent).

Looking further at the stack trace, we see that it crashes in `view.Position#_createAfter()`, which is called in `view.TreeWalker#_formatReturnValue()`. The only place where it could crash is here noted with `/* CRASH */`:

```js
	private _formatReturnValue(
		type: ViewTreeWalkerValueType,
		item: ViewItem,
		previousPosition: ViewPosition,
		nextPosition: ViewPosition,
		length?: number
	): IteratorYieldResult<ViewTreeWalkerValue> {
		// Text is a specific parent, because contains string instead of children.
		// Walker doesn't enter to the Text except situations when walker is iterating over every single character,
		// or the bound starts/ends inside the Text. So when the position is at the beginning or at the end of the Text
		// we move it just before or just after Text.
		if ( item.is( 'view:$textProxy' ) ) {
			// Position is at the end of Text.
			if ( item.offsetInText + item.data.length == item.textNode.data.length ) {
				if ( this.direction == 'forward' && !( this.boundaries && this.boundaries.end.isEqual( this.position ) ) ) {
/* CRASH */				nextPosition = ViewPosition._createAfter( item.textNode );
					// When we change nextPosition of returned value we need also update walker current position.
					this._position = nextPosition;
				} else {
					previousPosition = ViewPosition._createAfter( item.textNode );
				}
			}
```

This is weird, because it means that a view text node did not have a parent, which most probably means this is a removed text node, and maybe the original view range to-be-trimmed also was only on that text node.

[...]

PR #18958, opened 2025-08-13
No longer crash editor when selection change occurs within detached lement and during rendering of the editing.

Mati365:
### 🚀 Summary

No longer crash the editor when a selection change occurs within a detached element and during rendering of the editing.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/18744
---

### 💡 Additional information

Occasionally, such as when removing markers during a focus change, the selection may end up inside a view element whose parent has already been detached from the DOM but is still present in the view DOM mappings. In most cases this is harmless, but if the `selectionchange` event fires before the view is fully synchronized with the DOM converter, some elements in the selection may become orphans.
For example:

obraz

The `comment` marker has been removed from the view, but it's still present in DOM converter mappings.

### DIFF
--- a/packages/ckeditor5-engine/src/view/observer/selectionobserver.ts
+++ b/packages/ckeditor5-engine/src/view/observer/selectionobserver.ts
@@ -13,10 +13,10 @@ import { FocusObserver } from './focusobserver.js';
 import { env, type ObservableChangeEvent } from '@ckeditor/ckeditor5-utils';
 import { debounce, type DebouncedFunc } from 'es-toolkit/compat';
 
-import { type EditingView } from '../view.js';
-import { type ViewDocumentSelection } from '../documentselection.js';
-import { type ViewDomConverter } from '../domconverter.js';
-import { type ViewSelection } from '../selection.js';
+import type { EditingView } from '../view.js';
+import type { ViewDocumentSelection } from '../documentselection.js';
+import type { ViewDomConverter } from '../domconverter.js';
+import type { ViewSelection } from '../selection.js';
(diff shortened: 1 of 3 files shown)


Mati365:
I disabled `MapperCache`. It looks like it's not causing the issue.

packages/ckeditor5-engine/src/view/observer/selectionobserver.ts
```diff
@@ -371,6 +378,31 @@ export class SelectionObserver extends Observer {
...
+
+				if ( !element.parent ) {
+					return true;
+				}
+
+				element = element.parent;
+			}
+
+			return false;
+		} );
```
niegowski: You could use the `root` position property and verify if it is present and of the correct type, rather than manually iterating through parents.

arkflpc (approved):

rutek:
Hello, can I know what the schedule is for releasing this fix? It happens for my client all the time. It looks like if there are many marker changes, it happens almost every time. Thank you!

Mati365:
@rutek Hi! We don't have an ETA for the fix at the moment, but you can check if the issue is resolved in the latest nightly build.

Witoso:
We do actually, as this was already merged it waits for the next release. If everything goes smoothly, it should land next week.

rutek:
Thanks for the quick response! 💪 We use CDN-hosted CKEditor (we use premium plugins), so I guess we cannot use nightly build using the `useCKEditorCloud` helper, so we just need to wait for next week.

PR #18966, opened 2025-08-14
Migrate to pnpm

filipsobol:
### 🚀 Summary

This PR migrates the repository to pnpm. This change should not affect the production code.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4083.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
       - checkout_command
       - install_ssh_keys_command
       - node/install-pnpm:
-          version: 10
+          version: "10"
       - node/install-packages:
           pkg-manager: pnpm
       - run:
--- a/.circleci/template.yml
+++ b/.circleci/template.yml
@@ -56,7 +56,7 @@ commands:
     steps:
       - install_ssh_keys_command
       - node/install-pnpm:
-          version: 10
+          version: "10"
       - node/install-packages:
           pkg-manager: pnpm
       - prepare_environment_variables_commands


pnpm-workspace.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - '@ocular-d/vale-bin'
+  - '@parcel/watcher'
+  - '@swc/core'
+  - core-js
+  - esbuild
+  - hexo-util
+  - puppeteer
+  - sharp
+
+shellEmulator: true
```
filipsobol, line 16: This option enables bash-like shell. It simplifies cross-platform scripting, allowing for scripts like this (which will also work on Windows):

```json
"scripts": {
  "test": "NODE_ENV=test node test.js"
}
```

More information: https://pnpm.io/cli/run#shellemulator

filipsobol: The downside is that it tries to be smart about glob patterns like `**/node_modules` which in the following case crashed v8:

```
pnpx rimraf --glob **/node_modules
```

The workaround was to wrap the glob pattern in quotes:

```
pnpx rimraf --glob \"**/node_modules\"
```

scripts/release/preparepackages.mjs
```diff
@@ -183,6 +183,18 @@ const tasks = new Listr( [
...
+					task: ( ctx, task ) => {
+						return releaseTools.executeInParallel( {
+							packagesDirectory: PACKAGES_DIRECTORY,
+							packagesDirectoryFilter: directory => upath.basename( directory ).startsWith( 'ckeditor5' ),
+							listrTask: task,
+							taskToExecute: prepareDllBuildsCallback,
+							concurrency: cliArguments.concurrency
+						} );
+					}
+				},
```
filipsobol, line 201: Before, DLLs were built inside the `release` folder instead of the `packages` folder. However, this causes issues with TypeScript because it has two copies of some types (one in the `release` folder and the original in the `packages`).

I updated the release process to separate building DLLs and copying them to the CDN folder into individual and independent steps. This allows us to build DLLs in the `packages` folder (similar to other build steps) and still have all files in the `release/cdn` folder.

package.json
```diff
@@ -96,38 +30,18 @@
     "@ckeditor/ckeditor5-dev-docs": "^51.0.0",
     "@ckeditor/ckeditor5-dev-release-tools": "^51.0.0",
     "@ckeditor/ckeditor5-dev-stale-bot": "^51.0.0",
-    "@ckeditor/ckeditor5-dev-tests": "https://gitpkg.vercel.app/ckeditor/ckeditor5-dev/packages/ckeditor5-dev-tests?7a09e87676097130f5af04f324ee39a38ab04c37",
+    "@ckeditor/ckeditor5-dev-tests": "https://gitpkg.vercel.app/ckeditor/ckeditor5-dev/packages/ckeditor5-dev-tests?770b06d64bd3edfcd439cafff4f805cdc8742128",
```
filipsobol: Needs releasing before this PR can be merged.

package.json
```diff
@@ -96,38 +30,18 @@
...
-    "@ckeditor/ckeditor5-format-painter": "46.0.1",
-    "@ckeditor/ckeditor5-import-word": "46.0.1",
     "@ckeditor/ckeditor5-inspector": "^5.0.0",
-    "@ckeditor/ckeditor5-line-height": "46.0.1",
-    "@ckeditor/ckeditor5-list-multi-level": "46.0.1",
-    "@ckeditor/ckeditor5-merge-fields": "46.0.1",
-    "@ckeditor/ckeditor5-mermaid": "ckeditor/ckeditor5-mermaid#v0.0.4",
-    "@ckeditor/ckeditor5-pagination": "46.0.1",
-    "@ckeditor/ckeditor5-paste-from-office-enhanced": "46.0.1",
+    "@ckeditor/ckeditor5-mermaid": "ckeditor/ckeditor5-mermaid#internal/4086-migrate-to-pnpm",
```
filipsobol, line 36: Needs merging before this PR can be merged.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,20 @@
+packages:
```
filipsobol, line 1: _I can't add a comment to the `pnpm-lock.yaml` file, so I'll do it here._

The commercial and core repositories will have their own lock files. They are independent of each other, so they will also need to be updated individually whenever we change dependencies.

package.json
```diff
@@ -226,11 +139,11 @@
...
-    ]
-  },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "version": "^22.18.0",
+      "onFail": "download"
+    }
+  }
```
filipsobol: I'm not sure about this one.

This is a feature that (to my understanding) is standardized and will be added to all package managers and allows them to ensure the proper Node version is used. If the proper version is not available, they can either download it or fail (currently pnpm only supports downloading).

This is nice because if we change the target Node version, pnpm will automagically download it for everyone, so the entire team doesn't have to do it manually.

On the other hand this causes pnpm to use different Node version than the one that is available globally. See an example:

```bash
» pnpm run blah
> ckeditor5-commercial-root@46.0.1 blah /home/filip/Projects/commercial
> node -v

v22.18.0

» node -v
v22.15.0
```

This can make debugging some issues harder if we forget about this difference.

psmyrek:
Building content styles does not work:

```
$ pnpm docs:content-styles

> ckeditor5-root@46.0.2 docs:content-styles D:\tmp\pnpm\ckeditor5-commercial\external\ckeditor5
> node -e "import( './scripts/docs/build-content-styles.mjs' ).then( m => m.default() );"

file:///D:/tmp/pnpm/ckeditor5-commercial/external/ckeditor5/scripts/docs/build-content-styles.mjs:9
import mkdirp from 'mkdirp';
       ^^^^^^
SyntaxError: The requested module 'mkdirp' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:182:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:266:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:644:26)
    at async importModuleDynamicallyWrapper (node:internal/vm/module:436:15)
```

package.json
```diff
@@ -138,5 +138,12 @@
     "**/*.css": [
       "stylelint --quiet --allow-empty-input"
     ]
+  },
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "version": "^22.18.0",
+      "onFail": "download"
```
Witoso: will this work correctly with `nvm` which is a standard in current node versioning?

psmyrek:
The following undocumented pnpm feature (?) will be confusing for many of us: calling `pnpm docs` opens a browser with repository homepage 😱 

To build the docs, we must explicitly call `pnpm run docs`.

It looks like pnpm just executes the [`npm docs`](https://docs.npmjs.com/cli/v11/commands/npm-docs) command.

package.json
filipsobol: I reverted this change already. However, I don't think it would conflict with `nvm`, because when this option is used, pnpm uses its own Node instead of the one installed globally.

filipsobol: More info: https://github.com/ckeditor/ckeditor5/pull/18966#discussion_r2284987751

filipsobol:
> Building content styles does not work:

Fixed

psmyrek:
This will never match anything again:
- https://github.com/ckeditor/ckeditor5/blob/8607f5a3138736adc7530cd2255074771471e91b/scripts/release/utils/isckeditor5packagefactory.mjs#L18
- https://github.com/ckeditor/ckeditor5/blob/8607f5a3138736adc7530cd2255074771471e91b/scripts/release/switchlatestnpm.mjs#L19

psmyrek:
Regarding the docs, there are hundreds mentions about yarn that should be converted to pnpm.
Also, the [Mr. Git guide](https://ckeditor.com/docs/ckeditor5/latest/framework/develpment-tools/mrgit.html) needs an update too.

psmyrek:
Probably not needed anymore:
https://github.com/ckeditor/ckeditor5/blob/8607f5a3138736adc7530cd2255074771471e91b/tsconfig.typedoc.json#L19-L24

psmyrek:
The `scripts/collect-svg-icons.mjs` tries to find icons also from nested `external` directory which no longer exists. 

https://github.com/ckeditor/ckeditor5/blob/8607f5a3138736adc7530cd2255074771471e91b/scripts/collect-svg-icons.mjs#L60

Maybe it should be moved to commercial repository where it makes more sense now.

psmyrek:
The `external/**` exclusion from ESLint config should be moved to commercial repository:
https://github.com/ckeditor/ckeditor5/blob/7477cba9191d9c3e8a5cd2b392e698be15acecee/eslint.config.mjs#L32

ESLint works fine without this change, but it's a bit strange that `ckeditor5` repository defines an exclusion to `external/**`, which will then be used only in `ckeditor5-commercial` (which extends the `ckeditor5` configuration).

filipsobol:
> Regarding the docs, there are hundreds mentions about yarn that should be converted to pnpm. Also, the [Mr. Git guide](https://ckeditor.com/docs/ckeditor5/latest/framework/develpment-tools/mrgit.html) needs an update too.

I've already updated all the docs I could find. The only mentions of Yarn I left are either in migration guides or legacy installation guides. In both cases, I think we can or even should leave them as is.

PR #19016, opened 2025-08-26
Docs: Use `https` to enable `crypto` API.

filipsobol:
### 🚀 Summary

Docs: Use `https` to enable `crypto` API.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-commercial/issues/7888.
* See https://github.com/ckeditor/ckeditor5-commercial/pull/8209.

---

### 💡 Additional information

The new version of `@uploadcare/file-uploader` uses Crypto API, which requires the page to either use HTTPS or a localhost domain. Because we need to use the `fake.ckeditor.com` domain in docs, our only option might be using HTTPS.


filipsobol:
This PR also removes the WProofreader plugin from Fullpage snippets because they cause constant timeouts on CI.

psmyrek (approved):

Issue #18996, opened 2025-08-20
Multi-root example page redirects to the quick start for more details, and there are no related details there


dufipl:
## 📝 Provide a description of requested docs changes

[Multi-root example page](https://ckeditor.com/docs/ckeditor5/latest/examples/builds/multi-root-editor.html#editor-example-configuration) mentions Quick start in context that it contains more details about multi-root editor implementation:

> Check out the [Quick start](https://ckeditor.com/docs/ckeditor5/latest/getting-started/installation/cloud/quick-start.html) guide to learn more about implementing this kind of editor. You will find implementation steps there. You can see this example editor’s code below.

But this quick start guide does not contain any information about the multi-root editor at all. We should fix this.

Witoso:
@godai78 could you take a look?

godai78:
This one will need several minor changes throughout the setup section, including a passage in the QS and also mentioning all editor calls in the [editor types guide](https://ckeditor.com/docs/ckeditor5/latest/getting-started/setup/editor-types.html) as only `ClassicEditor` is explicitly shown in an example.

Witoso:
Maybe we should just remove it or point to the API docs. Other editors are not in Quick Start as well.

godai78:
We should at least add editor calls in the editor types guide -in the API section, the discoverability is lower. And then we can point it there. I'll get to it by the end of the week.

PR #19018, opened 2025-08-26
Docs: adding multi-root configuration.

godai78:
### 🚀 Summary

Docs: adding multi-root configuration.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/18996

---

### 💡 Additional information

### DIFF
--- a/docs/getting-started/integrations-cdn/quick-start.md
+++ b/docs/getting-started/integrations-cdn/quick-start.md
@@ -36,7 +36,7 @@ Then, you need to attach the script with the JavaScript code.
 <script src="https://cdn.ckeditor.com/ckeditor5/{@var ckeditor5-version}/ckeditor5.umd.js"></script>
 ```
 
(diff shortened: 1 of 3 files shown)


godai78:
Should we add it to the zip/npm too?

Witoso:
Let's discuss it. It's not clear why we want to add multi-root and not, for example, inline editor, and why we do it in the quick start. It multiplies plenty of cases: npm/ZIP/Cloud and may be convoluted for a user targeting a quick setup.

godai78:
Because classic, inline and document have all the same configuration (juist different import moniker) and only multi-root is different? And if we showed classic, anyone can use this to configure document or inline, but cannot use it to configure multi-root? What is here to discuss? That's pretty self-explanatory. We should have had multi-root explained long ago.

The QS is the logical place to do it, because it's the place where editor configuration is explained. We may reconsider putting it in the editor types guide maybe only.

Mati365 (approved):

PR #19034, opened 2025-08-29
Fixed an issue where Watchdog incorrectly restored collaboration data

### LINKED ISSUE #19033: Watchdog does not properly restore comments loaded on init

1. Create an editor with Watchdog, Comments, and load-and-save integration ( https://github.com/ckeditor/ckeditor5-collaboration-samples/tree/master/collaboration-for-react can be used)
2. Load content with existing comments and create a new comment after init
3. Crash the editor (for example, via https://github.com/ckeditor/ckeditor5/issues/18953)

### ✔️ Expected result

Editor restarts with all comments present.

### ❌ Actual result

- Existing comment markers are removed, and  comments are not visible
- Existing comments are not in an unlinked state (and do not display in the Comments Archive)
- New comment added in the editing session before the editor crash is intact.

## 📃 Other details

* Browser: …
* OS: …

DawidKossowski:
### 🚀 Summary

Fixed an issue where Watchdog incorrectly restored collaboration data in load&save integrations.

---

### 📌 Related issues

* Closes #19033

### DIFF
--- a/.changelog/20250829113424_feat_nextgenai.md
+++ b/.changelog/20250829113424_feat_nextgenai.md
@@ -1,0 +1,11 @@
+---
+type: Fix
+
+scope:
+  - ckeditor5-watchdog
+
+closes:
+  - 19033
+---
+
+Fixed an issue where Watchdog incorrectly restored collaboration data (comment threads and suggestions) after a crash in load and save integrations.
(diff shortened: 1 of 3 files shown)


scofalik (approved):
The changes make sense to me. But at this stage it is hard to imagine if there are some consequences that we don't see. But it seems better than previous implementation.

PR #18929, opened 2025-08-03
Replace mouse events with pointer events for widget and drag drop handling for better touch devices support

niegowski:
### 🚀 Summary

*A brief summary of what this PR changes.*

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

*Optional: Notes on decisions, edge cases, or anything helpful for reviewers.*

PR #19076, opened 2025-09-08
Fixed non-previewed widget dragging on iOS

### LINKED ISSUE #19068: It's not possible to drag/drop some block widgets on iOS in some editor configurations

1. Prepare some editor with widgets (e.g., `all-features-rtc.html` manual test)
    - HTML snippet with text only
    - Page break
    - Horizontal line
    - Table of Contents with some headers present

### ✔️ Expected result

All widgets can be dragged and dropped in Safari@iOS

### ❌ Actual result

- If widget does not have "draggable" content (determined by iOS?) it will not be moved:
    - Page break and Horizontal line do not have any content
    - HTML snippet with text only is not "draggable" by default
- If widget does have "draggable" content it can be moved but it will only display the "draggable" part in the iOS UI:
    - Table of Contents can be dragged via one of its links and only the link will be displayed as the "dragged" element
    - HTML snippet with "draggable" content like media elements can be dragged and dropped

- Does not seem to reproduce in all editor configurations, for example it's possible to drag/drop the page break element in Pagination demo in docs: https://ckeditor.com/docs/ckeditor5/latest/features/pagination/pagination.html

<details><summary>Recording</summary>
<p>

https://github.com/user-attachments/assets/e4cb5c8a-05f9-426e-a1f3-46fba840449f

</p>
</details> 

### ❓ Possible solution

- Handles for drag and drop could be introduced only on mobile for elements that do not have them

## 📃 Other details

* Browser: Safari
* OS: iOS
* First affected CKEditor version: n/a
* Installed CKEditor plugins: n/a

---

niegowski:
### 🚀 Summary

It's a fix for not yet released drag drop improvement: https://github.com/ckeditor/ckeditor5/pull/18929.

---

### 📌 Related issues

* Closes #19068.

---

### 💡 Additional information

The issue was that we used `DataTransfer#setDragImage()` on iOS. When the preview has no size, then the drag action does not start.

### DIFF
--- a/packages/ckeditor5-clipboard/src/dragdrop.ts
+++ b/packages/ckeditor5-clipboard/src/dragdrop.ts
@@ -705,8 +705,12 @@ export class DragDrop extends Plugin {
 				preview.style.backgroundColor = 'var(--ck-color-base-background)';
 			}
 		} else if ( env.isiOS ) {
-			// Custom preview for iOS.
+			// Custom preview for iOS. Note that it must have some dimensions for iOS to start dragging element.
 			preview.style.maxWidth = `${ editableWidth }px`;
+			preview.style.padding = '10px';
+			preview.style.minWidth = '200px';
+			preview.style.minHeight = '20px';
+			preview.style.boxSizing = 'border-box';
 			preview.style.backgroundColor = 'var(--ck-color-base-background)';
 		} else {
 			// If domTarget is inside the editable root, browsers will display the preview correctly by themselves.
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-clipboard/src/dragdrop.ts
```diff
@@ -705,8 +705,12 @@ export class DragDrop extends Plugin {
 				preview.style.backgroundColor = 'var(--ck-color-base-background)';
 			}
 		} else if ( env.isiOS ) {
-			// Custom preview for iOS.
+			// Custom preview for iOS. Note that it must have some dimensions for iOS to start dragging element.
```
Reinmar, line 708: Missing link to the ticket?

PR #19092, opened 2025-09-10
 Fix show changes icons and add dialog position function

l-lejman:


packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -467,9 +467,13 @@ export interface DialogDefinition {
 	 * Available dialog positions. By default `DialogViewPosition.EDITOR_CENTER` is used for {@link #isModal non-modals}
 	 * and `DialogViewPosition.SCREEN_CENTER` for modals.
 	 *
+	 * If set to a function, it will be called with the DOM root Rect and the dialog Rect as arguments.
+	 * It should return the coordinates of the dialog's position.
+	 *
 	 * {@link module:ui/dialog/dialogview#DialogViewPosition Learn more} about available positions.
 	 */
-	position?: typeof DialogViewPosition[ keyof typeof DialogViewPosition ];
+	position?: typeof DialogViewPosition[ keyof typeof DialogViewPosition ]
```
scofalik, line 475: Followup after merge: let's make this type sane. The function may become its own subtype and the string part of the type could possibly be simplified? Also, instead of writing `| undefined` we can add `?` next to the param name.

packages/ckeditor5-ui/src/dialog/dialogview.ts
```diff
@@ -139,9 +139,13 @@ export class DialogView extends /* #__PURE__ */ DraggableViewMixin( View ) imple
 	/**
 	 * The position of the dialog view.
 	 *
+	 * If set to a function, it will be called with the DOM root Rect and the dialog Rect as arguments.
+	 * It should return the coordinates of the dialog's position.
+	 *
 	 * @observable
 	 */
-	declare public position: typeof DialogViewPosition[ keyof typeof DialogViewPosition ] | null;
+	declare public position: typeof DialogViewPosition[ keyof typeof DialogViewPosition ] |
```
scofalik: Followup after merge: As above.

packages/ckeditor5-ui/src/dialog/dialog.ts
```diff
@@ -467,9 +467,13 @@ export interface DialogDefinition {
 	 * Available dialog positions. By default `DialogViewPosition.EDITOR_CENTER` is used for {@link #isModal non-modals}
 	 * and `DialogViewPosition.SCREEN_CENTER` for modals.
 	 *
+	 * If set to a function, it will be called with the DOM root Rect and the dialog Rect as arguments.
```
scofalik, line 470: Followup after merge: "DOM root Rect" needs clarification. What it is? This is how the param was named inside the function in `DialogView` but we should propose more meaningful name and describe what it actually is, so somebody can correctly use it to position their dialog.

packages/ckeditor5-icons/theme/icons/show-changes-off.svg
scofalik, line 1: Looks like the icon were changed by accident?

EDIT: Okay, they weren't changed by accident because that's even in PR's name :D. So, why such change? Looking at svg code it looks incorrect 🤔.

scofalik (approved):

scofalik:
Because `dialogRect` is always available, and `domRootRect` is sometimes `undefined` I decided to change the order of these parameters in the callback. However, I think it may make sense to pass `viewportRect` if `domRootRect` is not available, so that the custom callback can position in the middle of the screen.

scofalik:
I merged the PR but it seems to me that the icon change is incorrect. Let's discuss it tomorrow.

PR #19115, opened 2025-09-15
Tab/Shift+Tab generic handling.

### LINKED ISSUE #19083: Widget nested editable elements should implement default tab/shift+tab handling for cycling over nested fields.

Before fixing this issue: https://github.com/ckeditor/ckeditor5/issues/15506. 
There was a native tab/shift+tab handling. It sometimes behaved in unexpected ways, as selection would jump over, for example, caption elements of images in an unexpected way. But the native tab handler was used by 3rd-party plugins to navigate over custom widgets with multiple nested editable fields. Currently, plugin developers have to implement their own tab/shift+tab handlers, but this could be generalized and implemented on the editor side.
---

niegowski:
### 🚀 Summary

The Widget feature now implements the default `Tab`/`Shift+Tab` handling.

---

### 📌 Related issues

* Closes #19083

---

### 💡 Additional information

### DIFF
--- a/packages/ckeditor5-engine/tests/view/observer/bubblingemittermixin.js
+++ b/packages/ckeditor5-engine/tests/view/observer/bubblingemittermixin.js
@@ -823,12 +823,12 @@ describe( 'BubblingEmitterMixin', () => {
 					'$document @ lowest (bubbling @ $document)'
 				] );
 
-				// sinon.assert.calledOn( spyCapture, viewDocument );
-				// sinon.assert.calledOn( spyText, viewDocument );
-				// sinon.assert.calledOn( spyP, viewDocument );
-				// sinon.assert.calledOn( spyBlockQuote, viewDocument );
-				// sinon.assert.calledOn( spyRoot, viewDocument );
-				// sinon.assert.calledOn( spyDocument, viewDocument );
+				sinon.assert.calledOn( spyCapture, viewDocument );
+				sinon.assert.calledOn( spyText, viewDocument );
+				sinon.assert.calledOn( spyP, viewDocument );
+				sinon.assert.calledOn( spyBlockQuote, viewDocument );
+				sinon.assert.calledOn( spyRoot, viewDocument );
+				sinon.assert.calledOn( spyDocument, viewDocument );
 			} );
 		} );
 


packages/ckeditor5-engine/tests/view/observer/bubblingemittermixin.js
```diff
@@ -683,6 +687,149 @@ describe( 'BubblingEmitterMixin', () => {
...
+					'$document @ low (bubbling @ $document)',
+					'$document @ lowest (bubbling @ $document)'
+				] );
+
+				// sinon.assert.calledOn( spyCapture, viewDocument );
+				// sinon.assert.calledOn( spyText, viewDocument );
+				// sinon.assert.calledOn( spyP, viewDocument );
+				// sinon.assert.calledOn( spyBlockQuote, viewDocument );
+				// sinon.assert.calledOn( spyRoot, viewDocument );
+				// sinon.assert.calledOn( spyDocument, viewDocument );
```
Mati365:
```suggestion
				sinon.assert.calledOn( spyCapture, viewDocument );
				sinon.assert.calledOn( spyText, viewDocument );
				sinon.assert.calledOn( spyP, viewDocument );
				sinon.assert.calledOn( spyBlockQuote, viewDocument );
				sinon.assert.calledOn( spyRoot, viewDocument );
				sinon.assert.calledOn( spyDocument, viewDocument );
```

eliandoran:
Hi, @niegowski .

In our application (Trilium Notes), we've implemented a tab key to indent/outdent.

In some recent versions, the implementation no longer works as intended, since if you press Tab as soon as there is any widget, it simply jumps around the document to that element.

Is this intentional? I haven't personally seen it anywhere in visual editors.

Our implementation looked like this:

```javascript
export default class IndentBlockShortcutPlugin extends Plugin {

    init() {
        this.editor.keystrokes.set( 'Tab', ( _, cancel ) => {
            // In tables, allow default Tab behavior for cell navigation
            if (this.isInTable()) {
                return;
            }

            const command = this.editor.commands.get( 'indentBlock' );
            if (command?.isEnabled) {
                command.execute();
            }

            // Always cancel in non-table contexts to prevent widget navigation
            cancel();
        } );

        this.editor.keystrokes.set( 'Shift+Tab', ( _, cancel ) => {
            // In tables, allow default Shift+Tab behavior for cell navigation
            if (this.isInTable()) {
                return;
            }

            const command = this.editor.commands.get( 'outdentBlock' );
            if (command?.isEnabled) {
                command.execute();
            }

            // Always cancel in non-table contexts to prevent widget navigation
            cancel();
        } );
    }
```

I was able to find a work-around for it:

```javascript
this.listenTo<ViewDocumentTabEvent>(this.editor.editing.view.document, "tab", (ev, data) => {
    // In tables, allow default Tab behavior for cell navigation
    if (this.isInTable()) return;

    // Always cancel in non-table contexts to prevent widget navigation
    data.preventDefault();
    ev.stop();

    const commandName = data.shiftKey ? "outdentBlock" : "indentBlock";
    const command = this.editor.commands.get(commandName);
    if (command?.isEnabled) {
        command.execute();
    }
}, {
    priority: "highest",
    context: node => isWidget( node ) || node.is( 'editableElement' ),
});
```

niegowski:
> In some recent versions, the implementation no longer works as intended, since if you press Tab as soon as there is any widget, it simply jumps around the document to that element.

This is because you should not use `editor.keystrokes` for [bubbling events](https://ckeditor.com/docs/ckeditor5/latest/framework/deep-dive/event-system.html#view-events-bubbling), especially when many listeners are acting in different contexts for the same event. Please see:
* https://github.com/ckeditor/ckeditor5/blob/b699e0768f213a12705b203d2aa4ae6f13970c7b/packages/ckeditor5-list/src/list/listediting.ts#L415-L426
* https://github.com/ckeditor/ckeditor5/blob/b699e0768f213a12705b203d2aa4ae6f13970c7b/packages/ckeditor5-code-block/src/codeblockediting.ts#L122-L135
* https://github.com/ckeditor/ckeditor5/blob/b699e0768f213a12705b203d2aa4ae6f13970c7b/packages/ckeditor5-widget/src/widget.ts#L223-L233

Note that the listener registered by the widget feature listens on the `low` priority, so your snippet does not require the `highest` priority; the `normal` (default) would be sufficient.

eliandoran:
Hi, @niegowski .

Thanks for the explanation. Much clearer now!

PR #19147, opened 2025-09-24
Module export validator improvements.

### LINKED ISSUE #18993: Improve module export validator
The `!fileName.includes( '/external/' )` check in the following file causes some files to not be validated properly. If I remove it, some new, seemingly valid errors show up in the commercial repository.

https://github.com/ckeditor/ckeditor5/blob/c658326c4389eea9d840da516f43d5fa63496c68/scripts/ci/exports/utils/module.mjs#L35

We need to check if this part can be removed or whether it needs replacing with some other check.

martnpaneq:
### 🚀 Summary

Replaced broad `/external` check with a more precise for exporting public module members in `external/ckeditor5/packages`.

We want to keep the old behaviour, and mark all files from `ckeditor5/packages` as public API.

The rest of the files would have to be set manually to be marked as public API.

---

### 📌 Related issues

* Closes #18993

---

### 💡 Additional information

### DIFF
--- a/scripts/ci/exports/utils/module.mjs
+++ b/scripts/ci/exports/utils/module.mjs
@@ -17,8 +17,12 @@ import { globSync } from 'glob';
 import { CKEDITOR5_ROOT_PATH } from '../../../constants.mjs';
 import { PACKAGES_DIRECTORY } from '../../../release/utils/constants.mjs';
 
-// E.g. `packages/ckeditor5-core/` or `packages/ckeditor5/`.
-const PUBLIC_PACKAGES = globSync( `${ PACKAGES_DIRECTORY }/*/`, { cwd: CKEDITOR5_ROOT_PATH, mark: true } );
+// The trailing `/` is important to mark the end of a package name.
+const PUBLIC_PACKAGES = globSync( `${ PACKAGES_DIRECTORY }/*/`, {
+	cwd: CKEDITOR5_ROOT_PATH,
+	mark: true,
+	absolute: true
+} ).map( upath.normalize );
 
 export class Module {
 	static load( fileName, errorCollector ) {
@@ -378,12 +382,8 @@ export class Module {
 	}
 
 	_isFromPublicPackages( fileName ) {
-		// Checking if fileName contains the package that is marked as public API by default.
-		return PUBLIC_PACKAGES.some( publicPackageName => {
-			const publicPackageFullPath = upath.join( CKEDITOR5_ROOT_PATH, publicPackageName );
-
-			return fileName.includes( publicPackageFullPath );
-		} );
+		// Checking if the file belongs to the public package.
+		return PUBLIC_PACKAGES.some( publicPackagePath => fileName.startsWith( publicPackagePath ) );
 	}
 
 	resolveImportsExports( packages, modules ) {
--- a/scripts/ci/exports/utils/module.mjs
+++ b/scripts/ci/exports/utils/module.mjs
@@ -15,9 +15,10 @@ import { isInternalNode } from './misc.mjs';
 import { createExportResolutionError } from './errorutils.mjs';
 import { globSync } from 'glob';
 import { CKEDITOR5_ROOT_PATH } from '../../../constants.mjs';
+import { PACKAGES_DIRECTORY } from '../../../release/utils/constants.mjs';
 
-// E.g. `packages/ckeditor5-core/` or `packages/ckeditor5`.
-const PUBLIC_PACKAGES = globSync( 'packages/*', { cwd: CKEDITOR5_ROOT_PATH, mark: true } );
+// E.g. `packages/ckeditor5-core/` or `packages/ckeditor5/`.
+const PUBLIC_PACKAGES = globSync( `${ PACKAGES_DIRECTORY }/*/`, { cwd: CKEDITOR5_ROOT_PATH, mark: true } );
 
 export class Module {
 	static load( fileName, errorCollector ) {
@@ -378,7 +379,11 @@ export class Module {
 
 	_isFromPublicPackages( fileName ) {
 		// Checking if fileName contains the package that is marked as public API by default.
-		return PUBLIC_PACKAGES.some( publicPackageName => fileName.includes( publicPackageName ) );
+		return PUBLIC_PACKAGES.some( publicPackageName => {
+			const publicPackageFullPath = upath.join( CKEDITOR5_ROOT_PATH, publicPackageName );
+
+			return fileName.includes( publicPackageFullPath );
+		} );
 	}
 
 	resolveImportsExports( packages, modules ) {
--- a/scripts/ci/exports/utils/module.mjs
+++ b/scripts/ci/exports/utils/module.mjs
@@ -32,8 +32,7 @@ export class Module {
 
 	constructor( fileName, ast, { publicApiTag }, errorCollector ) {
 		this.fileName = fileName.replace( /\.d\.ts$/, '.ts' );
-		// TODO: `/external/` check doesn't seem correct, because it causes some errors to be silenced.
-		this.isPublicApi = fileName.includes( '/external/' ) || this.relativeFileName === 'index.ts' ? true : publicApiTag;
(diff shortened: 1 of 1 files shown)


martnpaneq:
How I tested the change.

1. I run the script `pnpm validate-module-re-exports`. 
2. I made sure it does not return any validation errors.

image

3. I checked that we return correct values for `PUBLIC_PACKAGES`. The suffix `/` is important, because without it, `packages/ckeditor5` path part can be mapped to any package starting with `ckeditor5-...`.

image

scripts/ci/exports/utils/module.mjs
```diff
@@ -13,6 +13,11 @@ import { Declaration } from './declaration.mjs';
 import { ExternalModule } from './externalmodule.mjs';
 import { isInternalNode } from './misc.mjs';
 import { createExportResolutionError } from './errorutils.mjs';
+import { globSync } from 'glob';
+import { CKEDITOR5_ROOT_PATH } from '../../../constants.mjs';
+
+// E.g. `packages/ckeditor5-core/` or `packages/ckeditor5`.
+const PUBLIC_PACKAGES = globSync( 'packages/*', { cwd: CKEDITOR5_ROOT_PATH, mark: true } );
```
pomek:
```suggestion
const PUBLIC_PACKAGES = globSync( `${ PACKAGES_DIRECTORY }/*/`, { cwd: CKEDITOR5_ROOT_PATH } );
```

- `PACKAGES_DIRECTORY` comes from `constants.mjs`
- To match directories only, `glob` suggests using the `/` suffix. Otherwise, you can match files, too.

scripts/ci/exports/utils/module.mjs
```diff
@@ -372,6 +376,11 @@ export class Module {
 		} );
 	}
 
+	_isFromPublicPackages( fileName ) {
+		// Checking if fileName contains the package that is marked as public API by default.
+		return PUBLIC_PACKAGES.some( publicPackageName => fileName.includes( publicPackageName ) );
```
pomek: An edge case, but if `fileName` includes a name of any public package, it will be marked as public.

I would match the packages directory and a package name, e.g., `packages/ckeditor5-core/`.

scripts/ci/exports/utils/module.mjs
martnpaneq: Added in [6976ad4](https://github.com/ckeditor/ckeditor5/pull/19147/commits/6976ad4d501fd2eb3f9af7665e6ec0fae030af58)

The only caveat is that I left `mark: true`, because it is needed for avoiding a bug where `packages/ckeditor5` matches all of existing packages, and `packages/ckeditor5/` matches only this package.

scripts/ci/exports/utils/module.mjs
martnpaneq: I'm not sure if the issue could exist in the proposed implementation. But to be sure, I added full path matching:

```js
		return PUBLIC_PACKAGES.some( publicPackageName => {
			const publicPackageFullPath = upath.join( CKEDITOR5_ROOT_PATH, publicPackageName );

			return fileName.includes( publicPackageFullPath );
		} );`
```

pomek: I hope all paths follow a Unix-style format.

@psmyrek, could you check it, too?

scripts/ci/exports/utils/module.mjs
```diff
@@ -13,6 +13,12 @@ import { Declaration } from './declaration.mjs';
 import { ExternalModule } from './externalmodule.mjs';
 import { isInternalNode } from './misc.mjs';
 import { createExportResolutionError } from './errorutils.mjs';
+import { globSync } from 'glob';
+import { CKEDITOR5_ROOT_PATH } from '../../../constants.mjs';
+import { PACKAGES_DIRECTORY } from '../../../release/utils/constants.mjs';
+
+// E.g. `packages/ckeditor5-core/` or `packages/ckeditor5/`.
```
psmyrek: Since the `mark: true` is important, I would write it down:

```suggestion
// E.g. `packages/ckeditor5-core/` or `packages/ckeditor5/` with a trailing `/` to mark the end of a package name.
```

scripts/ci/exports/utils/module.mjs
```diff
@@ -13,6 +13,12 @@ import { Declaration } from './declaration.mjs';
...
+const PUBLIC_PACKAGES = globSync( `${ PACKAGES_DIRECTORY }/*/`, { cwd: CKEDITOR5_ROOT_PATH, mark: true } );
```
psmyrek: We can prepare here an array of public packages in the format we want to use later. Right now `_isFromPublicPackages()` calculates the absolute path for each found file.

```suggestion
const PUBLIC_PACKAGES = globSync( `${ PACKAGES_DIRECTORY }/*/`, {
	cwd: CKEDITOR5_ROOT_PATH,
	mark: true,
	absolute: true
} ).map( upath.normalize );
```

scripts/ci/exports/utils/module.mjs
```diff
@@ -372,6 +377,15 @@ export class Module {
 		} );
 	}
 
+	_isFromPublicPackages( fileName ) {
+		// Checking if fileName contains the package that is marked as public API by default.
```
psmyrek: We are not checking here the `@publicApi` annotation as this comment suggests.

```suggestion
		// Checking if the file belongs to the public package.
```

scripts/ci/exports/utils/module.mjs
```diff
@@ -372,6 +377,15 @@ export class Module {
...
+		return PUBLIC_PACKAGES.some( publicPackageName => {
+			const publicPackageFullPath = upath.join( CKEDITOR5_ROOT_PATH, publicPackageName );
+
+			return fileName.includes( publicPackageFullPath );
+		} );
```
psmyrek: When we prepare the target data format for the `PUBLIC PACKAGES` array, then we can simplify this util:

```suggestion
		return PUBLIC_PACKAGES.some( publicPackagePath => fileName.startsWith( publicPackagePath ) );
```

psmyrek (commented):
It works, but I found some places to improve.

scripts/ci/exports/utils/module.mjs
martnpaneq: Adressed in [4b24bd2](https://github.com/ckeditor/ckeditor5/pull/19147/commits/4b24bd2aa39e98a7304fc7d1ffe9acaeb8726a0a).

scripts/ci/exports/utils/module.mjs
martnpaneq: Adressed in [4b24bd2](https://github.com/ckeditor/ckeditor5/pull/19147/commits/4b24bd2aa39e98a7304fc7d1ffe9acaeb8726a0a).

scripts/ci/exports/utils/module.mjs
martnpaneq: Adressed in [4b24bd2](https://github.com/ckeditor/ckeditor5/pull/19147/commits/4b24bd2aa39e98a7304fc7d1ffe9acaeb8726a0a).

scripts/ci/exports/utils/module.mjs
martnpaneq: Adressed in [4b24bd2](https://github.com/ckeditor/ckeditor5/pull/19147/commits/4b24bd2aa39e98a7304fc7d1ffe9acaeb8726a0a).

scripts/ci/exports/utils/module.mjs
martnpaneq: Yeah, we port all paths with `upath`: https://github.com/ckeditor/ckeditor5/blob/5c4e8d30a39e65dc747f7613b8f73d3808f1fad6/scripts/ci/exports/utils/getfilepaths.mjs#L22.

Adressed in: [6976ad4](https://github.com/ckeditor/ckeditor5/pull/19147/commits/6976ad4d501fd2eb3f9af7665e6ec0fae030af58).

PR #19165, opened 2025-09-26
Fix scrolling issue for quick actions when dom root is out of view

l-lejman:
### 🚀 Summary

When dom root element is out of view we should hide ai balloon instead of pinning it to 0,0.

---

### 📌 Related issues

* Closes #8491

---

### 💡 Additional information


scofalik (approved):
1. Please add a changelog entry.

Issue #19172, opened 2025-09-29
Use `minimumReleaseAge` option in pnpm


filipsobol:
See [v10.16](https://pnpm.io/blog/releases/10.16) and [v10.17](https://pnpm.io/blog/releases/10.17).

Using this configuration option can help reduce the risk of installing a compromised npm package. I suggest setting the minimum package age to between 3 and 7 days, since in most cases automated tools (like Snyk) or the community will detect issues within that time.

At the same time, packages such as `ckeditor` and `cksource` should be excluded to avoid problems when bumping our own tools.

```yaml
minimumReleaseAge: 4320 # 3 days
minimumReleaseAgeExclude:
  - '@ckeditor/*'
  - '@cksource/*'
  - '*ckeditor5*' # Should match eslint-config-ckeditor5, stylelint-plugin-ckeditor5-rules, etc.
```

The downside is that everyone on the team will need to update to the latest pnpm version using the following command:

```bash
pnpm self-update
```

PR #19180, opened 2025-09-30
Add `minimumReleaseAge` to protect against malicious packages

filipsobol:
### 🚀 Summary

Add `minimumReleaseAge` to protect against malicious packages.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5/issues/19172.
* See https://github.com/ckeditor/ckeditor5-commercial/issues/8385.
* See https://github.com/ckeditor/ckeditor5-commercial/pull/8542.

---

### 💡 Additional information

N/A

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -18,6 +18,7 @@ minimumReleaseAgeExclude:
   - '@ckeditor/*'
   - '@cksource/*'
   - '*ckeditor5*'
+  - 'umberto'
 
 shellEmulator: true
 shamefullyHoist: true


psmyrek (approved):

pnpm-workspace.yaml
pomek, line 1: I do not feel comfortable looking at these changes. I do not like `dangerouslyAllowAllBuilds=true` in the _production_ mode, but we also miss `umberto` in `minimumReleaseAgeExclude`.

Witoso:
Forgive my ignorance, is this option overriding the lock, or is it only considered when we update the lock. If so, shouldn't the `minimumReleaseAge` be a part of the `lockfile` update procedure, and not used in other cases?

psmyrek (approved):

pomek (approved):

PR #19189, opened 2025-10-03
Docs: Add a CKEditor 5 CDN + MathType setup

gorzelinski:
### 🚀 Summary

As the title suggests, it adds a CKEditor 5 CDN + MathType setup.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/8561

---

### 💡 Additional information

### DIFF
--- a/docs/features/math-equations.md
+++ b/docs/features/math-equations.md
@@ -52,7 +52,7 @@ MathType lets you choose between two editing modes:
 
 If you visit a page using MathType with your mobile device, the handwriting interface will appear by default. However, if you visit the same page with a laptop or desktop computer, the classic input will be displayed. The user is always free to change between the two interfaces.
 
-## Installation
+## Installation from npm
 
 MathType is delivered as a CKEditor&nbsp;5 plugin, so it can be combined into an editor preset just like other features. To add this feature to your editor, install the [`@wiris/mathtype-ckeditor5`](https://www.npmjs.com/package/@wiris/mathtype-ckeditor5) package:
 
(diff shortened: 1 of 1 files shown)


godai78 (approved):

docs/features/math-equations.md
```diff
@@ -70,7 +72,53 @@ ClassicEditor
...
+
+Once the scripts are added, get MathType from the UMD file loaded via the script tag. Use the MathType plugin directly from exports, then add it to the editor.
+
+```js
+const { ClassicEditor } = window.CKEDITOR;
+
+const MathTypeExports = window['@wiris/mathtype-ckeditor5'];
+const MathType = MathTypeExports.default;
+
+ClassicEditor
```
Witoso, line 123: Maybe add a note below on how to attach plugins in setup, like react? Those were a bit different, right?

docs/features/math-equations.md
```diff
@@ -70,7 +72,53 @@ ClassicEditor
...
+			// Configuration.
+		}
+	} )
+	.then( /* ... */ )
+	.catch( /* ... */ );
+```
+
+## Installation from CDN
+
+Our CDN does not provide external packages like MathType. Hence, after attaching the editor scripts, you cannot access the plugin from the global variable like other plugins. However, there is a way to use the CKEditor&nbsp;5 CDN with the MathType plugin. It requires a custom setup.
```
Witoso: Maybe without the first sentence to not diss us at the start :D

godai78: "Because the MathType package is provided by an external partner..."

Let's keep the explanation anyway.

gorzelinski: Sure, makes sense :D

docs/features/math-equations.md
gorzelinski: You mean a note like “hey, the setup may differ depending on the framework you use”? Yeah, in the case of React, we have a hook for loading packages. That may be a good idea.

docs/features/math-equations.md
```diff
@@ -79,6 +81,56 @@ ClassicEditor
...
+│	├── index.js.map
+│	├── index.umd.js
+│	└── index.umd.js.map
+├── ...
+```
+
+Next, attach the MathType script to your page after the editor scripts. This will expose MathType as a global variable, similar to other plugins.
+
+```html
+<script src="./mathtype/index.umd.js"></script>
```
Witoso, line 106: Should CSS be loaded?

gorzelinski: Yeah, yeah, I forgot. Fixed ✅

godai78 (dismissed):

docs/features/math-equations.md
```diff
@@ -70,7 +72,61 @@ ClassicEditor
...
+```
+
+Besides the script, you also need to attach the CSS files.
+
+```html
+<link rel="stylesheet" href="./mathtype/index.css" />
+<link rel="stylesheet" href="./mathtype/index-editor.css" />
+<link rel="stylesheet" href="./mathtype/index-content.css" />
+```
```
Witoso, line 113: I think only an index.css is needed. Editor and content are split into two when you need to use them separately.

gorzelinski: I've just tested it and it seems to work fine - removed

PR #19208, opened 2025-10-09
Fixed restricted imports rule

przemyslaw-zan:
### 🚀 Summary

`@ckeditor/ckeditor5-*` packages other than local and whitelisted packages are forbidden from being imported.

---

### 💡 Additional information

- Script detected dummy invalid imports properly, in source files as well as in snippets.
- Script also ran in related repositories and no errors were thrown.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -16,13 +16,12 @@ const projectPackages = readdirSync( CKEDITOR5_PACKAGES_PATH, { withFileTypes: t
 	.filter( dirent => dirent.isDirectory() )
 	.map( dirent => dirent.name );
 
-const packageWhitelist = [
-	'inspector',
-	'mermaid'
-];
 const allowedPackageNames = [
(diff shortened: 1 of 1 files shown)


eslint.config.mjs
```diff
@@ -10,17 +10,24 @@ import ckeditor5Rules from 'eslint-plugin-ckeditor5-rules';
...
 	.map( dirent => dirent.name );
 
+const packageWhitelist = [
+	'inspector',
+	'mermaid'
+];
+const allowedPackageNames = [
+	...projectPackages.map( p => p.replace( /ckeditor5-?/, '' ) ).filter( Boolean ),
+	...packageWhitelist
+];
```
filipsobol:
```suggestion
const allowedPackageNames = [
	...projectPackages.map( p => p.replace( /ckeditor5-?/, '' ) ).filter( Boolean ),
	'inspector',
	'mermaid'
];
```

This avoids using controversial variable names.

PR #19221, opened 2025-10-13
Remove Vale integration

pomek:
### 🚀 Summary

- Remove the Vale integration.
- Fresh lockfile.

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4132.


pnpm-workspace.yaml
```diff
@@ -13,6 +13,13 @@ onlyBuiltDependencies:
   - puppeteer
   - sharp
 
+# This section must be synchronized with `ckeditor/ckeditor5`.
```
psmyrek:
```suggestion
# This section must be synchronized with `ckeditor/ckeditor5-commercial`.
```

package.json
```diff
@@ -33,11 +33,10 @@
     "@ckeditor/ckeditor5-dev-tests": "^53.1.0",
     "@ckeditor/ckeditor5-dev-web-crawler": "^53.1.0",
     "@ckeditor/ckeditor5-inspector": "^5.0.0",
-    "@ckeditor/ckeditor5-mermaid": "ckeditor/ckeditor5-mermaid#v0.0.5",
+    "@ckeditor/ckeditor5-mermaid": "ckeditor/ckeditor5-mermaid#v0.0.6",
     "@ckeditor/ckeditor5-react": "^9.5.0",
     "@inquirer/prompts": "^7.5.0",
     "@listr2/prompt-adapter-inquirer": "^2.0.16",
-    "@ocular-d/vale-bin": "^2.29.1",
```
psmyrek, line 40: So also `@ocular-d/vale-bin` should be removed from `pnpm-workspace.yaml`.

Also, we should remove `docs:lint` and `docs:vale` scripts since the binary will not be available anymore.

psmyrek (commented):
This PR partially removes Vale. Should we get rid of it completely?

package.json
pomek: If you have [`vale`](https://vale.sh/) installed globally, it would work.

pomek:
> Should we get rid of it completely?

~@psmyrek, to avoid blocking the merge, I would keep it for now because I am unsure. @marburek, could you verify if the team uses it?~

Following up on our internal discussion, we agreed to remove the Vale integration.

cc @godai78

package.json
pomek: We agreed to remove all Vale scripts.

psmyrek (commented):
A few Vale things are left:

- https://github.com/ckeditor/ckeditor5/blob/6445f42b713f7d48cc0980cd016e102b400c752b/packages/ckeditor5-typing/docs/features/text-transformation.md#L25
- https://github.com/ckeditor/ckeditor5/blob/6445f42b713f7d48cc0980cd016e102b400c752b/packages/ckeditor5-typing/docs/features/text-transformation.md#L56
- `./.vale.ini` file.

Issue #19276, opened 2025-10-27
Update styling of schema table


arkflpc:
## 📝 Provide a description of requested docs changes

Update [schema table](https://ckeditor.com/docs/ckeditor5/latest/framework/deep-dive/schema.html#defining-additional-semantics) so it looks like [Feature Comparison table](https://ckeditor.com/docs/ckeditor5/latest/features/converters/import-word/features-comparison.html#collaboration-features).

PR #19287, opened 2025-10-28
Style schema table

gorzelinski:
### 🚀 Summary

This PR styles the Schema table to be more aligned with the new design system.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/19276

---

### 💡 Additional information


gorzelinski:
Icons with superscript numbers looked, well, bad. So, I modified the table. As you look closely, all reference points concern the same elements (object elements). That's why I extracted them into a list and added a description - no need for reference points, as it preserves the information and the table looks clearer. WDYT?

Issue #19038, opened 2025-09-01
CKEditor 4 migration: convert border="0" to border none style


FilipTokarski:
### Steps:
1. Use content from CKEditor 4, where table border was set to 0:
```html
<table border="0" cellpadding="1" cellspacing="1" style="width:500px">
	<tbody>
		<tr>
			<td>a</td>
			<td>b</td>
		</tr>
		<tr>
			<td>c</td>
			<td>d</td>
		</tr>
	</tbody>
</table>
```
<img width="579" height="183" alt="Image" src="https://github.com/user-attachments/assets/55c592aa-0016-4edc-b4ca-3b2a17419b2b" />

### Expected:
Table has all borders set to `none`. Displays only visual helper:

<img width="843" height="225" alt="Image" src="https://github.com/user-attachments/assets/e7d87b50-19bb-47ed-ac14-0bdc411c5f40" />

### Actual:
Borders are set to `double` and on table and `solid` on cells:

<img width="732" height="214" alt="Image" src="https://github.com/user-attachments/assets/a85a465b-3923-4779-b5c8-73eaba6afc43" />

---

If you'd like to see this improvement implemented, add a 👍 reaction to this post.

# Scope

- [x] Implement upcast converter for `border="0"`. It should use current model attributes for borders.
- [x] Check how it works with the content from CKEditor 4.
- [x] Check if it works with layout tables.
- [x] Check that export with inlined styles works.
- [x] Check if the content pasted from MS Word looks the same.
- [x] Discuss if it's a breaking change or not.

Note: we do not need to check other values for `border` than `0`.

Reinmar:
It seems that we simply don't have a converter for `border=0`. It's kept there if GHS allows it, but since it's an attribute it's overridden by a stylesheet.

So this is how CKEditor 5 loads this content:

```html
<figure class="table" style="width:500px;">
    <table border="0" cellpadding="1" cellspacing="1">
        <tbody>
            <tr>
                <td>
                    a
                </td>
                <td>
                    b
                </td>
            </tr>
            <tr>
                <td>
                    c
                </td>
                <td>
                    d
                </td>
            </tr>
        </tbody>
    </table>
</figure>
```

While this is what would work:

```html
<figure class="table" style="width:500px;">
    <table style="border-style:none;" cellpadding="1" cellspacing="1">
        <tbody>
            <tr>
                <td style="border-style:none;">
                    a
                </td>
                <td style="border-style:none;">
                    b
                </td>
[...]

PR #19313, opened 2025-11-03
Add `border="0"` to `border: none` tables converter.

Mati365:
### 🚀 Summary

Added support for converting the legacy HTML `border="0"` attribute to the modern border style representation in CKEditor 5 tables. This improves compatibility with content from CKEditor 4 and other legacy HTML sources.

---

### 📌 Related issues

* Closes  https://github.com/ckeditor/ckeditor5/issues/19038

### DIFF
--- a/packages/ckeditor5-table/src/augmentation.ts
+++ b/packages/ckeditor5-table/src/augmentation.ts
@@ -67,6 +67,22 @@ import type {
 	TableWidthCommand
 } from './index.js';
 
+declare module '@ckeditor/ckeditor5-engine' {
+	interface ExperimentalFlagsConfig {
+
+		/**
+		 * When enabled, the {@link module:table/tableproperties/tablepropertiesediting~TablePropertiesEditing table properties feature}
+		 * and {@link module:table/tablecellproperties/tablecellpropertiesediting~TableCellPropertiesEditing table cell properties feature}
+		 * upcast `border="0"` attributes on `<table>` and `<td>`/`<th>` elements as `tableBorderStyle` and `tableCellBorderStyle` styles
+		 * with value set to `none`.
+		 *
+		 * This is useful when migrating content from the CKEditor 4, which used `border="0"` to represent tables and cells without borders.
+		 * This will be enabled by default in the future CKEditor 5 releases.
+		 */
+		upcastTableBorderZeroAttributes?: boolean;
+	}
+}
+
 declare module '@ckeditor/ckeditor5-core' {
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-table/src/converters/tableproperties.ts
```diff
@@ -219,6 +229,45 @@ export function upcastBorderStyles(
...
+				const viewTable = (
+					viewItem.is( 'element', 'table' ) ?
+						viewItem :
+						viewItem.findAncestor( 'table' )
+				)!;
+
+				// If something already consumed the border attribute on the nearest table element, skip the conversion.
+				if ( !conversionApi.consumable.test( viewTable, { attributes: 'border' } ) ) {
+					return;
+				}
```
niegowski, line 249: How does this affect setting `border:none` on both cells and the table itself? Doesn't it set it only to the first cell?

packages/ckeditor5-table/src/tablelayout/tablelayoutediting.ts
```diff
@@ -247,6 +247,11 @@ function upcastLayoutTable( preferredExternalTableType: TableType | undefined )
 			conversionApi.consumable.consume( viewTable, { attributes: [ 'role' ] } );
 			conversionApi.consumable.consume( viewTable, { classes: [ 'layout-table' ] } );
 
+			// Layout tables have no border, so it's equitable to consume the border="0" attribute.
+			if ( viewTable.getAttribute( 'border' ) === '0' ) {
+				conversionApi.consumable.consume( viewTable, { attributes: [ 'border' ] } );
+			}
```
niegowski: This would consume this attribute even if the experimental flag is disabled. GHS would not handle it anymore.

packages/ckeditor5-table/src/converters/tableproperties.ts
Mati365: It works fine, `upcastBorderStyles` is used for `td` and `th` elements, and consume of `border` attribute is performed after upcasting of table children. Checked there: https://github.com/ckeditor/ckeditor5/pull/19313/files#diff-e478b75c9ed00d13199f5eeab0e5efb93e27edabf2978d6ad6fe04ac25e01fa6R318

niegowski: I'm not sure if this does not work correctly be accident. Conversion fires events for outer-most elements first and later for children of that element so I'm not so sure.

PR #19340, opened 2025-11-05
Strip figure element in clipboard output

### LINKED ISSUE #19319: Strip the figure element in clipboard pipeline

- Strip the figure element.
  - Extract content of Plain table output to table converters.
  - Plain table output converters should always be enabled (data and clipboard pipelines) when `PlainTableOutput` plugin is loaded.
  - Always enable plain table converters in clipboard pipeline (there is a flag in options defining clipboard pipeline).

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

mmotyczynska:
### 🚀 Summary

Strips `figure` element in clipboard output.

---

### 📌 Related issues

* Closes #19319

---

### 💡 Additional information

### DIFF
--- a/packages/ckeditor5-table/src/tableediting.ts
+++ b/packages/ckeditor5-table/src/tableediting.ts
@@ -235,7 +235,7 @@ export class TableEditing extends Plugin {
 	}
 
 	/**
-	 * Adds converters for plain table output. They are used either when the PlainTableOutput plugin is loaded
+	 * Adds converters for plain table output. These converters are used either when the `PlainTableOutput` plugin is loaded
 	 * or when content is processed by the clipboard pipeline, ensuring that pasted tables are not wrapped in a <figure> element.
 	 */
 	private _addPlainTableOutputConverters(): void {


packages/ckeditor5-table/src/tableediting.ts
```diff
@@ -222,6 +233,35 @@ export class TableEditing extends Plugin {
 	public registerAdditionalSlot( slotHandler: TableConversionAdditionalSlot ): void {
 		this._additionalSlots.push( slotHandler );
 	}
+
+	/**
+	 * Adds converters for plain table output. They are used either when the PlainTableOutput plugin is loaded
+	 * or when content is processed by the clipboard pipeline, ensuring that pasted tables are not wrapped in a <figure> element.
```
pszczesniak:
```suggestion
	 * Adds converters for plain table output. These converters are used either when the `PlainTableOutput` plugin is loaded
	 * or when content is processed by the clipboard pipeline, ensuring that pasted tables are not wrapped in a <figure> element.
```

Issue #3225, opened 2018-09-04
Implement table styles (alignment) like image styles


jodator:
The table feature lacks defining table styles similar to the [Image style](https://ckeditor.com/docs/ckeditor5/latest/features/image.html#image-styles) plugin.

The alignment of table cells is a separate feature: ckeditor/ckeditor5#3218.

**NOTE**: Starting from 2020 Q2 it is possible to align tables by using the [table styles panel](https://ckeditor.com/docs/ckeditor5/latest/features/table.html#table-and-cell-styling-tools). This option will work differently than predefined image styles mentioned in this ticket – table styles will be based on inline CSS styles, while image styles use predefined classes (more semantical approach). The semantic approach can still be applied to tables one day, though, so this ticket still makes sense.

- [ ] Re-check and close related issues that are solved by this one.

------

If you'd like to see this feature implemented, add 👍 to this post.

Reinmar:
Just to make it more visible:

**NOTE**: Starting from 2020 Q2 it will be possible to align tables by using the [table styles panel](https://github.com/ckeditor/ckeditor5/issues/3287).

See the note in the first post in this thread.

Reinmar:
@Mgsy, we're not going to work on this. The feature that we're working on in table styles is separate from this.

Reinmar:
It'd be good to review the conflict that arose in the image feature between resize and image styles: https://ckeditor.com/docs/ckeditor5/latest/features/image.html#image-styles. The same conflict will cause our headache in the case of tables. If we want to change something in case of images, it'd be good to make that change before touching tables.

Reinmar:
Media styles (alignment) is tracked in #2781.

Reinmar:
DUP reported: #10289.

bsmilling:
I completely agree that most of the table styles, like colors and widths, should be inline. But I think alignment should be based on classes, as you have multiple CSS properties (float, margin-left/-right) that need to be changed to get it to look good. Float alone doesn't cut it.

Witoso:
Breaking changes notes:

1. We know that this implementation will be a breaking change (UI, placement of buttons).
2. Someone could write a feature that consumes margins, when we introduce this, we will be consuming this first. Let's say this is a BC.

Options:
* let's only add this to v48,
* Backported to v47 
  * A "Glue" plugin for v47 with some custom logic or
  * Separate plugins `TablePropertiesV2` (too complex) or
[...]

PR #19345, opened 2025-11-07
Table block alignment

Mati365:
### 🚀 Summary

This is a collective branch for table alignment changes. We are introducing them first as experimental to the `v47`. They will be later merged to `v48`.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5/issues/3225


packages/ckeditor5-table/src/converters/tableproperties.ts
```diff
@@ -329,3 +332,207 @@ function reduceBoxSidesValue( style?: Style ): undefined | string | Style {
...
+		name: RegExp | string;
+		styles?: Record<string, RegExp | string>;
+		attributes?: Record<string, RegExp | string>;
+		key?: string;
+		value?: RegExp | string;
+	};
+	getAlign: ( ( viewElement: ViewElement ) => string | undefined ) | ( () => string );
+	getConsumables: ( viewElement: ViewElement ) => Consumables;
+};
+
```
Mati365, line 587: @mmotyczynska / @pszczesniak I'm trying to understand the reasoning behind this configuration-based approach. Could you explain what specific problem this is solving? I'd be grateful for any context on why this direction was chosen over regular converter functions. Thanks!

pszczesniak: Thanks for the question — happy to give more context.

The main reason behind moving to a configuration-based approach was to eliminate a growing amount of duplicated converter code. Most of the converters were following exactly the same logic and differed only in a few parameters. Instead of maintaining many nearly identical functions, the shared behavior is now expressed once, while the varying parts are provided as data.

This approach has a few benefits:

- DRY: we avoid repeating the same control flow across multiple converters.

- Data-driven design: all differences between converters are captured in a declarative config object, making the code easier to read, extend, and validate.

- Easier maintenance: adding or modifying a converter now means updating configuration rather than creating another boilerplate function.

- Reduced risk of inconsistencies: since the logic lives in one place.

Regular standalone converter functions worked fine at first, but as their number grew, maintaining them became repetitive and error-prone. Consolidating the shared logic and driving the differences through configuration made the codebase more scalable and easier to reason about.

Witoso (approved):

PR #19354, opened 2025-11-10
Introducing an automatic command for creating restricted editing exceptions (both block and inline).

### LINKED ISSUE #19353: Restricted editing exceptions should auto-choose either block or inline exception

The restricted editing in the standard editing mode provides separate buttons for inline and block exceptions. This might be confusing for users, as it might be unclear what that means. We should prepare a single button for creating both types of exceptions.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

---

niegowski:
### 🚀 Summary

Introducing an automatic command for creating restricted editing exceptions (both block and inline).

---

### 📌 Related issues

* Closes #19353

---

### 💡 Additional information

### DIFF
--- a/packages/ckeditor5-restricted-editing/docs/features/restricted-editing.md
+++ b/packages/ckeditor5-restricted-editing/docs/features/restricted-editing.md
@@ -66,7 +66,7 @@ After {@link getting-started/integrations-cdn/quick-start installing the editor}
 
 ### Running the standard editing mode
 
-To initialize the editor in the standard editing mode, add the {@link module:restricted-editing/standardeditingmode~StandardEditingMode} plugin and add the `'restrictedEditingException:dropdown'` button to the toolbar:
+To initialize the editor in the standard editing mode, add the {@link module:restricted-editing/standardeditingmode~StandardEditingMode} plugin and add the `'restrictedEditingException:auto'` button to the toolbar:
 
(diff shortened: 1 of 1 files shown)


spacelist-ca:
We started mocking the same method in a custom plugin - would be great to see this offered off the shelf. It should detect all kinds of highlighted placement, including tables (cells and entire table, rows, columns, etc), images, numbering/lists, mergefields, etc. Supporting Blocks is a huge improvement!

Witoso:
> highlighted placement, including tables (cells and entire table, rows, columns, etc), images, numbering/lists, mergefields, etc.

I think it's safe to say this PR won't do that in such a big scope, as more advanced cases would need a deeper rewrite.

spacelist-ca:
If they're either blocks or not blocks though, shouldn't that work regardless of the type of block? We've tested a bunch of these scenarios with the Block setup, and we've made the button inline so when you highlight text/block it shows a balloon with the two buttons so you can select, and the idea is depending on the type of content you selected/highlighted it'll restrict the options to show inline or block accordingly.

Images and tables seems to be the most common use case though, but numbered lists would be helpful to include in automatic detection as well if you highlighted multiple list items to enable editing it should let you add more to the numbered list...as opposed to hgihlighting each individual item to be editable.

PR #19357, opened 2025-11-12
Add allowed description block attributes to todo list editing.

Mati365:
### 🚀 Summary

Add ability to specify attributes that work fine with to-do list description.

---

### 📌 Related issues

* Caused by https://github.com/ckeditor/ckeditor5-commercial/issues/8804


gorzelinski (approved):
For me, it looks ok. But I'm unsure how we ultimately want to handle this. It's a strange case. I struggled with it on support.

Issue #19318, opened 2025-11-04
Table alignment - clipboard


mmotyczynska:
## Provide a description of the task

- Clipboard output: use a mix (classes + inline + align) for compatibility.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …
* Installed CKEditor plugins: …

pszczesniak:
Solved in https://github.com/ckeditor/ckeditor5/pull/19368.

PR #19368, opened 2025-11-13
Use mix of all attributes responsible for table alignment in clipboard output.

pszczesniak:
### 🚀 Summary

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/19318

---

### 💡 Additional information


pszczesniak:
Thanks @mmotyczynska for the comment — I’ve actually been thinking about this the whole time. I tried to quickly change that implementation using a downcast in `TablePropertiesEditing`, but unfortunately `attributeToAttribute` can’t return more than one key value, e.g. either `style` or `class`. On top of that, we’d have to repeat it there for `left`, `right`, `blockLeft`, etc.

It’s possible that when we revisit [the clipboard task](https://github.com/ckeditor/ckeditor5/issues/19369), something interesting will come up that will let us refactor this code.

PR #19378, opened 2025-11-17
Using new CI scripts config with expilit arguments instead of environment variables.

martnpaneq:
### 🚀 Summary

Using new CI scripts config with expilit arguments instead of environment variables.

---

### 💡 Additional information

### DIFF
--- a/.circleci/template.yml
+++ b/.circleci/template.yml
@@ -60,7 +60,6 @@ commands:
           environment:
             PUPPETEER_SKIP_DOWNLOAD: true
           command: pnpm install --frozen-lockfile
-      - prepare_environment_variables_commands
 
   checkout_command:
     description: "Clone the CKEditor 5 repository with limited depth and branches"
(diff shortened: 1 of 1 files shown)


.circleci/template.yml
```diff
@@ -216,9 +216,7 @@ jobs:
             fi
       - run:
           name: Trigger CKEditor 5 release job
-          environment:
-            CKE5_GITHUB_RELEASE_BRANCH: << pipeline.git.branch >>
-          command: pnpm ckeditor5-dev-ci-trigger-circle-build
+          command: pnpm ckeditor5-dev-ci-trigger-circle-build --release-branch << pipeline.git.branch >>
```
martnpaneq: In other projects `slug` is provided to the `trigger-circle-build` script. But here it was not provided before, using an environmnet variable.

Need to double check.

martnpaneq: `slug` was added, because it was being set at the `Prepare environment variables` command level.

.circleci/template.yml
```diff
@@ -216,9 +203,7 @@ jobs:
             fi
       - run:
           name: Trigger CKEditor 5 release job
-          environment:
-            CKE5_GITHUB_RELEASE_BRANCH: << pipeline.git.branch >>
-          command: pnpm ckeditor5-dev-ci-trigger-circle-build
+          command: pnpm ckeditor5-dev-ci-trigger-circle-build --slug ckeditor/ckeditor5 --release-branch << pipeline.git.branch >>
```
psmyrek, line 206: `--slug` is not valid and `--trigger-repository-slug` is missing.

Also please note that there is another `pnpm ckeditor5-dev-ci-trigger-circle-build` call below that does not have any parameters (so it will crash).

pomek (approved):
LGTM.

.circleci/template.yml
martnpaneq: Addressed in https://github.com/ckeditor/ckeditor5/pull/19385.

Issue #19369, opened 2025-11-13
[Nice to have] Support table block alignment in clipboard output.


pszczesniak:
## Provide a description of the task

As we wanted to use [mix of properties for compability](https://github.com/ckeditor/ckeditor5/issues/19318), we would also like to support block alignemnt so pasting content from editor will look the same (aligmnet) for example in MS Word.

To achieve it we must wrap a table into a div with `align` atribute.

[Here is a PoC](https://github.com/ckeditor/ckeditor5/pull/19359) of how we could do it by manipulating table element in clipboard pipeline.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …
* Installed CKEditor plugins: …

pszczesniak:
Solved in https://github.com/ckeditor/ckeditor5/pull/19388.

PR #19388, opened 2025-11-18
Extended support of table alignment in clipboard output.

pszczesniak:
### 🚀 Summary

*Upcast tables wrpapped with div with align attribute and wrapping a table in a div with align attribute in the clipboard content while copy from editor.*

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/19369

---

### 💡 Additional information

### DIFF
--- a/packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
+++ b/packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
@@ -453,10 +453,7 @@ function convertToTableAlignment( divAlign: string, tableAlign: string, defaultV
 			case 'center':
 				return 'center';
 			case 'left':
-				if ( tableAlign === undefined ) {
-					return 'blockLeft';
-				}
-				break;
+				return tableAlign === undefined ? 'blockLeft' : 'left';
 			default:
 				return defaultValue;
 		}
--- a/packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
+++ b/packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
@@ -2949,6 +2949,21 @@ describe( 'table properties', () => {
 				expect( table.getAttribute( 'tableAlignment' ) ).to.equal( 'blockLeft' );
 			} );
 
+			it( 'should upcast align `left` attribute from div wrapped on table on table with align `left` to `left`', () => {
+				editor.setData(
+					'<div align="left">' +
+						'<table align="left">' +
+							'<tr>' +
+								'<td>foo</td>' +
+							'</tr>' +
+						'</table>' +
+					'</div>'
+				);
+				const table = model.document.getRoot().getNodeByPath( [ 0 ] );
+
+				expect( table.getAttribute( 'tableAlignment' ) ).to.equal( 'left' );
+			} );
+
 			it( 'should upcast align `right` attribute from div wrapped on table with align `right`', () => {
 				editor.setData(
 					'<div align="right">' +


packages/ckeditor5-table/src/utils/structure.ts
```diff
@@ -566,3 +566,16 @@ export function adjustLastColumnIndex(
...
+ */
+export function getViewTableFromWrapper( figureView: ViewElement ): ViewElement | undefined {
+	for ( const figureChild of figureView.getChildren() ) {
+		if ( figureChild.is( 'element', 'table' ) ) {
+			return figureChild;
+		}
+	}
+
+	return undefined;
+}
```
mmotyczynska: I believe the explicit `return undefined` isn't necessary here - the function will return `undefined` if no table is found. The TS return type can stay as it is.
Additionally, since the original function name was updated (`figure` → `wrapper`), it might be worth renaming related variables as well (e.g., `figureView` → `wrapperView`) for consistency.

packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
```diff
@@ -156,7 +164,50 @@ export class TablePropertiesEditing extends Plugin {
 			'tableBackgroundColor',
 			new TableBackgroundColorCommand( editor, defaultTableProperties.backgroundColor )
 		);
+
+		// Adjust clipboard output to wrap tables in divs if needed (for alignment).
+		const viewDoc = editor.editing.view.document;
+		this.listenTo<ViewDocumentClipboardOutputEvent>( viewDoc, 'clipboardOutput', ( evt, data ) => {
```
mmotyczynska, line 171:
```suggestion
		const viewDoc = editor.editing.view.document;
		
		this.listenTo<ViewDocumentClipboardOutputEvent>( viewDoc, 'clipboardOutput', ( evt, data ) => {
```

packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
```diff
@@ -156,7 +164,50 @@ export class TablePropertiesEditing extends Plugin {
...
+			editor.editing.view.change( writer => {
+				const children = Array.from( data.content.getChildren() ).map( item => wrapInDivIfNeeded( item, writer ) );
+				const documentFragment = writer.createDocumentFragment( children );
```
mmotyczynska: Because this may run on rather large documents, I have some concerns about performance here. We end up creating two arrays, and even if none of the children require wrapping, we still create a new document fragment from a copied list of the original nodes. That work could potentially be avoided. Not sure if it's a big issue in practice, just something to consider 🤔

packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
```diff
@@ -156,7 +164,50 @@ export class TablePropertiesEditing extends Plugin {
...
+			} );
+		}, { priority: 'lowest' } );
+	}
+}
+
+/**
+ * Checks whether the view node is a table and if it needs to be wrapped in a div for alignment purposes.
+ * If so, it wraps it in a div and returns the new element. Otherwise, it returns the original view node.
+ */
+function wrapInDivIfNeeded( ViewNode: ViewNode, writer: ViewDowncastWriter ): ViewNode {
```
mmotyczynska:
```suggestion
function wrapInDivIfNeeded( viewNode: ViewNode, writer: ViewDowncastWriter ): ViewNode {
```

packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
```diff
@@ -2897,6 +2903,421 @@ describe( 'table properties', () => {
 			} );
 		} );
 
+		describe( 'upcast table wrapped by div', () => {
+			it( 'should upcast align `left` attribute from div wrapped on table', () => {
+				editor.setData(
+					'<div align="right">' +
```
mmotyczynska: `left` or `right`? The description does not match the test 😉

packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
```diff
@@ -295,6 +346,108 @@ function enableAlignmentProperty( schema: ModelSchema, conversion: Conversion, d
...
+			case 'right':
+				if ( tableAlign === 'right' ) {
+					return 'right';
+				} else {
+					return 'blockRight';
+				}
+			case 'center':
+				return 'center';
+			default:
+				return defaultValue;
```
pszczesniak, line 458: Add fallback for `div` with `align="left"` to support `blockLeft` alignment (support for for example MS Word in RTL mode when table is left aligned).

packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
```diff
@@ -156,7 +164,51 @@ export class TablePropertiesEditing extends Plugin {
 			'tableBackgroundColor',
 			new TableBackgroundColorCommand( editor, defaultTableProperties.backgroundColor )
 		);
+
+		const viewDoc = editor.editing.view.document;
+
+		// Adjust clipboard output to wrap tables in divs if needed (for alignment).
+		this.listenTo<ViewDocumentClipboardOutputEvent>( viewDoc, 'clipboardOutput', ( evt, data ) => {
+			editor.editing.view.change( writer => {
+				const children = Array.from( data.content.getChildren() ).map( item => wrapInDivIfNeeded( item, writer ) );
```
pszczesniak: It only takes into account the children without going deeper.

packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
pszczesniak: Thank you for your comment :+1: Now i've rewrite it to support table nesting.

packages/ckeditor5-table/src/tableproperties/tablepropertiesediting.ts
pszczesniak: Done.

mmotyczynska (approved):
👌

PR #19406, opened 2025-11-21
Introduced Sentry configuration in the docs

psmyrek:
### 🚀 Summary

Introduced Sentry configuration in the docs.

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

Requires https://github.com/cksource/umberto/pull/1359.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "svgo": "^3.3.2",
     "table": "^6.8.1",
     "typescript": "5.3.3",
-    "umberto": "^9.0.0",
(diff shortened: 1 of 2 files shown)


filipsobol:
Shouldn't this PR target the `stable` branch?

psmyrek:
> Shouldn't this PR target the `stable` branch?

Probably not I guess, as this is a new feature that also requires bumping `umberto` version.

Moreover, after merging to `master` instead of `stable`, we will be able to observe for some time whether it works in nightly docs before it goes to production.

PR #19412, opened 2025-11-24
Add support for block table alignment in Paste from Office in Safari.

mmotyczynska:
### 🚀 Summary

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

### DIFF
--- a/packages/ckeditor5-paste-from-office/src/filters/table.ts
+++ b/packages/ckeditor5-paste-from-office/src/filters/table.ts
@@ -47,10 +47,9 @@ export function transformTables(
 		// so the surrounding `<div>` with the `align` attribute may end up being the table's grandparent.
 
 		if ( hasTablePropertiesPlugin && item.is( 'element', 'table' ) ) {
-			const directParent = item.parent && item.parent.is( 'element', 'div' ) ? item.parent : null;
-			const grandParent = item.parent && item.parent.parent && item.parent.parent.is( 'element', 'div' ) ?
-				item.parent.parent : null;
-			const divParent = directParent || grandParent;
+			const directParent = item.parent?.is( 'element', 'div' ) ? item.parent : null;
+			const grandParent = item.parent?.parent?.is( 'element', 'div' ) ? item.parent.parent : null;
+			const divParent = directParent ?? grandParent;
 
 			// Center block table alignment.
 			if ( divParent && divParent.getAttribute( 'align' ) === 'center' && !item.getAttribute( 'align' ) ) {


packages/ckeditor5-paste-from-office/src/filters/table.ts
```diff
@@ -35,13 +35,35 @@ export function transformTables(
...
+		// In MS Word, left-aligned tables (default) have no align attribute on the table and are not wrapped in a div.
+		// In such cases, we need to set `margin-left: 0` and `margin-right: auto` to indicate to the editor that
+		// the table is block-aligned to the left.
+		//
+		// Center- and right-aligned tables in MS Word are wrapped in a div with the `align` attribute set to
+		// `center` or `right`, respectively with no align attribute on the table itself.
+		//
+		// Additionally, the structure may change when pasting content from MS Word.
+		// Some browsers (e.g., Safari) may insert extra elements around the table (e.g., a <span>),
+		// so the surrounding div with the `align` attribute may end up being the table's grandparent.
```
pszczesniak:
```suggestion
		// In MS Word, left-aligned tables (default) have no align attribute on the `<table>` and are not wrapped in a `<div>`.
		// In such cases, we need to set `margin-left: 0` and `margin-right: auto` to indicate to the editor that
		// the table is block-aligned to the left.
		//
		// Center- and right-aligned tables in MS Word are wrapped in a `<div>` with the `align` attribute set to
		// `center` or `right`, respectively with no align attribute on the `<table>` itself.
		//
		// Additionally, the structure may change when pasting content from MS Word.
		// Some browsers (e.g., Safari) may insert extra elements around the table (e.g., a <span>),
		// so the surrounding `<div>` with the `align` attribute may end up being the table's grandparent.
```

packages/ckeditor5-paste-from-office/src/filters/table.ts
```diff
@@ -35,13 +35,35 @@ export function transformTables(
...
+
 		if ( hasTablePropertiesPlugin && item.is( 'element', 'table' ) ) {
-			const hasDivParent = item.parent && item.parent.is( 'element', 'div' );
+			const directParent = item.parent && item.parent.is( 'element', 'div' ) ? item.parent : null;
+			const grandParent = item.parent && item.parent.parent && item.parent.parent.is( 'element', 'div' ) ?
```
pszczesniak: Maybe we could use here the optional chaining? 🤔

mmotyczynska: Good idea, thanks 👍

Issue #19407, opened 2025-11-21
Watchdog restarts all editors on a page instead of only the one with error


FilipTokarski:
## 📝 Provide detailed reproduction steps (if any)

1. Go to `watchdog.html` manual test
2. Follow instructions to crash editor 1

### ✔️ Expected result

Only editor 1 gets restarted.

### ❌ Actual result

Both editors get restarted.

## 📃 Other details

* Browser: …
* OS: …
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

PR #19418, opened 2025-11-25
Watchdog no longer restarts other editors that might share some configuration or plugins but were not affected by the error.

Mati365:
### 🚀 Summary

Watchdog no longer restarts other editors that might share some configuration or plugins but were not affected by the error.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/19407

### DIFF
--- a/packages/ckeditor5-watchdog/src/editorwatchdog.ts
+++ b/packages/ckeditor5-watchdog/src/editorwatchdog.ts
@@ -679,6 +679,8 @@ export type EditorWatchdogCreatorFunction<TEditor = Editor> = (
  */
 function getEditorSpecificEntries( editor: Editor ) {
 	return [
+		editor.locale,
+		editor.accessibility,
 		editor.data,
 		editor.plugins,
 		editor.config,


packages/ckeditor5-watchdog/src/editorwatchdog.ts
```diff
@@ -628,3 +673,22 @@ export type EditorWatchdogCreatorFunction<TEditor = Editor> = (
...
+		editor.config,
+		editor.conversion,
+		editor.commands,
+		editor.plugins,
+		editor.model.document,
+		editor.model,
+		editor.model.schema,
+		editor.ui?.view,
+		editor.editing?.view
+	].filter( Boolean );
```
niegowski, line 699: How did you select this list of objects? Maybe we should add some more like `editor.editing`, `editor.ui`, `editor.editing.mapper`?

niegowski (changes requested):
I'm not sure about this solution. Maybe we should discuss this in a larger group.

packages/ckeditor5-watchdog/src/editorwatchdog.ts
Mati365: I selected these fields because they’re the core editor properties that are always present and each of them is an instance of a CKEditor 5 specific class. In other words, this list isn’t random - it includes only the stable subsystems that exist in practically every editor, regardless of the loaded features.

Issue #19431, opened 2025-11-26
`thead` should not be gready and should not move all th to the `thead`


Mati365:
## 📝 Provide detailed reproduction steps (if any)

Spotted in: https://github.com/ckeditor/ckeditor5/issues/14911
Reproducible here: http://localhost:8125/ckeditor5/external/ckeditor5/packages/ckeditor5-table/tests/manual/tickets/14911.html

### ✔️ Expected result

`Rowgroup Header 2`  should be placed below `Row Header 1`.

### ❌ Actual result

It's moved to `thead`.

<img width="1861" height="862" alt="Image" src="https://github.com/user-attachments/assets/31dec9c0-d8f3-40d5-a508-7f2d875a74b3" />

PR #19448, opened 2025-11-26
Fix incorrect move of table row containing only `th` elements to `thead` if previous row was not header row (v47.x)

Mati365:
### 🚀 Summary

Fix incorrect move of table row containing only `th` elements to `thead` if previous row was not header row.
It's change for `v47` version of the editor.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/19431

### DIFF
--- a/packages/ckeditor5-table/src/converters/upcasttable.ts
+++ b/packages/ckeditor5-table/src/converters/upcasttable.ts
@@ -8,6 +8,7 @@
  */
 
 import type { ModelElement, UpcastDispatcher, UpcastElementEvent, ViewElement, ViewNode } from 'ckeditor5/src/engine.js';
+import type { Editor } from 'ckeditor5/src/core.js';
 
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-table/src/converters/upcasttable.ts
```diff
@@ -81,7 +82,9 @@ export function upcastTableFigure() {
  * @returns Conversion helper.
  * @internal
  */
-export function upcastTable() {
+export function upcastTable( editor: Editor ) {
```
pszczesniak, line 85: Maybe would be good to pass only the config flag value with default set to false instead of editor instance?

This function Is re-exported as an internal function but maybe we can add a fallback for the parameter.

pszczesniak: Ok, it will be removed in v48.

pszczesniak (approved):

PR #19459, opened 2025-12-01
Set default table header alignment to left in the output data.

### LINKED ISSUE #19454: `th` elements not explicitly left-aligned in output, causing inconsistent rendering.

1. Create a table in the editor containing at least one `<th>` cell.
2. Export or render the table using output data (e.g. on a page outside the editor).
3. Observe the text alignment inside `<th>` elements in the output.

### ✔️ Expected result

Content inside `<th>` elements should be explicitly left-aligned in the output, matching how it appears inside the editor.

### ❌ Actual result

The alignment is not explicitly applied in the output data.  
As a result, `<th>` content becomes center-aligned in environments where no default override is present, causing visual inconsistencies compared to the editor.

Image

### ❓ Possible solution

Append additional style:

```css
.ck-content .table th {
    text-align: start;
}
```

While these styles are no longer needed, we'll keep them for backward compatibility: https://github.com/ckeditor/ckeditor5/blob/bf0b902a89274f840bd4fc4c0df71bc186d35a86/packages/ckeditor5-table/theme/table.css#L9-L15

Related issue: https://github.com/ckeditor/ckeditor5/issues/6638

Mati365:
### 🚀 Summary

Set default table header alignment to left in the output data.

---

### 📌 Related issues

* Closes #19454

### DIFF
--- a/packages/ckeditor5-table/theme/table.css
+++ b/packages/ckeditor5-table/theme/table.css
@@ -10,7 +10,7 @@
 	}
 
 	/* Text alignment of the table header should match the editor settings and override the native browser styling,
-	when content is available outside the editor. See https://github.com/ckeditor/ckeditor5/issues/6638 */
+	when content is available outside the editor. See https://github.com/ckeditor/ckeditor5/issues/6638. */
 	&[dir="rtl"] .table th {
 		text-align: right;
 	}


packages/ckeditor5-table/theme/table.css
```diff
@@ -4,6 +4,11 @@
  */
 
 .ck-content {
+	/* Ensure that table header text is aligned to the start by default. See https://github.com/ckeditor/ckeditor5/issues/19454 */
```
pszczesniak:
```suggestion
	/* Ensure that table header text is aligned to the start by default. See https://github.com/ckeditor/ckeditor5/issues/19454. */
```

packages/ckeditor5-table/theme/table.css
```diff
@@ -4,6 +4,11 @@
...
+	& .table th {
+		text-align: start;
+	}
+
 	/* Text alignment of the table header should match the editor settings and override the native browser styling,
 	when content is available outside the editor. See https://github.com/ckeditor/ckeditor5/issues/6638 */
```
pszczesniak:
```suggestion
	when content is available outside the editor. See https://github.com/ckeditor/ckeditor5/issues/6638. */
```

pszczesniak (commented):
Just two suggestions (in comments).

Other than that :+1:

PR #19461, opened 2025-12-01
Handle title element attributes by redirecting to title-content element

DawidKossowski:
### 🚀 Summary

The `title` element in the model doesn't have a corresponding view element. Only its child `title-content` is converted to `<h1>` in the view. When attributes are set on the `title` element, the downcast conversion fails because there's no view element to apply them to.

Added a downcast converter that intercepts attribute changes on `title` elements and redirects them to the `title-content` corresponding view element. This ensures that attributes set on `title` in the model are correctly rendered in the view without causing conversion errors.

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

These changes are required for the AI feature, as it handles elements by adding the `$elementId` attribute to each one.

### DIFF
--- a/packages/ckeditor5-engine/src/conversion/downcastdispatcher.ts
+++ b/packages/ckeditor5-engine/src/conversion/downcastdispatcher.ts
@@ -916,9 +916,19 @@ function shouldMarkerChangeBeConverted(
 
 	const hasCustomHandling = ( ancestors as Array<ModelElement> ).some( element => {
 		if ( range.containsItem( element ) ) {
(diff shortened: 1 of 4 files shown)


scofalik (approved):
LGTM.

I dislike that we need to make a change in `downcastdispatcher.ts` because title feature is incorrectly written. But it is what it is. Title feature breaks the rule that every model element should be mapped to a counterpart in the view. The feature should be fixed. I will add this as a followup ticket.

scofalik:
https://github.com/ckeditor/ckeditor5/issues/19484

PR #19508, opened 2025-12-12
Throw an error when an editor config is not plain object.

### LINKED ISSUE #18072: Throw an error when an editor class is included in the plugin list

If an editor class (e.g., `ClassicEditor`) is mistakenly included in the plugins list when initializing `Editor`, it causes a silent misconfiguration leading to license issues (throws `license-key-missing`). `Editor` should explicitly detect this case and throw an informative error instead.

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

---

arkflpc:
### 🚀 Summary

Throws an error when editor config is not a plain JS object. This happens when there is an editor class (like `ClassicEditor`) in `config.plugins`. In that case, the error thrown suggests a license problem. The purpose of this change is to provide better error in that case. Additionally, `PluginCollection.init()` is refactored to consistently return a rejected promise in case of failure. Before it returned a rejected promise or threw an exception synchronously.

---

### 📌 Related issues

* Closes #18072.

---

### 💡 Additional information

### DIFF
--- a/.changelog/20251212123121_ck_18072_plain_object_config.md
+++ b/.changelog/20251212123121_ck_18072_plain_object_config.md
@@ -1,0 +1,13 @@
+---
+type: Other
+
+scope:
+  - ckeditor5-core
+
+closes:
+  - https://github.com/ckeditor/ckeditor5/issues/18072
+---
+
+The `Editor` class constructor now detects if the provided `EditorConfig` is not an object.
+
+The common source of this error is when an editor class (e.g., `ClassicEditor`) is mistakenly included in the plugins list when initializing Editor.


packages/ckeditor5-core/src/editor/editor.ts
```diff
@@ -306,6 +306,20 @@ export abstract class Editor extends /* #__PURE__ */ ObservableMixin() {
 	constructor( config: EditorConfig = {} ) {
 		super();
 
+		if ( !isPlainObject( config ) ) {
+			/**
+			 * Editor configuration must be a plain JavaScript object.
```
Reinmar, line 311: Not true, at least not in the way how I understand this. POJO for me isn't just about top-level but also other levels. And we have e.g. also callbacks or elements in our config if we look deeper.

So, this error message may be confusing, IMO.

niegowski: Does `isPlainObject()` verify argument recursively?

Mati365: @niegowski nope https://github.com/toss/es-toolkit/blob/main/src/predicate/isPlainObject.ts

PR #19513, opened 2025-12-12
Dropped the `src/` directory from published packages

przemyslaw-zan:
### 🚀 Summary

Dropped the `src/` directory from published packages.

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4331,9 +4331,6 @@ importers:
       '@ckeditor/ckeditor5-engine':
         specifier: workspace:*
         version: link:../ckeditor5-engine
-      typescript:
-        specifier: 5.3.3
-        version: 5.3.3
 
   packages/ckeditor5-watchdog:
     dependencies:


przemyslaw-zan:
TBD: What to do with the [update-icons-typings.mjs](https://github.com/ckeditor/ckeditor5/blob/5c6a87c81210fd2f2de5bc93ce705dc3e92a5a88/scripts/update-icons-typings.mjs) script? Here is the current output format:

### `index.js`
image

### `index.d.ts`
image

---

### Updated `index.d.ts`
image

packages/ckeditor5-adapter-ckfinder/package.json
psmyrek, line 1: As the `src` directory is no longer published, we should remove it from the `files`:

https://github.com/ckeditor/ckeditor5/blob/2ab69d481192976232794cbf09deceb6e61befd3/packages/ckeditor5-adapter-ckfinder/package.json#L49-L50

This should be done for all packages.

scripts/release/utils/updatepackageentrypoint.mjs
```diff
@@ -47,20 +48,9 @@ export default async function updatePackageEntryPoint( packagePath ) {
...
 			types: './' + types,
-			import: './dist/*',
-			default: './dist/*'
-		},
-		'./src/*': {
-			types: './src/*.d.ts',
-			import: './src/*',
-			default: './src/*'
+			import: './' + main,
+			default: './' + main
```
psmyrek: Not sure, but probably we should keep the existing paths targeting `./dist/*` for `import` and `default` keys.

przemyslaw-zan: @filipsobol How should this output look like?

filipsobol: @psmyrek is right. The `import` and `default` keys inside the `./dist/*'` should be left unchanged. Only the comment should be removed and the `types` field updated to target the `dist` instead of the `src` directory.

przemyslaw-zan: Example package output - please confirm that this is as expected:

```JSON
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/index.js",
      "default": "./dist/index.js"
    },
    "./dist/*": {
      "types": "./dist/index.d.ts",
      "import": "./dist/*",
      "default": "./dist/*"
    },
    "./ckeditor5-metadata.json": "./ckeditor5-metadata.json",
    "./package.json": "./package.json"
  },
```

packages/ckeditor5-utils/package.json
filipsobol, line 22: Can we remove `typescript` from `devDependencies` the same as in other packages?

scripts/release/utils/updatepackageentrypoint.mjs
filipsobol: Look fine :+1:

PR #19528, opened 2025-12-16
Fix error thrown when creating code block via backticks on some keyboard layouts.

### LINKED ISSUE #18926: Creating a codeblock via backticks on some keyboard layouts causes TypeError: Cannot read properties of null (reading 'parent')

1. On Windows, set up the US International keyboard layout (not the standard US one).
2. Go to https://ckeditor.com/ckeditor-5/demo/feature-rich/
3. Type three backticks to open a code block (```).
4. Observe the logs for an error.

### ✔️ Expected result

Inserting the code block via the auto-formatting should not cause any exceptions.

### ❌ Actual result

An unexpected error is triggered:

```
ck-demo-external-fea…293eec72de.js:15999 
 Uncaught CKEditorError: unexpected-error
Read more: https://ckeditor.com/docs/ckeditor5/latest/support/error-codes.html#error-unexpected-error
Original error: TypeError: Cannot read properties of null (reading 'parent')
    at hg.findMappedViewAncestor (ck-demo-external-fea…72de.js:16002:27714)
    at hg.<anonymous> (ck-demo-external-fea…72de.js:16055:41843)
    at hg.fire (ck-demo-external-fea…c72de.js:16000:1669)
    at hg.toModelPosition (ck-demo-external-fea…72de.js:16002:27163)
    at hg.toModelRange (ck-demo-external-fea…72de.js:16002:26957)
    at ck-demo-external-fea…2de.js:16020:104647
    at Array.map (<anonymous>)
    at Object.<anonymous> (ck-demo-external-fea…2de.js:16020:104638)
    at Object.fire (ck-demo-external-fea…c72de.js:16000:1669)
    at So (ck-demo-external-fea…72de.js:16000:80780)
```

## 📃 Other details

* Browser: Edge
* OS: Windows 11
* First affected CKEditor version: 46.0.0
* Installed CKEditor plugins: Happens in vanilla too.

---

Mati365:
### 🚀 Summary

Fix error thrown when creating code block via backticks on some keyboard layouts.

---

### 📌 Related issues

* Closes #18926 

---

### 💡 Additional information

1. The first `insertText` inserts the backtick character. This triggers the code block feature, which detects the markdown syntax (e.g., `code`) and replaces the current paragraph with a `codeBlock`.
2. The second `insertText` event arrives shortly after. However, its `viewSelection` still points to the original paragraph element, which is now detached from the document root. 

The fix ensures that we ignore such ranges by checking if the range's root is a valid.
This behaviour is not reproducible when CKEditor 5 Inspector or Devtools debugger is attached.

### DIFF
--- a/packages/ckeditor5-typing/src/input.ts
+++ b/packages/ckeditor5-typing/src/input.ts
@@ -128,7 +128,17 @@ export class Input extends Plugin {
 			// If view selection was specified, translate it to model selection.
 			if ( viewSelection ) {
 				modelRanges = Array.from( viewSelection.getRanges() )
-					.filter( viewRange => viewRange.root.is( 'rootElement' ) )
+					.filter( viewRange => {
+						// On Windows 11 with the US International keyboard, events are batched.
+						// In other words, the first backtick press does not send an `insertText` event,
+						// and only the second one sends it double (batched).
+						// This causes a race condition if during the first backtick insert the element is removed from the tree,
+						// and the second event flushes changes, and it's original targetRanges,
+						// which were initially good, now refer to a removed element.
+						// This is not reproducible on Mac/Linux as they enter composition mode.
+						// See more: https://github.com/ckeditor/ckeditor5/issues/18926.
+						return viewRange.root.is( 'rootElement' );
+					} )
 					.map( viewRange => mapper.toModelRange( viewRange ) )
 					.map( modelRange => _tryFixingModelRange( modelRange, model.schema ) || modelRange );
 			}


niegowski (changes requested):
I believe the fix should be located in the `InputObserver` as target ranges are validated there and also used in other cases (delete, enter, etc).

https://github.com/ckeditor/ckeditor5/blob/25f6fd457c6fe44304c71760423a2920e280ce5f/packages/ckeditor5-engine/src/view/observer/inputobserver.ts#L94-L100

PR #19536, opened 2025-12-18
Enabled reading ckbox version from umberto config

przemyslaw-zan:
### 🚀 Summary

Enabled reading ckbox version from umberto config.

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

### DIFF
--- a/scripts/docs/snippetadapter.mjs
+++ b/scripts/docs/snippetadapter.mjs
@@ -8,6 +8,9 @@ import { constants, readFile, writeFile, copyFile, access, mkdir } from 'node:fs
 import upath from 'upath';
 import { build as esbuild } from 'esbuild';
 import { CKEDITOR5_COMMERCIAL_PATH, CKEDITOR5_ROOT_PATH } from '../constants.mjs';
+import umbertoJson from '../../docs/umberto.json' with { type: 'json' };
+
+const { CKBOX_VERSION } = umbertoJson.variables;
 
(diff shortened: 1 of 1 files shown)


scripts/docs/snippetadapter.mjs
```diff
@@ -8,6 +8,9 @@ import { constants, readFile, writeFile, copyFile, access, mkdir } from 'node:fs
 import upath from 'upath';
 import { build as esbuild } from 'esbuild';
 import { CKEDITOR5_COMMERCIAL_PATH, CKEDITOR5_ROOT_PATH } from '../constants.mjs';
+import umbertoJson from '../../docs/umberto.json' with { type: 'json' };
+
+const CKBOX_VERSION = umbertoJson.variables.CKBOX_VERSION;
```
pomek:
```suggestion
const { CKBOX_VERSION } = umbertoJson.variables;
```

- Cosmetic, but in this case, enough to shorten it.
- I do not see any confirmation that the built documentation loads a proper resource.
- There are three (or even four, soon) commits. Can we merge all-in?

przemyslaw-zan: I've built docs, visited https://172.24.224.54:8080/ckeditor5/latest/features/file-management/ckbox.html. CKBox demo works, and `CKBox.version` in console returns `"2.9.2"`

Issue #18433, opened 2025-04-25
[iOS] Inline editor sticky toolbar not sticky horizontally to the editor edge


niegowski:
1.  Open the inline editor on an iPhone (Safari) with or without custom viewportOffset
2.  Zoom the page so that the toolbar becomes wider than the visible area
3.  Scroll horizontally to the edge of the page

**Result:**   
While scrolling horizontally, the toolbar moves sideways and it's not aligned with the editor.  
See the video:  [https://drive.google.com/file/d/1xz6mBS1osxIXkm1gTKF2oDHSGVvCTuml/view?usp=drive_link](https://drive.google.com/file/d/1xz6mBS1osxIXkm1gTKF2oDHSGVvCTuml/view?usp=drive_link)

_Originally posted by @juliaflejterska in https://github.com/ckeditor/ckeditor5/issues/18320#issuecomment-2815865544_

CKEditorBot:
There has been no activity on this issue for the past year. We've marked it as stale and will close it in 30 days. We understand it may still be relevant, so if you're interested in the solution, leave a comment or reaction under this issue.

CKEditorBot:
We've closed your issue due to inactivity. We understand that the issue may still be relevant. If so, feel free to open a new one (and link this issue to it).

PR #19540, opened 2025-12-19
Fixed an issue where the sticky toolbar would scroll off-screen on Android Chrome when the on-screen keyboard was opened.

### LINKED ISSUE #19514: Visual viewport toolbar fix should apply to Android Chrome 108+ (not just iOS/Safari)

1. Open any CKEditor 5 demo with a sticky toolbar on an Android device running Chrome 108 or later (e.g., https://ckeditor.com/docs/ckeditor5/latest/examples/builds/classic-editor.html)
2. Tap on the editor to focus it and bring up the on-screen keyboard (OSK)
3. Observe the toolbar position

### ✔️ Expected result

The sticky toolbar should remain visible at the top of the visual viewport, similar to how it behaves on iOS Safari. The toolbar should stay "pinned" to the top of the visible area even when the keyboard opens.

### ❌ Actual result

The toolbar scrolls off-screen (upward) and becomes invisible when the on-screen keyboard opens. The toolbar appears to be following the layout viewport rather than adjusting for the visual viewport offset.

### ❓ Possible solution

The fix implemented in PR #18320 (merged April 2025, released in v45.1.0) correctly handles this issue on iOS/Safari by using the Visual Viewport API. However, it appears to only apply to Safari/iOS.

Since Chrome 108 (released November 2022), Chrome on Android now uses the same `resizes-visual` viewport behavior as iOS Safari by default. See: https://developer.chrome.com/blog/viewport-resize-behavior#changing_the_default_behavior_in_chrome_108

**The same Visual Viewport API approach used for iOS should also work on Android Chrome 108+**, as both platforms now share identical viewport resize behavior when the keyboard opens.

Suggested implementation:
- Remove or expand any browser detection that limits the Visual Viewport fix to Safari/iOS only
- Apply the `getVisualViewportOffset()` logic to all browsers that support the Visual Viewport API
- Test on Android Chrome 108+ to verify the toolbar stays visible when the OSK opens

## 📃 Other details

* Browser: Chrome 108+ on Android (tested on Chrome 131)
* OS: Android 11+
* First affected CKEditor version: All versions (the iOS fix in v45.1.0 doesn't extend to Android)
* Installed CKEditor plugins: All default builds affected (Classic, Inline, Balloon, etc.)

**Additional context:**
- This works correctly on iOS Safari (after v45.1.0)
- This affects PWAs and mobile web apps significantly
- The root cause appears to be that the Visual Viewport fix is Safari/iOS-specific when it should be platform-agnostic

---

Mati365:
### 🚀 Summary

Fixed an issue where the sticky toolbar would scroll off-screen on Android Chrome when the on-screen keyboard was opened.

### Before

https://github.com/user-attachments/assets/1d6edf23-3b77-45f4-a138-a932894a8ae2

### After

https://github.com/user-attachments/assets/43831630-c7fe-49df-ad16-bcc162f7969c

### 📌 Related issues

* Closes #19514 
* Known issue that was not fixed https://github.com/ckeditor/ckeditor5/issues/18433

### DIFF
--- a/packages/ckeditor5-ui/tests/panel/sticky/stickypanelview.js
+++ b/packages/ckeditor5-ui/tests/panel/sticky/stickypanelview.js
@@ -417,18 +417,10 @@ describe( 'StickyPanelView', () => {
 		} );
 
 		describe( 'after scrolling', () => {
-			let visualViewportStub;
-
 			beforeEach( () => {
 				view.isActive = true;
-				visualViewportStub = sinon.stub( global.window, 'visualViewport' ).get( () => ( {
-					offsetLeft: 0,
-					offsetTop: 0
-				} ) );
-			} );
-
-			afterEach( () => {
-				visualViewportStub.restore();
+				sinon.stub( visualViewport, 'offsetLeft' ).get( () => 0 );
+				sinon.stub( visualViewport, 'offsetTop' ).get( () => 0 );
 			} );
 
 			describe( 'if there is only window scrollable', () => {
@@ -629,10 +621,9 @@ describe( 'StickyPanelView', () => {
 			describe( 'if there is window scrollable and visual viewport', () => {
 				beforeEach( () => {
 					view.isActive = true;
-					visualViewportStub.get( () => ( {
-						offsetLeft: 15,
-						offsetTop: 25
-					} ) );
+
+					sinon.stub( visualViewport, 'offsetLeft' ).get( () => 15 );
+					sinon.stub( visualViewport, 'offsetTop' ).get( () => 25 );
 				} );
 
 				it( 'should make panel sticky to the top if the limiter top is not visible', () => {
@@ -787,10 +778,8 @@ describe( 'StickyPanelView', () => {
 				beforeEach( () => {
 					view.viewportTopOffset = 5;
 					view.isActive = true;
-					visualViewportStub.get( () => ( {
-						offsetLeft: 15,
-						offsetTop: 25
-					} ) );
+					sinon.stub( visualViewport, 'offsetLeft' ).get( () => 15 );
+					sinon.stub( visualViewport, 'offsetTop' ).get( () => 25 );
 				} );
 
 				it( 'should make panel sticky to the top if the limiter top is not visible', () => {


packages/ckeditor5-ui/tests/panel/sticky/stickypanelview.js
```diff
@@ -418,11 +417,21 @@ describe( 'StickyPanelView', () => {
...
 		describe( 'after scrolling', () => {
-			describe( 'if there is only window scrollable', () => {
-				beforeEach( () => {
-					view.isActive = true;
-				} );
+			let visualViewportStub;
 
+			beforeEach( () => {
+				view.isActive = true;
+				visualViewportStub = sinon.stub( global.window, 'visualViewport' ).get( () => ( {
```
niegowski: This is already stubbed in line 274. Can we unify this?

niegowski:
This would require QA testing. It should include the viewport offset editor configuration and scroll to the selection handling. Also, the balloon positioning should be tested. Testing should include all browsers, not only mobile.

packages/ckeditor5-ui/tests/panel/sticky/stickypanelview.js
Mati365: I think the stub is correct. There is separate `describe` that tests behaviour of sticky position calculation especially with visual viewport.

charlttsie:
### [Android] Selecting a widget when sticky toolbar is full visible scrolls the page down and hides the toolbar
#### Steps
1. Open [all-features.html](http://localhost:8125/ckeditor5/tests/manual/all-features.html) on an Android device
2. Ensure the toolbar fits into the viewport
3. Tap a widget

#### Result
The page scrolls down slightly so the upper edge of the editable aligns with the top of the viewport, and the sticky toolbar is no longer visible:

https://github.com/user-attachments/assets/e4a31ecc-c163-4e89-a29f-54802e353f2d

#### Additional info
- Occurs also if the keyboard is already opened before tapping the widget, so might be a separate issue
- On Safari, after tapping a widget, the scrolling doesn't occur and the toolbar is fully visible

charlttsie:
### [Android] Viewport offset is not respected when scrolling down with the on-screen keyboard open
#### Steps
1. Set `editor.ui.viewportOffset.top` in [all-features.html](http://localhost:8125/ckeditor5/tests/manual/all-features.html) to e.g. `100` px
2. Open the manual test and focus the editor to open the keyboard
3. Scroll down and up

#### Result
The offset is absent when scrolling down but present when scrolling up or when the keyboard is closed:

https://github.com/user-attachments/assets/353071da-c14a-460d-9de2-dd3f28bdfebe

charlttsie:
### [Android] Typing after scrolling moves the caret behind the sticky toolbar
#### Steps
1. Open [all-features.html](http://localhost:8125/ckeditor5/tests/manual/all-features.html)
2. Place the caret somewhere in the editor
3. Scroll down so the caret is not visible
4. Start typing

#### Result
The caret and newly typed text are hidden behind the toolbar:

https://github.com/user-attachments/assets/34ae9349-3302-4cde-83ff-900ce116769f

#### Additional info
- Regression compared to `master`, where the caret and newly typed text are visible immediately after typing

charlttsie:
I've tested the PR and the main focus was on points listed in the [comment](https://github.com/ckeditor/ckeditor5/pull/19540#issuecomment-3674382915). The issues I found are reported above and they are all specific to Android. None of them occur on `master`, where the toolbar is hidden in described cases. The changes look good on iOS and desktop browsers, where I didn't find any regression it would introduce 👌

Mati365:
We have decided to move it back to the backlog as these issues may be difficult to resolve at the moment.

PR #19551, opened 2025-12-22
Add table cell type scope support.

Mati365:
### 🚀 Summary

Add table cell type scope support.

obraz

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/3175 https://github.com/ckeditor/ckeditor5-commercial/issues/9049

Issue #3175, opened 2019-10-09
Heading rows and columns might be missing markup in the view.


jodator:
The heading rows and heading columns cells[ should have "scope" attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Tables/Advanced#Tables_for_visually_impaired_users) set for accessibility reasons:

> Screenreaders will recognize markup structured like this, and allow their users to read out the entire column or row at once, for example.

Example proper markup of a table with one heading 

```html
<table>
	<thead>
		<tr>
			<th scope="col">a</th>
			<th scope="col">b</th>
		</tr>
	</thead>
	<tbody>
		<tr>
			<th scope="row">c</th>
			<td>d</td>
		</tr>
	</tbody>
</table>
```

madhaze:
`scope="col"` and `scope="row"` are accessibility issues and get flagged by accessibility scans.  
Example from Level Access scan

> HTML Element:
<th>Mixtape</th>
Description
One or more th elements that is in a table in which other elements contain scope attributes set to non-null values, excluding those elements with a rowspan or colspan attribute set to a value of '0'; or '2' or more, without an ARIA-assigned role, not intentionally hidden in the DOM and available to assistive technologies, does not have a scope attribute set to a text value of 'row' or 'col'.
Why it matters
Without an appropriate scope attribute on table header cells, header and data cells will not be properly associated as the correct header will not be clear. When data cells are not properly associated with header cells, users of assistive technology may not be able to reliably identify headers and navigate data tables.
What to do
Set the scope attribute to either scope=col (column headers) or scope=row (row headers). Tables that use scope to associate data and header cells need to use this scope approach to associate all data and header cells and must not mix use of scope and use of id and header attributes to associate header and data cells.

jonnyhocks:
Hi all, dropping in here to see if there is any update on this or if anyone has a solution for getting these attributes added?

Thanks in advance 👍

Witoso:
From our side, we haven't done it yet, but with [General HTML Support](https://ckeditor.com/docs/ckeditor5/latest/features/html/general-html-support.html) enabled those attributes are preserved.

jonnyhocks:
Thanks for the quick response; I noticed there was a 'yet'!  Would that mean there is a roadmap for it?

Witoso:
There's always some roadmap :) This most likely would be a part of some larger initiative around the table enhacements but we don't have timeline for it right now.

[...]

PR #19556, opened 2025-12-23
Add table cell scope documentation.

Mati365:
### 🚀 Summary

Add table cell scope docs.

---

### 📌 Related issues

* Feature introduced in https://github.com/ckeditor/ckeditor5/pull/19551
* Related issues https://github.com/ckeditor/ckeditor5/issues/3175 https://github.com/ckeditor/ckeditor5-commercial/issues/9049


Witoso:
But it was added later than `th`, so maybe that was a cause? We are introducing them both at the same time. To consider if `true` shouldn't be the default.

PR #19582, opened 2026-01-05
Add support for table footers. 

### LINKED ISSUE #12952: `<tfoot>` element is lost during the conversion

*   Go to 
*   Open the source editing.
*   Paste the following markup:

```
<table>
    <thead>
        <tr>
            <th>Items</th>
            <th scope="col">Expenditure</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <th scope="row">Donuts</th>
            <td>3,000</td>
        </tr>
        <tr>
            <th scope="row">Stationery</th>
            <td>18,000</td>
        </tr>
    </tbody>
    <tfoot>
        <tr>
            <th scope="row">Totals</th>
            <td>21,000</td>
        </tr>
    </tfoot>
</table>
```

*   Switch back to the editing mode.
*   Switch back to the source editing mode.

### ✔️ Expected result

`<tfoot>` is preserved.

### ❌ Actual result

`<tfoot>` is lost and merged to `<tbody>`.

---

Mati365:
### 🚀 Summary

Add support for table footers. 

---

### 📌 Related issues

* Closes #12952 
* Based on https://github.com/ckeditor/ckeditor5/pull/14276 (Huge thanks to @star-szr). This PR adds UI, RTC conflicts resolution, integrates it with cell types, and adds command to the editor.

---

### 🧾 Checklists

#### Author checklist

- [x] Is performance impacted?
- [x] Are there any changes the team should be informed about (e.g. architectural, difficult to revert in future versions or having impact on other features)?
- [x] Were these changes documented (in Logbook)?

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-table/src/converters/table-structure-refresh-handler.ts
+++ b/packages/ckeditor5-table/src/converters/table-structure-refresh-handler.ts
@@ -27,6 +27,7 @@ import { TableWalker } from '../tablewalker.js';
  */
 export function tableStructureRefreshHandler( model: Model, editing: EditingController ): void {
 	const differ = model.document.differ;
+	const movedRows = new Set<ModelElement>();
 	const rowsToReconvert = new Set<ModelElement>();
 	const cellsToReconvert = new Set<ModelElement>();
 
@@ -59,14 +60,13 @@ export function tableStructureRefreshHandler( model: Model, editing: EditingCont
 		}
 
 		// Mark row to be reconverted when it was moved around so that `<th>` and `<td>` elements can be updated.
-		// TODO This should be handled in the engine. A moved row was reused in the view
-		//  and differ does not allow refresh of cells in inserted row.
+		// See https://github.com/ckeditor/ckeditor5/issues/19671.
 		if (
 			change.type == 'insert' &&
 			change.name == 'tableRow' &&
 			editing.mapper.toViewElement( change.position.nodeAfter as ModelElement )
 		) {
-			rowsToReconvert.add( change.position.nodeAfter as ModelElement );
+			movedRows.add( change.position.nodeAfter as ModelElement );
 		}
 
 		const headingRows = table.getAttribute( 'headingRows' ) as number || 0;
@@ -86,6 +86,12 @@ export function tableStructureRefreshHandler( model: Model, editing: EditingCont
 
 			if ( viewElement.name != expectedElementName ) {
 				cellsToReconvert.add( tableSlot.cell );
+
+				// Reconvert rows that were just inserted (moved) as marking cells inside for reconversion does not work.
+				// See https://github.com/ckeditor/ckeditor5/issues/19671.
+				if ( movedRows.has( tableSlot.cell.parent as ModelElement ) ) {
+					rowsToReconvert.add( tableSlot.cell.parent as ModelElement );
+				}
 			}
 		}
 	}


Witoso:
There are a few areas here worth considering:
1. As we discussed in the past, this needs UI similar to table headers.
2. If we do it _without UI_, just to support normalization, is it enough? We would have users guess this is loaded as a specific markup (visible only via source mode) What happens if someone adds a row below `tfoot`? Would this row turn to `tfoot`? Is this understandable?
3. If we add UI, would it be graspable by the majority of users? While headers appear across apps like Word or GDocs, table footer is not understandable to 99%, and most would confuse them with captions at first glance.

I know I showed a lot of hesitancy in the topic in the past. To cut my doubts and to not make this stale again, I would approve a change without the UI when 1) it's a separate plugin **or** 2)  it's hidden under a config flag in tables and `tfoot` is opt-in for now.

Mati365:
@Witoso 

> As we discussed in the past, this needs UI similar to table headers.

That was the original plan. A UI toggle identical to the one used for table headers should be fully sufficient and consistent from a UX point of view.

> What happens if someone adds a row below tfoot?

A new row should be added to `tfoot`, and the `footerRows` counter should be incremented accordingly. This keeps the behavior predictable and aligned with how header rows work.

> [...] most would confuse them with captions at first glance.

I don’t think so. It should look like a regular table row. If needed later, we could slightly differentiate it via styling (e.g. different cell paddings), but for now I’d prototype it in an untouched, neutral form and validate assumptions based on feedback rather than upfront fears.

> I know I showed a lot of hesitancy in the topic in the past. To cut my doubts and to not make this stale again, I would approve a change without the UI when 1) it's a separate plugin or 2) it's hidden under a config flag in tables and tfoot is opt-in for now.

Between these options, a config flag seems less problematic. The logic operating on heading rows is already deeply rooted across multiple plugins (for example, cell types), and my goal is to keep this functionality as close as possible-both UX-wise and behavior-wise-to how header rows work today.

This approach also has a practical upside: if we ever need to backport the change to an LTS version, doing so would be trivial thanks to the opt-in nature and minimal surface impact.

Witoso:
Btw there's a PR https://github.com/ckeditor/ckeditor5/pull/14276, I'm not sure if you are taking a new direction, but this was once reviewed by @arkflpc, so we shouldn't do anything new here I guess.

Witoso:
>That was the original plan. A UI toggle identical to the one used for table headers should be fully sufficient and consistent from a UX point of view.

For now I don't agree with exposing this feature to all users by default; this needs to be an opt-in behavior.

Mati365:
@Witoso 

> Btw there's a PR https://github.com/ckeditor/ckeditor5/pull/14276, I'm not sure if you are taking a new direction, but this was once reviewed by @arkflpc, so we shouldn't do anything new here I guess.

I’m aware of PR https://github.com/ckeditor/ckeditor5/pull/14276. However, due to changes in the Table plugin - like the introduction of cell types and modifications in the column scanner - there are now a lot of conflicts, and that PR can no longer be used as-is. While the underlying logic can still be aligned in spirit, the code itself is fundamentally different and requires a more thorough approach to handling edge cases.

> For now I don't agree with exposing this feature to all users by default; this needs to be an opt-in behavior.

A config flag can easily control this, keeping the feature opt-in without any problem.

olstjos:
We've moved on, we created a module for this.  https://drupal.org/project/ckeditor5_table_fix .  While it was a shock to have our tables (and other markup) corrupted by the move from ck4 to ck5 , we're pleased that the api it'self (hopefully stable now) allows default and plugin up and down to be corrected/overridden.  The opinion of ck5 is that our markup doesn't fit however we have our own opinions and the api allows us to adjust.  It took us 9 months to iron things out and come up with an acceptible solution.

packages/ckeditor5-table/src/augmentation.ts
```diff
@@ -140,6 +125,7 @@ declare module '@ckeditor/ckeditor5-core' {
 		selectTableRow: SelectRowCommand;
 		setTableColumnHeader: SetHeaderColumnCommand;
 		setTableRowHeader: SetHeaderRowCommand;
+		setFooterRow: SetFooterRowCommand;
```
niegowski:
```suggestion
		setTableRowFooter: SetFooterRowCommand;
```
or
```suggestion
		setTableFooterRow: SetFooterRowCommand;
```

packages/ckeditor5-table/src/commands/mergecellcommand.ts
```diff
@@ -203,20 +203,31 @@ function getVerticalCell( tableCell: ModelElement, direction: ArrowKeyCodeDirect
...
-	const isMergeWithBodyCell = direction == 'down' && ( rowIndex + rowspan ) === headingRows;
-	const isMergeWithHeadCell = direction == 'up' && rowIndex === headingRows;
+	const footerRows = table.getAttribute( 'footerRows' ) as number || 0;
+	const footerIndex = rows - footerRows;
+
+	const isMergeUpWithBodyCell = direction == 'up' && rowIndex === footerIndex;
+	const isMergeUpWithHeadCell = direction == 'up' && rowIndex === headingRows;
+
+	const isMergeDownWithBodyCell = direction == 'down' && ( rowIndex + rowspan ) === headingRows;
+	const isMergeDownWithFootCell = direction == 'down' && rowIndex === footerIndex - 1;
```
niegowski: Is this correct? What about spanned cells?

packages/ckeditor5-table/src/converters/downcast.ts
```diff
@@ -51,13 +52,32 @@ export function downcastTable( tableUtils: TableUtils, options: DowncastTableOpt
...
 				writer.createPositionAt( tableElement, 'end' ),
 				writer.createContainerElement(
 					'tbody',
 					null,
-					writer.createSlot( element => element.is( 'element', 'tableRow' ) && element.index! >= headingRows )
+					writer.createSlot( element =>
+						element.is( 'element', 'tableRow' ) &&
+							element.index! >= headingRows &&
+							element.index! < tableUtils.getRows( table ) - footerRows
+					)
```
niegowski, line 66:
```suggestion
					writer.createSlot( element =>
						element.is( 'element', 'tableRow' ) &&
						element.index! >= headingRows &&
						element.index! < tableUtils.getRows( table ) - footerRows
					)
```

packages/ckeditor5-table/src/commands/mergecellcommand.ts
Mati365: Good catch, it lacks `rowspan` handling. I'll adjust it.

Mati365: Fixed in https://github.com/ckeditor/ckeditor5/pull/19582/changes/2c9e16d526e8f1e4fe7470b4f6ebf3a630f64664

.changelog/20260113113541_ck_12952.md
```diff
@@ -0,0 +1,13 @@
...
+type: Feature
+scope:
+  - ckeditor5-table
+closes:
+  - https://github.com/ckeditor/ckeditor5/issues/12952
+communityCredits:
+  - star-szr
+---
+
+Added support for table footers. It should be now possible upcast `<tfoot>` elements from HTML to the model and use the `footerRows` attribute to define the number of footer rows in a table. The footer row dropdown toggle has also been added to the table properties UI.
```
niegowski: This description is too technical. It should be user-facing.

.changelog/20260113113541_ck_12952.md
```diff
@@ -0,0 +1,13 @@
...
+
+This feature is disabled by default and can be enabled by setting the `table.useTfootElement` configuration to `true`.
```
niegowski: The `useTfootElement` looks odd. Could the `enableFooters` be better?

packages/ckeditor5-table/src/commands/inserttablecommand.ts
```diff
@@ -51,6 +51,9 @@ export class InsertTableCommand extends Command {
 	 * {@link module:table/tableconfig~TableConfig#defaultHeadings `config.table.defaultHeadings.rows`} table config.
 	 * @param options.headingColumns The number of heading columns. If not provided it will default to
 	 * {@link module:table/tableconfig~TableConfig#defaultHeadings `config.table.defaultHeadings.columns`} table config.
+	 * @param options.footerRows The number of footer rows. If not provided it will default to
+	 * {@link module:table/tableconfig~TableConfig#defaultFooters `config.table.defaultFooters.footerRows`} table config.
```
niegowski: The `config.table.defaultFooters.footerRows` does not match the code below.

packages/ckeditor5-table/src/converters/downcast.ts
```diff
@@ -51,13 +52,32 @@ export function downcastTable( tableUtils: TableUtils, options: DowncastTableOpt
...
 			writer.insert(
 				writer.createPositionAt( tableElement, 'end' ),
 				writer.createContainerElement(
 					'tbody',
 					null,
-					writer.createSlot( element => element.is( 'element', 'tableRow' ) && element.index! >= headingRows )
+					writer.createSlot( element =>
+						element.is( 'element', 'tableRow' ) &&
+						element.index! >= headingRows &&
+						element.index! < tableUtils.getRows( table ) - footerRows
```
niegowski: The `tableUtils.getRows( table )` function is repeated multiple times in this function; perhaps we should extract it into a variable.

packages/ckeditor5-table/src/converters/downcast.ts
```diff
@@ -309,17 +340,24 @@ export function downcastPlainTable(
...
+
 	// Table contents element containing <thead> and <tbody> when necessary.
 	const tableContentElements: Array<ViewContainerElement> = [];
 
 	if ( headingRows ) {
 		tableContentElements.push( theadElement );
 	}
 
-	if ( headingRows < table.childCount ) {
+	if ( headingRows + footerRows < table.childCount ) {
```
niegowski: The `table.childCount` might be invalid for a table with a caption. Maybe we should fix this to use real rows count, especially that we already have it above.

packages/ckeditor5-table/src/tableclipboard.ts
```diff
@@ -344,8 +352,12 @@ export class TableClipboard extends Plugin {
 		const headingRows = parseInt( selectedTable.getAttribute( 'headingRows' ) as string || '0' );
 		const headingColumns = parseInt( selectedTable.getAttribute( 'headingColumns' ) as string || '0' );
 
+		const footerRows = parseInt( selectedTable.getAttribute( 'footerRows' ) as string || '0' );
+		const footerIndex = tableUtils.getRows( selectedTable ) - footerRows;
+
 		const areHeadingRowsIntersectingSelection = selection.firstRow < headingRows && headingRows <= selection.lastRow;
 		const areHeadingColumnsIntersectingSelection = selection.firstColumn < headingColumns && headingColumns <= selection.lastColumn;
+		const areFooterRowsIntersectingSelection = selection.lastRow >= footerIndex && footerIndex >= selection.firstRow;
```
niegowski: I'm not sure about this condition; shouldn't it be reversed as those 2 above? What about `>=` vs `>`? Are edge cases covered correctly here?

packages/ckeditor5-table/src/tableconfig.ts
```diff
@@ -47,6 +47,37 @@ export interface TableConfig {
...
+	/**
+	 * Enables support for table footers (`<tfoot>`).
+	 *
+	 * When set to `true`, the editor will upcast and downcast `<tfoot>` elements, and the footer toggle will be visible
+	 * in the table row dropdown.
+	 * When set to `false` (default), footer rows are ignored and the footer toggle is hidden in the table row dropdown.
+	 *
+	 * @default false
+	 */
+	useTfootElement?: boolean;
```
niegowski: The `useTfootElement` seems too technical. Could the `enableFooters` be better?

packages/ckeditor5-table/src/tableconfig.ts
```diff
@@ -47,6 +47,37 @@ export interface TableConfig {
...
+	/**
+	 * Number of footer rows to render by default when inserting new tables.
+	 *
+	 * You can configure it like this:
+	 *
+	 * ```ts
+	 * const tableConfig = {
+	 * 	defaultFooters: {
+	 * 		rows: 1
+	 * 	}
```
niegowski: Do we need this nested object with a single `rows` property?

niegowski (changes requested):
Partial review.

.changelog/20260113113541_ck_12952.md
Mati365: Fixed in https://github.com/ckeditor/ckeditor5/pull/19582/commits/c7bfff6eada505be283b4d983da169fbffdc8e65

packages/ckeditor5-table/src/converters/downcast.ts
Mati365: Fixed in https://github.com/ckeditor/ckeditor5/pull/19582/changes/e13033f5c9992c9e8cf0e66ffb265a0f611dff75

.changelog/20260113113541_ck_12952.md
Mati365: Fixed in https://github.com/ckeditor/ckeditor5/pull/19582/changes/7693b3a330edd76d2f34dd2668f6db7f2be5dc2d

packages/ckeditor5-table/src/tableconfig.ts
Mati365: Fixed in https://github.com/ckeditor/ckeditor5/pull/19582/changes/c7bfff6eada505be283b4d983da169fbffdc8e65

packages/ckeditor5-table/src/commands/inserttablecommand.ts
Mati365: Fixed in https://github.com/ckeditor/ckeditor5/pull/19582/commits/01e643ec1fa5974e01f80c5df0228b7c0d40a9ac

packages/ckeditor5-table/src/tableconfig.ts
Mati365: Fixed in https://github.com/ckeditor/ckeditor5/pull/19582/changes/85606d401e272901f1fe65010aa23d226e66ee28

packages/ckeditor5-table/src/tableclipboard.ts
Mati365: I realigned the condition to make it identical to the two above and added a test. From what I can see, while `footerIndex >= selection.firstRow` makes a difference to the covered tests (and is covered), there is no difference between `selection.lastRow >= footerIndex` and `selection.lastRow > footerIndex`, as the internal logic of the `doHorizontalSplit` function seems to ignore them anyway. 

https://github.com/ckeditor/ckeditor5/pull/19582/changes/c8762bbcebbcb6dab6e7216d615ff5e53651699f
https://github.com/ckeditor/ckeditor5/pull/19582/changes/ef98d4a04957e501dfb7aeb2096b5f0a3c312da6

packages/ckeditor5-table/src/converters/downcast.ts
Mati365: I adjusted condition https://github.com/ckeditor/ckeditor5/pull/19582/commits/80f9fd9b49b44cabb06adadfa1d32776db38ca29

packages/ckeditor5-table/src/tableutils.ts
```diff
@@ -1083,69 +1164,76 @@ export class TableUtils extends Plugin {
...
+			autoExpand = true
+		} = options;
+
+		const totalColumns = this.getColumns( table );
 		const oldHeadingColumns = table.getAttribute( 'headingColumns' ) as number || 0;
 		const scopedHeaders = !!this.editor.config.get( 'table.tableCellProperties.scopedHeaders' );
 
+		let truncatedHeadingColumns = Math.min( headingColumns, totalColumns );
+
 		if ( headingColumns === oldHeadingColumns ) {
```
niegowski:
```suggestion
		if ( truncatedHeadingColumns === oldHeadingColumns ) {
```

packages/ckeditor5-table/src/tableutils.ts
```diff
@@ -142,11 +144,15 @@ export class TableUtils extends Plugin {
...
+			this.setHeadingRowsCount( writer, table, options.headingRows );
 		}
 
 		if ( options.headingColumns ) {
-			this.setHeadingColumnsCount( writer, table, Math.min( options.headingColumns, columns ) );
+			this.setHeadingColumnsCount( writer, table, options.headingColumns );
+		}
+
+		if ( options.footerRows ) {
+			this.setFooterRowsCount( writer, table, options.footerRows );
```
niegowski: As those are clamped and the latter one takes precedence, maybe we should first set footer rows and then heading rows as probably headers are more important than footers.

packages/ckeditor5-table/src/tableutils.ts
```diff
@@ -1332,33 +1420,60 @@ export class TableUtils extends Plugin {
...
 }
 
+/**
+ * Options for the {@link module:table/tableutils~TableUtils#setHeadingRowsCount} and
+ * {@link module:table/tableutils~TableUtils#setHeadingColumnsCount} methods.
+ */
+export type UpdateTableHeadingsOptions = {
+	updateCellType?: boolean;
+	resetFormerHeadingCells?: boolean;
+	autoExpand?: boolean;
```
niegowski, line 1503: Please add docs for those options.

packages/ckeditor5-table/src/converters/table-structure-refresh-handler.ts
```diff
@@ -58,16 +61,47 @@ export function tableHeadingsRefreshHandler( model: Model, editing: EditingContr
...
+				const oldValue = ( change.attributeOldValue as number | null ) || 0;
+				const newValue = ( change.attributeNewValue as number | null ) || 0;
+
+				const minFooterRows = Math.min( oldValue, newValue );
+				const maxFooterRows = Math.max( oldValue, newValue );
+
+				if ( tableSlot.row >= totalRows - maxFooterRows && tableSlot.row <= totalRows - minFooterRows ) {
+					shouldReconvert = true;
+				}
+			}
```
niegowski: Is this part required? AFAIK, footer rows do not change TD -> TH, so there should not be any cell reconversion needed. But maybe this is somehow needed to move the row from TBODY to THEAD? If so, maybe we should verify if the view element name is correct in a similar fashion as we check table cells above?

packages/ckeditor5-table/src/tableediting.ts
```diff
@@ -230,17 +247,22 @@ export class TableEditing extends Plugin {
...
 		injectTableLayoutPostFixer( model );
 		injectTableCellParagraphPostFixer( model );
+		injectTableStructurePostFixer( editor );
 
 		this.listenTo( model.document, 'change:data', () => {
 			// It's no longer needed to refresh table headings on every data change if table cell type feature is enabled.
 			// It's because headings rows / columns are updated based on cell types which triggers their own refresh handler.
 			if ( !isTableCellTypeEnabled( editor ) ) {
-				tableHeadingsRefreshHandler( model, editor.editing );
+				tableStructureRefreshHandler( model, editor.editing, tableUtils );
```
niegowski: If the cell type is not enabled, do we still need it for THEAD, TBODY, and TFOOT changes? I'm not sure

packages/ckeditor5-table/tests/commands/inserttablecommand.js
```diff
@@ -359,6 +362,132 @@ describe( 'InsertTableCommand', () => {
...
+				_setModelData( model, '[]' );
+
+				command.execute( { rows: 3, columns: 3 } );
+
+				expect( _getModelData( model, { withoutSelection: true } ) ).to.equalMarkup(
+					modelTable( [
+						[ '', '', '' ],
+						[ '', '', '' ],
+						[ '', '', '' ]
+					], { headingRows: 1, footerRows: 2 } )
```
niegowski: I'd expect this to be the opposite, 2 heading rows and a single footer row.

packages/ckeditor5-table/tests/commands/setfooterrowcommand.js
```diff
@@ -0,0 +1,481 @@
...
+			// forceValue is false.
+			// Should do nothing as it matches intent? Or just set value to false (unset all footers)?
+			// Wait, forceValue means "Set footer state to false".
+			// If I forceValue=false on a non-footer row, it should probably ensure this row is NOT a footer.
+			// But the command sets the footer attribute for the table.
+			// If forceValue=false, it means we want to UNSET footer for the selected row.
+			// Since row 1 is NOT a footer, nothing should happen?
+			// Let's check logic:
+			// if ( options.forceValue === this.value ) return;
+			// Here value is false. forceValue is false. Return. Correct.
```
niegowski: Why such long descriptions? In multiple tests in this file. I'm not sure if those are needed. Perhaps some ASCII art tables would provide a clearer explanation.

packages/ckeditor5-table/tests/commands/inserttablecommand.js
Mati365: yep, good catch

packages/ckeditor5-table/src/tableediting.ts
Mati365: That condition must be moved to refresh handler, or activated when footers are enabled.

PR #19642, opened 2026-01-20
Fix GPL URL in the `ckeditor5-emoji` package

pomek:
### 🚀 Summary

Internal change, before this PR, there was an invalid URL (`http` instead of `https`).


pomek:
_Bypassing the admin approval because there is no admin who can do that._

Issue #15979, opened 2024-03-05
`figure` and `figcaption` aria improvements in tables


Witoso:
> side question what about `table > caption`? It looks like another choice do describe table.

There are [several techniques for describing `table`](https://www.w3.org/TR/html5/tabular-data.html#sec-techniques-for-describing-tables). In our case we use `figcaption` as also the mean to describe the whole `figure` element. Probably it would be nice to additionally bind `figcaption` with `table` using `[aria-labelledby]`/`[aria-describedby]`.

_Originally posted by @Comandeer in https://github.com/ckeditor/ckeditor5/issues/3194#issuecomment-397245583_

Rel: https://www.w3.org/WAI/tutorials/tables/caption-summary/

PR #19653, opened 2026-01-21
Add `aria-label` based on `figcaption` to table editing.

Mati365:
### 🚀 Summary

Add `aria-label` based on `figcaption` to table editing.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/15979

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-table/src/tablecaption/tablecaptionediting.ts
+++ b/packages/ckeditor5-table/src/tablecaption/tablecaptionediting.ts
@@ -149,6 +149,7 @@ export class TableCaptionEditing extends Plugin {
 				// Remove `aria-labelledby` from the table if there is no caption.
 				if ( !modelCaption ) {
 					writer.removeAttribute( 'aria-labelledby', viewTable );
+
 					return;
 				}
 


Mati365:
@niegowski 

> In the issue there was a suggestion to use [aria-labelledby]/[aria-describedby]. 

`aria-labeleldby` and `aria-describedby` require adding `id` parameter on caption, which is not always present on caption.

packages/ckeditor5-table/src/converters/table-caption-aria-label-handler.ts
```diff
@@ -0,0 +1,73 @@
...
+				.find( child => child.is( 'element', 'table' ) ) as ViewElement | undefined;
+
+			/* istanbul ignore if -- @preserve */
+			if ( !viewTable ) {
+				return;
+			}
+
+			const modelCaption = Array
+				.from( modelTable.getChildren() )
+				.find( child => child.is( 'element', 'caption' ) ) as ModelElement | undefined;
```
niegowski: Please use the existing helper.

packages/ckeditor5-table/src/converters/table-caption-aria-label-handler.ts
```diff
@@ -0,0 +1,73 @@
...
+import { uid } from 'ckeditor5/src/utils.js';
+
+/**
+ * Handles `aria-labelledby` attribute on the table element in the editing view to
+ * make sure it always points to the caption of the table if there is one.
+ *
+ * @internal
+ */
+export function injectTableCaptionAriaLabelHandler( editor: Editor ): void {
+	const captionIdsMapping = new WeakMap<ModelElement, string>();
```
niegowski: Please store it in a private property of the table caption editing plugin.

packages/ckeditor5-table/src/converters/table-caption-aria-label-handler.ts
```diff
@@ -0,0 +1,73 @@
...
+
+	editor.conversion.for( 'editingDowncast' ).add( dispatcher => {
```
niegowski: This should be registered directly inside the table caption editing.

packages/ckeditor5-table/src/tablecaption/tablecaptionediting.ts
```diff
@@ -116,6 +127,54 @@ export class TableCaptionEditing extends Plugin {
...
+				if ( !viewTable ) {
+					return;
+				}
+
+				const modelCaption = getCaptionFromTableModelElement( modelTable );
+
+				// Remove `aria-labelledby` from the table if there is no caption.
+				if ( !modelCaption ) {
+					writer.removeAttribute( 'aria-labelledby', viewTable );
+					return;
```
niegowski, line 153:
```suggestion

					return;
```

charlttsie:
[JAWS][Firefox] Table caption is not announced when entering a table cell from outside the table
#### Steps
1. Open the [all-features.html](http://localhost:8125/ckeditor5/external/ckeditor5/packages/ckeditor5/tests/manual/all-features.html) manual test on Firefox on Windows
2. Run the JAWS screen reader
3. Insert a table caption
4. Click somewhere inside the editor but outside the table
5. Click inside a table cell

#### Result
Table caption is not announced even though the `aria-labelledby` attribute and the `ck-editor__caption` id are present and reference the same id:

https://github.com/user-attachments/assets/9f830292-8ead-48e6-94e4-d2df76ab6cfe

The same caption is announced on Chrome:

https://github.com/user-attachments/assets/5f78ebd3-e4a9-4621-a42f-b68dee72e244

charlttsie:
I've tested the PR on desktop browsers, as well as the NVDA, JAWS, Apple VoiceOver, and Windows Narrator screen readers. The `aria-labelledby` attribute and the `ck-editor__caption` id were present in all cases where the caption was on and absent in all cases where the caption was off. The caption was also announced by all listed screen readers apart from the case with JAWS and Firefox reported above. Aside from this scenario, everything else seems to be working correctly 👌

Mati365:
@charlttsie Sounds good enough to me, and it looks like the vast majority of cases work as expected. Thanks a lot for the thorough testing!

PR #19749, opened 2026-02-06
Add indent block list command

### LINKED ISSUE #19676: [List indent] List indent/outdent commands
- Command name: `indentBlockList` / `outdentBlockList`.
 - Enabled when:
     - Selection at start of the topmost list (collapsed or not collapsed).
 - Can be aplied only on the first list (placed at selection start) - when using TAB.
     - But can indent multiple lists with **toolbar button**.
 - Support for `(Shift + ) TAB`.
 - Support for toolbar buttons (Increase indent / Decrease indent).
 - Possibility to reset negative values (e.g. coming from MS Word) with indent.

mmotyczynska:
### 🚀 Summary

---

### 📌 Related issues

* Closes #19676

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-indent/src/integrations/indentblocklistcommand.ts
+++ b/packages/ckeditor5-indent/src/integrations/indentblocklistcommand.ts
@@ -75,9 +75,14 @@ export class IndentBlockListCommand extends Command {
 	}
 
 	/**
-	 * @inheritDoc
+	 * Executes the command.
+	 *
+	 * @fires execute
+	 * @param options Command options.
+	 * @param options.firstListOnly When `true`, indentation is applied only to the first list at the beginning of the selection.
+	 * When `false` or omitted, indentation is applied to all lists of the selection.
 	 */
-	public override execute( options: { source?: string } = {} ): void {
+	public override execute( options: { firstListOnly?: boolean } = {} ): void {
 		const editor = this.editor;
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-indent/src/integrations/indentblocklistcommand.ts
```diff
@@ -0,0 +1,137 @@
...
+		const model = editor.model;
+		const selection = model.document.selection;
+
+		model.change( writer => {
+			const listItem = this._getFirstListItemIfSelectionIsAtListStart( selection )!;
+
+			const listItems = [];
+
+			// If the command was not triggered by keyboard (i.e. from the toolbar), include all list from the selection.
+			if ( options.source !== 'keyboard' ) {
```
Mati365: I’m not sure if it wouldn't be better to use an attribute with a name that clearly defines the behavior instead of source. Right now, looking at the command and seeing `source: "keyboard"`, I don't really understand what happens under the hood.

mmotyczynska: Agree 👍  Thanks for pointing that out. Changed in 8770e6c2910bf87aad61374515cf34cc0ca1303e.

PR #19792, opened 2026-02-16
No longer return incorrect keys in `attributes` result of `getModelDataWithAttributes` test util.

### LINKED ISSUE #19757: `getModelDataWithAttributes` handles empty string attributes incorrectly

1. Create a model element with two attributes:
    * One attribute with an empty string value (e.g., `htmlContent: ''`).
    * Another attribute with an object value (e.g., `htmlInputAttributes: { classes: ... }`).
2. Run `getModelDataWithAttributes()` on this model.
3. Check the returned `data` string and the `attributes` object.

### ✔️ Expected result

The function should consistently reference attributes by their ID. If the empty string is stored in the `attributes` object at index `1`, the data string should refer to it as `(1)`, and subsequent attributes should follow the sequence (e.g., `(2)`).

```js
{
    data: '<paragraph><htmlInput htmlContent="(1)" htmlInputAttributes="(2)"></htmlInput></paragraph>',
    attributes: {
        1: '',
        2: {
            classes: [ 'foo', 'bar' ]
        }
    }
}
```

### ❌ Actual result

The empty string is inlined in the `data` string (as `htmlContent=""`), but it still consumes an index in the `attributes` object (`1: ''`).

This causes a mismatch for subsequent attributes. The second attribute in the string uses `(1)`, which points to the empty string in the object, instead of pointing to the correct object data (which is stored at index `2`).

```js
{
    data: '<paragraph><htmlInput htmlContent="" htmlInputAttributes="(1)"></htmlInput></paragraph>',
    attributes: {
        1: '', // Assigned ID 1, but not referenced in data string
        2: {
            classes: [ 'foo', 'bar' ] // Assigned ID 2, but data string refers to (1)
        }
    }
}
```

Mati365:
### 🚀 Summary

No longer return incorrect keys in `attributes` result of `getModelDataWithAttributes` test util.

---

### 📌 Related issues

* Closes #19757 

---

### 🧾 Checklists

#### Author checklist

- [x] Is the changelog entry intentionally omitted?

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-html-support/tests/_utils/utils.js
+++ b/packages/ckeditor5-html-support/tests/_utils/utils.js
@@ -45,7 +45,7 @@ export function getModelDataWithAttributes( model, options = {} ) {
 	let attributes = [];
 	for ( const item of range.getItems() ) {
 		for ( const [ key, value ] of sortAttributes( item.getAttributes() ) ) {
-			if ( key.startsWith( 'html' ) && !excludeAttributes.includes( key ) && typeof value === 'object' ) {
+			if ( key.startsWith( 'html' ) && !excludeAttributes.includes( key ) && value !== null && typeof value === 'object' ) {
 				attributes.push( value );
 			}
 		}


packages/ckeditor5-html-support/tests/_utils/utils.js
```diff
@@ -45,7 +45,7 @@ export function getModelDataWithAttributes( model, options = {} ) {
 	let attributes = [];
 	for ( const item of range.getItems() ) {
 		for ( const [ key, value ] of sortAttributes( item.getAttributes() ) ) {
-			if ( key.startsWith( 'html' ) && !excludeAttributes.includes( key ) ) {
+			if ( key.startsWith( 'html' ) && !excludeAttributes.includes( key ) && typeof value === 'object' ) {
```
gorzelinski: I don't know if the edge case applies, but `typeof null === 'object'`. Maybe:

```suggestion
			if ( key.startsWith( 'html' ) && !excludeAttributes.includes( key ) && value && typeof value === 'object' ) {
```

PR #19803, opened 2026-02-18
Reduce the `notify_ci_failure` resource class to `small`

pomek:
### 🚀 Summary

As in the title, nothing more to add.

---

### 📌 Related issues

* See: https://github.com/ckeditor/ckeditor5-internal/issues/4292.


pomek:
There are no issues with limited resources when installing dependencies.

image

Issue #19747, opened 2026-02-06
Always use plugin constructor instead of a soft requirement in features


pomek:
## 📝 Provide a description of the improvement

In the past, because of the DLL nature, each package had to be separate, and imports between feature packages were not allowed.

Since we are going to drop support for DLL builds (see #17779), could we bring back proper imports to improve DX?

Example: https://github.com/ckeditor/ckeditor5/blob/d6006e048fa8192ea08e7d0b312fba5475a82c47/packages/ckeditor5-emoji/src/emojimention.ts#L59-L61

### Definition of Done

- Review all of the soft requirements and decide if they can be replaced with a plugin constructor instead.
- Review ESLint rules that were blocking imports between packages. They are no longer needed.
- Review the documentation to remove the soft requirement section (_if defined_).
- Changes should be made in both repositories.

### Technical details

- Tree shaking.
- Circular dependencies.

pomek:
PRs:

1. https://github.com/ckeditor/ckeditor5/pull/19823
2. https://github.com/ckeditor/ckeditor5-commercial/pull/9373

pomek:
Both pull requests are merged.

PR #19823, opened 2026-02-23
Replace soft requirements with plugin constructors

pomek:
### 🚀 Summary

To avoid duplicating a changelog entry, see: https://github.com/ckeditor/ckeditor5/blob/ck/19747-drop-soft-require/.changelog/20260223151517_ck_19747_drop_soft_require.md.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/19747.

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist


Dumluregn:
Does this make this code obsolete? https://github.com/ckeditor/ckeditor5/blob/b699e0768f213a12705b203d2aa4ae6f13970c7b/packages/ckeditor5-core/src/plugincollection.ts#L346-L375 Can it be removed as well?

pomek:
During the arch guild, we were talking about "simplifying" plugins while keeping the soft-requirements concept as is. So, answering the question, I believe we want to save it.

psmyrek (approved):
LGTM 🎉 

I found one mention about soft requirements:
- https://github.com/ckeditor/ckeditor5/blob/363c4e560db1a179dc5f582f2733ff9060b411a1/docs/framework/contributing/package-metadata.md#L30

But since we keep the code for supporting this feature, I guess no change in the docs is needed.

PR #19838, opened 2026-02-24
Preserve HTML `<figure>` wrapper on content tables when layout tables plugin is enabled.

### LINKED ISSUE #19771: Allow preserving <figure> wrapper for content tables when using together with layout tables
Currently, when using both **content tables** and **layout tables** in the same editor setup, the `<figure>` element is stripped from tables in both cases.

This happens because layout tables require the `PlainTableOutput` plugin, which removes the `<figure>` wrapper from all tables globally. As a result, the wrapper is also removed from content tables.

It should be possible to:

- Keep PlainTableOutput behavior for layout tables
- Preserve the `<figure>` wrapper for content tables

In other words, we could add a configuration option that allows disabling `<figure>` stripping for content tables while still supporting layout tables.

The exact API shape and implementation approach should be discussed in this issue.

pszczesniak:
### 🚀 Summary

*Preserved the `<figure>` wrapper on content tables when `table.tableLayout.stripFigureFromContentTable` is set to `false` and the layout tables plugin is enabled.*

---

### 📌 Related issues

* Closes #19771

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

- [x] PR description explains the changes and the chosen approach (especially when  performance, API, or UX is affected).
- [x] The changelog entry is clear, user‑ or integrator-facing, and it describes any breaking changes.
- [x] The target branch is correct.

### DIFF
--- a/packages/ckeditor5-table/src/converters/downcast.ts
+++ b/packages/ckeditor5-table/src/converters/downcast.ts
@@ -253,11 +253,15 @@ export function convertPlainTable( editor: Editor ): DowncastElementCreatorFunct
 
 		const stripFigureTagWithLayoutTable = shouldStripFigureTagWithLayoutTable( editor, table );
 
-		if ( !hasPlainTableOutput && !stripFigureTagWithLayoutTable && !( useExtendedAlignment && isClipboardPipeline ) ) {
-			return null;
+		if (
+			hasPlainTableOutput ||
+			stripFigureTagWithLayoutTable ||
+			( useExtendedAlignment && isClipboardPipeline )
+		) {
+			return downcastPlainTable( table, conversionApi, editor );
 		}
 
-		return downcastPlainTable( table, conversionApi, editor );
+		return null;
 	};
 }
 
@@ -270,7 +274,15 @@ export function convertPlainTableCaption( editor: Editor ): DowncastElementCreat
 		const isClipboardPipeline = options.isClipboardPipeline;
 		const useExtendedAlignment = editor.config.get( 'experimentalFlags.useExtendedTableBlockAlignment' ) as boolean;
 
-		if ( !hasPlainTableOutput && !( useExtendedAlignment && isClipboardPipeline ) ) {
+		const stripFigureTagWithLayoutTable = shouldStripFigureTagWithLayoutTable( editor, modelElement );
+
+		if (
+			!(
+				hasPlainTableOutput ||
+				stripFigureTagWithLayoutTable ||
+				( useExtendedAlignment && isClipboardPipeline )
+			)
+		) {
 			return null;
 		}
 
@@ -396,9 +408,11 @@ export function downcastTableBorderAndBackgroundAttributes( editor: Editor ): vo
 				const stripFigureTagWithLayoutTable = shouldStripFigureTagWithLayoutTable( editor, item );
 
 				if (
-					!hasPlainTableOutput &&
-					!stripFigureTagWithLayoutTable &&
-					!( useExtendedAlignment && isClipboardPipeline )
+					!(
+						hasPlainTableOutput ||
+						stripFigureTagWithLayoutTable ||
+						( useExtendedAlignment && isClipboardPipeline )
+					)
 				) {
 					return;
 				}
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
```diff
@@ -1995,13 +1995,11 @@ describe( 'table properties', () => {
...
-									'<tbody>' +
-										'<tr><td>content table</td></tr>' +
-									'</tbody>' +
-								'</table>' +
-							'</figure>'
+							'<table class="table content-table" style="float:none;">' +
+								'<tbody>' +
+									'<tr><td>content table</td></tr>' +
+								'</tbody>' +
+							'</table>'
```
niegowski, line 2002: Why such a change? This is surprising as I'd expect no degradation in current content.

packages/ckeditor5-table/src/converters/downcast.ts
```diff
@@ -251,7 +251,9 @@ export function convertPlainTable( editor: Editor ): DowncastElementCreatorFunct
 		const isClipboardPipeline = conversionApi.options.isClipboardPipeline;
 		const useExtendedAlignment = editor.config.get( 'experimentalFlags.useExtendedTableBlockAlignment' ) as boolean;
 
-		if ( !hasPlainTableOutput && !( useExtendedAlignment && isClipboardPipeline ) ) {
+		const stripFigureTagWithLayoutTable = shouldStripFigureTagWithLayoutTable( editor, table );
+
+		if ( !hasPlainTableOutput && !stripFigureTagWithLayoutTable && !( useExtendedAlignment && isClipboardPipeline ) ) {
```
niegowski: This condition is hard to understand. Could we flip it? Would it make it more readable?

packages/ckeditor5-table/src/converters/downcast.ts
```diff
@@ -391,7 +393,13 @@ export function downcastTableBorderAndBackgroundAttributes( editor: Editor ): vo
...
 				const useExtendedAlignment = editor.config.get( 'experimentalFlags.useExtendedTableBlockAlignment' ) as boolean;
 
-				if ( !hasPlainTableOutput && !( useExtendedAlignment && isClipboardPipeline ) ) {
+				const stripFigureTagWithLayoutTable = shouldStripFigureTagWithLayoutTable( editor, item );
+
+				if (
+					!hasPlainTableOutput &&
+					!stripFigureTagWithLayoutTable &&
+					!( useExtendedAlignment && isClipboardPipeline )
+				) {
```
niegowski, line 416: Same as above. This condition is hard to understand. Could we flip it? Would it make it more readable?

packages/ckeditor5-table/src/converters/downcast.ts
```diff
@@ -411,6 +419,21 @@ export function downcastTableBorderAndBackgroundAttributes( editor: Editor ): vo
 	}
 }
 
+/**
+ * Returns `true` if the figure tag should be stripped when using layout tables and when `tableType` is `layout`
+ * or `stripFigureFromContentTable` option is set to `true`, `false` otherwise.
```
niegowski, line 438:
```suggestion
 * or `stripFigureFromContentTable` option is set to `true`, `false` otherwise.
 * 
```

packages/ckeditor5-table/tests/tableproperties/tablepropertiesediting.js
pszczesniak: Content tables should be stripped when layout tables are enabled, as the `LayoutTable` plugin depends on `PlainTableOutput`. Manual testing confirmed that after adding the `LayoutTable` glue plugin, the behavior is correct.

However, the referenced test relies on `TableLayoutEditing`, which does not require `PlainTableOutput`. As a result, the test setup does not reflect the actual integration scenario, making the tests incorrect.

packages/ckeditor5-table/src/converters/downcast.ts
pszczesniak: I've tune it a bit. It will be much more readable in v48.

packages/ckeditor5-table/src/converters/downcast.ts
pszczesniak: I've tune it a bit.

Issue #9398, opened 2021-03-31
Show blocks feature


Mgsy:
## 📝 Provide a description of the new feature

Implement a [show blocks feature](https://ckeditor.com/cke4/addon/showblocks) that works the same way as in CKEditor 4.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

oleq:
Note: Both CKE4 and TinyMCE use background images to display "P", "H2", "BLOCKQUOTE" in the blocks. 

I suppose they are used instead of `::after/::before` to avoid breaking the integrator's styles. After all, content like `<p>` or `<h2>` is unlikely to have a `background-image` but it may use `::after/::before` to display some extras.

oleq:
A solution to avoid a tedious generation of SVGs that takes advantage of our PostCSS system.

TODOs:
* final styling,
* [license of the font used by text-to-svg](https://www.npmjs.com/package/text-to-svg)

A new function mixin in `postcss-mixins`

```js
/**
 * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
 * For licensing, see LICENSE.md.
 */

'use strict';

const TextToSVG = require( 'text-to-svg' );
const textToSVG = TextToSVG.loadSync();

/* eslint-env node */

/**
 * Returns a PostCSS configuration to build the editor styles (e.g. used by postcss-loader).
 *
 * @param {Object} options
 * @param {Boolean} options.sourceMap When true, an inline source map will be built into the output CSS.
 * @param {Boolean} options.minify When true, the output CSS will be minified.
 * @param {ThemeImporterOptions} options.themeImporter Configuration of the theme-importer PostCSS plugin.
 * See the plugin to learn more.
 * @returns {Object} A PostCSS configuration object, e.g. to be used by the postcss-loader.
 */
module.exports = function getPostCssConfig( options = {} ) {
	const config = {
		plugins: [
			require( 'postcss-import' )(),
			require( './themeimporter' )( options.themeImporter ),
			require( 'postcss-mixins' )( {
				mixins: {
					'text-to-svg-background': function( mixin, text ) {
						const attributes = { fill: 'black' };
						const options = { x: 0, y: 0, fontSize: 10, anchor: 'top', attributes: attributes };
						const svg = textToSVG.getSVG( text, options );
						const base64svg = Buffer.from( svg ).toString('base64');

[...]

PR #19855, opened 2026-02-26
Replace PostCSS mixins with plain CSS alternatives

filipsobol:
### 🚀 Summary

Replace PostCSS mixins with plain CSS alternatives.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4263.

---

#### `stylelint-disable` usage

In a few places I've added the comments disabling the `no-descending-specificity` rule in stylelint. This was done to both make the diff much easier to read and understand and also to prevent accidental regressions.

#### Testing

The best way to test these changes is to build and run the documentation (`pnpm run docs --skip-api && pnpm run docs:serve`) and compare them with the production version at https://ckeditor.com/docs/ckeditor5/latest/index.html.

While the scope of the changes is big, most of them are trivial. The components that received non-trivial changes are the following:
* `ckeditor5-highlight`
* `ckeditor5-list` (todo list specifically)
* `ckeditor5-show-blocks`
* `ckeditor5-ui` (button, dropdown, splitbutton components specifically)

These plugins and components should be reviewed (both from a visual and implementation perspective) more thoroughly.

#### High-impact mixin migrations

<details>
<summary>ck-dir</summary>

```css
/* before */
@mixin ck-dir ltr { margin-left: 8px; }
@mixin ck-dir rtl { margin-right: 8px; }

/* after */
[dir="ltr"] & { margin-left: 8px; }
[dir="rtl"] & { margin-right: 8px; }
```

</details>

<details>
<summary>ck-media-phone</summary>

```css
/* before */
@mixin ck-media-phone { ... }

/* after */
@media screen and (max-width: 600px) { ... }
```

</details>

<details>
<summary>ck-button-colors</summary>

```css
/* before */
@mixin ck-button-colors --ck-color-button-default;

/* after */
background: var(--ck-color-button-default-background);
&:not(.ck-disabled) {
	&:hover { background: var(--ck-color-button-default-hover-background); }
	&:active { background: var(--ck-color-button-default-active-background); }
}
/* https://github.com/ckeditor/ckeditor5/issues/3376 */
&.ck-disabled { background: var(--ck-color-button-default-disabled-background); }
```

</details>

<details>
<summary>ck-focus-ring, ck-box-shadow, ck-drop-shadow </summary>

```css
/* before */
@mixin ck-focus-ring;
@mixin ck-box-shadow var(--ck-drop-shadow), var(--ck-focus-outer-shadow);

/* after */
outline: none;
border: var(--ck-focus-ring);
box-shadow: var(--ck-drop-shadow), var(--ck-focus-outer-shadow), 0 0;
```

</details>

<details>
<summary>ck-rounded-corners</summary>

```css
/* before */
.some-component {
	@mixin ck-rounded-corners;
}

/* after */
.some-component {
	border-radius: var(--ck-rounded-corners-radius);
}
```

</details>

#### One-off / localized mixin migrations

* `ck-media-forced-colors` / `ck-media-default-colors` → `@media (forced-colors: active|none)`.
* `ck-unselectable` → -`webkit-user-select: none; user-select: none;` (removed `-moz/-ms` prefixes as they are no longer needed).
* `ck-disabled` → `opacity: var(--ck-disabled-opacity);`.
* `ck-widget-type-around-button-visible/hidden` → explicit `opacity` + `pointer-events`.
* `highlight-marker-color` / `highlight-pen-color` → explicit static class rules in `packages/ckeditor5-highlight/theme/highlight.css`.
* `todo-list-checkbox` → shared plain selector block in `packages/ckeditor5-list/theme/todolist.css`.

---

### 🧾 Checklists

#### Author checklist

- [x] Is the changelog entry intentionally omitted?
- [x] Are there any changes the team should be informed about (e.g. architectural, difficult to revert in future versions or having impact on other features)?

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-show-blocks/package.json
+++ b/packages/ckeditor5-show-blocks/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@ckeditor/ckeditor5-core": "workspace:*",
-    "@ckeditor/ckeditor5-engine": "workspace:^",
+    "@ckeditor/ckeditor5-engine": "workspace:*",
     "@ckeditor/ckeditor5-icons": "workspace:*",
     "@ckeditor/ckeditor5-ui": "workspace:*",
     "@ckeditor/ckeditor5-utils": "workspace:*"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2955,7 +2955,7 @@ importers:
         specifier: workspace:*
         version: link:../ckeditor5-core
       '@ckeditor/ckeditor5-engine':
-        specifier: workspace:^
+        specifier: workspace:*
         version: link:../ckeditor5-engine
       '@ckeditor/ckeditor5-icons':
         specifier: workspace:*


packages/ckeditor5-ui/theme/components/button/switchbutton.css
```diff
@@ -52,21 +48,18 @@
...
 			/* Make sure the toggle is always to the left as far as possible. */
 			margin-right: auto;
 		}
 
 		& .ck-button__toggle__inner {
-			@mixin ck-rounded-corners {
-				border-radius: calc(.5 * var(--ck-border-radius));
-			}
-
+			border-radius: calc(.5 * var(--ck-rounded-corners-radius));
```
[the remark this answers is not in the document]
filipsobol: I double-checked this with AI:

> I checked, and this reviewer comment is a **false positive**.
> 
> - Old `ck-rounded-corners` mixin semantics were **not** “always use `--ck-border-radius`”.
> - Historically it started with `border-radius: 0`, and only under `.ck-rounded-corners` switched to `var(--ck-border-radius)` (then applied `@mixin-content`).
> - So this in `switchbutton`:
>   - old: `calc(.5 * var(--ck-border-radius))` **inside** `@mixin ck-rounded-corners { ... }`
>   - effectively meant: rounded only when `.ck-rounded-corners` is active.

The old implementation:

```css
@define-mixin ck-rounded-corners {
	border-radius: 0;

	.ck-rounded-corners &,
	&.ck-rounded-corners {
		border-radius: var(--ck-border-radius);
		@mixin-content;
	}
}
```

> With the new model:
> 
> - `packages/ckeditor5-ui/theme/globals/_rounded.css:26` sets `--ck-rounded-corners-radius: 0`.
> - `packages/ckeditor5-ui/theme/globals/_rounded.css:29` maps it to `var(--ck-border-radius)` when `.ck-rounded-corners` is present.
> - Therefore `calc(.5 * var(--ck-rounded-corners-radius))` is equivalent:
>   - no class → `calc(.5 * 0)` → `0`
>   - class present → `calc(.5 * var(--ck-border-radius))`
> 
> So behavior is preserved, not regressed.

The new implementation (the global part):

```css
:root {
	--ck-border-radius: 2px;
	--ck-rounded-corners-radius: 0;
}

.ck-rounded-corners {
	--ck-rounded-corners-radius: var(--ck-border-radius);
}
```

Then we add the following in the `switchbutton.css`:

```css
border-radius: calc(.5 * var(--ck-rounded-corners-radius));
```

packages/ckeditor5-show-blocks/theme/showblocks.css
```diff
@@ -3,129 +3,155 @@
...
 	& address {
-		@mixin block-name-background ADDRESS;
+		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>ADDRESS</text></svg>");
+		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>ADDRESS</text></svg>");
```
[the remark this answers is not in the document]
filipsobol: **This is false;** the output from the old implementation was larger—it was around 67 kB compared to 35 kB now.

While I also wanted to replace this with a much smaller `::before`-based implementation, the current solution was used over it intentionally (see https://github.com/ckeditor/ckeditor5/issues/9398#issuecomment-1511041695).

packages/ckeditor5-bookmark/theme/bookmarkform.css
```diff
@@ -3,13 +3,12 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-licensing-options
  */
 
-@import "@ckeditor/ckeditor5-ui/theme/mixins/_rwd.css";
 
 :root {
 	--ck-bookmark-form-width: 340px;
 }
 
-@mixin ck-media-phone {
```
oleq, line 12: If we're moving away from mixins, let's at least preserve this 600px value as a custom property so it can be reused across files.

packages/ckeditor5-bookmark/theme/bookmarktoolbar.css
```diff
@@ -22,6 +21,7 @@
 	text-align: center;
 	overflow: hidden;
 
-	@mixin ck-unselectable;
```
oleq, line 25: I'd turn this mixin into `.ck-unselectable` class and apply where relevant.

packages/ckeditor5-clipboard/theme/clipboard.css
```diff
@@ -93,14 +92,14 @@
 		height: 0;
 		border-style: solid;
 
-		@mixin ck-dir ltr {
+		[dir="ltr"] & {
```
oleq, line 95: For now, I think this is fine, but down the road we should get interested in 
* https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:dir
* using direction-aware properties (like `margin-inline-start`, `text-align: start`) that don't require integration with `dir` in the first place

filipsobol: This year we still target `Baseline 2022`. Next year we will update to `Baseline 2023` and be able to use `:dir()`.

packages/ckeditor5-ui/theme/components/button/button.css
```diff
@@ -141,15 +147,30 @@ a.ck.ck-button {
...
+
+			&:active {
+				background: var(--ck-color-button-on-active-background);
+			}
+		}
+
+		/* https://github.com/ckeditor/ckeditor5/issues/3376 */
+		&.ck-disabled {
+			background: var(--ck-color-button-on-disabled-background);
+		}
```
oleq: This repeats a lot of https://github.com/ckeditor/ckeditor5/pull/19855/changes#diff-2008d46f03fe2ec00d19eb84313a26b7266401a8c7c782014ce884e802b2534bR8-R18. I suppose we should have backgrounds set once using custom properties, and in variants (such as .ck-on) only change the values of those custom properties to other custom properties.

```css
--ck-color-button-default-background: var(--ck-color-button-default-background);
--ck-color-button-not-disabled-hover-background: var(--ck-color-button-default-hover-background);
--ck-color-button-not-disabled-active-background: var(--ck-color-button-default-active-background);

.ck-button {
    background: var(--ck-color-button-default-background);

	&:not(.ck-disabled) {
		&:hover {
			background: var(--ck-color-button-not-disabled-hover-background);
		}

		&:active {
			background: var(--ck-color-button-not-disabled-active-background);
		}
	}
}
```

so that the following is just

```css
.ck-button.ck-on {
    --ck-color-button-default-background: color for .ck-on state;
    --ck-color-button-not-disabled-hover-background: color for .ck-on state;
    --ck-color-button-not-disabled-active-background: color for .ck-on state;
}
```
without repeating selectors.

packages/ckeditor5-editor-classic/theme/classiceditor.css
```diff
@@ -3,18 +3,14 @@
...
 
 .ck.ck-editor__main > .ck-editor__editable {
 	/* https://github.com/ckeditor/ckeditor5/issues/3384 */
 	background: var(--ck-color-base-background);
-
-	@mixin ck-rounded-corners {
-		border-top-left-radius: 0;
-		border-top-right-radius: 0;
-	}
+	border-radius: var(--ck-rounded-corners-radius);
```
oleq, line 11: I suppose there should be `.ck-rounded-corners` class to be used whenever necessary.

packages/ckeditor5-engine/tests/manual/nestededitable.css
```diff
@@ -15,8 +13,9 @@ figure {
 		outline: none;
 
 		&.focused {
-			@mixin ck-focus-ring;
-			@mixin ck-box-shadow var(--ck-inner-shadow);
+			outline: none;
+			border: var(--ck-focus-ring);
+			box-shadow: var(--ck-inner-shadow), 0 0;
```
oleq, line 18: We have a couple of shadow cases in the UI: inner ones and outer ones. I suppose outer ones should be controlled by a CSS class shared between all panels, balloons, etc.

packages/ckeditor5-editor-classic/theme/classiceditor.css
filipsobol: This class is still necessary, and it sets the value for the `--ck-rounded-corners-radius` variable. See the changes in `packages/ckeditor5-ui/theme/globals/_rounded.css`.

It was necessary to both avoid repetition and preserve backward compatibility with the old selector.

packages/ckeditor5-bookmark/theme/bookmarkform.css
pszczesniak: Correct me if im wrong but there is no way to use CSS variables in @media queries 🤔

packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
```diff
@@ -35,13 +33,13 @@
...
 			& > * + * {
 				margin-left: var(--ck-spacing-standard);
 			}
 		}
 
-		@mixin ck-dir rtl {
+		[dir="rtl"] & {
 			& > * + * {
 				margin-right: var(--ck-spacing-standard);
 			}
```
pszczesniak, line 45: What do you think about removing the `[dir]` selector where possible? We could use logical properties like `margin-inline-start` / `margin-inline-end` instead of `margin-left` / `margin-right` (and the same for padding). That might help simplify the styles and better support RTL/LTR without extra selectors.

packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
```diff
@@ -78,11 +76,11 @@
...
 			color: var(--ck-color-base-border);
 
-			@mixin ck-dir ltr {
+			[dir="ltr"] & {
 				right: var(--ck-spacing-standard);
 			}
 
-			@mixin ck-dir rtl {
+			[dir="rtl"] & {
 				left: var(--ck-spacing-standard);
```
pszczesniak, line 84: Similar here - there are [`inset-inline-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline-start) and [`inset-inline-end`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline-end) that we could use.

packages/ckeditor5-bookmark/theme/bookmarkform.css
filipsobol: Exactly, https://bholmes.dev/blog/alternative-to-css-variable-media-queries/.

In the future, we may use the [`@custom-media`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@custom-media) rule, but for now we unfortunately may need to hardcode this value.

packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
filipsobol: I'm all for it, but I believe this is out of the scope of this task.

WDYT? If you agree, could you create a new issue for this?

packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
filipsobol: See https://github.com/ckeditor/ckeditor5/pull/19855#discussion_r2859007987.

packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
filipsobol: Also see https://github.com/ckeditor/ckeditor5/pull/19855#discussion_r2858768684.

packages/ckeditor5-ui/theme/components/button/button.css
filipsobol: Adressed in https://github.com/ckeditor/ckeditor5/pull/19855/commits/489652b95ad0156ef7a68231d3c56bdc38043fe4.

packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
pszczesniak: > WDYT? If you agree, could you create a new issue for this?

Fine for me 👍  

Will create a new issue for it.

filipsobol:
> [...] correct me if I'm wrong, I assume this PR is just the first step for us.
> 
> Getting it done right down the road coincides with the development of a design system and a design system handbook/component library that is organized and maintained.

Correct. While there's still some work to be done from the Platform team's side, modernization of the existing styles is outside our scope and will likely be tackled as part of other initiatives.

packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
filipsobol: https://github.com/ckeditor/ckeditor5/issues/19860

packages/ckeditor5-find-and-replace/theme/findandreplaceform.css
filipsobol: https://github.com/ckeditor/ckeditor5/issues/19860

packages/ckeditor5-clipboard/theme/clipboard.css
filipsobol: https://github.com/ckeditor/ckeditor5/issues/19860

packages/ckeditor5-bookmark/theme/bookmarktoolbar.css
filipsobol: I think it's slightly outside the scope of this task and more prone to regressions than adding these two CSS properties in 4 places.

packages/ckeditor5-engine/tests/manual/nestededitable.css
filipsobol: https://github.com/ckeditor/ckeditor5/issues/19861

packages/ckeditor5-show-blocks/theme/showblocks.css
pomek, line 1: - pnpm run manual -f show-blocks
- http://localhost:8125/ckeditor5-show-blocks/tests/manual/showblockseditortypes.html

Image

Image

In other words, this feature works only in ClassicEditor.

pomek, line 1: - [ ] **Blocker** → **Move show-block label generation to the view layer instead of keeping per-tag/per-direction CSS SVG duplication.**
  Why it matters: The linked issue calls this out as the key migration hotspot; leaving it in CSS keeps the same maintainability and bundle-size risk the task is supposed to remove.
  Smallest safe fix: Generate label SVGs from a single template in show-blocks view logic and keep CSS as a generic positioning/styling layer.

```diff
diff --git a/packages/ckeditor5-show-blocks/src/showblockscommand.ts b/packages/ckeditor5-show-blocks/src/showblockscommand.ts
index 26d9ce7892..964028b2c9 100644
--- a/packages/ckeditor5-show-blocks/src/showblockscommand.ts
+++ b/packages/ckeditor5-show-blocks/src/showblockscommand.ts
@@ -8,6 +8,42 @@
  */
 
 import { Command, type Editor } from '@ckeditor/ckeditor5-core';
+import type { ViewDowncastWriter, ViewRootEditableElement } from '@ckeditor/ckeditor5-engine';
+
+const CLASS_NAME = 'ck-show-blocks';
+
+const SHOW_BLOCKS_LABELS = [
+	[ 'address', 'ADDRESS' ],
+	[ 'aside', 'ASIDE' ],
+	[ 'blockquote', 'BLOCKQUOTE' ],
+	[ 'details', 'DETAILS' ],
+	[ 'div', 'DIV' ],
+	[ 'footer', 'FOOTER' ],
+	[ 'h1', 'H1' ],
+	[ 'h2', 'H2' ],
+	[ 'h3', 'H3' ],
+	[ 'h4', 'H4' ],
+	[ 'h5', 'H5' ],
+	[ 'h6', 'H6' ],
+	[ 'header', 'HEADER' ],
+	[ 'main', 'MAIN' ],
+	[ 'nav', 'NAV' ],
+	[ 'pre', 'PRE' ],
+	[ 'ol', 'OL' ],
+	[ 'ul', 'UL' ],
+	[ 'p', 'P' ],
+	[ 'section', 'SECTION' ],
+	[ 'figcaption', 'FIGCAPTION' ]
+] as const;
+
+type ShowBlocksLabelDirection = 'ltr' | 'rtl';
+
+const SHOW_BLOCKS_LABEL_SVG_TEMPLATE: Record<ShowBlocksLabelDirection, string> = {
+	ltr: "<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>__TEXT__</text></svg>",
+	rtl: "<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>__TEXT__</text></svg>"
+};
+
+const SHOW_BLOCKS_LABEL_CACHE = new Map<string, string>();
 
 /**
  * The show blocks command.
@@ -36,13 +72,13 @@ export class ShowBlocksCommand extends Command {
 	 * Toggles the visibility of content blocks.
 	 */
 	public override execute(): void {
-		const CLASS_NAME = 'ck-show-blocks';
 		const view = this.editor.editing.view;
 
 		view.change( writer => {
 			// Multiroot support.
 			for ( const root of view.document.roots ) {
 				if ( !root.hasClass( CLASS_NAME ) ) {
+					setBlockLabelStyles( writer, root );
 					writer.addClass( CLASS_NAME, root );
 					this.value = true;
 				} else {
@@ -53,3 +89,28 @@ export class ShowBlocksCommand extends Command {
 		} );
 	}
 }
+
+function setBlockLabelStyles( writer: ViewDowncastWriter, root: ViewRootEditableElement ): void {
+	if ( root.getStyle( '--ck-show-blocks-label-address-ltr' ) ) {
+		return;
+	}
+
+	for ( const [ key, label ] of SHOW_BLOCKS_LABELS ) {
+		writer.setStyle( `--ck-show-blocks-label-${ key }-ltr`, getLabelUrl( label, 'ltr' ), root );
+		writer.setStyle( `--ck-show-blocks-label-${ key }-rtl`, getLabelUrl( label, 'rtl' ), root );
+	}
+}
+
+function getLabelUrl( label: string, direction: ShowBlocksLabelDirection ): string {
+	const cacheKey = `${ direction }:${ label }`;
+
+	if ( SHOW_BLOCKS_LABEL_CACHE.has( cacheKey ) ) {
+		return SHOW_BLOCKS_LABEL_CACHE.get( cacheKey )!;
+	}
+
+	const labelUrl = `url("data:image/svg+xml;utf8,${ SHOW_BLOCKS_LABEL_SVG_TEMPLATE[ direction ].replace( '__TEXT__', label ) }")`;
+
+	SHOW_BLOCKS_LABEL_CACHE.set( cacheKey, labelUrl );
+
+	return labelUrl;
+}
diff --git a/packages/ckeditor5-show-blocks/tests/showblockscommand.js b/packages/ckeditor5-show-blocks/tests/showblockscommand.js
index 34227a0c16..09a946bbb8 100644
--- a/packages/ckeditor5-show-blocks/tests/showblockscommand.js
+++ b/packages/ckeditor5-show-blocks/tests/showblockscommand.js
@@ -59,6 +59,18 @@ describe( 'ShowBlocksCommand', () => {
 			expect( editor.editing.view.document.roots.get( 'main' ).hasClass( 'ck-show-blocks' ) ).to.be.true;
 		} );
 
+		it( 'should set block label styles on the root when executed for the first time', () => {
+			editor.execute( 'showBlocks' );
+
+			const root = editor.editing.view.document.roots.get( 'main' );
+			const paragraphLabelLtr = root.getStyle( '--ck-show-blocks-label-p-ltr' );
+			const headingLabelRtl = root.getStyle( '--ck-show-blocks-label-h2-rtl' );
+
+			expect( paragraphLabelLtr ).to.contain( 'data:image/svg+xml;utf8' );
+			expect( paragraphLabelLtr ).to.contain( '>P</text></svg>' );
+			expect( headingLabelRtl ).to.contain( '>H2</text></svg>' );
+		} );
+
 		it( 'should remove "ck-show-blocks" class on the root when executed for the second time', () => {
 			editor.execute( 'showBlocks' );
 			editor.execute( 'showBlocks' );
@@ -122,6 +134,17 @@ describe( 'ShowBlocksCommand', () => {
 				}
 			} );
 
+			it( 'should apply block label styles to all roots', () => {
+				multirootEditor.execute( 'showBlocks' );
+
+				for ( const root of multirootEditor.editing.view.document.roots ) {
+					expect(
+						root.getStyle( '--ck-show-blocks-label-blockquote-ltr' ),
+						`Label style was not set properly on ${ root.rootName } root`
+					).to.contain( '>BLOCKQUOTE</text></svg>' );
+				}
+			} );
+
 			it( 'should remove class from all roots', () => {
 				multirootEditor.execute( 'showBlocks' );
 				multirootEditor.execute( 'showBlocks' );
diff --git a/packages/ckeditor5-show-blocks/theme/showblocks.css b/packages/ckeditor5-show-blocks/theme/showblocks.css
index e7c520100f..6bedc65b2b 100644
--- a/packages/ckeditor5-show-blocks/theme/showblocks.css
+++ b/packages/ckeditor5-show-blocks/theme/showblocks.css
@@ -9,108 +9,108 @@
 
 .ck.ck-editor__editable.ck-editor__editable_inline.ck-show-blocks:not(.ck-widget) {
 	& address {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>ADDRESS</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>ADDRESS</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-address-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-address-rtl);
 	}
 
 	& aside {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>ASIDE</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>ASIDE</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-aside-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-aside-rtl);
 	}
 
 	& blockquote {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>BLOCKQUOTE</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>BLOCKQUOTE</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-blockquote-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-blockquote-rtl);
 	}
 
 	& details {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>DETAILS</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>DETAILS</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-details-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-details-rtl);
 	}
 
 	& div:not(.ck-widget, .ck-widget *) {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>DIV</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>DIV</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-div-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-div-rtl);
 	}
 
 	& footer {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>FOOTER</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>FOOTER</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-footer-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-footer-rtl);
 	}
 
 	& h1 {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H1</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H1</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-h1-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-h1-rtl);
 	}
 
 	& h2 {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H2</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H2</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-h2-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-h2-rtl);
 	}
 
 	& h3 {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H3</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H3</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-h3-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-h3-rtl);
 	}
 
 	& h4 {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H4</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H4</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-h4-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-h4-rtl);
 	}
 
 	& h5 {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H5</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H5</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-h5-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-h5-rtl);
 	}
 
 	& h6 {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H6</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>H6</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-h6-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-h6-rtl);
 	}
 
 	& header {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>HEADER</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>HEADER</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-header-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-header-rtl);
 	}
 
 	& main {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>MAIN</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>MAIN</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-main-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-main-rtl);
 	}
 
 	& nav {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>NAV</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>NAV</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-nav-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-nav-rtl);
 	}
 
 	& pre {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>PRE</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>PRE</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-pre-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-pre-rtl);
 	}
 
 	& ol {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>OL</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>OL</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-ol-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-ol-rtl);
 	}
 
 	& ul {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>UL</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>UL</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-ul-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-ul-rtl);
 	}
 
 	& p {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>P</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>P</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-p-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-p-rtl);
 	}
 
 	& section {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>SECTION</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>SECTION</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-section-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-section-rtl);
 	}
 
 	& :where(figure.image, figure.table) figcaption {
-		--ck-show-blocks-label-ltr: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='3' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>FIGCAPTION</text></svg>");
-		--ck-show-blocks-label-rtl: url("data:image/svg+xml;utf8,<svg width='120' height='12' xmlns='http://www.w3.org/2000/svg' ><text style='paint-order:stroke fill; clip-path: inset(-3px); transform:translate(-2px, 0)' stroke='%23EAEAEA' stroke-width='13' dominant-baseline='middle' fill='black' x='100%' text-anchor='end' y='7' font-size='9px' font-family='Consolas, %22Lucida Console%22, %22Lucida Sans Typewriter%22, %22DejaVu Sans Mono%22, %22Bitstream Vera Sans Mono%22, %22Liberation Mono%22, Monaco, %22Courier New%22, Courier, monospace'>FIGCAPTION</text></svg>");
+		--ck-show-blocks-label-ltr: var(--ck-show-blocks-label-figcaption-ltr);
+		--ck-show-blocks-label-rtl: var(--ck-show-blocks-label-figcaption-rtl);
 	}
 
 	& address,
@@ -137,21 +137,41 @@
 	{
 		background-repeat: no-repeat;
 		padding-top: 15px;
-
-		/* Fix for Multi-root editor
-		https://github.com/ckeditor/ckeditor5/issues/15969 */
-		[dir="ltr"] & {
-			background-image: var(--ck-show-blocks-label-ltr);
-			background-position: 1px 1px;
-		}
-
-		[dir="rtl"] & {
-			background-image: var(--ck-show-blocks-label-rtl);
-			background-position: calc(100% - 1px) 1px;
-		}
+		background-image: var(--ck-show-blocks-label-ltr);
+		background-position: 1px 1px;
 
 		&:not(.ck-widget_selected):not(.ck-widget:hover) {
 			outline: 1px dashed var(--ck-show-blocks-border-color);
 		}
 	}
 }
+
+/* Fix for Multi-root editor
+https://github.com/ckeditor/ckeditor5/issues/15969 */
+[dir="rtl"] .ck.ck-editor__editable.ck-editor__editable_inline.ck-show-blocks:not(.ck-widget),
+.ck.ck-editor__editable.ck-editor__editable_inline.ck-show-blocks:not(.ck-widget)[dir="rtl"] {
+	& address,
+	& aside,
+	& blockquote,
+	& details,
+	& div:not(.ck-widget, .ck-widget *),
+	& footer,
+	& h1,
+	& h2,
+	& h3,
+	& h4,
+	& h5,
+	& h6,
+	& header,
+	& main,
+	& nav,
+	& pre,
+	& ol,
+	& ul,
+	& p,
+	& section,
+	& :where(figure.image, figure.table) figcaption {
+		background-image: var(--ck-show-blocks-label-rtl);
+		background-position: calc(100% - 1px) 1px;
+	}
+}
```

But, since the feature does not work on other editors, I cannot say the change makes sense.

pszczesniak: it's casued by the `dir` attribute. In `ClassicEdtitor` there is also a wrapper with a `dir` attribute so this:

https://github.com/ckeditor/ckeditor5/blob/489652b95ad0156ef7a68231d3c56bdc38043fe4/packages/ckeditor5-show-blocks/theme/showblocks.css#L143-L151

will match and and because of this it works.

In other editors there is no such a wrapper, the `dir` attribute is on same level as `.ck.ck-editor__editable.ck-editor__editable_inline.ck-show-blocks:not(.ck-widget)`

The selector should look like this:

```css
.ck.ck-editor__editable.ck-editor__editable_inline.ck-show-blocks:not(.ck-widget)[dir="ltr"] {
   /* HTML tag element selectors with LTR stuff */
}

.ck.ck-editor__editable.ck-editor__editable_inline.ck-show-blocks:not(.ck-widget)[dir="rtl"] {
   /* HTML tag element selectors with RTL stuff */
}
```

pomek: It turns out that on `#master` it does not work as well. At least, in a manual test.

filipsobol: Addressed in https://github.com/ckeditor/ckeditor5/pull/19855/commits/66acb8421302510890aa5d63bcb58bd207528bbb.

pomek: @filipsobol, could you report a bug with these details? I would like to track it.

pomek (approved):
LGTM.

packages/ckeditor5-show-blocks/theme/showblocks.css
filipsobol: https://github.com/ckeditor/ckeditor5/issues/19866

PR #19893, opened 2026-03-04
Cursor should not jump over paragraph when navigating with arrow up keys

niegowski:
### 🚀 Summary

Cursor should not jump over paragraph when navigating with arrow up keys,

---

### 📌 Related issues

* Closes #19812

---

### 💡 Additional information

*Optional: Notes on decisions, edge cases, or anything helpful for reviewers.*

---

### 🧾 Checklists

Use the following checklists to ensure important areas were not overlooked.
This does not apply to feature-branch merges.
If an item is **not relevant** to this type of change, simply leave it unchecked.

#### Author checklist

- [ ] Is the changelog entry intentionally omitted?
- [x] Is the change backward-compatible?
- [x] Have you considered the impact on different editor setups and core interactions? _(e.g., classic/inline/multi-root/many editors, typing, selection, paste, tables, lists, images, collaboration, pagination)_
- [x] Has the change been manually verified in the relevant setups?
- [x] Does this change affect any of the above?
- [ ] Is performance impacted?
- [x] Is accessibility affected?
- [x] Have tests been added that fail without this change (against regression)?
- [ ] Have the API documentation, guides, feature digest, and related feature sections been updated where needed?
- [ ] Have metadata files (ckeditor5-metadata.json) been updated if needed?
- [ ] Are there any changes the team should be informed about (e.g. architectural, difficult to revert in future versions or having impact on other features)?
- [ ] Were these changes documented (in Logbook)?

#### Reviewer checklist

- [x] PR description explains the changes and the chosen approach (especially when  performance, API, or UX is affected).
- [x] The changelog entry is clear, user‑ or integrator-facing, and it describes any breaking changes.
- [ ] All new external dependencies have been approved and mentioned in LICENSE.md (if any).
- [ ] All human-readable, translateable strings in this PR been introduced using `t()` (if any).
- [ ] I manually verified the change (e.g., in manual tests or documentation).
- [x] The target branch is correct.

PR #19917, opened 2026-03-06
Fix markers OT for a scenario when multi-element marker is removed and then restored using undo

### LINKED ISSUE #19916: Markers are incorrectly restored on undo in some specific cases

1. Use a sample with track changes and a few paragraphs.
2. Turn on track changes.
4. Make a selection starting in the middle of one paragraph, and ending in the middle of third paragraph.
5. Press backspace to create a suggestion.
6. Accept the suggestion.
7. Undo

### ✔️ Expected result

Editor state is correctly restored. Especially, the marker for the deletion suggestion is as it was before accepting the suggestion.

### ❌ Actual result

Deletion suggestion marker is not correctly restored.

This affects track changes in a clear way, and also affects comments in similar scenarios.

This is a regression introduced in v43.2.0.

scofalik:
### 🚀 Summary

**TLDR:**

OT uses `ModelRange._createFromRanges()` to evaluate final marker range for marker operation, after a marker operation is fully processed.

This method prohibits passing intersecting ranges, as it works incorrectly in such case. However, in scenario described in the issue, the OT algorithm used the method on ranges array, where one range contained other range. The fix is to filter out such ranges.

**Broader explanation:**

When restoring a marker range during OT, the original marker operation may be split into multiple smaller markers operations. A lot of times during undo, these small markers ends up next to each other after all transformations are done (we actually expect that in most cases if everything went well). The OT mechanism, at the end of processing, joins these small parts together, if they ended up next to each other.

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/model/operation/transform.ts#L377

In this case, the original marker, e.g. `[ 0, 20 ] - [ 2, 10 ]`, is split into four markers during the processing, as removing such a range takes place in three steps. At the end of the processing, we get such partial-markers, as expected:

```
// From the original operation:
[ 0, 30 ] - [ 2, 0 ] // Assuming there were 30 characters total in the first paragraph.

// Created due to parts of the range being (re)moved:
[ 0, 20 ] - [ 0, 30 ] // Some text in the first paragraph.
[ 1 ] - [ 2 ] // The middle paragraph.
[ 2, 0 ] - [ 2, 10 ] // Part in the third paragraph.
```

When you sort them, you get such ranges:

```
[ 0, 20 ] - [ 0, 30 ]
[ 0, 30 ] - [ 2, 0 ]
[ 1 ] - [ 2 ]
[ 2, 0 ] - [ 2, 10 ]
```

As you can see, they are continuous and can be join together to the original `[ 0, 20 ] - [ 2, 10 ]` marker.

The problem is with `[ 1 ] - [ 2 ]` marker which is included in the marker before. It makes sense that OT produced such marker, and since it is contained in another marker, from logical point of view, it does not affect the "total range" of the marker. It's simply unnecessary information.

To join ranges, we use `ModelRange._createFromRanges()`. Unfortunately, due to how this function works, it does not take intersecting ranges. This is specified in the [method's API docs](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/model/range.ts#L932).

Because we pass `[ 1 ] - [ 2 ]` range, we make the algorithm [return too soon](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-engine/src/model/range.ts#L1000), and the last range (`[ 2, 0 ] - [ 2, 10 ]`) does not have a chance to be joined to the original range. **Due to that, the final result of OT is `[ 0, 20 ] - [ 2, 0 ]`. As a result, the marker is incorrectly restored (the part in the last paragraph is missing).**

I initially wanted to fix this in `ModelRange._createFromRanges()`. However, since the API docs specify that this method does not take intersecting ranges, I decided to respect that, and applied the fix directly in OT. I decided to filter out the ranges that are already contained in the original range. This is exactly the described scenario.

This way, the fix has a smaller splash zone.

However, I believe it would be the best at some point to improve `_createFromRanges()` so that it can take any ranges. This improvement will not be difficult and will not affect performance in a significant way. At the same, it may help fix some unknown bugs in other places which use `_createFromRanges()` and did not consider if input has intersecting ranges.

---

### 📌 Related issues

* Closes #19916 

---

### 🧾 Checklists

#### Author checklist

- [X] Is the change backward-compatible?
- [X] Has the change been manually verified in the relevant setups?
- [X] Have tests been added that fail without this change (against regression)?

#### Reviewer checklist

- [x] PR description explains the changes and the chosen approach (especially when  performance, API, or UX is affected).
- [x] The changelog entry is clear, user‑ or integrator-facing, and it describes any breaking changes.
- [x] I manually verified the change (e.g., in manual tests or documentation).
- [x] The target branch is correct.

### DIFF
--- a/packages/ckeditor5-engine/tests/model/operation/transform/undo.js
+++ b/packages/ckeditor5-engine/tests/model/operation/transform/undo.js
@@ -541,7 +541,7 @@ describe( 'transform', () => {
 		} );
 
 		// https://github.com/ckeditor/ckeditor5/issues/19916
-		it( 'multi-element marker moved then undo', () => {
+		it( 'multi-element marker moved then undo redo', () => {
 			john.setData( '<paragraph>Fo[o</paragraph><paragraph>Bar</paragraph><paragraph>Ab]c</paragraph>' );
 
 			john.setMarker( 'm1' );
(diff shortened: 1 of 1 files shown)


pomek:
Target the `#release` branch, please. Do not forget to rebase.

scofalik:
Changed target, squashed commits, rebeased onto release, force pushed.

pomek:
And, due to red CI, please take a look here: https://github.com/ckeditor/ckeditor5/pull/19893#issuecomment-4010770816.

The mentioned PR will most probably be merged on Monday.

scofalik:
Done.

scofalik:
I added a few more unit tests.

While doing this, I found out one interesting case.

Due to how deletion works when selection is over multiple elements (it's a multi-step process), for a case when selection is bigger than the marker:

image

We end up with a collapsed marker, rather than removed marker (which may be intuitive).

I added a test case for this with `.skip()`.

PR #19918, opened 2026-03-06
Fixed selection attributes handling after softBreak.

### LINKED ISSUE #1068: Should `br` break inline elements?

Other

At this moment if you press <kbd>Shift</kbd> + <kbd>Enter</kbd> inside an inline element, `<br>` element won't be inserted as a chidren of the inline element. The `<br>` will break the inline element:

image

I'm not sure how it should behave. Is it correct? What if we would like to have the same link in both lines? Editing inline styles (like bold, highlights, font size, etc.) is not a problem. The issue appears if we would like to edit links that are split by the `<br>` element. But there is the separate ticket for it – https://github.com/ckeditor/ckeditor5-link/issues/192.

### LINKED ISSUE #19853: Double Shift+Enter should preserve selection attributes.

1. Write some text with `bold` (any basic formatting or font properties).
2. Type `Shift+Enter` twice.
3. Continue writing text.

### ✔️ Expected result

Text written in 3rd step should be `bold`.

### ❌ Actual result

Text written in 3rd step is not `bold`.

### ❓ Possible solution

## 📃 Other details

* Browser: …
* OS: …

---

arkflpc:
### 🚀 Summary

Improved document selection attribute inheritance around `<softBreak>` so returning the caret after a soft break restores expected formatting.

When selection attributes are recalculated across `<softBreak>`, only attributes marked with `copyOnEnter` are inherited. Other inline non-object elements still act as hard boundaries.

---

### 📌 Related issues

* Closes #19853.
* Closes #1068.

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist


Mati365:
@arkflpc Please merge recent master, or cherry-pick https://github.com/ckeditor/ckeditor5/pull/19893/changes/1d846d2378158118a61eb65d5a0ecccd275fe1aa in order to fix failing manual tests.

packages/ckeditor5-engine/src/model/documentselection.ts
```diff
@@ -1247,37 +1238,83 @@ class LiveSelection extends ModelSelection {
...
 
-	if ( !schema.isInline( node ) ) {
-		return null;
+	let crossedSoftBreak = false;
+
+	for ( let node: ModelNode | null = startNode; node; node = getNextNode( node ) ) {
+		if ( node.is( 'element', 'softBreak' ) ) {
+			crossedSoftBreak = true;
+			continue;
+		}
```
Mati365: I'm not sure whether the engine should be aware of the existence of the `softBreak` element. It's registered in an external plugin. Would a schema flag be better? What do you think?

Mati365: We discussed with @niegowski, and it might be good idea to store some info in schema / attribute properties. However, there might be another workaround related to storing attributes on soft-breaks itself. We need to discuss it a little bit more.

PR #19984, opened 2026-03-20
Disabled mention, emoji and slash commands in code blocks.

### LINKED ISSUE #19146: Disabling mentions (slash commands, emoji) when in a code block
## 📝 Ask a question

Some of our users on TriliumNext/Trilium have complained about the slash commands and emoji functionality getting in the way while editing code blocks.

We would like to disable the mention plugin while the cursor is under a code block. What's the best approach to doing so?

I've tried:

* `forceDisable`ing the mention plugins.
* Listening to the key and preventing the event (which causes the character not to be inserted; even if I add it manually then the slash commands/mentions still show up).

martnpaneq:
### 🚀 Summary

Disabled mention, emoji and slash commands in code blocks, because code blocks are designed to contain plain, unformatted code where rich-text features like autocompletion and formatting are not applicable.

Before:
image

After:
image

---

### 📌 Related issues

* Closes #19146 

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

- [x] Is the changelog entry intentionally omitted? (Changelog entry included.)
- [x] Is the change backward-compatible? ( Yes. Mentions were never useful in code blocks, so disabling them there is purely additive behavior. No existing functionality breaks.)
- [x] Have you considered the impact on different editor setups and core interactions? _(e.g., classic/inline/multi-root/many editors, typing, selection, paste, tables, lists, images, collaboration, pagination)_ (  The change is at the schema level, so it works universally across classic/inline/multi-root/decoupled editors. No impact on
  typing, selection, paste, tables, lists, images, collaboration, or pagination. Text input in code blocks still works — only
  the mention autocomplete popup is suppressed.)
- [x] Has the change been manually verified in the relevant setups? (Yes, it has.)
- [x] Does this change affect any of the above? (No.)
- [x] Is performance impacted? (  No. The attribute check is a lightweight callback that only runs for the mention attribute - thanks to the second parameter.)
- [x] Is accessibility affected? (No.)
- [x] Have tests been added that fail without this change (against regression)? (Yes — two tests were added.)
- [x] Have the API documentation, guides, feature digest, and related feature sections been updated where needed? (Docs were updated.)
- [x] Have metadata files (ckeditor5-metadata.json) been updated if needed? (Not needed.)
- [x] Are there any changes the team should be informed about (e.g. architectural, difficult to revert in future versions or having impact on other features)? (Good to notice that mention not working in code blocks is the expected behaviour - not a bug.)
- [x] Were these changes documented (in Logbook)? (No.)

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-mention/tests/mentionui.js
+++ b/packages/ckeditor5-mention/tests/mentionui.js
@@ -715,7 +715,7 @@ describe( 'MentionUI', () => {
 				_setModelData( codeBlockModel, '<codeBlock language="plaintext">foo []</codeBlock>' );
 
 				codeBlockModel.change( writer => {
-					writer.insertText( '@Ka', codeBlockModel.document.selection.getFirstPosition() );
+					writer.insertText( '@Ba', codeBlockModel.document.selection.getFirstPosition() );
 				} );
 
 				await waitForDebounce();


martnpaneq:
Docs for `slash-commands` updated here: https://github.com/ckeditor/ckeditor5-commercial/pull/9554.

.changelog/20260320114754_ck_19146_disable_mentions_like_in_code_blocks.md
```diff
@@ -0,0 +1,9 @@
+---
+type: Fix
+scope:
+  - ckeditor5-mention
```
pomek, line 4: I would also add the integration packages:

- ckeditor5-code-block
- ckeditor5-emoji
- ckeditor5-slash-command

pomek:
Could we add a single integration test per affected package that checks if the mention dropdown can be activated in a code block element?

- emoji: type `:ro` (🚀)
- slash command: type `/he` (_heading_)
- mention: type `@ka` (or other letters that match an entry).

This would be the final confirmation.

pomek:
Is there any option to restore the behavior before this change? Example: https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-ModelSchema.html#event-checkAttribute.

martnpaneq:
> Is there any option to restore the behavior before this change? Example: https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-ModelSchema.html#event-checkAttribute.

Yes, users can restore the previous behavior by listening to the `Schema#checkAttribute` event and re-allowing the mention attribute in code blocks:

```js
  editor.model.schema.on( 'checkAttribute', ( evt, args ) => {
        const context = args[ 0 ];
        const attributeName = args[ 1 ];

        if ( attributeName === 'mention' && context.endsWith( 'codeBlock $text' ) ) {
                evt.stop();
                evt.return = true;
        }
  }, { priority: 'high' } );
```

The high priority ensures this listener runs before our addAttributeCheck callback (which uses default priority), effectively overriding the restriction.

martnpaneq:
I created a followup, with a potential fix already suggested. I decided to not include the fix here, because it would cause the discussion to grow unnecessarily.

https://github.com/ckeditor/ckeditor5/issues/19992

.changelog/20260320114754_ck_19146_disable_mentions_like_in_code_blocks.md
martnpaneq: Updated here [8ab9d41](https://github.com/ckeditor/ckeditor5/pull/19984/commits/8ab9d41f41598797388f5c80adc8c60278284dfd).

packages/ckeditor5-mention/tests/mentionui.js
```diff
@@ -698,6 +700,33 @@ describe( 'MentionUI', () => {
...
+					mention: staticConfig
+				} );
+
+				const codeBlockModel = codeBlockEditor.model;
+				const codeBlockPanelView = codeBlockEditor.plugins.get( ContextualBalloon ).view;
+
+				_setModelData( codeBlockModel, '<codeBlock language="plaintext">foo []</codeBlock>' );
+
+				codeBlockModel.change( writer => {
+					writer.insertText( '@Ka', codeBlockModel.document.selection.getFirstPosition() );
```
pomek:
```suggestion
					writer.insertText( '@Ba', codeBlockModel.document.selection.getFirstPosition() );
```

There is no mention starting with `@Ka`, sorry.

https://github.com/ckeditor/ckeditor5/blob/1a205b514b559efc1dbce1d07d1e0c7df370a44b/packages/ckeditor5-mention/tests/mentionui.js#L27

martnpaneq: Fixed in [52f79b9](https://github.com/ckeditor/ckeditor5/pull/19984/commits/52f79b94f516640b1926b699463fd7525ae28c85).

Issue #19994, opened 2026-03-23
Allowing list attributes on text nodes causes editor crash


niegowski:
## 📝 Provide detailed reproduction steps (if any)

1. Enable all attributes on `$text`:
  ```js
editor.model.schema.addAttributeCheck( context => {
	if( context.endsWith( '$text' ) ) {
		return true;
	}
} );
```
2. Create a list item and try to type in it.

### ✔️ Expected result

List attributes are not applied to text nodes. The editor does not try to downcast unexpected attributes.

### ❌ Actual result

Editor crashes while trying to downcast list attribute applied to text.

### ❓ Possible solution

The `insertContent` listener causes this: 

https://github.com/ckeditor/ckeditor5/blob/d3ba0536823b4e0510a1d0dd6df81a4cded9a66e/packages/ckeditor5-list/src/list/listediting.ts#L568

It tries to align pasted content to the list where it is pasted into to avoid breaking existing lists.

While typing, a `model.insertContent()` is called, and the above listener triggers, since it detects that it is inside a list item, it tries to add list attributes to every node that accepts it. Since the schema is invalid, as it allows all attributes on text, it applies those attributes to text.

We could add a check in the https://github.com/ckeditor/ckeditor5/blob/d3ba0536823b4e0510a1d0dd6df81a4cded9a66e/packages/ckeditor5-list/src/list/listediting.ts#L901-L903 to accept only elements for this fixing, as otherwise it does not make sense, and relying on the Schema itself could be problematic in such cases.

## 📃 Other details

* Browser: …
* OS: …
* First affected CKEditor version: …
* Installed CKEditor plugins: …

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

PR #20002, opened 2026-03-25
Prevent list attributes from being applied to text nodes

niegowski:
### 🚀 Summary

Prevent list attributes from being applied to text nodes and inline objects during content insertion, which could crash the editor when a permissive `Schema#addAttributeCheck()` is used.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/19994

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/.changelog/20260325155747_ck_19994.md
+++ b/.changelog/20260325155747_ck_19994.md
@@ -1,0 +1,9 @@
+---
+type: Other
+scope:
+  - ckeditor5-list
+closes:
+  - https://github.com/ckeditor/ckeditor5/issues/19994
+---
+
+Prevent list attributes from being applied to text nodes and inline objects during content insertion, which could crash the editor when a permissive `Schema#addAttributeCheck()` is used.


niegowski:
While working on this issue, I noticed yet another problem with allowing all attributes on text. When all attributes are allowed on `$text` (by the `Schema`), the selection placed after an inline object inherits its attributes. For example:

```xml
<paragraph><imageInline src="abc"/><$text src="abc">[]</$text></paragraph>
```

This causes the `src` attribute from the inline object to be applied to text typed right after it. If there were also a generic downcast converter, it would downcast to something like `<span src="abc">`, which does not make any sense.

packages/ckeditor5-list/tests/list/integrations/clipboard.js
```diff
@@ -428,6 +428,86 @@ describe( 'ListEditing integrations: clipboard copy & paste', () => {
...
+			);
+		} );
+
+		// https://github.com/ckeditor/ckeditor5/issues/19994
+		it( 'should not apply list attributes to an inline object even if schema allows all attributes on it', () => {
+			model.schema.addAttributeCheck( context => {
+				if ( context.endsWith( '$text' ) ) {
+					return true;
+				}
+			} );
```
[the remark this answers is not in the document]
niegowski: That was made on purpose to verify if allowing all attributes on text does not lead to inline objects, which are defined as allow-attributes-of: text.

.changelog/20260325155747_ck_19994.md
```diff
@@ -0,0 +1,9 @@
+---
+type: Other
```
mmotyczynska, line 2: I'd consider this a `Fix` as it was causing an editor crash.

niegowski: But to happen, it required an invalid schema configuration.

mmotyczynska: Fair enough 👍

PR #20011, opened 2026-03-26
Mention plugin should not generate random model attribute values

### LINKED ISSUE #20010: Mention feature generates random IDs on upcast, leading to model mismatch problems

Currently, on upcast and on creation, mentions generate a random ID and set it into model as `modelAttribute.uid`. This is necessary, to make sure that two mentions next to each other won't get merged into one.

The problem is that these IDs are not passed to data. So, upcasting same HTML will result in different models (as the attribute is randomly generated).

This violates one of our main rules and directly leads to crashes when using CKEditor AI.

### ❓ Possible solution

The easiest solution is to store these UIDs in data, and read them on upcast, to maintain stability between multiple data loads.

It is a bit of a breaking change. We even have guides showing how to customize that converter. Well, if people won't update, they won't have a regression, but they will still encounter crashes when using AI.

Additionally, we will need to have clipboard handling -- to prevent duplication on copy-paste we need to remove these IDs when clipboard is used.

---

A different approach would be to remove UIDs entirely, but then, of course, the view attribute elements will get merged. To prevent that, we could create a new view element custom property, e.g. `preventMerging`:

```
writer.setCustomProperty( 'preventMerging', true, mentionElement );
```

Then, in the writer, in few places where we check `ViewAttribute#isSimilar` we would need to check also check this property.

This change may seem a bit broader, but will lead to a cleaner model and cleaner HTML output.

This would be a more impactful breaking change for integrators that overwrite the converter, as `modelAttribute.uid` will no longer be available, so their mentions will merge or the editor will crash.

scofalik:
### 🚀 Summary

Instead of generating random values for mentions attributes in model on upcast, store UIDs in data and read them from data on upcast.

---

### 📌 Related issues

* Closes #20010

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist


packages/ckeditor5-mention/src/mentionediting.ts
```diff
@@ -94,7 +94,7 @@ export function _addMentionAttributes(
 	baseMentionData: { id: string; _text: string },
 	data?: Record<string, unknown>
 ): MentionAttribute {
-	return Object.assign( { uid: uid() }, baseMentionData, data || {} );
+	return Object.assign( { uid: uid().slice( 0, 8 ) }, baseMentionData, data || {} );
```
oleq, line 97: Why just 8 characters? This significantly reduces the entropy of the UID and increases a chance of collision.

packages/ckeditor5-mention/docs/_snippets/examples/chat-with-mentions.js
```diff
@@ -171,11 +171,11 @@ function MentionLinks( editor ) {
 			return writer.createAttributeElement( 'a', {
 				class: 'mention',
 				'data-mention': modelAttributeValue.id,
-				href
+				href,
+				// Omit `data-uid` in clipboard (copy/cut) to prevent UIDs duplication.
+				...( !options.isClipboardPipeline && { 'data-uid': modelAttributeValue.uid } )
 			}, {
-				// Make mention attribute to be wrapped by other attribute elements.
 				priority: 20,
```
oleq, line 179: Why removing those comments? They may be useful to anyone working on it.

packages/ckeditor5-mention/docs/_snippets/features/mention-customization.js
```diff
@@ -101,11 +101,11 @@ function MentionCustomization( editor ) {
 				class: 'mention',
 				'data-mention': modelAttributeValue.id,
 				'data-user-id': modelAttributeValue.userId,
-				'href': modelAttributeValue.link
+				'href': modelAttributeValue.link,
+				// Omit `data-uid` in clipboard (copy/cut) to prevent UIDs duplication.
+				...( !options.isClipboardPipeline && { 'data-uid': modelAttributeValue.uid } )
 			}, {
-				// Make mention attribute to be wrapped by other attribute elements.
 				priority: 20,
```
oleq, line 109: Same: why remove these comments?

packages/ckeditor5-mention/docs/examples/chat-with-mentions.md
```diff
@@ -200,11 +200,11 @@ function MentionLinks( editor ) {
 			return writer.createAttributeElement( 'a', {
 				class: 'mention',
 				'data-mention': modelAttributeValue.id,
-				href
+				href,
+				// Omit `data-uid` in clipboard (copy/cut) to prevent UIDs duplication.
+				...( !options.isClipboardPipeline && { 'data-uid': modelAttributeValue.uid } )
 			}, {
-				// Make mention attribute to be wrapped by other attribute elements.
 				priority: 20,
```
oleq, line 208: Why remove comments that may be useful for integrators?

packages/ckeditor5-mention/src/mentionediting.ts
```diff
@@ -118,12 +118,14 @@ export function _toMentionAttribute(
 		return;
 	}
 
+	const dataUid = viewElementOrMention.getAttribute( 'data-uid' ) as string;
```
oleq: I'm a little bit apprehensive about using a generic `data-uid` name which may easily collide with integrator's code.

oleq (changes requested):
Out of curiosity, why did you ditch the second approach described in the issue:

> A different approach would be to remove UIDs entirely, but then, of course, the view attribute elements will get merged. To prevent that, we could create a new view element custom property, e.g. preventMerging:

It seems more elegant, at least at first glance.

packages/ckeditor5-mention/src/mentionediting.ts
scofalik: Good catch. This should be `data-mention-uid`.

packages/ckeditor5-mention/src/mentionediting.ts
scofalik: I think this is actually a lot and I was even thinking to make it smaller. Reasoning:

1) Mentions aren't very common element in one document. I'd not expect more than a few dozens mentions. UIDs should be unique in a scope of a one document, of course.
2) Aparat from UID, mentions are differentiated by other attributes too.
3) This mechanism is used only when two mentions are next to each other.

Although UIDs should be, well, unique, I believe 8 characters is more enough for our needs. And the bug (merging two mentions) would be reproducible only if you are unlucky enough to have two same mentions with the same UID generated, and next to each other.

I wanted to keep the UID small so it doesn't make too much mess in output data, and is easier to preserve (and not use too many tokens ;)) for AI service.

scofalik:
I made fixes and added changelog files. Ready for re-review, @oleq.

packages/ckeditor5-mention/src/mentionediting.ts
```diff
@@ -118,12 +118,14 @@ export function _toMentionAttribute(
...
 
+	const dataUid = viewElementOrMention.getAttribute( 'data-mention-uid' ) as string;
+
 	const baseMentionData = {
 		id: dataMention,
 		_text: textNode.data
 	};
 
-	return _addMentionAttributes( baseMentionData, data );
+	return _addMentionAttributes( baseMentionData, dataUid ? { uid: dataUid, ...data } : data );
```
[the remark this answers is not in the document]
scofalik: Low severity.

PR #20028, opened 2026-03-31
Introducing `config.root.modelElement`.

niegowski:
### 🚀 Summary

Allow passing `config.root.modelElement` etc. in the configuration to define the model root.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/20029

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-core/src/editor/utils/normalizerootsconfig.ts
+++ b/packages/ckeditor5-core/src/editor/utils/normalizerootsconfig.ts
@@ -208,6 +208,9 @@ export function normalizeRootsConfig(
 		rootConfig.placeholder ??= getLegacyPlainConfigValue( config, 'placeholder', rootName );
 		rootConfig.label ??= getLegacyPlainConfigValue( config, 'label', rootName );
 
+		// Default model element to `$root` so all callers can rely on `rootConfig.modelElement` being a string.
+		rootConfig.modelElement ||= '$root';
+
 		rootsConfig[ rootName ] = rootConfig;
 	}
 
--- a/packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
+++ b/packages/ckeditor5-editor-multi-root/src/multirooteditor.ts
@@ -168,7 +168,7 @@ export class MultiRootEditor extends Editor {
 
 		for ( const [ rootName, rootConfig ] of rootsConfig ) {
 			// Create root and `UIView` element for each editable container.
-			const root = this.model.document.createRoot( rootConfig.modelElement || '$root', rootName );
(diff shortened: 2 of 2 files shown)


Mati365 (dismissed):
Looks fine.

packages/ckeditor5-core/src/editor/editor.ts
```diff
@@ -868,6 +872,32 @@ export abstract class Editor extends /* #__PURE__ */ ObservableMixin() {
 				editor._showLicenseError( 'pluginNotAllowed', disallowedPlugin.pluginName );
 			}
 		}
+
+		function verifyRootElements( editor: Editor ): void {
+			const schema = editor.model.schema;
+
+			for ( const rootName of editor.model.document.getRootNames() ) {
+				const root = editor.model.document.getRoot( rootName )!;
```
mmotyczynska: Maybe it would be nice to add a short description of what this function actually verifies.

Also, since `schema` is extracted into a variable even though it's only used once, we could consider doing the same for `document` (used twice) - it might make the code a bit more consistent and easier to read.

packages/ckeditor5-horizontal-line/tests/horizontallinecommand.js
```diff
@@ -46,6 +46,26 @@ describe( 'HorizontalLineCommand', () => {
 			} );
 		} );
 
+		it( 'should be true when the selection directly in a root that has a non-$root element name', () => {
```
mmotyczynska:
```suggestion
		it( 'should be true when the selection is directly in a root that has a non-$root element name', () => {
```

packages/ckeditor5-image/tests/imageutils.js
```diff
@@ -397,6 +397,25 @@ describe( 'ImageUtils plugin', () => {
 			} );
 		} );
 
+		it( 'should return true when the selection directly in a root that has a non-$root element name', () => {
```
mmotyczynska:
```suggestion
		it( 'should return true when the selection is directly in a root that has a non-$root element name', () => {
```

packages/ckeditor5-page-break/tests/pagebreakcommand.js
```diff
@@ -46,6 +46,26 @@ describe( 'PageBreakCommand', () => {
 			} );
 		} );
 
+		it( 'should be true when the selection directly in a root that has a non-$root element name', () => {
```
mmotyczynska:
```suggestion
		it( 'should be true when the selection is directly in a root that has a non-$root element name', () => {
```

mmotyczynska (approved):

PR #20038, opened 2026-04-03
No longer throw unclear error when source element is not attached to DOM.

### LINKED ISSUE #20017: Verify DOM element attachment during editor creation
### Problem

When `ClassicEditor.create()` is called with a DOM element that is not attached to the document (i.e., has no `parentNode`), the editor crashes with a cryptic, unhelpful error:

```
  TypeError: Cannot read properties of null (reading 'insertBefore')
      at ElementReplacer.replace (elementreplacer.ts:38:1)
      at ClassicEditorUI.init (classiceditorui.ts:114:1)
      at classiceditor.ts:310:1
```

This happens because `ElementReplacer.replace()` (`ckeditor5-utils/src/elementreplacer.ts:38`) calls `element.parentNode!.insertBefore(...)` with a non-null assertion, but `parentNode` is `null` when the element is detached from the DOM tree.

This is a common integrator mistake — for example, calling `ClassicEditor.create()` before the element is rendered into the DOM, or passing an element that has been removed.

### Expected behavior

The editor should detect a detached element early and throw a clear `CKEditorError` with a descriptive error code, similar to the existing `editor-source-element-already-used` check in `secureSourceElement()` (`ckeditor5-core/src/editor/utils/securesourceelement.ts`).

  ### Scope

  - **ClassicEditor** — directly affected (uses `ElementReplacer` which calls `parentNode.insertBefore`)
  - **InlineEditor, BalloonEditor**, **DecoupledEditor**, **MultiRootEditor** — not affected by the `ElementReplacer` crash (they don't use it), but they would still benefit from an early validation

Consider using a warning instead of error to avoid breaking exotic integrations.

  ### Reproduction

  ```js
  // Element not in DOM
  const div = document.createElement( 'div' );
  ClassicEditor.create( div ); // TypeError: Cannot read properties of null (reading 'insertBefore')

  // Element removed before editor creation
  const el = document.querySelector( '#editor' );
  el.remove();
  ClassicEditor.create( el ); // Same error
```

### PR DESCRIPTION
**Pull request by Mati365**
### 🚀 Summary

No longer throw unclear error when source element is not attached to DOM.

---

### 📌 Related issues

* Closes #20017 

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-editor-classic/src/classiceditor.ts
+++ b/packages/ckeditor5-editor-classic/src/classiceditor.ts
@@ -85,7 +85,7 @@ export class ClassicEditor extends /* #__PURE__ */ ElementApiMixin( Editor ) {
 		this.config.define( 'menuBar.isVisible', false );
 
 		if ( isElement( sourceElement ) ) {
-			if ( !sourceElement.parentElement ) {
+			if ( !sourceElement.isConnected ) {
 				/**
 				 * Cannot initialize the editor because the provided source element is not attached to the DOM and cannot be replaced.
 				 *

### CODE REVIEW

File: packages/ckeditor5-editor-classic/src/classiceditor.ts [hunk 1]
```diff
@@ -84,6 +85,15 @@ export class ClassicEditor extends /* #__PURE__ */ ElementApiMixin( Editor ) {
 		this.config.define( 'menuBar.isVisible', false );
 
 		if ( isElement( sourceElement ) ) {
+			if ( !sourceElement.parentElement ) {
```
niegowski: Is this enough to verify this way? Shouldn't we check the `isConnected` property?

Btw, this was supposed to be handled in all editor types, not only in the classic editor.

packages/ckeditor5-editor-classic/src/classiceditor.ts
```diff
@@ -84,6 +85,15 @@ export class ClassicEditor extends /* #__PURE__ */ ElementApiMixin( Editor ) {
...
 
 		if ( isElement( sourceElement ) ) {
+			if ( !sourceElement.isConnected ) {
+				/**
+				 * Cannot initialize the editor because the provided source element is not attached to the DOM and cannot be replaced.
+				 *
+				 * @error editor-source-element-not-attached
+				 */
+				throw new CKEditorError( 'editor-source-element-not-attached', null );
+			}
```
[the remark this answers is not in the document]
Mati365: This comment is incorrect. The `ClassicEditor` replaces the provided element; however, the other types of editors attach to it. This means that the parent node is required in such a scenario. In other cases, it is not required.

packages/ckeditor5-editor-classic/src/classiceditor.ts
Mati365: https://github.com/ckeditor/ckeditor5/issues/20017#issuecomment-4182711578

PR #20066, opened 2026-04-13
Fix native undo/redo functionality in source editing mode

### LINKED ISSUE #13700: Do it can undo or redo in source editing mode?
is it possible do undo /redo in source editing?
it look like only can do it after close souce code mode,

_What steps should be taken to fulfill the task?_

## 📃 Other details

* Browser: …
* OS: …
* CKEditor version: …

jacek-cke:
### 🚀 Summary

When source editing mode is active, the editor's keystroke handler intercepts Ctrl/Cmd+Z/Y and blocks native undo/redo in the textarea. This PR adds a keydown listener that stops propagation of these keystrokes before they reach the editor, restoring native browser undo/redo.

---

### 📌 Related issues

* Closes #13700

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist


Reinmar:
Two findings:
* Cmd+Y does not redo (on macOS). CKEditor does that (we support both keystrokes:  Cmd+Y and Cmd+Shift+Z) so it may be slightly confusing that the former does not work in source area. We could try to plug a listener that tries to do `execCommand( 'redo' )` but only for macOS (because Ctrl+Y works already fine on Linux and Windows).
* When doing Cmd+Z on an empty undo stack, at least in Arc, this triggers a native Arc action (opens a previously closed tab). This might not be safely fixable because we'd need to do `preventDefault()` and that would stop native undo 🤷  And I'd avoid trying to use `execCommand()` here too (too unsafe).

Reinmar (changes requested):
Let's try to fix the Cmd+Y scenario.

packages/ckeditor5-source-editing/src/sourceediting.ts
```diff
@@ -266,6 +266,18 @@ export class SourceEditing extends Plugin {
 				editor.ui.update();
 			} );
 
+			// Allow native undo/redo in the textarea. The editor's keystroke handler (attached via
+			// setEditableElement()) intercepts Ctrl+Z/Y and calls preventDefault(), blocking browser
+			// undo/redo while editor commands are force-disabled in source editing mode.
+			// See: https://github.com/ckeditor/ckeditor5/issues/13700
+			domSourceEditingElementTextarea.addEventListener( 'keydown', evt => {
+				const key = evt.key.toLowerCase();
```
Reinmar, line 277: Missing a comment why it's needed here.

jacek-cke: Makes sense, added.

jacek-cke:
> Let's try to fix the Cmd+Y scenario.

Fixed. Wondering if we should also cover the iOS/iPadOS case where a user might have a keyboard connected?

Reinmar (approved):

Reinmar:
> > Let's try to fix the Cmd+Y scenario.
> 
> Fixed. Wondering if we should also cover the iOS/iPadOS case where a user might have a keyboard connected?

I think it's ok to keep the scope under control. There are dragons in adding more to it :D

PR #20071, opened 2026-04-14
Downcast skip-level lists to valid HTML

mmotyczynska:
### 🚀 Summary

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/9702

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-list/src/list/converters.ts
+++ b/packages/ckeditor5-list/src/list/converters.ts
@@ -370,7 +370,7 @@ export function listItemDowncastConverter(
 	attributeNames: Array<string>,
 	strategies: Array<ListDowncastStrategy>,
 	model: Model,
-	{ dataPipeline }: { dataPipeline?: boolean } = {}
+	{ dataPipeline, allowSkipLevels }: { dataPipeline?: boolean; allowSkipLevels?: boolean } = {}
 ): GetCallback<DowncastAttributeEvent<ListElement>> {
 	const consumer = createAttributesConsumer( attributeNames, strategies );
 
@@ -390,7 +390,8 @@ export function listItemDowncastConverter(
 
 		const options = {
 			...conversionApi.options,
-			dataPipeline
+			dataPipeline,
+			allowSkipLevels
 		};
 
 		// Use positions mapping instead of mapper.toViewElement( listItem ) to find outermost view element.
@@ -681,6 +682,10 @@ function unwrapListItemBlock( viewElement: ViewElement, viewWriter: ViewDowncast
 
 /**
  * Wraps the given list item with appropriate attribute elements for ul, ol, and li.
+ *
+ * For skip-level lists (where indent gaps exist, e.g. indent 0 → indent 2), this function
+ * generates intermediate wrapper pairs (ul/ol + li) at each missing level. These intermediate
+ * wrappers are invisible (`list-style-type: none` on the li) and carry no model-backed attributes.
  */
 function wrapListItemBlock(
 	listItem: ListElement,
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -694,40 +699,81 @@ function wrapListItemBlock(
...
+			// model element backing this level.
+			//
+			// The <li> uses a fixed ID per indent (`list-item-skip-N`) so that sibling items sharing
+			// the same skipped level merge into one <li> (e.g. two items at indent 1 with no parent
+			// at indent 0 share one intermediate <li> at indent 0).
+			//
+			// The <ul/ol> uses the default real ID (list-type-indent) so it can merge with real list
+			// elements from other items at the same indent (e.g. a skip-level item at indent 2 followed
+			// by a real item at indent 0 — their <ul> at indent 0 must merge into one list).
+			const listType = currentListItem.getAttribute( 'listType' );
```
mmotyczynska, line 724: Note: a [follow-up PR](https://github.com/ckeditor/ckeditor5/pull/20079) refines the intermediate wrapper type logic — it uses the type of a real sibling at that indent if one exists (so the intermediate merges with it), otherwise the ancestor's type (so all intermediates at that level agree). This fixes merging for both mixed-type siblings and deeply nested skip-level items.

packages/ckeditor5-list/src/list/listediting.ts
```diff
@@ -481,9 +481,14 @@ export class ListEditing extends Plugin {
...
 			.add( dispatcher => {
+				const allowSkipLevels = !!editor.config.get( 'list.allowSkipLevels' );
+
 				dispatcher.on<DowncastAttributeEvent<ListElement>>(
 					'attribute',
-					listItemDowncastConverter( attributeNames, this._downcastStrategies, model )
+					listItemDowncastConverter(
+						attributeNames, this._downcastStrategies, model,
+						allowSkipLevels ? { allowSkipLevels } : undefined
+					)
```
niegowski, line 491:
```suggestion
					listItemDowncastConverter(
						attributeNames,
						this._downcastStrategies,
						model,
						allowSkipLevels ? { allowSkipLevels } : undefined
					)
```
Or even maybe we could pass the options object without a condition, as the value itself carries the meaning.

packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -694,40 +699,81 @@ function wrapListItemBlock(
...
-					currentListItem.getAttribute( strategy.attributeName ),
-					strategy.scope == 'list' ? listViewElement : listItemViewElement,
-					options,
-					currentListItem
-				);
+		// When allowSkipLevels is enabled and ListWalker jumps over indent levels (e.g. from indent 2
+		// to indent 0), the levels in between have no corresponding model element. We detect these
+		// "intermediate" levels by checking if currentListItem's indent doesn't match the current
+		// loop indent.
+		const isIntermediate = allowSkipLevels && currentListItem.getAttribute( 'listIndent' ) !== indent;
```
niegowski, line 710: Do we need to know if the `allowSkipLevels` is enabled, or maybe we should always handle model data properly?

packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -694,40 +699,81 @@ function wrapListItemBlock(
...
-		currentListItem = ListWalker.first( currentListItem, { lowerIndent: true } );
-
-		// There is no list item with lower indent, this means this is a document fragment containing
-		// only a part of nested list (like copy to clipboard) so we don't need to try to wrap it further.
-		if ( !currentListItem ) {
-			break;
+		// Only look for a parent when at a real (non-intermediate) level. At intermediate levels
+		// currentListItem already points to the nearest found ancestor and won't change.
+		if ( !isIntermediate ) {
+			const nextListItem = ListWalker.first( currentListItem, { lowerIndent: true } );
```
niegowski, line 763: `nextListItem` is misleading, it should be parent list item or sth like this.

packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -694,40 +699,81 @@ function wrapListItemBlock(
...
+			if ( nextListItem ) {
+				currentListItem = nextListItem;
+			} else if ( !allowSkipLevels ) {
+				// There is no list item with lower indent, this means this is a document fragment
+				// containing only a part of nested list (like copy to clipboard) so we don't need
+				// to try to wrap it further.
+				//
+				// When allowSkipLevels is enabled, the loop continues to indent 0 producing
+				// intermediate wrappers at every remaining level to ensure a complete HTML structure
+				// regardless of whether the item has a preceding parent.
```
niegowski, line 774: Maybe we should check what cases this covers; maybe this would not be needed.

packages/ckeditor5-list/src/list/converters.ts
mmotyczynska: I checked this case again, and it seems we're safe to always handle the model data this way. Without skip-level lists, the command and post-fixer ensure that there are no intermediate wrappers anyway.

packages/ckeditor5-list/src/list/converters.ts
mmotyczynska: We need to create missing intermediate wrappers in the data/editing downcast for skip-level lists, so we can't "`break`" here and this change is required.

However, this also changes clipboard behavior for skip-level lists - it will start adding missing intermediate wrappers (towards indent level 0) for partial selection, unlike how lists behave today without skip levels.

I'd keep the current behavior for now and discuss whether we want to change the clipboard behavior during refinement.

packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -370,7 +370,7 @@ export function listItemDowncastConverter(
 	attributeNames: Array<string>,
 	strategies: Array<ListDowncastStrategy>,
 	model: Model,
-	{ dataPipeline }: { dataPipeline?: boolean } = {}
+	{ dataPipeline, allowSkipLevels }: { dataPipeline?: boolean; allowSkipLevels?: boolean }
```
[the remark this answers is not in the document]
mmotyczynska: This is an internal function, and we always pass the correct arguments. Therefore, there is no need to assign {} now.

packages/ckeditor5-list/tests/list/converters-data.js
```diff
@@ -2613,4 +2614,227 @@ describe( 'ListEditing - converters - data pipeline', () => {
...
+								'</ol>' +
+							'</li>' +
+							'<li>C</li>' +
+						'</ol>' +
+					'</li>' +
+				'</ol>'
+			);
+		} );
+
+		it( 'should handle a block widget at a skip level', () => {
```
niegowski:
```suggestion
		it( 'should handle a container at a skip level', () => {
```

packages/ckeditor5-list/tests/list/converters.js
```diff
@@ -1099,6 +1099,236 @@ describe( 'ListEditing - converters', () => {
...
+								'</ol>' +
+							'</li>' +
+							'<li><span class="ck-list-bogus-paragraph">C</span></li>' +
+						'</ol>' +
+					'</li>' +
+				'</ol>'
+			);
+		} );
+
+		it( 'should handle a block widget at a skip level', () => {
```
niegowski:
```suggestion
		it( 'should handle a container at a skip level', () => {
```

niegowski (approved):

packages/ckeditor5-list/src/list/converters.ts
mmotyczynska: Extracted to https://github.com/ckeditor/ckeditor5-commercial/issues/9777

PR #20072, opened 2026-04-14
Upcast skip level lists.

mmotyczynska:
### 🚀 Summary

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/9703

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist


packages/ckeditor5-list/src/list/listediting.ts
```diff
@@ -463,6 +465,12 @@ export class ListEditing extends Plugin {
 				converterPriority: 'high'
 			} )
 			.add( dispatcher => {
+				if ( allowSkipLevels ) {
+					dispatcher.on<UpcastElementEvent>(
+						'element:li', listItemSkipLevelUpcastConverter(), { priority: 'high' }
```
niegowski, line 470: Is it required to be "high" priority, as it is registered before the following converter? Or, as this is a consuming converter, is it needed for other features to know it in advance? Please add a code comment.

packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -58,6 +58,36 @@ import type {
...
+ * with `list-style-type:none` (generated by the skip-level downcast or external sources) and unwraps them
+ * so they don't produce empty list items in the model.
+ *
+ * The wrapper `<li>` is consumed without creating a model element, but its children (nested lists) are
+ * converted normally. Because `getIndent()` counts all ancestor `<li>` elements (including the consumed wrapper),
+ * the nested items receive the correct indent values that reflect the skip-level gap.
+ *
+ * @internal
+ */
+export function listItemSkipLevelUpcastConverter(): GetCallback<UpcastElementEvent> {
```
niegowski, line 72: Maybe we should rename it to indicate that it just consumes intermediate items.

packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -58,6 +58,36 @@ import type {
 	ListDowncastStrategy
 } from './listediting.js';
 
+/**
+ * Returns the upcast converter for skip-level list item wrappers. It detects intermediate `<li>` elements
+ * with `list-style-type:none` (generated by the skip-level downcast or external sources) and unwraps them
```
niegowski, line 63: I'd avoid "unwrap" as it has a different meaning in our engine.

packages/ckeditor5-list/src/list/listediting.ts
mmotyczynska: Technically, this isn't required - `listItemUpcastConverter` is registered right after this in the same `.add()` call with `normal` priority, so the registration order already ensures it runs first.

But I'd keep the `high` priority just in case other plugins register their own `element:li` upcast converters. Added a code comment to explain this.

packages/ckeditor5-list/tests/list/converters-data.js
```diff
@@ -2836,5 +2836,299 @@ describe( 'ListEditing - converters - data pipeline', () => {
...
+						'</li>' +
+					'</ul>'
+				);
+
+				expect( _getModelData( skipModel, { withoutSelection: true } ) ).to.equalMarkup(
+					'<paragraph listIndent="0" listItemId="a00" listType="bulleted">A</paragraph>'
+				);
+			} );
+
+			it( 'should not upcast single skip-level wrapper to the model and preserve indent gap', () => {
```
niegowski, line 2871: I'd suggest adding a test for:
```html
<ol>
    <li style="list-style-type:none;">
        <ol>
            <li style="list-style-type:none;">
                foobar
            </li>
        </ol>
    </li>
</ol>
```

mmotyczynska: Added.

niegowski (approved):

PR #20079, opened 2026-04-15
Fix skip levels in multi level and todo lists

mmotyczynska:
### 🚀 Summary

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/9731
* See https://github.com/ckeditor/ckeditor5-commercial/issues/9704

---

### 💡 Additional information

⚠️ **Commercial PR**: https://github.com/ckeditor/ckeditor5-commercial/pull/9740

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist


packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -747,23 +750,50 @@ function wrapListItemBlock(
...
+			// classes and styles as real list wrappers (e.g. multi-level-list class, todo-list class,
+			// list-style-type:none). Without this, intermediate and real list elements at the same
+			// indent would have mismatched attributes and could not merge correctly, causing browsers
+			// to render unwanted default markers.
+			//
+			// Use the reference item (sibling or ancestor) as the strategy source so the attributes
+			// match whatever this intermediate is supposed to merge with.
+			for ( const strategy of strategies ) {
+				if (
+					strategy.scope == 'list' &&
```
niegowski, line 784: The item strategy should most probably be applied too, as for example, GHS uses it.

packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -814,6 +844,31 @@ function wrapListItemBlock(
 	}
 }
 
+/**
+ * Walks forward from the given list item through model siblings to find the first list item block
+ * at exactly the specified indent level. Stops when it encounters a non-list block or a list item
+ * at a lower indent (which means we left the current subtree).
+ */
+function findSiblingListItemAt( listItem: ListElement, targetIndent: number ): ListElement | null {
```
niegowski, line 852: Maybe this could be replaced with `ListWalker`, but forward direction?

siyor11andraows:
/

packages/ckeditor5-list/src/list/converters.ts
mmotyczynska: I'm not sure about that. The intermediate wrapper has no representation in the model, so:

- there is no list item (model paragraph) we could read such information from
- there is no place to store this data back when setting the data

Adding a custom data attribute (via GHS) to an empty list item with `list-style-type: none` will not affect the model, and that's expected - we agreed that intermediate wrappers should have no representation in the model.

packages/ckeditor5-list/src/list/converters.ts
mmotyczynska: I couldn't find an efficient way to use `ListWalker` here. The current function is short and clear, so I suggest keeping it as is.

packages/ckeditor5-list/src/list/converters.ts
mmotyczynska: Added an issue to be discussed https://github.com/ckeditor/ckeditor5-commercial/issues/9778

packages/ckeditor5-list/tests/list/converters.js
```diff
@@ -1327,6 +1352,126 @@ describe( 'ListEditing - converters', () => {
...
+									'<li><span class="ck-list-bogus-paragraph">B</span></li>' +
+								'</ol>' +
+							'</li>' +
+						'</ol>' +
+					'</li>' +
+				'</ol>'
+			);
+
+			await editor.destroy();
+		} );
```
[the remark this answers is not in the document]
mmotyczynska: Not an issue. _getViewData strips data-list-item-id by default (skipListItemIds defaults to true), same as in the other tests in this block.

PR #20112, opened 2026-04-27
Backspace before a nested list item should not fail when there is no previous item to merge with (skip level list).

mmotyczynska:
### 🚀 Summary

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/9762.

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-list/src/list/listmergecommand.ts
+++ b/packages/ckeditor5-list/src/list/listmergecommand.ts
@@ -75,6 +75,13 @@ export class ListMergeCommand extends Command {
 		model.change( writer => {
 			const { firstElement, lastElement } = this._getMergeSubjectElements( selection, shouldMergeOnBlocksContentLevel );
 
+			// A defensive guard. When `_getMergeSubjectElements()` cannot determine a valid pair of elements
+			// to merge (for example because there is no list block before/after the current one),
+			// the command should be a no-op instead of throwing on `null.getAttribute()`.
+			if ( !firstElement || !lastElement ) {
+				return;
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-list/src/list/listmergecommand.ts
```diff
@@ -213,6 +220,12 @@ export class ListMergeCommand extends Command {
 					//    * b
 					//    c
 					firstElement = ListWalker.first( positionParent, { sameIndent: true, lowerIndent: true } );
+
+					// Fallback for "skip-level" structures where the previous list block in the model
+					// happens to live at a *higher* indent (no same-or-lower-indent block precedes it).
+					if ( !firstElement && isListItemBlock( positionParent.previousSibling ) ) {
+						firstElement = positionParent.previousSibling;
+					}
```
niegowski, line 228: Looking at this and keeping in mind the previous [PR](https://github.com/ckeditor/ckeditor5/pull/20111) makes me wonder if the check from the previous PR could be simplified to such a check as here?

niegowski (approved):

packages/ckeditor5-list/src/list/listmergecommand.ts
mmotyczynska: For future us: we checked this during the call and it turned out that this kind of simplification wouldn't be enough in practice, so we decided to keep the current approach.

PR #20116, opened 2026-04-28
 Modified `superscript` and `subscript` text styles to be mutually exclusive.

### LINKED ISSUE #20000: Superscript and subscript can be active at the same time (formatting conflict)
In the text formatting (e.g. RTE/editor), “Superscript” and “Subscript” can be active at the same time.
Enabling one function does not disable the other, although only one of these two formatting options should be applied to a given character/selection at a time.

Type some text  and select part of the text.
Click on “Superscript”.
Result: The selected text becomes superscript.
Without changing the selection, additionally click on “Subscript”.

### ✔️ Expected result

The new choice (Subscript) replaces the previous one (Superscript). The text is only subscript, not superscript anymore.
### ❌ Actual result

Both buttons can be active at the same time, leading to inconsistent formatting and potentially incorrect rendering in the editor/output.

### ❓ Possible solution

Perhaps a toogle could or a select field is possible

martnpaneq:
### 🚀 Summary

The `superscript` and `subscript` text styles are now mutually exclusive: applying one to text that already has the other replaces it, matching the behavior of common word processors. The previous behavior, where both attributes could coexist on the same text, can be restored by setting `config.basicStyles.superscript.allowNesting` or `config.basicStyles.subscript.allowNesting` to `true`, which is useful for content such as isotope notation or tensor indices.

---

### 📌 Related issues

* Closes #20000

---

### 💡 Additional information

https://github.com/user-attachments/assets/84aae06e-1d1a-44f0-ab1d-0164fe234169

---

### 🧾 Checklists

#### Author checklist

- [x] Is the changelog entry intentionally omitted? (Changelog entry included.)
- [x] Is the change backward-compatible? (No, but it is treated as a fix. It changes the behaviour of `superscript` and `subscript` to be aligned with other solutions on the market.)
- [x] Have you considered the impact on different editor setups and core interactions? _(e.g., classic/inline/multi-root/many editors, typing, selection, paste, tables, lists, images, collaboration, pagination)_ (Should not impact other editor setups in unpredicted way.)
- [x] Has the change been manually verified in the relevant setups? (Verified in a manual test.)
- [x] Does this change affect any of the above? (Not relevant.)
- [x] Is performance impacted? (Performance not impacted.)
- [x] Is accessibility affected? (Accessibility not affected.)
- [x] Have tests been added that fail without this change (against regression)? (Appropriate tests added.)
- [x] Have the API documentation, guides, feature digest, and related feature sections been updated where needed? (Documentation updated.)
- [x] Have metadata files (ckeditor5-metadata.json) been updated if needed? (Not relevant.)
- [x] Are there any changes the team should be informed about (e.g. architectural, difficult to revert in future versions or having impact on other features)? (This change does not have a big impact.)

#### Reviewer checklist


packages/ckeditor5-basic-styles/src/augmentation.ts
```diff
@@ -28,6 +32,23 @@ import type {
...
+		 */
+		superscript?: BasicStyleSuperscriptConfig;
+
+		/**
+		 * The configuration of the {@link module:basic-styles/subscript~Subscript subscript feature}.
+		 *
+		 * Read more in {@link module:basic-styles/subscriptconfig~BasicStyleSubscriptConfig}.
+		 */
+		subscript?: BasicStyleSubscriptConfig;
+	}
```
pomek, line 40: These new options appear at the _root_ level, next to `plugins`, etc. I wonder if they could be wrapped into a namespace, e.g., `basicStyles`.

packages/ckeditor5-basic-styles/src/mutuallyexclusiveattributecommand.ts
```diff
@@ -0,0 +1,91 @@
...
+		} );
+	}
+}
+
+function _isNestingAllowed( editor: Editor ): boolean {
+	return Boolean(
+		editor.config.get( 'superscript.allowNesting' ) ||
+		editor.config.get( 'subscript.allowNesting' )
+	);
+}
```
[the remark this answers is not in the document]
martnpaneq: Resolved in [3717685](https://github.com/ckeditor/ckeditor5/pull/20116/commits/371768592079a9e935e18c60b969c398fa6f7177).

packages/ckeditor5-basic-styles/src/subscript/subscriptcommand.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @module basic-styles/subscript/subscriptcommand
+ */
+
+import type { Editor } from '@ckeditor/ckeditor5-core';
+import { ModelDocumentSelection, type ModelElement, type ModelRange } from '@ckeditor/ckeditor5-engine';
+
+import { AttributeCommand } from '../attributecommand.js';
+
+const SUBSCRIPT = 'subscript';
+const SUPERSCRIPT = 'superscript';
```
jacek-cke: Could we move these to a shared file, since they’re also declared in other files (e.g., `subscriptediting.ts`, `superscriptcommand.ts`)?

.changelog/20260428123311_ck_20000.md
jacek-cke, line 1: nit: It may not be important, but I believe the changelog filename format is `YYYYMMDDHHMMSS_{prefix}_{issue}.md`.

packages/ckeditor5-basic-styles/src/augmentation.ts
martnpaneq: Introduced a `basicStyles` namespace in [518e4a0](https://github.com/ckeditor/ckeditor5/pull/20116/commits/518e4a0c53782db290817751185899693cc073f6).

packages/ckeditor5-basic-styles/src/subscript/subscriptcommand.ts
martnpaneq: Extracted to a common place in [b921abc](https://github.com/ckeditor/ckeditor5/pull/20116/commits/b921abc1eb11c7d80caabee1bcf25d5c4ccaf093).

.changelog/20260428123311_ck_20000.md
martnpaneq: Updated in [32a625e](https://github.com/ckeditor/ckeditor5/pull/20116/commits/32a625e79c53bdf951bad9252c69a945299d5a7d).

.changelog/20260428123311_ck_20000.md
```diff
@@ -0,0 +1,7 @@
+---
+type: Fix
+scope:
+  - ckeditor5-basic-styles
+---
+
+The `superscript` and `subscript` text styles are now mutually exclusive.
```
pomek: It would be good to mention that this is not the default behavior. A user can configure the editor to work this way. And based on that, it results in being more of a feature than a fix, but I am not forcing the type change.

martnpaneq: Actually it is the default behaviour now. If you think we should make it not default - please let me know.

martnpaneq: This comment suggests it should be the default: https://github.com/ckeditor/ckeditor5/issues/20000#issuecomment-4132659783.

martnpaneq: Updated the changeset content to be more descriptive in [71425c2](https://github.com/ckeditor/ckeditor5/pull/20116/commits/71425c2e08d2cba6d615b15133d7b9f9c8ccc538).

pomek: Fair point, thanks.

jacek-cke (approved):
LGTM

pomek:
Edge case, but what would happen when both the subscript and superscript attributes are defined in the data on the editor load?

```html
<p>
  Test: <sup><sub>2</sub></sup>.
</p>
```

I am wondering about the alternative approach:

Use a schema to control both attributes. This way, we do not have to provide additional configuration options. Instead, our docs can share a code snippet that would turn off this check.

By adding a postfixer, this approach resolves the potential issue with _invalid_ HTML when creating the editor.

One caveat: we need to decide which attribute is more important. For the following conflict: `<sup><sub>2</sub></sup>`, do we expect that editor keeps

- `<sup>2</sup>`, or
- `<sub>2</sub>`

Maybe we can solve it at the upcast level, but I am not sure.

PR #20139, opened 2026-05-12
Improve detecting skip-level lists

mmotyczynska:
### 🚀 Summary

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/9778

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-list/src/list/converters.ts
+++ b/packages/ckeditor5-list/src/list/converters.ts
@@ -98,9 +98,9 @@ export function listItemSkipLevelConsumer(): GetCallback<UpcastElementEvent> {
 }
 
 /**
- * Checks whether a `<li>` view element is a skip-level intermediate wrapper, i.e. its only meaningful content
- * is a nested `<ul>`/`<ol>`. Whitespace-only text nodes are ignored. Any other content (real text, `<br>`,
- * `<p>`, custom elements, etc.) disqualifies the element, so it is upcast as a regular list item.
+ * Checks whether a `<li>` view element is a skip-level intermediate wrapper, i.e. its only child is a nested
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-list/src/list/converters.ts
```diff
@@ -88,6 +97,26 @@ export function listItemSkipLevelConsumer(): GetCallback<UpcastElementEvent> {
...
+		if ( child.is( 'element', 'ul' ) || child.is( 'element', 'ol' ) ) {
+			hasNestedList = true;
+			continue;
+		}
+
+		return false;
+	}
+
+	return hasNestedList;
+}
```
Mati365, line 118: This might incorrectly identify the list wrapper and strip its additional attributes or class names. See the example below:

```html
<li style="list-style-type:none;" data-foo="foo" class="some-class">
  <ul>
    <li>
    ...
```

It’s a rare edge case, but it’s worth considering if we want to maintain full attribute support.

Mati365:
On second thought, despite the concerns I raised [here](https://github.com/ckeditor/ckeditor5/pull/20139#discussion_r3234233005), let's try not to overcomplicate the logic. It might be better to stick with a simpler solution for now.

Mati365 (approved):

packages/ckeditor5-list/src/list/converters.ts
mmotyczynska: I agree, but we need to propose a compromise solution that is not too aggressive while still being reasonably logical and covering the most common use cases.

Determining when a list item is an actual list item and when it is only a skip-level wrapper is not that straightforward, so we need to make a clear decision here as well.

PR #20150, opened 2026-05-14
Fix missing `This link has no URL` label on links with an empty href.

### LINKED ISSUE #20136: Link preview button shows empty label when the selected link has no href (regression vs v41)
The link toolbar's preview button (LinkPreviewButtonView, class ck-link-toolbar__preview) renders an empty, disabled <a> with an empty <span class="ck-button__label"> whenever the selected link has no href. In
  v41 the same button showed the localized message "This link has no URL"; in v45.2.2 both the fallback binding and the source string have been removed from the build, so neither English source nor any translation is available.

  1. Use any v45.x CKEditor 5 build with the Link plugin enabled.
  2. Insert an <a> element with no href into the editor content — for example, paste HTML containing <a>Click me</a>, or have an upstream filter/style produce a link without a URL.
  3. Place the cursor inside that link.
  4. Observe the link toolbar balloon.

### ✔️ Expected result

  The first item in the toolbar shows the text "This link has no URL" (or its localized equivalent), as in v41.

### ❌ Actual result

The first item is an empty, disabled <a class="ck ck-button ck-disabled ck-off ck-button_with-text ck-link-toolbar__preview" target="_blank" rel="noopener noreferrer" aria-disabled="true"> containing an empty
  <span class="ck ck-button__label">. The DOM has no text content — this is not a CSS / translation gap; the source string is no longer emitted.
### Root cause
  In packages/ckeditor5-link/src/linkui.ts, the linkPreview component factory's setHref helper explicitly clears the label when href is falsy, with no fallback:
```
  const setHref = ( href: string | undefined ) => {
    if ( !href ) {
      button.label = undefined;   // ← regression
      button.icon = undefined;
      button.tooltip = t( 'Open link in new tab' );
      return;
    }
    // ...
  };
```
  For comparison, in v41 the same view was wired with a bind('label') callback that always provided a source string for the no-href case:
```
button.bind( 'label' ).to( this, 'href', href => href || t( 'This link has no URL' ) )
button.bind( 'isEnabled' ).to( this, 'href', href => !!href )
```
Both halves of the regression are observable in the published build:

  - t( 'This link has no URL' ) call is gone from linkui.ts
  - The source string "This link has no URL" is no longer present anywhere in ckeditor5-link/build/translations/*.js, so existing Transifex translations are stranded

### ❓ Possible solution
  Restore the fallback in setHref:
```
   const setHref = ( href: string | undefined ) => {
     if ( !href ) {
  -    button.label = undefined;
  +    button.label = t( 'This link has no URL' );
       button.icon = undefined;
       button.tooltip = t( 'Open link in new tab' );
       return;
     }
```
This re-emits the source string for the translation pipeline and restores parity with v41. Happy to open a PR with the fix + a unit test in tests/linkui.js asserting button.label after linkCommand.value changes to an empty string — let me know if you'd prefer that.
## 📃 Other details

  - Confirmed broken in: v45.2.2 (bundled with Drupal core 10.6.7)
  - Last confirmed working: v41 (bundled with Drupal core 10.4.9)
  - Regression introduced: likely v44 (the LinkActionsView → LinkPreviewButtonView refactor); haven't bisected exactly
  - Browser / OS: not browser-specific (the source string is missing from the build)

Mati365:
### 🚀 Summary

Fix missing `This link has no URL` label on links with an empty href.

---

### 📌 Related issues

* Closes #20136 

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/.changelog/20260514133321_ck_20136.md
+++ b/.changelog/20260514133321_ck_20136.md
@@ -1,0 +1,9 @@
+---
+type: Fix
+scope:
+  - ckeditor5-link
+closes:
+  - https://github.com/ckeditor/ckeditor5/issues/20136
+---
+
+The link preview button for links with an empty href now correctly displays a no url label.
--- a/.changelog/20260514134321_ck_20136.md
+++ b/.changelog/20260514134321_ck_20136.md
@@ -1,0 +1,9 @@
+---
+type: Fix
+scope:
+  - ckeditor5-link
+see:
+  - https://github.com/ckeditor/ckeditor5/issues/20136
+---
+
+The link edit form back button was hidden when editing a link with an empty URL via the balloon toolbar.
--- a/.changelog/20260514140028_ck_20136.md
+++ b/.changelog/20260514140028_ck_20136.md
@@ -6,4 +6,4 @@ see:
   - https://github.com/ckeditor/ckeditor5/issues/20136
 ---
 
-Text centering in the link preview button.
+Fix text centering in the link preview button.
--- a/.changelog/20260514140028_ck_20136.md
+++ b/.changelog/20260514140028_ck_20136.md
@@ -1,0 +1,9 @@
+---
+type: Fix
+scope:
+  - ckeditor5-link
+see:
+  - https://github.com/ckeditor/ckeditor5/issues/20136
+---
+
+Text centering in the link preview button.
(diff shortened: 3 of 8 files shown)


pszczesniak (changes requested):
I think that we should also remove the tooltip completely when there is no url.

Image

Other than that :+1:

pszczesniak (approved):

PR #20153, opened 2026-05-14
Fix duplicated wrap text and break text translation contexts

jacek-cke:
### 🚀 Summary

Move shared "Wrap text" and "Break text" translation contexts to core.

The new media embed alignment feature reuses the "Wrap text" and "Break text" source strings already declared by ckeditor5-image, which made `translations:validate` fail with a "Duplicated context" error. Since the validator only accepts a context from the message's own package or from ckeditor5-core, the keys are moved to the core package. 

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist

### DIFF
--- a/packages/ckeditor5-core/lang/contexts.json
+++ b/packages/ckeditor5-core/lang/contexts.json
@@ -1,6 +1,8 @@
 {
 	"Cancel": "Label for the Cancel button.",
 	"Clear": "Label for the Clear button.",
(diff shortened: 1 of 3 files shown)


mmotyczynska (approved):

pomek:
Could we please also move existing translations to avoid generating additional costs?

See: https://www.notion.so/Internationalization-a-developer-perspective-f27a39c4ec234fe98b5aab5fad5269fa?source=copy_link#26bbb9a67232805aa1f3cb7061c24bf6.

PR #20162, opened 2026-05-16
[codex] Keep external drop effect as copy

### LINKED ISSUE #20161: External drag/drop can force move dropEffect in Chromium
## Summary

Dragging external content into CKEditor 5 in Chromium-based browsers can be negotiated as a `move` operation instead of `copy`.

This is especially dangerous when the drag source is Outlook Classic: Outlook may treat the completed drop as a move and remove the original email from the mailbox. Firefox does not appear to hit the same path.

## Relevant code

In `packages/ckeditor5-clipboard/src/dragdrop.ts`, the `dragging` handler first detects an external drag and sets the operation to copy:

```ts
if ( !this._draggedRange ) {
	data.dataTransfer.dropEffect = 'copy';
}
```

But immediately after that, the non-Gecko path can overwrite it:

```ts
if ( !env.isGecko ) {
	if ( data.dataTransfer.effectAllowed == 'copy' ) {
		data.dataTransfer.dropEffect = 'copy';
	} else if ( [ 'all', 'copyMove' ].includes( data.dataTransfer.effectAllowed ) ) {
		data.dataTransfer.dropEffect = 'move';
	}
}
```

So for external drags where `effectAllowed` is `all` or `copyMove`, Chrome/Edge can end up with:

```ts
dropEffect = 'move'
```

even though the previous branch identified the drag as external and set `copy`.

## Why this seems wrong

`!this._draggedRange` means the drag did not originate from this editor instance. That includes content dragged from native apps such as Outlook or from the file manager.

For external content, CKEditor should not request/advertise a move operation. A web editor generally cannot safely move native source data. It should either:

- accept the drop as `copy`, or
- reject it with `none`.

Advertising `move` can cause native drag sources to remove/move their original item after the drop succeeds.

## Observed behavior

Environment where this was observed:

- Outlook Classic as the drag source
- CKEditor 5 embedded in a third-party app
- Edge/Chrome: dragging an email into the rich text editor can move/delete the source email
- Firefox: the same workflow does not delete the source email

The browser difference matches the code path because Firefox skips the overwrite via `env.isGecko`.

## Expected behavior

For external drags (`!this._draggedRange`), CKEditor should preserve `dropEffect = 'copy'` in Chrome/Edge instead of letting the non-Gecko fallback rewrite it to `move`.

Internal editor drags can still use move semantics where appropriate.

## Notes

`ClipboardObserver` maps native `dragover` to CKEditor's `dragging` event and calls `preventDefault()`, so this code participates in the native browser drag/drop negotiation.

There is also related final-effect logic in `getFinalDropEffect()`:

```ts
return [ 'all', 'copyMove' ].includes( dataTransfer.effectAllowed ) ? 'move' : 'copy';
```

but the highest-risk part is the `dragging` handler setting the native `dropEffect` to `move` during `dragover`.

HardyPotato:
## Summary

Keeps external drag operations advertised as `copy` in the clipboard drag/drop plugin.

## Root cause

The `dragging` handler already detects external drags with `!this._draggedRange` and sets `dataTransfer.dropEffect = 'copy'`.

In non-Gecko browsers, the next branch could immediately overwrite that value to `move` when `effectAllowed` was `all` or `copyMove`. That made external drags behave like editor-internal moves in Chrome/Edge.

## Change

Only run the non-Gecko `effectAllowed` fallback for internal editor drags. External drags keep the explicit `copy` drop effect.

Fixes ckeditor/ckeditor5#20161.

### DIFF
--- a/packages/ckeditor5-clipboard/src/dragdrop.ts
+++ b/packages/ckeditor5-clipboard/src/dragdrop.ts
@@ -376,10 +376,8 @@ export class DragDrop extends Plugin {
 			// is not possible until 'dragend' event case will be fixed.
 			if ( !this._draggedRange ) {
 				data.dataTransfer.dropEffect = 'copy';
-			}
-
-			// In Firefox it is already set and effect allowed remains the same as originally set.
-			if ( !env.isGecko ) {
+			} else if ( !env.isGecko ) {
+				// In Firefox it is already set and effect allowed remains the same as originally set.
 				if ( data.dataTransfer.effectAllowed == 'copy' ) {
 					data.dataTransfer.dropEffect = 'copy';
 				} else if ( [ 'all', 'copyMove' ].includes( data.dataTransfer.effectAllowed ) ) {


HardyPotato:
Superseded by #20180. The original PR had an incorrect commit author email, which kept the CLA check tied to the wrong identity.

PR #20164, opened 2026-05-17
Skipl-level list PFO integration.

mmotyczynska:
### 🚀 Summary

Apart from the PFO integration in this PR I also fixed a bug:

In `applyListItemMarginLeftAndUpdateTopLevelInfo` (the helper that converts Word's absolute `margin-left` into the editor's relative one).

This function only subtracted the immediate parent `<li>`'s `margin-left` when converting Word's absolute indent to the editor's relative one. For 3+ levels deep, every ancestor margin cumulates via default `<ol>` padding, so non-zero margins on grandparents (or higher) were silently ignored and descendants drifted right.
Triggers only when a Word document has 3+ list levels AND at least one non-immediate ancestor has explicit indentation. Typical pastes are unaffected.

Fix: sum the relative `margin-left` from every ancestor stack frame, not just the immediate parent. Skip-level intermediate wrappers contribute 0 and drop out naturally.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/9964.
* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/10014.
* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/9965.

---

### 💡 Additional information

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist


Mati365:
I checked it for this document:  [incorrect-list.docx](https://github.com/user-attachments/files/27954842/incorrect-list.docx)

It looks like paste from office breaks structure of more complex skip level-list. What's interesting is that it's not happening when Paste from Office plugin is disabled. Can you take a look?

PFO **Off**:

obraz

PFO **On**:

obraz

Mati365 (changes requested):
Let's mark it as request changes for now.

packages/ckeditor5-list/src/list/listediting.ts
```diff
@@ -465,6 +465,56 @@ export class ListEditing extends Plugin {
...
+						return;
+					}
+
+					// Empty <p> (truly empty, e.g. `<li><p></p></li>`) must go through the standard
+					// `elementToElement('p')` path so the EmptyBlock upcast (priority 'lowest') takes care of it.
+					if ( viewElement.isEmpty ) {
+						return;
+					}
+
+					// Bogus only when <p> carries no own attributes…
```
pszczesniak, line 486:
```suggestion
					// Bogus only when `<p>` carries no own attributes.
```

We do not use this `…` in comments, so lets do not break that convention :)

packages/ckeditor5-paste-from-office/src/filters/list.ts
```diff
@@ -38,7 +38,8 @@ import {
 export function transformListItemLikeElementsIntoLists(
 	documentFragment: ViewDocumentFragment,
 	stylesString: string,
-	hasMultiLevelListPlugin: boolean
+	hasMultiLevelListPlugin: boolean,
+	allowSkipLevels: boolean = false
```
pszczesniak, line 42: Please add those params to API docs.

pszczesniak (changes requested):
There are two things that would be good to update.

Other than that, 👍 works as expected.

mmotyczynska:
Moved to https://github.com/ckeditor/ckeditor5-commercial/pull/10056.

PR #20192, opened 2026-06-12
Fix (mention): The mention suggestions panel no longer overflows the viewport on narrow screens

### LINKED ISSUE #20182: Mention suggestions dropdown overflows viewport on mobile screen sizes

1. Open https://ckeditor.com/docs/ckeditor5/latest/examples/framework/chat-with-mentions.html
2. Resize browser to mobile width (~390px) or open on a real mobile device
3. Click into the editor and type `@`
4. Observe the mention suggestions dropdown position

### ✔️ Expected result

The dropdown should stay within the visible viewport bounds and not overflow outside the screen.

### ❌ Actual result

The mention suggestions dropdown partially overflows outside the viewport — 
the list extends beyond the screen edge, cutting off suggestion items and 
making them unreachable without scrolling.

### ❓ Possible solution

The dropdown position should be recalculated when it would overflow the viewport, 
flipping or repositioning it to stay within visible bounds (similar to how 
most tooltip/popover libraries handle edge collision).

## 📃 Other details

* Browser: Safari / Chrome
* OS: iOS
* First affected CKEditor version: 48.1.1
* Installed CKEditor plugins: Mentions (ckeditor5-premium-features 48.1.1)

## 🎥 Screen recording

https://github.com/user-attachments/assets/909fd70e-3bc0-4371-8913-393721b0e42f

ELHart05:
Closes #20182.

On narrow screens the mention suggestions panel could stick out of the viewport. All four balloon positions anchor the panel to the caret, so when a wide panel did not fit on either side of the caret, whichever position got picked still overflowed the screen and the suggestions were cut off.

The position callbacks now clamp the panel's left coordinate to the viewport rect, which `getOptimalPosition()` already passes to positioning functions. If the panel would stick out, it is shifted just enough to stay fully visible. Vertical behavior is unchanged.

Tested on the `mention-custom-renderer` manual test at a 390px viewport. Before, the panel rect ended at 414px with its right side cut off. After, it ends right at the 390px edge:

Before and after at a 390px viewport

The existing position tests were updated because the callbacks now take the viewport rect as a third argument, and a new test covers the clamping on both screen edges. The `ckeditor5-mention` suite passes with 100% coverage. A changelog entry is included.

### DIFF
--- a/.changelog/20260612135133_i_20182.md
+++ b/.changelog/20260612135133_i_20182.md
@@ -1,0 +1,9 @@
+---
+type: Fix
+scope:
+  - ckeditor5-mention
+closes:
+  - https://github.com/ckeditor/ckeditor5/issues/20182
+---
+
+The mention suggestions panel no longer overflows the viewport on narrow screens. When the caret is close to the screen edge, the panel is shifted horizontally so the whole list stays visible.
--- a/packages/ckeditor5-mention/src/mentionui.ts
+++ b/packages/ckeditor5-mention/src/mentionui.ts
@@ -601,10 +601,10 @@ function getBalloonPanelPositions(
 ): DomOptimalPositionOptions['positions'] {
 	const positions: Record<string, DomOptimalPositionOptions['positions'][0]> = {
 		// Positions the panel to the southeast of the caret rectangle.
-		'caret_se': ( targetRect: Rect ) => {
+		'caret_se': ( targetRect: Rect, balloonRect: Rect, viewportRect: Rect ) => {
 			return {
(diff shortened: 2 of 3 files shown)


ELHart05:
Already signed the CLA...

Reinmar:
Thanks for the PRs! We're on them. They look really solid and we'll let you know once our QA team finishes assessing them.

ELHart05:
Just flagging that this fix already landed on master as 674d3cb74f, so this PR can be closed. Thanks!

jacek-cke:
Closing the PR then. Thanks for the contribution!

PR #20193, opened 2026-06-12
Fix (widget): Type around buttons now insert a paragraph when tapped on touch devices

### LINKED ISSUE #20103: Keyboard opens when clicked on images and adding newlines not working

Use any demo on the cke5 website. F.e. this one: https://ckeditor.com/ckeditor-5/demo/merge-fields/#demo
Use smartphone and click on the image. 

### ✔️ Expected result
The keyboard stays closed.
I can tap the blue "add newline" buttons above and below the image. 

### ❌ Actual result
The keyboard opens and tapping the buttons does nothing.

* Browser: Vivaldi(Chrome) 
* OS: Android 16

Image

ELHart05:
Closes #20103.

On Android and iOS, tapping the type around buttons (the blue insert paragraph buttons above and below a selected widget) did nothing. The buttons are activated by a `mousedown` listener, but `Widget._onPointerdown()` calls `preventDefault()` on `pointerdown` for taps that resolve to a widget, and the buttons live inside the widget DOM. Canceling `pointerdown` suppresses the compatibility `mousedown` for touch input, so the button handler never ran and the tap only kept the widget selected.

`_onPointerdown()` now returns early when the tap lands on a type around button. The `pointerdown` event is not canceled, the browser dispatches the compatibility `mousedown`, and the existing button handler inserts the paragraph and puts the caret in it. The mouse path is unaffected because the early return only skips selecting the widget, which the button's `mousedown` handler was already stopping.

Verified with emulated touch (Chrome, Android user agent, 390px viewport) on the `type-around` manual test. Before, tapping the button changed nothing. After, it inserts a paragraph before the image and moves the caret there:

Before and after tapping the insert paragraph button on touch

A regression test is added next to the existing Android and iOS pointerdown tests, and the `ckeditor5-widget` suite passes with 100% coverage. A changelog entry is included.

One note on scope: the issue also mentions that the keyboard opens when tapping the image itself. That comes from the editor focusing the editable on widget selection and is unchanged here. Once the buttons work, tapping one inserts the paragraph and the keyboard opens with the caret in text, which is the expected flow on mobile.


Mati365:
Hello! We merged fix in https://github.com/ckeditor/ckeditor5/commit/048e75cd6c8492836e156c86a6882879355232bd. It'll be part of current release. Thanks!

PR #20239, opened 2026-09-16
Fix createEventNamespace to handle Safari's miscompilation of substr …

### LINKED ISSUE #20237: [Safari 27.0][Intel Macs] CKEditor 5 fails to initialize with "Maximum call stack size exceeded"

1. Open Safari Version 27.0 (21625.1.29.18.28).
2. Navigate to https://ckeditor.com/ckeditor-5/demo/feature-rich/.
3. Wait for the editor to initialize. The editor does not load.

The issue also occurs across a local Drupal site using CKEditor 5, including regular content forms and nested Paragraphs fields.

### ✔️ Expected result

The editor initializes, displays its toolbar and editable content, and allows typing.

### ❌ Actual result

The editors on the official demo do not load in this Safari version.

On the Drupal site, editor initialization fails with:
RangeError: Maximum call stack size exceeded.

Drupal logs the error in the .catch() handler following ClassicEditor.create(). The underlying failing function has not yet been identified.

### ❓ Possible solution

A Safari/WebKit compatibility regression is suspected because the official CKEditor demo also fails, independently of Drupal.

The local Drupal form initialized successfully in Chromium and Playwright WebKit 26.5. These tests did not reproduce the failure seen in
Safari 27.0. No confirmed workaround or fix has been identified for the affected Safari build.

## 📃 Other details

- Browser: Safari Version 27.0 (21625.1.29.18.28).
- OS: macOS; exact version to be added.
- First affected CKEditor version: Unknown. Reproduced locally with CKEditor 5 v47.6.2; the official demo’s version has not been verified.
- Installed CKEditor plugins: The official feature-rich demo’s default plugins. The local Drupal site also reproduces the issue with its
custom editor plugin disabled.

- Extension interference: Not yet ruled out.
- Full error stack from the official demo: To be added.

ld3nl:
Fix (utils): Avoid `substr` with negative length in `createEventNamespace`

### 🚀 Summary

Stops Safari 27 on Intel Macs from failing editor init with `RangeError: Maximum call stack size exceeded`.

`createEventNamespace()` truncated event names with `name.substr( 0, name.lastIndexOf( ':' ) )`. When there is no `:`, `lastIndexOf` is `-1`. Some JSC builds (Safari 27 x86_64) miscompile `substr` with a negative length and return the original string instead of `''`. That can add a self-edge in `childEvents`, so `getCallbacksListsForNamespace()` recurses until the stack overflows.

This PR uses the same `lastIndexOf` + `substring` pattern already used later in the file (never calls `substr` with a negative length), and skips pushing a `childEvents` self-edge.

---

### 📌 Related issues

* Closes #20237

---

### 💡 Additional information

Implementation requested in https://github.com/ckeditor/ckeditor5/issues/20237#issuecomment-5704157499 — the `emittermixin.ts` workaround that loaded the all-features manual test on the affected Intel Safari 27 machine.

- Replace `substr( 0, lastIndexOf( ':' ) )` with `colonIndex > -1 ? name.substring( 0, colonIndex ) : ''`.
- Guard `events[ name ].childEvents.push( … )` with `childEventName && childEventName !== name`.

Verified on Safari 27.0 (21625.1.29.18.28), Intel x86_64, `pnpm run manual` → `packages/ckeditor5/manual/all-features.manual.html`. Without the change the editor does not load (unbounded recursion in `getCallbacksListsForNamespace`). With it, the editor initializes.

Changelog entry: `.changelog/20260916210000_i_20237.md`.

No tests added in this basic PR. A unit test cannot reproduce the JSC miscompile; the self-edge guard is the defensive part.

---

### 🧾 Checklists

#### Author checklist

#### Reviewer checklist


packages/ckeditor5-utils/src/emittermixin.ts
```diff
@@ -792,8 +792,13 @@ function createEventNamespace( source: EmitterInternal, eventName: string ): voi
...
 		childEventName = name;
-		// If `.lastIndexOf()` returns -1, `.substr()` will return '' which will break the loop.
-		name = name.substr( 0, name.lastIndexOf( ':' ) );
+
+		// Truncate to the parent namespace. Do not use `substr( 0, lastIndexOf( ':' ) )`:
+		// when there is no `:`, `lastIndexOf` is -1 and some JSC builds (Safari 27 x86_64)
+		// miscompile `substr` with a negative length, returning the original string instead of `''`.
+		const colonIndex = name.lastIndexOf( ':' );
+
+		name = colonIndex > -1 ? name.substring( 0, colonIndex ) : '';
```
pomek, line 801: Consider adding a test:

```js
// See https://github.com/ckeditor/ckeditor5/issues/20237.
it( 'should not depend on String#substr() clamping a negative length', () => {
      const originalSubstr = String.prototype.substr;

      // Some JSC builds (Safari 27 on x86_64) miscompile `substr()` with a negative length
      // and return the whole string instead of an empty one. Event names without a colon
      // made the namespace node reference itself and recursion never ended.
      vi.spyOn( String.prototype, 'substr' ).mockImplementation( function( start, length ) {
              return length < 0 ? String( this ) : originalSubstr.call( this, start, length );
      } );

      const spy = vi.fn();

      emitter.on( 'foo', spy );
      emitter.fire( 'foo' );

      expect( spy ).toHaveBeenCalledOnce();
} );
```

ld3nl: > Thanks, added the regression test as suggested. It emulates the broken String#substr() behavior and verifies that firing a non-namespaced
  > event terminates and calls the listener once. All ckeditor5-utils tests pass.

pomek:
See: 6fd7d99a70ed8e220446db10b408e9e5248e8a77.

